PR #366, opened 2017-05-23
[CSL-1002] Try gold linker for macOS

ChShersh:


deepfire:
@ChShersh, by the way, what is the motivation for this?

Wasn't the consensus that `gold` produces worse results than `lld` on OS X?

ChShersh:
@deepfire Motivation for this is to speed up builds on macOS. Sure `lld` is faster but building with `lld` is not working :( So we at least can try to use `gold` on macOS. Probably making `lld` work on macOS is bigger task. Making `gold` enabled is an intermediate step and partial solution of the problem but at least there's some hope it can be done fast.

ChShersh:
Turns out it's not possible to use `gold` linker on _macOS_:
https://stackoverflow.com/questions/22087368/installing-gold-linker-on-os-x

I googled a lot and tried after saw this answer. But seems like this is really impossible. So waiting for `lld` linker.

PR #369, opened 2017-05-23
[CSL-780] Move Pos.Ssc.Extra.Logic to ssc/

neongreen:


ssc/Pos/Ssc/RichmenComponent.hs
```diff
@@ -0,0 +1,33 @@
...
+       ) where
+
+import           Universum
+
+import           Pos.Core               (EpochIndex)
+import           Pos.DB.Class           (MonadDB)
+import           Pos.Lrc.Class          (RichmenComponent (..))
+import           Pos.Lrc.DB.RichmenBase (getRichmen, putRichmen)
+import           Pos.Lrc.Types          (FullRichmenData, RichmenStake)
+import           Pos.Update.Constants   (genesisMpcThd)
```
gromakovsky: Is it the only reason we need update in ssc? If it is, then I think it makes sense to move this constant somewhere deeper and get rid of this dependency.

src/Pos/Block/Logic/Header.hs
```diff
@@ -70,13 +70,13 @@ mkCHRinvalid = CHInvalid . T.intercalate "; "
 -- | Classify new header announced by some node. Result is represented
 -- as ClassifyHeaderRes type.
 classifyNewHeader
-    :: (MonadSlots m, SscHelpersClass ssc, MonadDBPure m)
+    :: forall m ssc. (MonadSlots m, DB.MonadBlockDB ssc m)
```
gromakovsky: `MonadBlockDB` is more severe constraint than `MonadDBPure`, but probably we can live with it.

PR #371, opened 2017-05-24
[DEVOPS-107]  Travis: resource diagnostics, cosmetic resource usage improvement

deepfire:


domenkozar:
@int-index @georgeee so now we could try to enable haddock.

PR #372, opened 2017-05-24
[CSL-1081] Refactor Pos.Block

gromakovsky:
This PR brings a new portion of refactoring. The task is not finished yet, but it's getting there. Again, I don't want to diverge too much, hence I created this PR.
Notable changes:
• moved more core stuff into core;
• made smart constructors for block/header more consistent with other code;
• split `Pos.Block.Logic` into multiple modules;
• made some constraints in `Pos.Block.Logic` less strict;
• introduced `MonadDBPure` and `MonadBlockDBGeneric`.

Addition of `MonadDBPure` and `MonadBlockDBGeneric` is the most notable change. You can read rationale behind these classes in `Pos.DB.Class`. The main reason I did it was that soon we will want to test some functionality from `Pos.Block` which uses `MonadDB` a lot and we can't use `MonadDB` in pure tests (which are our current goal).
I made some experiments and that's the result. It's not ideal, but I think it's the right direction.
For example, `Pos.Block.Logic.Header` doesn't use `MonadDB` at all as a result.
Addition of `MonadBlockDBGeneric` should also be useful for @neongreen to be able to get block/header from some block component without knowledge of type of block/header.


core/Pos/Util/Util.hs
```diff
@@ -260,6 +261,10 @@ type instance ChannelT (Ether.TaggedTrans tag t m) = ChannelT m
 maybeThrow :: (MonadThrow m, Exception e) => e -> Maybe a -> m a
 maybeThrow e = maybe (throwM e) pure
 
+-- | Fail or return result depending on what is stored in 'Either'.
+eitherToFail :: (MonadFail m, ToString s) => Either s a -> m a
```
volhovM: Don't we usually choose `MonadError Text m` instead?

src/Pos/Block/Core/Genesis/Misc.hs
```diff
@@ -96,3 +105,41 @@ instance HasDifficulty (GenesisBlock ssc) where
...
+    -> EpochIndex
+    -> SlotLeaders
+    -> GenesisBlock ssc
+mkGenesisBlock prevHeader epoch leaders =
+    either disaster identity $ recreateGenericBlock header body extra
+  where
+    header = mkGenesisHeader prevHeader epoch body
+    body = GenesisBody leaders
+    extra = GenesisExtraBodyData $ mkAttributes ()
+    disaster = error . mappend "mkGenesisBlock: "
```
volhovM: I guess it's better to use explicit type-level exceptions here (TODO next part of this refactoring).

src/Pos/Block/Core/Main/Misc.hs
```diff
@@ -4,15 +4,18 @@
 -- | Miscellaneous instances, etc. Related to the main blockchain of course.
 
 module Pos.Block.Core.Main.Misc
```
volhovM: OMG I don't think it's a proper naming. Why put important creation helpers into `Misc`? You could leave it as `Pure`, couldn't you?

db/Pos/DB/Class.hs
```diff
@@ -1,24 +1,67 @@
...
+--
+-- 'MonadDB' is the most featured class (actually just a set of
+-- constraints) which wraps 'NodeDBs' which contains RocksDB
+-- databases. This class can be used to manipulate RocksDB
+-- directly. It may be useful when you need an access to advanced
+-- features of RocksDB.
+--
+-- Apart from that we have three more classes here.
+--
+-- 'MonadDBPure' contains only 'dbGet' method.  The advantage of it is
```
volhovM: Doesn't say anything about the fact it's the main class for DB emulation. And I don't like the naming (already told it in slack).

src/Pos/DB/Block.hs
```diff
@@ -210,13 +218,49 @@ blockIndexKey :: HeaderHash -> ByteString
...
+instance (MonadDBPure m, MonadDB m, t ~ IdentityT, SscHelpersClass ssc) =>
+         MonadBlockDBGeneric (BlockHeader ssc) (Block ssc) Undo (Ether.TaggedTrans BlockDBRedirectTag t m) where
+    dbGetBlock = getBlock
+    dbGetUndo = getUndo
+    dbGetHeader = blkGetHeader
+
+blkGetBlock ::
+       forall ssc m. MonadBlockDB ssc m
+    => HeaderHash
+    -> m $ Maybe (Block ssc)
```
volhovM: Why not `m $ Maybe $ Block ssc`?
:trollface:

core/Pos/Util/Util.hs
gromakovsky: What exactly do you mean? If you have `f :: MonadError Text m => m ()` and you want to use it inside `MonadFail`, `eitherToFail` is your friend.

volhovM: Why do you use `MonadFail` at all?

gromakovsky: Because `Get` forces me to do it.

volhovM: Reasonable.

src/Pos/Block/Core/Genesis/Misc.hs
gromakovsky: I don't know how to do it, you can do it if you want.

volhovM: type-level ~ use maybe

src/Pos/Block/Core/Main/Misc.hs
gromakovsky: Naming is too hard, sadly.

core/Pos/Util/Util.hs
pva701: @gromakovsky seems, `Get` implements `MonadFail`.
https://github.com/input-output-hk/cardano-sl/pull/372/commits/9fc286a29078426234e7b1fc51243bd7181fd4e3#diff-0c3ba35b3dd7ef1323f51e1695a28047L61

gromakovsky: Right.

src/Pos/Ssc/GodTossing/Toss/Logic.hs
```diff
@@ -46,10 +46,11 @@ verifyAndApplyGtPayload
...
     => Either EpochIndex (Some IsMainHeader) -> GtPayload -> m ()
 verifyAndApplyGtPayload eoh payload = do
-    -- We can't trust payload from mempool, so we must call @sanityChecksGtPayload@.
+    -- We can't trust payload from mempool, so we must call
+    -- @sanityChecksGtPayload@.
     whenLeft eoh $ const $ sanityChecksGtPayload eoh payload
-    -- We perform @sanityChecksGtPayload@ for block when deserialize it (in the recreateMainBlock).
-    -- So this check just in case.
+    -- We perform @sanityChecksGtPayload@ for block when we construct it
+    -- it (in the 'recreateGenericBlock').  So this check is just in case.
```
pva701: `it` duplication

cardano-sl.cabal
```diff
@@ -174,7 +174,11 @@ library
                         Pos.Block.Core.Union.Misc
                         Pos.Block.Core.Union.Types
                         Pos.Block.Error
+                        Pos.Block.Logic.Creation
+                        Pos.Block.Logic.Header
                         Pos.Block.Logic.Internal
+                        Pos.Block.Logic.Util
+                        Pos.Block.Logic.VAR
```
pva701: THIS IS VAR!

src/Pos/Block/Logic/Util.hs
```diff
@@ -34,6 +34,9 @@ import           Pos.Util            (_neHead)
 import           Pos.Util.Chrono     (NE, OldestFirst (getOldestFirst))
 import           Pos.WorkMode.Class  (WorkMode)
 
+-- | This function can be used to create a message when tip mismatch
+-- is detected (usually between tip stored in DB and some other tip
+-- received from somewhere).
 tipMismatchMsg :: Text -> HeaderHash -> HeaderHash -> Text
 tipMismatchMsg action storedTip attemptedTip =
```
pva701: maybe should move this function to `Pos.Util` (or something like this), because it's also needed in other components.

src/Pos/Ssc/GodTossing/Toss/Logic.hs
gromakovsky: I will fix in the next part.

src/Pos/DB/DB.hs
```diff
@@ -112,18 +113,24 @@ initNodeDBs = do
...
+getTipBlock = getTipSomething @ssc "block" getBlock
 
 -- | Get BlockHeader corresponding to tip.
--- TODO don't load tip block, fix it.
 getTipBlockHeader
-    :: (SscHelpersClass ssc, MonadDB m, MonadDBPure m)
+    :: forall ssc m. (SscHelpersClass ssc, MonadDB m, MonadDBPure m)
     => m (BlockHeader ssc)
-getTipBlockHeader = getBlockHeader <$> getTipBlock
+getTipBlockHeader = getTipSomething @ssc "header" getBlockHeader
```
pva701: :+1:

db/Pos/DB/Class.hs
```diff
@@ -34,40 +60,40 @@ import           Pos.Core                       (BlockVersionData (..))
...
+dbTagToLens GStateDB     = gStateDB
+dbTagToLens LrcDB        = lrcDB
+dbTagToLens MiscDB       = miscDB
+
+-- | Pure interface to the database.
+-- TODO: add some ways to put something.
+class MonadThrow m => MonadDBPure m where
+    dbGet :: DBTag -> ByteString -> m (Maybe ByteString)
+
+    default dbGet :: (MonadTrans t, MonadDBPure n, t n ~ m) =>
```
pva701: lol, seems it is not required due to
https://github.com/input-output-hk/cardano-sl/pull/372/commits/8191024f4163d9540c377cf31405293460f5fd0f#diff-1acc0f10680712fa9b8e376a57bfe413R90

db/Pos/DB/Class.hs
```diff
@@ -34,40 +60,40 @@ import           Pos.Core                       (BlockVersionData (..))
...
+
+dbTagToLens :: DBTag -> Lens' NodeDBs DB
+dbTagToLens BlockIndexDB = blockIndexDB
+dbTagToLens GStateDB     = gStateDB
+dbTagToLens LrcDB        = lrcDB
+dbTagToLens MiscDB       = miscDB
+
+-- | Pure interface to the database.
+-- TODO: add some ways to put something.
+class MonadThrow m => MonadDBPure m where
```
pva701: more abstractions!
It's not clear necessity of this type class for me

db/Pos/DB/Class.hs
volhovM: Line 86 is default method which is used to derive the instance you're pointing to. So everything is fine.

db/Pos/DB/Class.hs
volhovM: Mocking database for testing i think.

gromakovsky: It's needed to use DB w/o `IO`.

db/Pos/DB/Class.hs
```diff
@@ -127,6 +144,40 @@ gsMaxProposalSize = bvdMaxProposalSize <$> gsAdoptedBVData
...
+
+-- | Convenient wrapper which combines 'dbGetBlock' and 'dbGetUndo' to
+-- read 'Blund'.
+dbGetBlund ::
+       forall header blk undo m. MonadBlockDBGeneric header blk undo m
+    => HeaderHash
+    -> m $ Maybe (blk, undo)
+dbGetBlund x =
+    runMaybeT $
+    (,) <$> MaybeT (dbGetBlock @header @blk @undo x) <*>
```
pva701: К Р А С U B O

src/Pos/Block/Logic/Header.hs
```diff
@@ -205,7 +205,7 @@ classifyHeaders headers = do
 -- checkpoint that's in our main chain to the newest ones.
 getHeadersFromManyTo
     :: forall ssc m.
-       (MonadDB m, MonadDBPure m, SscHelpersClass ssc, CanLog m, HasLoggerName m)
+       (MonadDB m, DB.MonadBlockDB ssc m, CanLog m, HasLoggerName m)
```
pva701: `MonadDB m, DB.MonadBlockDB` looks a little bit inconsistent

src/Pos/Ssc/GodTossing/Toss/Logic.hs
gromakovsky: Well, I've fixed it already.

db/Pos/DB/Class.hs
pva701: why not to write
```
instance {-# OVERLAPPABLE #-}
MonadThrow (t m)) =>
    (MonadDBPure m, MonadTrans t, MonadThrow (t m)) => MonadDBPure (t m)
    dbGet tag = lift . dbGet tag
``` 
and remove this default method

gromakovsky: Why is it better?

pva701: lol, because it's more reasonable/unified use one approach over all code: or default method or new approach introduced by @int-index.
the current code mix two these approaches

gromakovsky: Code written before my changes also mixes them, so it's not for this issue anyway.

gromakovsky: For example, `Pos.Txp.Toil.Class`.

pva701: lol, I don't get what is the reason use 
```
instance {-# OVERLAPPABLE #-}
    (MonadTxPool m, MonadTrans t, Monad (t m)) =>
        MonadTxPool (t m)
```
this code with default methods at all. cc @int-index

core/Pos/Util/Util.hs
pva701: >Because `Get` forces me to do it.
so what does it mean?

don't get why you didn't use `MonadFail`.
because you didn't want to use it?

pva701: o, fuck.
it's the wrong place to write a comment.

I noticed that you used `MonadError` somewhere, and don't get why did you use it

gromakovsky: Didn't use where?

pva701: https://github.com/input-output-hk/cardano-sl/pull/372/files/35aa23e762b8e8ee1a8ef3d7aa69884787d4a5b3#diff-43c2c21b43694f1827d7691b01cdb3bcR78
for example, why did you prefer `MonadError` here?

gromakovsky: >  because you didn't want to use it?

Right

gromakovsky: One person asks why I don't use `MonadFail`, another one asks why I use `MonadFail` :D

pva701: ok, but `MonadError Text m` looks strange to me, `MonadError Text m ~ MonadFail m`

db/Pos/DB/Class.hs
int-index: @pva701  It doesn't matter if you write this with or without `default` signatures. I didn't care enough to remove them, so they are still used in many places, they're not necessary when you use `{-# OVERLAPPABLE #-}`.

core/Pos/Util/Util.hs
int-index: > MonadError Text m ~ MonadFail m

Not true. `MonadError` also gives you the `catch` operation. I'd avoid `MonadFail` whenever possible.

Instead of `MonadError` I recommend creating a type per exception and using them with `Ether.MonadExcept'`. One disadvantage is that each layer of `ExceptT` adds a bit of performance overhead, but I can implement flattening for `Ether.MonadExcept` like I did for `Ether.MonadReader`, so don't worry about this.

If you don't want to rewrite error handling just yet, I recommend using `MonadError`.

gromakovsky:
Travis failed more than 10 times in a row (two different travises together) with either OOM or hanging tests. I would like to stop monitoring this page constantly and restarting build every hour, so I am merging it.

core/Pos/Util/Util.hs
pva701: @int-index but `MonadError Text m` is used like `MonadFail` in this PR (`catch` isn't used).
so I don't understand advantages `MonadError Text` here

PR #376, opened 2017-05-25
[CSL-1081] Refactor Pos.Block

gromakovsky:
Yet another part of refactoring. I think it's the last one. Notable changes:
1. I made some checks internal invariants of some types, cleaned up `Pos.Block.Pure`.
2. Moved `verifyBBlock` into `BlockchainHelpers` class, also added `verifyBBlockHeader` there.
3. Introduced a `Slog` component to `Pos.Block` which does VAR logic not contained in any other component.
4. Made some constraints less severe and more explicit (in `Pos.Lrc` and `Pos.Block.Logic`).
5. Commented and cleaned-up some obscure places.
6. Slightly refactored `Pos.Block.Logic.Creation`.

I think it is enough for this task.


core/Pos/Binary/Core/Types.hs
```diff
@@ -38,17 +35,16 @@ instance Bi T.CoinPortion where
     get = label "CoinPortion" $ get >>= T.mkCoinPortion
 
 instance Bi T.LocalSlotIndex where
-    get = label "LocalSlotIndex" $ T.LocalSlotIndex . getUnsignedVarInt <$> get
-    put (T.LocalSlotIndex c) = put (UnsignedVarInt c)
+    get =
+        label "LocalSlotIndex" $
+        (eitherToFail . T.mkLocalSlotIndex) . getUnsignedVarInt =<< get
```
rockbmb: This is insignificant, but are the parens needed here?

gromakovsky: Not needed, but it's insignificant. Maybe I will fix, maybe someone else will do it.

core/Pos/Core/Block.hs
```diff
@@ -75,6 +77,9 @@ class Blockchain p where
     default checkBodyProof :: Eq (BodyProof p) => Body p -> BodyProof p -> Bool
     checkBodyProof body proof = mkBodyProof body == proof
 
+-- | Extension of 'Blockchain' type class with helper functions.
+class Blockchain p => BlockchainHelpers p where
+    verifyBBlockHeader :: MonadError Text m => GenericBlockHeader p -> m ()
```
volhovM: Add description to these -- which checks are performed. Because it does obviously less than `Pure.verifyX` can.

src/Pos/Block/Arbitrary.hs
```diff
@@ -378,20 +391,18 @@ newtype HeaderAndParams ssc = HAndP
...
         -- 'skip' is the random number of headers that should be skipped in the header
         -- chain. This ensures different parts of it are chosen each time.
         skip <- choose (0, num - 1)
-        let atMost3HeadersAndLeaders = take 3 $ drop skip headers
-            (prev, header, next) =
-                case atMost3HeadersAndLeaders of
-                    [h] -> (Nothing, h, Nothing)
-                    [h1, h2] -> (Just h1, h2, Nothing)
-                    (h1 : h2 : h3 : _) -> (Just h1, h2, Just h3)
+        let atMost2HeadersAndLeaders = take 2 $ drop skip headers
```
volhovM: Why don't you check `vhpNextHeader` and `vhpConsensus` now?

src/Pos/Block/Core/Main/Helpers.hs
```diff
@@ -0,0 +1,74 @@
...
+            sscVerifyPayload @ssc
+                (Right (Some _gbHeader))
+                (_mbSscPayload _gbBody)
+        dlgVerifyPayload (_gbHeader ^. epochIndexL) (_mbDlgPayload _gbBody)
+
+verifyMainBlockHeader ::
+       (Ssc ssc, MonadError Text m, Bi $ BodyProof $ MainBlockchain ssc)
+    => MainBlockHeader ssc
+    -> m ()
+verifyMainBlockHeader UnsafeGenericBlockHeader { _gbhConsensus = MainConsensusData { _mcdLeaderKey = leaderPk
```
volhovM: Seriously? I will refactor this the next time i see this code if you don't :)

src/Pos/Block/Core/Main/Misc.hs
```diff
@@ -164,22 +166,28 @@ mkMainHeader
...
+    makeSignature toSign (Left psk) =
+        BlockPSignatureLight $ proxySign SignMainBlockLight sk psk toSign
+    makeSignature toSign (Right psk) =
+        BlockPSignatureHeavy $ proxySign SignMainBlockHeavy sk psk toSign
     signature prevHash proof =
         let toSign = MainToSign prevHash proof slotId difficulty extra
-        in maybe (BlockSignature $ sign SignMainBlock sk toSign)
-                 (makeSignature toSign)
-             pSk
+        in maybe
```
volhovM: 😒

src/Pos/Block/Logic/Creation.hs
```diff
@@ -252,30 +279,57 @@ createMainBlockPure
...
+--
+-- Given limit applies only to body, not to other data from block.
+createMainBody
+    :: forall m ssc .
+       (MonadError Text m, SscHelpersClass ssc)
+    => Byte  -- ^ Body limit
+    -> SlotId
+    -> RawPayload ssc
+    -> m $ Body $ MainBlockchain ssc
+createMainBody bodyLimit sId payload =
```
volhovM: I find this particular function refactoring out to be useless. It only adds extra 15 lines of code and makes function even more complicated.

core/Pos/Core/Block.hs
gromakovsky: All checks which can be done without any extra information. Ok, I will comment.

src/Pos/Block/Arbitrary.hs
gromakovsky: `vhpConsensus` is fully checked in constructor.
`vhpNextHeader` wasn't used in code, so I removed it completely. There is no need to test something that is not used in actual code.

src/Pos/Block/Core/Main/Misc.hs
gromakovsky: I think it was `hindent`.

src/Pos/Block/Logic/Creation.hs
gromakovsky: Old `createMainBlockPure` was very big and contained a lot of logic. I split this logic into to functions more or less evenly, I think it's generally useful refactoring.

gromakovsky: I don't know where it's more complicated, it does part of what it was doing before. Also it unwraps `DlgPayload` and `RawPayload`, but that's because, first, previously there was no `DlgPayload`, second, I added `RawPayload` to avoid passing 4 arguments to a function which already has too many arguments.

core/Pos/Core/Slotting.hs
```diff
@@ -51,33 +55,44 @@ instance Enum SlotId where
...
+crucialSlot epochIdx = SlotId {siEpoch = epochIdx - 1, ..}
+  where
+    siSlot =
+        leftToPanic "crucialSlot: " $
+        mkLocalSlotIndex (epochSlots - slotSecurityParam - 1)
 
 instance Enum EpochOrSlot where
     succ (EpochOrSlot (Left e)) =
-        EpochOrSlot (Right SlotId {siEpoch = e, siSlot = 0})
+        EpochOrSlot (Right SlotId {siEpoch = e, siSlot = minBound})
```
pva701: why did you replace 0 with `minBound` everywhere?

src/Pos/SafeCopy.hs
```diff
@@ -120,6 +121,8 @@ deriveSafeCopySimple 0 'base ''TxProof
 deriveSafeCopySimple 0 'base ''TxPayload
 deriveSafeCopySimple 0 'base ''SharedSeed
 
+deriveSafeCopySimple 0 'base ''DlgPayload
```
pva701: memniy file

cardano-sl.cabal
```diff
@@ -108,6 +108,7 @@ library
```
pva701: did you forget to squash this commit with the previous commit?

PR #378, opened 2017-05-25
[CSL-969] Add helpers to work with `scrypt`

Martoon-00:
It was intended to change replace hashing of passphrase for secret key, but it will be done under CSL-1169 + some other issue, only helpers remained

### DIFF
--- a/core/Pos/Crypto/Scrypt.hs
+++ b/core/Pos/Crypto/Scrypt.hs
@@ -1,0 +1,44 @@
+-- | Wrapper over scrypt library
+
+module Pos.Crypto.Scrypt
+    ( S.ScryptParams
+    , S.Salt (..)
+    , S.Pass (..)
+    , S.EncryptedPass (..)
+
+    , S.scryptParams
+    , S.scryptParamsLen
+
+    , mkSalt
+    , genSalt
+    , encryptPassIO
+    , encryptPassWithSalt
+    , verifyPass
+    ) where
+
+import           Universum
+
+import qualified Crypto.Scrypt     as S
+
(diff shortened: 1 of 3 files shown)


Martoon-00:
Requires hashing parameters tuning

core/Pos/Crypto/Scrypt.hs
```diff
@@ -0,0 +1,66 @@
...
+import           Pos.Binary.Class  (Bi, encodeStrict)
+import           Pos.Crypto.Random (secureRandomBS)
+
+-- | This corresponds to 'ScryptParams' datatype.
+-- These parameters influence on resulting hash length, memory and time
+-- consumption. See documentation for exact details.
+data ScryptParamsBuilder = ScryptParamsBuilder
+    { spLogN    :: Word
+    , spR       :: Word
+    , spP       :: Word
```
rockbmb: Could you rename these 3 fields, or add some Haddock documentation to each so that their purpose becomes clearer?

Martoon-00: You're right, it became unclear. 
In your opinion, if I attach this link, will it be enough? https://hackage.haskell.org/package/scrypt-0.5.0/docs/Crypto-Scrypt.html

Martoon-00: Okay, I'll do my best to make it understandable in 2 minutes

vincenthz:
another redundant package with cryptonite

Martoon-00:
That `scrypt` library provides a more convenient way to go than the single function in `cryptonite`, but perhaps it doesn't worth having extra dependency, I don't even know whether will it be required later. So closing

PR #379, opened 2017-05-26
[CSL-1042] Do not treat exceptions like UnknownBlockForLrc or OutdatedSlottingData as errors while we're in recovery

pva701:
* Don't run `onNewSlot` workers when node is in recovery mode

### DIFF
--- a/infra/Pos/Communication/Protocol.hs
+++ b/infra/Pos/Communication/Protocol.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -30,8 +31,8 @@ import qualified Data.HashMap.Strict              as HM
 import qualified Data.List.NonEmpty               as NE
 import qualified Data.Text.Buildable              as B
 import           Formatting                       (bprint, build, sformat, (%))
-import           Mockable                         (Delay, Fork, Mockable, SharedAtomic,
-                                                   Throw, throw)
+import           Mockable                         (Delay, Fork, Mockable, Mockables,
+                                                   SharedAtomic, Throw, throw)
 import qualified Node                             as N
 import           Node.Message                     (Message (..), MessageName (..),
                                                    messageName')
@@ -190,18 +191,10 @@ worker' outSpecs h =
 
 
 type OnNewSlotComm m =
-    ( MonadIO m
-    , MonadSlots m
-    , MonadMask m
-    , WithLogger m
-    , Mockable Fork m
-    , Mockable Delay m
+    ( LocalOnNewSlotComm m
     , Mockable Throw m
     , WithPeerState m
     , Mockable SharedAtomic m
-    , MonadReportingMem m
-    , MonadShutdownMem m
-    , MonadDiscovery m
     )
 
 type LocalOnNewSlotComm m =
@@ -209,8 +202,7 @@ type LocalOnNewSlotComm m =
     , MonadSlots m
     , MonadMask m
     , WithLogger m
-    , Mockable Fork m
-    , Mockable Delay m
+    , Mockables m [Fork, Delay]
     , MonadReportingMem m
     , MonadShutdownMem m
     , MonadDiscovery m
--- a/infra/Pos/Communication/Protocol.hs
+++ b/infra/Pos/Communication/Protocol.hs
@@ -204,6 +204,18 @@ type OnNewSlotComm m =
     , MonadDiscovery m
     )
 
+type LocalOnNewSlotComm m =
+    ( MonadIO m
+    , MonadSlots m
+    , MonadMask m
+    , WithLogger m
+    , Mockable Fork m
+    , Mockable Delay m
+    , MonadReportingMem m
+    , MonadShutdownMem m
+    , MonadDiscovery m
+    )
+
 onNewSlot'
     :: OnNewSlotComm m
     => Bool -> Bool -> (SlotId -> WorkerSpec m, outSpecs) -> (WorkerSpec m, outSpecs)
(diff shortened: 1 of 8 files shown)


infra/Pos/Communication/Protocol.hs
```diff
@@ -204,6 +204,18 @@ type OnNewSlotComm m =
     , MonadDiscovery m
     )
 
+type LocalOnNewSlotComm m =
```
volhovM: Nope. 
1. Merge with `OnNewSlotComm`
2. Use `Mockables`

src/Pos/Context/Functions.hs
```diff
@@ -83,3 +85,11 @@ recoveryInProgress :: (MonadIO m, MonadRecoveryHeader ssc m) => m Bool
 recoveryInProgress = do
     var <- Ether.ask @RecoveryHeaderTag
     isJust <$> atomically (STM.tryReadTMVar var)
+
+
+
+recoveryCommGuard
```
volhovM: 1. Three empty lines.
2. No documentation.
3. Misplacement: you've just added import of `Pos.Communication.Protocol` into `Context.Functions`. `Context.*` are supposed to be lightweight types and not import anything heavy. Another file -- `Context.Context` does some imports from `Communication`, but they are type-oriented. Solution: import `Protocol.Types` instead (if it's possible, idk) or move `recoveryCommGuard` elsewhere.

infra/Pos/Communication/Protocol.hs
pva701: why?

pva701: ladno

volhovM: To remove redundant code duplication obv.

src/Pos/Context/Functions.hs
```diff
@@ -83,3 +85,11 @@ recoveryInProgress :: (MonadIO m, MonadRecoveryHeader ssc m) => m Bool
 recoveryInProgress = do
     var <- Ether.ask @RecoveryHeaderTag
     isJust <$> atomically (STM.tryReadTMVar var)
+
+-- | This function helper for workers.
```
volhovM: This PR helper to coding very many good thanks <3 <3

pva701: blya, zabil "is". shas forcepushu

src/Pos/Lrc/Worker.hs
```diff
@@ -56,17 +57,17 @@ import           Pos.WorkMode.Class         (WorkMode)
...
-    (lrcSingleShot siEpoch `catch` reportError) `catch` onLrcError
+        lrcSingleShot siEpoch `catch` onLrcError
   where
-    reportError (SomeException e) = do
-        reportMisbehaviourMasked version $ "Lrc worker failed with error: " <> show e
-        throwM e
-    onLrcError UnknownBlocksForLrc =
+    onLrcError e@UnknownBlocksForLrc = do
+        reportError e
         logInfo
```
gromakovsky: I still have this comment http://i.imgur.com/i7cLxye.png in my mailbox. Perhaps it's a good time to change it to `logWarning`?

pva701:
>1.What about logNewSlotWorker?

che ita? will see tomorrow.

>2. Why did you decide to do it in many files, instead of changing few lines in Pos.Worker (where all workers are aggregated)?

At first: for example, there are two blocks and only one need this stuff (not both) 
Secondly: this logic is an own logic of each worker, today worker mustn't be run during recovery, but tomorrow it must. or today component contains only one worker, which doesn't need `recoveryCommGuard`, but tomorrow two and one of them need `recoveryCommGuard`. so part of usages of `recoveryCommGuard` will be at `Pos.Worker` and another part in components.

src/Pos/Lrc/Worker.hs
pva701: maybe maybe, who knows...

gromakovsky:
> che ita? will see tomorrow.

You can use `util-scripts/grep.sh` to find it in source code.

gromakovsky:
> At first: for example, there are two block workers and only one needs this stuff (not both)
> Secondly: this logic is an own logic of each worker, today worker mustn't be run during recovery, but tomorrow it must be. or today component contains only one worker, which doesn't need recoveryCommGuard, but tomorrow two, and one of them needs recoveryCommGuard.
so part of usages of recoveryCommGuard would be at Pos.Worker and another part in components.

Ok, makes sense. I think it's subjective and both approaches have advantages. Yes, some workers should run even when we are not in recovery mode. But I would say these workers are exceptional. AFAIK, there are only two such workers and both retrieve blocks and are in `Pos.Block.Worker` (`retrievalWorker` and `queryBlocksWorker`). Normally worker shouldn't do anything in recovery mode.
If we put all these `recoveryCommGuard`s into `Pos.Worker` (except for `blkWorkers`), then it will be clear which workers are exceptions. Also it makes it harder to forget to add such guard somewhere (like you apparently did for `logNewSlotWorker`.

But ok, what you are saying also makes sense and since you are doing this task, decide yourself 😉

src/Pos/Lrc/Worker.hs
gromakovsky: Just do it 😉

pva701:
just hater :P

pva701:
@gromakovsky fixed

src/Pos/Lrc/Worker.hs
gromakovsky: Okay, I did it and archived that message in my mailbox 🎉

gromakovsky:
I approved, but what about `relayWorkers`? Are they guarded? If not, why not?

pva701:
@gromakovsky to be honest, I suppose there are a lot of places which can be guarded but didn't.
I think the problem is more general: we should consider the lifecycle of a node during recovery mode, such points as:
* listeners (what should they do? should we listen during recovery mode?)
* non-`onNewSlot` workers (what should they do? and should work or not)
* mempool (should we accept any data?) also relay
* maybe something also
Now we don't care about it, maybe it is ok, maybe not.
Does it make sense or меня плющит?

Concretely about relay-worker: seems, if relay worker will work during recovery, nothing bad shouldn't happen, it only propagates received messages. Or I don't get something?

gromakovsky:
```
Now we don't care about it, maybe it is ok, maybe not.
Does it make sense or меня плющит?
```

What's that?

> I suppose there are a lot of places which can be guarded but didn't. I think the problem is more general: we should consider the lifecycle of a node during recovery

Yes, I agree, this issue is not about it.

> Concretely about relay-worker: seems, if relay worker works during recovery, nothing bad shouldn't happen, it only propagates received messages. Or I don't get something?

Well, if node's GState is outdated, she won't be able to properly verify data she receives and chances of ignoring valid data or accepting invalid data (I mean, invalid from other nodes' point of view, this data will be removed after blocks are fetched) become higher.

I created https://issues.serokell.io/issue/CSL-1182 just in case.

PR #380, opened 2017-05-26
[CSL-1154] Split `Pos.Launcher.Launcher` into submodules

Martoon-00:
We are running out of memory, sir!


pva701:
@Martoon-00  WE NEED MORE MEMORY

Martoon-00:
Lil, it doesn't help anymore :shrug:

PR #382, opened 2017-05-27
Kademlia: get rid of DHT key parameter for node CLI/configs

pva701:


infra/Pos/DHT/Real/Real.hs
```diff
@@ -147,9 +147,13 @@ kademliaGetKnownPeers
...
 kademliaGetKnownPeers inst = do
     let myId = kdiKey inst
-    let explicitInitial = kdiExplicitInitial inst
-    let initPeers = bool [] (kdiInitialPeers inst) explicitInitial
-    buckets <- liftIO (K.viewBuckets $ kdiHandle inst)
+    let kInst = kdiHandle inst
+    let initNetAddrs = map toKPeer (kdiInitialPeers inst)
+    initPeers <- map toDHTNode <$>
+                     bool (pure [])
+                     (catMaybes <$> liftIO (K.peersToNodeIds kInst initNetAddrs))
```
georgeee: `peersToNodeIds` will return `[]` if node just initialized and there were no dump. We should use different here. If this is the case that we need return peers as `[DHTNode]`, then my proposal is to rewrite this method to use `[NetworkAddress]` and instead of `(K.peersToNodeIds kInst initNetAddrs)` have `initNetAddrs`

stack.yaml
```diff
@@ -23,7 +23,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/kademlia.git
-    commit: b4b0a6774368b9bbea96d7a03a1a451b6fc3cfea
+    commit: 2a51c89d39faef920d11137490294f34a8969601
```
georgeee: Don't forget to update this commit & one for time-warp after PR for kademlia gets merged to master

PR #383, opened 2017-05-27
Feature/csl1125 heavy redelegation

volhovM:
This PR introduces:
* Two more storages to dlg database -- transitive and transitive inverse, thus supporting redelegation
* Lifted wrappers over `concurrent-extra` (see Util/Concurrent in core). Delegation was rewriten from TVar to RVWar to combine state modification and db access.
* Minor refactorings, some checks that missed before (now you can't issue block if you've delegated to somebody).
* Fixes of dlg rollbacks (there was a broken place with revokation certs).

The logic may be a little bit tricky in some places, but at least i've left comments.


src/Pos/Delegation/DB.hs
```diff
@@ -1,101 +1,239 @@
...
+    => Either PublicKey StakeholderId -> m DlgMemPool
+getPskChain = getPskChainInternal HS.empty
+
+-- See doc for 'getPskTree'. This function also stops traversal if
+-- encounters anyone in 'toIgnore' set.
+getPskChainInternal
+    :: MonadDBPure m
+    => HashSet StakeholderId -> Either PublicKey StakeholderId -> m DlgMemPool
+getPskChainInternal toIgnore (either addressHash identity -> issuer) =
+    fmap (view _1) $ flip execStateT (HM.empty, [issuer], HS.empty) trav
```
pva701: can you use datatype instead of `(,,)` or at least write what does this tuple mean?

src/Pos/Delegation/DB.hs
```diff
@@ -1,101 +1,239 @@
...
+    :: MonadDBPure m
+    => Either PublicKey StakeholderId -> m DlgMemPool
+getPskChain = getPskChainInternal HS.empty
+
+-- See doc for 'getPskTree'. This function also stops traversal if
+-- encounters anyone in 'toIgnore' set.
+getPskChainInternal
+    :: MonadDBPure m
+    => HashSet StakeholderId -> Either PublicKey StakeholderId -> m DlgMemPool
+getPskChainInternal toIgnore (either addressHash identity -> issuer) =
```
pva701: seems, this function is wrong: you perform only one iteration, remove one element from list and break, because you don't add any elements to `_2`

pva701: seems, you should add issuers to `_2` and break when `getPskByIssuer $ Right x` returns `Nothing`

src/Pos/Delegation/DB.hs
```diff
@@ -1,101 +1,239 @@
...
+-- leaves. Executes 'getPskChain' for every issuer and merges. This
+-- function must be used under outside shared lock.
+getPskForest
+    :: (MonadDBPure m)
+    => Either [PublicKey] [StakeholderId] -> m DlgMemPool
+getPskForest (either (fmap addressHash) identity -> issuers) =
+    foldlM foldFoo HM.empty (map Right issuers)
+  where
+    -- Don't revisit branches we retrieved earlier.
+    foldFoo cur = getPskChainInternal (HS.fromList $ map addressHash $ HM.keys cur)
```
pva701: seems, you should merge returned chain with `cur`

src/Pos/Delegation/DB.hs
volhovM: Why do i perform one iteration? I call `trav` on 2/3 case matches.

src/Pos/Delegation/DB.hs
volhovM: You're right. This function is not used anyway, so i forgot to fix it I guess. Will just delete.

src/Pos/Delegation/DB.hs
volhovM: I'll add type, ok.

src/Pos/Delegation/Helpers.hs
```diff
@@ -21,3 +34,60 @@ dlgVerifyPayload epoch (getDlgPayload -> proxySKs) =
...
+    trav cur = ifM (uses identity $ HS.member cur) (pure $ Just cur) $ do
+        next <- lift $ resolve cur
+        identity %= HS.insert cur
+        let stop = pure Nothing
+        maybe stop (\psk -> bool stop (trav $ pskDelegatePk psk) $ isRevokePsk psk) next
+
+-- | Given an 'DlgMemPool', issuer, delegate and his psk, checks if
+-- delegate is allowed to issue the block. This does not check that
+-- delegate didn't issue a psk to somebody else.
+dlgReachesIssuance
```
pva701: is this function overkill?
I suppose that only root-delegate can publish blocks, so it's not needed to pass to this function other delegates.
am I wrong?

volhovM: No, the idea was to transfer "issue block" permission transitively as well. Does it answer your question?

pva701: @volhovM 
hm, suppose there three nodeы: A, B, C
A -> B (A delegated stake to B)
B -> C
1. can B publish blocks at all?
2. can B publish block instead of A? or only C can do it?

volhovM: No, only `C` can. For every chain `I -> X1 -> ... ->Xn -> D` only `D` can publish, and it's forbidden to do so for `{I}∪{Xi}_i`.

pva701: I am saying about it, but you pass to this function any node.
So I am confused by this comment:
>Given an 'DlgMemPool', issuer, delegate and his psk, checks if delegate is allowed to issue the block.

Only root-delegate can issue block, but you will return true in other cases

volhovM: This particular function indeed checks that "D reaches issuance" as it's written. The check "delegate didn't delegate to elseone" is performed in `delegationVerifyBlocks` step 1.

pva701: Ya uje ponyal, can you fix comment plz?

volhovM: Sure. I'll forcepush though.

src/Pos/Delegation/Logic.hs
```diff
@@ -171,40 +133,294 @@ initDelegation = do
...
+    changeset <- transChangeset
+    let toTransOp iPk (This _)      = GS.DelTransitiveDlg iPk
+        toTransOp iPk (That dPk)    = GS.AddTransitiveDlg iPk dPk
+        toTransOp iPk (These _ dPk) = GS.AddTransitiveDlg iPk dPk
+    let transOps = map (uncurry toTransOp) (HM.toList changeset)
+
+    -- Bulid reverse transitive set and convert it to reverseTransOps
+    let reverseTrans = buildReverseTrans changeset
+    reverseOps <- forM (HM.toList reverseTrans) $ \(k, (ad0, dl0)) -> do
+        let ad = HS.toList ad0 -- view patterns are for the weak ones
```
pva701: mda uj.

volhovM: The line would get too long otherwise :(

src/Pos/Delegation/Logic.hs
```diff
@@ -171,40 +133,294 @@ initDelegation = do
...
+
+    pure $ SomeBatchOp $ transOps <> reverseOps
+  where
+    {-
+    Get the transitive changeset.
+
+    To scare the reader, we suggest the example. Imagine the following
+    transformation (all arrows are oriented to the right). First graph
+    G is one we store in DB. We apply the list of edgeActions {del DE,
+    add CF} ~ F to it. Note that order of applying edgeActions
```
pva701: rly? seems only the order of delegation and a corresponding revocation.
so if there aren't revocations, the order doesn't matter, right?

src/Pos/Delegation/Logic.hs
```diff
@@ -171,40 +133,294 @@ initDelegation = do
...
+    matters, so we can't go incrementally building graph G' = G ∪ F
```
pva701: why can't?

src/Pos/Delegation/Logic.hs
volhovM: It does, because, for instance, one can have subgraph of changes `F' ⊂ F` such that `G ∪ F'` has cycles. Consider this:
```
G: a -> b -> c -> d
F: {remove c -> d, add d -> a}
```

src/Pos/Delegation/Logic.hs
volhovM: See previous comment.

src/Pos/Delegation/Logic.hs
```diff
@@ -171,40 +133,294 @@ initDelegation = do
...
+
+    A   F--G      A   F--G
+     \             \ /
+      C--D--E  ⇒    C  D  E
+     /             /
+    B             B
+
+    Let dlg_H(a) denote the transitive delegation relation, returning
+    Nothing or Just dPk. Then we want to:
+    1. Find affected users af = {e ∈ E_G ∪ E_F | dlg_G(e) ≠ dlg_G'(e)}
```
pva701: I don't get these denotations.
E_G = set of edges of graph G? usually it is denoted as `E(G)`.
And I don't get what does mean `dlg_G(e)` seems `dlg_G :: V -> V`, but `e` is edge

volhovM: (1) Yes.
(2) dlg_G(e) = dlg_G(uv) ~ dlg_G(u), we identify edge by its beginning. I'll fix this comment.

src/Pos/Delegation/Logic.hs
```diff
@@ -171,40 +133,294 @@ initDelegation = do
...
+        dlgNew <- execStateT (for_ af calculateDlgNew) HM.empty
+        -- Let's check that sizes of af and dlgNew match (they should).
+        -- We'll need it to merge in (3).
+        let dlgKeys = keysMap dlgNew
+        when (dlgKeys /= af) $ throwM $ DBMalformed $
+            sformat ("transChangeset: dlgNew keys "%listJson%" /= af "%listJson)
+                    (HS.toList dlgKeys) (HS.toList af)
+
+        -- Step 3.
+        -- Some unsafe functions (чтобы жизнь медом не казалась)
```
pva701: russian swine :pig:

volhovM: It's "pig" :peka:

src/Pos/Delegation/Logic.hs
```diff
@@ -171,40 +133,294 @@ initDelegation = do
...
+            let retHm = HM.fromList $ map (,Just dPk) ret
+            -- Just a sanity check. If you're optimizing, delete it.
+            unless (HM.size retHm /= length ret) $ throwM $ DBMalformed $
+                sformat ("calculateLocalAf for iPk "%build%" has duplicates: "%listJson)
+                        iPk ret
+            pure retHm
+
+    eActionsHM :: HashMap PublicKey GS.DlgEdgeAction
+    eActionsHM =
+        HM.fromList $ map (\x -> (GS.dlgEdgeActionIssuer x, x)) $ HS.toList eActions
```
pva701: loooooooooool, I supposed prefix `GS.` implies action w/ DB

volhovM: It does, what's wrong?

pva701: but `eActionsHM` is pure, but internally it uses `GS.dlgEdgeActionIssuer`

volhovM: Well, it's a `DlgEdgeAction` helper. I think the most appropriate place for it is in `DB` because that's where it is used mostly.

src/Pos/Delegation/Logic.hs
pva701: ok, got it. thx

core/Pos/Core/Types.hs
```diff
@@ -248,11 +247,11 @@ type ProxySigHeavy a = ProxySignature EpochIndex a
 -- | Heavy delegation psk.
 type ProxySKHeavy = ProxySecretKey EpochIndex
 
--- | Some proxy secret key.
-type ProxySKEither = Either ProxySKLight ProxySKHeavy
-
--- | Map from issuers to heavy certs.
-type ProxySKHeavyMap = HashMap PublicKey ProxySKHeavy
+-- | Lightweight PSK or heavyweight PSK with real leader public key
```
gromakovsky: What is `real leader public key`? Leader's key is stored in header. Is it necessary here?

P. S. I just started reviewing, maybe later I will answer by myself.

core/Pos/Crypto/Signing.hs
```diff
@@ -247,27 +247,24 @@ verifyProxySecretKey ProxySecretKey{..} =
...
+-- because otherwise we can't verify sig in heavyweight psk transitive
+-- delegation: i -> x -> d, we have psk from x to d, slot leader is
+-- i.
 data ProxySignature w a = ProxySignature
-    { pdOmega      :: w
-    , pdDelegatePk :: PublicKey
-    , pdCert       :: ProxyCert w
-    , pdSig        :: CC.XSignature
+    { pdPsk :: ProxySecretKey w
+    , pdSig :: CC.XSignature
```
gromakovsky: Btw, what's the meaning of `pd` prefix? It seems a little confusing in other places (hard to understand it's from `ProxySignature`).

lrc/Pos/Lrc/Core.hs
```diff
@@ -18,14 +18,17 @@ import           Pos.Util.Iterator   (MonadIterator (..))
 import           Pos.Util.Util       (getKeys)
 
 
--- | Function helper for delegated richmen.
--- Iterate by Delegate -> [Issuer] map and compute two set:
--- 1. Old richmen who delegated own stake and isn't richman more.
+-- | Function helper for delegated richmen. Iterates @Delegate ->
+-- [Issuer]@ map and computes the following two set:
```
gromakovsky: s/set/sets

src/Pos/Delegation/Listeners.hs
```diff
@@ -84,7 +84,7 @@ pskHeavyRelay = Data $ DataParams $ handlePsk
   where
     handlePsk :: forall ssc m. WorkMode ssc m => ProxySKHeavy -> m Bool
     handlePsk pSk = do
-        logDebug $ sformat ("GLightLightot request to handle heavyweight psk: "%build) pSk
```
gromakovsky: wut

core/Pos/Core/Types.hs
volhovM: Yes, it's needed here. Before we were passing `(sk, Maybe (Either PskLight PskHeavy))`, where `sk` because psk has `issuerPk` inside. But now we can have slot leader `i` in the chain `i -> x -> d`, where we (sk) are `d`, and `x -> d` is heavy cert we pass, so we don't have any info about `i`. That's why it's added to this datatype.

core/Pos/Crypto/Signing.hs
```diff
@@ -247,27 +247,24 @@ verifyProxySecretKey ProxySecretKey{..} =
 
 -- | Delegate signature made with certificate-based permission. @a@
 -- stays for message type used in proxy (ω in the implementation
--- notes), @b@ for type of message signed.
+-- notes), @b@ for type of message signed. We add whole psk as a field
```
volhovM: Indeed, I will.

core/Pos/Crypto/Signing.hs
volhovM: Lol, no idea. Proxy Delegation? Got used to it. Will fix in https://issues.serokell.io/issue/CSL-1171

core/Pos/Core/Types.hs
gromakovsky: Okay, I grepped where it's used and now I see that it's used only for block creation.

> This is used to create a block header mainly.

So, not mainly, but only.

Probably it shouldn't be in `core` then, but somewhere in `Pos.Block` (perhaps `Pos.Block.Core.Main.Types`, near `BlockSignature`).

Secondly, all usages of it are `Maybe ProxySKEither`. So perhaps we don't need this type without `Maybe`. Also I never liked this `Maybe $ Either b c`, I think it should be a new datatype.

src/Pos/Block/Worker.hs
```diff
@@ -115,18 +115,20 @@ blkOnNewSlotImpl (slotId@SlotId {..}) sendActions = do
                       map (bprint pairF) (zip [0 :: Int ..] $ toList leaders)
         logLeadersF $ sformat ("Current slot leader: "%build) leader
         logDebugS $ sformat ("Available to use lightweight PSKs: "%listJson) validCerts
-        heavyPskM <- getPSKByIssuerAddressHash leader
-        logDebug $ "Does someone have cert for this slot: " <> show (isJust heavyPskM)
-        let heavyWeAreDelegate = maybe False ((== ourPk) . pskDelegatePk) heavyPskM
-        let heavyWeAreIssuer = maybe False ((== ourPk) . pskIssuerPk) heavyPskM
+        heavyPskIssuerM <- getPskByIssuer (Right leader)
+        let heavyWeAreIssuer = maybe False ((== ourPk) . pskIssuerPk) heavyPskIssuerM
```
gromakovsky: I don't understand this check. I think we should check whether **we** delegated to somebody?!

Currently you check whether leader delegated to us. But I don't get it. For example, suppose we are leader and we delegated to somebody (not to us, obviously). Then `heavyWeAreIssuer` will be `False`. But in this case we shouldn't create a block, because we delegated to somebody.

Another example. We are not a leader, but leader delegated to us. `heavyWeAreIssuer` will be `True`. And we won't create a block. But we should do it unless we delegated further!

gromakovsky: *OLD COMMENT:*

I don't understand this check. I think we should check whether **we** delegated to somebody?!

Currently you check whether leader delegated to us. But I don't get it. For example, suppose we are leader and we delegated to somebody (not to us, obviously). Then `heavyWeAreIssuer` will be `False`. But in this case we shouldn't create a block, because we delegated to somebody.

Another example. We are not a leader, but leader delegated to us. `heavyWeAreIssuer` will be `True`. And we won't create a block. But we should do it unless we delegated further!

*UPDATE*:

I read it again more carefully. Again, I don't understand what's going on. We get a cert issued by leader and then compare `pskIssuerPk` with `ourPk`. But `pskIssuerPk` must be leader's PK. Am I saying nonsense?

gromakovsky: Also we shouldn't create a block if we lightly delegated to somebody, because it will lead to a fork. We can't enforce it by protocol, but we should enforce it by implementation.

core/Pos/Core/Types.hs
volhovM: Agree. I will do it in the next refactoring, noted.

src/Pos/Delegation/Logic.hs
```diff
@@ -590,18 +590,26 @@ delegationVerifyBlocks blocks = do
...
+                    sformat ("heavy proxy signature's "%build%" "%
+                             "related proxy cert can't be found/doesn't "%
+                             "match the one in current allowed heavy psks set")
                             pSig
+            (BlockPSignatureLight pSig) -> do
+                let pskIPk = pskIssuerPk $ pdPsk pSig
+                unless (pskIPk == issuer) $ throwError $
+                    sformat ("light proxy signature's "%build%" issuer "%
+                             build%" doesn't match block slot leader "%build)
+                            pSig pskIPk issuer
```
gromakovsky: Why not do this check in `verifyBBlockHeader`?

src/Pos/Delegation/DB.hs
```diff
@@ -119,19 +118,25 @@ isIssuerByAddressHash = fmap isJust . getPskByIssuer . Right
...
-        psks <- filter (\psk -> pskDelegatePk psk == dPk) . HM.elems <$>
-                getPskChain issuer
-        when (length psks > 1) $ throwM $ DBMalformed $
-            sformat ("getDlgTransPk: found several psks with a same dlg: "%listJson)
-                    psks
+        chain <- HM.elems <$> getPskChain issuer
+        let finalPsk = find (\psk -> pskDelegatePk psk == dPk) chain
+        let issuer' = toStakeholderId issuer
+        let iPk = pskIssuerPk <$>
+                  find (\psk -> addressHash (pskIssuerPk psk) == issuer') chain
```
gromakovsky: Do I understand correctly that it doesn't make sense if `issuer` is `Left`?

P. S. I know that it's always `Right` (actually it's used only in one place anyway).

src/Pos/Delegation/Helpers.hs
```diff
@@ -21,3 +34,70 @@ dlgVerifyPayload epoch (getDlgPayload -> proxySKs) =
...
+dlgReachesIssuance _ i d _ | i == d = pure True
+dlgReachesIssuance resolve i d psk = reach i
+  where
+    -- Delegate 'd' has right to issue block instead of issuer 'i' if
+    -- there's a delegation chain:
+    --
+    -- i → x₁ → x₂ → … xₖ → d
+    --
+    -- where every arrow is psk present in the 'pskm', and the last
+    -- psk xₖ → d has the same certificate as 'cert'.
```
gromakovsky: What is `pskm` and what is `cert`?

src/Pos/Delegation/Logic.hs
volhovM: I guess it'd be strange to very heavy and light signatures in different places, so i put light check to the place we had heavy already.

src/Pos/Delegation/DB.hs
volhovM: Yes, If somebody passes `Left pk`, we can just return `pk` but for the sake of compactnes we will still retrieve this `pk` from the beginning of the chain.

src/Pos/Delegation/Helpers.hs
volhovM: `cert` is a subfield of psk that _is_ actually the delegation signature. Comment is outdated, thanks.

core/Pos/Crypto/Signing.hs
```diff
@@ -1,4 +1,5 @@
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE IncoherentInstances #-}
```
int-index: Bad, bad extension

src/Pos/Delegation/DB.hs
```diff
@@ -1,101 +1,259 @@
...
+-- | Action on delegation database, used commonly. Generalizes
+-- applications and rollbacks.
+data DlgEdgeAction
+    = DlgEdgeAdd !ProxySKHeavy
+    | DlgEdgeDel !PublicKey
+    deriving (Show, Eq, Generic)
+
+instance Hashable DlgEdgeAction
+
+-- | Converts mempool to set of database actions.
```
gromakovsky: There is no mempool and no set here.

src/Pos/Delegation/Logic.hs
```diff
@@ -170,40 +134,305 @@ initDelegation = do
...
+    2. We're on u, next vertex is v, we know dlg(v). Set dlg(u) =
+    dlg(v) and apply it to all the traversal chain before. If dlg(v) =
+    Nothing, set dlg(u) = v instead.
+
+
+    Step 3 is trivial.
+    -}
+    transChangeset :: m TransChangeset
+    transChangeset = do
+        let keysMap = HS.fromMap . HM.map (const ())
```
gromakovsky: There is `Pos.Util.getKeys`, I will make use of it.

src/Pos/Delegation/Logic.hs
```diff
@@ -170,40 +134,305 @@ initDelegation = do
...
+
+    -- Once we figure out this piece of code works like a charm we
+    -- can delete this logging.
+    logDebug $ sformat ("Trans changeset: "%mapJson) $ HM.toList changeset
+
+    -- Bulid reverse transitive set and convert it to reverseTransOps
+    let reverseTrans = buildReverseTrans changeset
+
+    reverseOps <- forM (HM.toList reverseTrans) $ \(k, (ad0, dl0)) -> do
+        let ad = HS.toList ad0 -- view patterns are for the weak ones
```
gromakovsky: Why are you converting everything to list, then do operations like `intersect` and `\\` (which should be cheaper for set) and then convert back to set?

src/Pos/Delegation/DB.hs
```diff
@@ -1,101 +1,259 @@
...
+
+-- | Given an issuer, retrieves all certificate chains starting in
+-- issuer. This function performs a series of consequental db reads so
+-- it must be used under the shared lock.
+getPskChain
+    :: MonadDBPure m
+    => Either PublicKey StakeholderId -> m DlgMemPool
+getPskChain = getPskChainInternal HS.empty
+
+-- See doc for 'getPskTree'. This function also stops traversal if
```
gromakovsky: > See doc for 'getPskTree'.

Can't find it.

src/Pos/Delegation/Logic.hs
```diff
@@ -170,40 +134,305 @@ initDelegation = do
...
+
+----------------------------------------------------------------------------
 -- Heavyweight PSK
 ----------------------------------------------------------------------------
 
 -- | Retrieves current mempool of heavyweight psks plus undo part.
 getDlgMempool
-    :: (MonadIO m, MonadDBPure m, MonadDelegation m)
+    :: (MonadIO m, MonadDBPure m, MonadDelegation m, MonadMask m)
     => m (DlgPayload, [ProxySKHeavy])
```
gromakovsky: s/[ProxySKHeavy]/DlgUndo, I suppose.

PR #384, opened 2017-05-27
[CSL-780] Start moving GodTossing

neongreen:


godtossing/Pos/Ssc/GodTossing/SecretStorage.hs
```diff
@@ -6,12 +6,12 @@ module Pos.Ssc.GodTossing.SecretStorage
 
 import           Universum
 
-import           Pos.Binary.Ssc           ()
-import           Pos.DB                   (MonadDB)
-import           Pos.DB.Misc.Common       (miscGetBi, miscPutBi)
-import           Pos.Ssc.GodTossing.Core  (Opening, SignedCommitment)
-import           Pos.Ssc.GodTossing.Types (GtSecretStorage (..))
-import           Pos.Types                (EpochIndex)
+import           Pos.Binary.Ssc.GodTossing.Types ()
```
gromakovsky: Perhaps we can get rid of `Ssc` in `Pos.Binary.Ssc.GodTossing`, especially now when `GodTossing` is in a separate package.

neongreen: Okay.

neongreen: (I think that it'd be better to get rid of it once all or most of GodTossing is moved, though.)

neongreen:
The build failed due to OOM so I'm going to merge it with the big red button.

godtossing/Pos/Ssc/GodTossing/SecretStorage.hs
neongreen: Ah, sorry, I didn't notice that you said `Pos.Binary.Ssc.GodTossing`. Okay, I'll get rid of it in the next PR.

PR #385, opened 2017-05-29
{WIP} [CSL-780] Move more of GodTossing

neongreen:
* [x] Remove `Ssc` from `Pos.Binary.Ssc.GodTossing`
* [x] Move `genesisN` and dev keys to core/
* [x] Split `genesis.bin` into certificates and the rest of data (`genesis-general.bin` and `godtossing/genesis-gt.bin`; the latter has to be in `godtossing/` because otherwise Nix will probably break)
* [x] Move modules that depend on VSS genesis
* [x] Move other easily movable modules


src/keygen/Main.hs
```diff
@@ -142,38 +149,48 @@ reassignBalances GenesisData{..} = GenesisData
...
+    do let bin = encode genGtData
+           name = "godtossing" </> "genesis-gt.bin"
+           path = koGenesisDir </> name
+       liftIO $ createDirectoryIfMissing True (takeDirectory path)
+       case decodeFull bin of
+           Right (_ :: GenesisGtData) -> do
+               liftIO $ BSL.writeFile path bin
+               putText (toText name <> " generated successfully")
+           Left err ->
+               putText ("Generated GenesisGtData can't be read: " <> toText err)
```
gromakovsky: Copy-paste 😒

neongreen: I'll eliminate copy-paste when there are >2 generated files

gromakovsky:
Merge with 0 checks 😕

PR #386, opened 2017-05-29
[CSL-1121] Kademlia use external address

georgeee:


infra/Pos/DHT/Real/Real.hs
```diff
@@ -100,10 +100,10 @@ startDHTInstance kconf@KademliaDHTInstanceConfig {..} = do
     catchErrors x = liftIO x `catchAll` catchErrorsHandler
 
     log' logF =  usingLoggerName ("kademlia" <> "messager") . logF . toText
-    createKademlia host port key cfg =
-        K.createL host (fromIntegral port) key cfg (log' logDebug) (log' logError)
-    createKademliaFromSnapshot host port cfg snapshot =
-        K.createLFromSnapshot host (fromIntegral port)
+    createKademlia bA eA key cfg =
+        K.createL bA eA key cfg (log' logDebug) (log' logError)
+    createKademliaFromSnapshot bA eA cfg snapshot =
```
pva701: Longer names would be more understandable, imho

PR #387, opened 2017-05-29
Feature/csl937 body hash to header

7sedam7:
CSL-937 Add `Hash MainExtraBodyData` to `MainExtraHeaderData`


src/Pos/Block/Core/Main/Types.hs
```diff
@@ -76,6 +76,7 @@ data MainExtraHeaderData = MainExtraHeaderData
       _mehSoftwareVersion :: !SoftwareVersion
     , -- | Header attributes
       _mehAttributes      :: !BlockHeaderAttributes
+    , _mehEBDataProof     :: !(Hash MainExtraBodyData)
```
gromakovsky: Can you add a haddock comment for consistency?

src/Pos/Block/Logic/Creation.hs
```diff
@@ -46,7 +46,8 @@ import           Pos.Core                   (Blockchain (..), EpochIndex,
                                              SlotId (..), SlotLeaders, crucialSlot,
                                              epochOrSlot, flattenSlotId, getEpochOrSlot,
                                              getSlotIndex, headerHash, mkLocalSlotIndex)
-import           Pos.Crypto                 (SecretKey, WithHash (WithHash))
+import           Pos.Crypto                 (SecretKey, WithHash (WithHash), Hash)
+import           Pos.Crypto.Hashing         (hash)
```
gromakovsky: I think `hash` can also be imported from `Pos.Crypto` (without `Hashing`), but it's not really important.

georgeee: Right, it can be

7sedam7:
I don't know should I write here when changes are done, but I see that nobody commented on the changes so @gromakovsky

src/Pos/Block/Core/Main/Types.hs
```diff
@@ -76,7 +76,8 @@ data MainExtraHeaderData = MainExtraHeaderData
       _mehSoftwareVersion :: !SoftwareVersion
     , -- | Header attributes
       _mehAttributes      :: !BlockHeaderAttributes
-    , _mehEBDataProof     :: !(Hash MainExtraBodyData)
+    , -- | Body Hash
```
gromakovsky: It's not Body hash, it's hash of extra data, which is different from Body.

gromakovsky:
@7sedam7 I saw your commit, but you didn't add check that hash is correct as I wrote above:
> You should also check that hash from header matches `MainExtraBodyData` from body. The most appropriate place for it is `Pos.Block.Core.Main.Helpers.verifyBBlock`.

rockbmb:
@7sedam7 All you need to do now is merge `master` and CI will pass.

src/Pos/Block/Core/Main/Helpers.hs
```diff
@@ -36,12 +37,13 @@ instance ( BiHeader ssc
...
          BlockchainHelpers (MainBlockchain ssc) where
     verifyBBlockHeader = verifyMainBlockHeader
-    verifyBBlock UnsafeGenericBlock {..} = do
+    verifyBBlock block@UnsafeGenericBlock {..} = do
         either (throwError . pretty) pure $
             sscVerifyPayload @ssc
                 (Right (Some _gbHeader))
                 (_mbSscPayload _gbBody)
         dlgVerifyPayload (_gbHeader ^. epochIndexL) (_mbDlgPayload _gbBody)
+        unless (hash (block ^. gbExtra) == (block ^. mainBlockEBDataProof)) $ throwError ""
```
gromakovsky: What is it? Why empty message?

src/Pos/Block/Core/Main/Lens.hs
```diff
@@ -258,6 +264,11 @@ mainBlockHeaderAttributes ::
        Lens' (MainBlock ssc) BlockHeaderAttributes
 mainBlockHeaderAttributes = gbHeader . mainHeaderAttributes
 
+-- | Lens from 'MainBlock' to 'BlockHeaderAttributes'.
```
gromakovsky: Wrong comment.

PR #886, opened 2017-05-31
Fix csl-1129

naryl:
by creating an external bat script and terminating cardano-launcher
before running updater

### DIFF
--- a/src/launcher/Main.hs
+++ b/src/launcher/Main.hs
@@ -39,17 +39,18 @@ import           Pos.Reporting.Methods     (retrieveLogFiles, sendReport)
 import           Pos.ReportServer.Report   (ReportType (..))
 
 data LauncherOptions = LO
-    { loNodePath       :: !FilePath
-    , loNodeArgs       :: ![Text]
-    , loNodeLogConfig  :: !(Maybe FilePath)
-    , loNodeLogPath    :: !(Maybe FilePath)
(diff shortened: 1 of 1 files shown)


src/launcher/Main.hs
```diff
@@ -220,16 +227,45 @@ runUpdater (path, args, updateArchive) = do
...
+                 , Process.std_out = Process.CreatePipe
+                 , Process.std_err = Process.CreatePipe
+                 }
+    phvar <- liftIO newEmptyMVar
+    system' phvar cr mempty
+
+writeWindowsUpdaterRunner :: FilePath -> IO ()
+writeWindowsUpdaterRunner runnerPath =
+    writeFile (toString runnerPath) $ unlines
+        [ "TaskKill /IM cardano-launcher.exe /F"
```
georgeee: Is this necessary? Can't we make launcher close itself?

src/launcher/Main.hs
```diff
@@ -220,16 +227,45 @@ runUpdater (path, args, updateArchive) = do
...
+        -- Run updater
+        , "%*"
+        -- Delete updater
+        , "del %1"
+        -- Delete the bat file
+        , "(goto) 2>nul & del \"%~f0\""
```
georgeee: After running process, launcher with same options as originally should be run.

PR #887, opened 2017-05-31
Feature/csl1168 forbid rerevoking psks

volhovM:
* Now you can't post revocation cert if you have nothing to revoke
* You can now also revoke even if you don't have money (not a richman anymore)
* It's forbidden to sign blocks with revoke certificates

### DIFF
--- a/src/Pos/Block/Worker.hs
+++ b/src/Pos/Block/Worker.hs
@@ -34,6 +34,7 @@ import           Pos.Crypto                  (ProxySecretKey (pskDelegatePk, psk
 import           Pos.DB.Class                (MonadDBCore)
 import           Pos.DB.GState               (getDlgTransPsk, getPskByIssuer)
 import           Pos.DB.Misc                 (getProxySecretKeys)
+import           Pos.Delegation.Helpers      (isRevokePsk)
 import           Pos.Lrc.DB                  (getLeaders)
 import           Pos.Slotting                (currentTimeSlotting,
                                               getSlotStartEmpatically)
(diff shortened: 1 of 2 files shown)


src/Pos/Block/Worker.hs
```diff
@@ -112,7 +113,11 @@ blkOnNewSlotImpl (slotId@SlotId {..}) sendActions = do
         proxyCerts <- getProxySecretKeys -- TODO rename it with "light" suffix
         let validCerts =
                 filter (\pSk -> let (w0,w1) = pskOmega pSk
-                                in siEpoch >= w0 && siEpoch <= w1) proxyCerts
+                                in and [ siEpoch >= w0
+                                       , siEpoch <= w1
+                                       , not $ isRevokePsk pSk
+                                       ])
+                       proxyCerts
```
gromakovsky: I think it makes more sense to just prohibit putting revocation lighweight PSKs into MiscDB, but it's minor.

volhovM: It's already done. Just in case we get malformed misc db.

gromakovsky: Ok, great.

src/Pos/Delegation/Logic.hs
```diff
@@ -480,15 +480,19 @@ processProxySKHeavy psk = do
...
     runDelegationStateAction $ do
         memPoolSize <- use dwPoolSize
-        exists <- uses dwProxySKPool (\m -> HM.lookup issuer m == Just psk)
+        exists <- uses dwProxySKPool (\m -> HM.lookup iPk m == Just psk)
         cached <- isJust . snd . LRU.lookup msg <$> use dwMessageCache
-        alreadyPosted <- uses dwThisEpochPosted $ HS.member issuer
+        alreadyPosted <- uses dwThisEpochPosted $ HS.member iPk
         epochMatches <- (headEpoch ==) <$> use dwEpochId
+        hasPskInDB <- fmap isJust $ GS.getPskByIssuer (Left $ pskIssuerPk psk)
+        let rerevoke = isRevokePsk psk && hasPskInDB
```
pva701: I don't get this logic, why is it rerevokation?
Seems, when you just revoke the certificate, this statement (`isRevokePsk psk && hasPskInDB`) also will true.

volhovM: Where did my `not` go? :\

volhovM: Fixed

pva701: SPS BRATAN

PR #888, opened 2017-05-31
[CSM-249] More wallet renamings

Martoon-00:
Some stuff remained untouched or was renamed incorrectly.
It was already fixed in cardano docs, now fixing it here.


pva701:
YOU ARE THE WINNER!!!!
GETGETGETGET888

PR #892, opened 2017-06-01
Tests/alexrb/csl717 toss

rockbmb:


test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+        -- 'checkCommitmentsPayload' is called with the same epoch as the 'lastKnownEoS'
+        -- in 'gsVssCertificate'.
+        -- This is because rolling back slots is and should be tested elsewhere.
+        _gsVssCertificates <- VssCertData
+            <$> (pure . EpochOrSlot . Right . crucialSlot $ epoch)
+            <*> pure stableCerts
+            <*> arbitrary
+            <*> arbitrary
+            <*> arbitrary
+            <*> arbitrary
```
gromakovsky: I don't like the idea of creating `VssCertData` with violated internal invariants.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+
+        commsMap <- mkCommitmentsMapUnsafe . HM.fromList <$>
+                mapM (\k -> (,) <$> pure k <*> arbitrary) richmenWithComms
```
gromakovsky: I also don't like that invariants may be violated here.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+        _gsCommitments <- mkCommitmentsMapUnsafe <$> customHashMapGen
+            (arbitrary `suchThat` (not . flip elem richmenWithComms))
+            arbitrary
```
gromakovsky: And here.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+    arbitrary = GoodComms <$> do
+        -- These fields won't be needed for anything, so they can be entirely arbitrary.
+        _gsOpenings <- arbitrary
+        _gsShares <- arbitrary
+
+        -- The epoch used in the tests is generated separately to make sure it exists.
+        (epoch, m) <- arbitrary :: Gen (EpochIndex, MultiRichmenStake)
+        richmen <- getValid <$> (arbitrary :: Gen (ValidRichmenStake GenesisMpcThd))
+        let richmenIds = HM.keys richmen
+            mrs = HM.insert epoch richmen m
```
gromakovsky: I think it doesn't make sense to generate `mrs` with more than one epoch here. It may slow tests down a little bit.

If you really want to test that adding richmen for other epochs doesn't spoil anything, you can create `mrs` for two epochs, for example.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+    -> GtGlobalState
+    -> Bool
+emptyPayload pureToss mrs gtgs =
+    isRight $ tossRunner mrs gtgs $ pureToss mempty
+
+newtype GoodCommsPayload = GoodComms
+    { getGoodComms :: (EpochIndex, GtGlobalState, CommitmentsMap, MultiRichmenStake)
+    } deriving (Show, Eq)
+
+instance Arbitrary GoodCommsPayload where
```
rockbmb: @gromakovsky I don't think I mentioned this in the meeting, but I think `Arbitrary` instances used only in tests should stay in the modules where they're used. These ones are especially large, and I think it'd just be adding noise to the signal, to try and differentiate between actual `Arbitrary` instances and test ones.

If you disagree, I can move them there, no problem with that as it's also a viable solution - keeping these instances here also has its problems.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
rockbmb: This reply is for this comment and the following 2 as well:
Since the purpose of these datatypes is just to test the `check*Payload` functions, does it matter if some unrelated invariants are violated? They're unrelated to the checks performed in those functions, and therefore to the functions themselves.

If in the future we change these checking functions so that these invariants become relevant, it will make sense to change these `Arbitrary` instances to reflect this. Otherwise, is it really necessary?

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
gromakovsky: I was thinking about it before I wrote that comment and I found it questionable actually. On the one hand, you're right, these `Arbitrary` instances and helpers are used only in tests. On the other hand, I can imagine that some of these `Arbitrary` instances may be useful later to implement `Arbitrary` for some complex data type (which is used in other code, not some ad-hoc type for tests). And `Arbitrary` for this complex type may be needed not only in tests.

> try and differentiate between actual Arbitrary instances and test ones.

Ok, let's do it for now. Even if later we notice that we need some of these `Arbitrary` in places other than tests, we will move it, it shouldn't be a problem.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
gromakovsky: I think unsafe functions (e. g. unsafe constructors) should be avoided as much as possible. And if it's impossible, there should be comments explaining why this unsafe is safe.

> does it matter if some unrelated invariants are violated? 

Yes, I think it does for two reasons.
• Function may rely on the fact that invariants are not violated. It's an implementation detail whether it does or not. Even if you know it doesn't, you shouldn't rely on it if it's not written in the function's contract. If later someone changes implementation, they will have to also change these tests, even though they didn't break anything. But tests shouldn't start failing because of changes which don't break contract of the function.
• Currently we have `mkCommitmentsMapUnsafe` which is defined as `CommitmentsMap`, so it doesn't check anything. It assumes that the caller has already verified all input data. But later we may add a sanity check there and use `error` if it fails.

Apart from these two points, I would like to add that this way of constructing `VssCertData` looks strange to me anyway. What you want to create is `VssCertData` with given set of certificates and EoS. You can create an empty `VssCertData` (there is `empty`), use `setLastKnownEoS` to set EoS and then use `insert` to insert all certificates. This way you don't rely on the internal representation of `VssCertData` at all. You can't violate invariants of `VssCertData` and this code will compile if we change number of fields in `VssCertData` or change order of fields. Actually I don't like that `VssCertData`'s constructor is exported at all, but that's a separate question. And note that combining `empty`, `insert` and `setLastKnownEoS` isn't something hard to do.
For `CommitmentsMap` it's harder, because you need to know `SecretKey` to generate a `SignedCommitment` with given `StakeholderId` (which is needed to use `mkCommitmentsMap`). So you will need to write multiple generators. If it appears to be too much work, it can be ignored for now.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+                mkCommitmentsMapUnsafe .
+                -- over _2 (view _1) (a,(b,c)) = (a,b)
+                HM.fromList $ fmap (over _2 $ view _1) stakeHoldersAndCOs
+        openingPldList <- sublistOf stakeHoldersAndCOs
+
+        -- For the test data to be correct, none of the stakeholders with an opening in
+        -- openings payload can have an opening in the global state i.e. be a key in
+        -- '_gsOpenings'.
+        _gsOpenings <- customHashMapGen
+            (arbitrary `suchThat` (not . flip elem (map fst openingPldList)))
```
gromakovsky: I think it makes sense to convert `openingPldList` to a `HashSet` first and then use quick lookup, rather than `O(n)` for list. It applies to other places too.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+        sid' = addressHash . vcSigningKey $ cert
+        newerCertsMap = HM.insert sid' cert certsMap
+        certNoRichmen = tossRunner mrs gtgs $ checkCertificatesPayload epoch newerCertsMap
+        res3 = case certNoRichmen of
+            Left (CertificateNotRichmen nes) -> sid' `elem` nes
+            _ -> False
+
+    in not (HM.member sid' $ mrs HM.! epoch) ==> res1 && res2 && res3
+
+-- | Utility functions for this module
```
gromakovsky: You wrote a haddock comment to the `tossRunner` function. But it applies not to these functions, but to all functions below (currently there are 2 functions).

So it should be a regular comment. We often use such formatting:
```
-- Utility functions for this module
```

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+            tossRunner mrs' gtgs' $ checkSharesPayload epoch newestSharesMap
+        res4 = case sharesAlreadySent of
+            Left (SharesAlreadySent nes) -> sid `elem` nes
+            _ -> False
+
+    in not (HM.member sid _gsShares) && not (null sharesMap) ==>
+        res1 && res2 && res3 && res4
+
+newtype GoodCertsPayload = GoodCertsPayload
+    { getGoodCerts :: (EpochIndex, GtGlobalState, VssCertificatesMap, MultiRichmenStake)
```
gromakovsky: This is the fourth newtype for a tuple consisting of `(EpochIndex, GtGlobalState, payload, MultiRichmenStake)`.
Don't you think it would be better to have
```
data GoodPayload p = GoodPayload {
    gpEpoch :: !EpochIndex,
    gpGlobalState :: !GtGlobalState
…
}
```

And then `GoodCertsPayload = GoodPayload VssCertificatesMap` etc.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+        -- These fields of 'GtGlobalState' are irrelevant for the
+        -- 'checkCertificatesPayload' function.
+        _gsShares <- arbitrary
+        _gsOpenings <- arbitrary
+        _gsCommitments <- arbitrary
+
+        -- We'll need an 'EpochIndex' to run 'checkCertificatesPayload'. This epoch index
+        -- will also need an accompanying 'RichmenStake', but because we'll need the
+        -- public keys to generate valid 'VssCertificates' w.r.t.
+        -- 'checkCertificatesPayload', a list with public keys in tuples in generated
```
gromakovsky: Typo probably (_is_ generated).

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
rockbmb: You're right, there's a lot of refactoring potential in this module but I preferred just getting it done first, and then focus on that.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+    -> NonEmpty (AsBinary Share)
+    -> VssCertificate
+    -> Property
+checksBadSharesPayload
+    (getGoodShares -> (epoch, g@GtGlobalState {..}, sm, mrs))
+    sid
+    ne
+    cert =
+    let -- This property assumes the existence of a stakeholder not in the commitments or
+        -- shares map. Instead of writing a new 'Arbitrary' type which will only be useful
```
gromakovsky: > the existence of a stakeholder not in the commitments or shares map

It sounds strange to me, but maybe that's because I am not a native speaker. Perhaps it can be reformulated more clearly using word `absence` :wink:

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
```diff
@@ -0,0 +1,573 @@
...
+                       (gsVssCertificates %~ \vcd@VssCertData{..} ->
+                           vcd { certs = HM.insert sid cert certs})
+        mrs' = HM.update (Just . HM.insert sid (mkCoin 0)) epoch mrs
+        sharesAlreadySent =
+            tossRunner mrs' gtgs' $ checkSharesPayload epoch newestSharesMap
+        res4 = case sharesAlreadySent of
+            Left (SharesAlreadySent nes) -> sid `elem` nes
+            _ -> False
+
+    in not (HM.member sid _gsShares) && not (null sharesMap) ==>
```
gromakovsky: I find it strange that you have a check `not (HM.member sid _gsShares)` here and also there is `gtgs = g & (gsShares %~ HM.delete sid) `.

rockbmb: Oh, good catch. Essentially, I do not enforce the absence of any particular stakeholder in the `Arbitrary` instances, so if I want these tests that check failure to work I have to delete the `StakeholderId` passed as an argument to the property. Otherwise, if the stakeholder is, by chance, in the global state, one of the failures won't be triggered and the property won't work.

rockbmb:
@gromakovsky Agreed. I'll fix the style and cosmetic issues right now.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
rockbmb: I thought about that, but then I remembered what you told me almost a year ago when I wanted to use `IntSet`s instead of regular `Set Int`s for some trivial task: premature optimization is the root of all evil.

Don't get me wrong, you're right, `elem` is bad when compared to `O(log n)` lookup, especially since this hashmap may be quite large, which implies a large number of lookups to this possibly large list too.

I decided that I'd just do it this way quickly and move on with the issue (it was already overdue for several weeks) than to ponder for half an hour whether or not the `O(n * log n)` conversion would slow down tests enough to hang CI.

rockbmb:
@gromakovsky CI shouldn't have any problems with this, so you can merge when possible.

test/Test/Pos/Ssc/GodTossing/Toss/BaseSpec.hs
gromakovsky: Actually it's just one more line:

> openingPldSet = HS.fromList $ map fst openingPldList

And in general usage of `elem` for list is suspicious (sometimes it makes sense, but usually not). I wouldn't call it a premature optimization, because it's only a local variable inside this function and it's only 1 extra line of code.

gromakovsky:
https://issues.serokell.io/issue/CSL-1208

PR #893, opened 2017-06-02
[CSM-264] Decode endpoints arguments automatically

Martoon-00:
This introduces special argument specifiers of servant API similar to `QueryParam`, `Capture` and other, which decode many CTypes (e.g. `CPassPhrase -> PassPhrase`) on a fly and allows for cleaner and faster code in endpoints.

It's in only plans to use this feature for some types like `CId`, this will be done later I hope


src/Pos/Util/Servant.hs
```diff
@@ -0,0 +1,139 @@
...
+
+   -- | Way to encode to @CType@.
+   encodeCType :: ToOriginType c -> c
+
+
+-- | For arguments-specifiers of API, get argument type.
+type family ApiArg (apiType :: * -> *) a :: *
+type instance ApiArg (Capture s) a = a
+type instance ApiArg (QueryParam s) a = Maybe a
+type instance ApiArg (ReqBody ct) a = a
```
Martoon-00: @gromakovsky Need to write this code makes me think that there is smth I don't understand in this life.
Do you know where could such type family be defined?

Martoon-00:
I suppose this servant-api fun lacks of comprehensive explanation. Suggestions regarding comments are welcome

cardano-sl.cabal
```diff
@@ -120,6 +120,7 @@ library
                         Pos.Util.Arbitrary
                         Pos.Util.BackupPhrase
                         Pos.Util.JsonLog
+                        Pos.Util.Servant
```
gromakovsky: It should be included only if `with-web` is enabled.

src/Pos/Util/Servant.hs
gromakovsky: Last time I had fun with servant was more than a year ago (Feb 2016) and since then a lot of things changed. So unfortunately I don't know servant internals nowadays. I recommend you to ask in IRC. That's what they suggest here: https://github.com/haskell-servant/servant#getting-started

cardano-sl.cabal
Martoon-00: @akegalj wants it to be used in explorer some day as well.
Does it worth separate condition `if flag(with-web) || flag (with-explorer)`?

gromakovsky: I think `if flag(with-web) || flag (with-explorer)` should be ok for now.

src/Pos/Util/Servant.hs
Martoon-00: Thanks, I didn't know where to find all servant libraries listed. 
After a quick look I didn't manage to find anything similar, so let it be

cardano-sl.cabal
Martoon-00: Ok, fixed.
I just discovered that I'd broken build with `--no-wallet` before. Will deal with it tomorrow

gromakovsky: Keep in mind that #895 is not merged yet.

Martoon-00:
Fixed

cardano-sl.cabal
Martoon-00: Ah, rly.
Then everything here seems to work fine. All modules that depend on `Util.Servant` presence are built successfully with `--no-wallet`

Martoon-00:
I tested it sending JSON requests via Postman, api hasn't changed

Martoon-00:
I don't see any reason not to merge it.
If merge linger some more, it will produce not pleasant conflicts with new important features

PR #896, opened 2017-06-02
Feature/csl1171 dlg refactoring

volhovM:
* Split delegation logic into `Common`, `VAR` and `Mempool`
* Rename `ProxySignature` fields inappropriate `pd` prefix (it's `psig` now)
* Move `ProxySKEither` from core to `Delegation.Types`. It's `ProxySKBlockInfo` now also.


src/Pos/Delegation/Types.hs
```diff
@@ -68,3 +69,9 @@ type DlgMemPool = HashMap PublicKey ProxySKHeavy
 
 -- | Confirmation of light cert type.
 type ProxySKLightConfirmation = (ProxySKLight, ProxySigLight ProxySKLight)
+
+-- | Lightweight PSK or heavyweight PSK with real leader public key
+-- (because heavyweight psks have redelegation feature, so pskIssuerPk
+-- hPsk /= leader in general case). This is used to create a block
+-- header only.
+type ProxySKBlockInfo = Either ProxySKLight (ProxySKHeavy, PublicKey)
```
gromakovsky: I wrote it in older commit and I will write again.
This type is **always** used with `Maybe`. This commit confirms it. And it makes sense, because to sign block this thing is indeed optional (we can use regular secret key or we can use PSK). And if we add `Maybe` here, it will be too cumbersome type which should be `data`.
This type denotes the way to sign a block. And there are three possibilities. So why use `Maybe $ Either bla (patak, kek)`? Why not `SignRegular | SignProxyLight ProxySKLight | SignProxyHeavy { sphPSK :: !ProxySKHeavy, sphIssuerKey :: !PublicKey }`?

src/Pos/Delegation/Logic/Mempool.hs
```diff
@@ -0,0 +1,284 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
+
+-- | Transactions processing, in-mem state and mempool-related
```
gromakovsky: Transactions processing? 😒

src/Pos/Delegation/Types.hs
volhovM: 1. I don't really want to turn it into the datatype, because it'll make things more complicated IMO. More constructors, similar to `BlockSignature`. Eh. 
2. I thought about wrapping it into the newtype, but i didn't get the point. It's anyway wrapped only once and unwrapped once as well.
3. Wrapping it into `Maybe` makes sense to me.

src/Pos/Delegation/Logic/Mempool.hs
volhovM: I write words, 0 level thinking.

PR #883, opened 2017-05-30
[DEVOPS-107] Fix CI

domenkozar:
Superseeds https://github.com/input-output-hk/cardano-sl/pull/346

PR #898, opened 2017-06-03
[CSL-1173] Reduce the size of .hi-files

int-index:


domenkozar:
@int-index https://github.com/input-output-hk/cardano-sl/pull/883 includes these changes, enabled linux build again and turns on the haddock. All passing

nh2:
Could we document some before/after behaviour of this fix?

* The number of terms/types/coercions as output by `ghc -v`
* A directory tree listing showing the size of all `.hi` files and their text equivalent
* An excerpt from `relevant .hi-dump` files

PR #900, opened 2017-06-03
[CSL-780] Move SSC listeners/workers

neongreen:
See commit history.


src/Pos/Lrc/Consumers.hs
```diff
@@ -14,6 +18,8 @@ import           Pos.Ssc.Class.Workers (SscWorkersClass (sscLrcConsumers))
 import           Pos.Update.Lrc        (usLrcConsumer)
 
 allLrcConsumers
-    :: (LrcMode ssc m, SscWorkersClass ssc)
+    :: forall ssc m.
+       (LrcMode ssc m, SscWorkersClass ssc)
     => [LrcConsumer m]
-allLrcConsumers = [delegationLrcConsumer, usLrcConsumer] ++ untag sscLrcConsumers
+allLrcConsumers = [delegationLrcConsumer, usLrcConsumer] ++
+                  untag (sscLrcConsumers @ssc)
```
gromakovsky: It's strange to see usage of `@ssc` and `untag` at the same time. AFAIK, they should solve same problem.

ssc/Pos/Ssc/Mode.hs
```diff
@@ -0,0 +1,60 @@
...
+import           Pos.Ssc.Class.Types         (MonadSscContext)
+import           Pos.Ssc.Extra               (MonadSscMem)
+
+-- | Mode used for all SSC listeners, workers, and the like.
+type SscMode ssc m
+    = ( WithLogger m
+      , MonadIO m
+      , MonadMask m
+      , MonadSlots m
+      , Mockable M.CurrentTime m      -- TODO: if I use Mockables, GHC fails
```
gromakovsky: Use `MonadMockable` please. Btw, is GHC failure a bug in GHC? Maybe it's worth reporting it?

src/Pos/Lrc/Consumers.hs
neongreen: `Tagged` doesn't solve the problem here because `ssc` isn't mentioned in `=> [LrcConsumer m]`. So, I'm going to remove `Tagged` here completely because it's useless.

(Though previously it *did* solve the problem... no idea why.)

ssc/Pos/Ssc/Mode.hs
neongreen: It's a bug that is probably fixed in 8.2, and even if it's not it could likely be solved by tweaking some constraint-solving flags (which I didn't want to do).

neongreen:
Here's a cat to distract you from the fact that I merged it with the red button:

PR #902, opened 2017-06-07
[CSM-273] Add isOutgoing flag to returned transaction info

Martoon-00:
This also fixes smth in genesis HD addresses

### DIFF
--- a/src/Pos/Wallet/Web/ClientTypes.hs
+++ b/src/Pos/Wallet/Web/ClientTypes.hs
@@ -61,6 +61,7 @@ import           Control.Arrow          ((&&&))
 import qualified Data.ByteString.Lazy   as LBS
 import           Data.Default           (Default, def)
 import           Data.Hashable          (Hashable (..))
+import qualified Data.Set               as S
 import           Data.Text              (Text, isInfixOf, splitOn, toLower)
 import           Data.Text.Buildable    (build)
 import           Data.Time.Clock.POSIX  (POSIXTime)
@@ -164,8 +165,9 @@ mkCTx
     :: ChainDifficulty    -- ^ Current chain difficulty (to get confirmations)
     -> TxHistoryEntry     -- ^ Tx history entry
     -> CTxMeta            -- ^ Transaction metadata
+    -> [CId Addr]         -- ^ Addresses of wallet
     -> CTx
-mkCTx diff THEntry {..} meta = CTx {..}
+mkCTx diff THEntry {..} meta wAddrs = CTx {..}
   where
     ctId = txIdToCTxId _thTxId
     outputs = toList $ _txOutputs _thTx
@@ -174,6 +176,7 @@ mkCTx diff THEntry {..} meta = CTx {..}
     ctMeta = meta
     ctInputAddrs = map addressToCId _thInputAddrs
     ctOutputAddrs = map addressToCId _thOutputAddrs
+    ctIsOutgoing = not . null $ S.fromList wAddrs `S.intersection` S.fromList ctInputAddrs
 
 newtype CPassPhrase = CPassPhrase Text
     deriving (Eq, Generic)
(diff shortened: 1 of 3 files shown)


akegalj:
From https://issues.serokell.io/issue/CSM-273#comment=96-4165 :

Transactions can have list of transaction inputs and list of transaction outputs. We have to define is transaction incoming or outgoing. Lets see which use cases we have to cover. There are users A, B and C. We are user A. There are these scenarios:
```
 * user A sends money to B
     - A doesn't have change                         txInputs = [A], txOutputs = [B]        - outgoing
     - change is sent back to A                      txInputs = [A], txOutputs = [B, A]     - outgoing
 * user B sends money to A
     - B doesn't have a change                       txInputs = [B], txOutputs = [A]        - incoming
     - change is sent back to B                      txInputs = [B], txOutputs = [A, B]     - incoming
 * user A sends money to B and C
     - A doesn't have change                         txInputs = [A], txOutputs = [B, C]     - outgoing
     - change is sent back to A                      txInputs = [A], txOutputs = [B, C, A]  - outgoing
 * user B sends money to A and C
     - B doesn't have a change                       txInputs = [B], txOutputs = [A, C]     - incoming
     - change is sent back to B                      txInputs = [B], txOutputs = [A, C, B]  - incoming
 * user B and C send money to A
     - we are not supporting this feature atm
 * user B sends money to C
     - user A is not included so he doesn't care
```

@martoon is https://github.com/input-output-hk/cardano-sl/pull/902 working as this ? I think we didn't capture it right with my proposal:
```
* it is incoming transaction if any addresses from txOutputs is from from this wallet. so something like `elem txOutputs $ map getAddresses $ map getAccounts getAllWallets`
* it is outgoing otherwise  - this is a bit fuzzy as transaction can be input and output in the same time
```
because the change address for A will mark it as incoming, where it should be outgoing. This ^ proposed algorithm would work only if we take change addresses out of equation.

akegalj:
By looking at rules:
```
txInputs = [A], txOutputs = [B]        - outgoing
txInputs = [A], txOutputs = [B, A]     - outgoing

txInputs = [B], txOutputs = [A]        - incoming
txInputs = [B], txOutputs = [A, B]     - incoming

txInputs = [A], txOutputs = [B, C]     - outgoing
txInputs = [A], txOutputs = [B, C, A]  - outgoing

txInputs = [B], txOutputs = [A, C]     - incoming
txInputs = [B], txOutputs = [A, C, B]  - incoming

```
I can tell that this applies:
 * it is outgoing if it is txInputs
 * it is incoming otherwise

Martoon-00:
Yep, I did like in your second post, as it was AFAIR your last proposal yesterday

src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -391,11 +399,27 @@ data CTx = CTx
...
     } deriving (Show, Generic, Typeable)
 
 txContainsTitle :: Text -> CTx -> Bool
 txContainsTitle search = isInfixOf (toLower search) . toLower . ctmTitle . ctMeta
 
+-- TODO [CSM-288] Rename?
+-- | In case of A -> A tranaction, we have to return two similar 'CTx's:
+-- one with 'ctIsOutgoing' set to /true/, one with the flag set to /false/.
+-- This type gathers these two together.
+data CTxs = CTxs
```
akegalj: I think this solution is ok. We could have also just use `[CTx]` as we did before. Now we have `[CTxs]` in searchHistory but we can easily fallback to `catMaybes $ concatMap (\tx -> [ctsOutgoing, ctsIncoming]) [CTxs]` .

This way [CTxs] -> [CTx] we would get the same representation in getHistory and searchHistory. So before returning [CTxs] in returnHistory I would convert it to [CTx]. (because we already have information is it outgoing or incoming in CTx datatype as a flag)

what do you think @Martoon-00 ?

akegalj: Aha, seems to me that you are already doing this?

Martoon-00: Yep, `CTxs` is internal item only

PR #907, opened 2017-06-08
Tests/alexrb/csl1206 binary

rockbmb:


test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
             binaryTest @(B.FixedSizeInt Int)
             binaryTest @(B.FixedSizeInt Int64)
         describe "TinyVarInt" $ do
-            -- TODO: check that these actually take 1 byte
             describe "1 byte" $ do
+                it "serializing 0 takes 1 byte" $ serializeFail 0 1
                 prop "0" $ binaryEncodeDecode (B.TinyVarInt 0)
```
neongreen: I think it would make sense to change these lines to e.g. `"roundtrip 0"` instead of `"0"`

test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
...
+                it "serializing 1 takes 1 byte" $ serializeFail 1 1
                 prop "1" $ binaryEncodeDecode (B.TinyVarInt 1)
+                it "serializing 0x75 takes 1 byte" $ serializeFail 0x75 1
                 prop "0x75" $ binaryEncodeDecode (B.TinyVarInt 0x75)
+                it "serializing 0x79 takes 1 byte" $ serializeFail 0x79 1
                 prop "0x79" $ binaryEncodeDecode (B.TinyVarInt 0x79)
-            -- TODO: check that these actually take 2 bytes
             describe "2 bytes" $ do
+                it "serializing 0x80 takes 2 bytes" $ serializeFail 0x80 2
```
neongreen: Given that these all are really repetitive, perhaps writing helpers for them in a `let` would be better?

```
let test_roundtrip n_name n = prop n_name $ binaryEncodeDecode (B.TinyVarInt 1)
```

```
let test_length n_name n b = it ("serializing " ++ n_name ++ " takes " ++ show n ++ " bytes") $
        serializeFail n b
```

test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
...
                 prop "0x81" $ binaryEncodeDecode (B.TinyVarInt 0x81)
+                it "serializing 0x82 takes 2 bytes" $ serializeFail 0x82 2
                 prop "0x82" $ binaryEncodeDecode (B.TinyVarInt 0x82)
+                it "serializing 0x100 takes 2 bytes" $ serializeFail 0x100 2
                 prop "0x100" $ binaryEncodeDecode (B.TinyVarInt 0x100)
+                it "serializing 0x1000 takes 2 bytes" $ serializeFail 0x1000 2
                 prop "0x1000" $ binaryEncodeDecode (B.TinyVarInt 0x1000)
                 let maxnum = 2^(14::Int)-1
+                it "serializing (2^14)-1 takes 2 bytes" $
+                    B.biSize (B.TinyVarInt maxnum) `shouldBe` 2
```
neongreen: Why not `serializeFail` here?

Also, why is `serializeFail` called like it's called? The name seems non-intuitive to me.

test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
...
+                    shouldThrowException B.encode anyErrorCall n
+            describe "deserialize more than 2 bytes" $ do
+                prop "fails to deserialize more than 2 bytes" $ do
+                    let wordGen =
+                            generate $
+                                (arbitrary :: Gen Word8) `suchThat` (flip testBit 7)
+                    w1 <- wordGen
+                    w2 <- wordGen
+                    let bs =
+                            uncurry BS.append . bimap BS.singleton BS.singleton $ (w1, w2)
```
neongreen: This is just `BS.singleton w1 <> BS.singleton w2`.

(An even better option is `BS.pack [w1, w2]`.)

test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
...
-            -- TODO: check that deserializing an ambiguous 0 (i.e. @80 00@)
-            --       fails as well
+                prop "fails to serialize more than 2^14 - 1" $ do
+                    n <-  generate $ B.TinyVarInt <$> choose (maxnum + 1, maxBound)
+                    shouldThrowException B.encode anyErrorCall n
+            describe "deserialize more than 2 bytes" $ do
+                prop "fails to deserialize more than 2 bytes" $ do
+                    let wordGen =
+                            generate $
+                                (arbitrary :: Gen Word8) `suchThat` (flip testBit 7)
```
neongreen: `wordGen = flip setBit 7 <$> arbitrary`

test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
...
+                    w1 <- wordGen
+                    w2 <- wordGen
+                    let bs =
+                            uncurry BS.append . bimap BS.singleton BS.singleton $ (w1, w2)
+                    shouldThrowException (B.decode @B.TinyVarInt) anyErrorCall bs
```
neongreen: I don't think this is right. If it eats the second byte and throws a “not enough input” error, the test will pass even though it shouldn't. Instead, you should generate random bytestrings with 3+ bytes and check that `decodeFull` fails on them.

test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
...
+            describe "deserialize ambiguous zero" $ do
+                prop "fails to deserialize an ambiguous zero (i.e. @80 00@)" $ do
+                    word8 <- generate $
+                        (arbitrary :: Gen Word8) `suchThat` (flip testBit 7)
+                    let zero = '\NUL'
+                        bs = (BS.singleton word8) `BS.append` (singleton zero)
```
neongreen: Why not `BS.singleton word8 <> BS.singleton 0x00`?

test/Test/Pos/BinarySpec.hs
```diff
@@ -29,22 +38,51 @@ spec = describe "Bi" $ do
...
+                prop "fails to deserialize more than 2 bytes" $ do
+                    let wordGen =
+                            generate $
+                                (arbitrary :: Gen Word8) `suchThat` (flip testBit 7)
+                    w1 <- wordGen
+                    w2 <- wordGen
+                    let bs =
+                            uncurry BS.append . bimap BS.singleton BS.singleton $ (w1, w2)
+                    shouldThrowException (B.decode @B.TinyVarInt) anyErrorCall bs
+            describe "deserialize ambiguous zero" $ do
```
neongreen: Please rename to “deserialize inefficiently encoded data”.

test/Test/Pos/BinarySpec.hs
rockbmb: I did this, but then it felt that it'd be hard to understand what the property was without taking a look at the function that implements it. Plus, I did it as a top level binding, which means I had to use a type we don't use very often to represent the test, so I just undid that.

With a `let`, that problem disappears.

test/Test/Pos/BinarySpec.hs
rockbmb: I did that, and the test failed because that's not the behavior set by the implementation. I left this in the code I submitted for review so that if someone (ie you) noticed it, I could be sure the `get` for `TinyVarInt` needed to be changed. I'll do it.

core/Pos/Binary/Class.hs
```diff
@@ -218,7 +218,9 @@ getTinyVarInt' = do
     if testBit a 7
         then do
             b <- getWord8
-            if | testBit b 7 -> fail "getTinyVarInt': more than 2 bytes"
+            emptyBuf <- isEmpty
```
rockbmb: @neongreen if you don't think there's a problem with these lines I'll keep them here.

neongreen: `isEmpty` looks at the whole remaining buffer, so it will return `False` most of the time and mess the decoding up. I'll make a commit with the correct behavior, a sec.

PR #909, opened 2017-06-08
[CSL-1196] Add DiscoveryContextSum, refactor wallet

gromakovsky:
I added `DiscoveryContextSum` as a sum of all possible context used by 'MonadDiscovery' implementations. It allowed us to remove some boilerplate.

Also I refactored some stuff in `Pos.Wallet` (moved light part into `Pos.Wallet.Light`), because I couldn't understand anything in `Pos.Wallet`.


src/Pos/Context/Context.hs
```diff
@@ -108,6 +109,8 @@ data NodeContext ssc = NodeContext
     -- ^ Context needed for the update system
     , ncLrcContext          :: !LrcContext
     -- ^ Context needed for LRC
+    , ncDiscoveryContext    :: !DiscoveryContextSum
```
volhovM: Do I understand correctly that:
1. It's not used anywhere at all
2. It can be retrieved with `Ether.ask` anyway

gromakovsky: 1. It's used by `DiscoveryRedirect` which needs `Ether.MonadReader' DiscoveryContextSum` for `MonadDiscovery` instance.

2. Yes, it can be.

PR #911, opened 2017-06-08
[CSM-274] Create all wallets with initial account

Martoon-00:


Martoon-00:
This breaks everything, to be done and tested after #902 is merged

PR #913, opened 2017-06-08
[CSM-278] Search tx history by wallet ID

Pastafarianist:


src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -626,69 +626,89 @@ sendMoney sendActions passphrase moneySource dstDistr title desc = do
...
-    pure (paginate cHistory, fromIntegral $ length cHistory)
+getFullHistory :: WalletWebMode m => ObjectWithHistory -> m ([CTx], Word)
+getFullHistory owh = do
+    accIds <- case owh of
+        OWHWallet cWalId -> getWalletAccountIds cWalId
+        OWHAccount accId -> pure [accId]
+    accAddrs <- concatMapM (getAccountAddrsOrThrow Ever) accIds
+    addrs <- mapM (decodeCIdOrFail . cwamId) accAddrs
+    cHistory <- do
+        (mInit, cachedTxs) <- transCache <$> getHistoryCache owh
```
flyingleafe: Having a separate independent cache for wallet given that every account in wallet already have its own cache is space-consuming and error-prone. It would be better to leave caches only for account ids, and here, instead of getting cache by `ObjectWithHistory`, call a helper which aggregates cached histories for `accIds` and merges them.

Pastafarianist: Would you suggest merging them by converting `[CTx]` to `[Tx]` (somehow, no idea how exactly) and topsorting them, or in some other way?

flyingleafe: I think simple sorting by timestamp of creation will do. Such timestamps should be available in according `CTxMeta`s in wallet acid-state. Timestamp is stored as `POSIXTime`, so the probability of two transactions having the same creation time is very close to zero.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -626,69 +626,86 @@ sendMoney sendActions passphrase moneySource dstDistr title desc = do
...
+    mAccountId <- mapM decodeCAccountIdOrFail mCAccountId
+    cWalId <- case (mCWalId, mAccountId) of
+        (Nothing, Nothing)        -> throwM errorSpecifySomething
+        (Just _, Just _)          -> throwM errorDontSpecifyBoth
+        (Just cWalId', Nothing)   -> pure $ cWalId'
+        (Nothing, Just accountId) -> pure $ aiWSId accountId
+    first (applySkipLimit . filter fits) <$> getFullWalletHistory cWalId
   where
-    fits ctx = txContainsTitle search ctx
+    fits ctx = maybe True (containsInTitle ctx) mSearch
```
Martoon-00: you can `('txContainsTitle' ctx)`

src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -104,6 +104,7 @@ instance Default WalletStorage where
         , _wsWalletInfos  = mempty
         , _wsProfile      = def
         , _wsReadyUpdates = mempty
+        , _wsTxHistory = mempty
```
Martoon-00: stylish haskell negodue

src/Pos/Wallet/Web/Server/Methods.hs
Pastafarianist: Can you give a full example? I'm not familiar with Template Haskell.

Martoon-00: `maybe True ('txContainsTitle' ctx) mSearch`
it's not TH, it's just infix form of function )

src/Pos/Wallet/Web/Api.hs
```diff
@@ -257,20 +256,13 @@ type UpdateTx =
     :> ReqBody '[JSON] CTxMeta
     :> WRes Post ()
 
-type GetHistory =
```
flyingleafe: Are you sure that `getHistory` is really never used by Daedalus, for example?

src/Pos/Wallet/Web/Api.hs
```diff
@@ -257,20 +256,13 @@ type UpdateTx =
...
 type SearchHistory =
        "txs"
     :> "histories"
-    :> CCapture "walletId" CAccountId
-    :> Capture "search" Text
-    :> QueryParam "account" (CId Addr)
+    :> QueryParam "walletId" (CId Wal)
+    :> QueryParam "accountId" CAccountId
+    :> QueryParam "address" (CId Addr)
+    :> QueryParam "search" Text
```
flyingleafe: Such changes require according changes in `daedalus-bridge`

Pastafarianist: @flyingleafe `daedalus-bridge` builds successfully with this PR. Specifically, here is what I did to build it:

```
stack exec -- cardano-wallet-hs2purs
cd daedalus
npm install
npm run build:prod
```

There were no errors.

src/Pos/Wallet/Web/Api.hs
Pastafarianist: @flyingleafe no. However, the [search](https://github.com/input-output-hk/daedalus/search?utf8=%E2%9C%93&q=getHistory&type=) in the repository yields no results, and `daedalus-bridge` can be built successfully. Is there anything I'm missing here?

src/Pos/Wallet/Web/Api.hs
flyingleafe: @Pastafarianist unfortunately, purescript-bridge is not as tightly interconnected with haskell code, and you won't detect mismatches in server api and bridge methods in compile time. 
`src/Daedalus/BackendApi.purs:184` contains `searchHistory` method, which now doesn't comply with modified API method in Haskell code.

src/Pos/Wallet/Web/Api.hs
flyingleafe: @Pastafarianist anyway, if it does, you can now rewrite method `getHistory` in bridge (`src/Daedalus/BackendApi.purs:181`) in such a way that it uses `searchHistory` api method without search query.

PR #914, opened 2017-06-08
[CSM-282] Don't add default wallet in non- wallet-debug mode

Martoon-00:


Martoon-00:
Need to check that it's ok when launched in the way frontenders use

PR #915, opened 2017-06-09
Add flag to indicate a change address

ajnsit:


src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -272,6 +272,7 @@ data CAddress = CAddress
     { cadId     :: !(CId Addr)
     , cadAmount :: !CCoin
     , cadIsUsed :: !Bool
+    , cadIsChange :: !Bool -- ^ Is this a change address
```
akegalj: we have formatter that, when applied, formats the code. Mine is applied on every file save in the project. Ask in serokell general channel for details how to set it up

Martoon-00:
Well, as far as I can see, effectively, you added flag to `getWAddress`, and sometimes pass `False` to it, sometimes pass `True` (it doesn't depend neither on user input nor on stored information about wallets).
All actual endpoints always pass `False` there, so nothing actually changed )) 
(under _actual_ I mean these endpoints https://github.com/input-output-hk/cardano-sl/blob/a1174e4690f9507d9934faa20eda265a9f516bfb/src/Pos/Wallet/Web/Server/Methods.hs#L295).

Could you explain for me which address should be considered _change_ address?

akegalj:
@Martoon-00 how it doesn't depend on user's api? When user makes a new payment this will be triggered https://github.com/input-output-hk/cardano-sl/pull/915/files#diff-309f72d4bc7dc71a30a34bbda61ef9bfR574 if there is a change address. So it will set `True` to a change address field in `newPayment` . This is how I see it. Right?

Martoon-00:
Aha, okay, I wasn't correct here.
But still, the fact of whether this address is change address is not stored in database, just look at how `isChangeAddress` argument is used in `getWAddress'` and `newWAddress'`. @ajnsit @akegalj do you agree?

If it makes sense for you, I suppose, the easiest way to add this information to database it to add some `_wsChangeAddresses :: Set CWAddressMeta` here:
https://github.com/input-output-hk/cardano-sl/blob/976d42a7a109214d7df67db1721c9974e3c0fc0a/src/Pos/Wallet/Web/State/Storage.hs#L90
and functions which reflect correspondent database requests

ajnsit:
Thanks @akegalj for the clarification! @Martoon-00 you are correct the flag isn't being stored in the db right now. I will add those bits in.

ajnsit:
@akegalj @Martoon-00 Should I merge this in now?

src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -93,6 +96,7 @@ data WalletStorage = WalletStorage
     , _wsReadyUpdates :: [CUpdateInfo]
     , _wsTxHistory    :: !(HashMap (CId Wal) TransactionHistory)
     , _wsHistoryCache :: !(HashMap (CId Wal) (HeaderHash, Utxo, [TxHistoryEntry]))
+    , _wsChangeAddresses :: CAddresses
```
Martoon-00: Please, make them aligned (and in other places as well). Most of us have `stylish-haskell` applied automatically on save which does this, I recommend )

Martoon-00:
Cool, thanks!

PR #918, opened 2017-06-09
[CSL-1167] Report error in applyBlocksUnsafe if passed more than epoch

volhovM:


src/Pos/Block/Logic/Internal.hs
```diff
@@ -111,8 +114,12 @@ type BlockApplyMode ssc m
 applyBlocksUnsafe
     :: forall ssc m . BlockApplyMode ssc m
     => OldestFirst NE (Blund ssc) -> Maybe PollModifier -> m ()
-applyBlocksUnsafe blunds0 pModifier =
-    reportingFatal version $
+applyBlocksUnsafe blunds0 pModifier = reportingFatal version $ do
+    when (length blunds /= length blunds0) $ assertionFailed $
```
gromakovsky: I don't like how it was done. This `spanSafe` is confusing. This function does something only if `blunds0 == blunds`. So `spanSafe` must return `(blunds0, [])`. So why not just check that all blocks have same epoch as the first one (head)? Also these names complicate understanding.

volhovM: I can instead check that `b` is null in  `(_,b) = spanSafe ...`. Will it be better?

gromakovsky: Yes, it's better than current approach and I will approve if you do it this way. But also can you rename `blunds0` to `blunds` and `b` to `blundsOtherEpoch` or something more descriptive?

volhovM: Done.

PR #919, opened 2017-06-09
[CSL-1174] Get rid of wallet-specific transformers in `RawRealMode`

gromakovsky:


src/Pos/Wallet/Light/Web/Server.hs
```diff
@@ -2,145 +2,117 @@
...
-walletServeWebLite _ sendActions dbPath dbRebuild port =
-    bracketWalletWebDB dbPath dbRebuild $ \db ->
-        bracketWalletWS $ \conn -> do
-            let runner = runWalletWebDB db . runWalletWS conn
-            let hoistedSA :: SendActions (WalletWebHandler LightWalletMode)
-                hoistedSA = hoistSendActions (lift . lift) runner sendActions
-            let action :: WalletWebHandler LightWalletMode Application
-                action = walletApplication $ walletServer hoistedSA nat
-            runner $ walletServeImpl action port
+walletServeWebLite __sendActions __dbPath __dbRebuild __port =
```
pva701: It makes me sad because we can't test update/tx/etc now.

src/Pos/Wallet/Web/Server/Full/R.hs
```diff
@@ -52,9 +54,11 @@ runWRealMode
 runWRealMode db conn =
     runRealBasedMode
         unwrapWPMode
-        (lift . lift)
+        -- doesn't work
+        -- (powerLift @(RealMode WalletSscType) @WalletRealWebMode)
+        (liftWalletRedirects . lift . lift)
```
pva701: why not `coerce`?

gromakovsky: `liftWalletRedirects` is just an alias for `coerce`. 
`WalletWebSockets` is `Ether.ReaderT' ConnectionsVar`. `WalletWebDB` is `Ether.ReaderT' WalletState`.

I can't use `coerce` to lift something into reader.

PR #922, opened 2017-06-11
Fix/csl1209 resourcet without forkio

volhovM:
This fixes terrible critical bug.


volhovM:
IT will be reopened after bug is indeed fixed.

PR #924, opened 2017-06-11
[CSL-1105] Split slotting implementations akin to Discovery

gromakovsky:
Previously we had slotting implementation based on `ReaderT (Bool, NtpSlottingVar)`. It was bad for two reasons:
• `(Bool, NtpSlottingVar) doesn't make sense, it should be `Maybe NtpSlottingVar` or something equivalent.
• We couldn't use simple slotting separately. But we want to use it in tests.
Also it was strange to see simple slotting in NTP module.

In order to fix it I've chosen approach similar to `Discovery` implementations. So we have different transformers and we have a wrapper which stores sum of all contexts and chooses the appropriate one. Also I removed workers from type class, because otherwise I wouldn't be able to get all workers without monadic context and because workers require much more constraints. There is a function which chooses appropriate workers based on contexts sum.

Note that I had to add `SlottingContextSum` as an argument in many places. I don't like it and I think nobody does. I will improve it within another issue. So please don't complain about this argument.

### DIFF
--- a/infra/Pos/Slotting/Impl/Ntp.hs
+++ b/infra/Pos/Slotting/Impl/Ntp.hs
@@ -1,18 +1,27 @@
-{-# LANGUAGE DataKinds           #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TemplateHaskell     #-}
-{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE DataKinds       #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies    #-}
 
--- | 'MonadSlots' implementation which uses Ntp servers.
+-- | NTP-based implementation of slotting.
 
-module Pos.Slotting.Ntp
-       ( NtpSlottingState
+module Pos.Slotting.Impl.Ntp
+       (
+         -- * State
+         NtpSlottingState
        , NtpSlottingVar
+
+         -- * Mode
+       , NtpMode
+       , NtpWorkerMode
+
+         -- * MonadSlots, redirects, etc.
        , askNtpSlotting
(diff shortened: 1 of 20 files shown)


infra/Pos/Slotting/Impl/Ntp.hs
```diff
@@ -65,64 +92,78 @@ type NtpSlottingVar = TVar NtpSlottingState
...
-type SlotsRedirect =
-    Ether.TaggedTrans SlotsRedirectTag IdentityT
+----------------------------------------------------------------------------
+-- MonadSlots implementation
+----------------------------------------------------------------------------
 
-runSlotsRedirect :: SlotsRedirect m a -> m a
-runSlotsRedirect = coerce
+-- | Monad which implements NTP-based solution for slotting.
+-- Flag means whether to use real NTP servers or rely on local time.
```
pva701: seems, this comment is outdated

infra/Pos/Slotting/Impl/Ntp.hs
```diff
@@ -211,99 +252,20 @@ ntpGetCurrentSlotBlocking = ntpGetCurrentSlotImpl >>= \case
...
 -- Worker for synchronization of local time and global time.
 ntpSyncWorker
-    :: (SlottingConstraint m, MonadNtpSlotting m)
-    => m ()
-ntpSyncWorker =
-    ifNtpUsed (askNtpSlotting >>= void . startNtpClient . ntpSettings) pass
+    :: NtpWorkerMode m
+    => NtpSlottingVar -> m ()
+ntpSyncWorker = void . startNtpClient . ntpSettings
+    -- ifNtpUsed (askNtpSlotting >>= void . startNtpClient . ntpSettings) pass
```
pva701: remove it plz

PR #928, opened 2017-06-12
[CSM-291] Add `DefaultApiArg`

Martoon-00:
This allows to nicely support optional parameters of requests by means of servant-api

### DIFF
--- a/src/Pos/Util/Servant.hs
+++ b/src/Pos/Util/Servant.hs
@@ -12,17 +12,22 @@ module Pos.Util.Servant
     , VerbMod
 
     , ReportDecodeError (..)
-    , CDecodeArg
+    , CDecodeApiArg
 
+    , OriginType
     , FromCType (..)
     , ToCType (..)
+
+    , WithDefaultApiArg
 
     , CQueryParam
     , CCapture
     , CReqBody
+    , DCQueryParam
     ) where
 
 import           Control.Monad.Except (ExceptT (..))
+import           Data.Default         (Default (..))
 import           Formatting           (formatToString, sformat, shown, stext, string, (%))
 import           Servant.API          ((:>), Capture, QueryParam, ReqBody, Verb)
 import           Servant.Server       (Handler (..), HasServer (..), ServantErr, Server)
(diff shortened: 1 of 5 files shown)


Martoon-00:
Looks like no one actually can review it, because during last couple of months (except last week or two) I was mostly the only one who worked in backend, and no one delved (or enjoyed the result) much of these servant tricks.

To justify this PR I can say that this instance was removed:
https://github.com/input-output-hk/cardano-sl/pull/928/files#diff-7fc892f6182c1909c46af1c7bb826809L260
That's not much, but previous approach required, for given pair of types `Xxx` and `Cxxx`, to define `FromCType` for both `Cxxx` and `Maybe Cxxx` in case if `Cxxx` is used as `Capture` and `QueryParam` at the same time, which will become nasty boilerplate if we decide to use these tricks for other client types

gromakovsky:
> Looks like no one actually can review it

~It's bad if people who are working on the code can't review it as long as they have eyes (and if they don't have eyes it's even worse, but for other reasons). The goal of review is not only to check whether everything is correct, but also to understand what's going on and to ask questions if you don't understand. Because if you don't understand something, it's better to ask questions in PR than to get stuck later when you need to modify something there.~

Okay 😞 

> if we decide to use these tricks for other client types

Well, I don't have enough expertise to understand how likely it is, so I will assume that later you will use these tricks more and approve.

Martoon-00:
Well, if look at this from perspective of learning the codebase, I agree, would be nice if someone else also review.
This time I looked at review usefulness only from position of finding this change reasonable, and you're the only one who followed the history of these tricks and has knowledge of internals. But this logic now seems strange even to me :slightly_smiling_face:

akegalj:
@ajnsit @Pastafarianist  can you take a look at this as well. Its some magic that @Martoon-00 introduced. It might be hard to get it well at first but take your time to grasp whats going on here and ask @Martoon-00 questions if something is not clear or to point you to some links where you can learn more about these techniques that he used here.

@Martoon-00 is your mentor - ask him questions about this PR. For example I don't have much hands on experience with type families and other techniques that he used. But its useful to share this knowledge across the team

akegalj:
Lets not merge it until @ajnsit and @Pastafarianist get some understanding of this codebase @Martoon-00

src/Pos/Util/Servant.hs
```diff
@@ -93,31 +98,29 @@ instance (HasServer (Verb mt st ct $ ApiModifiedRes mod a) ctx,
...
+type family OriginType ctype :: *
 
+class FromCType c where
    -- | Way to decode from @CType@.
-   decodeCType :: c -> Either Text (FromOriginType c)
+   decodeCType :: c -> Either Text (OriginType c)
 
--- This is not used for now
 class ToCType c where
-   -- | Same as 'FromOriginType'. For some types they actually differ,
```
Martoon-00: Yes, previously the difference was in Maybe only, because there were mandatory and optional parameters in request, we wanted to specify special value in the latter case when parameter is not set (e.g. `emptyPasspharse`, if passphrase is not specified)

ksaric:
@Martoon-00 if I got this right, these type-level constructs you introduced are used for the transformation of the query parameters.
They remove some of the boilerplate, like using `Maybe` in your instance definitions.

But at the end of the day, didn't you just push the same functionality, which is (query) parameter transformation one layer deeper?

Weren't there **Servant** constructs like `FromHttpApiData` and `ToHttpApiData`? Couldn't you extend these definitions to connect them and effectively extend Servant with custom (automatic/semi-automatic) serializations of parameters?

Sorry, I don't know much about this, just trying to understand.

Martoon-00:
You propose, instead of `FromCType` and `ToCType` just use existing `FromHttpApiData` and `ToHttpApiData` type classes, do I understand it correctly?

This is worth discussing.
But if we go this way, client types passed request body won't be supported (this requires `ToJSON` and `FromJSON` instances rather than `FromHttpApiData` and `ToHttpApiData`). Can't say is it really needed though

Pastafarianist:
I somewhat understand what's going on. I know nothing about the internals of Servant, and hence I understand the part starting at https://github.com/input-output-hk/cardano-sl/pull/928/files#diff-a048dee31900132c54e3588128756b14R165 very poorly. I understand that the justification for this PR is https://github.com/input-output-hk/cardano-sl/pull/928/files#diff-a048dee31900132c54e3588128756b14R121 and the changes at https://github.com/input-output-hk/cardano-sl/pull/928/files#diff-7fc892f6182c1909c46af1c7bb826809L274 and https://github.com/input-output-hk/cardano-sl/pull/928/files#diff-7fc892f6182c1909c46af1c7bb826809L229. But I don't see the bigger picture just yet.

ajnsit:
@Pastafarianist I also don't understand servant internals, but regarding the part at https://github.com/input-output-hk/cardano-sl/pull/928/files#diff-a048dee31900132c54e3588128756b14R165. -

We have basically created a type wrapper called `WithDefaultApiArg` that allows us to provide a different instance for servant's `HasServer` class. In the handler for that instance (defined in the code from line 182 onwards) we define the type of the `Server` (i.e. the handler) to have the type `a -> b` even though the type allowed by Servant in the requests is defined as `Maybe a -> b`. If a Nothing is supplied by the request, a default value is provided to the handler (we declare the type to have a `Default` instance).

While it's easy to construct a type of `Maybe a` from an arbitrary type `a` in Haskell, it's not as easy to construct an `a` from a `Maybe a` because there is no guarantee that the supplied type is a `Maybe`. We need such a type function to work on arbitrary types because we apply it to an arbitrary type `ServerT a` . To do this we have `UnmaybeArg` which is a closed type family and is only defined for `Maybe` wrapped values. This allows the compiler to give an error if used for any other type.

HTH

PR #929, opened 2017-06-12
[CSM-279] Don't remove spent wallet addresses

Martoon-00:
This also implements `isUsed` wallet address field properly

### DIFF
--- a/src/Pos/Wallet/Web/Tracking.hs
+++ b/src/Pos/Wallet/Web/Tracking.hs
@@ -273,7 +273,6 @@ applyModifierToWSet
     -> m ()
 applyModifierToWSet wAddr newTip mapModifier = do
     -- TODO maybe do it as one acid-state transaction.
-    mapM_ WS.removeWAddress (MM.deletions mapModifier)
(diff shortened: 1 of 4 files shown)


Martoon-00:
Guys?

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -273,7 +273,6 @@ applyModifierToWSet
     -> m ()
 applyModifierToWSet wAddr newTip mapModifier = do
     -- TODO maybe do it as one acid-state transaction.
-    mapM_ WS.removeWAddress (MM.deletions mapModifier)
```
flyingleafe: Simply ignoring deletions seem like dirty hack -- summoning @pva701 for an opinion about how to do it better.

flyingleafe: @pva701 told that it's okay though

Martoon-00: Why not? Addresses deletions occur only when address is spent (from blockchain's point of view). That's exactly what we don't want to do since now.

PR #931, opened 2017-06-12
Brunjlar/tw 150

brunjlar:
Moved JSON logging from cardano-sl to time-warp-nt.


src/Pos/Launcher/Runner.hs
```diff
@@ -199,6 +203,7 @@ runRealModeDo discoveryCtx slottingCtx transport np@NodeParams {..} sscnp
                runProduction .
                    runResourceT .
                    usingLoggerName lpRunnerTag .
+                   runJsonLogT' mJLHandle .
```
gromakovsky: @brunjlar I see that `JsonLogT` stores `Maybe (MVar Handle, JLTimedEvent -> IO Bool)` internally. So, as far as I understand, it means that here both monad stacks will use same handle but different `MVar`, so it may happen that they will print to same handle concurrently. I find it suspicious.
If `ekg` doesn't need JSON logging, then why not use `runWithoutJsonLogT`? If it needs JSON logging, then how do we solve the problem I described?

brunjlar: @gromakovsky good point, I think you're right, this is indeed a problem. We never ran into this problem during benchmarking, so I'm sure ekg monitoring doesn't need json logging, so using `runWithoutJsonLogT` sounds like the right solution.

brunjlar: @gromakovsky I changed it to `runWithoutJsonLogT` and created a pull request: https://github.com/input-output-hk/cardano-sl/pull/937

PR #943, opened 2017-06-14
[CSM-297] Catch from importWallet nicely

Martoon-00:
Only `File does not exist` now is reported as `RequestError`, all other exceptions are `InternalError`s

### DIFF
--- a/src/Pos/Util/UserSecret.hs
+++ b/src/Pos/Util/UserSecret.hs
@@ -26,8 +26,8 @@ module Pos.Util.UserSecret
        , ensureModeIs600
        ) where
 
-import           Control.Exception          (onException)
 import           Control.Lens               (makeLenses, to)
+import           Control.Monad.Catch        (handle, onException)
 import           Data.Binary.Get            (label)
 import qualified Data.ByteString.Lazy       as BSL
 import           Data.Default               (Default (..))
@@ -36,6 +36,7 @@ import qualified Prelude
 import           Serokell.Util.Text         (listJson)
 import           System.FileLock            (FileLock, SharedExclusive (..), lockFile,
                                              unlockFile, withFileLock)
+import           System.IO.Error            (isDoesNotExistError)
 import qualified Turtle                     as T
 import           Universum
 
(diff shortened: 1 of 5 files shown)


src/Pos/Wallet/Web/Error/Util.hs
```diff
@@ -36,8 +33,11 @@ catchEndpointErrors action = catchOtherError $ tryWalletError action
...
-    catchOtherError = handleAll $ \e -> do
-        usingLoggerName logName $
-            logError $ sformat ("Uncaught error in wallet method: "%shown) e
-        throwM err500
+    catchOtherError = flip catches
+        [ Handler $ throwM @_ @ServantErr
+        , Handler $ \(SomeException e) -> do
+            usingLoggerName logName $
+                logError $ sformat ("Uncaught error in wallet method: "%shown) e
+            throwM err500
```
gromakovsky: I think error body should be customized by putting formatted error into it. Here is an example: https://www.stackage.org/haddock/lts-8.13/servant-server-0.9.1.1/Servant-Server.html#v:err500

src/Pos/Util/UserSecret.hs
```diff
@@ -176,14 +177,18 @@ initializeUserSecret secretPath = do
 
 -- | Reads user secret from file, assuming that file exists,
 -- and has mode 600, throws exception in other case
-readUserSecret :: (MonadIO m, WithLogger m) => FilePath -> m UserSecret
-readUserSecret path = do
+readUserSecret :: (MonadIO m, MonadCatch m, WithLogger m) => FilePath -> m UserSecret
```
gromakovsky: To be honest, I don't understand why this function throws wallet-specific errors and why this module imports something from `Pos.Wallet.Web`. For me this function looks like quite general function which doesn't have to be used by wallet's web server. And I see that it's currently used by `src/keygen/Avvm.hs`, for example. It's quite strange to see `WalletError` (especially if one knows that `WalletError` is mostly for web part) in keygen. Also in principle I think it's ok to use this function in regular full node (not wallet), because it also needs to store some secret somewhere (I don't know much details).

So I think this function shouldn't catch anything internally and it should be done at higher level. Exceptions like `DoesNotExistError` are the most reasonable for this function.

src/Pos/Wallet/Web/Error/Util.hs
```diff
@@ -38,6 +38,6 @@ catchEndpointErrors action = catchOtherError $ tryWalletError action
         , Handler $ \(SomeException e) -> do
             usingLoggerName logName $
                 logError $ sformat ("Uncaught error in wallet method: "%shown) e
-            throwM err500
+            throwM err500 { errBody = "Internal error occured, please report this bug." }
```
gromakovsky: Why don't you include `e` into this message? It's not better than the previous one.

PR #945, opened 2017-06-14
[CSL-1123] Reorganize scripts, remove unnecessary ones, renaming.

denisshevchenko:


domenkozar:
I think you'll need to fix `.travis.yaml` as well

denisshevchenko:
@domenkozar Fixed. Thank you!

scripts/build/daedalus-bridge.sh
```diff
@@ -0,0 +1,21 @@
...
+set -o pipefail
+
+echo "Building Daedalus Bridge..."
+
+# Cleaning node modules, bower components, etc.
+./scripts/clean/daedalus-bridge.sh
+
+# We have to build Cardano SL because there's a code in Haskell
+# codebase that needs to be recompiled for Bridge building.
+./scripts/build/cardano-sl.sh
```
gromakovsky: It's better to pass `sl` argument here for two reasons:
• it should be faster than `b` without arguments, because in this case packages will be built in parallel;
• new reason: after recent PR (which was merged this night) we also have `lwallet` which shouldn't be needed for daedalus.

scripts/clean/db.sh
```diff
@@ -1,9 +1,15 @@
...
 #     https://cardanodocs.com/technical/wallets/
 
 # TODO: support purescript API rebuilding?
 
-rm -rf ./run/*
+echo "Cleaning Cardano SL db..."
+
+rm -rf run/*
 rm -rf wallet-db/
 rm -rf db-abc/
```
gromakovsky: I don't like this `db-abc` name. It's obscure. Can we have a better name for it?

gromakovsky:
@denisshevchenko what do you think about all of this?

scripts/clean/db.sh
denisshevchenko: Yes, we can. This name `abc` is really old (and you're right, it's strange name). So what name can we use here?

denisshevchenko:
> Also I would like to see description of what these scripts do. Maybe as comments in scripts, maybe in some separate file.

@gromakovsky as comments in scripts, of course.

scripts/clean/db.sh
gromakovsky: These are all `abc`s in scripts:
```
scripts/clean/db.sh:rm -rf db-abc/
scripts/clean/db.sh:rm -rf kademlia-abc.dump
scripts/launch/qa.sh:    --log-config scripts/log-templates/log-config-abc.yaml \
scripts/generate/genesis.sh:function abc {
scripts/generate/genesis.sh:abc 2>&1 | tee genesis.info
```

So `db-abc` appears only in clean script and hence it is useless. Same for `kademlia-abc.dump`.

`qa.sh` uses `db-qanet` and `kademlia-qanet.dump`. I think these names are reasonable. But cleaning script should be fixed. Also there is `log-config-abc.yaml` and this name it bad.

PR #959, opened 2017-06-15
[CSL-1201] Guard the state modifying STM action with a mutex

arybczak:
Version of #953 for cardano-sl-0.5-rc.

### DIFF
--- a/src/Pos/WorkMode.hs
+++ b/src/Pos/WorkMode.hs
@@ -17,6 +17,8 @@ module Pos.WorkMode
 import           Universum
 
 import           Control.Monad.Fix
+import           Control.Monad.Base             (MonadBase)
+import           Control.Monad.Trans.Control
 import qualified Control.Monad.Trans.Lift.Local as Lift
 import           Data.Coerce
 import           Data.Tagged                    (Tagged)
@@ -78,17 +80,19 @@ type RealMode' ssc =
     LoggerNameBox Production
     ))))))))))
 
-newtype RealMode ssc a = RealMode (RealMode' ssc a)
+newtype RealMode ssc a = RealMode { unRealMode :: RealMode' ssc a }
   deriving
     ( Functor
     , Applicative
     , Monad
     , MonadIO
+    , MonadBase IO
     , MonadThrow
     , MonadCatch
     , MonadMask
     , MonadFix
     )
+
 type instance ThreadId (RealMode ssc) = ThreadId Production
 type instance Promise (RealMode ssc) = Promise Production
 type instance SharedAtomicT (RealMode ssc) = SharedAtomicT Production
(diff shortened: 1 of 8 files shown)


src/Pos/WorkMode.hs
```diff
@@ -17,6 +17,8 @@ module Pos.WorkMode
 import           Universum
 
 import           Control.Monad.Fix
+import           Control.Monad.Base             (MonadBase)
+import           Control.Monad.Trans.Control
```
gromakovsky: [Our style guide](https://github.com/input-output-hk/cardano-sl-style-guides/blob/master/haskell-style-guide.md#imports) prohibits implicit imports.

txp/Pos/Txp/MemState/Class.hs
```diff
@@ -73,11 +76,15 @@ getMemPool = getTxpLocalData (STM.readTVar . txpMemPool)
 getTxpExtra :: (MonadIO m, MonadTxpMem e m) => m e
 getTxpExtra = getTxpLocalData (STM.readTVar . txpExtra)
 
+txpLocalDataLock :: MVar ()
+txpLocalDataLock = unsafePerformIO $ newMVar ()
```
gromakovsky: Is it really necessary to use global variable here? Can't it be added to reader?

txp/Pos/Txp/MemState/Class.hs
```diff
@@ -91,13 +98,15 @@ modifyTxpLocalData f =
         STM.writeTVar txpTip newTip
         STM.writeTVar txpExtra newExtra
         pure res
+ where
+   withLock = L.bracket_ (takeMVar txpLocalDataLock) (putMVar txpLocalDataLock ())
```
gromakovsky: We already have two ways of doing `bracket`: `MonadMask` and `Mockable Bracket` (which is bad and we are planning to leave only one). It doesn't make sense to add the third one.

src/Pos/WorkMode.hs
arybczak: Alright, I didn't know it existed. Do you list and modify them by hand or use some sort of automation (afair GHC can generate explicit import lists, but I don't remember exactly how)? If it's the former, that sounds quite inefficient.

txp/Pos/Txp/MemState/Class.hs
arybczak: As I mentioned in #953, this function is instantiated with transformer stack that contains `ExceptT`, hence `MonadMask` won't work here. Perhaps `Mockable Bracket` will.

src/Pos/WorkMode.hs
gromakovsky: Currently we do it by hand, but we have a group of people (actually only two people) working on a separate utility to facilitate work with imports.

> If it's the former, that sounds quite inefficient.

It makes writing code less efficient, but reading code more efficient. The latter is usually more important.

PR #965, opened 2017-06-15
[CSM-271] Detect change addresses from the transaction history instead of from the DB

ajnsit:


src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -395,16 +395,21 @@ getWAddressBalance addr =
...
-            (Just $ addrMetaToAccount cAddr)  -- just to specify addrId is not enough
-            (Just cwamId)
+            (Just addrAccount)  -- just to specify addrId is not enough
+            (Just aId)
             Nothing
     let isUsed = not (null ctxs) || balance > minBound
-    CAddress cwamId (mkCCoin balance) isUsed <$> isChangeAddress cAddr
+    -- Suppose we have transaction with inputs A1, A2, A3. Output address B_j is referred to as change address if it belongs to same account as one of A_i
+    acctAddrs <- map cwamId <$> getAccountAddrsOrThrow Ever addrAccount
+    let isChange = elem aId $ filter (`elem` acctAddrs) $ concatMap ctOutputAddrs ctxs
```
Martoon-00: 1. `filter (`elem` acctAddrs)` seems redundant because it's implied by `elem aId` - if `x == aId`, then `x` indeed belongs to the set of all ever existed accounts `acctAddrs`.
2. This seem not to work with example above. For instance, if `cAddr` is `B_j` from example, then `aId` indeed belongs to `concatMap ctOutputAddrs ctxs` which is set of outputs of all related transactions, but we want `cAddr` to be considered _change_ only if it participates as input of some transaction as well

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -395,16 +395,22 @@ getWAddressBalance addr =
...
-            (Just cwamId)
+            (Just addrAccount)  -- just to specify addrId is not enough
+            (Just aId)
             Nothing
     let isUsed = not (null ctxs) || balance > minBound
-    CAddress cwamId (mkCCoin balance) isUsed <$> isChangeAddress cAddr
+    -- Suppose we have transaction with inputs A1, A2, A3. Output address B_j is referred to as change address if it belongs to same account as one of A_i
+    acctAddrs <- map cwamId <$> getAccountAddrsOrThrow Ever addrAccount
+    let containsChangeAddr ctx = aId `elem` (ctOutputAddrs ctx) && not (null $ filter (`elem` acctAddrs) (ctInputAddrs ctx))
+    let isChange = not $ null $ filter containsChangeAddr ctxs
```
Martoon-00: Can be replaced with `any`

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -395,16 +395,22 @@ getWAddressBalance addr =
...
-            (Just $ addrMetaToAccount cAddr)  -- just to specify addrId is not enough
-            (Just cwamId)
+            (Just addrAccount)  -- just to specify addrId is not enough
+            (Just aId)
             Nothing
     let isUsed = not (null ctxs) || balance > minBound
-    CAddress cwamId (mkCCoin balance) isUsed <$> isChangeAddress cAddr
+    -- Suppose we have transaction with inputs A1, A2, A3. Output address B_j is referred to as change address if it belongs to same account as one of A_i
+    acctAddrs <- map cwamId <$> getAccountAddrsOrThrow Ever addrAccount
+    let containsChangeAddr ctx = aId `elem` (ctOutputAddrs ctx) && not (null $ filter (`elem` acctAddrs) (ctInputAddrs ctx))
```
Martoon-00: `('elem' acctAddrs)` for each `ctInputAddrs ctx` will work for too long here, maybe put `actAddrs` to Set?

Martoon-00:
Looks very good, except some small issues

src/Pos/Wallet/Web/Server/Methods.hs
Pastafarianist: Also, you can replace `not (null $ filter ...)` here with `any` as well.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -391,16 +390,23 @@ getWAddressBalance :: WalletWebMode m => CWAddressMeta -> m Coin
 getWAddressBalance addr =
     getBalance <=< decodeCIdOrFail $ cwamId addr
 
+
```
Pastafarianist: Please remove this extra blank line.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -391,16 +390,23 @@ getWAddressBalance :: WalletWebMode m => CWAddressMeta -> m Coin
...
         getHistory
             Nothing
-            (Just $ addrMetaToAccount cAddr)  -- just to specify addrId is not enough
-            (Just cwamId)
+            (Just addrAccount)  -- just to specify addrId is not enough
+            (Just aId)
     let isUsed = not (null ctxs) || balance > minBound
-    CAddress cwamId (mkCCoin balance) isUsed <$> isChangeAddress cAddr
+    -- Suppose we have transaction with inputs A1, A2, A3. Output address B_j is referred to as change address if it belongs to same account as one of A_i
+    acctAddrs <- (S.fromList . map cwamId) <$> getAccountAddrsOrThrow Ever addrAccount
```
Martoon-00: Extra brackets
(but if you find it more convenient to read, remain it as is)

src/Pos/Wallet/Web/Server/Methods.hs
ajnsit: Ah yeah, that was me trying to coax github into updating the PR (github was having issues yesterday).

PR #966, opened 2017-06-15
[CSL-1217] Fix 'PollState' test

rockbmb:
@gromakovsky this fix is needed in every branch, including both `master` and `cardano-sl-0.5-rc`. Can it be merged to `master` first and then cherrypicked onto wherever else it's needed?


gromakovsky:
Does it solve the problem? Do you know the seed which triggered failure before?

> Can it be merged to master first and then cherrypicked onto wherever else it's needed?

It is more correct to make this PR for `-0.5-rc`. It will be eventually merged to master.

PR #967, opened 2017-06-16
[CSL-1154] ExecMode base monad

int-index:
This PR solves the built-time performance issues caused by Ether 0.5 migration. The temporary hack (disabled specialization) now safely removed, so run-time performance should be improved as well.

Highlights:

* The monad transformer stacks are now flat, consisting of a single `ExecMode`. Before we had `IdentityT` layers, so even if the run-time representation was flat, at compile-time the types were syntactically huge, choking the compiler.
* `HasLens` instances are now auto-generated. Yay TemplateHaskell.


src/Pos/Util/JsonLog.hs
```diff
@@ -77,3 +89,30 @@ showHash = sformat hashHexF
...
+jsonLogConfigFromHandle :: MonadIO m => Handle -> m JsonLogConfig
+jsonLogConfigFromHandle h = do
+    v <- newMVar h
+    return $ JsonLogConfig v (\_ -> return True)
+
+jsonLogReal
+    :: (ToJSON a, Ether.MonadReader' JsonLogConfig m, Mockable Catch m,
+        MonadIO m, WithLogger m)
+    => a -> m ()
+jsonLogReal x = do
```
gromakovsky: Copy-paste 😒

src/Pos/WorkMode.hs
```diff
@@ -11,83 +12,119 @@ module Pos.WorkMode
...
+import           Pos.Ssc.Class.Types         (SscBlock)
+import           Pos.Ssc.Extra               (SscMemTag, SscState)
+import           Pos.Txp.MemState            (GenericTxpLocalData, TxpHolderTag)
+import           Pos.Util                    (Some (..))
+import           Pos.Util.JsonLog            (JsonLogConfig, jsonLogReal)
+import           Pos.Util.TimeWarp           (CanJsonLog (..))
+import           Pos.Util.Util               (PowerLift (..))
+import           Pos.WorkMode.Class          (MinWorkMode, TxpExtra_TMP, WorkMode)
+
+data RealModeContext ssc = RealModeContext
```
gromakovsky: Hm, but you wanted to make it newtype over a big tuple to avoid writing all this `HasLens` boilerplate :/

gromakovsky: Also currently it's absolutely unclear why we can't put all of this (perhaps except `rmcPeerState`) into `NodeContext`.

src/Pos/Util/JsonLog.hs
int-index: Yep. Unfortunately, this code isn't exported as a function, so I had to copy it.

src/Pos/WorkMode.hs
gromakovsky: But I am reviewing per-commit, so maybe things will change later.

int-index: This is a WIP branch. I haven't rewritten the context data types to tuples. Actually, it won't be that simple, because with tuples I won't be able to embed one context into another.

I think the best way to go is to write some TH that would generate those records together with `HasLens` instances.

src/Pos/ExecMode.hs
```diff
@@ -0,0 +1,77 @@
...
+import           Mockable                    (ChannelT, Counter, Distribution, Gauge,
+                                              MFunctor' (..), Mockable (..), Promise,
+                                              SharedAtomicT, SharedExclusiveT, ThreadId)
+import           System.Wlog                 (CanLog)
+
+import           Pos.Util.Util               (PowerLift (..))
+
+type family ExecModeM mode :: * -> *
+
+newtype ExecMode mode a = ExecMode { unExecMode :: ExecModeM mode a }
```
gromakovsky: Do we still need this wrapper? Is its purpose to decrease resources consumption by compiler or something else?

int-index: The `mode` parameter to this type determines the instances. For each mode there's a tag:

```
data REAL ssc -- real mode
data LW -- light wallet
data WWEB -- wallet web mode
data WEB -- web mode
```

So, the way to think about this type is that it's a base redirection layer. The alternative would be to use a `TaggedTrans` here, but this newtype indeed hides the type to decrease resource consumption.

int-index: Furthermore, it has a special `Ether.MonadReader` instance that doesn't do propagation. It's safe because `ExecMode` is always the base layer, and this allowed me to delete the `Tags` instances for `NodeContext` (which were rather atrocious).

src/Pos/Client/Txp/Balances.hs
```diff
@@ -65,39 +67,47 @@ type BalancesRedirect =
 runBalancesRedirect :: BalancesRedirect m a -> m a
 runBalancesRedirect = coerce
 
+type BalancesMonad ext m =
```
gromakovsky: Previously we had two `MonadBalances`, now we have two `MonadBalances` and one `BalancesMonad` :/

src/Pos/Launcher/Mode.hs
```diff
@@ -0,0 +1,168 @@
...
+data InitRef a = InitRef
+    { initRefVar :: MVar a
+    , initRefVal :: a }
+
+makeLensesFor [ ("initRefVal", "initRefValL") ] ''InitRef
+
+newInitRef :: IO (InitRef a)
+newInitRef = do
+    m <- newEmptyMVar
+    r <- unsafeInterleaveIO (readMVar m)
```
gromakovsky: It definitely needs some comments. Also we put haddock comment at the beginning of each module.

src/Pos/Launcher/Resource.hs
```diff
@@ -117,55 +114,33 @@ hoistNodeResources nat nr =
...
     -> SscParams ssc
-    -> m $ NodeResources ssc m
-allocateNodeResources np@NodeParams {..} sscnp = do
+    -> Production (NodeResources ssc Production)
+allocateNodeResources np@NodeParams {..} sscnp = runInitMode $ do
+    putInitRef imcNodeParams np
+    putInitRef imcGenesisUtxo (GenesisUtxo npCustomUtxo)
+    putInitRef imcGenesisLeaders (GenesisLeaders (genesisLeaders npCustomUtxo))
     db <- openNodeDBs npRebuildDb npDbPathM
-    -- TODO: usage of 'runResourceT' here is strange.
```
gromakovsky: I have no idea how this comment survived.

src/Pos/Launcher/Resource.hs
```diff
@@ -117,55 +114,33 @@ hoistNodeResources nat nr =
...
-    let runDBAction = flip Ether.runReaderT' db . runDBPureRedirect
-    let genLeaders = genesisLeaders npCustomUtxo
-    -- DB initialization is hard.
```
gromakovsky: DB initialization is still hard.

update/Pos/Update/Poll/RollTrans.hs
```diff
@@ -77,6 +77,8 @@ instance MonadPoll m => MonadPoll (RollT m) where
         insertIfNotExist id unChangedPropsL getProposal
         deactivateProposal id
 
+    setSlottingData = lift . setSlottingData
```
gromakovsky: He-he, it could be bug :)

Better ask someone when you such suspicious places. I am not sure whether my intent was to not write this definition because I want it to look like this.
(maybe it didn't look suspicious to you, dunno)

Anyway I will check whether this logic is correct.

gromakovsky: Right, it is a bug. I will create an issue.

infra/Pos/Discovery/Holders.hs
```diff
@@ -112,27 +118,23 @@ data DiscoveryContextSum
...
+    (MonadDiscoverySum m, DiscoveryKademliaEnv m)
+
+getPeersReal :: DiscoveryRealMonad m => m (Set NodeId)
+getPeersReal =
+    Ether.ask' >>= \case
+        DCStatic nodes -> runDiscoveryConstT nodes getPeers
+        DCKademlia inst -> runDiscoveryKademliaT inst getPeers
+
+findPeersReal :: DiscoveryRealMonad m => m (Set NodeId)
+findPeersReal =
```
gromakovsky: I don't like suffix `Real` in this module. `DiscoveryKademlia` and `DiscoveryConstT` are as real as their sum. Why not change `Real` to `Sum`. What does this `Real` actually mean?

infra/Pos/Communication/PeerState.hs
```diff
@@ -60,34 +59,36 @@ peerStateFromSnapshot (PeerStateSnapshot snapshot) = do
...
+    atomically $ nodeId `STM.delete` m
 
-    clearPeerState nodeId = Ether.ask @PeerStateTag >>= \m -> atomically $ nodeId `STM.delete` m
-    getAllStates = Ether.ask @PeerStateTag >>= \m -> do
-        stream <- atomically $ LT.toList $ STM.stream m
-        PeerStateSnapshot <$> forM stream (mapM readSharedAtomic)
+getAllStatesReal
+    :: PeerStateRealMonad m
+    => m PeerStateSnapshot
+getAllStatesReal = do
```
gromakovsky: I don't like `Real` in this module too, but less than for discovery. It's quite strange to see `Real` given that it's based on `Mockable SharedAtomic` (so in principle at least part of logic can be emulated). What does this `Real` mean here?

infra/Pos/Slotting/Impl/Sum.hs
```diff
@@ -48,18 +53,33 @@ askSlottingContextSum = Ether.ask'
 
 type SlottingModeSum m = (NtpMode m, SimpleSlottingMode m)
 
+type SlotsRealMonad m =
```
gromakovsky: Same as above, don't understand this `Real`.

src/Pos/Client/Txp/Balances.hs
```diff
@@ -42,62 +39,52 @@ class Monad m => MonadBalances m where
...
     map (txOutValue . toaOut) . toList <$> getOwnUtxo addr
 
-data BalancesRedirectTag
+type BalancesMonad ext m =
+    (MonadRealDB m, MonadDBRead m, MonadGState m, MonadMask m, WithLogger m, MonadTxpMem ext m)
 
-type BalancesRedirect =
-    Ether.TaggedTrans BalancesRedirectTag IdentityT
+getOwnUtxosWebWallet :: BalancesMonad ext m => [Address] -> m Utxo
+getOwnUtxosWebWallet addr = do
```
gromakovsky: 1. I don't like `BalancesMonad` name, but I don't have a good alternative. Maybe `MonadBalancesBase`?
2. I don't like `WebWallet` in names. This module doesn't know anything about web wallet. And this function are not specific to web wallet. They can be used by stuff like smart-generator or whatever crazy utility.

src/Pos/Client/Txp/History.hs
```diff
@@ -195,57 +181,58 @@ instance
...
-                cachedUtxo
-
-            let lastCachedHash = fromMaybe bot $ head cachedHashes
-            return $ TxHistoryAnswer lastCachedHash (length cachedTxs) cachedUtxo result
-
-        maybe (error "deriveAddrHistory: Nothing") pure mres
-
+    )
+
+getTxHistoryWebWallet
```
gromakovsky: Basically same comments as for `Balances` module.

src/Pos/DB/Block.hs
```diff
@@ -247,19 +254,40 @@ runBlockDBRedirect = coerce
...
 
 -- instance MonadBlockDBGeneric (SscBlock ssc)
 
+dbGetBlockReal' :: forall ssc m . BlockDBGenericRealMonad ssc m => HeaderHash -> m (Maybe (SscBlock ssc))
+dbGetBlockReal' = fmap (toSscBlock <$>) . getBlock
+
+dbGetUndoReal' :: forall ssc m . BlockDBGenericRealMonad ssc m => HeaderHash -> m (Maybe ())
+dbGetUndoReal' = fmap (const () <$>) . getUndo
+
+dbGetHeaderReal' :: forall ssc m . BlockDBGenericRealMonad ssc m => HeaderHash -> m (Maybe (Some IsHeader))
```
gromakovsky: I think these functions should have `ssc` in their names, because their are basically for `ssc`. At least the first two.

gromakovsky: If we add such function for `txp`, will we call them `dbGetBlockReal''`? Hopefully no.

src/Pos/ExecMode.hs
```diff
@@ -0,0 +1,82 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
+{-# LANGUAGE TypeFamilies        #-}
+
+module Pos.ExecMode
```
gromakovsky: I am very sad that this module doesn't have any comments. Maybe that's because it's WIP.

src/Pos/Launcher/Resource.hs
gromakovsky: (earlier of course)

src/Pos/Util/JsonLog.hs
gromakovsky: If it gets merged with this copy-paste, create an issue about it at least.

src/Pos/Wallet/Web/Mode.hs
```diff
@@ -0,0 +1,185 @@
...
+    getSystemStart = getSystemStartReal
+    getSlottingData = getSlottingDataReal
+    waitPenultEpochEquals = waitPenultEpochEqualsReal
+    putSlottingData = putSlottingDataReal
+
+instance MonadSlots WalletWebMode where
+    getCurrentSlot = getCurrentSlotReal
+    getCurrentSlotBlocking = getCurrentSlotBlockingReal
+    getCurrentSlotInaccurate = getCurrentSlotInaccurateReal
+    currentTimeSlotting = currentTimeSlottingReal
```
gromakovsky: Do you have ideas how to avoid writing it, because it's already defined for `RealMode` and it's basically a super mode of `RealMode`?

src/Pos/Wallet/Web/Server/Full/Common.hs
```diff
@@ -70,40 +59,29 @@ convertHandler
...
-           . runBListenerStub
-           $ act
+    walletRunner :: forall a . WalletWebMode a -> IO a
+    walletRunner act = runProduction $ do
+        peerStateCtx <- peerStateFromSnapshot psCtx
+        Mtl.runReaderT (unWalletWebMode act) $
+            WalletWebModeContext
+                ws
+                conn
+                (RealModeContext
```
gromakovsky: I suppose it should become unnecessary to pass so much stuff to this function and recreate already existing `RealNodeContext`. Probably it's not done because it's WIP.

gromakovsky:
Write more comments please (in general).
Maybe you were planning to write them in the end, I don't know.

src/Pos/Client/Txp/Balances.hs
int-index: Yes, the naming is horrible. Please help.

infra/Pos/Discovery/Holders.hs
int-index: "Real" because it's the implementation used in the "real mode". I know, these names are bad, but I didn't want to think about them too much at a proof-of-concept phase. Do suggest a better alternative if you can think of one.

src/Pos/Wallet/Web/Mode.hs
int-index: I have a few thoughts, but maybe we should follow Snoyman's advise and keep this code stupid (at the cost of some boilerplate). I can investigate more type-heavy and less boilerplate alternatives later. The situation with Ether taught me that we must keep the types syntactically small, so the usual tricks I tend to use won't work.

infra/Pos/Discovery/Holders.hs
gromakovsky: Well, I proposed changing `Real` to `Sum`. Why not? E. g. `findPeersSum`.

infra/Pos/Communication/PeerState.hs
gromakovsky: And here it can be `getAllStatesDefault`, for example. Because that's just the simplest possible implementation I guess.

src/Pos/Util/JsonLog.hs
```diff
@@ -77,3 +89,33 @@ showHash = sformat hashHexF
 -- | Returns event of created 'Block'.
 jlAdoptedBlock :: SscHelpersClass ssc => Block ssc -> JLEvent
 jlAdoptedBlock = JLAdoptedBlock . showHash . headerHash
+
+-- FIXME: definitions below were copied (with slight modifications)
```
volhovM: Why not fix the export scope in tw? It's not that hard i believe.

gromakovsky:
Please don't merge it until `-0.5-rc` is merged.

src/Pos/ExecMode.hs
```diff
@@ -0,0 +1,149 @@
...
+-- transformer stacks.
+newtype ExecMode mode a = ExecMode { unExecMode :: ExecModeM mode a }
+
+makePrisms ''ExecMode
+
+deriving instance Functor (ExecModeM mode) => Functor (ExecMode mode)
+deriving instance Applicative (ExecModeM mode) => Applicative (ExecMode mode)
+deriving instance Monad (ExecModeM mode) => Monad (ExecMode mode)
+deriving instance MonadIO (ExecModeM mode) => MonadIO (ExecMode mode)
+deriving instance (Monad b, MonadBase b (ExecModeM mode)) => MonadBase b (ExecMode mode)
```
int-index: It should, but GHC doesn't see it for some reason — simply removing the `Monad b` constraint here doesn't work (I tried).

src/Pos/Launcher/Mode.hs
```diff
@@ -0,0 +1,125 @@
+{-# LANGUAGE ScopedTypeVariables #-}
```
pva701: @int-index can you write a clarifying comment about `InitMode`?

int-index: Added some comments.

src/Pos/ExecMode/Context.hs
```diff
@@ -0,0 +1,120 @@
...
+
+{- | Generate a mode context record with 'HasLens' instances.
+
+The 'modeContext' function takes a declaration like this:
+
+@
+    data PatakCtx ssc = PatakCtx
+        !(Buba    ::: Buba)
+        !(HuhaTag ::: Huha ssc)
+        !Bardaq
```
gromakovsky: When should I use `:::` and when not?

gromakovsky:
```
Notice that types annotated with '(:::)' get turned into fields. The rest
of the fields are inherited from the parent (a field without a tag annotation)
using an overlappable instance. Only one parent is allowed.
```

I see it but is there a rule of thumb?

src/Pos/Util/JsonLog.hs
gromakovsky: LOL :peka:

src/Pos/ExecMode/Context.hs
int-index: To add a field to the context, use `:::`, it's like record fields, but tags instead of field names.

If you want to extend an existing context (like `RealModeContext` extends `NodeContext`), don't use `(:::)`. The limitation is that you can use it at most once.

src/Pos/Util/JsonLog.hs
int-index: I fixed this.

PR #969, opened 2017-06-16
[CSM-307] Catch all types of errors

Pastafarianist:


src/Pos/Wallet/Web/Error/Util.hs
```diff
@@ -35,7 +37,8 @@ catchEndpointErrors action = catchOtherError $ tryWalletError action
   where
     tryWalletError
         | isDevelopment = try
-        | otherwise     = tryJust $ \e -> (e ^? _RequestError) $> e
+        | otherwise     = tryJust $ guarded $ \e ->
+            has _RequestError e || has _InternalError e || has _DecodeError e
```
Martoon-00: Isn't it just `try`?

PR #970, opened 2017-06-16
[CSL-1189] Restrict nonempty passphrase again

Martoon-00:
Use only 32-byte length pasphrases.
Default is set to `replicate 32 0`

### DIFF
--- a/core/Pos/Binary/Crypto.hs
+++ b/core/Pos/Binary/Crypto.hs
@@ -196,17 +196,11 @@ instance (Bi w) => Bi (ProxySignature w a) where
 
 instance Bi PassPhrase where
     put pp = do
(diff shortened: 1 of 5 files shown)


Martoon-00:
Need to regenerate genesis.bin

rockbmb:
@Martoon-00 this makes tests hang again. It really is because the empty passphrase being `BS.replicate 32 0` causes a massive slowdown. Maybe @neongreen can help here.

Martoon-00:
Loool, I hoped it will work this time

akegalj:
@rockbmb @Martoon-00 what is the state of this ?

Martoon-00:
Perhaps it's worth investigating how absence of empty passphrase influences on tests performance, some day.
It's quite cosmetical, right?

PR #972, opened 2017-06-16
[CSL-1217] Fix 'PollState' test

rockbmb:
@gromakovsky changed the base. It works, the seed is 999417177, and to run the tests do the following:
`util-scripts/build.sh; stack test --test-arguments '--match "Pos.Update.Poll" --seed 999417177'`
(@neongreen for future reference, here's how to use seeds)

Don't just use `stack test` because it'll be unbearably slow.


rockbmb:
@domenkozar I keep seeing this error in various different branches, can you shed some light on this? My hunch is that it is unrelated to my changes, if you can confirm this I'll ask @gromakovsky to override the CI to close the PR as it's a fix long overdue. If the CI is not working here, then it probably doesn't work in `cardano-0.5-rc` either, and the sooner this fix is merged the better, so it won't make any difference to CI in the original branch.

```
~/build/input-output-hk/cardano-sl
Jun 20 23:49:44 Packing up daedalus-bridge ...
Jun 20 23:49:44 tar (child): s3/daedalus-bridge-linux-tests-csl1217/fixpolltest.tar.xz: Cannot open: No such file or directory
Jun 20 23:49:44 tar (child): Error is not recoverable: exiting now
Jun 20 23:49:44 tar: s3/daedalus-bridge-linux-tests-csl1217/fixpolltest.tar.xz: Cannot write: Broken pipe
Jun 20 23:49:44 tar: Child returned status 2
Jun 20 23:49:44 tar: Error is not recoverable: exiting now
```

domenkozar:
Looking into it.

domenkozar:
There's probably a bug in travis-ci code with branch names that include slashes. Let me try to fix it.

domenkozar:
@rockbmb I pushed a fix that should resolve the problem. Thanks for patience!

rockbmb:
@domenkozar No problem. I'll rename my other branches to prevent this issue from happening again.

PR #974, opened 2017-06-16
build against time-warp-nt master 79acc87f

avieth:
Message limits are now computed within the listener, rather than having
the listener itself be computed by way of a type reflection.


georgeee:
* We try not to use `FIXME` annotations much, general agreement is to create an issue and put a `-- TODO CSL-XXX <title of issue> <desc of fix for line>`. We leave FIXME/TODO w/o issue id only for very minor changes (like `move instance to time-warp`)
* We certainly don't want to have Qs like this in code: https://github.com/input-output-hk/cardano-sl/pull/974/files#diff-86d43f2a2c26aa174be7ebb69db80664R25

georgeee:
I merge this, MacOS CI needs to be fixed in `master`
Minor comments above should be fixed and proposed as new PR

PR #976, opened 2017-06-16
[CSL-1222] Fix slotting data processing in 'RollT'

gromakovsky:
I forgot to add this logic to 'RollT' when I added 'SlottingData' to Poll.


gromakovsky:
It's good, but `unSlottingData` is not used anywhere to read from it, only to write :)

I will fix it within CSL-1106.

PR #978, opened 2017-06-17
[CSL-1125] Don't print trans changeset if empty

volhovM:


src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -87,7 +87,8 @@ calculateTransCorrections eActions = do
 
     -- Once we figure out this piece of code works like a charm we
     -- can delete this logging.
-    logDebug $ sformat ("Trans changeset: "%mapJson) $ HM.toList changeset
+    unless (HM.null changeset) $
+        logDebug $ sformat ("Nonempty dlg trans changeset: "%mapJson) $ HM.toList changeset
```
gromakovsky: FYI: it should work w/o `HM.toList` https://www.stackage.org/haddock/lts-8.13/unordered-containers-0.2.8.0/src/Data.HashMap.Base.html#line-1369

PR #979, opened 2017-06-17
[CSL-1229] Fix US data processing

gromakovsky:


gromakovsky:
I was requested to fix it quickly, but for some reason it's not merged yet and it seems nobody cares about this fix. I wanted to test it, build CSL and then noticed that steps to reproduce contain this one:
> 3. Check logs of node0

And it's too hard for me. So I am merging this PR without testing. In any case, it's a certain bug and even if it doesn't fix CSL-1229, it's still useful.

PR #917, opened 2017-06-09
Feature/csl1190 monaddbread iterators

volhovM:
* `MonadDBRead` now supports iteration: `dbIterSource`
* DB iteration is rewritten using `conduits` everywhere. Previously used `MonadIterator` classes are removed completely.

PR #982, opened 2017-06-17
Feature/csl1190 1209 conduit iterators

volhovM:
This finally introduces conduit-based iterators.
https://github.com/input-output-hk/cardano-sl/pull/917
See this PR for more info ^
`demo.sh` works for 3 and 4 nodes.


db/Pos/DB/Redirect.hs
```diff
@@ -48,3 +64,56 @@ instance
...
+       , MonadResource m
+       , DBIteratorClass i
+       , Bi (IterKey i)
+       , Bi (IterValue i)
+       )
+    => DBTag
+    -> Proxy i
+    -> Source m (IterType i)
+iteratorSource tag _ = do
+    putText $ ("Iterator source, prefix " <> show (iterKeyPrefix @i))
```
gromakovsky: Do we need it? We usually use `log-warper` for logging.

db/Pos/DB/Redirect.hs
```diff
@@ -48,3 +64,56 @@ instance
...
+    DB{..} <- view (dbTagToLens tag) <$> lift getNodeDBs
+    let createIter = Rocks.createIter rocksDB rocksReadOpts
+    let releaseIter i = Rocks.releaseIter i
+    let onExc (e :: SomeException) = do
+            putText $ "Exception arised in redirect handler: " <> show e
```
gromakovsky: I think it would be good to use `log-warper` logging here for consistency.

gromakovsky: Moreover, it's usually considered a bad idea to catch exception, log it and throw further. I don't see a good reason why it's necessary here.

db/Pos/DB/Class.hs
```diff
@@ -99,19 +106,47 @@ dbTagToLens GStateDB     = gStateDB
...
     -- | This function takes tag and key and reads value associated
     -- with given key from DB corresponding to given tag.
     dbGet :: DBTag -> ByteString -> m (Maybe ByteString)
 
+    -- | Source producing iteration over given 'i'.
+    dbIterSource ::
+        ( DBIteratorClass i
+        , Bi (IterKey i)
+        , Bi (IterValue i)
+        ) => DBTag -> Proxy i -> Source (ResourceT m) (IterType i)
```
gromakovsky: It's a little strange that you deleted `Proxy` from `iterKeyPrefix` but are using it here.

lrc/Pos/Lrc/Core.hs
```diff
@@ -26,25 +27,28 @@ import           Pos.Util.Util       (getKeys)
...
-    :: forall m . MonadIterator (StakeholderId, [StakeholderId]) m
-    => (StakeholderId -> m Bool)         -- ^ Check if user is issuer?
-    -> (StakeholderId -> m (Maybe Coin)) -- ^ Gets effective stake.
-    -> Coin
-    -> m (RichmenSet, RichmenStake) -- old richmen, new richmen
+    :: forall m . Monad m
+    => (StakeholderId -> m Bool)                   -- ^ Check if user is issuer?
+    -> (StakeholderId -> m (Maybe Coin))           -- ^ Gets effective stake.
+    -> Coin                                        -- ^ Coin threshold
+    -> Sink (StakeholderId, HashSet StakeholderId) -- ^ Old richmen, new richmen
```
gromakovsky: I think I already pointed out that I don't like this line. I want this haddock comment to be two lines below.
1. I am not sure if it works at all, because this comment is somewhere in the middle of this type. Even if it works, it looks a little weird.
2. This comment applies to `(RichmenSet, RichmenStake)`, but it makes the impression that it's about `(StakeholderId, HashSet StakeholderId)` part.

src/Pos/Block/Logic/Creation.hs
```diff
@@ -217,7 +217,9 @@ createMainBlockFinish slotId pske prevHeader = do
     (block, undo, pModifier) <-
         verifyCreatedBlock uncheckedBlock (pure unchecked) fallbackCreateBlock
     logDebug "Created main block/undos, applying"
-    lift $ block <$ applyBlocksUnsafe (one (Right block, undo)) (Just pModifier)
+    lift $ applyBlocksUnsafe (one (Right block, undo)) (Just pModifier)
+    logDebug "Applied successfully"
```
gromakovsky: I am afraid this message will look confusing for some people. There are many things which can be applied and we don't have good logger names nowadays, so one will have to search it in code. Maybe it's worth adding that we applied **a block ** successfully? Or maybe even attach header hash.

src/Pos/DB/GState/GState.hs
```diff
@@ -50,11 +50,12 @@ prepareGStateDB initialTip = do
...
 sanityCheckGStateDB
     :: forall m.
-       (MonadRealDB m, MonadDBRead m, MonadMask m, WithLogger m)
+       (MonadDBRead m, MonadMask m, WithLogger m)
     => m ()
 sanityCheckGStateDB = do
     sanityCheckBalances
-    sanityCheckUtxo =<< getRealTotalStake
+    totalStake <- getRealTotalStake
+    sanityCheckUtxo totalStake
```
gromakovsky: Was it too hard? :)

src/Pos/Lrc/FtsPure.hs
```diff
@@ -42,14 +43,12 @@ followTheSatoshi seed stakes
...
+                           followTheSatoshiM seed totalCoinsCoin
   where
     totalCoins = sumCoins $ map snd stakes
-    totalCoinsCoin = mkCoin (fromInteger totalCoins)
+    totalCoinsCoin = mkCoin $ fromInteger totalCoins
 
 followTheSatoshiUtxo :: SharedSeed -> Utxo -> NonEmpty StakeholderId
 followTheSatoshiUtxo seed utxo =
-    followTheSatoshi seed (HM.toList (utxoToStakes utxo))
+    followTheSatoshi seed $ HM.toList $ utxoToStakes utxo
```
gromakovsky: @neongreen look, this guy hates brackets.

src/Pos/Lrc/Worker.hs
```diff
@@ -217,14 +218,17 @@ richmenComputationDo
     => EpochIndex -> [LrcConsumer m] -> m ()
 richmenComputationDo epochIdx consumers = unless (null consumers) $ do
     total <- GS.getEffectiveTotalStake
+    logDebug $ "Effective total stake: " <> show total
```
gromakovsky: Why did you use `show`? AFAIU, `total` is `Coin` and it looks better when formatted based on `Buildable`. You could use `pretty`, for example.

stack.yaml
```diff
@@ -49,7 +49,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: 79acc87fcb6fc21998ee2bf1d2ae17d8894bc007
+    commit: 5614dbd3db4313029285fdf144df06e8b3e8c304 # master
```
gromakovsky: of science

db/Pos/DB/Class.hs
volhovM: I couldn't do it without proxy for some reason.

PR #984, opened 2017-06-17
[CSM-307] Take txs from other wallet into account for tx amount evaluation

Martoon-00:
Previously this case wan't considered, and error was thrown

### DIFF
--- a/src/Pos/Wallet/Web/ClientTypes.hs
+++ b/src/Pos/Wallet/Web/ClientTypes.hs
@@ -161,21 +161,22 @@ convertTxOutputs :: [TxOut] -> [(CId w, CCoin)]
 convertTxOutputs = map (addressToCId . txOutAddress &&& mkCCoin . txOutValue)
 
 -- | Get all addresses of source account of given transaction.
-getLocalAccountAddrs
-    :: [CWAddressMeta]  -- ^ all addresses in wallet
-    -> [CId Addr]       -- ^ Input addresses of transaction
-    -> Either Text [CId Addr]
-getLocalAccountAddrs walAddrMetas inputAddrs = do
+getChangeAccountAddrs
+    :: [CWAddressMeta]   -- ^ All addresses in wallet
+    -> [CId Addr]        -- ^ Input addresses of transaction
+    -> Either Text (Maybe [CId Addr])
+                         -- ^ `Just` change addrs if the wallet is source of
+                         --   transaction, `Nothing` otherwise
+getChangeAccountAddrs walAddrMetas inputAddrs = do
     someInputAddr <-
         head inputAddrs `whenNothing`
         throwError "No input addresses in transaction"
-    someSrcAddrMeta <-
-        find ((== someInputAddr) . cwamId) walAddrMetas `whenNothing`
-        throwError "Address doesn't belong to any wallet"
-    let srcAccount =
-            addrMetaToAccount someSrcAddrMeta
-    return $
-        map cwamId $ filter ((srcAccount ==) . addrMetaToAccount) walAddrMetas
+    return $ do
+        someSrcAddrMeta <- find ((== someInputAddr) . cwamId) walAddrMetas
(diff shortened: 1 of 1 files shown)


src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -161,21 +161,22 @@ convertTxOutputs :: [TxOut] -> [(CId w, CCoin)]
...
-getLocalAccountAddrs walAddrMetas inputAddrs = do
+getChangeAccountAddrs
+    :: [CWAddressMeta]   -- ^ All addresses in wallet
+    -> [CId Addr]        -- ^ Input addresses of transaction
+    -> Either Text (Maybe [CId Addr])
+                         -- ^ `Just` change addrs if the wallet is source of
+                         --   transaction, `Nothing` otherwise
+getChangeAccountAddrs walAddrMetas inputAddrs = do
     someInputAddr <-
         head inputAddrs `whenNothing`
```
Pastafarianist: This will cease working as soon as we support multiple input accounts for a single transaction. Can you register this as an issue in the bugtracker?

src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -161,21 +161,22 @@ convertTxOutputs :: [TxOut] -> [(CId w, CCoin)]
...
         throwError "No input addresses in transaction"
-    someSrcAddrMeta <-
-        find ((== someInputAddr) . cwamId) walAddrMetas `whenNothing`
-        throwError "Address doesn't belong to any wallet"
-    let srcAccount =
-            addrMetaToAccount someSrcAddrMeta
-    return $
-        map cwamId $ filter ((srcAccount ==) . addrMetaToAccount) walAddrMetas
+    return $ do
```
Pastafarianist: It might be just me, but it takes some mental effort to figure out that the monad here is `Maybe`. Perhaps you could add a comment to clarify that?

src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -161,21 +161,22 @@ convertTxOutputs :: [TxOut] -> [(CId w, CCoin)]
 convertTxOutputs = map (addressToCId . txOutAddress &&& mkCCoin . txOutValue)
 
 -- | Get all addresses of source account of given transaction.
-getLocalAccountAddrs
-    :: [CWAddressMeta]  -- ^ all addresses in wallet
-    -> [CId Addr]       -- ^ Input addresses of transaction
-    -> Either Text [CId Addr]
-getLocalAccountAddrs walAddrMetas inputAddrs = do
+getChangeAccountAddrs
```
Pastafarianist: Perhaps a name like `getTxChangeAddresses` would be more appropriate? I don't see why it should mention `Account`.

src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -184,12 +185,23 @@ mkCTxs
...
+                -- if given wallet is source of tx, /changes addresses/
+                -- can be fetched according to definition
+                let changeAddrsSet = S.fromList changeAddrs
+                flip S.member changeAddrsSet
+           Nothing -> do
+                -- if given wallet is *not* source of tx, then it's incoming
+                -- transaction, and only addresses of given wallet are *not*
+                -- change addresses
+                let nonChangeAddrsSet = S.fromList $ cwamId <$> wAddrMetas
+                flip S.member nonChangeAddrsSet
```
Pastafarianist: Looks like you have forgotten `not` here.

src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -184,12 +185,23 @@ mkCTxs
...
+        isChangeTxOutput = isChangeAddr . addressToCId . txOutAddress
```
Pastafarianist: This does not seem to be used anywhere.

PR #993, opened 2017-06-19
[CSL-1202] Allow non-empty distribution for PubKeyAddress

neongreen:
I'm not sure whether anything else needs to be done regarding this task, but I grepped for `PubKeyAddress` and these were the only places that looked like they needed to be changed.


neongreen:
@gromakovsky done.

gromakovsky:
@georgeee do you want to review or we can merge?

PR #994, opened 2017-06-19
Switch to time-warp-nt/master

int-index:
prosti @gromakovsky

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -45,7 +45,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: 50d664e676ff41d2176ee9e24d84b52475deef5c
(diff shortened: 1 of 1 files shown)


gromakovsky:
Timeouts are back :(

int-index:
> Timeouts are back :(

Haddock causes them, notice that `travis-ci/push` doesn't time out, but `travis-ci/pr` does — the difference between them is whether `--haddock` is used.

gromakovsky:
Maybe we can separate haddock as we did it for hlint? Does haddock need anything to be compiled?

int-index:
As far as I know, Haddock needs the entire project (and deps) to be built with `--haddock`.

gromakovsky:
I merged it because currently all travis builds fail because of time limit because of haddock. Everything else is correct.

PR #996, opened 2017-06-19
Merge cardano-sl-0.5-rc

gromakovsky:


gromakovsky:
Travis constantly times out because of Haddock (three times in a row), it seems impossible to make it pass within given time, so I am merging.

PR #1000, opened 2017-06-20
[CSM-306] Fix grammar

Pastafarianist:


Martoon-00:
Takaya moya gramota ))

PR #1002, opened 2017-06-20
Cosmetic changes

Pastafarianist:


akegalj:
@Martoon-00 should we discard these changes ?

Martoon-00:
There is no sense in resolving conflicts, it will be much easier to do from scratch (and those haddock comments are gone anyway)

PR #1003, opened 2017-06-20
[CSM-317] Fix changing password

flyingleafe:
New version of `cardano-crypto`, where bug with password changing has been fixed, is used.
Cumbersome workaround with copying the wallet is removed as redundant.

Testing via CLI in dev mode has shown that CSM-314 has been fixed by this too, but more profound testing with Daedalus is planned.

### DIFF
--- a/src/Pos/Wallet/Web/Server/Methods.hs
+++ b/src/Pos/Wallet/Web/Server/Methods.hs
@@ -66,10 +66,10 @@ import           Pos.Core                         (Address (..), Coin, addressF,
                                                    makeRedeemAddress, mkCoin, sumCoins,
                                                    unsafeAddCoin, unsafeIntegerToCoin,
                                                    unsafeSubCoin)
-import           Pos.Crypto                       (EncryptedSecretKey, PassPhrase,
-                                                   aesDecrypt, changeEncPassphrase,
-                                                   checkPassMatches, deriveAesKeyBS,
-                                                   emptyPassphrase, encToPublic, hash,
+import           Pos.Crypto                       (PassPhrase, aesDecrypt,
+                                                   changeEncPassphrase, checkPassMatches,
+                                                   deriveAesKeyBS, emptyPassphrase,
+                                                   encToPublic, hash,
                                                    redeemDeterministicKeyGen,
                                                    redeemToPublic, withSafeSigner,
                                                    withSafeSigner)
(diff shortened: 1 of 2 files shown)


src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -853,86 +835,19 @@ instance Buildable AccountsSnapshot where
...
-
 changeWalletPassphrase
     :: WalletWebMode m
-    => SendActions m -> CId Wal -> PassPhrase -> PassPhrase -> m ()
-changeWalletPassphrase sa wid oldPass newPass = do
+    => CId Wal -> PassPhrase -> PassPhrase -> m ()
+changeWalletPassphrase wid oldPass newPass = do
     oldSK <- getSKByAddr wid
 
     -- 'cloneWalletSetWithPass' will work badly if accounts / addresses ids
```
Martoon-00: Please remove this comment as well

flyingleafe:
Tested with Daedalus in dev mode and prod mode on qanet. All bugs listed in CSM-314 seem to be gone in both cases.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -853,86 +835,17 @@ instance Buildable AccountsSnapshot where
...
-                `E.onException` do
-                    logError "Money transmition to new wallet failed"
-                    mapM_ totallyRemoveWAddress $ asDeleted <> asExisting $ newAddrMeta
-            return oldAddrMeta
-        mapM_ removeWAddress $ asExisting oldAddrMeta
         setWalletPassLU wid =<< liftIO getPOSIXTime
-        deleteSK oldPass
   where
     badPass = RequestError "Invalid old passphrase given"
     deleteSK passphrase = do
```
georgeee: > isJust (checkPassMatches passphrase k)
This check is redundant

This is checked above at:
> newSK <- maybeThrow badPass $ changeEncPassphrase oldPass newPass oldSK

Martoon-00: They are not the same.
First one you mentioned checks whether new pass fits to old secret key, while second one - that _old_ pass fits to that secret key

Martoon-00: However first check can be removed painlessly

flyingleafe: https://issues.serokell.io/issue/CSM-319
^ Created a separate issue to get rid of such comparisons

PR #1005, opened 2017-06-21
Disable Haddock, again :(

int-index:
Fixes build time outs.

### DIFF
--- a/scripts/ci/travis.sh
+++ b/scripts/ci/travis.sh
@@ -11,7 +11,8 @@ fi
 
 if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
     if [[ "$with_haddock" == "true" ]]; then
-      export EXTRA_STACK="--haddock";
+      # export EXTRA_STACK="--haddock";
+      :
     fi
 
     export EXTRA_STACK="--test $EXTRA_STACK";


int-index:
Some back story. I enabled Haddock in #967 because build-time performance was fixed. Indeed, you can witness how https://travis-ci.org/input-output-hk/cardano-sl/jobs/244405570 succeeds. However, right after merging into `master` we started to get time outs again.

cc @domenkozar

int-index:
Another build where Haddock didn't cause trouble:

https://travis-ci.org/input-output-hk/cardano-sl/jobs/245233457


Seems to be non-deterministic.

gromakovsky:
Hm, it's strange that this build passed in 1 hr 5 min 24 sec, while other builds timeout on building `cardano-sl-update` (which is not even close to the end) at 1:19:++.

Also I see this interesting thing in this build: http://i.imgur.com/rN0d8NO.png

https://travis-ci.org/input-output-hk/cardano-sl/jobs/245233457#L1935

domenkozar:
We have too much dependencies and compilation steps, our build takes 80min+. We will have to strip things off, we can disable documentation but I don't think that's the right place to start.

A place to start would be https://issues.serokell.io/issue/CSM-195

gromakovsky:
Currently it seems that more than 80% builds time out. So PRs get merged with very high delays or via red button. I think it's not acceptable and we should disable haddock until another compile time improvement is done (e. g. CSM-195).

PR #1006, opened 2017-06-21
[CSL-1240] + [CSL-1242] Introduce TxFeePolicy

int-index:


core/Pos/Core/Constants/Typed.hs
```diff
@@ -61,6 +61,7 @@ genesisBlockVersionData =
     , bvdUpdateProposalThd = genesisUpdateProposalThd
     , bvdUpdateImplicit = genesisUpdateImplicit
     , bvdUpdateSoftforkThd = genesisUpdateSoftforkThd
+    , bvdTxFeePolicy = Nothing
```
gromakovsky: I think it should be configurable by `constants.yaml`. Probably it's for CSL-1242, so ok.

core/Pos/Core/Fee.hs
```diff
@@ -0,0 +1,44 @@
...
+newtype Coeff = Coeff Nano
+    deriving (Eq, Show, Generic, NFData)
+
+instance Buildable Coeff where
+    build (Coeff x) = fromString (showFixed True x)
+
+-- | A linear equation on the transaction size. Represents the @\s -> a + b*s@
+-- function where @s@ is the transaction size, @a@ and @b@ are constant
+-- coefficients.
+data TxSizeLinear = TxSizeLinear !Coeff !Coeff
```
gromakovsky: Actually now I think that one value should be `CoinPortion` and another one should be `Coin` (when I wrote `Nano` in slack I didn't think at all).

src/Pos/Types/Arbitrary/Core.hs
```diff
@@ -37,3 +39,21 @@ instance Arbitrary Types.EpochOrSlot where
...
+deriving instance Arbitrary Fee.Coeff
+
+instance Arbitrary Fee.TxSizeLinear where
+    arbitrary = genericArbitrary
+    shrink = genericShrink
+
+instance Arbitrary Fee.TxFeePolicy where
+    arbitrary = do
+        v <- arbitrary
+        case v of
```
gromakovsky: I don't like it. It will almost always be `Unknown`. I think probabilities of `TxSizeLinear` and `Unknown` should be similar.

core/Pos/Core/Fee.hs
int-index: `Nano` makes more sense to me because these values must be fixed-precision fractionals and the computations will be performed accordingly.

gromakovsky: `CoinPortion` is fixed-precision fractional too. There is `applyCoinPortion` for multiplication.

gromakovsky: And adding `Nano` to `Coin` doesn't make sense at all, because `Coin` is integer.

int-index: The entire check will be `a + b * txSize <= txFee`. Nothing prevents `a` and `b` from being fractional, although `txSize` and `txFee` are integer.

int-index: 1. Initially the design was with a quadratic term, `a + b * txSize + c * txSize * txSize`, so I'd rather treat those coefficients numerically, without units (otherwise `c` would have to be `coin/bytes^2`.
2. Why do porpose `applyCoinPortion` if multiplication is with bytes, not coins?

gromakovsky: > Why do you propose applyCoinPortion if multiplication is with bytes, not coins?

Oh lol, indeed.

Okay then, `Nano` should be fine, units are mixed anyway. Maybe `Nano` is too precise, but it's hard to say.

core/constants.yaml
```diff
@@ -25,6 +25,11 @@ dev: &dev
   genesisUpdateProposalThd: 0.1 # 10% of total stake
   genesisUpdateImplicit: 10 # slots
   genesisUpdateSoftforkThd: 0.75 # 75% of total stake
+  genesisTxFeePolicy:
+    txSizeLinear:
+      a: 1.2
```
int-index: Random values to test the parser. Can anyone suggest more realistic ones?

txp/Pos/Txp/Toil/Logic.hs
```diff
@@ -130,8 +168,10 @@ verifyAndApplyTx
     :: (MonadUtxo m, MonadToilEnv m, MonadError ToilVerFailure m)
     => Bool -> (TxId, TxAux) -> m TxUndo
 verifyAndApplyTx verifyVersions tx@(_, txAux) = do
-    verifyToilEnv txAux
-    Utxo.verifyTxUtxo ctx txAux <* applyTxToUtxo' tx
+    applyTxToUtxo' tx
```
int-index: Accidentally reordered actions here, so that `applyTxToUtxo'` is called before verification. This is fixed in a later commit.

core/Pos/Core/Fee.hs
```diff
@@ -0,0 +1,90 @@
...
+import qualified Data.HashMap.Strict        as HM.S
+
+-- | A fractional coefficient of fixed precision.
+newtype Coeff = Coeff Nano
+    deriving (Eq, Show, Generic, NFData)
+
+instance Buildable Coeff where
+    build (Coeff x) = fromString (showFixed True x)
+
+instance JSON.FromJSON Coeff where
```
gromakovsky: We keep config parsing code stupid and don't write aeson instances for our types, but rather parse them as plain numbers (and later convert them).

Probably that's not for good reason, but because everybody was lazy and you just happened to be the first person who is not lazy to do it. So ok.

core/Pos/Core/Fee.hs
```diff
@@ -0,0 +1,90 @@
...
+            <$> o JSON..: "a"
+            <*> o JSON..: "b"
+
+calculateTxSizeLinear :: TxSizeLinear -> Byte -> Nano
+calculateTxSizeLinear
+    (TxSizeLinear (Coeff a) (Coeff b))
+    (fromInteger . toBytes -> txSize) =
+        a + b * txSize
+
+data TxFeePolicy
```
gromakovsky: Can you add a Haddock comment to this type? It's quite important type.

core/Pos/Core/Types.hs
```diff
@@ -283,6 +285,7 @@ instance Bounded Coin where
 maxCoinVal :: Word64
 maxCoinVal = 45000000000000000
 
+-- FIXME: This operation is unsafe because it doesn't check 'maxCoinVal'.
```
gromakovsky: Funny thing is that `getCoin` has `unsafe` prefix despite being safe, while `mkCoin` is unsafe and doesn't have it :)

stack.yaml
```diff
@@ -96,6 +96,7 @@ extra-deps:
 - lens-4.15.3                     # not yet in lts-8
 - th-abstraction-0.2.2.0          # needed for lens
 - rocksdb-haskell-1.0.0
+- generic-arbitrary-0.1.0
```
gromakovsky: More dependencies! Maybe we can get rid of `derive` package then? I think we use them in more or less same way.

txp/Pos/Txp/Toil/Failure.hs
```diff
@@ -30,6 +30,8 @@ data ToilVerFailure
     | ToilInvalidInputs ![Text] -- [CSL-814] TODO: make it more informative
     | ToilTooLargeTx { ttltSize  :: !Byte
                      , ttltLimit :: !Byte}
+    | ToilInsufficientFee { tifPolicy :: !TxFeePolicy
+                          , tifSize   :: !Byte }
```
gromakovsky: I think it's more important to include fee into this error than size. Though size is useful too.

core/Pos/Core/Fee.hs
gromakovsky: Also I just realized there **is** is a good reason why we didn't define aeson instances for core types. That's because some of them should be transferred to frontend and they may need different serialization formats. So middleware guys are kings of JSON serialization currently.

txp/Pos/Txp/Toil/Logic.hs
```diff
@@ -115,13 +119,47 @@ normalizeToil txs = mapM_ normalize ordered
...
-    let txSize = biSize txAux
+    => TxAux -> TxFee -> m ()
+verifyToilEnv txAux txFee = do
+    toilEnv <- getToilEnv
+    let
+        limit = teMaxTxSize toilEnv
+        mtxFeePolicy = teTxFeePolicy toilEnv
+        txSize = biSize txAux
+    case mtxFeePolicy of
+        Nothing ->
```
gromakovsky: Actually it should never happen (because policy is always `Just` in genesis `BlockVersionData`), but it doesn't seem easy to enforce it on type-level without something cumbersome.

It would be good to log error in such case, but not necessary (probably it will involve propagating `WithLogger` constraint into too many places).

core/Pos/Core/Fee.hs
int-index: I don't know how to represent a fee policy using a single number (except for some kind of Gödel numbering). Those instances are specifically for the config. I can create a newtype `ConfigTxFeePolicy` so that the core type would be without an instance.

stack.yaml
int-index: I don't like `derive` because it's based on `haskell-src-exts` and can't handle `PolyKinds`. But it doesn't matter here, I just went for a solution I like more.

gromakovsky: I am just thinking loudly. Ok, I think I will create a minor issue.

core/Pos/Core/Fee.hs
gromakovsky: > I don't know how to represent a fee policy using a single number

In stupid code it could be represented as a list of two numbers in JSON :)

> I can create a newtype ConfigTxFeePolicy so that the core type would be without an instance.

It would be good, but let's not do it until it happens to be necessary (maybe never).

gromakovsky:
So, Haddock comment and fees to `ToilInsufficientFee`.

int-index:
Okay, I think I addressed the outlined problems.

core/Pos/Core/Fee/Config.hs
```diff
@@ -0,0 +1,56 @@
...
+       ) where
+
+import           Universum
+
+import qualified Data.Aeson          as JSON
+import qualified Data.Aeson.Types    as JSON
+import           Data.Fixed          (Fixed (..), resolution)
+import qualified Data.HashMap.Strict as HM.S
+
+import           Pos.Core.Fee
```
gromakovsky: Implicit imports are prohibited, sorryan.

txp/Pos/Txp/Toil/Failure.hs
```diff
@@ -55,5 +62,17 @@ instance Buildable ToilVerFailure where
     build (ToilTooLargeTx {..}) =
         bprint ("transaction's size exceeds limit "%
                 "("%memory%" > "%memory%")") ttltSize ttltLimit
+    build (ToilInvalidMinFee {..}) =
+        bprint ("policy"%build%" generates invalid minimal fee on a "%
```
gromakovsky: 1. You forgot space.
2. Word `policy` seems to be redundant, because format of `TxFeePolicy` already includes it.

int-index:
> Also it would be good to test it, did you do it?

I don't know a proper way to test it, but I've run the `demo` script and it didn't output any errors.

gromakovsky:
`demo` script is not really useful to test transaction processing, because it doesn't send any transactions. It only proves that you didn't make something absolutely weird.

Proper way to test it is to use `scripts/launch/wallet.sh` and send some transactions, but:
1. Transactions will be invalid, because code which creates them doesn't take fee policy into account.
2. `wallet.sh` is most likely broken.

Regarding the first point: you should set fee policy to always return `0` as minimal fee, until transaction creation code starts considering them. Currently transaction sending is likely broken.

int-index:
> This issue can be merged after fee policy is changed to zero.

Do you mean in `constants.yaml`?

gromakovsky:
Yes

PR #1007, opened 2017-06-21
Feature/csl1224 report logging

volhovM:
This introduces the small reporting fix: if no servers are specified we won't try to retrieve `.pub` logs so we won't fail if they are not present (the setting used in QA cluster).
I've also removed util-scripts dir as it's redundant after scripts refactoring and renamed a `reportMisbehaviorMasked` into `reportMisbehaviorSilent` as recommended in the TODO.

### DIFF
--- a/infra/Pos/Reporting/Methods.hs
+++ b/infra/Pos/Reporting/Methods.hs
@@ -7,7 +7,7 @@ module Pos.Reporting.Methods
        , sendReportNodeNologs
        , getNodeInfo
        , reportMisbehaviour
-       , reportMisbehaviourMasked
+       , reportMisbehaviourSilent
        , reportingFatal
        , sendReport
        , retrieveLogFiles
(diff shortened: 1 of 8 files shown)


infra/Pos/Reporting/Methods.hs
```diff
@@ -59,6 +60,13 @@ import           Pos.Reporting.MemState   (MonadReportingMem, rcLoggingConfig,
     lhs' = reverse $ dropWhile isSlash $ reverse lhs
     rhs' = dropWhile isSlash rhs
 
+type ReportingMode m =
```
gromakovsky: FYI: we decided not to use `Mode` for constraints. Mode is execution mode. `MonadReporting` would be more consistent with the rest of the code.

volhovM:
One CI failed with weird error not related to the PR. I'm merging.

PR #1008, opened 2017-06-21
[CSL-1239] Show full leaders list only once

volhovM:
* Full leader list is shown once per epoch with INFO
* Every other slot s we show [s-10..s+10] window of leaders with DEBUG verbosity

### DIFF
--- a/src/Pos/Block/Worker.hs
+++ b/src/Pos/Block/Worker.hs
@@ -98,16 +98,26 @@ blkOnNewSlotImpl (slotId@SlotId {..}) sendActions = do
   where
     onNoLeader =
         logWarning "Couldn't find a leader for current slot among known ones"
-    logLeadersF = if siSlot == minBound then logInfo else logDebug
-    logLeadersFS = if siSlot == minBound then logInfoS else logDebugS
+    logOnEpochFS = if siSlot == minBound then logInfoS else logDebugS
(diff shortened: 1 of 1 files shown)


src/Pos/Block/Worker.hs
```diff
@@ -98,16 +98,26 @@ blkOnNewSlotImpl (slotId@SlotId {..}) sendActions = do
...
         let ourPkHash = addressHash ourPk
         logNoticeS "This is a test debug message which shouldn't be sent to the logging server."
-        logLeadersFS $ sformat ("Our pk: "%build%", our pkHash: "%build) ourPk ourPkHash
-        logLeadersF $ sformat ("Slot leaders: "%listJson) $
-                      map (bprint pairF) (zip [0 :: Int ..] $ toList leaders)
-        logLeadersF $ sformat ("Current slot leader: "%build) leader
+        logOnEpochFS $ sformat ("Our pk: "%build%", our pkHash: "%build) ourPk ourPkHash
+        logOnEpochF $ sformat ("Current slot leader: "%build) leader
+
+
```
gromakovsky: Why two empty lines?

gromakovsky:
Also, if @georgeee really wants to see leaders only once per epoch (even if we show only subset of them), it makes sense to make sure that some special logger name is used to print them. Then it will really be a matter of configuring logging properly.

PR #1009, opened 2017-06-21
[CSM-148] All transactions have the same date and time

ajnsit:
Initialise the transaction history metadata for a particular wallet, if it didn't exist previously.

### DIFF
--- a/core/Pos/Core/Timestamp.hs
+++ b/core/Pos/Core/Timestamp.hs
@@ -3,14 +3,16 @@
 module Pos.Core.Timestamp
        ( Timestamp (..)
        , timestampF
+       , getCurrentTimestamp
        ) where
 
 import           Universum
 
-import           Data.Text.Buildable (Buildable)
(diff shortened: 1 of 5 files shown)


ajnsit:
@Martoon-00 Yeah I had to use `non'` instead of `non` because it complained about a missing `Eq` instance otherwise.

Martoon-00:
Could you explain why it fixes CSM-148?

ajnsit:
@Martoon-00 It fixes atleast a part of the problem. When we send some money, we store the time of the transaction in the metadata in the wallet storage and not on the blockchain. However, it would fail if the wallet id was not pre-initialised in the store. This edit initialises the wallet storage to an empty hashmap if it didn't already exist.

Martoon-00:
Thanks, makes sense

Martoon-00:
By the way, it's not the only place there `wsTxHistory . ix cWalId` is encountered )
How do you think about fixing `addWalletTxHistory` as well?

PR #1010, opened 2017-06-21
Tests-alexrb-csl1160-structure

rockbmb:


core/cardano-sl-core.cabal
```diff
@@ -43,13 +43,17 @@ library
                        Pos.Core.Slotting
                        Pos.Core.Timestamp
                        Pos.Core.Types
+                       Pos.Core.Types.Arbitrary
+                       Pos.Core.Types.Arbitrary.Unsafe
```
gromakovsky: I think `Types` is redundant here and one line above.

src/Pos/Ssc/Arbitrary.hs
```diff
@@ -2,12 +2,13 @@ module Pos.Ssc.Arbitrary
...
 import           Universum
 
-import           Pos.Ssc.Class.Types (Ssc (..))
-import           Pos.Ssc.NistBeacon  (SscNistBeacon)
-import           Pos.Types           (SlotId)
+import           Test.QuickCheck          (Arbitrary (..), Gen)
+
+import           Pos.Ssc.Class.Types      (Ssc (..))
+import           Pos.Ssc.NistBeacon       (SscNistBeacon)
+import           Pos.Types                (SlotId)
```
gromakovsky: I think this module `Pos.Ssc.Arbitrary` should go to `cardano-sl-ssc`. It imports two modules from `cardano-sl`: `Pos.Ssc.NistBeacon` and `Pos.Types`. The latter can be changed to `Pos.Core`. The former can be moved to `cardano-sl-ssc`, it only imports `Pos.Types` which can be changed to `Pos.Core` as well.

gromakovsky: Ah, `NistBeacon` also imports `Pos.Binary.Relay`, but it's completely useless.

src/Pos/Ssc/GodTossing/Arbitrary.hs
```diff
@@ -43,7 +43,7 @@ import           Pos.Ssc.GodTossing.Types.Message (GtTag (..), MCCommitment (..)
 import           Pos.Ssc.GodTossing.Types.Types   (GtGlobalState (..),
                                                    GtSecretStorage (..))
 import           Pos.Ssc.GodTossing.VssCertData   (VssCertData (..))
-import           Pos.Types.Arbitrary.Core.Unsafe  ()
+import           Pos.Core.Types.Arbitrary.Unsafe  ()
 import           Pos.Util.Arbitrary               (Nonrepeating (..), makeSmall, sublistN,
                                                    unsafeMakePool)
```
gromakovsky: Is there anything preventing this module from being moved to `cardano-sl-godtossing`?

gromakovsky: Now I realized that I am reviewing only subset of changes, maybe it's already moved. I will see it soon :)

Ignore if it's really moved.

src/Pos/Ssc/Arbitrary.hs
gromakovsky: Alright, I see that you already did it in the next commit, nevermind.

core/cardano-sl-core.cabal
rockbmb: Alright, I'll make the quick fix.

rockbmb:
@gromakovsky By the way, a heads-up: the only remaining `Arbitrary` module in `src` is in `src/Pos/Block/Arbitrary.hs`, but this one cannot be moved into `core` because it depends on block modules that themselves cannot be moved outside of `src`.

gromakovsky:
> By the way, a heads-up: the only remaining Arbitrary module in src is in src/Pos/Block/Arbitrary.hs, but this one cannot be moved into core because it depends on block modules that themselves cannot be moved outside of src.

Yes, it's fine to `Pos.Block.Arbitrary` where it is now, except that `TxPayload` and related should be moved to `txp` as I wrote above.

gromakovsky:
`Pos.DHT.Arbitrary` is probably a leftover.

rockbmb:
@gromakovsky Done. As soon as CI passes you can merge this.

PR #1011, opened 2017-06-21
[CSL-1234] Merge cardano-sl-0.5-rc into master

gromakovsky:


rockbmb:
@gromakovsky the `travis-ci/pr` build will fail here for the same reason as it does in my CSL-1160 branch, although in different points.

@domenkozar near line 1708 of this build (https://travis-ci.org/input-output-hk/cardano-sl/jobs/245509716) there appear to be these logs:

```
Timeout (20 minutes) reached. Terminating "sleep 4600"

The command sleep 4600 exited with 137.
```

The same also happens in the `tests-alexrb-csl1160-structure` branch. Is this `sleep` necessary here? I don't think it's possible to alter compilation to avoid timeouts in any other way than to remove this instruction. The other option is to just override CI here.

gromakovsky:
Now there is something weird: http://i.imgur.com/9EQLW9R.png

domenkozar:
Try restarting the travis job

PR #1013, opened 2017-06-22
Port ekg statsd to cardano-sl-0.5-rc

arybczak:
Port of #368.

Where do you want the `--with-rtsopts=-T`? For all the executables in cardano-sl.cabal?


domenkozar:
It's ok to leave out `-T` for now, we add it as part of https://github.com/input-output-hk/iohk-nixops/pull/71

PR #1018, opened 2017-06-23
[CSL-1268] Remove Ether.MonadReader & ExecMode

int-index:
* Remove `ExecMode`
* Remove `modeContext`
* Remove some redirection layers
* Remove `runReaderT` on top of base monad
* Use `(Mtl.MonadReader ctx m, HasFoo ctx, HasBar ctx)` instead of `(Ether.MonadReader' Foo m, Ether.MonadReader' Bar m)`.
* Replace some of `HasLens Foo` constraints with dedicated `HasFoo` classes. TBD: finish this.


infra/Pos/Slotting/Impl/Sum.hs
```diff
@@ -40,46 +40,34 @@ data SlottingContextSum
...
-currentTimeSlottingSum = helper currentTimeSlotting
-
-instance (SlotsSumEnv m, t ~ IdentityT) =>
-         MonadSlots (Ether.TaggedTrans SlotsRedirectTag t m) where
-    getCurrentSlot = getCurrentSlotSum
-    getCurrentSlotBlocking = getCurrentSlotBlockingSum
-    getCurrentSlotInaccurate = getCurrentSlotInaccurateSum
-    currentTimeSlotting = currentTimeSlottingSum
-
-helper
```
gromakovsky: Why did you remove `helper`?

int-index: It won't survive further changes, I think.

src/Pos/Context/Context.hs
```diff
@@ -36,169 +35,225 @@ import           Universum
...
+    sscContext = ncSscContext_L
+
+instance HasDiscoveryContextSum (NodeContext ssc) where
+    discoveryContextSum = ncDiscoveryContext_L
+
+instance HasSlottingVar (NodeContext ssc) where
+    slottingTimestamp = ncSlottingVar_L . _1
+    slottingVar = ncSlottingVar_L . _2
+
+instance HasLens SlottingContextSum (NodeContext ssc) SlottingContextSum where
```
gromakovsky: After reviewing majority of this PR I suddenly decided to ask: why do you use `HasPatak` in some places and `HasLens Bardaq` in other places? Is it WIP or planned for another task or http://cs5.pikabu.ru/post_img/big/2015/12/04/5/1449210847155432089.jpg?

src/Pos/Launcher/Param.hs
```diff
@@ -64,3 +76,26 @@ data NodeParams = NodeParams
...
+makeLensesWith postfixLFields ''NodeParams
+
+instance HasLens UpdateParams NodeParams UpdateParams where
+    lensOf = npUpdateParams_L
+
+instance HasLens SecurityParams NodeParams SecurityParams where
+    lensOf = npSecurityParams_L
+
+instance HasLens GenesisUtxo NodeParams GenesisUtxo where
+    lensOf = npCustomUtxo_L . coerced
```
gromakovsky: I think it would be better to make the type of `npCustomUtxo` `GenesisUtxo`.

src/Pos/Wallet/WalletMode.hs
```diff
@@ -56,9 +56,9 @@ instance {-# OVERLAPPABLE #-}
 -- Composite restrictions
 ---------------------------------------------------------------
 
-type WalletMode m
-    = ( TxMode m
-      , MonadKeys m
+type MonadWallet ssc ctx m
+    = ( TxMode ssc m
```
gromakovsky: I think `ssc` should be replaced with `WalletSscType` here.

src/Pos/Wallet/Web/Mode.hs
```diff
@@ -137,17 +138,17 @@ instance MonadDB WalletWebMode where
 instance MonadBlockDBWrite WalletSscType WalletWebMode where
     dbPutBlund = dbPutBlundDefault
 
-instance MonadBlockDBGeneric (BlockHeader WalletSscType) (Block WalletSscType) Undo WalletWebMode
+instance SscHelpersClass ssc => MonadBlockDBGeneric (BlockHeader ssc) (Block ssc) Undo WalletWebMode
```
pva701: `WalletSscType` -> `ssc`. yet another war

lwallet/Pos/Wallet/Light/Redirect.hs
```diff
@@ -27,7 +26,7 @@ import qualified Pos.Wallet.Light.State    as LWS
 -- MonadBalances
 ----------------------------------------------------------------------------
 
-getOwnUtxosWallet :: (MonadIO m, Ether.MonadReader' LWS.WalletState m) => [Address] -> m Utxo
+getOwnUtxosWallet :: (MonadIO m, MonadReader ctx m, HasLens LWS.WalletState ctx LWS.WalletState) => [Address] -> m Utxo
```
int-index: This should be replaced with `LWS.HasWalletState ctx`, I didn't remove all `HasLens` uses yet.

src/Pos/Context/Context.hs
int-index: Planned for another task (it's routine and beginner friendly).

infra/Pos/Communication/Relay/Types.hs
```diff
@@ -51,3 +56,17 @@ data RelayContext = RelayContext
     { _rlyIsPropagation    :: !Bool
     , _rlyPropagationQueue :: !RelayPropagationQueue
     }
+
+makeLenses ''RelayContext
```
int-index: `makeClassy` creates classes of this shape:

```
class HasPropagationQueue s a | s -> a
```

and I want them to be of this shape:

```
class HasPropagationQueue ctx where
```

One parameter less.

db/Pos/DB/Class.hs
```diff
@@ -80,7 +80,7 @@ import           Control.Monad.Trans.Lift.Local (LiftLocal (..))
 import           Control.Monad.Trans.Resource   (ResourceT)
 import           Data.Conduit                   (Source)
 import qualified Database.RocksDB               as Rocks
-import qualified Ether
+import           Ether.Internal                 (HasLens (..))
```
int-index: It turned out to be difficult to remove in a single task, this PR is concerned with `MonadReader` only.

int-index: Initially I planned on removing `ether` completely, but it's just too much work to do in one take.

infra/Pos/Communication/Relay/Types.hs
gromakovsky: But then why do I see only one parameter here? https://www.stackage.org/haddock/lts-8.13/lens-4.15.1/Control-Lens-TH.html#v:makeClassy

georgeee: @int-index ?

int-index: Because I was looking at docs for `makeFields`, not `makeClassy`.

nh2:
I tried to load this branch into ghci but it didn't work, complaining that this module doesn't exist:

https://github.com/input-output-hk/cardano-sl/commit/ada4cfbd06de29222a94aacfde7feff73a76ea9b#diff-d1ef8c10ae5c27f2e2e30ea9a4e092ee

And it doesn't seem to exist in my tree either. I asked on Slack about that.

infra/Pos/Communication/Relay/Types.hs
int-index: I went for whichever notation worked best with existing field names. Doesn't really matter because those lenses are not exported (and not meant to be exported).

lwallet/Pos/Wallet/Light/Mode.hs
```diff
@@ -44,37 +49,48 @@ import           Pos.Wallet.Light.State.Core      (gsAdoptedBVDataWallet)
...
-        !(PeerStateTag     ::: PeerStateCtx Production)
-        !(KeyData          ::: KeyData)
-        !(WalletState      ::: WalletState)
-        !(ReportingContext ::: ReportingContext)
-        !(DiscoveryTag     ::: Set NodeId)
-        !(JsonLogConfig    ::: JsonLogConfig)
-        !(LoggerName       ::: LoggerName)
-    |]
+type LightWalletSscType = SscGodTossing
+-- type LightWalletSscType = SscNistBeacon
```
int-index: No, it's just a showcase how it could be done alternatively. This is how it's done for `WalletSscType`:

```
-- | We can also use CPP flags that I proposed that would enable switching between
-- implementations without changing the code.
type WalletSscType = SscGodTossing
--type WalletSscType = SscNistBeacon
```

nh2:
> And it doesn't seem to exist in my tree either. I asked on Slack about that.

Figured out together with @rockbmb, the reason is an unused file that's checked in, but not in the .cabal file, and it is planned to be fixed:

> I delete this file again in my #1045 PR, so when it's merged to `master` this should go away.

PR #1022, opened 2017-06-23
[CSL-1179] Basic blockchain test

gromakovsky:
Now you can write such properties:
```
verifyEmptyMainBlock :: BlockProperty ()
verifyEmptyMainBlock = do
    let mainBlock :: MainBlock SscGodTossing
        mainBlock = …
        block = Right mainBlock
    lift (verifyBlocksPrefix (one block)) >>= assert . isRight
```

`BlockProperty` is `PropertyM`, so you can use things like `pick` and `assert`. It's also instance of many classes (not directly, but one layer below, so use `lift`), so you can use functions like `verifyBlocksPrefix`.

You can check that it works by running `b --for-installer sl -t '--test-arguments --match"Block.Logic.Var"'`.

Outcomes of this issue:
• added `BlockProperty` with `Testable` instance which allows one to write monadic properties which use functions from the main code and also can use `PropertyM` features;
• added property which states that empty main block for the 0-th slot of the 0-th epoch is always successfully verified if it's created by the slot leader. You can treat this property as a proof of concept. It doesn't test much logic, but it demonstrates how to write future properties and it already prevents some breakages.


gromakovsky:
Hello, dear reviewers! I removed `WIP` label and I think that issue is done. Please take a look. I also updated the description so that you can see overview of what has been done.

gromakovsky:
Tests fail in CI, but don't fail for me locally if I use same seed 😒

gromakovsky:
Alright, it fails locally if I run all tests, but doesn't fail if I run that single test. Weird.

gromakovsky:
Okay, I was a little stupid. Of course it doesn't matter whether seeds are same or not if I run different set of tests.

gromakovsky:
Okay, I fixed most of the problems which could cause tests to fail, but currently they still can fail. For example, `b --for-installer sl -t '--test-arguments --match"Block.Logic.Var"-a"5000"--seed"1701344398"'` leads to the following error:
> internal error: Richmen stakes less than threshsold (after 2891 tests):

I will investigate tomorrow probably.

gromakovsky:
I got rid of some problems and realized why CI problems happen. That's because CI uses production mode. I will find a solution for it.

core/Pos/Core/Genesis.hs
```diff
@@ -93,7 +93,6 @@ genesisStakeDistribution
     | isDevelopment = def
     | otherwise     = gcdDistribution compileGenCoreData
 
-genesisBalances :: HashMap StakeholderId Coin
-genesisBalances
-    | isDevelopment = mempty
-    | otherwise     = gcdBootstrapBalances compileGenCoreData
+-- | Genesis stakes which should be used only in production mode.
+genesisStakesProdMod :: StakesMap
```
volhovM: Motivation of this change?

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -0,0 +1,296 @@
...
+                    (stakeDistribution tpStakeDistribution)
+        return TestParams {..}
+
+----------------------------------------------------------------------------
+-- Main context
+----------------------------------------------------------------------------
+
+modeContext [d|
+    data BlockTestContext = BlockTestContext
+        { btcDBs               :: !(NodeDBs     ::: NodeDBs)
```
volhovM: Nice prefix.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -0,0 +1,84 @@
...
+-- • it's 'fail' from 'Monad', not 'MonadFail';
+-- • I am not a fan of 'fail'.
+-- TODO: maybe these functions will become popular, then it makes
+-- sense to move them somewhere.
+stopProperty :: Monad m => Text -> PropertyM m a
+stopProperty msg = stop failed {reason = toString msg}
+
+maybeStopProperty :: Monad m => Text -> Maybe a -> PropertyM m a
+maybeStopProperty msg =
+    \case
```
volhovM: Indentation.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -0,0 +1,84 @@
...
+        "created by the leader of the 0-th slot " <>
+        "always succeeds for initial GState"
+
+-- Note, 'fail' does the same thing, but:
+-- • it's quite trivial, almost no copy-paste;
+-- • it's 'fail' from 'Monad', not 'MonadFail';
+-- • I am not a fan of 'fail'.
+-- TODO: maybe these functions will become popular, then it makes
+-- sense to move them somewhere.
+stopProperty :: Monad m => Text -> PropertyM m a
```
volhovM: I was using this instead:
```
leftToCounter :: (ToString s, Testable p) => Either s a -> (a -> p) -> Property
leftToCounter x c = either (\t -> counterexample (toString t) False) (property . c) x
```
Which looks terrible. Will rewrite someday.

test/Test/Pos/Block/Logic/VarSpec.hs
gromakovsky: What's wrong?

core/Pos/Core/Genesis.hs
gromakovsky: 1. Previously one could accidentally pass `genesisBalances` somewhere in development mode, but empty genesis stakes don't make sense at all (they make system broken). Now name makes it clear that they shouldn't be used in development mode.
2. Name `Balances` is wrong.
3. Using type aliases is a good practice.
4. I want to get rid of `isDevelopment` in as many places as possible. It's not my current goal, but when I see such opportunity, I do it.

test/Test/Pos/Block/Logic/VarSpec.hs
volhovM: `\case` can be on the line 55 obv.

gromakovsky: But we all know that most likely eventually someone will press `M-q` and move it to the line 56. Then someone will move it back. And so on.

I prefer to format code like `hindent` does as long as it doesn't contradict to our style-guide.

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
```diff
@@ -382,15 +382,19 @@ checkPayload
...
-        OpeningsPayload     opens  _ ->
-            checkOpeningsPayload opens
-        SharesPayload       shares _ ->
-            checkSharesPayload epoch shares
-        CertificatesPayload        _ ->
-            pass
+        CommitmentsPayload comms _
+            | null comms -> pass
+            | otherwise -> checkCommitmentsPayload epoch comms
+        OpeningsPayload opens _ -> checkOpeningsPayload opens
```
pva701: why didn't you add similar checks for openings and shares?

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -0,0 +1,296 @@
...
+                    (toList tpSecretKeys)
+                    (stakeDistribution tpStakeDistribution)
+        return TestParams {..}
+
+----------------------------------------------------------------------------
+-- Main context
+----------------------------------------------------------------------------
+
+modeContext [d|
+    data BlockTestContext = BlockTestContext
```
pva701: I need something like that for generation of a blockchain, doesn't it?
Can I reuse this context? :thinking:

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
gromakovsky: Because it's not related to this task. Also it's not really important for openings and shares, because they do almost nothing when openings/shares are empty.

test/Test/Pos/Block/Logic/Mode.hs
gromakovsky: I think you can use `RealMode`. Why do you need this context? It will be changed to use pure db after Misha implements pure DB, I think. And it doesn't make sense to use pure DB in generator.

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
pva701: >changing of names of slotting related functions
>it's not related to this task

gromakovsky: Yes, but name changes are only cosmetics, while this change affects behavior of the program.

PR #1025, opened 2017-06-24
[CSL-1009] - Move gitHash and gitBranch printing to executables

7sedam7:
It was building, tests were passing
after merge of master, I can not build any more, so if it doesn't build please let me know

### DIFF
--- a/src/addr-convert/Main.hs
+++ b/src/addr-convert/Main.hs
@@ -1,9 +1,12 @@
 {-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell     #-}
 
 module Main (main) where
 
 import           Universum
 
+import           Development.GitRev           (gitBranch, gitHash)
+import           System.Wlog                  (logInfo, usingLoggerName)
 import qualified Data.ByteString              as BS
 import           Data.String.QQ               (s)
 import qualified Data.Text                    as T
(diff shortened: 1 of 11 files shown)


src/addr-convert/Main.hs
```diff
@@ -81,3 +84,7 @@ fromAvvmPk addrText = do
 
 convertAddr :: Text -> IO Text
 convertAddr addr = pretty . makeRedeemAddress <$> fromAvvmPk (toText addr)
+
+
+main :: IO ()
```
gromakovsky: Why did you add one more `main` here?

src/node/Main.hs
```diff
@@ -4,18 +4,20 @@
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE TemplateHaskell     #-}
 
 module Main
        ( main
        ) where
 
+import           Development.GitRev  (gitBranch, gitHash)
```
gromakovsky: According to the [style-guide](https://github.com/serokell/serokell-util/blob/master/serokell-style.md) `Universum` should be a separate block of imports and 3rd-party libraries should be all in the same block.

gromakovsky:
@7sedam7 FYI: I force-pushed to this branch, reset your local version please.

gromakovsky:
I think it can be closed, because we need a different solution anyway, see https://issues.serokell.io/issue/CSL-1328#comment=96-5292

ChShersh:
Agree. If we need different solution so why spend time on non-viable solution? Does it improve anyone's life?

gromakovsky:
Well, this solution still improves our lifes, because it makes compilation faster than what we currently have. So if we know that we will not do CSL-1328 in near future, we can finish this effort at least (which seems much easier than CSL-1328).

PR #1026, opened 2017-06-25
[CSL-1286] Fix Production in node-sketch

gromakovsky:
I took the commit which was used before (a85eee0610) and
cherry-picked Misha's fix there. My commit is some arbitary commit,
but it's not worse than it was before.


volhovM:
Why?
https://github.com/serokell/time-warp-nt/pull/74
It's already in master. Set the commit dep to master.

PR #1029, opened 2017-06-26
[CSL-781] Move almost all executables to cardano-sl-tools

gromakovsky:
The main reason I am doing it is that currently every time one wants to change
something in tests, they have to wait until all executables are preprocessed.
It can be partially solved by `--for-install` flag, but:
• it's easier to forget to pass this flag than to pass the proper target;
• there still were about 5 exeuctables needed for installer.


gromakovsky:
One appveyor build failed with a strange error, but `pr` build passed and `branch` build is not required to pass, so I think it's fine to leave it as is.

domenkozar:
Since macOS is disabled, this introduced a regression in osx, leaving one `for-installer` flag condition around

PR #1031, opened 2017-06-26
[CSM-328] Wallet restore is not working

flyingleafe:
Wallet itself has been created deterministically from backup phrase, but underlying account and addresses were for some reason generated using random seed, which is completely wrong. This PR fixes it.


darko-mijic:
@akegalj and @Martoon-00: Is this now fixed? Can you do your reviews and verify that the bug is fixed?

PR #1033, opened 2017-06-26
CSE-162 explorer slow blocks

ksaric:
https://issues.serokell.io/issue/CSE-162

The backend work for CSE-162. The code could be better, I need to check `onApplyPageBlocksExplorer`, but I presume this is a general idea?

### DIFF
--- a/src/Pos/Block/Logic/Slog.hs
+++ b/src/Pos/Block/Logic/Slog.hs
@@ -30,7 +30,7 @@ import qualified Ether
 import           Formatting            (build, sformat, (%))
 import           Serokell.Util         (Color (Red), colorize)
 import           Serokell.Util.Verify  (formatAllErrors, verResToMonadError)
-import           System.Wlog           (WithLogger, logWarning)
+import           System.Wlog           (WithLogger)
 
 import           Pos.Binary.Core       ()
 import           Pos.Block.BListener   (MonadBListener (..))
@@ -159,14 +159,15 @@ slogApplyBlocks blunds = do
     -- Note: it's important to put blunds first
     mapM_ dbPutBlund blunds
     -- If the program is interrupted at this point (after putting on block),
-    -- we will rollback all wallet sets at the next launch.
-    onApplyBlocks blunds `catch` logWarn
+    -- we won't save any blocks.
+    callbackBlocks <- onApplyBlocks blunds
+
     let putTip =
             SomeBatchOp $
             GS.PutTip $ headerHash $ NE.last $ getOldestFirst blunds
     sanityCheckDB
     putSlottingData =<< GS.getSlottingData
-    return $ SomeBatchOp [putTip, forwardLinksBatch, inMainBatch]
+    return $ SomeBatchOp [putTip, callbackBlocks, forwardLinksBatch, inMainBatch]
   where
     blocks = fmap fst blunds
     forwardLinks = map (view prevBlockL &&& view headerHashG) $ toList blocks
@@ -175,9 +176,6 @@ slogApplyBlocks blunds = do
     inMainBatch =
         SomeBatchOp . getOldestFirst $
         fmap (GS.SetInMainChain True . view headerHashG . fst) blunds
-    -- ↓ was written by @pva701
-    logWarn :: SomeException -> m ()
-    logWarn = logWarning . sformat ("onApplyBlocks raised exception: " %build)
 
 -- | This function does everything that should be done when rollback
 -- happens and that is not done in other components.
@@ -191,14 +189,15 @@ slogRollbackBlocks blunds = do
         assertionFailed $
         colorize Red "FATAL: we are TRYING TO ROLLBACK 0-TH GENESIS block"
     -- If program is interrupted after call @onRollbackBlocks@,
-    -- we will load all wallet set not rolled yet at the next launch.
-    onRollbackBlocks blunds `catch` logWarn
+    -- we won't save any blocks.
+    callbackBlocks <- onRollbackBlocks blunds
+    
     let putTip = SomeBatchOp $
                  GS.PutTip $
                  headerHash $
                  (NE.last $ getNewestFirst blunds) ^. prevBlockL
     sanityCheckDB
-    return $ SomeBatchOp [putTip, forwardLinksBatch, inMainBatch]
+    return $ SomeBatchOp [putTip, callbackBlocks, forwardLinksBatch, inMainBatch]
   where
     blocks = fmap fst blunds
     inMainBatch =
@@ -209,6 +208,3 @@ slogRollbackBlocks blunds = do
         fmap (GS.RemoveForwardLink . view prevBlockL) blocks
     isGenesis0 (Left genesisBlk) = genesisBlk ^. epochIndexL == 0
     isGenesis0 (Right _)         = False
-    -- ↓ was written by @pva701
-    logWarn :: SomeException -> m ()
-    logWarn = logWarning . sformat ("onRollbackBlocks raised exception: "%build)
--- a/src/Pos/Explorer/BListener.hs
+++ b/src/Pos/Explorer/BListener.hs
@@ -39,6 +39,7 @@ import           Pos.Util.Chrono              (NE, NewestFirst (..),
 -- Declarations
 ----------------------------------------------------------------------------
 
+
 data ExplorerBListenerTag
 
 type ExplorerBListener = Ether.TaggedTrans ExplorerBListenerTag IdentityT
@@ -65,9 +66,11 @@ instance ( MonadDBRead m
     onApplyBlocks     blunds = onApplyCallGeneral blunds
     onRollbackBlocks  blunds = onRollbackCallGeneral blunds
 
+
 ----------------------------------------------------------------------------
 -- General calls
 ----------------------------------------------------------------------------
+
 
 onApplyCallGeneral
     :: forall ssc m . 
@@ -88,9 +91,11 @@ onRollbackCallGeneral blunds = do
     pageBlocks  <- onRollbackPageBlocksExplorer blunds
     pure $ SomeBatchOp [epochBlocks, pageBlocks]
 
+
 ----------------------------------------------------------------------------
 -- Function calls
 ----------------------------------------------------------------------------
+
 
 -- For @EpochBlocks@
 onApplyEpochBlocksExplorer
@@ -101,7 +106,7 @@ onApplyEpochBlocksExplorer
 onApplyEpochBlocksExplorer blunds = do
 
     -- Get existing @HeaderHash@es from the keys so we can merge them.
-    existingBlocks <- getExistingBlocks $ M.keys newBlocks
+    existingBlocks <- getExistingBlocks keys
 
     -- Merge the new and the old
     let mergedBlocksMap = M.unionWith (<>) existingBlocks newBlocks
@@ -114,46 +119,14 @@ onApplyEpochBlocksExplorer blunds = do
     pure $ SomeBatchOp mergedBlocks
   where
 
+    keys :: [Epoch]
+    keys = M.keys newBlocks
+
     newBlocks :: M.Map Epoch [HeaderHash]
     newBlocks = epochBlocksMap blocksNE
 
     blocksNE :: NE (Block ssc)  
     blocksNE = fst <$> getOldestFirst blunds
-
-    -- Get exisiting key blocks paired with the key.
-    getExistingBlocks
-        :: (MonadDBRead m) 
-        => [Epoch]
-        -> m (M.Map Epoch [HeaderHash])
-    getExistingBlocks keys = do
-        keyBlocks <- sequence [ getExistingKeyBlocks key | key <- keys ]
-        pure $ M.unions keyBlocks
-      where
-        -- Get exisiting key blocks paired with the key. If there are no
-        -- saved blocks on the key return an empty list.
-        getExistingKeyBlocks 
-            :: (MonadDBRead m)
-            => Epoch 
-            -> m (M.Map Epoch [HeaderHash])
-        getExistingKeyBlocks key = do
-            keyBlocks        <- DB.getEpochBlocks key
-            let mKeyBlocks = fromMaybe [] keyBlocks
-            pure $ M.singleton key mKeyBlocks
-
-    -- For each (k, [v]) pair, create a database operation.
-    putKeysBlocks 
-        :: M.Map Epoch [HeaderHash]
-        -> [DB.ExplorerOp]
-    putKeysBlocks keysBlocks = putKeyBlocks <$> M.toList keysBlocks
-      where
-        putKeyBlocks 
-            :: (Epoch, [HeaderHash])
-            -> DB.ExplorerOp
-        putKeyBlocks keyBlocks = DB.PutEpochBlocks key uniqueBlocks
-          where
-            key           = keyBlocks ^. _1
-            blocks        = keyBlocks ^. _2
-            uniqueBlocks  = ordNub blocks
 
 
 -- For @PageBlocks@
@@ -165,7 +138,7 @@ onApplyPageBlocksExplorer
 onApplyPageBlocksExplorer blunds = do
 
     -- Get existing @HeaderHash@es from the keys so we can merge them.
-    existingBlocks <- getExistingBlocks $ M.keys newBlocks
+    existingBlocks <- getExistingBlocks keys
 
     -- Merge the new and the old
     let mergedBlocksMap = M.unionWith (<>) existingBlocks newBlocks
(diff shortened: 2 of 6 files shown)


src/Pos/Block/Logic/Slog.hs
```diff
@@ -159,14 +159,15 @@ slogApplyBlocks blunds = do
     -- Note: it's important to put blunds first
     mapM_ dbPutBlund blunds
     -- If the program is interrupted at this point (after putting on block),
-    -- we will rollback all wallet sets at the next launch.
-    onApplyBlocks blunds `catch` logWarn
+    -- we won't save any blocks.
+    callbackBlocks <- onApplyBlocks blunds
```
gromakovsky: 1. I would call this variable `bListenerBatch` or whatever else with `Batch` suffix. When I see variable named `***Callback` I expect it to be a function (but here it is `SomeBatchOp`).
2. This `logWarn` looked strange to me, so I don't mind removing it.
3. @pva701 I think something should be done with these comments. Your comments were bad because they are about `BListener` used in wallet. But we have other `BListener`s and it doesn't make sense to have it here.

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,247 @@
...
+import           Control.Monad.Trans.Identity (IdentityT (..))
+import           Data.Coerce                  (coerce)
+import qualified Ether
+
+import           Pos.Block.Types              (Blund)
+import           Pos.DB.BatchOp               (SomeBatchOp)
+import           Pos.Ssc.Class.Helpers        (SscHelpersClass)
+import           Pos.Util.Chrono              (NE, NewestFirst (..),
+                                               OldestFirst (..))
+import           Universum
```
gromakovsky: Please follow our [style-guide](https://github.com/serokell/serokell-util/blob/master/serokell-style.md#imports) to group imports.

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,247 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | Blockchain listener for Explorer.
+-- Callbacks on application and rollback.
+
+module Pos.Explorer.BListener
+       ( runExplorerBListener -- (+) BListener instance
```
gromakovsky: See how exported instances are documented in `core/Pos.Util.Util` for example.

src/Pos/Explorer/DB.hs
```diff
@@ -28,6 +30,17 @@ import           Pos.Txp.Toil          (Utxo)
 import           Pos.Util.Chrono       (NewestFirst (..))
 
 ----------------------------------------------------------------------------
+-- Types
+----------------------------------------------------------------------------
+
+type Page = Integer
```
gromakovsky: What is `Page`?

src/Pos/Explorer/DB.hs
```diff
@@ -28,6 +30,17 @@ import           Pos.Txp.Toil          (Utxo)
...
+type Epoch = Integer
```
gromakovsky: Why not use `EpochIndex`?

src/Pos/Block/Logic/Slog.hs
pva701: @gromakovsky which comments?

ksaric: @pva701 I thinks he meant line 161-162 (look at the previous version of the code). But I don't know what to do about the comments, but it's a good idea to see if you need to move this comment elsewhere.

src/Pos/Explorer/DB.hs
ksaric: It's a simpler construct introduced in `explorer`, a `Page` contains 10 `Block`s - https://github.com/input-output-hk/cardano-sl-explorer/blob/master/src/Pos/Explorer/Web/Api.hs#L48-L60
This simplified the `explorer` socket.io implementation quite a bit, as well as made it much simpler from the frontend perspective.

src/Pos/Explorer/DB.hs
ksaric: I don't like to wrap and unwrap code too much, the client is sending an `Integer`, so I decided that `Integer` is the most simple thing to do. I can change it if you think `EpochIndex` is better.

src/Pos/Explorer/BListener.hs
ksaric: Cool. I was actually thinking of proposing something like this, didn't even know you had this already.

src/Pos/Block/Logic/Slog.hs
gromakovsky: @pva701 yes, I mean comment from lines 161-162.

src/Pos/Explorer/DB.hs
gromakovsky: Ok, I don't mind having this alias, but probably it can be bounded (e. g. `Int64`)? I don't think it will ever be greater than 10¹⁸.

src/Pos/Explorer/DB.hs
gromakovsky: > I don't like to wrap and unwrap code too much, the client is sending an Integer,

Then transformation should be done quickly after receiving it from the client. I don't know that part well, but virtually all `cardano-sl` uses `EpochIndex` to refer to epoch. And block stores it as `EpochIndex`. So it doesn't make sense to use something different, it will most likely involve even more wrapping and unwrapping eventually and it's inconsistent.
If epoch arrives as `Integer` from the client, it's ok, but it should be converted to appropriate type as soon as possible.
Also `Integer` doesn't make sense here, because it doesn't seem possible that epoch will ever be greater than 10¹⁸.

src/Pos/Explorer/BListener.hs
gromakovsky: Well, the link is in [README](https://github.com/input-output-hk/cardano-sl#for-contributors), also it can be found in [cardano-sl-style-guides repository](https://github.com/input-output-hk/cardano-sl-style-guides/blob/master/haskell-style-guide.md).

src/Pos/Explorer/DB.hs
ksaric: Fixed.

src/Pos/Explorer/BListener.hs
ksaric: Fixed.

src/Pos/Explorer/DB.hs
ksaric: Fixed.

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+    -- add the (Epoch, [HeaderHash]) pair back to the database).
+    mBlocksHHs <- DB.getEpochBlocks epoch
+    let blocksHHsOrEmpty = fromMaybe mempty mBlocksHHs
+    let uniqueHHs = ordNub $ blocksHHsOrEmpty <> blockHHs
+
+    -- In the end make sure we place this under @SomeBatchOp@ in order to
+    -- preserve atomicity.
+    pure $ SomeBatchOp $ DB.PutEpochBlocks epoch uniqueHHs
+  where
+    epochBlocks = minEpochBlocks blocksNE
```
gromakovsky: `minEpochBlocks` seems useless here, because we know that all blocks are from the same epoch (which is false for rollback, btw).

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+----------------------------------------------------------------------------
+-- Apply
+----------------------------------------------------------------------------
+
+-- For @EpochBlocks@
+onApplyEpochBlocksExplorer
+    :: forall ssc m .
+    ( SscHelpersClass ssc
+    , WithLogger m
+    , MonadRealDB m
```
gromakovsky: I guess `MonadRealDB` shouldn't be needed here.

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+onApplyPageBlocksExplorer
+    :: forall ssc m .
+    ( SscHelpersClass ssc
+    , WithLogger m
+    , MonadRealDB m
+    , MonadDBRead m
+    )
+    => OldestFirst NE (Blund ssc) -> m SomeBatchOp
+onApplyPageBlocksExplorer blunds = do
+    -- [(Integer, [HeaderHash])]
```
gromakovsky: 1. It's not `Integer`, it's `Page`.
2. You can write real type signature instead of comment:
```
    let newPageBlocks :: [(Page, [HeaderHash])]
        newPageBlocks = …
```

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+    blockIndexBlock = blockPages `zip` blocksHHs
+      where 
+        blocksHHs :: [HeaderHash]
+        blocksHHs = headerHash <$> blocks 
+
+        blockPages :: [Page]
+        blockPages = currentPage <$> blockIndexes
+
+        currentPage :: Int -> Page
+        currentPage blockIndex = blockIndex `rem` pageSize
```
gromakovsky: Shouldn't it be `div` instead?

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+    => OldestFirst NE (Blund ssc) -> m SomeBatchOp
+onApplyPageBlocksExplorer blunds = do
+    -- [(Integer, [HeaderHash])]
+    let newPageBlocks = blocksPagePages blocksNE
+    let pages = fst <$> newPageBlocks
+    -- Get existing @HeaderHash@es from the pages so we can merge them.
+    existingPageBlocks <- sequence [ getPagePageBlocks page | page <- pages]
+
+    -- convert to Map
+    let newPageBlocksMap      = M.fromList newPageBlocks
```
gromakovsky: You apply `M.toList` in the end of `blocksPagePages` and here you apply `M.fromList`. It seems quite strange to me. Why not return map from `blocksPagePages`?

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+    let existingPageBlocksMap = M.fromList existingPageBlocks
+
+    -- Merge the old and the new
+    let newExistingPageBlocks = M.toList $ M.unionWith (<>) newPageBlocksMap existingPageBlocksMap
+
+    let deletedPages = [ DB.DelPageBlocks page | page <- pages]
```
gromakovsky: What's the purpose of deleting pages? `PutPageBlocks` will override existing page if it's stored already.

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+    :: forall ssc m .
+    ( SscHelpersClass ssc
+    , WithLogger m
+    , MonadRealDB m
+    , MonadDBRead m
+    )
+    => NewestFirst NE (Blund ssc) -> m SomeBatchOp
+onRollbackEpochBlocksExplorer blunds = do
+    let epoch    = epochBlocks ^. _1
+    pure $ SomeBatchOp $ DB.DelEpochBlocks epoch
```
gromakovsky: Wait, what is it? It's completely wrong. For example, suppose we have 4 blocks for epoch 3. And we rollback only one of these blocks. You will delete whole index for epoch 3. But you should remove only one element. You can get whole list, remove one item from there and put it back (or delete if it becomes empty).

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,255 @@
...
+
+onRollbackPageBlocksExplorer
+    :: forall ssc m .
+    ( SscHelpersClass ssc
+    , WithLogger m
+    , MonadRealDB m
+    , MonadDBRead m
+    )
+    => NewestFirst NE (Blund ssc) -> m SomeBatchOp
+onRollbackPageBlocksExplorer blunds = do
```
gromakovsky: This function is wrong too, see comments to `onRollbackEpochBlocksExplorer`.

src/Pos/Explorer/BListener.hs
gromakovsky: Actually I think `MonadRealDB` is not needed in this whole module, but I am not sure.

src/Pos/Block/Logic/Slog.hs
gromakovsky: @ksaric @pva701 I pushed a commit which makes comments and variable name more reasonable to me.

src/Pos/Explorer/BListener.hs
ksaric: Ah. Did not know that, this was a general solution. Will remove.

src/Pos/Explorer/BListener.hs
ksaric: Ah, dammit. You are right.

src/Pos/Explorer/BListener.hs
ksaric: Yes, correct again.

src/Pos/Explorer/BListener.hs
ksaric: I forgot about that comment, will add type signature.

src/Pos/Explorer/BListener.hs
ksaric: Fair enough.

src/Pos/Explorer/BListener.hs
ksaric: Yes, it should. :man_facepalming:

src/Pos/Explorer/BListener.hs
ksaric: I'll check that out, probably there isn't a good reason.

ksaric:
You can see, there is a lot of repetition. I can't make types align, tried with something like this:
```
class BlockKey k where
    newBlocksMapF :: (SscHelpersClass ssc) => (NE (Block ssc) -> M.Map Epoch [HeaderHash])
    getKeyBlocksF :: (MonadDBRead m) => (k -> m (Maybe [HeaderHash]))
    putKeyBlocksF :: (k -> [HeaderHash] -> ExplorerOp)

instance BlockKey Epoch where
    newBlocksMapF = epochBlocksMap
    getKeyBlocksF = DB.getEpochBlocks
    putKeyBlocksF = DB.PutEpochBlocks

instance BlockKey Page where
    newBlocksMapF = pageBlocksMap
    getKeyBlocksF = DB.getPageBlocks
    putKeyBlocksF = DB.PutPageBlocks
```

In these functions there is a general shape that can be reused in all the project, it's in the form `k -> [v]`. Since the results of `BListener` end up in the key value database, things align nicely.
In this case it's `k -> [HeaderHash]`, where `k € {Page, Epoch}`.

What we could have is:
- a way to produce keys `[Blund] -> [k]`
- a way to produce mapping from key to values `[Blund] -> k -> [v]`

This would be general enough for most DB operations I guess, we can add specializations (on the function above) like:
- `k -> v`
- `[v]`

It's just an idea, but it evolves from these functions and general functions could be moved to `BListener`.

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,481 @@
...
+    :: forall ssc m .
+    MonadBListenerT m ssc
+    => NewestFirst NE (Blund ssc) -> m SomeBatchOp
+onRollbackEpochBlocksExplorer blunds = do
+    
+    -- Get existing @HeaderHash@es from the keys so we can merge them.
+    existingBlocks <- getExistingBlocks keys
+
+    -- Diffrence between the new and the old
+    let mergedBlocksMap = rollbackedBlocks keys newBlocks existingBlocks
```
gromakovsky: I see `rollbackedBlocks keys' oldMap' newMap'`, but here you pass `newBlocks` before `existingBlocks`. Is it by mistake?

src/Pos/Explorer/BListener.hs
```diff
@@ -0,0 +1,481 @@
...
+    :: forall ssc m .
+    MonadBListenerT m ssc
+    => NewestFirst NE (Blund ssc) -> m SomeBatchOp
+onRollbackPageBlocksExplorer blunds = do
+    
+    -- Get existing @HeaderHash@es from the keys so we can merge them.
+    existingBlocks <- getExistingBlocks keys
+
+    -- Diffrence between the new and the old
+    let mergedBlocksMap = rollbackedBlocks keys newBlocks existingBlocks
```
gromakovsky: Here the order of arguments looks wrong too.

src/Pos/Explorer/DB.hs
```diff
@@ -77,17 +99,38 @@ putGenesisBalances genesisUtxo =
 data ExplorerOp
     = AddTxExtra !TxId !TxExtra
     | DelTxExtra !TxId
+
+    | PutPageBlocks !Page ![HeaderHash]
+    | DelPageBlocks !Page
+
+    | PutEpochBlocks !Epoch ![HeaderHash]
+    | DelEpochBlocks !Epoch
```
gromakovsky: Btw, `Del` operations are not used. There are two options:
1. In rollback you can check whether the list you put is empty (it might be empty sometimes) and use `Del` operation if it is.
2. You can never delete, because even if rollback happens and we put empty list, we know that it will be refilled very soon, so it's negligible overhead to store empty lists. In this case I propose to remove these operations altogether.

ksaric: I agree. I forgot about those two, to be honest.

src/Pos/Explorer/BListener.hs
ksaric: Yes. A stupid one.

src/Pos/Explorer/BListener.hs
ksaric: Yep. 2/2 :man_facepalming:

gromakovsky:
@ksaric you should also merge `cardano-sl-0.5-rc` to `master`.

PR #1039, opened 2017-06-27
[CSM-330] Fix redeeming throwing an error

Pastafarianist:


Pastafarianist:
Tests pending, but code is no longer WIP.

PR #1045, opened 2017-06-28
Tests alexrb csl1161 arbitrary

rockbmb:


core/Pos/Core/Arbitrary.hs
```diff
@@ -62,6 +62,8 @@ instance Arbitrary Types.Address where
         Types.UnknownAddressType <$> choose (3, 255) <*> scale (min 150) arbitrary
         ]
 
+-- NOTE: Deriving an 'Arbitrary' instance for wrappers over integral types does not use
```
int-index: I don't think this has to do with the type being an instance of `Integral`, as your comment may suggest. What's happening is that `deriving` for newtypes uses the `GND` deriving strategy, so the `shrink` method is inherited from the underlying type. Since `ChainDifficulty` is a newtype over `Word64`, we get a `shrink` impl that tries to bring the value to `0` (which is not strictly a "better" instance, depending on what values you consider to be trivial in your domain. For example, if you represent unique identifiers as numbers, you'd probably want no shrinking to take place).

core/Pos/Core/Arbitrary.hs
```diff
@@ -71,25 +73,30 @@ deriving instance Random Types.EpochIndex
...
     arbitrary = choose (0, maxReasonableEpoch)
+    shrink = genericShrink
 
 instance Arbitrary Types.LocalSlotIndex where
     arbitrary =
         leftToPanic "arbitrary@LocalSlotIndex: " . Types.mkLocalSlotIndex <$>
         choose (Types.getSlotIndex minBound, Types.getSlotIndex maxBound)
+    shrink = genericShrink
 
+-- NOTE: Deriving an 'Arbitrary' instance for wrappers over integral types with 'derive
```
int-index: First, this instance uses `genericArbitrary`, not `derive makeArbitrary`. Second, you are repeating yourself. To have the same commentary in multiple places there's a good practice popularized by GHC devs to leave so called "NOTE"s — you can read about them here - https://ghc.haskell.org/trac/ghc/wiki/Commentary/CodingStyle#Commentsinthesourcecode

I also recall that @pva701  used them somewhere in our codebase (grep the sources for `~~~~`).

core/Pos/Core/Arbitrary.hs
```diff
@@ -318,26 +332,42 @@ instance Arbitrary G.StakeDistribution where
...
       ]
+    shrink = genericShrink
 
 ----------------------------------------------------------------------------
 -- Arbitrary miscellaneous types
 ----------------------------------------------------------------------------
 
 instance Arbitrary Millisecond where
     arbitrary = fromMicroseconds <$> choose (0, 600 * 1000 * 1000)
+    shrink = fmap fromIntegral . shrink . toInteger
```
int-index: How do you know that this is safe? What if the shrinking strategy for `Integer` changes and we start to get under/overflow? I'm not saying that this will happen, but I'm saying that this code is correct by coincidence.

What you should do instead is use `shrinkIntegral` and newtype wrapping/unwrapping. This criticism applies to both `Millisecond` and `Microsecond` instances (slightly below).

core/Pos/Util/Util.hs
```diff
@@ -349,3 +352,18 @@ _neTail f (x :| xs) = (x :|) <$> f xs
...
+-- compilation, as a GHC warning). Useful for debugging.
+--
+-- For instance, you can dump splices generated with 'makeLenses' by
+-- replacing a top-level invocation of 'makeLenses' in your code with:
+--
+-- @dumpSplices $ makeLenses ''Foo@
+--
+dumpSplices :: TH.DecsQ -> TH.DecsQ
+dumpSplices x = do
+  ds <- x
```
int-index: Indent with 4 spaces.

godtossing/Pos/Ssc/GodTossing/Arbitrary.hs
```diff
@@ -103,18 +107,22 @@ commitmentsAndOpenings =
...
 instance Arbitrary CommitmentOpening where
     arbitrary = elements commitmentsAndOpenings
+    shrink = genericShrink
 
 instance Nonrepeating CommitmentOpening where
     nonrepeating n = sublistN n commitmentsAndOpenings
 
 instance Arbitrary Commitment where
     arbitrary = coCommitment <$> arbitrary
+    shrink Commitment {..} = [ Commitment { commShares = shrunkShares, .. }
```
int-index: How do you know that this shrinking preserves the invariant of the data type (see comment where `Commitment` is defined)?

```
'checkSig vcSigningKey (vcVssKey, vcExpiryEpoch) vcSignature'.
```

int-index: Actually, you should check invarians on all data types. If you risk breaking it, it's better to have no shrinking (otherwise you can't rely on the invariant in tests).

int-index: This applies to both `shrink` and `arbitrary`.

src/Pos/Block/Arbitrary.hs
```diff
@@ -167,12 +157,13 @@ instance Arbitrary (SscPayloadDependsOnSlot ssc) =>
...
 
 instance Arbitrary (SscPayload ssc) => Arbitrary (T.Body (T.MainBlockchain ssc)) where
-    arbitrary = makeSmall $ do
-        txPayload   <- arbitrary
-        mpcData     <- arbitrary
-        mpcProxySKs <- arbitrary
-        mpcUpload   <- arbitrary
-        return $ T.MainBody txPayload mpcData mpcProxySKs mpcUpload
+    arbitrary = makeSmall genericArbitrary
+    shrink mb = [ T.MainBody txp sscp dlgp updp
```
int-index: No way, I don't think that the list monad (or list comprehension) is the right approach to writing shrink. This will try to shrink each field separately, creating a blowing up the space of shrinking choices. 

I refer you to the Haddock documentation of the `shrink` method where this issue is mentioned:
image

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -128,7 +129,7 @@ syncWalletsAtStartWebWallet :: WalletTrackingEnv ext m => [EncryptedSecretKey] -
 syncWalletsAtStartWebWallet = syncWalletsWithGStateLock @WalletSscType
 
 syncOnImportWebWallet :: WalletTrackingEnv ext m => EncryptedSecretKey -> m ()
-syncOnImportWebWallet = selectAccountsFromUtxoLock @WalletSscType . one
+syncOnImportWebWallet = (() <$) . selectAccountsFromUtxoLock @WalletSscType . one
```
int-index: How come that this change (and quite a few of other changes) is displayed in this PR (#1045)? It has no relation to `Arbitrary` and belongs to a different branch. You should've forked from a recent `master`.

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -173,6 +174,15 @@ selectAccountsFromUtxoLock encSKs = withBlkSemaphore_ $ \tip -> do
         guard $ length derPath == 2
         pure $ CWAddressMeta wAddr (derPath !! 0) (derPath !! 1) (addressToCId addr)
 
+    addMetaInfo :: CWAddressMeta -> m ()
+    addMetaInfo cwMeta = do
```
int-index: All of this appears to have no relation to CSL-1161.

txp/Pos/Txp/Arbitrary.hs
```diff
@@ -51,9 +63,16 @@ instance Arbitrary TxInWitness where
         ScriptWitness <$> arbitrary <*> arbitrary,
         RedeemWitness <$> arbitrary <*> arbitrary,
         UnknownWitnessType <$> choose (3, 255) <*> scale (min 150) arbitrary ]
+    shrink = genericShrink
```
int-index: Invariants. I see `choose (3, 255)` for `UnknownWitnessType` above in the `arbitrary` method impl. How do you know that it's not violated by `genericShrink` which would try to bring this numeric value down to `0`?

src/Pos/Block/Arbitrary.hs
rockbmb: Is the hanging colon supposed to be here?

src/Pos/Wallet/Web/Tracking.hs
rockbmb: I rebased a change into a commit (fixing CI by removing unused imports), and now some other commits show up with my name under them. This has happened once before, I'll need to look into it to fix the history.

src/Pos/Block/Arbitrary.hs
int-index: There's a picture attached to my comment. Probably takes a few seconds to load, but it's a screenshot of the docs.

rockbmb: My browser hid it (at my request), I see the warning now.

core/Pos/Core/Arbitrary.hs
rockbmb: I wrote this and the next comment as a reminder of this information, but I quickly realized they're not worded very well because I'm not specific and detailed enough for them to make sense to anyone that drops into the codebase and takes a look. I'll write them on my Slack DMs to myself, this probably shouldn't be in the codebase.

core/Pos/Core/Arbitrary.hs
rockbmb: Ah, I just now realized I wrote this comment after I used `dumpSplices` to verify that `derive makeArbitrary` does NOT define `shrink` in these case, changed the instances, and then forgot to update the verb's tense.

src/Pos/Block/Arbitrary.hs
rockbmb: 🤦‍♂️ 
Thanks for spotting this. That's the Haddock page I visit the most, yet I forgot that.

core/Pos/Core/Arbitrary.hs
```diff
@@ -318,26 +354,42 @@ instance Arbitrary G.StakeDistribution where
...
 instance Arbitrary SmallHashMap where
     arbitrary = SmallHashMap <$> makeSmall arbitrary
+    shrink = genericShrink
 
-derive makeArbitrary ''UnsignedVarInt
-derive makeArbitrary ''SignedVarInt
-derive makeArbitrary ''FixedSizeInt
+instance Arbitrary a => Arbitrary (UnsignedVarInt a) where
+    arbitrary = genericArbitrary
+    shrink = genericShrink
```
gromakovsky: Perhaps it makes sense to add `Integral` constraint to VarInt instances and use `shrinkIntegral`.

godtossing/Pos/Ssc/GodTossing/Arbitrary.hs
```diff
@@ -128,6 +136,9 @@ instance Arbitrary Opening where
 
 instance Arbitrary VssCertificate where
     arbitrary = mkVssCertificate <$> arbitrary <*> arbitrary <*> arbitrary
+    shrink VssCertificate {..} = [ VssCertificate { vcExpiryEpoch = shrunkEpoch, .. }
+                                 | shrunkEpoch <- shrink vcExpiryEpoch
```
gromakovsky: It seems broken to me. If you change `vcExpiryEpoch` in `VssCertificate`, the signature inside it becomes invalid.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
```diff
@@ -22,7 +22,7 @@ module Pos.Ssc.GodTossing.Core.Types
        , SharesDistribution
 
          -- * Vss certificates
-       , VssCertificate (vcVssKey, vcExpiryEpoch, vcSignature, vcSigningKey)
+       , VssCertificate (..)
```
gromakovsky: It's totally bad. Constructor of `VssCertificate` shouldn't be exported, it's unsafe.

txp/Pos/Txp/Arbitrary.hs
```diff
@@ -118,9 +141,9 @@ buildProperTx triplesList (inCoin, outCoin) = fmap newTx txList
 -- | Well-formed transaction 'Tx'.
 newtype GoodTx = GoodTx
     { getGoodTx :: NonEmpty ((Tx, TxDistribution), TxIn, TxOutAux, TxInWitness)
-    } deriving (Show)
+    } deriving (Generic, Show)
 
-newtype SmallGoodTx = SmallGoodTx { getSmallGoodTx :: GoodTx } deriving (Show)
+newtype SmallGoodTx = SmallGoodTx { getSmallGoodTx :: GoodTx } deriving (Generic, Show)
```
gromakovsky: Btw, I think we should get rid of these `Small…` types and have `newtype Small a = Small {getSmall :: a}`.

gromakovsky: Probably not in this task. Just saying (wanted to say long ago).

rockbmb: I'll address this in a different task.

godtossing/Pos/Ssc/GodTossing/Arbitrary.hs
```diff
@@ -136,9 +136,6 @@ instance Arbitrary Opening where
 
 instance Arbitrary VssCertificate where
     arbitrary = mkVssCertificate <$> arbitrary <*> arbitrary <*> arbitrary
-    shrink VssCertificate {..} = [ VssCertificate { vcExpiryEpoch = shrunkEpoch, .. }
```
int-index: Write a comment instead, so that no one thinks that you just forgot to write `shrink`.

godtossing/Pos/Ssc/GodTossing/Arbitrary.hs
int-index: I still expect a comment for each non-generic implementation. Here I want to see why we're not shrinking `vcSigningKey`, `vcVssKey` and `vcExpiryEpoch` (because of the invariant).

rockbmb: You're right, but it's largely due to not having `shrink`s for crypto datatypes.

rockbmb:
@int-index Should be better now.

core/Pos/Core/Arbitrary.hs
```diff
@@ -320,26 +354,42 @@ instance Arbitrary G.StakeDistribution where
...
+
+instance (Arbitrary a, Integral a) => Arbitrary (SignedVarInt a) where
+    arbitrary = genericArbitrary
+    shrink = genericShrink
+
+instance (Arbitrary a, Integral a) => Arbitrary (FixedSizeInt a) where
+    arbitrary = genericArbitrary
+    shrink = genericShrink
```
gromakovsky: So you added `Integral` constraint, but didn't change it to `shrinkIntegral`, it seems strange.

gromakovsky: Hm, after thinking about it for a while, I think it's better to just derive `Arbitrary` instance using `GeneralizedNewTypeDeriving`. My advice to use `shrinkIntegral` is bad, because underlying type may use more complicated shrinking and we would like to preserve it for varint type.
So I would write
```
deriving instance Arbitrary a => Arbitrary (UnsignedVarInt a)
deriving instance Arbitrary a => Arbitrary (SignedVarInt a)
deriving instance Arbitrary a => Arbitrary (FixedSizeInt a)
```

rockbmb: Last time I had to merge `master`, I messed up 3 or 4 times and forgot to fix several little things each time, with my last force push I must have forgotten something.

PR #1046, opened 2017-06-28
[DEVOPS-32] initial hydra jobset

cleverca22:
this PR allows hydra to build every PR for cadrano-sl, and report test pass/failure back to github

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,14 @@ language: nix
 sudo: true
 matrix:
   include:
-    - env: HLINT=true
-    - env: HLINT=false
+    - env: MODE=hlint
+    - env: MODE=normal
+    - env: MODE=nix
+    - env: MODE=stack2nix
     - os: osx
       osx_image: xcode8
       # don't use sierra (xcode 8.1+) because of too restrictive linker path (DEVOPS-168)
-env:
-  global:
-    - NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/8bed8fb53227932886ab23e5f5f9eabe139f8e9f.tar.gz
+
 cache:
   directories:
   - "$HOME/.stack"
@@ -31,14 +31,25 @@ before_install:
 - sudo /etc/init.d/memcached stop || true
 - sudo /etc/init.d/mysql stop || true
 - sudo /etc/init.d/postgresql stop || true
+- cat /etc/nix/nix.conf > ~/nix.conf
+- echo "binary-caches = https://cache.nixos.org https://hydra.iohk.io" >> ~/nix.conf
+- echo "binary-cache-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" >> ~/nix.conf
+- export NIX_CONF_DIR=~
+
+install:
+  - source scripts/set_nixpath.sh
 
 after_failure:
 - sudo dmesg
 
 script:
 # DEVOPS-107: wait 20min for output due to haddock generation
-- if [[ "$HLINT" == "true" ]]; then
+- if [[ "$MODE" == "nix" ]]; then
+    nix-build -A cardano-sl --no-build-output --keep-going || nix-build -A cardano-sl -j 1 ;
+  elif [[ "$MODE" == "hlint" ]]; then
     nix-shell -p haskellPackages.hlint moreutils --run "scripts/haskell/lint.sh --nix | ts";
+  elif [[ "$MODE" == "stack2nix" ]]; then
+    nix-shell -p cabal2nix stack cabal-install ghc -Q -j 4 --run scripts/check-stack2nix.sh ;
   else
     nix-shell -p moreutils expect --run "unbuffer ./scripts/ci/travis.sh | ts";
   fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ after_failure:
 script:
 # DEVOPS-107: wait 20min for output due to haddock generation
 - if [[ "$MODE" == "nix" ]]; then
-    nix-build -A cardano-sl --no-build-output --keep-going || nix-build -A cardano-sl -j 1 ;
+    nix-build -A cardano-sl-static --no-build-output --keep-going || nix-build -A cardano-sl -j 1 ;
   elif [[ "$MODE" == "hlint" ]]; then
     nix-shell -p haskellPackages.hlint moreutils --run "scripts/haskell/lint.sh --nix | ts";
   elif [[ "$MODE" == "stack2nix" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ matrix:
       osx_image: xcode8
       # don't use sierra (xcode 8.1+) because of too restrictive linker path (DEVOPS-168)
 
+  allow_failures:
+    - env: MODE=nix
+    # https://github.com/input-output-hk/stack2nix/issues/15
+    # stack2nix will randomly omit sections of the generated file
+    - env: MODE=stack2nix
+
 cache:
   directories:
   - "$HOME/.stack"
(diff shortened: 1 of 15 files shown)


cleverca22:
```
Pos/Binary/Crypto.hs:113:1: error:
    Not in scope: ‘Store.makeStore’
    Neither ‘Data.Store’ nor ‘Data.Store.TH’ exports ‘makeStore’.
```
is this something i havent configured right, or an unknown bug?

domenkozar:
```
$ git grep nixpkgs
.travis.yml:    - NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/8bed8fb53227932886ab23e5f5f9eabe139f8e9f.tar.gz
.travis.yml:# TODO: https://github.com/NixOS/nixpkgs/issues/21200
daedalus/default.nix:with (import (fetchTarball https://github.com/NixOS/nixpkgs/archive/8bed8fb53227932886ab23e5f5f9eabe139f8e9f.tar.gz) {});
shell.nix:with import (fetchTarball https://github.com/NixOS/nixpkgs/archive/8bed8fb53227932886ab23e5f5f9eabe139f8e9f.tar.gz) { };
```

cleverca22:
@domenkozar updated the PR to get more of those nixpkgs references

scripts/set_nixpath.sh
```diff
@@ -0,0 +1,6 @@
+update_NIX_PATH() {
+  local scriptDir=$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")
+  export NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/$(jq .rev < ${scriptDir}/../nixpkgs-src.json -r).tar.gz
```
domenkozar: better to use `nix-shell -p jq --run "jq..."

domenkozar:
I have restarted the `mode=nix` to see how long it takes with the binary cache.

default.nix
```diff
@@ -0,0 +1,34 @@
...
+        "-f-asserts"
+        "-f-dev-mode"
+        "-fwith-explorer"
+        # https://github.com/NixOS/nixpkgs/pull/24692#issuecomment-306509337
+        "--ghc-option=-optl-lm"
+      ];
+    });
+    cardano-sl-static = justStaticExecutables self.cardano-sl;
+    # Gold linker fixes
+    cryptonite = addConfigureFlags ["--ghc-option=-optl-pthread"] super.cryptonite;
```
nh2: We should file an issue for this for cryptonite. It's maintained by @vincenthz

default.nix
```diff
@@ -0,0 +1,34 @@
...
+      patchPhase = ''
+       export CSL_SYSTEM_TAG=linux64
+      '';
+      # production full nodes shouldn't use wallet as it means different constants
+      configureFlags = [
+        "-f-asserts"
+        "-f-dev-mode"
+        "-fwith-explorer"
+        # https://github.com/NixOS/nixpkgs/pull/24692#issuecomment-306509337
+        "--ghc-option=-optl-lm"
```
nh2: We should probably add this already to cardano.

pkgs/generate.sh
```diff
@@ -0,0 +1,30 @@
...
+  nix-shell -j 4 -p cabal2nix nix-prefetch-scripts coreutils cabal-install stack --run "$*"
+}
+
+set -xe
+set -v
+
+# Get relative path to script directory
+scriptDir=$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")
+
+source ${scriptDir}/../scripts/set_nixpath.sh
```
nh2: Might be good to have all paths with vars inside `"quoted"` as we don't know if some people like spaces in their dir names

pkgs/generate.sh
```diff
@@ -0,0 +1,30 @@
+
+#!/bin/sh
```
nh2: For me hashbangs don't work when not in the first line.

scripts/check-stack2nix.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -xe
+
+fail_stack2nix_check() {
+  echo "ERROR: you need to run ./pkgs/generate.sh and commit the changes"
```
nh2: stderr

pkgs/generate.sh
```diff
@@ -0,0 +1,30 @@
...
+
```
nh2: For each script added, can you add a short explanation at the top about what it does and when it's to be used?

scripts/set_nixpath.sh
```diff
@@ -0,0 +1,6 @@
+update_NIX_PATH() {
+  local scriptDir=$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")
+  export NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/$(nix-shell -p jq --run "jq .rev < ${scriptDir}/../nixpkgs-src.json -r").tar.gz
+  export NIX_PATH_LOCKED=1
```
nh2: What does this env var do?

.travis.yml
```diff
@@ -31,14 +33,25 @@ before_install:
...
+  - source scripts/set_nixpath.sh
 
 after_failure:
 - sudo dmesg
 
 script:
 # DEVOPS-107: wait 20min for output due to haddock generation
-- if [[ "$HLINT" == "true" ]]; then
+- if [[ "$MODE" == "nix" ]]; then
+    nix-build -A cardano-sl-static --no-build-output --keep-going || nix-build -A cardano-sl -j 1 ;
```
cleverca22: the parallelism enabled by default makes the logs impossible to read (all output is interleaved)
`--no-build-output` solves that, but now you cant see the error from the last job that failed
so in the event of failures, run it again without `--no-build-output -j1`

that should probably be fixed within nix itself, as a new option

pkgs/generate.sh
cleverca22: ah, didnt notice that, and bash likes to cheat and just treat "invalid" files as shell scripts for you

scripts/set_nixpath.sh
cleverca22: at one point, i was going to have the nix expressions detect that `NIX_PATH` was already configured, and not have them re-import the nixpkgs specified in `nixpkgs-src.json`

that change has since been removed, but i am keeping the variable for if i bring it back later

pkgs/generate.sh
nh2: I think you'd have to quote all uses of `$scriptDir`, as in the above line and the below lines, as well.

scripts/set_nixpath.sh
nh2: OK, please add a comment so that from the code one can get that, otherwise the next person might delete it assuming it was forgotten to be removed.

.travis.yml
```diff
@@ -31,14 +33,25 @@ before_install:
 - sudo /etc/init.d/memcached stop || true
 - sudo /etc/init.d/mysql stop || true
 - sudo /etc/init.d/postgresql stop || true
+- cat /etc/nix/nix.conf > ~/nix.conf
+- echo "binary-caches = https://cache.nixos.org https://hydra.iohk.io" >> ~/nix.conf
+- echo "binary-cache-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=" >> ~/nix.conf
```
domenkozar: This is to avoid sudo usage, but we could use it with the workaround: https://github.com/NixOS/nixops/issues/693#issuecomment-312868027

cleverca22: ah, i'll keep that fix in mind, but i think the sudo version of travis has worse cpu performance

scripts/set_nixpath.sh
```diff
@@ -0,0 +1,13 @@
+set -x
+update_NIX_PATH() {
+  case $OSTYPE in darwin*)
```
domenkozar: Add a comment that this provides `readlink`

PR #1047, opened 2017-06-28
[CSM-335] Extrapolate transaction timestamp

ajnsit:
When it cannot be precisely determined from the blockchain. The extrapolation will be accurate as long as the slot duration doesn't change.

### DIFF
--- a/core/Pos/Core/Timestamp.hs
+++ b/core/Pos/Core/Timestamp.hs
@@ -4,6 +4,8 @@ module Pos.Core.Timestamp
        ( Timestamp (..)
        , timestampF
        , getCurrentTimestamp
+       , diffTimestamp
+       , addMicrosecondsToTimestamp
        ) where
 
 import           Universum
@@ -45,3 +47,9 @@ timestampF = build
 -- Get the current time as a timestamp
 getCurrentTimestamp :: IO Timestamp
 getCurrentTimestamp = Timestamp . round . (*1000000) <$> getPOSIXTime
+
+diffTimestamp :: Timestamp -> Timestamp -> Microsecond
+diffTimestamp t1 t2 = getTimestamp t1 - getTimestamp t2
+
+addMicrosecondsToTimestamp :: Microsecond -> Timestamp -> Timestamp
+addMicrosecondsToTimestamp m t = Timestamp { getTimestamp = (getTimestamp t) + m }
--- a/infra/Pos/Slotting/Util.hs
+++ b/infra/Pos/Slotting/Util.hs
@@ -32,6 +32,7 @@ import           System.Wlog            (WithLogger, logDebug, logError, logInfo
 import           Universum
 
 import           Pos.Core               (FlatSlotId, SlotId (..), Timestamp (..),
+                                         addMicrosecondsToTimestamp, diffTimestamp,
                                          flattenSlotId, getSlotIndex, slotIdF)
 import           Pos.Discovery.Class    (MonadDiscovery)
 import           Pos.Exception          (CardanoException)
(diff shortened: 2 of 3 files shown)


akegalj:
@gromakovsky pointed that we should use this solutionhttps://issues.serokell.io/issue/CSM-337#comment=96-4981

core/Pos/Core/Timestamp.hs
```diff
@@ -45,3 +47,9 @@ timestampF = build
 -- Get the current time as a timestamp
 getCurrentTimestamp :: IO Timestamp
 getCurrentTimestamp = Timestamp . round . (*1000000) <$> getPOSIXTime
+
+diffTimestamp :: Timestamp -> Timestamp -> Microsecond
+diffTimestamp t1 t2 = getTimestamp t1 - getTimestamp t2
+
+addMicrosecondsToTimestamp :: Microsecond -> Timestamp -> Timestamp
```
gromakovsky: Usually it's shorter to write `Timestamp ms + b` than `addMicrosecondsToTimestamp ms b` (which are same things). So I am not sure this function is needed.
If you think it's needed, I don't mind having it (it's subjective and I don't want insist on subjective changes).

infra/Pos/Slotting/Util.hs
```diff
@@ -50,15 +52,31 @@ getCurrentSlotFlat = fmap flattenSlotId <$> getCurrentSlot
...
        | siEpoch == sdPenultEpoch + 1 -> Just $ slotTimestamp siSlot sdLast
        | otherwise -> Nothing
   where
+    -- Extrapolate slotting data for arbitrary epochs
+    extrapolateSlottingData desiredEpoch =
+      let
+        -- Assuming these durations stay constant
+        blockDuration = diffTimestamp (esdStart sdLast) (esdStart sdPenult)
+        slotDuration = esdSlotDuration sdPenult
+        msDiff = fromIntegral (toInteger blockDuration * (toInteger (getEpochIndex desiredEpoch) - toInteger (getEpochIndex sdPenultEpoch)))
```
gromakovsky: `EpochIndex` is `Integral`, so `getEpochIndex` is redundant in two places.

infra/Pos/Slotting/Util.hs
```diff
@@ -50,15 +52,31 @@ getCurrentSlotFlat = fmap flattenSlotId <$> getCurrentSlot
...
+         Just $ slotTimestamp siSlot $ extrapolateSlottingData siEpoch
        | siEpoch == sdPenultEpoch -> Just $ slotTimestamp siSlot sdPenult
        | siEpoch == sdPenultEpoch + 1 -> Just $ slotTimestamp siSlot sdLast
        | otherwise -> Nothing
   where
+    -- Extrapolate slotting data for arbitrary epochs
+    extrapolateSlottingData desiredEpoch =
+      let
+        -- Assuming these durations stay constant
+        blockDuration = diffTimestamp (esdStart sdLast) (esdStart sdPenult)
```
gromakovsky: Why is it called `blockDuration`? It is `epochDuration`, isn't it?

gromakovsky:
@akegalj to be more precise: I am okay with this solution as a workaround and I think it's fine to merge it if we need it soon. But long-term solution is what I described.

core/Pos/Core/Timestamp.hs
ajnsit: Ah! I didn't realise that it was even possible to add two timestamps. Not sure how much sense it makes though. What would the unit of the result be? It's probably clearer and more future proof to do it this way.

ajnsit:
@gromakovsky I incorporated your review comments.

core/Pos/Core/Timestamp.hs
gromakovsky: Ok, makes sense.

PR #1049, opened 2017-06-28
Merge DEVOPS-74 to master

gromakovsky:


src/Pos/Statistics/Statsd.hs
```diff
@@ -0,0 +1,72 @@
+{-# LANGUAGE ApplicativeDo #-}
+
+module Pos.Statistics.Statsd
```
rockbmb: Not critical, but a comment explaining what "Statsd" means would make things clealer.

PR #1050, opened 2017-06-28
[CSM-320] Use accounts and addresses based on hardened keys

Martoon-00:
This PR makes, when it comes to generation of new accounts and addresses, generate **hardened** child key for that account / address (see CSLREQ-12 for more info). It still allows to import arbitrary accounts/addresses from already existing keyfiles, including ones based on non-hardened keys, i.e. current wallet version is compatible with previously created keys.

It was tested with postman, new keys work (genesis `tar` is uploaded here: https://drive.google.com/a/iohk.io/file/d/0B5lxtpMYMfwKZ25ucDhUTi1mVG8/view?usp=sharing).

WARNING: Once merged, cluster should be relaunched


flyingleafe:
What's up with this PR? It seems like it shouldn't go into `0.5` version, but straight to `master` instead -- we aren't going to relaunch 0.5 testnet, I suppose.

Martoon-00:
Yep, I'm going to make another PR, to master

PR #1051, opened 2017-06-28
[CSL-555] Use 'onException' in 'applyWithRollback'

neongreen:


src/Pos/Block/Logic/VAR.hs
```diff
@@ -13,13 +13,14 @@ module Pos.Block.Logic.VAR
 
 import           Universum
 
+import           Control.Exception.Lifted (onException)
```
gromakovsky: Why not `MonadMask` version?

gromakovsky: (I know about `ExceptT`, but there you use `lift`)

src/Pos/Block/Logic/VAR.hs
```diff
@@ -223,22 +223,47 @@ applyWithRollback
...
+    -- thrown). In pseudocode, something like this:
+    --
+    --     rollbackBlocksUnsafe `finally` do
+    --         newTip <- GS.getTip
+    --         when (newTip /= expectedTipApply) $ do
+    --             applyBack
+    --             throwError ...
+    --
+    -- However, we can't use 'throwError' in 'finally' because it won't work
+    -- (Either Text is not an instance of MonadMask), so we have to do it the
```
gromakovsky: Not `Either Text`, but `ExceptT …`.

src/Pos/Block/Logic/VAR.hs
```diff
@@ -223,22 +223,47 @@ applyWithRollback
...
+    --     newTip==tipAfterRollback, 'throwError' will happen too
+    tipAfterRollback <- lift $
+        (rollbackBlocksUnsafe toRollback >> GS.getTip) `finally` do
+            newTip <- GS.getTip
+            when (newTip /= expectedTipApply) applyBack
+    when (tipAfterRollback /= expectedTipApply) $
+        throwError (tipMismatchMsg "apply in 'apply with rollback'"
+                                   tip newestToRollback)
+
     lift (verifyAndApplyBlocks True toApply) >>= \case
```
gromakovsky: What if this `verifyAndApplyBlocks` throws exception?

db/Pos/DB/GState/Common.hs
```diff
@@ -143,8 +164,21 @@ getTipMaybe = gsGetBi tipKey
...
 putTip = gsPutBi tipKey
 
 putBot :: MonadDB m => HeaderHash -> m ()
 putBot = gsPutBi botKey
+
+putMaxSeenDifficulty :: MonadDB m => ChainDifficulty -> m ()
+putMaxSeenDifficulty = gsPutBi maxSeenDifficultyKey
+
+-- | Put difficulty only if it's bigger than existing one.
+updateMaxSeenDifficulty :: MonadDB m => ChainDifficulty -> m ()
```
gromakovsky: I think this function should be removed. It's not used and if it was used it would most likely be a bug. That's because we should update GState only in batches. Instead of using `putMaxSeenDifficulty` you can return `m (Maybe CommonOp`) (or `m [CommonOp]`).

gromakovsky:
@neongreen can you merge/rebase? There are multiple conflicts and I had a feeling that I was looking at old code when I was reviewing it. After that I will review again.

neongreen:
> Required statuses must pass before merging

I blame macOS.

gromakovsky:
Did you report it to DEVOPS-204?

neongreen:
I think I did. Not sure.

PR #1052, opened 2017-06-29
[CSL-1295] Request missing headers in a worker

int-index:
To be honest, I couldn't reproduce the issue in the exact way it was reported. In particular, I didn't get the "Header c3f52eb6 potentially represents good alternative chain, requesting more headers" message at all, and "Action Recv from NodeId 34.249.78.18:3000:0 in conversation took more than 4s" were not grouped.

Furthermore, I see errors like this in the log: "Error in conversation to NodeId 192.169.1.38:3001:0: NoParse".

Anyway, I don't know whether this is something wrong with the code or my testing setup. I've implemented the changes as discussed with @georgeee, but I'm hoping someone else could verify that they help (and don't break anything).

### DIFF
--- a/src/Pos/Block/Network/Listeners.hs
+++ b/src/Pos/Block/Network/Listeners.hs
@@ -78,8 +78,8 @@ handleBlockHeaders
     :: forall ssc m.
        (SscWorkersClass ssc, WorkMode ssc m)
     => (ListenerSpec m, OutSpecs)
-handleBlockHeaders = listenerConv $ \__ourVerInfo nodeId conv -> do
+handleBlockHeaders = listenerConv @Void $ \__ourVerInfo nodeId conv -> do
     logDebug "handleBlockHeaders: got some unsolicited block header(s)"
     mHeaders <- recvLimited conv
     whenJust mHeaders $ \(MsgHeaders headers) ->
-        handleUnsolicitedHeaders (getNewestFirst headers) nodeId conv
+        handleUnsolicitedHeaders (getNewestFirst headers) nodeId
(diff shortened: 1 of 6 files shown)


gromakovsky:
I am transferring review request from me to Misha as he is more familiar with the blocks' exchange machinery.

gromakovsky:
Also keep in mind https://issues.serokell.io/issue/CSL-1295#comment=96-5020

src/Pos/Block/Network/Listeners.hs
```diff
@@ -78,8 +78,8 @@ handleBlockHeaders
     :: forall ssc m.
        (SscWorkersClass ssc, WorkMode ssc m)
     => (ListenerSpec m, OutSpecs)
-handleBlockHeaders = listenerConv $ \__ourVerInfo nodeId conv -> do
+handleBlockHeaders = listenerConv @Void $ \__ourVerInfo nodeId conv -> do
```
georgeee: This change is dangerous. You should ensure other side of conversation also expects Void

georgeee: I would recommend to leave MshHeaders here. As far as I see bug is present

int-index: `Void` means that no messages will be actually send (there's no way to create a value of `Void`). It doesn't matter what the other side expects because it won't receive anything.

src/Pos/Block/Network/Retrieval.hs
```diff
@@ -109,19 +110,29 @@ retrievalWorkerImpl sendActions =
...
+                (Just (nodeId, task), _) ->
+                    case task of
+                        RetrieveBlocksByHeaders headers ->
+                            pure (handleBlockRetrieval nodeId headers)
+                        RetrieveHeadersByTip tip ->
+                            pure $ do
+                                mghM <- mkHeadersRequest (Just (headerHash tip))
+                                whenJust mghM $ \mgh ->
+                                    withConnectionTo sendActions nodeId $ \_ -> pure $ Conversation $
+                                        requestHeaders mgh nodeId (Just tip)
```
georgeee: How I understand this logic, requestHeaders will add another message to queue to download blocks. Which is wrong, we should download block right away

PR #1053, opened 2017-06-29
[CSM-336] Do not traverse blockchain on `newWallet`

Martoon-00:
Do it only on wallet restore.

Tested with daedalus and postman. 
Wallet restoration still takes 20-60 sec


akegalj:
@Martoon-00 we still have to rewire restoreWallet and newWallet to proper endpoints in client-api (the bridge)

Martoon-00:
That's true :man_facepalming:

Martoon-00:
@akegalj Wait a sec, that's not true. In bridge they are also different, with `TODO` to unite them one day

akegalj:
@Martoon-00 yes, I just checked. You are right. Approving

PR #1055, opened 2017-06-29
[CSL-1290] Add BlockVersionModifier

gromakovsky:
Previously 'BlockVersionData' was misused. It represented two different
things:
• description of the currently used block version;
• modification of the currently used block version proposed in update.

Eveyrthing was fine as long as we wanted to represent modification as the
whole 'BlockVersionData'. But recently we added tx fee policy which can be
potentially big and decided to put it into update proposal only if it differs
from the currently adopted one. It lead to necessity to separate
'BlockVersionData' and 'BlockVersionModifier' which was done in this PR.

Also now we can easily make other proposed fields optional.

Dear reviewers, how do you think, should I make other fields optional? I think it's a good idea, do you agree?


update/Pos/Update/Core/BlockVersion.hs
```diff
@@ -0,0 +1,30 @@
...
+import           Universum
+
+import           Pos.Core              (BlockVersionData (..))
+import           Pos.Update.Core.Types (BlockVersionModifier (..))
+
+-- | Apply 'BlockVersionModifier' to 'BlockVersionData'.
+applyBVM :: BlockVersionModifier -> BlockVersionData -> BlockVersionData
+applyBVM BlockVersionModifier {..} BlockVersionData {..} =
+    BlockVersionData
+    { bvdScriptVersion     =                        bvmScriptVersion
```
int-index: What kind of alignment is this? :smile:

gromakovsky:
> I don't like code duplication caused by this approach. BlockVersionModifier is a type of patches for BlockVersionData, but it's basically copy-pasted. This is evident in bvmMatchesBVD and applyBVM.
> It would be nice to figure out how to derive bvmMatchesBVD and applyBVM automatically.

I don't like it too, but I have no idea how to make it better. Well, currently I could move everything except tx fee policy into a separate type, but if we want to make other fields optional too (I think it makes sense), then all fields will be different.

PR #1057, opened 2017-06-29
[CSM-338] Sort wallets on order of creation on fetch from db

Martoon-00:
Make `getWallets` endpoint sort wallets by creation order (oldest first)

### DIFF
--- a/src/Pos/Wallet/Web/State/Storage.hs
+++ b/src/Pos/Wallet/Web/State/Storage.hs
@@ -59,6 +59,7 @@ import           Control.Monad.State.Class  (put)
 import           Data.Default               (Default, def)
 import qualified Data.HashMap.Strict        as HM
 import           Data.SafeCopy              (base, deriveSafeCopySimple)
+import           Data.Time.Clock.POSIX      (POSIXTime)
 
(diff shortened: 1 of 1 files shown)


src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -81,6 +81,7 @@ type CustomAddresses = HashMap (CId Addr) (HeaderHash)
 
 data WalletInfo = WalletInfo
     { _wiMeta         :: CWalletMeta
+    , _wiId           :: Int           -- ^ number of wallet in creation order
```
Martoon-00: It's a bit more complex and no one needs them for now )

ajnsit: I'll take your word for it :) It's strange though, looking at the code, it seems that we are already passing the current time to `createWallet`. I would think it could just be stored as-is, or if that param has a different meaning, we can pass another one.

Martoon-00: You're right, I didn't notice it :man_facepalming:

akegalj: @Martoon-00 will you slightly modify it based on ^ or should we merge ?

PR #1058, opened 2017-06-29
[CSM-321] Update transaction creation logic in wallet to include enough fees

pva701:
* Generate a transaction considering of fees


src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -711,6 +732,60 @@ prepareTxInfo passphrase moneySource dstDistr = do
     listF separator formatter =
         F.later $ fold . intersperse separator . fmap (F.bprint formatter)
 
+    unknownFeePolicy =
+        InternalError . sformat ("Unknown Fee Policy, tag: "%build)
+
+-- Accept all addresses in decrease order (by coins)
```
volhovM: Haddock format.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -649,44 +645,69 @@ prepareTxInfo
...
-                   selectSrcAccounts reqCoins accs
-               | balance < reqCoins ->
-                   do let remCoins = reqCoins `unsafeSubCoin` balance
-                      ((acc, balance) :|) . toList <<$>>
-                          selectSrcAccounts remCoins accs
-               | otherwise ->
-                   return
-                       (balance `unsafeSubCoin` reqCoins, (acc, reqCoins) :| [])
+        => TxSizeLinear
+        -> [(CWAddressMeta, Coin)] -- All inputs, immputable
```
volhovM: Typo.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -649,44 +645,69 @@ prepareTxInfo
...
     allAddrs <- getMoneySourceAddresses moneySource
     let dstAccAddrsSet = S.fromList $ map fst $ toList dstDistr
-        notDstAccounts = filter (\a -> not $ cwamId a `S.member` dstAccAddrsSet) allAddrs
+        notDstAddrs = filter (\a -> not $ cwamId a `S.member` dstAccAddrsSet) allAddrs
         coins = foldr1 unsafeAddCoin $ snd <$> dstDistr
-    distr@(remaining, spendings) <- selectSrcAccounts coins notDstAccounts
-    logDebug $ buildDistribution distr
-    mRemTx <- mkRemainingTxOut remaining
+    balances <- mapM getWAddressBalance notDstAddrs
+    -- Sort in decrease order
```
volhovM: Motivation?

gromakovsky:
1. Do we discard light wallet (CLI I mean)?
2. We definitely need tests for transaction formation, in particular test that transaction formation code creates a transaction which is valid w. r. t. fee policy. Probably it's not for this task, because code is the mess, but i just wanted to draw your attention.

src/Pos/Client/Txp/Util.hs
```diff
@@ -120,18 +119,20 @@ prepareInpsOuts
...
-            | otherwise = outputs
+    -- @pva701: @flyingleaf, we added fees and this code became invalid,
+    -- I just commented it. Remove these comments if it's ok or rewrite otherwise.
+    -- let inputSum =
+    --         unsafeIntegerToCoin $ sumCoins $ map (txOutValue . toaOut . snd) futxo
+        -- newOuts
+        --     | inputSum > totalMoney =
+        --         TxOutAux (TxOut someAddr (inputSum `unsafeSubCoin` totalMoney)) [] <|
+        --         outputs
+        --     | otherwise = outputs
```
flyingleafe: Ahem... elaborate on that, please. Maybe it's just me being stupid, but we have to have some change output no matter if we have fees or not. Where's that logic is handled now instead?

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -649,44 +645,69 @@ prepareTxInfo
...
+    let addrWBal = reverse $ sortWith snd $ zip notDstAddrs balances
```
flyingleafe: Also if you sort, you'd better sort it with `Down` instead of reversing.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -711,6 +732,60 @@ prepareTxInfo passphrase moneySource dstDistr = do
...
+    -> NonEmpty (EncryptedSecretKey)
+    -> (NonEmpty SafeSigner -> m a) -> m a
+withSafeSigners passphrase (sk :| sks) action =
+    withSafeSigner sk (pure passphrase) $ \mss -> do
+        ss <- maybeThrow (RequestError "Passphrase doesn't match") mss
+        case nonEmpty sks of
+            Nothing -> action (ss :| [])
+            Just sks' -> do
+                let action' = action . (ss :|) . toList
+                withSafeSigners passphrase sks' action'
```
flyingleafe: Why is such stuff in `Methods`, and not somewhere like `Pos.Client.Txp.Util`?

pva701: it's monadic

db/Pos/DB/Class.hs
```diff
@@ -185,9 +185,11 @@ gsUnlockStakeEpoch = bvdUnlockStakeEpoch <$> gsAdoptedBVData
 
 -- | Checks if provided epoch is in the bootstrap era.
 gsIsBootstrapEra :: MonadGState m => EpochIndex -> m Bool
-gsIsBootstrapEra epoch = do
-    unlockStakeEpoch <- gsUnlockStakeEpoch
-    return $ isBootstrapEra unlockStakeEpoch epoch
+gsIsBootstrapEra epoch =
+    if isDevelopment then pure False
```
int-index: This check doesn't belong here, perhaps move it to `isBootrstrapEra` itself.

gromakovsky: Or better, remove it altogether 😉

pva701: this code from CSM-322 and CSM-359, I propose to do it @flyingleafe 
I want to merge this PR ASAP

src/Pos/Wallet/Web/Server/Methods.hs
gromakovsky: So…?

pva701: the main reason not to do it: this function depends on `WalletError` which is in `Pos.Wallet.Web.Error.Types` module, so we would get cyclic deps

PR #1059, opened 2017-06-30
[CSL-1305] Allow to use different genesis.bin depending on DCONFIG

georgeee:
* Allow to specify genesis.bin as constant
* Get rid off aggregates: prod, wallet


core/constants.yaml
```diff
@@ -121,31 +123,38 @@ qanet: &qanet
 
   recoveryHeadersMessage: 100
 
+  genesisBinSuffix: qa
+
 testnet: &testnet
   <<: *prod_base
 
-  k: 60
+  k: 72
```
georgeee: This change is not related to this PR, it just adjusts testnet `k` to be such that epoch is one day (this is used in modern testnet staging)

core/Pos/Core/Constants.hs
```diff
@@ -75,6 +77,8 @@ data CoreConstants = CoreConstants
       -- | Size of mem pool will be limited by this value muliplied by block
       -- size limit.
     , ccMemPoolLimitRatio          :: !Word
+      -- | Suffix for genesis.bin files
+    , ccGenesisBinSuffix           :: ![Char]
```
neongreen: Why not `String`?

genesis-info/tn.log
```diff
@@ -0,0 +1,358 @@
...
+                                 Dload  Upload   Total   Spent    Left  Speed
+
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 2655k    0 2655k    0     0  4022k      0 --:--:-- --:--:-- --:--:-- 4023k
100 9150k    0 9150k    0     0  5508k      0 --:--:--  0:00:01 --:--:-- 5508k
100 14.8M    0 14.8M    0     0  5712k      0 --:--:--  0:00:02 --:--:-- 5711k
100 20.8M    0 20.8M    0     0  5822k      0 --:--:--  0:00:03 --:--:-- 5822k
100 22.0M    0 22.0M    0     0  5827k      0 --:--:--  0:00:03 --:--:-- 5826k
+
+Running command: /nix/store/1yhsgimnbfiyqssgmcfcg67s8if8hvg0-cardano-sl-0.4.4/bin/cardano-keygen -f /home/ielectric/dev/serokell/daedalus/cardano-sl/scripts/generate/../../secrets/secret-{}.key -m 14 -n 12000 --richmen-share 0.94 --testnet-stake 19072918462000000 --utxo-file /home/ielectric/dev/serokell/daedalus/cardano-sl/scripts/generate/../avvm-files/utxo-dump-last-new.json --randcerts --blacklisted /home/ielectric/dev/serokell/daedalus/cardano-sl/scripts/generate/../avvm-files/full_blacklist.js --fake-avvm-seed-pattern avvm/fake-{}.seed --fake-avvm-entries 100
```
gromakovsky: Why do we need anything before this line in this file in the repository?

core/constants.yaml
```diff
@@ -121,31 +123,38 @@ qanet: &qanet
...
+testnet_wallet: &testnet_wallet
+  <<: *testnet
+  applicationName: csl-daedalus
 
+testnet_staging_wallet: &testnet_staging_wallet
+  <<: *testnet_staging
+  applicationName: csl-daedalus
+
+qanet_wallet: &qanet_wallet
+  <<: *qanet
```
gromakovsky: I must confess that I don't understand anything in this file. Hope it's ok.

genesis-info/tn.log
georgeee: It's being logged automatically, we didn't want to write complex scripts for handling that (i.e. that will log accurately). This file is for double-checking mostly

core/Pos/Core/Constants.hs
georgeee: B/c it's not present in `universum`?

gromakovsky: Are you from 2016? 😱

PR #1063, opened 2017-06-30
[DEVOPS-100] Automate redemption certificate generation

domenkozar:
- launchable by Nix to provide correct dependencies
- automate postvend-app building (should probably be moved to the post-vend-app so we would just call `make`)
- other smaller changes for readability
- shellcheck only complains about unknown shebang

### DIFF
--- a/scripts/generate/certificates.sh
+++ b/scripts/generate/certificates.sh
@@ -1,34 +1,53 @@
-#!/bin/bash
+#!/usr/bin/env nix-shell
+#! nix-shell -j 4 -i bash -p nodejs-7_x stack
 
-set -e
+# NOTE: this file must be kept idempotent
+
+
+set -ex
 
 genesis=$1
 
 if [[ ! -d "$genesis" ]]; then
-  echo "No genesis"
+  echo "No such folder $genesis"
   exit 1
 fi
 
-genesis_=`cd $genesis && pwd`
+pushd "$genesis"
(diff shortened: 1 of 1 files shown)


scripts/generate/certificates.sh
```diff
@@ -1,32 +1,54 @@
...
+          npm install
+        popd
+      done
+      mkdir -p bin
+      stack build --nix --local-bin-path bin --copy-bins
+      echo "done" > build-done
+    fi
+
+    for i in ../avvm/*.seed; do
+      rm seeds.txt
```
georgeee: Nonsense

georgeee: Move it upwards

scripts/generate/certificates.sh
```diff
@@ -1,32 +1,54 @@
...
+  
+    mkdir ../certs
+  
+    mv paper-certs-mnem-* ../certs/paper-certs-mnem
+    mv paper-certs-* ../certs/paper-certs
+    mv redeem-certs-* ../certs/redeem-certs
+  popd
+popd
+
+tar -czf "$genesis-certs-secrets.tgz" "$genesis/certs" secrets
```
georgeee: I guess here should be `$genesis/secrets`

domenkozar:
@georgeee can we merge?

domenkozar:
@georgeee ping :)

PR #1065, opened 2017-06-30
Tests alexrb csl1162 binary

rockbmb:


rockbmb:
@gromakovsky me, @int-index and @neongreen had a discussion on the testing channel late Sunday about automating these tests, and some preliminary ideas are that it won't be worth the effort compared to just being disciplined and add the types as we go. I'll link the chat to you on Slack.

PR #1070, opened 2017-06-30
[CSL-1311] Enable mainnet constants in `master`

georgeee:


test/Test/Pos/Types/BlockSpec.hs
```diff
@@ -40,7 +40,7 @@ spec = describe "Block properties" $ do
     describe "verifyHeader" $ do
         prop verifyHeaderDesc (validateGoodMainHeader @SscNistBeacon)
         prop verifyHeaderDesc (validateGoodMainHeader @SscGodTossing)
-    describe "verifyHeaders" $ do
+    describe "verifyHeaders" $ modifyMaxSuccess (const 1) $ do
```
gromakovsky: Why?

georgeee: B/c this test requires verifying headers of blockchain of length > 4000 blocks. It's not fast and with `100` retries takes really long
And actually there is little sense to perform tests multiple time for reason checks are structural and we generate same structure with just different keys.

rockbmb: Agreed. Alternatively one could just add `makeSmall` to the `Arbitrary` instances, but this works better-

PR #1072, opened 2017-07-01
[CSL-1298] Fix misused 'recoveryCommGuard'

gromakovsky:
Previously 'recoveryCommGuard' was totally misused in «on new slot workers».
Specifically, it was checking whether we are doing recovery only once: before
the worker is actually launched. But we want to do it for each tick of the
worker.

### DIFF
--- a/src/node/Main.hs
+++ b/src/node/Main.hs
@@ -32,7 +32,7 @@ import qualified Pos.CLI                    as CLI
 import           Pos.Communication          (ActionSpec (..), NodeId, OutSpecs,
                                              WorkerSpec, worker, wrapActionSpec)
 import           Pos.Constants              (isDevelopment)
-import           Pos.Context                (MonadNodeContext, recoveryCommGuard)
+import           Pos.Context                (MonadNodeContext)
 import           Pos.Core.Types             (Timestamp (..))
 import           Pos.DHT.Real               (KademliaDHTInstance (..),
                                              foreverRejoinNetwork)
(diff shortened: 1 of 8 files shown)


src/node/Main.hs
```diff
@@ -88,8 +88,10 @@ action peerHolder args@Args {..} transport = do
 
     let vssSK = fromJust $ npUserSecret currentParams ^. usVss
     let gtParams = gtSscParams args vssSK
+    -- TODO: this code is impossible to maintain, hence we have 'identity' here.
+    -- Fortunately it's refactored in master.
     let wDhtWorkers :: WorkMode ssc m => KademliaDHTInstance -> ([WorkerSpec m], OutSpecs)
-        wDhtWorkers = (\(ws, outs) -> (map (fst . recoveryCommGuard . (, outs)) ws, outs)) . -- TODO simplify
+        wDhtWorkers = (\(ws, outs) -> (map (fst . identity . (, outs)) ws, outs)) . -- TODO simplify
```
georgeee: `. identity`? Why?

georgeee: Maybe we just remove this part

georgeee:
This bug is so much fun, can't believe we had it (facepalm)

pva701:
>This bug is so much fun, can't believe we had it (facepalm)

@georgeee You underestimate my power :peka:

src/node/Main.hs
gromakovsky: This code is removed in master, I don't want to fuck with it in rc branch (comment says it).

PR #1080, opened 2017-07-01
[CSL-1101] Introduce BlockAmount and SlotAmount newtypes

neongreen:


gromakovsky:
Why did you use word `Amount` for blocks and slots? They are countable, but `Amount` is usually used for uncountable nouns. Am I wrong?

core/Pos/Core/Slotting.hs
```diff
@@ -15,18 +15,20 @@ import           Universum
...
-    fromIntegral siEpoch * epochSlots + fromIntegral (getSlotIndex siSlot)
+flattenSlotId SlotId {..} = fromIntegral $
+    (fromIntegral siEpoch * epochSlots :: SlotAmount) +
+    (fromIntegral (getSlotIndex siSlot) :: SlotAmount)
 
 -- | Flattens 'EpochIndex' into a single number.
 flattenEpochIndex :: EpochIndex -> FlatSlotId
-flattenEpochIndex EpochIndex {..} = getEpochIndex * epochSlots
+flattenEpochIndex (EpochIndex i) = fromIntegral $
+    ((fromIntegral i * epochSlots) :: SlotAmount)
```
gromakovsky: Why do you write `SlotAmount` in expressions which include `epochSlots`? `epochSlots` is `SlotAmount`, so everything should be deduced automatically. Do you think it makes the code more understandable? I am not sure about it, because it's quite obvious that `epochSlots` has type `SlotAmount` as long as the reader knows about existence of such type.

godtossing/Pos/Ssc/GodTossing/Core/Core.hs
```diff
@@ -105,16 +104,23 @@ mkSignedCommitment
     => SecretKey -> EpochIndex -> Commitment -> SignedCommitment
 mkSignedCommitment sk i c = (toPublic sk, c, sign SignCommitment sk (i, c))
 
+toLocalSlotIndex :: SlotAmount -> LocalSlotIndex
+toLocalSlotIndex = either error identity . mkLocalSlotIndex . fromIntegral
```
gromakovsky: FYI: there is `unsafeMkLocalSlotIndex` in `Pos.Core`.

godtossing/Pos/Ssc/GodTossing/Toss/Logic.hs
```diff
@@ -62,11 +61,16 @@ verifyAndApplyGtPayload eoh payload = do
...
-            let slot = epochOrSlot (const 0) (getSlotIndex . siSlot) eos
-            -- We can freely clear shares after 'slotSecurityParam'
-            -- because it's guaranteed that rollback on more than 'slotSecurityParam'
+            -- We can freely clear shares after 'slotSecurityParam' because
+            -- it's guaranteed that rollback on more than 'slotSecurityParam'
             -- can't happen
-            when (slot >= slotSecurityParam && slot < 2 * slotSecurityParam) resetShares
+            let mkIndex :: SlotAmount -> LocalSlotIndex
+                mkIndex = toEnum . fromIntegral
+            let slot  = epochOrSlot (const (toEnum 0)) siSlot eos
```
gromakovsky: `toEnum 0` is `minBound`, but it's subjective what is better.

godtossing/Pos/Ssc/GodTossing/Toss/Logic.hs
```diff
@@ -62,11 +61,16 @@ verifyAndApplyGtPayload eoh payload = do
...
             setEpochOrSlot eos
-            let slot = epochOrSlot (const 0) (getSlotIndex . siSlot) eos
-            -- We can freely clear shares after 'slotSecurityParam'
-            -- because it's guaranteed that rollback on more than 'slotSecurityParam'
+            -- We can freely clear shares after 'slotSecurityParam' because
+            -- it's guaranteed that rollback on more than 'slotSecurityParam'
             -- can't happen
-            when (slot >= slotSecurityParam && slot < 2 * slotSecurityParam) resetShares
+            let mkIndex :: SlotAmount -> LocalSlotIndex
+                mkIndex = toEnum . fromIntegral
```
gromakovsky: I think it would be better to do comparison in `SlotAmount`s, because `toEnum` is unsafe, while conversion from `LocalSlotsIndex` to `SlotAmount` is not.

src/Pos/Aeson/Types.hs
```diff
@@ -42,6 +43,9 @@ instance ToJSON TxOut where
         "coin"    .= coinToInteger txOutValue,
         "address" .= sformat build txOutAddress ]
 
+deriving instance ToJSON SlotAmount
+deriving instance ToJSON BlockAmount
+
 -- NOTE: some of these types are used on frontend (PureScript).
```
gromakovsky: You might have broken something in communication with frontend, but I am not an expert in this area.

src/Pos/Wallet/Redirect.hs
```diff
@@ -78,7 +78,7 @@ networkChainDifficultyWebWallet = getLastKnownHeader >>= \case
         cSlot <- flattenSlotId <$> MaybeT getCurrentSlot
         th <- lift (getTipHeader @ssc)
         let hSlot = flattenEpochOrSlot th
-        when (hSlot <= cSlot - blkSecurityParam) $
+        when (hSlot + fromIntegral blkSecurityParam <= cSlot) $
```
gromakovsky: The task was to prohibit summing number of blocks and number of slots, but that's exactly what is done here :(

test/Test/Pos/FollowTheSatoshiSpec.hs
```diff
@@ -118,8 +118,9 @@ ftsAllStake fts key ah v =
 -- | Constant specifying the number of times 'ftsReasonableStake' will be
 -- run.
 numberOfRuns :: Int
-numberOfRuns = 300000 `div` blkSecurityParam -- Higher we have `blkSecurityParam`,
-                                             -- longer epochs and more time FTS takes
+-- The higher we have `blkSecurityParam`, the longer epochs will be and the
+-- more time FTS will take
```
gromakovsky: Thanks for fixing that comment.

neongreen:
@gromakovsky you're right about `Amount`, but somehow it sounds better to me than `Number` or `Count` (in fact, the former sounds like an index and not like amount). Any ideas?

src/Pos/Aeson/Types.hs
neongreen: Probably not because `deriving` will use GND here and so the instance will be the same as for the underlying type.

src/Pos/Wallet/Redirect.hs
neongreen: Mda.

gromakovsky:
> (in fact, the former sounds like an index and not like amount)

I had similar intuition for a long time, but last year it somehow disappeared and nowadays for me _number_ may sound like an index or like an amount depending on the context.

Google gives them same rectangles http://i.imgur.com/HBHNuSH.png

Regarding _count_ vs _number_: I don't know when one should what.

gromakovsky:
I asked in #global: https://serokell.slack.com/archives/C0EAY0GLR/p1499250495321863

PR #1082, opened 2017-07-01
[CSM-340] Wallet address ordering

Pastafarianist:


src/Pos/Wallet/Web/BListener.hs
```diff
@@ -115,15 +113,8 @@ logMsg
...
-                %", deleted accounts: "%listJson
-                %", used address: "%listJson
-                %", change address: "%listJson)
-        action bNums wAddr
-        (map fst $ MM.insertions camAddresses)
-        (MM.deletions camAddresses)
-        (map (fst . fst) $ MM.insertions camUsed)
-        (map (fst . fst) $ MM.insertions camChange)
+                %", "%build)
+        action bNums wAddr accModifier
```
pva701: I can't understand where is `Buildable` from `MapModifier`. can you explain&

src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -182,18 +190,26 @@ getWalletAddresses =
...
                   -> Query (Maybe [CWAddressMeta])
-getAccountWAddresses mode accId = do
-    let fetch which = toList <<$>> preview (wsAccountInfos . ix accId . which)
-    withAccLookupMode mode (fetch aiAccounts) (fetch aiRemovedAccounts)
+getAccountWAddresses mode accId =
+    withAccLookupMode mode (fetch aiAddresses) (fetch aiRemovedAddresses)
+  where
+    fetch :: MonadReader WalletStorage m => Lens' AccountInfo CAddresses -> m (Maybe [CWAddressMeta])
+    fetch which = do
+        -- `preview` operates within a `MonadReader m` and applies the lens
```
pva701: are you sure this comment is needed? :)

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -96,6 +115,18 @@ instance Monoid CAccModifier where
     (CAccModifier a b c) `mappend` (CAccModifier a1 b1 c1) =
         CAccModifier (a <> a1) (b <> b1) (c <> c1)
 
+instance Buildable CAccModifier where
```
pva701: ok, I see buildable

src/Pos/Wallet/Web/State/Storage.hs
Pastafarianist: Not here probably, but it would be nice to have this explanation _somewhere_, or else someone inexperienced with `Lens` (like me) is going to waste a lot of time (like me) figuring out what these functions do from their types.

PR #1083, opened 2017-07-02
[CSM-332] Optimize history derivation

flyingleafe:
Got rid of cached utxos, cached the history fully and used `BListener`s to update cache on block updates and rollbacks.


pva701:
why did you checkout from `master`, not from `cardano-0.5-rc`?

src/Pos/Wallet/Web/BListener.hs
```diff
@@ -66,8 +70,21 @@ onApplyTracking blunds = do
                        trackingApplyTxs encSK allAddresses $
                        zip txs (repeat newTip)
         applyModifierToWallet wAddr newTip mapModifier
+        addrs <- either addrDecodeErr pure $
```
pva701: it's implied you to add logic to `trackingApplyTxs` and add a new field into `CAccModifier` and can use `trackingApplyTxs` everywhere, for example, in order to consider mempool

flyingleafe: ok, it would be more consistent; I'll change it after I do some tests.

akegalj:
@flyingleafe this also implements https://issues.serokell.io/issue/CSM-331 right ?

flyingleafe:
@akegalj yes, CSM-331 is also done here.

akegalj:
@flyingleafe didn't we say we will merge this into 0.5-rc , not master?

flyingleafe:
@akegalj I'm preparing another PR for 0.5-rc right now

PR #1084, opened 2017-07-02
[CSL-1191] Pure monaddb

volhovM:
* Move rocks implementation to `db/../DB.Rocks`
* Move `MonadBlocksDB` to `db/` so that all interfaces are in the same place
* Implement pure db based on `RWVar X`, where `X` is in-memory datatype holding all mappings as `Data.Map`s.


src/Pos/Block/Logic/Slog.hs
```diff
@@ -144,7 +143,7 @@ slogVerifyBlocks blocks = do
 -- | Set of constraints necessary to apply/rollback blocks in Slog.
 type SlogApplyMode ssc m =
     ( SlogMode ssc m
-    , MonadBlockDBWrite ssc m
+    , MonadBlockDBWrite (BlockHeader ssc) (Block ssc) Undo m
```
gromakovsky: It's inconvenient to write `MonadBlockDBWrite (BlockHeader ssc) (Block ssc) Undo` and it's actually the only reasonable set of type parameters. Can you define an alias?

src/Pos/DB/DB.hs
```diff
@@ -104,7 +104,7 @@ initNodeDBs systemStart = do
 -- | Safely close all databases from 'NodeDBs'.
 closeNodeDBs :: MonadIO m => NodeDBs -> m ()
 closeNodeDBs NodeDBs {..} =
```
gromakovsky: Some functions in this module are specific to rocksdb and can be easily moved to `Pos.DB.Rocks` in `cardano-sl-db` I suppose.

src/Pos/Delegation/DB.hs
```diff
@@ -67,7 +67,7 @@ isIssuerByAddressHash :: MonadDBRead m => StakeholderId -> m Bool
 isIssuerByAddressHash = fmap isJust . getPskByIssuer . Right
 
 -- | Given issuer @i@ returns @d@ such that there exists @x1..xn@ with
--- @i->x1->...->xn->d@.
+-- @i→x1→...→xn→d@.
```
gromakovsky: Why do you sometimes remove non-ascii characters and sometimes add them? 😒 
WTF?

db/Pos/DB/Rocks/Functions.hs
```diff
@@ -0,0 +1,156 @@
...
+       , MonadRealDB ctx m
+       , DBIteratorClass i
+       , Bi (IterKey i)
+       , Bi (IterValue i)
+       )
+    => DBTag
+    -> Proxy i
+    -> Source m (IterType i)
+rocksIterSource tag _ = do
+    putText $ ("Iterator source, prefix " <> show (iterKeyPrefix @i))
```
volhovM: Hmm, i guess that's the result of a bad merge.

db/Pos/DB/Rocks/Functions.hs
```diff
@@ -0,0 +1,156 @@
...
+
+-- | Write ByteString to RocksDB for given key.
+rocksPutBytes :: (MonadIO m) => ByteString -> ByteString -> DB -> m ()
+rocksPutBytes k v DB {..} = Rocks.put rocksDB rocksWriteOpts k v
+
+-- | Delete element from RocksDB for given key.
+rocksDelete :: (MonadIO m) => ByteString -> DB -> m ()
+rocksDelete k DB {..} = Rocks.delete rocksDB rocksWriteOpts k
+
+-- garbage, should be abstracted and hidden
```
volhovM: Yes, i think everything should use `MonadDBRead` interface if it's possible.

db/Pos/DB/Pure.hs
```diff
@@ -0,0 +1,127 @@
...
+dbPutPureDefault :: MonadPureDB ctx m => DBTag -> ByteString -> ByteString -> m ()
+dbPutPureDefault (tagToLens -> l) key val =
+    view (lensOf @DBPureVar) >>= flip RWV.modifyPure (l . at key .~ Just val)
+
+dbDeletePureDefault :: MonadPureDB ctx m => DBTag -> ByteString -> m ()
+dbDeletePureDefault (tagToLens -> l) key =
+    view (lensOf @DBPureVar) >>= flip RWV.modifyPure (l . at key .~ Nothing)
+
+dbWriteBatchPureDefault :: MonadPureDB ctx m => DBTag -> [Rocks.BatchOp] -> m ()
+dbWriteBatchPureDefault tag = mapM_ processOp
```
volhovM: Hm, let me think and fix.

db/Pos/DB/Rocks/Redirect.hs
```diff
@@ -0,0 +1,54 @@
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+
+-- | Redirects enable instances of some type classes.
+-- Here it is related to type classes from 'Pos.DB.Class'.
+
+module Pos.DB.Rocks.Redirect
```
volhovM: You're right, it was used by Vlad as it is. Let's move the implementation into `Functions`.

src/Pos/Block/Logic/Slog.hs
volhovM: Alright, valid concern.

src/Pos/Delegation/DB.hs
volhovM: Because
I
Can
:tada:

src/Pos/DB/DB.hs
volhovM: Good!

db/Pos/DB/Pure.hs
```diff
@@ -0,0 +1,127 @@
...
+    , _pureMiscDB        :: DBPureMap
+    , _pureBlocksStorage :: Map HeaderHash ByteString
+    }
+
+makeLenses ''DBPure
+
+instance Default DBPure where
+    def = DBPure mempty mempty mempty mempty mempty
+
+type DBPureVar = RWV.RWVar DBPure
```
gromakovsky: Why isn't `IORef` enough here? I think the best practice is: if `IORef` is enough, use `IORef`. FYI: there is `atomicModifyIORef`, etc.
Another good practice is to use more widely used primitives, rather than our own ad-hoc modules whenever possible.

gromakovsky: > RWVar is better than both.

`Ether` is better than `mtl`, but we use `mtl` :peka:

db/Pos/DB/Rocks/Functions.hs
gromakovsky: Okay, it seems to be quite easy to fix, isn't it?

db/Pos/DB/Pure.hs
volhovM: It's not the best practice. In fact, I remember Snoyman once told that `atomicModifyIORef` is less preferable than `stm` methods. But `stm` is an overkill, so i use trivial wrapper over `IORef`. Anyway, this is minor as it's simple.

db/Pos/DB/Rocks/Functions.hs
volhovM: I will check it. I think i wrote this comment because i didn't figure it out. Maybe i will this time :)

db/Pos/DB/Pure.hs
gromakovsky: My impression is that Snoyman always advised to use `IORef` unless you need something more powerful (for instance, it's used in `ResourceT` implementation). Here you don't use anything that is not provided by `IORef`. I see no reason to complicate this code and add one more point of possible failure.
`IORef` has the following advantages:
1. It makes it clearer what operations you want from the primitive.
2. It is more understandable, because everybody knows what `IORef` is, but `Pos.Util.Concurrent.RWVar` is our own module.
3. It's more efficient, because you don't need to take locks, but rather leave atomicity concerns to low levels.

volhovM: Ok.

volhovM:
@gromakovsky I've fixed what you've asked me to.

db/Pos/DB/Pure.hs
```diff
@@ -0,0 +1,130 @@
...
+dbPutPureDefault :: MonadPureDB ctx m => DBTag -> ByteString -> ByteString -> m ()
+dbPutPureDefault (tagToLens -> l) key val =
+    view (lensOf @DBPureVar) >>= atomicModifyIORefPure (l . at key .~ Just val)
+
+dbDeletePureDefault :: MonadPureDB ctx m => DBTag -> ByteString -> m ()
+dbDeletePureDefault (tagToLens -> l) key =
+    view (lensOf @DBPureVar) >>= atomicModifyIORefPure (l . at key .~ Nothing)
+
+dbWriteBatchPureDefault :: MonadPureDB ctx m => DBTag -> [Rocks.BatchOp] -> m ()
+dbWriteBatchPureDefault tag = mapM_ processOp
```
gromakovsky: Good evening!

db/Pos/DB/Pure.hs
```diff
@@ -121,10 +121,13 @@ dbDeletePureDefault (tagToLens -> l) key =
...
 
 dbWriteBatchPureDefault :: MonadPureDB ctx m => DBTag -> [Rocks.BatchOp] -> m ()
-dbWriteBatchPureDefault tag = mapM_ processOp
+dbWriteBatchPureDefault (tagToLens -> l) batchOps =
+    view (lensOf @DBPureVar) >>= atomicModifyIORefPure action
   where
-    processOp (Rocks.Put k v) = dbPutPureDefault tag k v
-    processOp (Rocks.Del k)   = dbDeletePureDefault tag k
+    -- Apply actions from left to right (right fold, but reverse application order)
+    action = foldr (\batchop acc -> acc . processOp batchop) identity batchOps
```
volhovM: @gromakovsky this order is better imo

gromakovsky:
> Apart from my comments, do we still need MonadBaseControl IO as a subclass of MonadDBRead?

But what about this?

volhovM:
@gromakovsky sadly, yes. This PR wasn't dedicated to remove it and i still have no idea how to do it smoothly.

PR #1085, opened 2017-07-03
[CSL-1204] Add UnlockStakeEpoch

int-index:


src/Pos/Lrc/Worker.hs
```diff
@@ -84,6 +84,7 @@ type LrcModeFull ssc ctx m =
     , SscWorkersClass ssc
     , SscHelpersClass ssc
     , MonadSscMem ssc ctx m
+    , MonadSlotsData m
     , MonadSlots m
```
gromakovsky: IIRC, `MonadSlotsData` is a subclass of `MonadSlots`.

txp/Pos/Txp/Logic/Local.hs
```diff
@@ -54,7 +56,8 @@ txProcessTransaction
 txProcessTransaction itw@(txId, txAux) = do
     let UnsafeTx {..} = taTx txAux
     tipDB <- GS.getTip
-    bootEra <- gsIsBootstrapEra
+    epoch <- siEpoch <$> getCurrentSlotBlocking
```
int-index: I used `getCurrentSlotBlocking` here, but I don't know whether it's okay to block here.

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -698,7 +699,8 @@ prepareTxInfoRaw moneySource dstDistr = do
         coins = foldr1 unsafeAddCoin $ snd <$> dstDistr
     distr@(remaining, spendings) <- selectSrcAccounts coins notDstAccounts
     logDebug $ buildDistribution distr
-    bootEra <- gsIsBootstrapEra
+    epoch <- siEpoch <$> getCurrentSlotBlocking
```
int-index: I used `getCurrentSlotBlocking` here, but I don't know whether it's okay to block here.

gromakovsky: Hm, good question, I don't know, it should be discussed with wallet team. I see three options:
1. We can block and submit a transaction only after we are synchronized.
2. We can use `getCurrentSlotInaccurate`, then we don't need to block, but we may create invalid transaction.
3. We can simply tell user that we are not ready and do nothing.
I don't know who wallet behaves currently, but it has same options. It can block until we are synchronized. It can use current GState, but then it can create invalid transaction (if relevant parts of utxo changed). Or it can tell user that we are not ready.
It should be discussed with wallet team.

txp/Pos/Txp/Logic/Local.hs
gromakovsky: No, it's totally not ok. If we receive multiple transactions, we will have multiple threads waiting for slotting data.
If we don't know current slot, we should reject transactions.

Btw, Misha's PR (for CSL-1205) was so big that I didn't notice that it was very bad in this part.
1. He added this logic to `txProcessTransaction`, but didn't add it to explorer version.
2. He also didn't add these checks to block verification.
3. He didn't take this logic into account in normalization. So now after normalization we can have invalid transactions in mempool :(
3. In fact, it should be in Toil, not in Logic, in this case all three problems above would be resolved.

And in fact the task is more complicated.
After thinking about it for a while, I decided that it doesn't make sense to fix it within this issue. I will make a separate one.
As a workaround please use `getCurrentSlotInaccurate`.

gromakovsky: Created https://issues.serokell.io/issue/CSL-1341

src/Pos/Wallet/Web/Server/Methods.hs
gromakovsky: http://i.imgur.com/WWnfcJO.png

So let's use blocking version unless someone else enters this discussion and tells us not to do it.

int-index:
Alright, I merged `master` into this branch and addressed @gromakovsky's comments.

PR #1089, opened 2017-07-03
[CSL-1295] Request missing headers in a worker

int-index:


volhovM:
@pva701 tested it on his machine (block recovery). Works.

PR #1090, opened 2017-07-03
Port more EKG stuff

arybczak:
Ports the following:
- 3c5f59b4037ee71ce168279c94e50223ca30175c
- 25c83d7b09935fd5194ed37bf93cf0b4c721c81c
- 9cad0c00bebba638002e8cbe59ab29356b4d8195

For the record, subsequent related patches (I think):
- 7db4653736e759f5bcd0c5be357c7e67a6736553
- da6250fc87cc5d731a99050e4efe9069c4fd6579
- 72d219b2e3f7590a92e23b52494964946b5465e0
- b549bd139c35f6ebf2658e02047317322da84f22

But these deal with JSON logging and trying to incorporate these currently leads to circular dependency (as json logging is in cardano-sl, cardano-sl depends on cardano-sl-txp, but cardano-sl-txp also wants to log).

### DIFF
--- a/src/Pos/Launcher/Runner.hs
+++ b/src/Pos/Launcher/Runner.hs
@@ -55,10 +55,10 @@ import qualified System.Metrics.Gauge            as Metrics.Gauge
 import           System.Random                   (newStdGen)
 import qualified System.Remote.Monitoring        as Monitoring
 import qualified System.Remote.Monitoring.Statsd as Monitoring
-import           System.Wlog                     (LoggerConfig (..), WithLogger,
-                                                  getLoggerName, logError, logInfo,
-                                                  productionB, releaseAllHandlers,
-                                                  setupLogging, usingLoggerName)
+import           System.Wlog                     (LoggerConfig (..), WithLogger, logError,
+                                                  logInfo, logDebug, productionB,
+                                                  releaseAllHandlers, setupLogging,
+                                                  getLoggerName, usingLoggerName)
 import           Universum                       hiding (bracket, finally)
 
 import           Pos.Binary                      ()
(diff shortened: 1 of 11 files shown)


kantp:
@arybczak The CI is failing due to `hlint`. Could you please replace `pack` with `toText` in `src/Pos/Launcher/Runner.hs`, lines 218, 219, 220, 221?

arybczak:
Fixed.

src/Pos/Launcher/Runner.hs
```diff
@@ -201,6 +202,61 @@ runRealModeDo discoveryCtx transport np@NodeParams {..} sscnp listeners outSpecs
...
+        -- arbitrarily.
+        -- FIXME take alpha from configuration/CLI, or use a better
+        -- estimator.
+        let alpha :: Double
+            alpha = 0.75
+            -- This TxpMetrics specifies what to do when waiting on the
+            -- mempool lock, when the mempool lock has been granted, and
+            -- when that lock has been released. It updates EKG metrics
+            -- and also logs each data point at debug level.
+            txpMetrics = TxpMetrics
```
gromakovsky: Why is it defined here and not somewhere in `txp`?

kantp: I guess we could move it to `cardano-sl-txp`, at the price of adding `ekg` as a dependency for that package. Would you prefer that, or can we merge as-is?

gromakovsky: I would prefer moving to `cardano-sl-txp`, `Runner` shouldn't have non-trivial logic, this logic is specific to transaction processing. Adding `ekg` dependency to `cardano-sl-txp` is fine, because it's already a dependency of `cardano-sl` and I don't think someone will ever want to build `cardano-sl-txp` without `cardano-sl`. These separation into packages is mostly to have some boundaries between logical parts of the code, rather than because small parts (smaller than `cardano-sl` package) are useful by themselves.

arybczak: Should everything ekg related be moved there or just the txpMetrics stuff?

gromakovsky: @arybczak I think only `TxpMetrics` stuff. I expect it to be a function with type `Metrics.Store -> TxpMetrics`.

arybczak:
Moved the construction of TxpMetrics to cardano-sl-txp.

PR #1091, opened 2017-07-03
[CSM-275] Dummy endpoint for transaction fees

pva701:
* Fix fee api: passphrase isn't required anymore


akegalj:
@pva701 doesn't build in prod mode. Builds in dev mode. Note that 0.5-rc builds in both modes

akegalj:
>  doesn't build in prod mode. Builds in dev mode. Note that 0.5-rc builds in both modes
@georgeee said this is because we don't have staging run with master yet

PR #1093, opened 2017-07-03
[CSM-332] Rewrite tx history - version for 0.5 release

flyingleafe:
See https://github.com/input-output-hk/cardano-sl/pull/1083

### DIFF
--- a/src/Pos/Client/Txp/History.hs
+++ b/src/Pos/Client/Txp/History.hs
@@ -21,60 +21,56 @@ module Pos.Client.Txp.History
        -- * History derivation
        , getRelatedTxsByAddrs
        , deriveAddrHistory
-       , deriveAddrHistoryPartial
+       , deriveAddrHistoryBlk
        , TxHistoryRedirect
        , runTxHistoryRedirect
        ) where
 
 import           Universum
 
-import           Control.Lens                 (makeLenses, (%=))
-import           Control.Monad.Loops          (unfoldrM)
-import           Control.Monad.Trans          (MonadTrans)
-import           Control.Monad.Trans.Control  (MonadBaseControl)
+import           Control.Lens                (makeLenses)
+import           Control.Monad.Trans         (MonadTrans)
+import           Control.Monad.Trans.Control (MonadBaseControl)
 import           Control.Monad.Trans.Identity (IdentityT (..))
-import           Control.Monad.Trans.Maybe    (MaybeT (..))
+import           Control.Monad.Trans.Maybe   (MaybeT (..))
+import           Data.DList                  (DList)
 import           Data.Coerce                  (coerce)
-import qualified Data.DList                   as DL
-import qualified Data.HashSet                 as HS
-import           Data.Tagged                  (Tagged (..))
+import qualified Data.DList                  as DL
+import           Data.Tagged                 (Tagged (..))
 import qualified Ether
-import           System.Wlog                  (WithLogger)
+import           System.Wlog                 (WithLogger)
 
-import           Pos.Block.Core               (Block, mainBlockSlot, mainBlockTxPayload)
-import           Pos.Constants                (blkSecurityParam)
-import           Pos.Context.Context          (GenesisUtxo (..))
-import           Pos.Core                     (Address, ChainDifficulty, HeaderHash,
-                                               Timestamp (..), difficultyL, prevBlockL)
-import           Pos.Crypto                   (WithHash (..), withHash)
-import           Pos.DB                       (MonadDBRead, MonadRealDB)
-import qualified Pos.DB.Block                 as DB
-import           Pos.DB.Error                 (DBError (..))
-import qualified Pos.DB.GState                as GS
-import           Pos.Slotting                 (MonadSlots, SlottingData (..),
-                                               getSlotStartPure)
-import           Pos.Ssc.Class                (SscHelpersClass)
+import           Pos.Block.Core              (Block, mainBlockTxPayload)
+import           Pos.Block.Types             (Blund)
+import           Pos.Context.Context         (GenesisUtxo (..))
+import           Pos.Core                    (Address, ChainDifficulty, HeaderHash,
+                                              difficultyL)
+import           Pos.Crypto                  (WithHash (..), withHash)
+import           Pos.DB                      (MonadDBRead, MonadGState, MonadRealDB)
+import qualified Pos.DB.Block                as DB
+import qualified Pos.DB.GState               as GS
+import           Pos.Slotting                (MonadSlots)
+import           Pos.Ssc.Class               (SscHelpersClass)
+>>>>>>> 5af063da8... Rewrite tx history to `BlockExtra` routine.
 #ifdef WITH_EXPLORER
 import           Pos.Explorer.Txp.Local       (eTxProcessTransaction)
 #else
 import           Pos.Txp                      (txProcessTransaction)
 #endif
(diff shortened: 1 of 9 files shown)


src/Pos/Client/Txp/History.hs
```diff
@@ -102,86 +90,106 @@ data TxHistoryEntry = THEntry
...
+        :: [Address] -> m (DList TxHistoryEntry)
     saveTx :: (TxId, TxAux) -> m ()
 
-    default getTxHistory
+    default getBlockHistory
         :: (SscHelpersClass ssc, MonadTrans t, MonadTxHistory m', t m' ~ m)
-        => Tagged ssc ([Address] -> Maybe (HeaderHash, Utxo) -> m TxHistoryAnswer)
-    getTxHistory = fmap lift <<$>> getTxHistory
+        => Tagged ssc ([Address] -> m (DList TxHistoryEntry))
+    getBlockHistory = (lift .) <$> getBlockHistory
```
pva701: nice

src/Pos/Client/Txp/History.hs
```diff
@@ -102,86 +90,106 @@ data TxHistoryEntry = THEntry
...
+----------------------------------------------------------------------------
+-- GenesisToil
+----------------------------------------------------------------------------
+
+-- | Identity wrapper to use genesis utxo in context as `MonadUtxoRead` instance
+-- TODO: probably should be moved elsewhere; `Pos.Txp.Toil` is not possible, because
+-- of dependency on `Pos.Context` from main package
+data GenesisToilTag
+
+type GenesisToil = Ether.TaggedTrans GenesisToilTag IdentityT
```
pva701: minor:
you can use `UtxoReaderT`, like here
https://github.com/input-output-hk/cardano-sl/blob/cardano-sl-0.5-rc/src/Pos/Wallet/Web/Tracking.hs#L267

akegalj:
I have tested it in prod mode and it seems to be working fast at least. I can't be 100% about its correctness - but it seems to work

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -665,31 +663,20 @@ sendMoney sendActions passphrase moneySource dstDistr = do
...
-                cWalId
-                taLastCachedHash
-                taCachedUtxo
-                (cached <> cachedTxs)
-
-        ctxs <- forM fullHistory $ addHistoryTx cWalId False
-        pure $ concatMap toList ctxs
+
+    blockHistory <- getHistoryCache cWalId >>= \case
+        Just hist -> pure $ DL.fromList hist
```
pva701: does this code make a lot of sense?
I don't understand when it can be called. we synchronise our cache on an import of wallet so it's strange if it is called at all. looks like a piece of old logic. 
maybe make sense logging some message here to see that something strange happened

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -665,31 +663,20 @@ sendMoney sendActions passphrase moneySource dstDistr = do
...
+        Nothing -> do
```
pva701: does this code make a lot of sense?
I don't understand when it can be called. we synchronise our cache on an import of wallet so it's strange if it is called at all. looks like a piece of old logic.
maybe make sense logging some message here to see that something strange happened

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -201,117 +212,145 @@ syncWalletWithGStateUnsafe
...
+        rollbackBlock allAddresses (b, u) =
+            trackingRollbackTxs encSK allAddresses $
+            zip3 (gbTxs b) (undoTx u) (repeat $ headerHash b)
+
+        applyBlock :: (WithLogger m1, MonadUtxoRead m1)
+                   => [CWAddressMeta] -> Blund ssc -> ToilT () m1 CAccModifier
+        applyBlock allAddresses (b, _) = trackingApplyTxs encSK allAddresses mDiff blkHeaderTs $
+                                         zip (gbTxs b) (repeat $ getBlockHeader b)
+
+        sync :: HeaderHash -> m ()
```
pva701: to be honest I sgorel: don't like when functions are in `lets`, not in `when` sections...

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -111,12 +117,14 @@ data CAccModifier = CAccModifier
...
 
 instance Monoid CAccModifier where
-    mempty = CAccModifier mempty mempty mempty mempty
-    (CAccModifier a b c d) `mappend` (CAccModifier a1 b1 c1 d1) =
-        CAccModifier (a <> a1) (b <> b1) (c <> c1) (d <> d1)
+    mempty = CAccModifier mempty mempty mempty mempty mempty mempty
+    (CAccModifier a b c d ah dh) `mappend` (CAccModifier a1 b1 c1 d1 ah1 dh1) =
+        CAccModifier (a <> a1) (b <> b1) (c <> c1) (d <> d1) (ah1 <> ah) (dh <> dh1)
 
 instance Buildable CAccModifier where
```
pva701: what about updating of `Buildable` instance?

src/Pos/Client/Txp/History.hs
```diff
@@ -198,65 +206,49 @@ type TxHistoryRedirect =
...
+
+            blockFetcher :: HeaderHash -> GenesisHistoryFetcher m (DList TxHistoryEntry)
+            blockFetcher start = GS.foldlUpWhileM fromBlund start (const $ const True)
+                (deriveAddrHistoryBlk addrs getBlockTimestamp) mempty
+
+        DB.runBlockDBRedirect . runGenesisToil . evalToilTEmpty $
+            blockFetcher bot
+
+    getLocalHistory :: [Address] -> TxHistoryRedirect m (DList TxHistoryEntry)
+    getLocalHistory addrs = runDBTxp . evalToilTEmpty $ do
```
pva701: I don't understand why this function doesn't use `txMempoolToModifier`.
IMHO it's safer and more consistent

src/Pos/Wallet/Web/Server/Methods.hs
flyingleafe: @pva701 I think there's no true overhead and it makes sure that `getHistory` call yields correct result, even if something strange has happened with cache

src/Pos/Client/Txp/History.hs
flyingleafe: Yes, but I thought it would be cleaner to not get genesis utxo as a map explicitly

src/Pos/Wallet/Web/Tracking.hs
flyingleafe: i agree (only not in `when`, but `where`), but here is the problem of getting `slottingData` inside `where`-placed definition. Passing it as an argument everywhere would've been much uglier.

src/Pos/Wallet/Web/Server/Methods.hs
pva701: @flyingleafe getting history, in this case, takes a loooooong time.
I don't mind not removing this code, but add logging msg plz

src/Pos/Wallet/Web/Tracking.hs
pva701: @flyingleafe why didn't you use `MonadSlots`?

src/Pos/Client/Txp/History.hs
flyingleafe: @pva701 agree

PR #1032, opened 2017-06-26
[CSL-1205] Post bootstrap delegation

volhovM:
* Genesis stakes are now delegated to boot addresses.
* When sending tx from wallet in boot era, txdistr is automatically set to boot addresses.
* Txs in boot era that have any address other than boot in their txdistr are banned.
(+ cosmetics)

PR #1095, opened 2017-07-03
[CSL-1181] Remove delegation from poor guys

gromakovsky:


gromakovsky:
@volhovM I've just added rollback logic. Please review my last commit carefully.

src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -341,7 +361,11 @@ dlgVerifyBlocks blocks = do
         HashSet StakeholderId ->
         Block ssc ->
         ExceptT Text (MapCede (StateT DlgVerState m)) DlgUndo
-    verifyBlock _ (Left _) = [] <$ (dvCurEpoch .= HS.empty)
+    verifyBlock _ (Left genesisBlk) = do
+        dvCurEpoch .= HS.empty
+        let blkEpoch = genesisBlk ^. epochIndexL
+        noLongerRichmen <- getNoLongerRichmen blkEpoch
+        catMaybes <$> mapM getPsk noLongerRichmen
```
volhovM: I think you also need to remove it from `Cede` context (`modPsk`). Or not. Depending on whether we implement "don't verify over genesis block" policy. If we do, you should.

src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -512,15 +536,10 @@ removeNoLongerRichmen ::
...
-    let getRichmen e =
-            toList <$>
-            lrcActionOnEpochReason e "removeNoLongerRichmen" LrcDB.getRichmenDlg
-    oldRichmen <- getRichmen (newEpoch - 1)
-    newRichmen <- getRichmen newEpoch
-    let edgeActions = map DlgEdgeDel $ oldRichmen \\ newRichmen
-    calculateTransCorrections $ HS.fromList edgeActions
+    noLongerRichmen <- getNoLongerRichmen newEpoch
+    let edgeActions = map DlgEdgeDel noLongerRichmen
+    calculateTransCorrections (HS.fromList edgeActions)
```
volhovM: This is weird. You should return edge actions with trans corrections. Edge actions modify db (1), while trans corrections affect transitive and transitive reverse (2-3).

src/Pos/Delegation/Logic/VAR.hs
gromakovsky: It seems easy to do, so I will do it even if it's not necessary (I don't want to rely on it when it's avoidable).

gromakovsky:
@volhovM I've just fixed everything.

gromakovsky:
@volhovM do you think you will finish CSL-1205 today? If yes, I won't merge this PR before #1032.

PR #1096, opened 2017-07-03
[CSL-1276] Remove 'derive' package from project

rockbmb:


gromakovsky:
@rockbmb conflict appeared.

PR #1101, opened 2017-07-04
CSL-1255 Test update system with properly formed proposal

naryl:


src/wallet/Command.hs
```diff
@@ -129,8 +141,18 @@ proposeUpdate =
     lexeme parseIntegralSafe <*>
     lexeme parseIntegralSafe <*>
     lexeme parseSoftwareVersion <*>
+    many parseProposeUpdateSystem
```
georgeee: `many1` will fit better

src/wallet/Main.hs
```diff
@@ -242,6 +238,28 @@ runCmd _ (AddKeyFromFile f) = do
...
+dummyHash :: Hash Raw
+dummyHash = unsafeHash (0 :: Integer)
+
+hashFile :: MonadIO m => Maybe FilePath -> m (Maybe (Hash Raw))
+hashFile Nothing  = pure $ Just dummyHash
+hashFile filename = runMaybeT $ do
+    filePath <- MaybeT $ pure filename
+    fileData <- liftIO $ BS.readFile filePath
+    let h = unsafeHash fileData
+    liftIO $ putText $ sformat ("Read file succesfuly, its hash: "%hashHexF) h
```
georgeee: Print `Read file <filename> successfully...`

src/wallet/Main.hs
```diff
@@ -242,6 +238,28 @@ runCmd _ (AddKeyFromFile f) = do
     mapM_ addSecretKey $ secret ^. usKeys
 runCmd _ Quit = pure ()
 
+dummyHash :: Hash Raw
+dummyHash = unsafeHash (0 :: Integer)
+
+hashFile :: MonadIO m => Maybe FilePath -> m (Maybe (Hash Raw))
+hashFile Nothing  = pure $ Just dummyHash
+hashFile filename = runMaybeT $ do
```
georgeee: I believe we should get rid of `MaybeT` here and just throw IO exception

src/wallet/Main.hs
```diff
@@ -242,6 +238,28 @@ runCmd _ (AddKeyFromFile f) = do
...
+    liftIO $ putText $ sformat ("Read file succesfuly, its hash: "%hashHexF) h
+    pure h
+
+updateDataElement :: MonadIO m => ProposeUpdateSystem -> m (SystemTag, UpdateData)
+updateDataElement ProposeUpdateSystem{..} = do
+    diffHash <- hashFile pusBinDiffPath
+    installerHash <- hashFile pusInstallerPath
+    case liftA2 (,) diffHash installerHash of
+        Nothing ->
+            error "Failed to read prop file"
```
georgeee: This should be elimnated after we through IO exception in `hashFile`

PR #1103, opened 2017-07-04
[CSM-350] Redeem endpoint reports tx zero amount

Pastafarianist:


akegalj:
@Pastafarianist what is a result of this PR. In https://issues.serokell.io/issue/CSM-350 you said:
> Submitting a workaround for now.

What is idea behind the workaround ?

Pastafarianist:
@akegalj I suspect that the root cause of CSM-330 is CSM-343. If that's true, after we fix CSM-343 we can roll back CSM-350 and CSM-330. I'm waiting until @Martoon-00 gets back from vacation to have a look at that.

For now, my workaround is to handle redemption txs as a special case: I pass around a flag called `isRedemptionTx` (which used to be called `forceIncoming` earlier when I thought that it could be used only for CSM-330). This flag disables filtering of output addresses in `mkCTxs`.  This filtering is used to not report A→A txs twice. The code is fairly self-descriptive, and I've put a fair bit of comments there. You could check it out as well.

flyingleafe:
@Pastafarianist this won't fix current underlying problem which prevents redeem from happening at all. See https://issues.serokell.io/issue/CSM-351

Pastafarianist:
@flyingleafe these issues are unrelated. Here I'm fixing the fact that the `ctAmount` field in the `redeem` endpoint output is always zero. It is zero regardless of whether redeeming works or not.

akegalj:
Testing with @flyingleafe fix on top https://github.com/input-output-hk/cardano-sl/pull/1104 ...

PR #1104, opened 2017-07-04
[CSM-351] Use DB utxo for redeem addresses 

flyingleafe:
Cached filtered utxo doesn't contain outputs for redeem addresses, thus redeeming doesn't work.

### DIFF
--- a/src/Pos/Client/Txp/Balances.hs
+++ b/src/Pos/Client/Txp/Balances.hs
@@ -22,7 +22,7 @@ import           System.Wlog                  (WithLogger, logWarning)
 
 import           Pos.Core                     (AddressIgnoringAttributes (AddressIA))
 import           Pos.Crypto                   (WithHash (..), shortHashF)
-import           Pos.DB                       (MonadDBRead)
+import           Pos.DB                       (MonadDBRead, MonadRealDB)
 import qualified Pos.DB.GState.Balances       as GS
 import           Pos.Txp                      (GenericToilModifier (..), MonadTxpMem,
                                                TxAux (..), TxOutAux (..), Utxo,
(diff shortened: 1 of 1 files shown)


flyingleafe:
It works - tested on Daedalus for Mac.

src/Pos/Client/Txp/Balances.hs
```diff
@@ -68,17 +69,28 @@ runBalancesRedirect = coerce
...
     , WebWalletModeDB m
-    ,  MonadMask m
+    , MonadMask m
     , WithLogger m
     , MonadTxpMem ext m, t ~ IdentityT) =>
         MonadBalances (Ether.TaggedTrans BalancesRedirectTag t m)
   where
-    getOwnUtxos addr = do
+    getOwnUtxos addrs = do
+        let isRedeem (RedeemAddress _) = True
```
Pastafarianist: Can you move these to a `where` clause?

flyingleafe: It wouldn't make it much cleaner and actually the opposite - it would take more time for eyes to find `isRedeem`. Not worth restarting CI :)

src/Pos/Client/Txp/Balances.hs
```diff
@@ -68,17 +69,28 @@ runBalancesRedirect = coerce
...
+            redeemAddrs = filter isRedeem addrs
+            commonAddrs = filter (not . isRedeem) addrs
+
         updates <- getUtxoModifier
-        locUtxo <- MM.modifyMap updates <$> WS.getWalletUtxo
-        let addrsSet = HS.fromList $ AddressIA <$> addr
-        pure $ M.filter (`addrBelongsToSet` addrsSet) locUtxo
+        commonUtxo <- if null commonAddrs then pure mempty
+                      else WS.getWalletUtxo
+        redeemUtxo <- if null redeemAddrs then pure mempty
```
pva701: you can use `genesisUtxoM` to avoid iterating over whole Utxo, right?
but I think we can do it in the next PR
@flyingleafe

flyingleafe: @pva701 no, I can't. Redeem address can be already used.

PR #1105, opened 2017-07-04
[CSL-1307] Don't take block lock during wallet sync

pva701:


src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -192,8 +200,40 @@ syncWalletsWithGState
...
-    tip <$ mapM_ (syncWalletWithGStateUnsafe @ssc) encSKs
+syncWalletsWithGState encSKs = forM_ encSKs $ \encSK -> do
+    let wAddr = encToCId encSK
+    whenJustM (WS.getWalletSyncTip wAddr) $ \wTip ->
+        if wTip == genesisHash then syncDo encSK Nothing
+        else DB.blkGetHeader wTip >>= \case
+            Nothing ->
+                logWarning $
+                sformat ("Couldn't get block header of wallet "%build
+                            %" by last synced hh: "%build) wAddr wTip
```
flyingleafe: Such error seems somewhat fatal. Isn't it better to actually throw an error here?

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -192,8 +200,40 @@ syncWalletsWithGState
...
+            -- so we can sync wallet and GState without the block lock
+            -- to avoid blocking of blocks verification/application.
+            bh <- unsafeLast . getNewestFirst <$> DB.loadHeadersByDepth (blkSecurityParam + 1) (headerHash gstateTipH)
+            logDebug $
+                sformat ("Wallet's tip is far from GState tip. Syncing with "%build%" without the block lock")
+                (headerHash bh)
+            syncWalletWithGStateUnsafe encSK wTipH bh
+        withBlkSemaphore_ $ \tip -> do
+            tipH <- maybe (error "Wallet tracking: no block header corresponding to tip") pure =<< DB.blkGetHeader tip
+            tip <$ syncWalletWithGStateUnsafe encSK wTipH tipH
```
flyingleafe: It seems like here you should start from `bh`, not from `wTipH`, in case when previous section of code is executed. Fix, or explain why I'm wrong.

PR #1106, opened 2017-07-04
[CSL-1331] No input addresses in tx

flyingleafe:
This error is most likely emerged because transactions from mempool weren't topsorted before history derivation. This is fixed here


flyingleafe:
Tested with Daedalus on Mac OS - 3 different transactions, quickly sent over 15 seconds one after another, are displayed correctly in history.

pva701:
@flyingleaf btw, you should set correct tag for commit message (otherwise gromak will diss you as fuck)

gromakovsky:
🔪

PR #1107, opened 2017-07-05
[CSM-337] [WIP] Extrapolate timestamps in case of slot duration change

ajnsit:
This is WIP because I still need to test it more, and clean up warnings.

### DIFF
--- a/infra/Pos/Slotting/Class.hs
+++ b/infra/Pos/Slotting/Class.hs
@@ -22,13 +22,6 @@ class MonadSlotsData m => MonadSlots m where
     -- return until current slot is known.
     getCurrentSlotBlocking :: m SlotId
 
-    -- | This function tries to predict current slot as accurately as it can.
-    -- If 'getCurrentTime' returns unreliable time,
-    -- then function returns last known slot
-    -- If our slotting data into DB is outdated,
-    -- then function tries to extrapolate slot using last know slotting data
-    getCurrentSlotInaccurate :: m SlotId
-
     currentTimeSlotting :: m Timestamp
 
     slottingWorkers :: [m ()]
@@ -48,10 +41,6 @@ class MonadSlotsData m => MonadSlots m where
     default slottingWorkers :: (MonadTrans t, MonadSlots m', t m' ~ m) =>
         [m ()]
     slottingWorkers = map lift slottingWorkers
-
-    default getCurrentSlotInaccurate :: (MonadTrans t, MonadSlots m', t m' ~ m) =>
-        m SlotId
-    getCurrentSlotInaccurate = lift getCurrentSlotInaccurate
 
 instance {-# OVERLAPPABLE #-}
     (MonadSlots m, MonadTrans t, Monad (t m)) =>
--- a/infra/Pos/Slotting/Class.hs
+++ b/infra/Pos/Slotting/Class.hs
@@ -22,6 +22,13 @@ class MonadSlotsData m => MonadSlots m where
     -- return until current slot is known.
     getCurrentSlotBlocking :: m SlotId
 
+    -- | This function tries to predict current slot as accurately as it can.   
+    -- If 'getCurrentTime' returns unreliable time,   
+    -- then function returns last known slot    
+    -- If our slotting data into DB is outdated,    
+    -- then function tries to extrapolate slot using last know slotting data    
+    getCurrentSlotInaccurate :: m SlotId
+
     currentTimeSlotting :: m Timestamp
 
     slottingWorkers :: [m ()]
@@ -33,6 +40,10 @@ class MonadSlotsData m => MonadSlots m where
     default getCurrentSlotBlocking :: (MonadTrans t, MonadSlots m', t m' ~ m) =>
         m SlotId
     getCurrentSlotBlocking = lift getCurrentSlotBlocking
+
+    default getCurrentSlotInaccurate :: (MonadTrans t, MonadSlots m', t m' ~ m) =>   
+        m SlotId    
+    getCurrentSlotInaccurate = lift getCurrentSlotInaccurate    
 
     default currentTimeSlotting :: (MonadTrans t, MonadSlots m', t m' ~ m) =>
         m Timestamp
--- a/infra/Pos/Slotting/MemState/Holder.hs
+++ b/infra/Pos/Slotting/MemState/Holder.hs
@@ -21,26 +21,30 @@ import qualified Ether
 import           Pos.Core.Types               (Timestamp)
 
 import           Pos.Slotting.MemState.Class  (MonadSlotsData (..))
-import           Pos.Slotting.Types           (SlottingData, addEpochSlottingData,
-                                               getPenultEpochIndex, lookupEpochSlottingData, getLastEpochIndex)
+import           Pos.Slotting.Types           (SlottingData,
+                                               addEpochSlottingData,
+                                               getLastEpochIndex,
+                                               getPenultEpochIndex,
+                                               lookupEpochSlottingData)
 
 ----------------------------------------------------------------------------
 -- Transformer
 ----------------------------------------------------------------------------
 
 -- | System start and slotting data
-type SlottingVar = (Timestamp, TVar SlottingData)
+type SystemStart = Timestamp
+type SlottingVar = (SystemStart, TVar SlottingData)
 
 type MonadSlotting = Ether.MonadReader' SlottingVar
 
 askSlotting :: MonadSlotting m => m SlottingVar
 askSlotting = Ether.ask'
 
-askSlottingVar :: MonadSlotting m => m (TVar SlottingData)
-askSlottingVar = snd <$> askSlotting
-
 askSlottingTimestamp :: MonadSlotting m => m Timestamp
 askSlottingTimestamp  = fst <$> askSlotting
+
+askSlottingVar :: MonadSlotting m => m (TVar SlottingData)
+askSlottingVar = snd <$> askSlotting
 
 ----------------------------------------------------------------------------
 -- MonadSlotsData implementation
@@ -58,20 +62,25 @@ instance
     (MonadSlotting m, MonadIO m, t ~ IdentityT) =>
          MonadSlotsData (Ether.TaggedTrans SlotsDataRedirectTag t m)
   where
+
     getSystemStart = askSlottingTimestamp
+
     getEpochLastIndex = do
         var <- askSlottingVar
-        atomically $ (fromMaybe 0 . getLastEpochIndex) <$> readTVar var
+        atomically $ getLastEpochIndex <$> readTVar var
+
     getEpochSlottingData ei = do
         var <- askSlottingVar
         atomically $ lookupEpochSlottingData ei <$> readTVar var
+
     waitPenultEpochEquals target = do
         var <- askSlottingVar
         atomically $ do
             penultEpoch <- getPenultEpochIndex <$> readTVar var
             when (penultEpoch /= target) retry
+
     putEpochSlottingData ei esd = do
         var <- askSlottingVar
         atomically $ do
-            sd <- readTVar var
-            writeTVar var (addEpochSlottingData ei esd sd)
+            slottingData <- readTVar var
+            writeTVar var (addEpochSlottingData ei esd slottingData)
--- a/infra/Pos/Slotting/MemState/Holder.hs
+++ b/infra/Pos/Slotting/MemState/Holder.hs
@@ -21,7 +21,8 @@ import qualified Ether
 import           Pos.Core.Types               (Timestamp)
 
 import           Pos.Slotting.MemState.Class  (MonadSlotsData (..))
-import           Pos.Slotting.Types           (SlottingData (sdPenultEpoch))
+import           Pos.Slotting.Types           (SlottingData, addEpochSlottingData,
+                                               getPenultEpochIndex, lookupEpochSlottingData, getLastEpochIndex)
 
 ----------------------------------------------------------------------------
 -- Transformer
@@ -58,14 +59,19 @@ instance
          MonadSlotsData (Ether.TaggedTrans SlotsDataRedirectTag t m)
   where
     getSystemStart = askSlottingTimestamp
-    getSlottingData = atomically . readTVar =<< askSlottingVar
+    getEpochLastIndex = do
+        var <- askSlottingVar
+        atomically $ (fromMaybe 0 . getLastEpochIndex) <$> readTVar var
+    getEpochSlottingData ei = do
+        var <- askSlottingVar
+        atomically $ lookupEpochSlottingData ei <$> readTVar var
     waitPenultEpochEquals target = do
         var <- askSlottingVar
         atomically $ do
-            penultEpoch <- sdPenultEpoch <$> readTVar var
+            penultEpoch <- getPenultEpochIndex <$> readTVar var
             when (penultEpoch /= target) retry
-    putSlottingData sd = do
+    putEpochSlottingData ei esd = do
         var <- askSlottingVar
         atomically $ do
-            penultEpoch <- sdPenultEpoch <$> readTVar var
-            when (penultEpoch < sdPenultEpoch sd) $ writeTVar var sd
+            sd <- readTVar var
+            writeTVar var (addEpochSlottingData ei esd sd)
--- a/infra/Pos/Slotting/Ntp.hs
+++ b/infra/Pos/Slotting/Ntp.hs
@@ -35,7 +35,7 @@ import           Serokell.Util                (sec)
 import           System.Wlog                  (WithLogger, logDebug, logInfo, logWarning)
 
 import qualified Pos.Core.Constants           as C
-import           Pos.Core.Slotting            (unflattenSlotId)
+import           Pos.Core.Slotting            (flattenEpochIndex, unflattenSlotId)
 import           Pos.Core.Types               (EpochIndex, SlotId (..), Timestamp (..),
                                                mkLocalSlotIndex)
 
@@ -43,6 +43,7 @@ import           Pos.Slotting.Class           (MonadSlots (..))
 import qualified Pos.Slotting.Constants       as C
 import           Pos.Slotting.MemState.Class  (MonadSlotsData (..))
 import           Pos.Slotting.Types           (EpochSlottingData (..), findMatchingEpoch)
+import           Pos.Slotting.Util            (getLastEpochEmpatically)
 
 ----------------------------------------------------------------------------
 -- State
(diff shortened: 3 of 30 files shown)


update/Pos/Update/Poll/RollTrans.hs
```diff
@@ -74,8 +74,10 @@ instance MonadPoll m => MonadPoll (RollT m) where
         insertIfNotExist id unChangedPropsL getProposal
         deactivateProposal id
 
-    setSlottingData =
-        setValueWrapper unSlottingDataL getSlottingData setSlottingData
+    putEpochSlottingData ei esd = ether $ do
+        insertIfNotExist ei unSlottingDataL getEpochSlottingData
+        putEpochSlottingData ei esd
+
```
gromakovsky: You didn't provide an explicit definition for `delEpochSlottingData`. It's correct (it will be `delEpochSlottingData = lift . delEpochSlottingData` by default), but can you write it explicitly as it's done for `delConfirmedSV`?
```
    -- can't be called during apply
    delConfirmedSV = lift . delConfirmedSV
```

update/Pos/Update/Poll/PollState.hs
```diff
@@ -57,7 +58,9 @@ data PollState = PollState
       -- | Update proposals for each application
     , _psActivePropsIdx     :: !(HM.HashMap ApplicationName (HashSet UpId))
       -- | Slotting data for this node
-    , _psSlottingData       :: !SlottingData
+    , _psEpochSlottingData  :: !(HashMap EpochIndex EpochSlottingData)
+      -- | Last updated epoch index
+    , _psEpochLastIndex  :: !EpochIndex
```
gromakovsky: Hm, do you need to store it explicitly? I think in this case it makes sense to store slotting data in `Map` (from `containers`) rather than `HashMap`, in this case you can get last index using `findMax` function.
Update: while I was writing it I realized that your approach also makes sense and it's subjective. I think `findMax` approach makes code simpler, but `HashMap + EpochIndex` is slightly more efficient. So I don't mind if you leave it as is.

P. S. I started reviewing from bottom for some reason :) So don't be surprised to this comment _here_ (probably it can be applied to other places as well).

gromakovsky: Well, I payed more attention and I see that in this particular case (`PollState`) storing last epoch explicitly doesn't make sense at all. It's not used `getEpochLastIndex` (in `Pure.hs`) and is not update by `putEpochSlottingData`/`delEpochSlottingData`. So I suppose it's a leftover, isn't it?

update/Pos/Update/Poll/Logic/Base.hs
```diff
@@ -200,31 +200,29 @@ updateSlottingData
...
     let errFmt =
             ("can't update slotting data, stored penult epoch is "%int%
              ", while given epoch is "%int%
              ")")
-    if | sdPenultEpoch + 1 == epoch -> updateSlottingDataDo sd
+    let err = throwError $ PollInternalError $ sformat errFmt (-1) epoch
+    mres <- do
+      li <- getEpochLastIndex
+      fmap (li,) <$> getEpochSlottingData li
+    (lastIndex, esd) <- maybe err return mres
```
gromakovsky: FYI: there is `maybeThrow` function in `Pos.Util.Util`.

gromakovsky: Oh, sorry, `maybeThrow` is for `MonadThrow (throwM)`. But for this use case there is `note` in `Universum`, though I am not sure it's a good idea to use it. So I don't mind if you leave it as it is.

gromakovsky: Also I think that the error here should be different, it should explicitly say that there is no slotting data for the epoch returned by `getEpochLastIndex`. It's different from saying that our stored penult epoch is (-1), it means that some internal invariant should be violated.
But maybe this case will disappear completely, because it makes sense to ensure that there is `EpochSlottingData` for the last index at lower-level.

update/Pos/Update/Poll/Trans.hs
```diff
@@ -111,9 +110,8 @@ instance MonadPollRead m =>
             map (\dps -> (hash $ upsProposal $ dpsUndecided dps, dps)) <$>
             getDeepProposals cd
     getBlockIssuerStake e = lift . getBlockIssuerStake e
-    getSlottingData = ether $ do
-        new <- pmSlottingData <$> get
-        maybe getSlottingData pure new
+    getEpochSlottingData ei = ether $
+        MM.lookupM getEpochSlottingData ei =<< use pmSlottingDataL
```
gromakovsky: You forgot to redefine `getEpochLastIndex`. It should take `pmSlottingDataL` into account as well.

infra/Pos/Slotting/Class.hs
```diff
@@ -22,13 +22,6 @@ class MonadSlotsData m => MonadSlots m where
     -- return until current slot is known.
     getCurrentSlotBlocking :: m SlotId
 
-    -- | This function tries to predict current slot as accurately as it can.
-    -- If 'getCurrentTime' returns unreliable time,
-    -- then function returns last known slot
-    -- If our slotting data into DB is outdated,
-    -- then function tries to extrapolate slot using last know slotting data
-    getCurrentSlotInaccurate :: m SlotId
```
gromakovsky: Why did you remove this function? Even if it wasn't used, it is still quite useful.
Even if we store all known slotting data, this function is still needed.
We don't need to extrapolate time when we want to know timestamp of the _passed_ slot. That's exactly the case for transactions from blocks. If we processed block, we know slotting data for it. That's why I say extrapolation isn't needed to compute transaction's timestamp.
However, if you didn't run your node for a long time, you don't know the current slot until you receive all blocks. Even if you store whole slotting data. But you may want to know the current slot at least approximately. So this function is useful in such cases.

infra/Pos/Slotting/Types.hs
```diff
@@ -23,14 +33,65 @@ data EpochSlottingData = EpochSlottingData
...
-    { sdPenult      :: !EpochSlottingData
-    -- ^ 'EpochSlottingData' for penult (i. e. 'sdPenultEpoch') epoch.
-    , sdLast        :: !EpochSlottingData
-    -- ^ 'EpochSlottingData' for last epoch.
-    -- Last means 'sdPenultEpoch + 1'.
-    , sdPenultEpoch :: !EpochIndex
-    -- ^ Penult epoch for which 'EpochSlottingData' is known.
-    } deriving (Eq, Show, Generic)
+-- TODO: Reflect the fact that this will always have 2+ elements
+-- TODO: Make this a newtype wrapper
```
gromakovsky: Yes, can you do both TODOs within this issue? I see two approaches:
1. Make it `newtype` over `HashMap EpochIndex EpochSlottingData` and add some unsafe functions to this module which will use `error` if the invariant is violated. So that external code can use functions like `getLastEpochIndex` or `getLastEpochSlottingData` and not worry about cases where it doesn't exist (this module should be responsible for it).
Note that I find it important to use `error` rather than default values like `0`, because such cases indicate invariants violation and shouldn't be hidden behind default values.
2. Make it `data` and store two last `EpochSlottingData`s explicitly. Probably it will require you to write more code, but it will make the code safer and we will have less partial functions.

infra/Pos/Slotting/Types.hs
```diff
@@ -23,14 +33,65 @@ data EpochSlottingData = EpochSlottingData
...
+ensureEpochSlottingDataExists = fromMaybe $ EpochSlottingData 0 0
+
+-- PRIVATE
+-- AJ: TODO: DELETE
+-- ensureEpochIndexExists :: Maybe EpochIndex -> EpochIndex
+-- ensureEpochIndexExists = fromMaybe 0
+
+-- | Get the latest epoch index.
+getLastEpochIndex :: SlottingData -> Maybe EpochIndex
+getLastEpochIndex = fmap maximum . nonEmpty . HM.keys
```
gromakovsky: It's inefficient and this function is used quite often, so I propose two approaches, choose what you like more (I already wrote about it below):
1. Use ordered `Map` from `containers`.
2. Store last epoch explicitly.

src/Pos/Block/Logic/Slog.hs
```diff
@@ -171,7 +171,11 @@ slogApplyBlocks blunds = do
             SomeBatchOp $
             GS.PutTip $ headerHash $ NE.last $ getOldestFirst blunds
     sanityCheckDB
-    putSlottingData =<< GS.getSlottingData
+
+    slottingData <- GS.getAllSlottingData
+    -- TODO: Could this be more efficient?
```
gromakovsky: Yes, it can. I think it's ok to rely on the fact that slotting data changes only when we apply genesis block. Also there is invariant that all applied blocks in this function have same epoch (unfortunately, it's not written here, but it's written in `Pos.Block.Logic.Internal` which calls this function).
So we can't apply more than one genesis block and if we apply genesis block it must be head.
The simplest solution is:
• if we apply genesis block for the `i`-th epoch, then we should use `putEpochSlottingData` for the `i + 1`-th epoch;
• otherwise don't do anything.

Apart from that I just noticed that `slogRollbackBlocks` doesn't update in-memory slotting data, though it can be changed too. It will be great if you fix it too. But do it only if you want, it's not your mistake.

pva701:
@ajnsit btw you should name your commits like `[TAG] Short description`, like
`[CSL-134] Introduce 'Blah' datatype`, otherwise @gromakovsky will be sad

infra/Pos/Slotting/Types.hs
rockbmb: By the way, I'm not sure if this datatype has serialization tests, but when you turn a `newtype/data` into a `type`, remove it from any `BinarySpec/SafeCopySpec` files you can find as it's not worth testing type synonyms for `HashMap`s, for example.

update/Pos/Update/Poll/PollState.hs
rockbmb: @gromakovsky I think it should be removed.

src/Pos/Wallet/Web/BListener.hs
```diff
@@ -63,10 +63,17 @@ onApplyTracking
...
-        newTipH = NE.last oldestFirst ^. _1 . blockHeader
-    -- AJ: TODO: Efficiency
-    sd <- HM.fromList <$> GS.getAllSlottingData
+        txs         = concatMap (gbTxs . fst) oldestFirst
+        newTipH     = NE.last oldestFirst ^. _1 . blockHeader
+        newEpoch    = NE.last oldestFirst ^. _1 . epochIndexL
+
+    -- Apply blocks only from last epoch, since this can be called only with blocks
+    -- from a single epoch.
+    sd <- do
```
ksaric: I will add this to other instances if the code seems ok.

ksaric:
@gromakovsky Can someone review this and check if it's ok?

infra/Pos/Slotting/MemState/Holder.hs
```diff
@@ -21,26 +21,31 @@ import qualified Ether
...
+                                               getPenultEpochIndex,
+                                               lookupEpochSlottingData)
 
 ----------------------------------------------------------------------------
 -- Transformer
 ----------------------------------------------------------------------------
 
 -- | System start and slotting data
-type SlottingVar = (Timestamp, TVar SlottingData)
+type SystemStart = Timestamp
```
gromakovsky: Even better is to wrap it into `newtype`, but it is not necessary for this task of course.

infra/Pos/Slotting/Ntp.hs
```diff
@@ -219,46 +229,34 @@ ntpCurrentTime = do
...
-    | curTime < start + duration * C.epochSlots = Just $ SlotId epoch localSlot
-    | otherwise = Nothing
-  where
-    localSlotNumeric = fromIntegral $ (curTime - start) `div` duration
-    localSlot =
-        leftToPanic "computeSlotUsingEpoch: " $
-        mkLocalSlotIndex localSlotNumeric
-    duration = convertUnit esdSlotDuration
-    start = getTimestamp esdStart
+  li <- getEpochLastIndex
```
gromakovsky: 2 spaces (should be 4).

infra/Pos/Slotting/Types.hs
```diff
@@ -23,14 +33,56 @@ data EpochSlottingData = EpochSlottingData
...
-    -- ^ 'EpochSlottingData' for last epoch.
-    -- Last means 'sdPenultEpoch + 1'.
-    , sdPenultEpoch :: !EpochIndex
-    -- ^ Penult epoch for which 'EpochSlottingData' is known.
-    } deriving (Eq, Show, Generic)
+-- External code can use functions like getLastEpochIndex or getLastEpochSlottingData and 
+-- not worry about cases where it doesn't exist (this module should be responsible for it).
+-- Note that it's important to use error rather than default values like 0, because 
+-- such cases indicate invariants violation and shouldn't be hidden behind default values.
+type SlottingData = Map EpochIndex EpochSlottingData
```
gromakovsky: If you choose this approach, it should be wrapped into `newtype` and default constructor should be hidden. Instead there should be a constructor which takes two `EpochSlottingData`s.
Currently uncautious person can easily create invalid `SlottingData` (e. g. `mempty`) and some functions will fail unexpectedly.
In other words, it should be impossible to create invalid `SlottingData`.

infra/Pos/Slotting/Types.hs
```diff
@@ -23,14 +33,56 @@ data EpochSlottingData = EpochSlottingData
...
+
+-- | Get the latest epoch index.
+getLastEpochIndex :: SlottingData -> EpochIndex
+getLastEpochIndex = fst . M.findMax
+
+-- | Get the latest epoch slotting data.
+getLastEpoch :: SlottingData -> EpochSlottingData
```
gromakovsky: I would call it `getLastEpochData` or `getLastEpochSlottingData`. Otherwise it's not obvious from the name that it returns `EpochSlottingData` and not `EpochIndex` (if you don't see this module).

infra/Pos/Slotting/Types.hs
```diff
@@ -23,14 +33,56 @@ data EpochSlottingData = EpochSlottingData
...
+computeSlotStart :: SlotId -> EpochSlottingData -> Timestamp
+computeSlotStart SlotId{..} esd = slotTimestamp siSlot esd
+    where
+      slotTimestamp (getSlotIndex -> locSlot) EpochSlottingData{..} =
+        esdStart + Timestamp (fromIntegral locSlot * convertUnit esdSlotDuration)
 
-instance NFData SlottingData
+findMatchingEpoch :: Monad m => Timestamp -> EpochIndex -> (EpochIndex -> m (Maybe EpochSlottingData)) -> m (Maybe (EpochIndex, EpochSlottingData))
+findMatchingEpoch approxCurrentTime lastEpochIndex getterEpochSlottingData = go lastEpochIndex
+  -- Iterate over lastEpochIndex..0 and find the epoch
```
gromakovsky: It can be done more efficiently using binary search, but I think it's not essential now :)

infra/Pos/Slotting/Util.hs
```diff
@@ -88,9 +72,21 @@ getSlotStartEmpatically
 getSlotStartEmpatically slot =
     getSlotStart slot >>= maybeThrow (SEUnknownSlotStart slot)
 
+-- | Get last epoch empatically, which means
+-- that function throws exception when last epoch is unknown.
+getLastEpochEmpatically
```
gromakovsky: This function should use `getLastEpoch` and then it will not need to throw exception.

infra/Pos/Slotting/Util.hs
```diff
@@ -88,9 +72,21 @@ getSlotStartEmpatically
...
+    lastEpoch   <- getEpochLastIndex    
+    getEpochSlottingData lastEpoch >>= maybeThrow (SEUnknownLastEpoch lastEpoch)
+
 -- | Get last known slot duration.
 getLastKnownSlotDuration :: MonadSlotsData m => m Millisecond
-getLastKnownSlotDuration = esdSlotDuration . sdLast <$> getSlottingData
+getLastKnownSlotDuration = do
+  eli <- getEpochLastIndex
+  -- AJ: TODO: GOOD DEFAULT VALUE FOR SLOT DURATION?
+  maybe 0 esdSlotDuration <$> getEpochSlottingData eli
```
gromakovsky: This function should use `getLastEpoch` and then it will not need a default value.

src/Pos/Wallet/Web/BListener.hs
gromakovsky: It should be done differently. Instead of fetching whole slotting data from DB and then filtering it, you can iterate over DB and pick only desired data. Please don't do it before rebasing to master, because there everything changed a lot.

ksaric:
@gromakovsky regarding you comments above. If I create a newtype wrapper over the `Map EpochIndex EpochSlottingData` and hide it, I can't serialize the object.
If I use the pure `Map`, I can make it work. Also, even if I create a `Bi` instance, I'm not sure how the idea of saving a `Map` to a database will make sense.
I think it should be passed around as a `Map EpochIndex EpochSlottingData`, that's the most natural type. The problem is the potential invalid state. Even if I use `data SlottingData = SlottingData PenultSlottingData EpochSlottingData` I still have to create it using a function in order for it to make sense.
So other than some type-level perversions which could take me some time since I'm not really familiar with them (type level Map with a minimum of two elements), I don't know how to satisfy your requirements.

gromakovsky:
@ksaric 
> If I create a newtype wrapper over the Map EpochIndex EpochSlottingData and hide it, I can't serialize the object.

It's not true, you can define and export `SlottingData -> Map EpochIndex EpochSlottingData` (destructor). But you shouldn't export the opposite, i. e. constructor (`SlottingData -> Map EpochIndex EpochSlottingData`) because it's unsafe. You can export it if you really need it, but then add `Unsafe` prefix into the name.

If you want to add something into this map, you can export `insertSlottingData :: EpochIndex -> EpochSlottingData -> SlottingData -> SlottingData`.

> Even if I use data SlottingData = SlottingData PenultSlottingData EpochSlottingData I still have to create it using a function in order for it to make sense.

I don't see a problem with it too. You can have `SlottingData = SlottingData (Map EpochIndex EpochSlottingData) (EpochIndex, EpochSlottingData) (EpochIndex, EpochSlottingData)`. Then you can define all necessary operations easily.

ksaric:
@gromakovsky I get: ```No instance for (Pos.Binary.Class.Core.Bi
                             (Map EpochIndex EpochSlottingData))
            arising from a use of ‘Pos.Binary.Class.Core.size’```

gromakovsky:
@ksaric I think it can be closed, right?

ksaric:
@gromakovsky Yes. Closing in favor of https://github.com/input-output-hk/cardano-sl/pull/1217

PR #1109, opened 2017-07-05
[CSL-1330] Synchronised recovery header modification

int-index:


src/Pos/Block/Network/Retrieval.hs
```diff
@@ -137,14 +139,21 @@ retrievalWorkerImpl sendActions =
         handleBlockRetrieval nodeId $
             BlockRetrievalTask { brtHeader = rHeader, brtContinues = False }
 
+-- | Result of attempt to update recovery header.
 data UpdateRecoveryResult ssc
     = RecoveryStarted NodeId (BlockHeader ssc)
+      -- ^ Recovery header was absent, so we've net it.
```
int-index: s/net/set/

volhovM:
Do not delete the branch as it's going to be merged into master.

volhovM:
Or not, let's just merge csl-0.5-rc into master with these changes.

PR #1110, opened 2017-07-05
[CSM-353] Remove history cache when deleting wallet

flyingleafe:
When wallet is deleted, wallet history metadata and cache are not. This causes bugs when wallet is deleted and restored/imported again. This PR fixes it.


flyingleafe:
Tested on Mac OS with daedalus - works

PR #1112, opened 2017-07-05
Update start time in staging script

flyingleafe:
Subj


georgeee:
(Merged w/o waiting for CI because CI doesn't check scripts anyway)

PR #1115, opened 2017-07-05
[CSM-356] Enable sending to same set of addresses as senders addrs

akegalj:
If A is set of addresses we send from, and B set of addresses we
send to - this change enables that intersection of A and B can be
non empty. ie we are enabling A ∩ B = non-epty set .

Before it was forbidden to send from the same address that is in set B (see the diff)

Fixes: https://issues.serokell.io/issue/CSM-356

**Note** that I am not sure is there any case enabling this has cons. Some thoughts are here https://issues.serokell.io/issue/CSM-356#comment=96-5374 . In short, sending from address A to address A doesn't make sense (nor sending from account A to any address within the same account) and will only spend users *fee* in future. But I guess we don't care about that. UI might warn the user about such usecases as it clearly doesn't make sense and its most likely a mistake by user.

Tested and it works!

### DIFF
--- a/src/Pos/Wallet/Web/Server/Methods.hs
+++ b/src/Pos/Wallet/Web/Server/Methods.hs
@@ -34,7 +34,6 @@ import qualified Data.DList                       as DL
 import qualified Data.HashMap.Strict              as HM
 import           Data.List                        (findIndex, notElem)
 import qualified Data.List.NonEmpty               as NE
-import qualified Data.Set                         as S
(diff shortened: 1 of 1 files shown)


akegalj:
We decided we won't merge this in tn0.5 as it is ok not to spend from the same address to the same address.

We could/should improve error message in this case to let the user know that he can't spend to the same address (if that is the case).

Closing

PR #1118, opened 2017-07-06
[CSL-1330] Gracefully handle lack of checkpoints

int-index:


src/Pos/Block/Network/Logic.hs
```diff
@@ -150,20 +152,34 @@ requestTip _ nodeId conv = do
...
     uHdr <- blkGetHeader @ssc upto
-    runMaybeT $ do
-        -- no reason to make a request when we already have the latest header
-        guard (isNothing uHdr)
-        bHeaders <- MaybeT $ nonEmpty . toList <$> getHeadersOlderExp @ssc Nothing
-        pure $ MsgGetHeaders (toList bHeaders) (Just upto)
+    if isJust uHdr then return MhrrBlockAdopted else do
+        mCheckpoints <- nonEmpty . toList <$> getHeadersOlderExp @ssc Nothing
+        case mCheckpoints of
+            Nothing -> do
```
int-index: Not sure about this codepath. Can it ever happen? When does `getHeadersOlderExp` return an empty list? @volhovM

volhovM: I've removed it, will push soon.

PR #1126, opened 2017-07-06
CSL-554 Check chain quality explicitly 

gromakovsky:
What's done:
1. Slots of the last `k` blocks in our chain are stored in BlockExtra part of GState. Initially I wanted to store 1 slot per key (something like deque), but eventually I discarded this idea. It shouldn't be harmful, because we also store this data in memory (which is fine, because it's only `k` `FlatSlotId`s) and won't have to read and parse it every time.
2. I introduced `SlogContext` which currently stores only `IORef LastBlkSlots`. Its contents is the same as the value stored in DB. It's needed to avoid using DB too often.
3. I added chain quality checker which currently only logs `Info` or `Warning` if chain quality is low. This behaviour can be treated as a stub for now (there is TODO).
4. Main and block creation functions have been reworked with regards to checks whether we should create them. They are well documented, so please read documentation for details.

What's left (not for this issue):
1. Send messages to reporting server from checker, make messages more reasonable, constants tunable, etc. Will be done within CSL-1342.


src/Pos/Block/Slog/Context.hs
```diff
@@ -0,0 +1,33 @@
...
+
+import           Pos.Block.Slog.Types (HasSlogContext (..), LastBlkSlots,
+                                       SlogContext (..), scLastBlkSlots)
+import           Pos.DB.Class         (MonadDBRead)
+import           Pos.DB.GState        (getLastSlots)
+
+-- | Make new 'SlogContext' using data from DB.
+mkSlogContext :: (MonadIO m, MonadDBRead m) => m SlogContext
+mkSlogContext = do
+  _scLastBlkSlots <- getLastSlots >>= newIORef
```
volhovM: 2 spaces.

PR #1128, opened 2017-07-07
Merge CSL-1330 fix to cardano-sl-0.5

georgeee:


int-index:
Is this request competing with https://github.com/input-output-hk/cardano-sl/pull/1125 or something? What's the difference between those?

gromakovsky:
See #1125.

PR #1130, opened 2017-07-07
[CSL-1265] Integrate SCRAPE

neongreen:
Currently tests don't pass because pvss-haskell seems to be broken.


neongreen:
Okay, now tests pass.

One of the bigger changes in this PR is that now VSS keys are sorted and everything is expected to be sorted by VSS pubkey (shares, etc). Hopefully this doesn't break anything.

neongreen:
Also don't merge it before I test it locally (run e.g. 8 nodes, kill some of them, see if all secrets get recovered successfully).

neongreen:
> I don't mind if you leave this version, but I would like to know the reason if it exists.

I assumed that `getRichmenFromLrc` relies on the rollback to give the right answer. If it doesn't, I'll change it.

neongreen:
> Why did you simplify this comment? I think the previous version was fine and more descriptive :)

Okay, will unsimplify.

neongreen:
> I assumed that getRichmenFromLrc relies on the rollback to give the right answer. If it doesn't, I'll change it.

I don't understand it very well so I'll be thinking out loud. “Richmen for epoch N” means “richmen that are to be used during epoch N” = “richmen as calculated at the crucial slot for epoch N” = “richmen at a particular point during epoch N-1”. Correct?

Then, when we are at the beginning of epoch N+1, we'll be using richmen for epoch N. We already know them and don't need to do anything to calculate them.

Okay.

neongreen:
Meeeeeeeeeeeeerging...

stack.yaml
```diff
@@ -120,6 +116,8 @@ extra-deps:
 - happy-1.19.5                    # https://github.com/commercialhaskell/stack/issues/3151
 - entropy-0.3.7                   # https://github.com/commercialhaskell/stack/issues/3151
 - cborg-0.1.1.0
+- foundation-0.0.13               # pvss
+- memory-0.14.6                   # pvss
```
gromakovsky: LTS-9.1 has these versions. Maybe let's switch to it?

ssc/Pos/Ssc/NistBeacon.hs
```diff
@@ -49,7 +49,7 @@ instance Ssc SscNistBeacon where
     sscCreateNodeContext = Tagged $ const (pure ())
 
 instance SscHelpersClass SscNistBeacon where
-    sscVerifyPayload = const $ const $ Right ()
+    sscVerifyPayload _ _ = Right ()
```
gromakovsky: Very useful change!

core/Pos/Crypto/SecretSharing.hs
```diff
@@ -120,77 +125,143 @@ instance Bi SecretProof =>
...
+----------------------------------------------------------------------------
+-- Test
+----------------------------------------------------------------------------
+
+-- | You can use this to do debugging. If everything is okay with SCRAPE, it
+-- will print 'True's.
+testScrape
+    :: Int           -- ^ Threshold (number of participants = 2× threshold)
+    -> IO ()
+testScrape t = do
```
gromakovsky: Hm, why is it here and not in a regular test suite? Is it to be able to pass threshold?

core/Pos/Util/Util.hs
```diff
@@ -190,6 +196,33 @@ instance ToJSON Byte where
...
+MONADRANDOM( Rand.MonadRandom m            , Lazy.StateT s m)
+MONADRANDOM( Rand.MonadRandom m            , Strict.StateT s m)
+MONADRANDOM((Rand.MonadRandom m, Monoid w) , Lazy.RWST r w s m)
+MONADRANDOM((Rand.MonadRandom m, Monoid w) , Strict.RWST r w s m)
+MONADRANDOM( Rand.MonadRandom m            , ExceptT e m)
+MONADRANDOM( Rand.MonadRandom m            , NamedPureLogger m)
+
+instance (Monad m, Monad (t m), MonadTrans t, Rand.MonadRandom m)
+         => Rand.MonadRandom (Ether.TaggedTrans tag t m) where
+    getRandomBytes = lift . Rand.getRandomBytes
```
gromakovsky: 1. Why didn't you use the `MONADRANDOM` macro here?
2. Why not define overlappable instance for `MonadTrans`?

godtossing/Pos/Binary/GodTossing/Core.hs
```diff
@@ -20,16 +20,14 @@ import           Pos.Ssc.GodTossing.Core.Types (Commitment (..), CommitmentsMap
 import           Serokell.Util                 (allDistinct)
 
 instance Bi Commitment where
-  encode Commitment{..} = encodeListLen 3 <> encode commShares
```
gromakovsky: Was it a bug? Does it work in master? Do I understand correctly that it produces invalid CBOR but works by chance?

godtossing/Pos/Ssc/GodTossing/Workers.hs
```diff
@@ -322,32 +325,37 @@ generateAndSetNewSecret sk SlotId {..} = do
...
-                Just multiPS -> do
-                    mPair <- runMaybeT (genCommitmentAndOpening threshold multiPS)
-                    flip (maybe (reportDeserFail $> Nothing)) mPair $
-                        \(mkSignedCommitment sk siEpoch -> comm, open) ->
-                            Just comm <$ SS.putOurSecret comm open siEpoch
+                Nothing -> Nothing <$ logWarning (here "Couldn't compute participant's vss")
+                Just multiPS ->
+                    -- we use runErrorT and not runExceptT because we want
+                    -- to get errors produced by 'fail'. In the future we'll
+                    -- use MonadError everywhere and it won't be needed.
```
gromakovsky: _mda_

gromakovsky:
> Then, when we are at the beginning of epoch N+1, we'll be using richmen for epoch N. We already know them and don't need to do anything to calculate them, and the rollback won't change anything.

Yes, it seems correct, manual testing (demo) should confirm it.

ssc/Pos/Ssc/NistBeacon.hs
neongreen: yeah, it's one of those that I made for consistency and then was like “maybe I shouldn't”

godtossing/Pos/Ssc/GodTossing/Workers.hs
neongreen: IIRC, I actually fixed it in a different PR.

godtossing/Pos/Binary/GodTossing/Core.hs
neongreen: There was no bug. In the previous version of PVSS we had three fields in a commitment, now we have two.

stack.yaml
neongreen: Done.

core/Pos/Crypto/SecretSharing.hs
neongreen: No, it's because loading the test suite in repl is a pain (if you want to debug anything in scrape, after each change in the main project you hav to restart the repl).

godtossing/Pos/Ssc/GodTossing/Seed.hs
```diff
@@ -6,38 +8,40 @@ module Pos.Ssc.GodTossing.Seed
...
-calculateSeed commitments' openings lShares richmen = do
-    let commitments = getCommitmentsMap commitments' -- just unwrapping
-    let participants = getKeys commitments
+calculateSeed commitments' binVssKeys openings lShares richmen = do
+    let commitments :: HashMap StakeholderId SignedCommitment
+        commitments = getCommitmentsMap commitments' -- just unwrapping
+    let participants :: HashSet StakeholderId
+        participants = getKeys commitments
+    -- vssKeys :: HashMap StakeholderId VssPublicKey
+    vssKeys <- over _Left BrokenVssKey $ mapFailing fromBinaryM binVssKeys
```
georgeee: `vssKeys` is now passed because Scrape implementation doesn't guarantee ordering in some parts and thus we explicitly resolve stake holder against map, right?

test/Test/Pos/Ssc/GodTossing/SeedSpec.hs
```diff
@@ -43,23 +45,23 @@ spec = do
...
                return $ recoverSecretsProp n n 0 0
         prop
             "finds the seed when all shares are present" $
-            do n <- sized $ \size -> choose (1, max size 1)
+            do n <- sized $ \size -> choose (4, max size 4)
                return $ recoverSecretsProp n 0 n 0
         prop
             "finds the seed when all secrets can be recovered" $
-            do n <- sized $ \size -> choose (1, max size 1)
+            do n <- sized $ \size -> choose (4, max size 4)
```
georgeee: Can you describe effect of this change?

src/Pos/Lrc/Worker.hs
```diff
@@ -168,6 +168,23 @@ lrcDo epoch consumers tip = tip <$ do
     -- If there are blocks from 'epoch' it means that we somehow accepted them
     -- before running LRC for 'epoch'. It's very bad.
     unless (null blundsUpToGenesis) $ throwM LrcAfterGenesis
+    -- We don't calculate the seed inside 'withBlocksRolledBack' because
```
georgeee: Lool, I love it! Yummy-yummy!

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
```diff
@@ -37,42 +38,50 @@ module Pos.Ssc.GodTossing.Core.Types
...
--- Invariant which must be ensured: commShares is not empty.
+-- | Commitment is a message generated during the first stage of GodTossing.
+-- It contains encrypted shares and proof of secret.
+--
+-- There can be more than one share generated for a single participant.
 data Commitment = Commitment
-    { commExtra  :: !(AsBinary SecretSharingExtra)
-    , commProof  :: !(AsBinary SecretProof)
-    , commShares :: !(HashMap (AsBinary VssPublicKey) (NonEmpty (AsBinary EncShare)))
+    { commProof  :: !SecretProof
```
neongreen: I think yes, because `AsBinary` is only applied to things of known size and `SecretProof` can have arbitrary size. Maybe I'm wrong.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
```diff
@@ -37,42 +38,50 @@ module Pos.Ssc.GodTossing.Core.Types
...
+getCommShares :: Commitment -> Maybe [(VssPublicKey, NonEmpty EncShare)]
+getCommShares =
+    traverseOf (each . _1) fromBinaryM <=<          -- decode keys
+    traverseOf (each . _2 . each) fromBinaryM .     -- decode shares
+    HM.toList . commShares
+
 instance Ord Commitment where
-    compare = comparing commExtra <>
-              comparing commProof <>
+    compare = comparing (serialize' . commProof) <>
```
neongreen: I haven't checked whether the serialization is fast enough.

We need this instance because during CBOR decoding we need `Ord` to enforce canonicity, unfortunately. The best solution is to enforce canonicity on CBOR level but with `cborg` it's hard to do.

godtossing/Pos/Ssc/GodTossing/Seed.hs
neongreen: Right.

test/Test/Pos/Ssc/GodTossing/SeedSpec.hs
neongreen: SCRAPE doesn't work when there are less than 4 participants, because it requires the threshold to be between 2 and N−2 (N = number of participants).

node/test/Test/Pos/CryptoSpec.hs
```diff
@@ -481,18 +471,19 @@ keygenInequality a b =
...
 verifyEncShareGoodData SharedSecrets {..} =
     Crypto.verifyEncShare ssSecShare
                           (ssVSSPKs !! ssPos)
                           (fst $ ssShares !! ssPos)
 
-verifyEncShareBadSecShare :: SharedSecrets -> Crypto.SecretSharingExtra -> Property
-verifyEncShareBadSecShare SharedSecrets {..} secShare =
-    (ssSecShare /= secShare) ==>
-    not $ Crypto.verifyEncShare secShare (ssVSSPKs !! ssPos) (fst $ ssShares !! ssPos)
+verifyEncShareBadSecShare :: SharedSecrets -> Crypto.SecretProof -> Property
```
volhovM: Shouldn't it be just removed?

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
gromakovsky: `AsBinary` was also applied to `SecretSharingExtra` which is of unknown size. But it was checked carefully. In general using `AsBinary` for things of unknown size is legal as long as you are careful. I left some clarifying comments somewhere.

gromakovsky:
All checks have failed, btw.

test/Test/Pos/Ssc/GodTossing/SeedSpec.hs
gromakovsky: And what do we do if there are less than 4 participants in an epoch?

core/Pos/Arbitrary/Crypto.hs
```diff
@@ -150,10 +149,9 @@ instance (Bi w, Arbitrary w, Bi a, Arbitrary a) =>
...
 data SharedSecrets = SharedSecrets
-    { ssSecShare  :: SecretSharingExtra
-    , ssSecret    :: Secret
+    { ssSecret    :: Secret
     , ssSecProof  :: SecretProof
-    , ssShares    :: [(EncShare, Share)]
+    , ssShares    :: [(EncShare, DecShare)]
     , ssThreshold :: Threshold
     , ssVSSPKs    :: [VssPublicKey]
     , ssPos       :: Int            -- This field is a valid, zero-based index in the
```
gromakovsky: Pos of stake.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
gromakovsky: https://github.com/input-output-hk/cardano-sl/blob/db17c31720f7ec885576c6bc33b9c279d61f1450/core/Pos/Binary/Crypto.hs#L104
There is my note which you forgot to update.

core/Pos/Crypto/SecretSharing.hs
```diff
@@ -57,59 +55,66 @@ import           Pos.Crypto.Random   (deterministic, runSecureRandom)
...
 newtype VssPublicKey = VssPublicKey
-    { getVssPublicKey :: Pvss.PublicKey
-    } deriving (Show, Eq)
+    { getVssPublicKey :: Scrape.PublicKey
+    } deriving (Show, Eq, Binary.Binary)
+
+-- | Note: this is an important instance, don't change it! It dictates the
+-- order of participants when generating a commitment.
+instance Ord VssPublicKey where
+    compare = comparing Binary.encode
```
gromakovsky: Hm, perhaps it could be done more efficiently if `Ord` was defined for `Scrape.PublicKey` in `pvss-haskell`.

core/Pos/Crypto/SecretSharing.hs
```diff
@@ -57,59 +55,66 @@ import           Pos.Crypto.Random   (deterministic, runSecureRandom)
...
-    { getEncShare :: EncryptedShare
+    { getEncShareVal :: Scrape.EncryptedSi
     } deriving (Show, Eq)
 
--- | This extra data may be used to verify encrypted share.
-data SecretSharingExtra =
-    SecretSharingExtra !ExtraGen
-                       ![Commitment]
-    deriving (Show, Eq, Generic)
+-- | This extra data may be used to verify various stuff.
```
gromakovsky: Are you sure? :)

core/Pos/Crypto/SecretSharing.hs
```diff
@@ -120,77 +125,143 @@ instance Bi SecretProof =>
...
+getDhSecret (Scrape.DhSecret s) = s
 
 -- | Transform a Secret into a usable random value.
-secretToDhSecret :: Secret -> DhSecret
-secretToDhSecret = Pvss.secretToDhSecret . getSecret
+secretToDhSecret :: Secret -> Scrape.DhSecret
+secretToDhSecret = Scrape.secretToDhSecret . getSecret
 
 -- | Decrypt share using secret key. Doesn't verify if an encrypted
 -- share is valid, for this you need to use verifyEncShare.
```
gromakovsky: There is no `verifyEncShare` function anymore, so I guess this comment should be updated.

core/Pos/Crypto/SecretSharing.hs
```diff
@@ -120,77 +125,143 @@ instance Bi SecretProof =>
...
+    -- Then we check that we have enough shares and do secret recovery
+    guard (length ordered >= thr)
+    pure (coerce Scrape.recover (take thr ordered))
+
+-- | Like 'Scrape.reorderDecryptedShares', but handles the case when multiple
+-- shares were created for a single key.
+--
+-- TODO: move to pvss-haskell, maybe
+reorderDecryptedShares
+    :: [(VssPublicKey, Int)]            -- ^ Participants
```
gromakovsky: Participants + how many shares were sent to each

neongreen:
Note to self: is this ok?

```
Test.Pos.Lrc.Worker
  Lrc.Worker
    lrcSingleShotNoLock
SSC couldn't compute seed: NoSecrets for epoch epoch #1, going to reuse seed for previous epoch
SSC couldn't compute seed: NoSecrets for epoch epoch #1, going to reuse seed for previous epoch
SSC couldn't compute seed: NoSecrets for epoch epoch #1, going to reuse seed for previous epoch
SSC couldn't compute seed: NoSecrets for epoch epoch #1, going to reuse seed for previous epoch
      Computes richmen correctly according to the stake distribution right before the '8 * k'-th slot.
```

gromakovsky:
If you mean `NoSecrets` messages, then yes, it's totally ok.

gromakovsky:
We don't generate any ssc payload in tests.

godtossing/Pos/Ssc/GodTossing/Workers.hs
```diff
@@ -221,43 +226,67 @@ onNewSlotCommitment slotId@SlotId {..} sendActions
...
+        "We're not sending opening, because there is no commitment \
+        \from us in global state"
+    noOpenMsg =
+        "We don't know our opening, maybe we started recently"
+    sendOpening ourId open = do
+        mbOpen' <- case params of
+            GtOpeningNone   -> pure Nothing
+            GtOpeningNormal -> pure (Just open)
+            GtOpeningWrong  -> do
+                keys <- NE.fromList . map (asBinary . toVssPublicKey) <$>
```
georgeee: Rather complicated... Why not `arbitrary` opening?

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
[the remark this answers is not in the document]
gromakovsky: @neongreen will you update it?

test/Test/Pos/Ssc/GodTossing/SeedSpec.hs
gromakovsky: @neongreen can you answer?

godtossing/Pos/Binary/GodTossing/Core.hs
gromakovsky: Ok, I read diff incorrectly.

godtossing/Pos/Ssc/GodTossing/Core/Core.hs
```diff
@@ -79,24 +77,37 @@ import           Pos.Util.Limits               (stripHashMap)
 secretToSharedSeed :: Secret -> SharedSeed
 secretToSharedSeed = SharedSeed . getDhSecret . secretToDhSecret
 
+-- | Figure out the threshold (i.e. how many secret shares would be required
+-- to recover each node's secret) from the total number of shares.
+--
+-- We use this function to find out what threshold to use when creating
+-- shares, when verifying shares, and when recovering the secret.
+vssThreshold :: Integral a => a -> Threshold
+vssThreshold len = fromIntegral $ len `div` 2 + len `mod` 2
```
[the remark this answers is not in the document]
gromakovsky: @volhovM what concerns you? It wasn't changed it all, just moved into this module.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
```diff
@@ -119,18 +128,34 @@ type OpeningsMap = HashMap StakeholderId Opening
...
--- identified by key Y, this NonEmpty list will be at @sharesMap ! X !
--- Y@.
+-- has received and decrypted:
+--
+--   * Outer key = who decrypted the share
+--   * Inner key = who created the share
+--
+-- Let's say that there are participants {A, B, C}. If A has generated a
+-- secret and shared it, A's shares will be denoted as Ab and Ac (sent
+-- correspondingly to B and C). Then node B will decrypt the share and get
```
gromakovsky: Did you change the logic so that A doesn't generate shares for A?

godtossing/Pos/Ssc/GodTossing/Seed.hs
```diff
@@ -6,38 +8,40 @@ module Pos.Ssc.GodTossing.Seed
...
+                                           SignedCommitment, getCommShares,
+                                           secretToSharedSeed, verifyOpening,
+                                           vssThreshold)
+import           Pos.Ssc.GodTossing.Error (SeedError (..))
+import           Pos.Util.Util            (getKeys)
 
 -- | Calculate SharedSeed. SharedSeed is a random bytestring that all
 -- nodes generate together and agree on.
---
--- TODO: do we need to check secrets' lengths? Probably not.
```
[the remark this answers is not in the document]
gromakovsky: @neongreen 
> Where is this checked now?

If it's not checked, are you sure we don't need to check it?

godtossing/Pos/Ssc/GodTossing/Seed.hs
```diff
@@ -6,38 +8,40 @@ module Pos.Ssc.GodTossing.Seed
...
     -> Either SeedError SharedSeed
-calculateSeed commitments' openings lShares richmen = do
-    let commitments = getCommitmentsMap commitments' -- just unwrapping
-    let participants = getKeys commitments
+calculateSeed commitments' binVssKeys openings lShares richmen = do
+    let commitments :: HashMap StakeholderId SignedCommitment
+        commitments = getCommitmentsMap commitments' -- just unwrapping
+    let participants :: HashSet StakeholderId
+        participants = getKeys commitments
+    -- vssKeys :: HashMap StakeholderId VssPublicKey
```
gromakovsky: I think it would be better not to have this type commented out and instead have it on the next line.

godtossing/Pos/Ssc/GodTossing/Seed.hs
gromakovsky: @neongreen can you explain which order we use here and how?

core/Pos/Util/Util.hs
```diff
@@ -189,6 +187,20 @@ instance ToJSON Byte where
 instance IsString s => MonadFail (Either s) where
     fail = Left . fromString
 
+instance Rand.DRG drg => HasLoggerName (Rand.MonadPseudoRandom drg) where
+    getLoggerName = pure mempty
```
gromakovsky: 1. Are you sure it's needed? I see that pure toss uses named logger, for instance.
2. Why `pure mempty` instead of e. g. `pure "MonadPseudoRandom"`?

godtossing/Pos/Ssc/GodTossing/Types/Types.hs
```diff
@@ -97,22 +98,29 @@ instance Buildable GtGlobalState where
...
+    , gtpVssKeyPair :: !VssKeyPair  -- ^ Key pair used for secret sharing
+    , gtpBehavior   :: !GtBehavior  -- ^ Settings for the algorithm
     }
 
 data GtContext = GtContext
     {
       -- | Vss key pair used for MPC.
       gtcVssKeyPair     :: !VssKeyPair
     , -- | Flag which determines whether we want to participate in SSC.
+      -- TODO: consider putting it into GtBehavior?
```
gromakovsky: Note that it should be controlled by web API.

godtossing/Pos/Ssc/GodTossing/Workers.hs
```diff
@@ -1,5 +1,8 @@
 {-# LANGUAGE RankNTypes #-}
 
+-- Don't complain about deprecated ErrorT
+{-# OPTIONS -Wno-deprecations #-}
```
gromakovsky: It is sad.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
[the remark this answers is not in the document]
neongreen: @gromakovsky can you explain what exactly you mean by “careful”? I'm going to write that in the comment.

godtossing/Pos/Ssc/GodTossing/Seed.hs
neongreen: SCRAPE needs to know the order of shares (i.e. you can't give it shares 4, 1 and 3, you must necessarily give it 1,3,4). It also can't deduce the order of shares because the shares don't carry any IDs or anything. Thus we need to impose some arbitrary order on them. I chose to order the shares by stakeholder's VSS key, which means that we need to know stakeholders' VSS keys in addition to a `Stakeholder->[Share]` map if we want to recover the secret.

I have put this comment into the haddock for `calculateSeed`.

core/Pos/Crypto/SecretSharing.hs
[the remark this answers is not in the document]
neongreen: Unfortunately I can't document it because I have no idea how SCRAPE works.

core/Pos/Crypto/SecretSharing.hs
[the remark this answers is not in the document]
neongreen: I have added a comment.

core/Pos/Crypto/SecretSharing.hs
```diff
@@ -120,77 +125,143 @@ instance Bi SecretProof =>
...
+    map (first toInteger) $ go 1 participants
   where
-    convert (EncShare es, VssPublicKey p, Share s) = (es, p, s)
+    go :: Int                        -- ^ Index of current share
+       -> [(VssPublicKey, Int)]      -- ^ Remaining participants
+       -> [(Int, DecShare)]
+    go _ []         = []
+    go i ((k,n):ps) = case HM.lookup k shares of
+        Nothing -> go (i + n) ps
+        Just ss -> zip [i..] (take n ss) ++ go (i + n) ps
```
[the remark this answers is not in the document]
neongreen:
```haskell
-- __Description of the algorithm__
--
-- We know:
--   * the /original/ order of participants
--   * how many shares were generated for each participant
--   * a list of decrypted shares for each participant,
--     though some participants might be missing
--
-- /Note:/ we assume that if a participant isn't missing then their shares
-- are present in the right order and no shares are skipped. This is a valid
-- assumption to make because
--   * first we verify encrypted shares with 'verifyEncShares'
--     (this takes care of the order and count of shares)
--   * then we verify each decrypted share with 'verifyDecShare' against
--     the corresponding encrypted share
--   * and we don't forget to check that the counts of encrypted and
--     decrypted shares match.
--
-- After this is done, we 'go' through the list of participants and try to
-- recover a share with index 'i' for each 'i'. If we have the shares for
-- some participant (let's say 'n' shares), we can recover shares with
-- indices @[i..i+n-1]@, so we add them to the list and continue from index
-- @i+n@. If we don't have the shares for the participant, we just skip 'n'
-- shares and move to the next participant and try to find the @i+n@'th
-- share.
reorderDecryptedShares
    :: [(VssPublicKey, Int)]            -- ^ Participants + how many shares
                                        --    were sent to each
    -> HashMap VssPublicKey [DecShare]  -- ^ Decrypted shares
    -> [(Scrape.ShareId, DecShare)]
reorderDecryptedShares participants shares =
    map (first toInteger) $ go 1 participants
  where
    go :: Int                        -- ^ Index of current share
       -> [(VssPublicKey, Int)]      -- ^ Remaining participants
       -> [(Int, DecShare)]
    go _ []         = []
    go i ((k,n):ps) = case HM.lookup k shares of
        Nothing -> go (i + n) ps
        Just ss -> zip [i..] (take n ss) ++ go (i + n) ps
```

I have added an explanation, please review it.

core/Pos/Crypto/AsBinary.hs
```diff
@@ -54,31 +52,25 @@ checkLenImpl action name expectedLen len
...
 
 
-vssPublicKeyBytes, secretBytes, shareBytes, encShareBytes, secretProofBytes :: Int
+vssPublicKeyBytes, secretBytes, decShareBytes, encShareBytes :: Int
 vssPublicKeyBytes = 35   -- 33 data + 2 of CBOR overhead
 secretBytes       = 35   -- 33 data + 2 of CBOR overhead
-shareBytes        = 103  --4+33+64
-encShareBytes     = 103
-secretProofBytes  = 66   -- 64 data + 2 of CBOR overhead
+decShareBytes     = 99
```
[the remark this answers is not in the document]
neongreen: I have added a comment.

```haskell
decShareBytes     = 99   -- Point (33) + DLEQ.Proof (64) + CBOR overhead (2)
```

godtossing/Pos/Arbitrary/Ssc/GodTossing.hs
```diff
@@ -99,10 +90,13 @@ instance Arbitrary BadCommAndOpening where
...
     map (uncurry CommitmentOpening) $
-    unsafeMakePool "[generating Commitments and Openings for tests...]" 50 $
-       genCommitmentAndOpening 1 (one (asBinary vssPk))
-  where
-    vssPk = toVssPublicKey $ deterministicVssKeyGen "ababahalamaha"
+    unsafeMakePool "[generating Commitments and Openings for tests...]" 50 $ do
+      t <- R.randomRIO (3, 10)
+      n <- R.randomRIO (t*2-1, t*2)
+      vssKeys <- replicateM n $
+          toVssPublicKey . deterministicVssKeyGen <$> secureRandomBS 2
```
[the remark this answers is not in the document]
neongreen: This code doesn't exist anymore.

godtossing/Pos/Ssc/GodTossing/Seed.hs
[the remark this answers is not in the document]
neongreen: We don't need to check it because in SCRAPE we have

```haskell
-- | Secret
newtype Secret = Secret Point
    deriving (Show,Eq,NFData,Binary)

pointToDhSecret :: Point -> DhSecret
pointToDhSecret (Point p) =
    let (x, _) = SSL.ecPointToAffineGFp p256 p
     in DhSecret $ B.convert $ hashSHA256 $ i2ospOf_ 32 x
```

i.e. it will always have the same length and the attacker can't influence it. A secret is a SHA256 hash of math. MATH

godtossing/Pos/Ssc/GodTossing/Core/Core.hs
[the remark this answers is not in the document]
neongreen: [points] He is right.

godtossing/Pos/Ssc/GodTossing/Seed.hs
```diff
@@ -75,45 +79,46 @@ calculateSeed commitments' openings lShares richmen = do
...
+    -- A 'SharesMap' with decoded shares:
+    --
+    -- shares :: HashMap StakeholderId
+    --                   (HashMap StakeholderId (NonEmpty DecShare))
     shares <- over _Left BrokenShare $
               mapFailing (traverse (traverse fromBinaryM)) lShares
+    -- Get shares of some particular stakeholder's secret
+    let getShares :: StakeholderId -> HashMap VssPublicKey (NonEmpty DecShare)
+        getShares sender = HM.fromList $ catMaybes $ do
+            (recipient, innerShareMap) <- HM.toList shares
```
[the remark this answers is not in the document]
neongreen: No, we also turn a `HashMap StakeholderId` into a `HashMap VssPublicKey`.

node/test/Test/Pos/CryptoSpec.hs
neongreen: It might be possible to resurrect it. I'm going to try now.

core/Pos/Crypto/SecretSharing.hs
neongreen: Yep, but it's too late, I guess.

core/Pos/Crypto/SecretSharing.hs
neongreen: Updated.

godtossing/Pos/Ssc/GodTossing/Workers.hs
neongreen: Okay, changed.

godtossing/Pos/Ssc/GodTossing/Workers.hs
neongreen: I removed it.

node/test/Test/Pos/CryptoSpec.hs
neongreen: Okay, I resurrected this test and will resurrect the others too.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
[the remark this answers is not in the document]
gromakovsky: @neongreen I mean that for `AsBinary VssPublicKey` you can assume that `fromBinary` won't fail and can even use `error` when it fails. But for `SecretSharingExtra` you can't assume it and should handle errors gracefully (depends on the context, but for example `error` is most definitely not _gracefully_).

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
neongreen: It looks like I haven't. Oh well, I'll update the comment.

neongreen: I think that we already decrypt our own shares when we process our message, but I'm not 100% sure.

godtossing/Pos/Ssc/GodTossing/Seed.hs
neongreen: Okay, done.

core/Pos/Util/Util.hs
neongreen: It was needed at some point, might not be needed now (:shrug:) but maybe it'll be needed in the future again.

neongreen: > Why pure mempty instead of e. g. pure "MonadPseudoRandom"?

Done.

PR #1131, opened 2017-07-08
CSE-178 Reverse DB block order.

ksaric:
https://issues.serokell.io/issue/CSE-178

Now the order is correct.

### DIFF
--- a/txp/Pos/Txp/MemState/Class.hs
+++ b/txp/Pos/Txp/MemState/Class.hs
@@ -7,6 +7,7 @@
 module Pos.Txp.MemState.Class
        ( MonadTxpMem
        , askTxpMem
+       , askTxpMemAndMetrics
        , TxpHolderTag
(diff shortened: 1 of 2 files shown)


ksaric:
Still need to bump explorer side, since the fix is in the new version of `cardano-sl`.

txp/Pos/Txp/MemState/Class.hs
```diff
@@ -7,6 +7,7 @@
 module Pos.Txp.MemState.Class
        ( MonadTxpMem
        , askTxpMem
+       , askTxpMemAndMetrics
```
ksaric: https://github.com/input-output-hk/cardano-sl-explorer/pull/176#discussion_r126139776

ksaric:
Merging to `master` - https://github.com/input-output-hk/cardano-sl/pull/1134

PR #1132, opened 2017-07-09
[CSL-1307] Don't take block lock when synchronize wallet with blockchain

pva701:
* don't take the block lock while iterating over blocks w/ difficulty `<= tip's difficulty - k`
* this also solves: CSL-1319 Fix bug synchronisation of wallet


pva701:
@flyingleafe suuda idiii

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -248,67 +290,63 @@ syncWalletWithGStateUnsafe encSK = do
...
         computeAccModifier wHeader = do
             allAddresses <- getWalletAddrMetasDB Ever wAddr
             logDebug $
                 sformat ("Wallet "%build%" header: "%build%", current tip header: "%build)
-                wAddr wHeader tipHeader
-            if | diff tipHeader > diff wHeader -> runDBToil $ runGenesisToil $ evalToilTEmpty $ do
-                     -- If walletset syncTip before the current tip,
+                wAddr wHeader gstateH
+            if | diff gstateH > diff wHeader -> runWithWalletUtxo $ do
+                     -- If wallet's syncTip is before than the current tip in the blockchain,
```
flyingleafe: `than` isn't needed :)

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -248,67 +290,63 @@ syncWalletWithGStateUnsafe encSK = do
...
+              (maybe (error "No genesis block corresponding to header hash") pure <=< DB.blkGetHeader)
+
+
+runWithWalletUtxo
+    :: (MonadReader ctx m, HasLens GenesisUtxo ctx GenesisUtxo, WebWalletModeDB ctx m)
+    => ToilT () (GenesisToil m) a
+    -> m a
+runWithWalletUtxo action = do
+    walletUtxo <- WS.getWalletUtxo
+    runGenesisToil $ fst <$> runToilTLocal (fromUtxo walletUtxo) def mempty action
```
flyingleafe: So, instead of using `ToilT` over `DBToil` while getting `CAccModifier`, you use it over `GenesisToil`? This will break transaction history. If there's an incoming transactions in block which inputs don't belong to our wallet and aren't from genesis utxo, these inputs won't be found. So you cannot get rid of `DBToil` here.
You can still use the idea of pre-fetching locally cached utxo to reduce db queries, though.

flyingleafe:
We thought about that together with @pva701 and found out that we are stupid, and all sync process should be rewritten. See you in the next episode of "Tough Way To Mainnet".

PR #1133, opened 2017-07-09
[CSL-1318] Remove networkReceiveTimeout

georgeee:
+ remove `networkReceiveTimeout`
+ make `networkWaitLogInterval` configurable (not hardcoded 4s)

### DIFF
--- a/infra/Pos/Communication/Util.hs
+++ b/infra/Pos/Communication/Util.hs
@@ -8,21 +8,18 @@ module Pos.Communication.Util
 
 import           Universum
 
-import           Data.Time.Units             (Microsecond)
 import           Formatting                  (sformat, shown, (%))
 import           Mockable                    (Async, Bracket, Delay, Mockable)
 import qualified Node                        as N
 import           Serokell.Util.Base16        (base16F)
-import           System.Wlog                 (LoggerName, WithLogger, logWarning,
-                                              modifyLoggerName)
+import           System.Wlog                 (LoggerName, WithLogger, modifyLoggerName)
 
 import           Pos.Communication.BiP       (BiP)
-import           Pos.Communication.Constants (networkReceiveTimeout)
+import           Pos.Communication.Constants (networkWaitLogInterval)
 import           Pos.Communication.Protocol  (ActionSpec (..), Listener, Message (..),
                                               MessageName (..), PeerData, mapActionSpec,
                                               mapListener, mapListener')
-import           Pos.Util.TimeLimit          (CanLogInParallel, execWithTimeLimit,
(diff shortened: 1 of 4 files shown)


neongreen:
@georgeee Can you explain how CSL-1318 can cause a hardfork and how this PR addresses the problem?

infra/Pos/Infra/Constants.hs
```diff
@@ -37,8 +37,9 @@ data InfraConstants = InfraConstants
     -- ^ We consider node as known if it was pinged at most @ccEnhancedMessageTimeout@ sec ago
     , ccEnhancedMessageBroadcast :: !Word
       -- ^ Number of nodes from batch for enhanced bessage broadcast
-    , ccNetworkReceiveTimeout    :: !Int
-      -- ^ Network timeout on `recv` in milliseconds
+    , ccNetworkWaitLogInterval   :: !Int
+      -- ^ Network wait logging interval in seconds
+      --   (logging that some recv/send takes significant amount of time)
```
neongreen: I don't understand this comment. Can you rephrase it?

Does it mean that if an action takes more `ccNetworkWaitLogInterval` seconds, there'll simply be an entry in the log about it? Or are there other consequences as well?

georgeee: Only about log entry every subsequent `ccNetworkWaitLogInterval` s action continues to run

infra/Pos/Communication/Util.hs
```diff
@@ -64,54 +61,16 @@ convWithWaitLogL
...
-                            nodeId
-                    return Nothing
-                Just r  -> return r
-
-sendActionsWithTimeLimit
-    :: (Mockable Async m, Mockable Bracket m, Mockable Delay m, WithLogger m)
-    => Microsecond
-    -> N.SendActions BiP PeerData m
-    -> N.SendActions BiP PeerData m
-sendActionsWithTimeLimit timeout sendActions = sendActions
```
pva701: :triangular_ruler:

georgeee:
> @georgeee Can you explain how CSL-1318 can cause a hardfork and how this PR addresses the problem?
If a network has a cluster in China and internet there is around 60 KB/s. At some point all blocks in system were up to 1mb, but then limit changed to 2mb. And there was networkReceiveLimit set to 20s. With 60 KB/s 20s was enough to download 1mb block, but then with 2mb we're not more able to do this (lightweight connection will be dropped after 20s passed, block will never be downloaded, cluster will maintain own chain, never to meet original again (yes, this chain would end up soon due to chain quality property, so it's not truly a hardfork, but still a nasty problem))

PR #1138, opened 2017-07-10
[CSM-359] Fix not working sending tx + CSM-322 Check whether we can use purescript-bridge from LTS 

flyingleafe:
Sending transactions weren't working due to changes in txp processing related to bootstrap era logic. This is solved by explicitly stating that there's no bootstrap era in dev mode. Also error handling in `Pos.Communication.Tx` and related is rewritten to simplify things and avoid ignoring txp verification errors on API level.

In addition, in this PR LTS version of purescript-bridge is used (CSM-322). Didn't make another PR because it's minor change and it's unnecessary to run CI separately for that.

### DIFF
--- a/core/Pos/Util/Util.hs
+++ b/core/Pos/Util/Util.hs
@@ -16,6 +16,7 @@ module Pos.Util.Util
 
        , maybeThrow
        , eitherToFail
+       , eitherToThrow
        , getKeys
        , sortWithMDesc
        , leftToPanic
@@ -273,6 +274,12 @@ maybeThrow e = maybe (throwM e) pure
 -- | Fail or return result depending on what is stored in 'Either'.
 eitherToFail :: (MonadFail m, ToString s) => Either s a -> m a
 eitherToFail = either (fail . toString) pure
+
+-- | Throw exception or return result depending on what is stored in 'Either'
+eitherToThrow
+    :: (MonadThrow m, Exception e)
+    => (s -> e) -> Either s a -> m a
+eitherToThrow f = either (throwM . f) pure
 
 -- | Create HashSet from HashMap's keys
 getKeys :: HashMap k v -> HashSet k
--- a/src/Pos/Client/Txp/Util.hs
+++ b/src/Pos/Client/Txp/Util.hs
@@ -13,7 +13,7 @@ module Pos.Client.Txp.Util
        , createRedemptionTx
 
        -- * Additional datatypes
-       , TxError
+       , TxError (..)
        ) where
 
 import           Control.Lens        (traversed, (%=), (.=))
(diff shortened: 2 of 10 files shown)


core/Pos/Util/Util.hs
```diff
@@ -274,6 +275,12 @@ maybeThrow e = maybe (throwM e) pure
 eitherToFail :: (MonadFail m, ToString s) => Either s a -> m a
 eitherToFail = either (fail . toString) pure
 
+-- | Throw exception or return result depending on what is stored in 'Either'
+eitherToThrow
+    :: (MonadThrow m, Exception e)
+    => (s -> e) -> Either s a -> m a
```
gromakovsky: I think this function should have type `Either e a -> m a`. At least for consistency with `eitherToFail`. If you need to apply a function to the `Left` value, use `first`.

src/Pos/Client/Txp/Util.hs
```diff
@@ -111,15 +121,15 @@ makeRedemptionTx rsk txInputs = makeAbstractTx mkWit (map ((), ) txInputs)
...
+type InputPicker = StateT (Coin, FlatUtxo) (Either Text)
 
 -- | Given Utxo, desired source addresses and desired outputs, prepare lists
 -- of correct inputs and outputs to form a transaction
 prepareInpsOuts
     :: Utxo
     -> NonEmpty Address
     -> TxOutputs
-    -> Either TxError (TxOwnedInputs Address, TxOutputs)
+    -> Either Text (TxOwnedInputs Address, TxOutputs)
```
gromakovsky: I think it makes more sense to return `TxError` from this function if something goes wrong.

flyingleafe:
@gromakovsky though @pva701 accidentally merged it together with CSM-321, I'll answer a couple of questions.
1) The problem was that transactions didn't pass txp validation in dev mode and the error was `Transaction hasn't been added to storage: a5b7926f , reason: transaction has non-boot stake distr in boot era: []`. Thus disabling boot era in dev mode fixed this. I believe that there's a better way to fix that, and it may be the sign of deeper problem, but I needed a quick solution.
2) However, this error wasn't reported in API answer, and successful answer (with formed transaction) was returned, which led to confusion, and it took more time and thorough examination of logs to determine the cause of error. So I decided to rewrite error reporting to have all errors, including `ToilVerFailure`s, correctly reported in API answer. Also the code became simpler and shorter.

Your minor notes on `eitherToThrow` make sense, but PR is already merged) I don't think that these changes are worth separate PR.

gromakovsky:
>  I believe that there's a better way to fix that, and it may be the sign of deeper problem

Yes.

> I needed a quick solution.

Ok.

> Also the code became simpler and shorter.

Yes, I agree, I said that it's a useful change. But I don't understand why it didn't work before. For example, I am looking at [this](https://github.com/input-output-hk/cardano-sl/pull/1138/files#diff-309f72d4bc7dc71a30a34bbda61ef9bfL628) change and I see that previously it was throwing `RequestError` in `Left` case. So it sounds strange to me that successful answer could be returned.

I suppose you didn't investigate it deep enough and don't know the answer too. In this case ok, let's forget about it.

PR #1141, opened 2017-07-11
Port transaction generator to master

kantp:
We'll want to use the transaction generator in the light wallet for
performance and stress-tests.

### DIFF
--- a/lwallet/Command.hs
+++ b/lwallet/Command.hs
@@ -28,7 +28,12 @@ import           Pos.Types                  (Address (..), BlockVersion, Coin, E
                                              SoftwareVersion, decodeTextAddress, mkCoin)
 import           Pos.Update                 (SystemTag, UpId, mkSystemTag)
 
-data SendMode = SendNeighbours | SendRoundRobin | SendRandom deriving Show
+-- | Specify how transactions are sent to the network during benchmarks using 'SendToAllGenesis'.
+data SendMode =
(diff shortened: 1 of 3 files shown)


lwallet/Command.hs
```diff
@@ -27,10 +28,12 @@ import           Pos.Types                  (Address (..), BlockVersion, Coin, E
                                              SoftwareVersion, decodeTextAddress, mkCoin)
 import           Pos.Update                 (SystemTag, UpId, mkSystemTag)
 
+data SendMode = SendNeighbours | SendRoundRobin | SendRandom deriving Show
```
gromakovsky: Can you write a comment about this type? It's not obvious for unprepared reader.

kantp: Yes, good idea!

lwallet/Main.hs
```diff
@@ -69,7 +82,9 @@ import           Pos.Wallet.Light           (LightWalletMode, WalletParams (..),
 import           Pos.WorkMode               (RealMode, RealModeContext)
 
 import           Command                    (Command (..), ProposeUpdateSystem (..),
-                                             parseCommand)
+                                             SendMode (..), parseCommand)
+import           Prelude                    (id) -- TODO: this should be exported from Universum.
```
kantp: Ah, thank you.

PR #1142, opened 2017-07-11
[CSL-1336] Ensure wallet nodes work well with `--no-ntp`

pva701:
This PR doesn't solve all our problems w/ `--no-ntp`, it updates genesises and simple slotting.
* We won't use OUR PERFECT NTP in mainnet, hence I added `--no-ntp` flag to dev scripts
* I and @georgeee decided to add field to simple slotting with last returned slot, to avoid situation when we return descending sequence of slots from slotting-functions


pva701:
Goooosha, approve?!

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -298,9 +301,9 @@ instance MonadSlotsData (TestInitMode ssc) where
     putSlottingData = putSlottingDataDefault
 
 instance MonadSlots (TestInitMode ssc) where
-    getCurrentSlot = getCurrentSlotSimple
-    getCurrentSlotBlocking = getCurrentSlotBlockingSimple
-    getCurrentSlotInaccurate = getCurrentSlotInaccurateSimple
+    getCurrentSlot = getCurrentSlotSimple =<< mkSimpleSlottingVar
+    getCurrentSlotBlocking = getCurrentSlotBlockingSimple =<< mkSimpleSlottingVar
+    getCurrentSlotInaccurate = getCurrentSlotInaccurateSimple =<< mkSimpleSlottingVar
     currentTimeSlotting = currentTimeSlottingSimple
```
georgeee: Is it ok that `currentTimeSlotting` is potentially not coherent with `getCurrentSlotXX`?

pva701: dunno, I thought it's not a problem.
cc @gromakovsky

infra/Pos/Slotting/Impl/Simple.hs
```diff
@@ -24,30 +28,50 @@ import           Pos.Slotting.Types          (SlottingData (..))
...
+----------------------------------------------------------------------------
+
+data SimpleSlottingState = SimpleSlottingState
+    { _sssLastSlot :: !SlotId
+    }
+
+type SimpleSlottingVar = TVar SimpleSlottingState
+
+mkSimpleSlottingVar :: MonadIO m => m SimpleSlottingVar
+mkSimpleSlottingVar = atomically $ newTVar $ SimpleSlottingState $ unflattenSlotId 0
```
gromakovsky: There is `newTVarIO` IIRC.

test/Test/Pos/Block/Logic/Mode.hs
gromakovsky: I think we can't do better.

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -298,9 +301,9 @@ instance MonadSlotsData (TestInitMode ssc) where
     putSlottingData = putSlottingDataDefault
 
 instance MonadSlots (TestInitMode ssc) where
-    getCurrentSlot = getCurrentSlotSimple
-    getCurrentSlotBlocking = getCurrentSlotBlockingSimple
-    getCurrentSlotInaccurate = getCurrentSlotInaccurateSimple
+    getCurrentSlot = getCurrentSlotSimple =<< mkSimpleSlottingVar
```
gromakovsky: It doesn't make sense to create a new slotting var every time.

gromakovsky: On the other hand, in init mode it's probably fine, so I don't mind.

PR #1144, opened 2017-07-11
[CSL-1317] Integrate pure db into tests

volhovM:
Subj. Now pure db is used in block test mode. 

Another way would be to modify `InitContextMode`, but i tried it and it didn't work out.

### DIFF
--- a/src/Pos/DB/Block.hs
+++ b/src/Pos/DB/Block.hs
@@ -27,6 +27,9 @@ module Pos.DB.Block
        , dbGetHeaderPureDefault
        , dbGetBlockPureDefault
        , dbGetUndoPureDefault
+       , dbGetBlockSscPureDefault
+       , dbGetUndoSscPureDefault
+       , dbGetHeaderSscPureDefault
        , dbPutBlundPureDefault
 
        -- * MonadBlockDB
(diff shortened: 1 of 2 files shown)


src/Pos/DB/Block.hs
```diff
@@ -266,6 +269,24 @@ dbPutBlundPureDefault (blk,undo) = do
...
+    -> m (Maybe (SscBlock ssc))
+dbGetBlockSscPureDefault = fmap (toSscBlock <$>) . dbGetBlockPureDefault
+
+dbGetUndoSscPureDefault ::
+       forall ssc ctx m. (MonadPureDB ctx m, SscHelpersClass ssc)
+    => HeaderHash
+    -> m (Maybe ())
+dbGetUndoSscPureDefault = fmap (const () <$>) . dbGetUndoPureDefault @ssc
+
+dbGetHeaderSscPureDefault ::
```
gromakovsky: Btw, the functions in this module shouldn't have `Default` suffix, but it's minor.

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -152,11 +146,32 @@ instance Arbitrary TestParams where
...
+-- Init mode with instances
+----------------------------------------------------------------------------
+
+-- The fields are lazy on purpose: this allows using them with
+-- futures.
+data TestInitModeContext ssc = TestInitModeContext
+    { imcDBPureVar          :: DBPureVar
+    , imcGenesisUtxo        :: GenesisUtxo
+    , imcSlottingVar        :: (Timestamp, TVar SlottingData)
+    , imcSlottingContextSum :: SlottingContextSum
```
gromakovsky: It's not needed here.

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -180,22 +195,21 @@ makeLensesWith postfixLFields ''BlockTestContext
...
-            let initCtx =
-                    InitModeContext
-                        nodeDBs
-                        tpGenUtxo
-                        futureSlottingVar
-                        SCSimple
-                        futureLrcCtx
-            runInitMode @SscGodTossing initCtx $
-                bracketBlockTestContextDo nodeDBs putSlottingVar putLrcCtx
+bracketBlockTestContext testParams@TestParams {..} callback = do
```
gromakovsky: It's not bracket anymore and probably will never be bracket.

src/Pos/DB/Block.hs
volhovM: Why not? There are lots of `aoeuDefault`, I've added `aoeuPureDefault`s also.

gromakovsky: `Default` suffix is used when there is only one implementation. But here we have `Pure` and `Real`/`Rocks` implementations. So `Default` suffix is redundant.

volhovM: I thought that `Default` suffix means "put this method in the instance implementation" which makes a lot of sense for me.

PR #1145, opened 2017-07-11
[CSL-1354] Introduce dlgNormalize

volhovM:
🎉

### DIFF
--- a/src/Pos/Block/Logic/Internal.hs
+++ b/src/Pos/Block/Logic/Internal.hs
@@ -165,6 +165,9 @@ applyBlocksUnsafeDo blunds pModifier = do
         , sscBatch
         , slogBatch
         ]
+    -- We normalize all mempools except the delegation one.
+    -- That's because delegation mempool normalization is harder and is done
+    -- within block application.
     sscNormalize
 #ifdef WITH_EXPLORER
     eTxNormalize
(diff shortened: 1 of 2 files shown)


src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -591,3 +571,29 @@ dlgRollbackBlocks blunds = do
...
+       , MonadReader ctx m
+       , HasLens LrcContext ctx LrcContext
+       )
+    => m ()
+dlgNormalize = do
+    tip <- DB.getTipHeader @ssc
+    let tipEpoch = tip ^. epochIndexL
+    fromGenesisPsks <-
+        map pskIssuerPk <$> (getPSKsFromThisEpoch @ssc) (headerHash tip)
+    pure ()
```
gromakovsky: Useless line.

src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -591,3 +571,29 @@ dlgRollbackBlocks blunds = do
...
+    oldPool <- runDelegationStateAction $ do
+        dwEpochId .= tipEpoch
+        dwThisEpochPosted .= HS.fromList fromGenesisPsks
+        uses dwProxySKPool toList
```
gromakovsky: I suppose you should also clean the mempool before processing PSKs.

src/Pos/Block/Logic/Internal.hs
```diff
@@ -167,6 +168,7 @@ applyBlocksUnsafeDo blunds pModifier = do
         , slogBatch
         ]
     sscNormalize
+    dlgNormalize @ssc
```
gromakovsky: I don't like that it's done here. This normalization traverses all blocks from the epoch and might be quite expensive. And it should be done only after rollback. So I propose to move it to `rollbackBlocksUnsafe` and write few comments about all this stuff.

Also they way normalization is done after block application looks correct but is currently fragile too. If we add more rules, it might become broken. I see that it's not so easy to make it in the same way as in other components, so I don't insist on changing it.

volhovM:
So failed CI is totally irrelevant and is probably a CI bug, so i'm merging this.

PR #1146, opened 2017-07-12
[CSM-362] Sync progress changed datatype - fix api

akegalj:
Fixes https://issues.serokell.io/issue/CSM-362 . This introduces breaking change for UI as it has to take into account another Object nesting:
 * before `getChainDifficulty: 2`
 * now `getChainDifficulty: { getBlockCount: 2 } `

```
> api.syncProgress().then(console.log)
Promise {
  <pending>,
  domain: 
   Domain {
     domain: null,
     _events: { error: [Function: debugDomainError] },
     _eventsCount: 1,
     _maxListeners: undefined,
     members: [] } }
> { _spPeers: 0,
  _spNetworkCD: { getChainDifficulty: { getBlockCount: 2 } },
  _spLocalCD: { getChainDifficulty: { getBlockCount: 2 } } }

```

cc @darko-mijic

### DIFF
--- a/src/Pos/Aeson/Types.hs
+++ b/src/Pos/Aeson/Types.hs
@@ -44,7 +44,6 @@ instance ToJSON TxOut where
         "address" .= sformat build txOutAddress ]
 
 deriving instance ToJSON SlotCount
-deriving instance ToJSON BlockCount
 
 -- NOTE: some of these types are used on frontend (PureScript).
 -- We are automatically deriving instances there and they are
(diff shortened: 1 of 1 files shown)


akegalj:
@flyingleafe purescript-bridge is working ok. But aeson and argonaut (ie, purescripts aeson) are in very sensitive sync. Argonaut is configured to work well with deriveToJson which is using TH under the hood with some options tweaked. `deriving instance` won't work because of GeneralizedNewtypeDeriving which changes the implementation (if I remember correctly)

PR #1147, opened 2017-07-12
[never merge] testnet staging

domenkozar:


domenkozar:
@georgeee Let me elaborate why this is needed. Staging environment needs a different magic number, so cardano-sl needs to be compiled with correct constants, so that daedalus picks that binary up. It's the same constants that match with `testnet_staging` name deployed on staging.

If I should be doing things otherwise, let me know.

domenkozar:
@pva701 why did you merge this? It says *never merge* in the pr?

gromakovsky:
I guess it happened automatically when he merged #1198 which contains the commit from this PR. So the problem is with #1198, it shouldn't have this commit.

pva701:
@domenkozar ^ this

domenkozar:
Why did it pass the review then? :( I'll revert and reopen this PR.

gromakovsky:
It was approved only by @pva701 and he didn't have enough context. I wrote about this problem in comments, but didn't reject (and didn't approve) because I wasn't strict enough. Next time I will be more strict, noted.

PR #1149, opened 2017-07-12
CSL-1247 put explorer into csl

ksaric:
https://issues.serokell.io/issue/CSL-1247

As is, merge from `explorer` `master`.
Used `git subtree add -P explorer https://github.com/input-output-hk/cardano-sl-explorer master`.

**EDIT** do not merge before @domenkozar gives us green light for Nix build feature (TODO add link)


ksaric:
Not sure whether we should put the **explorer** Haskell source in the `cardano-node` _src_ directory or not?
Similar to the wallet?

gromakovsky:
@ksaric 
> Not sure whether we should put the explorer Haskell source in the cardano-node src directory or not?

No, let's keep it in `explorer` folder. 

>  Similar to the wallet?

Sooner or later we will move wallet in `wallet` folder.

> Can I add the modules to cardano-node or will it get in the way there?

We should put as much explorer logic to `explorer` directory as possible.

explorer/LICENSE
```diff
@@ -1,20 +0,0 @@
-Copyright (c) 2016 IOHK
```
gromakovsky: We have a `LICENSE` file in all directories with different packages (`tools`, `update`, etc.). I think it should be in `explorer` too.

gromakovsky:
1. Something should be done with CI. AFAIU, currently it doesn't build explorer.
2. I think `scripts/build/cardano-sl.sh` should build explorer too if `--explorer` is passed.

akegalj:
@ksaric 
> Something should be done with CI. AFAIU, currently it doesn't build explorer.
@domenkozar said he can help us with CI if help is needed

explorer/LICENSE
ksaric: Ok, will return the `LICENCE` back.

domenkozar:
Indeed. Before we add explorer, we should switch to building packages with Nix first. Hopefully this week.

PR #1150, opened 2017-07-12
[CSL-1356] Create the BlockEvent type

int-index:


int-index:
Okay, I changed the module name.

int-index:
> But why not Block.Event?

Because `Generator.BlockEvent` is going to import `Generator.Block`, and it's weird to import a module from upward in the hierarchy.

gromakovsky:
It's not necessary to import `Generator.Block` from there, you could do vice versa.

PR #1151, opened 2017-07-12
regen default.nix to update cardano-report-server

cleverca22:


cleverca22:
@domenkozar https://hydra.iohk.io/jobset/serokell/cardano-pr-1151
the tests on hydra now pass (they where previously failing on most PR's)

PR #1154, opened 2017-07-12
[CSL-1281] Use constants from database, not genesis constants

neongreen:
There were three places overall where genesis constants were used wrongly.

The first commit deals with two places in `update`. The first commit is small and easy.

The second commit deals with `computeSharesDistr`. The second commit is fat and introduces a new monad class and a new newtype. I don't like `computeSharesDistr`.

---

Questions:

1. Is it okay that `generateAndSetNewSecret` uses block version data for current slot and not the slot it was passed?

2. I used `MonadGState` instead of `MonadDBRead` in some places because I wasn't sure about adding an `cardano-sl-update` dependency for `cardano-sl-ssc`. I can change that if needed.

3. I use `getAdoptedBVData` from `MonadPollRead` in the first commit, but `gsAdoptedBVData` from `MonadGState` in the second commit. Since I don't understand the difference well enough (despite @pva701's explanation), I'm not sure that it's correct.

### DIFF
--- a/test/Test/Pos/Ssc/GodTossing/ComputeSharesSpec.hs
+++ b/test/Test/Pos/Ssc/GodTossing/ComputeSharesSpec.hs
@@ -13,6 +13,7 @@ import           Universum
 import           Pos.Constants         (genesisMpcThd)
 import           Pos.Core              (mkCoin)
 import           Pos.Core.Coin         (coinPortionToDouble, sumCoins)
+import           Pos.Lrc               (RichmenStake)
 import           Pos.Lrc.Arbitrary     (GenesisMpcThd, InvalidRichmenStake (..),
                                         ValidRichmenStake (..))
 import qualified Pos.Ssc.GodTossing    as T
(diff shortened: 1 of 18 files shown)


test/Test/Pos/Ssc/GodTossing/ComputeSharesSpec.hs
```diff
@@ -37,22 +38,25 @@ spec = describe "computeSharesDistr" $ do
     validRichmenStakeWorksDesc = "Given a valid distribution of stake, calculating the\
     \ distribution of shares successfully works."
 
+computeShares' :: RichmenStake -> Either T.TossVerFailure T.SharesDistribution
+computeShares' stake = runExcept $ T.computeSharesDistrPure stake genesisMpcThd
```
gromakovsky: What? It seems useless, because `Either Patak` is instance of `MonadError Patak`.

gromakovsky: Aha, it was in old code too, ok.

gromakovsky:
> Is it okay that generateAndSetNewSecret uses block version data for current slot and not the slot it was passed?

Hm, probably not okay, but shouldn't really matter, I think. The worst which can happen is that we will create invalid secret which will not be accepted. And later we will create valid secret after we synchronize with blockchain. Also it's unlikely to happen, because we have `lrcActionOnEpochReason` there.
We can improve it by checking whether our tip header is from the current epoch, but it's minor.

> I used MonadGState instead of MonadDBRead in some places because I wasn't sure about adding an cardano-sl-update dependency for cardano-sl-ssc. I can change that if needed.

It's ok. That's exactly the purpose of `MonadGState`.

> I use getAdoptedBVData from MonadPollRead in the first commit, but gsAdoptedBVData from MonadGState in the second commit. Since I don't understand the difference well enough (despite @pva701's explanation), I'm not sure that it's correct.

It's also ok. `MonadPollRead` accounts for modifications made by mempool/blocks which are being applied. But we know that protocol version can change only as a result of genesis block processing and we process them separately, so everything is ok.

PR #1155, opened 2017-07-12
CSL-1178 Implement basic block generator

gromakovsky:
I almost implemented a generator represented by the following function:
```
genBlocks ::
       MonadBlockGen ctx m
    => BlockGenParams
    -> m (OldestFirst [] (Blund SscGodTossing))
```

`MonadBlockGen` conceptually consists of two parts: constraints on base monad and constraints on context. Constraints on constant contain only persistent data (DB) and other data which replicates it (e. g. `SlogContext`, `SlottingData`, etc.). This data is cloned and is used be generator. Old data is still available. Other in-memory data (mempool and similar) is created from scratch.
Payload generation is not implemented. It's not for this task.

### DIFF
--- a/core/Pos/Core/Slotting.hs
+++ b/core/Pos/Core/Slotting.hs
@@ -9,6 +9,8 @@ module Pos.Core.Slotting
        , crucialSlot
        , unsafeMkLocalSlotIndex
        , isBootstrapEra
+       , epochOrSlot
+       , epochOrSlotToSlot
        ) where
 
 import           Universum
@@ -16,8 +18,8 @@ import           Universum
 import           Pos.Core.Class     (HasEpochOrSlot (..), getEpochOrSlot)
 import           Pos.Core.Constants (epochSlots, slotSecurityParam)
 import           Pos.Core.Types     (EpochIndex (..), EpochOrSlot (..), FlatSlotId,
-                                     LocalSlotIndex, SlotCount, SlotId (..), epochOrSlot,
-                                     getSlotIndex, mkLocalSlotIndex)
+                                     LocalSlotIndex, SlotCount, SlotId (..), getSlotIndex,
+                                     mkLocalSlotIndex)
 import           Pos.Util.Util      (leftToPanic)
 
 -- | Flatten 'SlotId' (which is basically pair of integers) into a single number.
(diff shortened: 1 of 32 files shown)


src/Pos/Generator/Block/Mode.hs
```diff
@@ -0,0 +1,245 @@
+-- | Execution mode used by blockchain generator.
+
+module Pos.Generator.Block.Mode
+       ( MonadBlockGenBase
+       , MonadBlockGen
+       , BlockGenContext (..)
+       , BlockGenMode
+       , mkBlockGenContext
+
+       , InitBlockGenContext (..) -- useless
```
volhovM: Yes, it is. I think i did it because otherwise record fields were not used. Will prefixing them with `_` work?

gromakovsky: You did it? 🙄 

I think it's fine to just export it as a workaround. I didn't want to think about smart solution.

src/Pos/Delegation/Class.hs
```diff
@@ -78,13 +81,19 @@ instance Default DelegationWrap where
         msgCacheLimit = Just dlgCacheParam
         confCacheLimit = Just (dlgCacheParam `div` 5)
 
+type DelegationVar = RWVar DelegationWrap
+
+-- | Make a new 'DelegationVar'.
+--
+-- FIXME: it creates 'DelegationVar' with '_dwThisEpochPosted =
+-- HS.empty', is it even legal?
```
gromakovsky: Okay, I would like to move it into `mkDelegationVar` to make it harder to create `DelegationVar` and forget to initialize it.

gromakovsky: Long ago I thought it would be better to have it in `Scenario`, but later I stopped doing so.

core/Pos/Core/Slotting.hs
```diff
@@ -71,6 +79,14 @@ crucialSlot epochIdx = SlotId {siEpoch = epochIdx - 1, ..}
         leftToPanic "crucialSlot: " $
         mkLocalSlotIndex (fromIntegral (epochSlots - slotSecurityParam - 1))
 
+instance HasEpochIndex EpochOrSlot where
+    epochIndexL = lens (epochOrSlot identity siEpoch) setter
```
int-index: There's a `HasEpochIndex` instance for `Either`, its implementation is way simpler because it uses the `choosing` combinator. Since `EpochOrSlot` is a newtype over `Either Epoch Slot`, I think you'd want to reuse that instance, perhaps like this:

```
epochIndexL = _EpochOrSlot . epochIndexL
```

where `_EpochOrSlot` is an isomorphism that can be generated by `makePrisms ''EpochOrSlot`.

gromakovsky: I think it's a mistake that `EpochOrSlot` is a newtype over `Either EpochIndex SlotId`. It should be `EpochOrSlot = EOSepoch !EpochIndex | EOSslot !SlotId`.

PR #1158, opened 2017-07-13
[CSL-1368] Mock CurrentTime and Delay

int-index:


test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -102,7 +102,9 @@ sudoLiftIO m = BaseMonad (liftIO m)
 
 instance MonadIO BaseMonad where
     liftIO m = BaseMonad . liftIO $ do
-        putStr ("Don't do IO in tests! " :: String)
+        -- if you see a lot of stars in the test log, a hunting season for
+        -- not-mocked operations is open.
+        putStr ("*" :: String)
```
pva701: lol, can you explain why this solution is better than `putStr ("Don't do IO in tests! " :: String)`? just meme?

txp/Pos/Txp/MemState/Class.hs
```diff
@@ -28,21 +28,22 @@ import           Control.Monad.Trans.Control (MonadBaseControl)
...
 import           Pos.Txp.Toil.Types          (MemPool (..), UtxoModifier)
-import           Pos.Util.TimeWarp           (currentTime)
 
 data TxpHolderTag
 
 -- | More general version of @MonadReader (GenericTxpLocalData mw, TxpMetrics) m@.
 type MonadTxpMem ext ctx m
      = ( MonadReader ctx m
        , HasLens TxpHolderTag ctx (GenericTxpLocalData ext, TxpMetrics)
+       , Mockable CurrentTime m
```
gromakovsky: Personally I don't like it here. When I see `MonadBlaMem` I treat it as an extensible version of `MonadReader BlaContext`.

test/Test/Pos/Block/Logic/Mode.hs
int-index: Too many messages, I needed to make it short.

PR #1159, opened 2017-07-13
[TW-154] bump network-transport revision

avieth:
This one has a fix to prependLength which checks for overflow. It's in
mainline master but we use a fork.


gromakovsky:
There were some bugs with CI, rebase on top of latest master to make CI pass (hopefully).

PR #1160, opened 2017-07-13
[CSL-1372] Fix/stabilize master

pva701:
* Got rid of endless sending of Inv

There is still message `Reporting misbehaviour "Poor chain quality for the last 'k' blocks, less than 0.654: 0.500"`


src/Pos/Block/Worker.hs
```diff
@@ -141,7 +141,7 @@ blockCreator (slotId@SlotId {..}) sendActions = do
             else logDebug $ sformat ("Trimmed leaders: "%listJson) $
                             dropAround (fromIntegral $ fromEnum $ siSlot) 10 strLeaders
 
-        proxyCerts <- getProxySecretKeys -- TODO rename it with "light" suffix
+        proxyCerts <- getProxySecretKeysLight
```
gromakovsky: 🎉

PR #1161, opened 2017-07-14
[CSL-1238] Store diff between epoch start and 'systemStart'

pva701:


infra/Pos/Slotting/Types.hs
```diff
@@ -16,8 +16,8 @@ import           Pos.Util.Util   ()
 data EpochSlottingData = EpochSlottingData
     { esdSlotDuration :: !Millisecond
     -- ^ Slot duration actual for given epoch.
-    , esdStart        :: !Timestamp
-    -- ^ Time when epoch starts.
+    , esdStartDiff    :: !Timestamp
```
gromakovsky: I don't like that `Timestamp` is used to represent difference.

gromakovsky:
Change type of `esdStartDiff`, fix warnings, ping me. Eventually system start shouldn't be needed to initialize DB.

pva701:
@gromakovsky sooda idi

PR #1163, opened 2017-07-14
PR to track replacing use of fork with async [do not merge]

chrisdone:
After putting together the doc entitled 'IOHK: Recommendation for thread management', we've been given approval to try to replace uses of `fork` in the codebase with `async` combinators. A brief recap of that doc is that:

* `fork` does not raise the exception in the parent thread, so exceptions are often lost and threads no longer running and nobody knows about it.
* `fork` also does not link the child and parent, so if the parent is killed or exits, *usually* you want the child to also be killed too, lest you create "leaked threads".  

In any case the meaning of a `fork` is usually not clear and the above factors aren't explicitly accounted for.

This PR will track instances found of `fork` in the codebase and whether I've successfully replaced them with approval, or whether the meaning of the instance isn't clear and needs clarification. In the worst case, we'll have those instances of concurrency documented in the code.

- [x] `runWallet` [launches](https://github.com/input-output-hk/cardano-sl/compare/master...fpco:async-fpco-rewrites#diff-7574fa4c86e44dfe90afe7d543dea188L67) threads with fork and then sleeps forever. Instead, I propose to replace this with a call to `mapConcurrently`.
- [x] `runNode` [forks n nodes and then waits](https://github.com/input-output-hk/cardano-sl/compare/master...fpco:async-fpco-rewrites#diff-aaaf713cd8f8f86c844cb1b1eb6d61e1L87). I also proposed to launch those threads with `mapConcurrently` and then concurrently run that and wait for the threads to shut down. Ivan Gromakovskii raised the concern that the `mapConcurrently` might run forever if one of the threads does not shut down; as the purpose of this shutdown procedure is to tell threads with an STM variable that they must shutdown and they are supposed to pay attention to this and do their shutdown process in time. With that in mind, I think that `race` is the correct combinator instead. So we would preserve existing behavior but also get error handling.
- [ ] We have [this `fork checkForUpdate`](https://github.com/input-output-hk/cardano-sl/blob/423f08f49a49d7e8aa307dbce55e31a4ddfee5e7/src/Pos/Update/Worker.hs#L23) which lacks justification. I asked around and Ivan believes it's not necessary:
    > but I think it was added because these actions can be done in parallel and we don't want one of them to wait for another. And now I see that it doesn't make sense because `fork` is done for the second action which will wait for the first one anyway

  I tried hard but my git-fu was unsuccessful at finding the original author of this code, lost in a tree of merge commits. In any case, this use of `fork` remains a mystery.
 
  **04/08/17 update**: 

   > The original commit is 73abb29. I asked the author of this commit about this change. I am 99% sure that it should be concurrently (processNewSlot s) checkForUpdate instead.

* [ ] There's a pattern of forking of peer to peer announcements [here](https://github.com/fpco/cardano-sl/blob/45f51bc473117b86a0c079b04d35d6dde9a33c6e/src/Pos/Block/Network/Logic.hs#L542) and [here](https://github.com/fpco/cardano-sl/blob/45f51bc473117b86a0c079b04d35d6dde9a33c6e/src/Pos/Block/Worker.hs#L217). Is there a reason for the fork? Does it have to be done in parallel? If it fails, I'm not sure whether that will be picked up immediately and rather later in the logs.

   **04/08/17 update**: 

  > I think fork is redundant here. This onNewSlotWhenLeader action is launched from onNewSlotWorker which forks threads for each action anyway. And this is basically the last IO, so even if it takes a lot of time, it will not delay anything.
  > Maybe earlier (long ago) this fork was used because there was another IO after this announceBlock and announceBlock could be long (it uses network so can take significant and unpredictable amount of time).
  > I think that now this fork should be removed.

I'll add more as I find them. I'll bump a maintainer when I'm ready for review & merge.


chrisdone:
I've pushed a commit [here](https://github.com/input-output-hk/cardano-sl/pull/1163/commits/c4201dbe8cd60a9ddc43f62de350fc4f86eb2765) which removes use of the shutdown counter, instead only using the shutdown variable and async functions. 

The logic being that the worker threads just run and do their business. If one of them fails, they are all killed. Or, if the shutdown signal is given to the custodian thread, then that branch wins the "race" and the worker threads are all killed. In either case threads are cleaned up.

chrisdone:
Has anybody had a chance to review this PR yet?

gromakovsky:
I reviewed, looks good to me so far.
> I tried hard but my git-fu was unsuccessful at finding the original author of this code, lost in a tree of merge commits. In any case, this use of fork remains a mystery.

The original commit is 73abb2934358d4230bc193dc7dfa0ad04c0fab3f. I asked the author of this commit about this change. I am 99% sure that it should be `concurrently (processNewSlot s) checkForUpdate` instead.

> https://github.com/fpco/cardano-sl/blob/45f51bc473117b86a0c079b04d35d6dde9a33c6e/src/Pos/Block/Worker.hs#L217

I think `fork` is redundant here. This `onNewSlotWhenLeader` action is launched from `onNewSlotWorker` which forks threads for each action anyway. And this is basically the last `IO`, so even if it takes a lot of time, it will not delay anything.
Maybe earlier (long ago) this `fork` was used because there was another `IO` after this `announceBlock` and `announceBlock` could be long (it uses network so can take significant and unpredictable amount of time).
I think that now this `fork` should be removed.

> https://github.com/fpco/cardano-sl/blob/45f51bc473117b86a0c079b04d35d6dde9a33c6e/src/Pos/Block/Network/Logic.hs#L542

I don't know this code well. I can take a deeper look at it, but I guess @volhovM or @georgeee should know it better and probably can clarify this `fork`.

chrisdone:
@gromakovsky Thanks for the updates and digging, very clarifying!

chrisdone:
One thing that I haven't been able to understand is this section of `onNewSlotDo`:  https://github.com/fpco/cardano-sl/blob/c4201dbe8cd60a9ddc43f62de350fc4f86eb2765/infra/Pos/Slotting/Util.hs#L151..L152

```haskell
    -- Fork is necessary because action can take more time than duration of slot.
    when startImmediately $ void $ fork $ action curSlot
    -- check for shutdown flag again to not wait a whole slot
    do
        let nextSlot = succ curSlot
        Timestamp curTime <- currentTimeSlotting
        Timestamp nextSlotStart <- getSlotStartEmpatically nextSlot
        let timeToWait = nextSlotStart - curTime
        when (timeToWait > 0) $ do
            when withLogging $ logTTW timeToWait
            delay timeToWait
        onNewSlotDo withLogging (Just nextSlot) True action
```

So my understanding is that the action has to be performed at *exactly* the right interval per slot. Assuming this is right, I have two observations:

1. It's written that `action` can take more time than duration of slot.
1. When using delay provided by `threadDelay`, it means "at least" as a minimum bound. The maximum bound of the delay could be larger, we don't know how much, because the GHC runtime might be doing other things. Normally, a loop using `threadDelay` suffers from a kind of "drift" where it would become out of sync with e.g. a cron job set to every minute. In this code, however, the delay is calculated by time until the next slot from the current time, so that shouldn't be a problem. Agreed?
2. There seems to be the possibility of running the same `action` concurrently, depending on the action time and slots available. Do we have some accounting for the fact that the function is possibly competing for resources with another thread doing the exact same code? Also, how long do these actions take? Presumably they cannot be cancelled ever. Without ordering, the `fork` here could possibly let new thread finish before an older thread, if the action takes longer than a slot.

It seems difficult to describe these semantics in terms of the async package, and that seems like a red flag. But if its intended behavior and limits can be clarified, I can at least document the code justifying it, and possibly test it.

gromakovsky:
> So my understanding is that the action has to be performed at _exactly_ the right interval per slot.

The semantics is that given action should be executed when a new slot starts. Ideally it should be done at _exactly_ that time, but there are always some inaccuracies with time, of course.
Another part of the semantics is that the action is executed in another thread. Initially it was done because it's hard to predict how much time the action can take (it can do arbitrary `IO`) and we didn't want this `onNewSlot` worker to be blocked if the action blocks once.
I see your points below that it might be bad idea to do it in a separate thread.
I can't give a general answer for all usages of `onNewSlot`. In some cases the code can account for resources competition, in some case not. In some cases it can be necessary to fork a thread (and desirable to finish a new thread even if an old one is not finished yet), in some cases not.

I will review usages of `onNewSlot` functions and will write something about it later.

Currently my intuition is that `fork` there is indeed bad. And even if it's useful in some cases, it can be achieved by putting `fork` into `action` itself, rather than having this behavior in `onNewSlotDo`.

I can already give one example where proceeding to the next action seems desirable even if the current action is not finished yet. Probably the implementation should differ though.
Each slot one node needs to create a block. So we have `onNewSlot` worker which checks whether we should create a block and if, we should, creates it and announces to the network. It can happen that this announcing takes a lot of time because it involves network. But if we are eligible to create blocks in multiple consequent slots, it's better to create them regardless of whether we have already announced the previous block.

> When using delay provided by threadDelay, it means "at least" as a minimum bound. The maximum bound of the delay could be larger, we don't know how much, because the GHC runtime might be doing other things. Normally, a loop using threadDelay suffers from a kind of "drift" where it would become out of sync with e.g. a cron job set to every minute. In this code, however, the delay is calculated by time until the next slot from the current time, so that shouldn't be a problem. Agreed?

Yes, I think this code is already resistant to this problem.

gromakovsky:
Btw, regarding very long actions, it's not a paranoiac precaution, we really had very long running actions (even those which look simple) in autumn. IIRC, even something like logging (small messages) could take more than 20 seconds (which is slot duration).
The exact reason is unknown, probably that was because we were using very cheap amazon machines and the load was very high.

gromakovsky:
I grepped for `onNewSlot` and `OnNewSlot`. There are the following usages:
1. `Pos.Security.Workers.checkForIgnoredCommitmentsWorker`.
This worker is bad, it does absolutely wrong thing and should be either rewritten and at least commented out. Anyway, the behavior won't be much different conceptually. However, I see a problem: it can potentially update a `TVar` on each tick. And if multiple actions run concurrently, the order of these updates can differ from the expected one. I suppose that's exactly what you meant by
> Do we have some accounting for the fact that the function is possibly competing for resources with another thread doing the exact same code?

Also here I don't see a reason to run a new action until the previous action finishes. If the previous action hangs, the new action will most likely hang too.
2. `Pos.Block.Worker.blkOnNewSlot`.
There are basically two actions which are run `concurrently`: `chainQualityChecker` and `blockCreator`.
2.1. For `chainQualityChecker` there doesn't seem to be any resource which multiple actions may want to acquire. Basically it just reads an `IORef` (`slogGetLastSlots`), reads data from DB (`gsIsBootstrapEra `), does some computations and then logs results and optionally sends a message to the report server.
Unlike `checkForIgnoredCommitmentsWorker`, if one action takes a lot of time, it still can be useful to run another one (for newer slot). For example, if the action for `slot1` logs warning and then hangs on `reportMisbehaviour`, it's still useful to run the action for `slot2`. Even if it hangs on `reportMisbehaviour` too, it will at least report newer chain quality. On the other hand, if `reportMisbehaviour` hangs indefinitely (which shouldn't happen because we have timeouts, but who knows), we may end up having many threads which hang. Apparently it can be solved by having a timeout at the _workers_ level (and not relying on the fact that the action will never hang indefinitely).
2.2. Regarding `blockCreator`, I've already written above why it might be useful to start the action for `slot2` before the action for `slot1` finishes. The reason is identical to the one for `chainQualityChecker`. The action can be executed partially and then hang, but this «partially» will already be useful.
Regarding resources contention, the logic is quite complex, but everything should be fine. If there are two threads which are trying to create a block for `slot1` and `slot2` concurrently (which can happen only if something takes unreasonable amount of time) and the action for `slot2` does it faster, then there will be no block for `slot1` (it will see that the newer block is already created). Also two threads can't create blocks concurrently, there is an `MVar` which allows only one thread to do it at given time (`BlkSemaphore`).
3. `Pos.Slotting.Util.logNewSlotWorker`.
This one is trivial, I don't think `fork` can be useful here, on the other hand it also shouldn't spoil anything.
4. `Pos.Ssc.GodTossing.Workers.onNewSlotSsc`.
I think it's not needed to run these actions concurrently if one of them takes too much time. The main purpose is to submit some data to the network and that's actually the last thing we do.
Regarding resources, it's quite complex, not so easy to understand whether something bad can happen quickly. I think won't delve into details, because I think we should just make it sequential. 
5. `Pos.Update.Worker.usWorkers`.
It contains two things:
5.1. `processNewSlot`. There is no sense in running it for more than one slot at given time. The ideal way is to stop all `processNewSlot`s when a new slot starts (if `processNewSlot` is still running for some reason) and run `processNewSlot` for this new slot.
5.2. `checkForUpdate`. This action must not run concurrently with itself, 100%. We don't need to download the same update concurrently. And this action can indeed take more than slot duration.
So I think `processNewSlot` and `checkForUpdate` must not be in the same action.
6. `Pos.Lrc.Worker.lrcOnNewSlotWorker`.
It only calls `lrcSingleShot` and it's safe to call it as much as possible. This function is idempotent (except that it can throw `UnknownBlocksForLrc` sometimes, but it's caught) and it takes lock, so multiple actions can't run concurrently.
7. `Pos.DHT.Workers.dumpKademliaStateWorker`.
This worker dumps some state into a file with constant name. Here we have a problem with shared resources. I think actions should be executed consequently (i. e. no `fork`).

volhovM:
@gromakovsky I do not have anything to argue with you about regarding your last two comments (telegas). 
> https://github.com/fpco/cardano-sl/blob/45f51bc473117b86a0c079b04d35d6dde9a33c6e/src/Pos/Block/Network/Logic.hs#L542
So block relaying here is executed in the main block retrieval loop. If this action takes too long we're blocked and we can't download blocks. In reality relaying won't happen while we're in recovery mode so it's not possible to block recovery at least. But in general synchronized mode we can block on this call (if something is wrong with the network timeouts) and miss new block arrival. So removing `fork` here is safe only if sending action takes less than slot duration (which should be true i hope).

gromakovsky:
I am closing it, see #1495.

PR #1166, opened 2017-07-14
[СSL-1359] Add some simple tests which use blockchain generator

gromakovsky:
Done: added some tests.
TODO: add more tests (in another PR).


test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -239,8 +260,13 @@ initBlockTestContext tp@TestParams {..} callback = do
             btcUpdateContext <- mkUpdateContext
             btcSscState <- mkSscState @SscGodTossing
             btcSlogContext <- mkSlogContext
+            btcTxpMem <- (,) <$> mkTxpLocalData <*> pure ignoreTxpMetrics
```
int-index: Why not `(,ignoreTxpMetrics) <$> mkTxpLocalData`?

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -5,87 +5,100 @@ module Test.Pos.Block.Logic.VarSpec
...
+    whenLeftM (lift $ verifyBlocksPrefix (one emptyBlock)) stopProperty
+
+verifyValidBlocks :: BlockProperty ()
+verifyValidBlocks = do
+    bpGoToArbitraryState
+    blocks <- map fst . toList <$> bpGenBlocks Nothing
+    pre (not $ null blocks)
+    let blocksToVerify =
+            case blocks of
+                [] -> error "verifyValidBlocks: impossible"
```
int-index: I don't like that this is without comment. Tracing from the code, it's impossible because we use QuickCheck's `sized` combinator to get the amount of blocks, but how do you know that `size` isn't `0`? I'd rather see a comment here that explains all of this.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -5,87 +5,100 @@ module Test.Pos.Block.Logic.VarSpec
...
-        mainBlock =
-            mkMainBlock
-                (Just tipHeader)
-                slot0
-                theSecretKey
-                Nothing
-                (emptyMainBody minBound)
-    block <- either stopProperty (pure . Right) mainBlock
-    whenLeftM (lift $ verifyBlocksPrefix (one block)) stopProperty
+    emptyBlock <- fst . unsafeHead . getOldestFirst <$> bpGenBlocks (Just 1)
```
int-index: I see why it's safe here, but I still want to see a comment that explains why it's safe. The reason is that when we use some unsafe function, we make some assumptions that make it safe, but if we don't write those assumptions in code, it's easy to modify the code in a way that breaks this assumption.

I have a rather strict rule that every invocation of an unsafe function should be accompanied with a comment that explains why in this particular case it's safe.

test/Test/Pos/Block/Logic/Mode.hs
gromakovsky: 🤷‍♂️

test/Test/Pos/Block/Logic/VarSpec.hs
gromakovsky: There is `pre (not $ null blocks)` right before this `let`.

PR #1168, opened 2017-07-15
[CSE-182] Extract utxoToAddressCoinPairs to a separate function

Pastafarianist:
This function is needed in Explorer.

### DIFF
--- a/src/Pos/Explorer/DB.hs
+++ b/src/Pos/Explorer/DB.hs
@@ -24,13 +24,13 @@ import qualified Ether
 import           Pos.Binary.Class      (UnsignedVarInt (..), encodeStrict)
 import           Pos.Context.Functions (GenesisUtxo, genesisUtxoM)
 import           Pos.Core              (unsafeAddCoin)
-import           Pos.Core.Types        (Address, Coin, HeaderHash, EpochIndex)
+import           Pos.Core.Types        (Address, Coin, EpochIndex, HeaderHash)
 import           Pos.DB                (DBTag (GStateDB), MonadDB, MonadDBRead (dbGet),
                                         RocksBatchOp (..))
 import           Pos.DB.GState.Common  (gsGetBi, gsPutBi, writeBatchGState)
 import           Pos.Explorer.Core     (AddrHistory, TxExtra (..))
 import           Pos.Txp.Core          (Tx, TxId, TxOutAux (..), _TxOut)
-import           Pos.Txp.Toil          (Utxo)
+import           Pos.Txp.Toil          (Utxo, utxoToAddressCoinPairs)
 import           Pos.Util.Chrono       (NewestFirst (..))
 
 ----------------------------------------------------------------------------
@@ -42,7 +42,7 @@ type Page = Int
 type Epoch = EpochIndex
 
 -- type PageBlocks = [Block SscGodTossing]
--- ^ this is much simpler but we are trading time for space 
+-- ^ this is much simpler but we are trading time for space
 -- (since space is an issue, it seems)
 
 -- TODO: In time if we have enough constants, maybe add to explorer Constants?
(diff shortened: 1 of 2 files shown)


txp/Pos/Txp/Toil/Utxo/Util.hs
```diff
@@ -33,3 +35,11 @@ utxoToStakes = foldl' putDistr mempty . M.toList
   where
     plusAt hm (key, val) = HM.insertWith unsafeAddCoin key val hm
     putDistr hm (_, toaux) = foldl' plusAt hm (txOutStake toaux)
+
+utxoToAddressCoinPairs :: Utxo -> [(Address, Coin)]
```
ksaric: I prefer the more verbose:
```haskell
utxoToAddressCoinPairs :: Utxo -> [(Address, Coin)]
utxoToAddressCoinPairs utxo = combineWith unsafeAddCoin txOuts
  where
    combineWith :: (Eq a, Hashable a) => (b -> b -> b) -> [(a, b)] -> [(a, b)]
    combineWith func = HM.toList . HM.fromListWith func

    txOuts :: [(Address, Coin)]
    txOuts = map utxoToaOut utxoElems
      where
        utxoToaOut :: TxOutAux -> (Address, Coin)
        utxoToaOut txOutAux = view _TxOut . toaOut $ txOutAux

        utxoElems :: [TxOutAux]
        utxoElems = M.elems utxo
```

src/Pos/Explorer/DB.hs
```diff
@@ -95,11 +95,8 @@ putInitFlag = gsPutBi balancesInitFlag True
 putGenesisBalances :: MonadDB m => Utxo -> m ()
 putGenesisBalances genesisUtxo =
```
ksaric: I prefer:
```
putGenesisBalances :: MonadDB m => Utxo -> m ()
putGenesisBalances genesisUtxo = writeBatchGState putAddrBalancaOp
  where
    putAddrBalancaOp :: [ExplorerOp]
    putAddrBalancaOp = map (uncurry PutAddrBalance) addressCoinsPairs

    addressCoinsPairs :: [(Address, Coin)]
    addressCoinsPairs = utxoToAddressCoinPairs genesisUtxo
```

Pastafarianist:
@ksaric thank you, I added all of your suggestions (slightly modified).

gromakovsky:
http://i.imgur.com/8AvB2Ty.png
This is an example of anti-pattern. It's better to rebase if your PR has only one commit. You have no benefits if you merge, because you have to resolve the same number of conflicts.
The only exception I have in mind is when you are working with other people in your branch, but I guess it's not the case.

txp/Pos/Txp/Toil/Utxo/Util.hs
gromakovsky: 1. I think this function should return `HashMap Address Coin`. It applies `HM.toList` in the end. If caller needs `HashMap`, they will apply `HM.fromList` to the result and it will lead to unnecessary conversion to and from `HashMap`. If caller needs `[(Address, Coin)]`, they can use `HM.toList` explicitly. And I can't say that the latter case will be much more popular.
2. It would be good to add a haddock comment.

Pastafarianist:
@gromakovsky in this case I merged because the Github interface said that my branch was "outdated" and `cardano-sl-0.5` needed to be merged into it before it could be merged into `cardano-sl-0.5`. I don't know under which circumstances this happens and I would be happy to find out.

txp/Pos/Txp/Toil/Utxo/Util.hs
gromakovsky: Another reason to do it is that it makes clearer to the caller that the returned value doesn't have any duplicated addresses (the function with such name could also return outputs converted to pairs which would have duplicates).

gromakovsky:
@Pastafarianist 
1. You could rebase and Github wouldn't say that your branch was outdated.
2. That's because this http://i.imgur.com/YtTxOsR.png checkmark is enabled for `cardano-sl-0.5` branch. I don't know who enabled it, I am 99% sure it wasn't me. For other branches it's not enabled. I don't know whether it's good or not. Perhaps it's good, because there shouldn't be many commits to 0.5 branch at this point.

txp/Pos/Txp/Toil/Utxo/Util.hs
ksaric: This reasoning is quite good. To inform the person looking at this code that the keys are unique is a good idea.

gromakovsky:
@Pastafarianist don't forget to make PR to master.

PR #1171, opened 2017-07-17
[CSL-1357] Block event generator

int-index:


gromakovsky:
http://i.imgur.com/btYsFh3.png
I call it anti-pattern.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+    = BlkEvApply (BlockEventApply' blund)
+    | BlkEvRollback (BlockEventRollback' blund)
+    deriving (Functor, Foldable)
+
+instance IsBlockEventFailure (BlockEvent' blund) where
+    isBlockEventFailure = \case
+        BlkEvApply    a -> isBlockEventFailure a
+        BlkEvRollback a -> isBlockEventFailure a
+
+type BlockEvent = BlockEvent' BlundDefault
```
volhovM: Why did you parametrize everything with it instead of hardcoding `SscGodTossing`?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+-- | A coefficient in the range [0,1]. Pass it to 'weighted' if you ever get
+-- the chance.
+newtype Chance = Chance {getChance :: Rational}
+
+-- | Generate a boolean that may happen to be of true value.
+byChance :: (Monad m, RandomGen g) => Chance -> RandT g m Bool
+byChance (Chance c) = weighted [(False, 1 - c), (True, c)]
+
+data BlockEventGenParams = BlockEventGenParams
+    { _begpSecrets         :: !AllSecrets
```
volhovM: Shouldn't it be the same as in `BlockGenParams`?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+genBlockEvent ::
+       (Monad m, RandomGen g)
+    => BlockEventGenParams
+    -> StateT GenBlockEventState (RandT g m) (Maybe (BlockEvent' BlockIndex))
+genBlockEvent begp = do
+    gbes <- get
+    failure <- lift $ byChance (begp ^. begpFailureChance)
+    when failure $ put GbesExpectedFailure
+    rollback <- lift $ byChance (begp ^. begpRollbackChance)
+    case gbes of
```
volhovM: Maybe you wanted to do `get >>= \case` here so `put GbesExpectedFailure` affects it?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+    * it's possible to encode an empty range as @(k, k)@
+    * to compute the amount of elements, one can use simple subtraction
+-}
+
+getBlockIndexRange ::
+       [BlockEvent' BlockIndex]
+    -> (BlockIndex, BlockIndex)
+getBlockIndexRange =
+    let minMax a = Smg.Option (Just (Smg.Min a, Smg.Max a))
+    in maybe (0, 0) (over _2 (+1)) . coerce . foldMap minMax . Compose
```
gromakovsky: It took me more than 10 minutes to understand what's going on (including parameterization of `BlockEvent'`). I don't know if it indicates that I am stupid or that the code is hard to understand. I would at least change `coerce` to `Smg.getOption` to make it at least slightly more understandable.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                blockIndices <- genBlockIndices 0
+                put (GbesBlockchain blockIndices)
+                return . BlkEvApply $ BlockEventApply
+                    { _beaInput    = blockIndices
+                    , _beaOutValid = BlockApplySuccess
+                    }
+        GbesBlockchain blockchain -> Just <$> case (failure, rollback) of
+            (True, True) -> do
+                -- Fail with rollback.
+                rft <- getRandom
```
volhovM: Save up some code space here and in similar situations:
```
(True, True) -> getRandom >>= \case 
    ...
```

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                            skip = 1
+                        blockIndices <- genBlockIndices (tip + 1 + skip)
+                        return . BlkEvApply $ BlockEventApply
+                            { _beaInput    = blockIndices
+                            , _beaOutValid = BlockApplyFailure
+                            }
+            (False, True) -> do
+                -- Success with rollback.
+                len <- getRandomR (1, length blockchain)
+                let
```
volhovM: Why leave the newline here 🤔

src/Pos/Generator/BlockEvent.hs
gromakovsky: To make it `Functor`.

gromakovsky: For instance, `genBlockEvents'` generates indices, not actual blocks.

src/Pos/Generator/BlockEvent.hs
volhovM: Yes, `coerce` here requires some time to think because you can't percieve it as a hint.

gromakovsky: Each function call here requires some time to think 😉 
Coerce is the one for which it's easy to reduce this time :)

Actually now I think that it took me more than 10 minutes because I was stupid and have never used this stuff before.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                return . BlkEvApply $ BlockEventApply
+                    { _beaInput = blockIndices
+                    , _beaOutValid = BlockApplySuccess
+                    }
+  where
+    genBlockIndices blockIndexStart = do
+        let blockIndexEndMax = blockIndexStart +
+                               fromIntegral (begp ^. begpBlockCountMax)
+        blockIndexEnd <- getRandomR (blockIndexStart,blockIndexEndMax)
+        return $ OldestFirst . NE.fromList $ [blockIndexStart..blockIndexEnd]
```
gromakovsky: Shouldn't it be `blockIndexEnd - 1` instead?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                -- invalid by itself.
+                blockIndices <- genBlockIndices 1
+                let
+                    blkZeroth     = BlockIndex 0 :| []
+                    blockIndices' =
+                        -- Those block indices are broken by construction
+                        -- because we append the 0-th block to the end.
+                        -- Sometimes we can violate the maximum bound on
+                        -- block count, but this is fine. See NOTE
+                        -- on reordering corner cases.
```
gromakovsky: I don't fully understand what these comments mean (I see a problem, then I see that the solution is proposed, but it has a drawback too and it's hard to understand what is done eventually), but do I understand correctly that we can generate more than `begpBlockCountMax` blocks? Even by 2.
1. `blockIndices` has at most `begpBlockCountMax` indices (see my comment to `genBlockIndices`).
2. We also add the 0-th block.

Update: I read this comment again and see this:
```
Sometimes we can violate the maximum bound on
                       -- block count, but this is fine.
```
Then I have another question: why is it fine?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                        let
+                            tip  = NE.last (getOldestFirst blockchain)
+                            -- The amount of blocks to skip. One is enough.
+                            skip = 1
+                        blockIndices <- genBlockIndices (tip + 1 + skip)
+                        return . BlkEvApply $ BlockEventApply
+                            { _beaInput    = blockIndices
+                            , _beaOutValid = BlockApplyFailure
+                            }
+            (False, True) -> do
```
gromakovsky: Why don't we update the state in this case?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                let
+                    -- 'NE.fromList' is valid here because 'len >= 1'.
+                    select = NE.fromList . NE.take len
+                    blockIndices = toNewestFirst blockchain &
+                        over _NewestFirst select
+                return . BlkEvRollback $ BlockEventRollback
+                    { _berInput    = blockIndices
+                    , _berOutValid = BlockRollbackSuccess
+                    }
+            (False, False) -> do
```
gromakovsky: Why don't we update the state in this case?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+
+When reordering blocks to get an invalid sequence, there's a corner case:
+if there's only one block, it's impossible to get an invalid ordering.
+The reason one might have only one block depends on the means of generation,
+but for 'genBlockIndices' there are two possible causes:
+
+* we rolled 'blockIndexEnd' equal to 'blockIndexStart'
+* 'blockIndexMax' is 1
+
+To avoid dealing with this, we may append one more block. The downside is that
```
gromakovsky: _we may append one more block_ sounds confusing for me. How should I treat it? It sounds like this is an open question. But apparently that's what actually done.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                    , _beaOutValid = BlockApplySuccess
+                    }
+        GbesBlockchain blockchain -> Just <$> case (failure, rollback) of
+            (True, True) -> do
+                -- Fail with rollback.
+                rft <- getRandom
+                case rft of
+                    RftRollbackExcess -> do
+                        -- Attempt to rollback the entire blockchain and then
+                        -- some more.
```
gromakovsky: What does this failure mean exactly?
Suppose we have blockchain `[0, 1, 2]`. We may generate rollback `[2, 1, 0, -1]`. In the end we will generate 4 blocks. Okay, I see, rollback of the `-1`-st block will be unsuccessful. Can you add this example here to make it clearer for readers?

src/Pos/Generator/BlockEvent.hs
gromakovsky: I _think_ no, but I am not sure too. If this code is valid, can you add a comment about it (because it indeed looks suspicious).

int-index: It's on purpose, I don't want `put GbesExpectedFailure` to affect it. I'll add a comment.

core/Pos/Util/Chrono.hs
```diff
@@ -37,7 +39,10 @@ newtype OldestFirst f a = OldestFirst {getOldestFirst :: f a}
             Binary, Bi,
             Arbitrary)
 
+makePrisms  ''NewestFirst
 makeWrapped ''NewestFirst
+
+makePrisms  ''OldestFirst
```
int-index: It's more clear than `_Wrapped`, you get to see the name of the constructor at call site.

src/Pos/Generator/BlockEvent.hs
int-index: I'll try to simplify it.

src/Pos/Generator/BlockEvent.hs
int-index: Why? We want at least one block index, otherwise `NE.fromList` won't be safe.

src/Pos/Generator/BlockEvent.hs
int-index: I forgot :(

src/Pos/Generator/BlockEvent.hs
int-index: I forgot :(

src/Pos/Generator/BlockEvent.hs
int-index: Yes, that's what I had in mind. I'll add the example.

src/Pos/Generator/BlockEvent.hs
gromakovsky: For example, few lines above you use `genBlockIndices` and put the result into `blockIndices`. And `genBlockIndices` can generate `begpBlockCountMax + 1` indices. But we want at most `begpBlockCountMax` blocks.

gromakovsky: For me it seems more natural to generate at most `begpBlockCountMax` indices here.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +91,334 @@ data BlockRollbackResult
...
+                    blockIndices = toNewestFirst blockchain &
+                        over _NewestFirst select
+                return . BlkEvRollback $ BlockEventRollback
+                    { _berInput    = blockIndices
+                    , _berOutValid = BlockRollbackSuccess
+                    }
+            (False, False) -> do
+                -- Success without rollback (with apply).
+                let tip = NE.last (getOldestFirst blockchain)
+                blockIndices <- genBlockIndices (tip + 1)
```
gromakovsky: Also I don't understand how you ensure that total number of blocks can't exceed `begpBlockCountMax`. For instance, here we may already have some blocks. And then `genBlockIndices` can generate `begpBlockCountMax + 1` indices. And all these indices will be in the result.
Will we trim them somewhere?

int-index: `begpBlockCountMax` is the maximum amount of blocks per apply event.

gromakovsky: Lol, that's why you should document fields even if you think the semantics is obvious 😉

src/Pos/Generator/BlockEvent.hs
int-index: It is the same. I pass it to `genBlocks` without modification.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +94,388 @@ data BlockRollbackResult
...
+            -- blocks for the lookup to succeed.
+            getOldestFirst blocks !! toZeroBased i
+    return $ fmap getBlock <$> preBlockEvents
+
+data GenBlockEventState
+    = GbesStartingState
+    | GbesBlockchain (OldestFirst NE BlockIndex)
+    | GbesExpectedFailure
+
+-- | A worker for 'genBlockEvents' that generates event with block indices
```
gromakovsky: Minor comments:
• usually we call `worker` something that is launched at the beginning and is periodically executed, usually in a separate thread;
• usually we use regular comments for not exported names.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -41,13 +94,388 @@ data BlockRollbackResult
...
+byChance :: (Monad m, RandomGen g) => Chance -> RandT g m Bool
+byChance (Chance c) = weighted [(False, 1 - c), (True, c)]
+
+data BlockEventGenParams = BlockEventGenParams
+    { _begpSecrets         :: !AllSecrets
+    , _begpBlockCountMax   :: !BlockCount {- ^ the maximum possible amount of
+    blocks in a BlockApply event. Must be 1 or more. Can be violated by 1 in
+    some cases (see Note on reordering corner cases), so if you specify '52'
+    here, some rare events may contain up to '53' blocks. -}
+    , _begpBlockEventCount :: !BlockEventCount
```
gromakovsky: Can you comment this field too? AFAIU, it's the **maximal** number of events to generate. The actual number of events can be smaller. I think it's worth a comment, because otherwise someone can think that it **always** generates as many events.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -168,8 +174,7 @@ getBlockIndexRange ::
        [BlockEvent' BlockIndex]
     -> (BlockIndex, BlockIndex)
 getBlockIndexRange =
-    let minMax a = Smg.Option (Just (Smg.Min a, Smg.Max a))
-    in maybe (0, 0) (over _2 (+1)) . coerce . foldMap minMax . Compose
+    maybe (0, 0) (\(lower, upper) -> (lower, upper+1)) . minMaxOf (folded.folded)
```
volhovM: So this is supposed to be the simpler version I guess? 😄

src/Pos/Generator/BlockEvent.hs
volhovM: plus for (1), i don't care for (2).

src/Pos/Generator/BlockEvent.hs
```diff
@@ -252,160 +259,210 @@ instance Random ApplyFailureType where
...
+        -- 'NE.fromList' assumes that 'len >= 1', which holds because we require
+        -- 'begpBlockCountMax >= 1'.
+        return $ OldestFirst . NE.fromList . take len $ [blockIndexStart..]
+
+    -- Fail with rollback. In the starting state it's easy,
+    -- because any rollback will fail when we don't have any
+    -- blocks yet.
+    genBlockStartingState (IsFailure True) (IsRollback True) = do
+        blockIndices <- genBlockIndices 0
+        let
```
volhovM: so sad

src/Pos/Generator/BlockEvent.hs
int-index: Yes, ha ha. I asked @neongreen how he would write it and he recommended lenses (unsurprisingly).

src/Pos/Generator/BlockEvent.hs
int-index: I meant "worker" in a "worker/wrapper" sense here, but I will use a different word.

gromakovsky:
@int-index do something (push empty commit or amend and force push and whatever else) to trigger CI.

PR #1176, opened 2017-07-17
[CSL-1359] Fix 'applyByOneOrAllAtOnce' test

gromakovsky:
This PR fixes `applyByOneOrAllAtOnce` test. The version in `master` is broken and does wrong thing. There are two problems:
1. The GState was cloned only partially: only DB was cloned, but in-memory data which replicates GState (e. g. `LrcContext`) wasn't cloned.
2. It should apply blocks all at once or one-by-one, but it only applies all at once.

Both problems have been fixed. I want it to be merged ASAP, because broken tests are bad. For some reason tests are broken only in dev mode, while in travis mode they pass (though it may be by accident).


core/Pos/Util/Util.hs
```diff
@@ -31,7 +31,9 @@ module Pos.Util.Util
        , ether
        , Ether.TaggedTrans
        , HasLens(..)
+       , HasLens'
```
pva701: If I recall correctly you hated `'` in global names, did the system break you?

gromakovsky: 1. I didn't hate it completely, only in many cases (but not all).
2. It's consistent with the naming used in `ether` which I am already used to.

PR #1182, opened 2017-07-18
[CSM-320] Regenerate genesis

georgeee:


gromakovsky:
Why is it needed btw?

georgeee:
@gromakovsky AFAIK, current master fails to start in prod mode

gromakovsky:
```
cardano-node: Failed to read genesis: Attempted to read too many bytes for StaticSize 28 Data.ByteString.ByteString. Needed 28, but only 22 remain.
GenesisCoreData
  error, called at src/Debug.hs:43:11 in universum-0.4.3-CkMiNAiq0KBB5vO0r3nIZM:Debug
```
Okay, indeed.

gromakovsky:
It works!

PR #1183, opened 2017-07-18
[DEVOPS-204] Require MacOS builds again

domenkozar:
The builds should not be timeouting anymore.

There are still going to be some transient **errors due to SSL**, those are fixed by restarting the build. The proper fix will be done as DEVOPS-203 is implement, which is expected to be this week.

cc @input-output-hk/serokell

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ matrix:
       # don't use sierra (xcode 8.1+) because of too restrictive linker path (DEVOPS-168)
 
   allow_failures:
-    - os: osx
     - env: MODE=nix
     # https://github.com/input-output-hk/stack2nix/issues/15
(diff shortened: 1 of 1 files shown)


gromakovsky:
@domenkozar it's funny, but MacOS build failed with this PR already :)
And it doesn't seem to be SSL error.
> error: file ‘nixpkgs’ was not found in the Nix search path (add it using $NIX_PATH or -I), at (string):1:13
> Sandbox: launchd(1) System Policy: deny(1) file-write-unlink /private/var/run/dyld_shared_cache_x86_64.map
> Sandbox: CommCenter(231) deny(1) file-read-metadata /private/var/folders

domenkozar:
Blocked by https://github.com/NixOS/nixpkgs/issues/27493

PR #1184, opened 2017-07-18
[CSL-1361] Blockgen tx payload

volhovM:
This PR adds transaction payload capabilities to the block generator. Most important code is contained in `Pos.Generator.Block.Payload`. Generation is parametrized by `TxGenParams`. 
Also arbitrary instances were moved: for every `X.Arbitrary` module it was renamed into `Arbitrary.X`. It simplifies grouping and was discussed with gromakovsky.


src/Pos/Generator/Block/Logic.hs
```diff
@@ -36,17 +42,18 @@ import           Pos.Util.Util             (maybeThrow, _neHead)
 -- valid with respect to the global state right before this function
 -- call.
 genBlocks ::
-       MonadBlockGen ctx m
+       (MonadBlockGen ctx m, RandomGen g)
     => BlockGenParams
+    -> g
```
int-index: Export it with `RandT`, not with `g` as a simple function argument. The reason is that if you're abstracting over `g`, you can't assume that it's read-only, it needs to be passed in a `StateT`-like manner. Therefore you'll need to return `(OldestFirst [] (Blund SscGodTossing), g)`. But it's not as convenient as `RandT` because then the caller doesn't have a choice between `evalRandT` and `runRandT`).

For instance, in my code I will need to call `genBlocks` in the `RandT` monad directly, and I don't want unnecessary wrapping/unwrapping.

src/Pos/Generator/Block/Payload.hs
```diff
@@ -0,0 +1,186 @@
...
+import qualified Pos.Txp.Toil.Utxo          as Utxo
+
+----------------------------------------------------------------------------
+-- Tx payload generation
+----------------------------------------------------------------------------
+
+-- | Generates list of distinct ints in the given range [a,b].
+selectDistinct :: (MonadRandom m, MonadIO m) => Int -> (Int,Int) -> m [Int]
+selectDistinct n0 p@(a, b)
+    | b - a < 0 = error $ "selectDistinct: b < a " <> show p
```
int-index: Why do you write `b - a < 0` instead of `b < a`?

src/Pos/Generator/Block/Payload.hs
```diff
@@ -0,0 +1,186 @@
...
+genTxPayload = do
+    utxo <- lift DB.getAllPotentiallyHugeUtxo
+    let gtd = GenTxData utxo (V.fromList $ M.keys utxo)
+    flip evalStateT gtd $ do
+        (a,d) <- lift $ view tgpTxCountRange
+        txsN <- fromIntegral <$> getRandomR (a, a + d)
+        void $ replicateM txsN genTransaction
+  where
+    genTransaction :: StateT GenTxData (BlockGenRandMode g m) ()
+    genTransaction = do
```
int-index: In this function alone I counted 8 `error` invocations and at least 5 unsafe functions (`V.unsafeIndex`, 3 times `NE.fromList`, and `unsafeIntegerToCoin`).

The way I'm used to read these unsafe things is that they are not meant to go wrong. I.e. if you call `error`, it's a sanity check (or tricking the type system) and it's never meant to happen at run-time. If you use `unsafeIndex`, you expect it to give you the value because you know that the index is valid.

However, I see nearly no comments as to why these `error` calls don't happen or why unsafe functions are safe in this context. It's hard to reason about code written this way.

To be more clear, I just want a short comment before each `error` or `unsafe-*` function, and in each comment a list of the assumptions which make it safe.

test/Test/Pos/Block/Logic/Emulation.hs
```diff
@@ -47,7 +47,11 @@ instance MonadIO Emulation where
     liftIO m = Emulation . liftIO $ do
         -- if you see a lot of stars in the test log, a hunting season for
         -- not-mocked operations is open.
-        putStr ("*" :: String)
+        --
+        -- It's currently commented because stars take too much buffer space
```
int-index: It's okay, but I'm sad anyway.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -59,13 +59,13 @@ verifyEmptyMainBlock :: BlockProperty ()
...
     -- generate exactly 1 block
-    emptyBlock <- fst . unsafeHead . getOldestFirst <$> bpGenBlocks (Just 1)
+    emptyBlock <- fst . unsafeHead . getOldestFirst <$> bpGenBlocks (Just 1) False
     whenLeftM (lift $ verifyBlocksPrefix (one emptyBlock)) stopProperty
 
 verifyValidBlocks :: BlockProperty ()
 verifyValidBlocks = do
     bpGoToArbitraryState
-    blocks <- map fst . toList <$> bpGenBlocks Nothing
+    blocks <- map fst . toList <$> bpGenBlocks Nothing True
```
int-index: It's not possible from this function call to understand what `True` means here. On the other hand, with a newtype it would look like this:

```
bpGenBlocks Nothing (EnableTxPayload True)
```

Huge readability gain in my opinion.

txp/Pos/Txp/Logic/Local.hs
```diff
@@ -48,18 +48,16 @@ type TxpLocalWorkMode ctx m =
...
     :: TxpLocalWorkMode ctx m
     => (TxId, TxAux) -> ExceptT ToilVerFailure m ()
 txProcessTransaction itw@(txId, txAux) = do
     let UnsafeTx {..} = taTx txAux
     tipDB <- GS.getTip
-    epoch <- siEpoch <$>
-        -- TODO: Don't use inaccurate slotting here. If we don't know the
-        -- current slot, we should reject transactions. See CSL-1341.
-        getCurrentSlotInaccurate
+    epoch <- maybe (throwError ToilSlotUnknown) (pure . siEpoch) =<< getCurrentSlot
```
int-index: this fix probably deserves a PR of its own, no?

src/Pos/Generator/Block/Payload.hs
volhovM: More intuitive to me :shrug:

src/Pos/Generator/Block/Payload.hs
volhovM: No problem, I'll add comments.

txp/Pos/Txp/Logic/Local.hs
volhovM: Too complicated, i think this will do. @gromakovsky ?

test/Test/Pos/Block/Logic/VarSpec.hs
volhovM: Wrapping every bool value into newtype is a tradeoff i would better avoid. Having lots of constructors + extra unpacking -- better go to definition and read the comment.

src/Pos/Generator/Block/Logic.hs
```diff
@@ -36,17 +42,17 @@ import           Pos.Util.Util             (maybeThrow, _neHead)
...
-       MonadBlockGen ctx m
+       (MonadBlockGen ctx m, RandomGen g)
     => BlockGenParams
-    -> m (OldestFirst [] (Blund SscGodTossing))
+    -> RandT g m (OldestFirst [] (Blund SscGodTossing))
 genBlocks params = do
-    ctx <- mkBlockGenContext params
-    runReaderT genBlocksDo ctx
+    ctx <- lift $ mkBlockGenContext params
+    mapRandT (`runReaderT` ctx) genBlocksDo
```
gromakovsky: There is `usingReaderT` in `Universum`, btw.

src/Pos/Generator/Block/Mode.hs
```diff
@@ -117,6 +125,7 @@ data BlockGenContext = BlockGenContext
     , bgcSystemStart       :: !Timestamp
     , bgcParams            :: !BlockGenParams
     , bgcDelegation        :: !DelegationVar
+    , bgcCustomUtxo        :: !Utxo
```
gromakovsky: Also, add a comment that it's needed here only to know bootstrap stakeholders.

src/Pos/Generator/Block/Mode.hs
```diff
@@ -165,6 +177,14 @@ mkBlockGenContext bgcParams = do
...
         bgcDelegation <- mkDelegationVar @SscGodTossing
         return BlockGenContext {..}
+  where
+    bgcCustomUtxo =
+        let addrs = map (makePubKeyAddress . toPublic) . toList $
+                    view (bgpSecrets . asSecretKeys) bgcParams
+            utxoTxHash = unsafeHash ("randomutxotx" :: Text)
+            txIns = map (TxIn utxoTxHash) [0..fromIntegral (length addrs) - 1]
+            txOuts = map (\addr -> TxOutAux (TxOut addr (mkCoin 10000)) []) addrs
+        in M.fromList $ txIns `zip` txOuts
```
gromakovsky: 1. Wtf is that? Why do you create utxo here? Block generator definitely shouldn't define genesis utxo by itself. It doesn't know which utxo was genesis! It should come from the outside!
2. Why does it have empty distribution? It doesn't make sense.
3. Actually I would remove genesis utxo and store only genesis stakeholders (and change constraints is some places to make it possible).
4. It seems that here you can just use current utxo and set it as genesis utxo. If we are in bootstrap era, it's guaranteed that all stake always belongs to all genesis stakeholders and only them. If we are not in bootstrap era, we don't care about genesis stakeholders. 

Anyway, the ideal solution for me is to change `HasLens GenesisUtxo` to `HasLens GenesisStakeholders` in some places, store only `GenesisStakeholders` in `BlockGenContext` and pass it here through `MonadBlockGen`.

src/Pos/Generator/Block/Mode.hs
```diff
@@ -317,7 +343,11 @@ instance MonadBlockGenBase m => MonadSlots (BlockGenMode m) where
     currentTimeSlotting = currentTimeSlottingSimple
 
 instance MonadBlockGenBase m => DB.MonadGState (BlockGenMode m) where
-    gsAdoptedBVData = gsAdoptedBVDataDefault
+    gsAdoptedBVData =
+        -- Currently we can't and don't want to test the system in boot mode
+        -- since in requires transaction to have specific distribution.
+        gsAdoptedBVDataDefault <&>
+        (\BlockVersionData{..} -> BlockVersionData { bvdUnlockStakeEpoch = 0, ..})
```
gromakovsky: 1. It's very bad, because it differs from the data we store in DB and leads to inconsistent behavior and invariants violation.
2. Can we stop this cancer already? I can't stand it anymore. It's July 19 already. Why do we disable bootstrap era when something doesn't work and pretend it's a proper fix? When will we test bootstrap era? We must consider bootstrap era in our tests. It's even more important than past-bootstrap era.
Instead of writing proper code we add more and more hacks and pretend it's fine. We will not have time to rewrite it. We need to do it properly from the beginning.

> we can't

we can

> don't want

we want

src/Pos/Generator/Block/Param.hs
```diff
@@ -39,16 +42,46 @@ instance Buildable AllSecrets where
             (length _asSecretKeys)
             (HM.keys _asSecretKeys)
 
+-- | Parameters for transactions payload generation.
+data TxGenParams = TxGenParams
```
gromakovsky: Why is it not in `txp`?

src/Pos/Generator/Block/Param.hs
```diff
@@ -39,16 +42,46 @@ instance Buildable AllSecrets where
...
+    { _tgpTxCountRange :: !(Word, Word)
+    -- ^ Such (a, d), that there block will include some x ∈ [a, a+d)
+    -- transactions. Set to (y,y) to disable tx generation.
```
gromakovsky: Hm, `(y, y)` means `[y, y+y)` according to the comment which is not empty as long as `y > 0`. Seems like there is a mistake in this comment.

src/Pos/Generator/Block/Param.hs
```diff
@@ -39,16 +42,46 @@ instance Buildable AllSecrets where
...
+    -- ^ Maximum number of tx outputs.
+    }
+
+makeClassy ''TxGenParams
+
+instance Buildable TxGenParams where
+    build TxGenParams {..} = do
+        let (a,d) = _tgpTxCountRange
+        bprint ("TxGenParams {\n"%
+                "  tx count (from,to): "%pairF%"\n"%
```
gromakovsky: Shouldn't it be `[from, to)`? Otherwise it can lead to confusion.

src/Pos/Generator/Block/Payload.hs
```diff
@@ -0,0 +1,199 @@
...
+makeLenses ''GenTxData
+
+instance (Monad m) => MonadUtxoRead (StateT GenTxData m) where
+    utxoGet txIn = uses gtdUtxo $ M.lookup txIn
+
+instance (Monad m) => MonadUtxo (StateT GenTxData m) where
+    utxoPut txIn txOutAux = gtdUtxo . at txIn ?= txOutAux
+    utxoDel txIn = gtdUtxo . at txIn .= Nothing
+
+-- TODO: move to txp, think how to unite it with 'Pos.Arbitrary.Txp'.
```
gromakovsky: Is it hard?

txp/Pos/Txp/Logic/Local.hs
gromakovsky: Well, it's only a small part of CSL-1341 and it was necessary for this task.
Also it's bad that we talk in private messages.

src/Pos/Generator/Block/Payload.hs
gromakovsky: Also `MonadBlockGenBase` contains `MonadThrow` constraint, so if you can't prove that some unsafe function is always safe, you'd better use `throwM`.

src/Pos/Generator/Block/Payload.hs
```diff
@@ -0,0 +1,199 @@
...
+        when (utxoSize == 0) $
+            -- Empty utxo means misconfigured testing settings most
+            -- probably, so that's a critical error
+            error "Utxo is empty already which is weird"
+
+        secrets <- view asSecretKeys <$> view blockGenParams
+        -- Unsafe hashmap resolving is used here because we suppose
+        -- utxo contains only related to these secret keys only.
+        let resolveSecret stId =
+                fromMaybe (error $ "can't find stakeholderId " <>
```
gromakovsky: For example, this `error` **can** happen if `genTxPayload` is called with wrong context. It should use `throwM` instead.

src/Pos/Generator/Block/Payload.hs
```diff
@@ -0,0 +1,199 @@
...
+        let taWitness =
+                V.fromList $ map (uncurry toWitness . second txOutAddress)
+                                 (txIns `zip` inputsResolved)
+        let txAux = TxAux { taTx = tx, .. }
+        res <- lift . lift $ runExceptT $ txProcessTransaction (txId, txAux)
+        case res of
+            Left e  -> error $ "genTransaction@txProcessTransaction: got left: " <> pretty e
+            Right () -> do
+                Utxo.applyTxToUtxo (WithHash tx txId) taDistribution
+                gtdUtxoKeys %= V.filter (`notElem` txIns)
```
gromakovsky: It's more efficient to use `HashSet` here.

src/Pos/Generator/Block/Payload.hs
```diff
@@ -0,0 +1,199 @@
...
+                            sign SignTxIn sk TxSigData { txSigInput = txIn
+                                                       , txSigOutsHash = hash txOuts
+                                                       , txSigDistrHash = hash taDistribution }
+                    in PkWitness (toPublic sk) txSig
+                other -> error $
+                    sformat ("Found non-pubkey address: "%addressDetailedF) other
+        let taWitness =
+                V.fromList $ map (uncurry toWitness . second txOutAddress)
+                                 (txIns `zip` inputsResolved)
+        let txAux = TxAux { taTx = tx, .. }
```
gromakovsky: I don't like that you create tx fully by yourself. It's better to reuse something from `Pos.Client` or maybe add something new there.
Currently it doesn't account for tx fee checks and bootstrap era.
`Pos.Client` also doesn't account for it, but that's another problem. You can pretend that it takes these rules into account for now.

txp/Pos/Txp/Logic/Local.hs
pva701: stop it @volhovM, discuss at the #general plz

volhovM: I think it's more than appropriate to discuss this kinds of changes in private messages.

pva701: you discuss *whole* things in private messages.
I want to see decisions about some design solutions, instead of seeing them when they are in master

src/Pos/Generator/Block/Mode.hs
volhovM: Alright, i thought this PR should be merged ASAP but it can be WIP for one more day if you want to. I'll implement boot era payload corrections.

src/Pos/Generator/Block/Param.hs
volhovM: Why would it be it `txp`? There's not much code and it's easier to read it when it's all in one place. I do not see any reason to move it outside for now.

src/Pos/Generator/Block/Param.hs
volhovM: Ok, yes, I meant `(x,0)`.

src/Pos/Generator/Block/Payload.hs
volhovM: I don't really want it, but left the comment in case somebody thinks otherwise.

src/Pos/Generator/Block/Payload.hs
volhovM: I consider tests as an executable that runs only once. Semantics for `error` is "something is misconfigured and we should fail". If tests are misconfigured and exit because of error -- it's good. There's no reason to catch this exception and process it (hspec does it anyway).

src/Pos/Generator/Block/Payload.hs
volhovM: size of `txIns` is small, so it doesn't matter.

txp/Pos/Txp/Logic/Local.hs
volhovM: Yes, but that's your subjective point only. I don't like to see tons of unrelated spam (issues that i don't assigned/related to at all) in general channels. PR is a good place to place the concise description of design decisions and features descriptions in IMO.

pva701: >Yes, but that's your subjective point only.

not only me, gromak's too

>PR is a good place to place the concise description of design decisions and features descriptions in IMO.

lol, it's a good place for *decided* decision, I would like to see related discussion, see cons and pros of approaches and maybe participate in them.

you can scroll through unrelated discussion or participate in discussions you want

pva701: public chat is a good place to see where the team is going to and issues teammates are working on.
maybe it's not related to you, but you can see what is going on

src/Pos/Generator/Block/Mode.hs
volhovM: It's copied from node params. Was called `customUtxo` there as well.

gromakovsky: Yes, but it should be called `npGenesisUtxo` there. It had to be renamed in November, but nobody had time since then. Let's stop copying nonsensical stuff.

txp/Pos/Txp/Logic/Local.hs
pva701: maybe it's not important for you, but important for managers (gromak, georgee, and I wonder because don't want to go out context). also, it might be unimportant for you for now, but may be important in the future when you take a related task, and if a related discussion is in the private chats you aren't able to read them

gromakovsky: > I don't like to see tons of unrelated spam (issues that i don't assigned/related to at all) in general channels. 

But nobody reads these messages anyway (including you).

src/Pos/Generator/Block/Mode.hs
gromakovsky: > Alright, i thought this PR should be merged ASAP but it can be WIP for one more day if you want to. I'll implement boot era payload corrections.

Yes, go ahead. It can wait now, because I am going to work on performance issues, Vlad is going to implement events executor probably, etc. It seems nobody is blocked.
Also I already allowed 4 things which make me sad and I don't want to allow more until at least one of those 4 is resolved, namely:
1. No bootstrap era in dev mode.
2. Bootstrap era logic in tx construction is in web server.
3. Fee logic in tx construction is in web server.
4. Bootstrap era checks are not in Toil.

src/Pos/Generator/Block/Param.hs
gromakovsky: Because transaction payload generation is `txp` thing. But I see there is a TODO about moving tx payload generator to txp and I think it's ok to postpone it.

src/Pos/Generator/Block/Payload.hs
gromakovsky: Let's postpone.

src/Pos/Generator/Block/Payload.hs
gromakovsky: But generator is not only for tests.

pva701:
@volhovM can you wait for my review plz?
I am going to review today

gromakovsky:
I dissed him pretty well, most likely it will take more than this day to make me approve this PR :peka:

core/Pos/Core/Coin.hs
```diff
@@ -47,7 +47,8 @@ coinToInteger = toInteger . unsafeGetCoin
 unsafeAddCoin :: Coin -> Coin -> Coin
 unsafeAddCoin (unsafeGetCoin -> a) (unsafeGetCoin -> b)
     | res >= a && res >= b && res <= unsafeGetCoin (maxBound @Coin) = mkCoin res
-    | otherwise = error "unsafeAddCoin: overflow"
+    | otherwise =
+      error $ "unsafeAddCoin: overflow when summing " <> show a <> " + " <> show b
```
pva701: tfw neosislil `sformat`

src/Pos/Client/Txp/Util.hs
```diff
@@ -86,7 +106,40 @@ makeAbstractTx mkWit txInputs outputs =
...
+                        " era and can't be distributed among genStakeholders") c
+          -- TODO CSL-1351 boot stakeholders' weights are not used
+          pure $ genesisSplitBoot (HM.fromList $ map (,1) genStakeholders) c
+
+-- | Same as 'overrideTxOutDistrBoot' but changes 'TxOutputs' all at once
+overrideTxDistrBoot ::
+       (TxCreateMode ctx m) => TxOutputs -> ExceptT Text m TxOutputs
+overrideTxDistrBoot outputs = do
+    forM outputs $ \TxOutAux{..} -> do
+        newStakeDistr <- overrideTxOutDistrBoot  (txOutValue toaOut) toaDistr
```
pva701: two spaces after `overrideTxOutDistrBoot`

src/Pos/Generator/Block/Logic.hs
```diff
@@ -55,35 +61,36 @@ genBlocks params = do
 -- Generate a valid 'Block' for the given epoch or slot (genesis block
 -- in the former case and main block the latter case) and apply it.
 genBlock ::
-       (MonadBlockGenBase m)
+       forall ctx m g. (RandomGen g, MonadBlockGen ctx m)
     => EpochOrSlot
-    -> BlockGenMode m (Blund SscGodTossing)
+    -> BlockGenRandMode g m (Blund SscGodTossing)
```
pva701: are you sure it's necessary to reflect fact that generator is random in the name `BlockGenRandMode`? it seems to be more or less obvious

pva701: nvm, I see `BlockGenRandMode = RandT BlockGenMode`

src/Pos/Generator/Block/Mode.hs
```diff
@@ -167,6 +180,18 @@ mkBlockGenContext bgcParams@BlockGenParams{..} = do
         bgcTxpMem <- (,ignoreTxpMetrics) <$> mkTxpLocalData
         bgcDelegation <- mkDelegationVar @SscGodTossing
         return BlockGenContext {..}
+  where
+    -- Genesis utxo is needed only for boot era stakeholders
+    bgcGenesisUtxo =
```
pva701: this code looks like bulshit because of underlying db in `GStateContext` is initialized by other real `GenesisUtxo`, but you use some arbitrary `GenesisUtxo`.

Maybe I don't understand anything or you should take `GenesisUtxo` as argument of function

src/Pos/Generator/Block/Param.hs
```diff
@@ -39,17 +42,48 @@ instance Buildable AllSecrets where
             (length _asSecretKeys)
             (HM.keys _asSecretKeys)
 
+-- | Parameters for transactions payload generation.
+data TxGenParams = TxGenParams
+    { _tgpTxCountRange :: !(Word, Word)
```
pva701: `range` sound counterintuitive because there are left bound and length

src/Pos/Launcher/Param.hs
```diff
@@ -61,7 +61,7 @@ data NodeParams = NodeParams
     , npSecretKey      :: !SecretKey            -- ^ Primary secret key of node
     , npUserSecret     :: !UserSecret           -- ^ All node secret keys
     , npBaseParams     :: !BaseParams           -- ^ See 'BaseParams'
-    , npCustomUtxo     :: !Utxo                 -- ^ Predefined genesis utxo
+    , npGenesisUtxo    :: !GenesisUtxo                 -- ^ Predefined genesis utxo
```
pva701: Format the comment plz

src/Pos/Generator/Block/Mode.hs
volhovM: See the comment. The `GenesisUtxo` used in `TestParams` shouldn't have the "genesis" prefix. It is a utxo snapshot corresponding to the beginning of block generation. 
`bgcGenesisUtxo` is _really genesis_. But it's used only for boot era stakeholders as it is written in the comment above.

src/Pos/Generator/Block/Param.hs
volhovM: I agree. Have a better suggestion? :\

src/Pos/Launcher/Param.hs
volhovM: Sure.

src/Pos/Generator/Block/Param.hs
pva701: hz, maybe take left and right bound, it's minor

src/Pos/Generator/Block/Payload.hs
```diff
@@ -0,0 +1,235 @@
...
+          res <- selectDistinct' [] n
+          if fromIntegral (length res) /= n
+              then error "selectDistinct is broken"
+              else pure res
+  where
+    n :: Int
+    n = min (b + 1 - a) n0
+    selectDistinct' :: [Int] -> Int -> m [Int]
+    selectDistinct' cur leftN = do
+        leftItems <- take leftN <$> getRandomRs (a, b)
```
pva701: 1. what does `getRandomRs` do?
2. which range `(a, b)` can be passed? how big?
3. it looks like very inefficient, we will be surprised afterwards why tests work slowly.

src/Pos/Generator/Block/Mode.hs
gromakovsky: Wait-wait-wait, `GenesisUtxo` in `TestParams` is really genesis. But at this point we can have different Utxo in DB! That's why we need bootstrap stakeholders.

src/Pos/Generator/Block/Payload.hs
volhovM: 1. Read documentation pls.
2. It doesn't matter, list is lazy.
3. It's totally alright, i don't understand your predjawa.

src/Pos/Generator/Block/Mode.hs
```diff
@@ -167,6 +180,18 @@ mkBlockGenContext bgcParams@BlockGenParams{..} = do
...
+        let addrs =
+                -- So we take three stakeholders in boot era.
+                take 3 $
+                map (makePubKeyAddress . toPublic) . toList $
+                view (bgpSecrets . asSecretKeys) bgcParams
+            utxoTxHash = unsafeHash ("randomutxotx" :: Text)
+            txIns = map (TxIn utxoTxHash) [0..fromIntegral (length addrs) - 1]
+            txOuts = map (\addr -> TxOutAux (TxOut addr (mkCoin 10000)) []) addrs
+        in GenesisUtxo $ M.fromList $ txIns `zip` txOuts
```
gromakovsky: It looks totally weird, but let's deal with it after merging this PR. I don't like that here you recreate genesis utxo which differs from the real genesis utxo.

src/Pos/Generator/Block/Param.hs
gromakovsky: Don't forget to fix it please.

src/Pos/Generator/Block/Payload.hs
pva701: 1. where? I couldn't find it.

pva701:
@volhovM can you integrate your code in the `block-gen`? or it's better to create another issue :thinking: ?

gromakovsky:
Another issue, let's merge this one ASAP please 🙏

src/Pos/Generator/Block/Param.hs
gromakovsky: Hello!

src/Pos/Generator/Block/Payload.hs
gromakovsky: Also I don't like this `error`, but don't want to postpone the merge further.

txp/Pos/Txp/Toil/Balances/Functions.hs
```diff
@@ -66,23 +65,22 @@ recomputeStakes plusDistr minusDistr = do
                         coinToInteger totalStake + positiveDelta - negativeDelta
     let newStakes
           = HM.toList $
-              calcNegStakes minusDistr
-                  (calcPosStakes $ zip needResolve resolvedStakes ++ plusDistr)
+            map unsafeIntegerToCoin $
```
gromakovsky: Also add a comment explaining why it's safe here.

PR #1189, opened 2017-07-19
[CSM-324] Call `txMempoolToModifier` once per api request

Martoon-00:
This allows not to recalculate mempool accounts modifier (which is somewhat expensive) many times for given request, but rather fixate it once and pass implicitly around

### DIFF
--- a/src/Pos/Wallet/Web/Server/Methods.hs
+++ b/src/Pos/Wallet/Web/Server/Methods.hs
@@ -109,7 +109,7 @@ import           Pos.Wallet.WalletMode            (applyLastUpdate,
                                                    localChainDifficulty,
                                                    networkChainDifficulty, waitForUpdate)
 import           Pos.Wallet.Web.Account           (AddrGenSeed, GenSeed (..),
-                                                   genSaveRootKey,
+                                                   MonadKeySearch (..), genSaveRootKey,
                                                    genUniqueAccountAddress,
                                                    genUniqueAccountId, getAddrIdx,
                                                    getSKByAccAddr, getSKByAddr,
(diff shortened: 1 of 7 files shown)


Martoon-00:
To be tested

src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -165,8 +165,8 @@ getProfile = view wsProfile
 setProfile :: CProfile -> Update ()
 setProfile cProfile = wsProfile .= cProfile
 
-getWAddressIds :: Query [AccountId]
-getWAddressIds = HM.keys <$> view wsAccountInfos
+getAccountIds :: Query [AccountId]
+getAccountIds = HM.keys <$> view wsAccountInfos
```
flyingleafe: thank you for getting rid of these bits of nonsense ❤️

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -175,6 +176,32 @@ import           Pos.Web.Server                   (serveImpl)
...
+    let ?cachedAccModifier = accMod
+    action key
+
+fixCachedAccModifierFor
+    :: (WalletWebMode m, MonadKeySearch key m)
+    => key
+    -> (WithCachedAccMod => m a)
+    -> m a
+fixCachedAccModifierFor key action =
+    fixingCachedAccModifier (const action) key
```
flyingleafe: `ImplicitParams` may look cool, but, as I guess, they are unfamiliar to most developers and whoever else will inevitably be confused when reading such code at the first time.

I think that using simple explicit parameter won't make the code worse, instead, the code will be more straightforward and readable, and not much would change, actually - `fixingCachedAccModifier` and `fixCachedAccModifierFor` can be trivially rewritten to explicit parameter, and all functions which now have `WithCachedAccMod` in type can have `CAccModifier ->` instead as well.

flyingleafe:
@Martoon-00 isn't `wip` label obsolete here?

Martoon-00:
Tested with:
* Wallet importing and money sending
* `getAllAccounts` endpoint (others were changed in trivial way)

PR #1191, opened 2017-07-19
[CSL-1283] Use applyCoinPortionDown/Up

neongreen:


src/Pos/Lrc/Logic.hs
```diff
@@ -74,12 +75,12 @@ data RichmenType
     = RTUsual
     | RTDelegation (HashMap StakeholderId (HashSet StakeholderId))
 
---- | Pure version of findRichmen which uses in-memory Utxo.
+-- | Pure version of 'findRichmen' which uses in-memory Utxo.
```
gromakovsky: This comment is a bit outdated, because this function doesn't necessarily use Utxo.
Given that you need to force push to fix appveyor, can you update this comment?

test/Test/Pos/Ssc/GodTossing/ComputeSharesSpec.hs
```diff
@@ -72,3 +78,19 @@ totalStakeIsZero (getValid -> richmen) =
 invalidStakeErrors :: InvalidRichmenStake GenesisMpcThd -> Bool
 invalidStakeErrors (getInvalid -> richmen) =
     isLeft $ computeShares' richmen
+
+lrcConsistency :: Property
+lrcConsistency =
```
gromakovsky: Did you check that it fails without your fix?

neongreen: No, because it was too much work. I've checked that it fails if I replace `Up` by `Down` in LRC, though (and that should be equivalent).

PR #1193, opened 2017-07-20
[CSL-1358] Block event interpreter

int-index:


db/Pos/DB/Pure.hs
```diff
@@ -94,6 +95,9 @@ type MonadPureDB ctx m =
     , MonadIO m
     )
 
+dbPureDump :: MonadPureDB ctx m => m DBPure
```
gromakovsky: It would be good to use it in `VarSpec` already (there this functionality is inlined).

src/Pos/Block/Logic/Event.hs
```diff
@@ -0,0 +1,69 @@
+module Pos.Block.Logic.Event
```
gromakovsky: I don't like that this module is here. It's not used and probably will never be used by the node itself (because it's generator-specific). We usually separate code which is used by the node and which is auxiliary.

src/Pos/Block/Logic/VAR.hs
```diff
@@ -203,16 +204,26 @@ applyBlocks calculateLrc pModifier blunds = do
                spanSafe ((==) `on` view (_1 . epochIndexL)) .
                getOldestFirst
 
+data RollbackException = RollbackTipMismatch HeaderHash HeaderHash
```
gromakovsky: I think it would be better to move this type somewhere. This module contains actual logic. For instance, we have `Pos.Block.Error` module.

int-index:
@gromakovsky I addressed your criticism.

cardano-sl.cabal
```diff
@@ -571,6 +571,7 @@ test-suite cardano-test
                        Test.Pos.Block.Logic.Emulation
                        Test.Pos.Block.Logic.Mode
                        Test.Pos.Block.Logic.Util
+                       Test.Pos.Block.Logic.Event
```
gromakovsky: Ok, fine.

PR #1196, opened 2017-07-20
[CSL-1408] Rework LRC types

gromakovsky:
1. Old 'Richmen' type was bad. It assumed that there is always at least
one richman, but I don't think we can rely on it. Even if we could, it would
be better to throw error in 'MonadThrow' rather than use 'error'. Also
you almost always need to do lookup, so list is not good.
2. 'RichmenStake' was renamed to 'RichmenStakes' because it makes more sense.

### DIFF
--- a/lrc/Pos/Lrc/Types.hs
+++ b/lrc/Pos/Lrc/Types.hs
@@ -1,33 +1,22 @@
 -- | Types used in LRC.
 
 module Pos.Lrc.Types
-       ( Richmen
-       , RichmenSet
-       , RichmenStake
-       , toRichmen
+       ( RichmenSet
+       , RichmenStakes
        , FullRichmenData
        ) where
 
-import qualified Data.HashMap.Strict as HM
 import           Universum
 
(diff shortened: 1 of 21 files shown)


lrc/Pos/Lrc/Types.hs
```diff
@@ -1,33 +1,25 @@
...
 
-import qualified Data.HashMap.Strict as HM
 import           Universum
 
-import           Pos.Core.Types      (Coin, StakeholderId)
+import           Pos.Core.Types (Coin, StakeholderId)
 
 -- | Addresses which have enough stake for participation in SSC.
-type Richmen = NonEmpty StakeholderId
+-- type Richmen = NonEmpty StakeholderId
```
volhovM: Why didn't you delete it at once?

gromakovsky: Good catch, fixed.

gromakovsky:
Again, only one appveyor build succeeded and another one hangs. Given that the branch is already synchronized with master, I think it's enough.

PR #1116, opened 2017-07-05
[TW-148] Use time-warp relay queueing

avieth:
Updated July 11.

Will extend this summary in the morning. The gist of it: time-warp-nt is patched to offer an asynchronous `enqueueConversation` on the `SendActions`, and listeners are given access to the `SendActions`. To start a node, you must give an `OutboundQueue` record. The simple `freeForAll :: OutboundQueue` does what we were always doing: concurrently sends to every peer listed. A more complex queue is Edsko's `Network.Broadcast.OutboundQueue`, which is slightly modified so that it enqueues conversations and can deliver the typed results of those conversations if the caller chooses to block.

These changes make it much easier to integrate outbound queueing here in cardano-sl. Instead of creating some (misnomer) relay context and putting it into the reader antecedent, and offering a `MonadRelay` as yet another way of doing communication, we can just use what time-warp already provides: an abstract way of enqueueing a conversation with multiple peers and optionally waiting for some or all of them to finish.

https://github.com/serokell/time-warp-nt/pull/78

PR #1197, opened 2017-07-20
[TW-148] remove MonadDiscovery; use Kademlia to update OutboundQ

avieth:
It's not needed anymore. Only the OutboundQ will need to deal with peer
discovery.

You can hit a known set of peers by using `immediateConcurrentConversations` on
a `SendActions`, as the light wallet does, but typically you won't specify any targets
at all and just use `enqueueMsg`.

Notice the merge target: pull request #1116

### DIFF
--- a/infra/Pos/Communication/Types/Protocol.hs
+++ b/infra/Pos/Communication/Types/Protocol.hs
@@ -39,6 +39,7 @@ module Pos.Communication.Types.Protocol
        , MsgType (..)
        , Origin (..)
        , Msg
+       , MsgSubscribe (..)
        ) where
 
 import           Data.Aeson                 (ToJSON (..), FromJSON (..), Value)
@@ -290,3 +291,11 @@ instance Monad m => Monoid (MkListeners m) where
     a `mappend` b = MkListeners act (inSpecs a `mappend` inSpecs b) (outSpecs a `mappend` outSpecs b)
       where
         act vI pD = (++) (mkListeners a vI pD) (mkListeners b vI pD)
+
+-- | 'Subscribe' message
+--
+-- This can be used by behind-NAT nodes to subscribe to the 'OutboundQueue'
+-- of a relay node. The node will remain subscribed for the duration of the
+-- conversation.
+data MsgSubscribe = MsgSubscribe
+    deriving (Generic, Show, Eq)
--- a/infra/Pos/DHT/Real/Real.hs
+++ b/infra/Pos/DHT/Real/Real.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Pos.DHT.Real.Real
-       ( foreverRejoinNetwork
+       ( kademliaJoinNetwork
        , K.lookupNode
        , kademliaGetKnownPeers
        , startDHTInstance
@@ -16,7 +17,7 @@ import qualified Data.Store                as Store
 import           Formatting                (build, int, sformat, shown, (%))
 import           Mockable                  (Async, Catch, Mockable, MonadMockable,
                                             Promise, Throw, catch, catchAll, throw,
-                                            waitAnyUnexceptional, withAsync)
+                                            waitAnyUnexceptional, withAsync, try)
 import qualified Network.Kademlia          as K
 import qualified Network.Kademlia.Tree     as K (toView)
 import qualified Network.Kademlia.Instance as K (KademliaState (sTree), KademliaInstance (state))
@@ -76,8 +77,9 @@ startDHTInstance
        )
     => KademliaParams
     -> NodeType -- ^ Type to assign to peers discovered via Kademlia
+    -> Bool     -- ^ True if Kademlia peers should populate known peers (MonadKnownPeers)
     -> m KademliaDHTInstance
-startDHTInstance kconf@KademliaParams {..} peerType = do
+startDHTInstance kconf@KademliaParams {..} peerType subscribe = do
     let bindAddr = first B8.unpack kpNetworkAddress
         extAddr  = maybe bindAddr (first B8.unpack) kpExternalAddress
     logInfo "Generating dht key.."
@@ -103,6 +105,7 @@ startDHTInstance kconf@KademliaParams {..} peerType = do
     let kdiExplicitInitial = kpExplicitInitial
     kdiKnownPeersCache <- atomically $ newTVar []
     let kdiPeerType = peerType
+        kdiSubscribe = subscribe
     pure $ KademliaDHTInstance {..}
   where
     catchErrorsHandler e = do
@@ -204,8 +207,6 @@ kademliaJoinNetwork
     :: ( MonadIO m
        , Mockable Catch m
        , Mockable Throw m
-       , Mockable Async m
-       , Eq (Promise m (Maybe ()))
        , WithLogger m
        , Bi DHTKey
        , Bi DHTData
@@ -214,11 +215,11 @@ kademliaJoinNetwork
     -> [NetworkAddress]
     -> m ()
 kademliaJoinNetwork _ [] = throw AllPeersUnavailable
-kademliaJoinNetwork inst nodes =
-    waitAnyUnexceptional (map (kademliaJoinNetwork' inst) nodes) >>= handleRes
-  where
-    handleRes (Just _) = pure ()
-    handleRes _        = throw AllPeersUnavailable
+kademliaJoinNetwork inst (node : nodes) = do
+    outcome <- try (kademliaJoinNetwork' inst node)
+    case outcome of
+        Left (e :: DHTException) -> kademliaJoinNetwork inst nodes
+        Right _ -> return ()
 
 kademliaJoinNetwork'
     :: ( MonadIO m
--- a/infra/Pos/DHT/Real/Real.hs
+++ b/infra/Pos/DHT/Real/Real.hs
@@ -8,6 +8,7 @@ module Pos.DHT.Real.Real
        , stopDHTInstance
        ) where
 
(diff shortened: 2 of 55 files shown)


infra/Pos/Reporting/Methods.hs
```diff
@@ -135,23 +134,22 @@ ipv4Local w =
...
-    peers <- getPeers
     (ips :: [Text]) <-
         map show . filter ipExternal . map ipv4 <$>
         liftIO getNetworkInterfaces
-    pure $ sformat outputF (pretty $ listBuilderJSON ips) peers
+    pure $ sformat outputF (pretty $ listBuilderJSON ips)
   where
     ipExternal (IPv4 w) =
         not $ ipv4Local w || w == 0 || w == 16777343 -- the last is 127.0.0.1
-    outputF = ("{ nodeParams: '"%stext%"', otherNodes: "%listJson%" }")
```
gromakovsky: cc @volhovM

avieth:
@edsko @dcoutts This branch has the kademlia worker to add/remove its known peers. Also includes some work on the configuration of kademlia: the P2P and transitional topology constructors get `KademliaParams` (parsed from a `"p2p"` or `"transitional"` key in the yaml).

edsko:
I haven't checked the actual Kademlia logic but the changes to Topology and the logic for adding the worker or not seems good. I see that we can even access the node resources when we create the list of workers, so we don't have to start the worker and then exit it if we discover that we don't actually need it. That's great; I will try to do the same for the subscription worker.

infra/Pos/Reporting/Methods.hs
volhovM: Ok, so yes, it's bad. I would like to see explanation of this change. Reports must include neighbors.

stack.yaml
```diff
@@ -28,8 +28,8 @@ packages:
     commit: 95fce1dbada62020a0b2d6aa2dd7e88eadd7214b
   extra-dep: true
 - location:
-    git: https://github.com/serokell/kademlia.git
-    commit: bbdca50c263c6dae251e67eb36a7d4e1ba7c1cb6
+    git: https://github.com/avieth/kademlia.git
+    commit: b86e3cb717c3398ddc1cc25cfb8f8ea3491e4026
```
volhovM: Why did you switch to `avieth`? Is it even legal? 🤔

stack.yaml
```diff
@@ -46,7 +46,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: dcc41ea61e32e74f005cdd40ea774c210e6cfe45
+    commit: 7e1c8d794269884fa8740a347eece701a4efa1e8
```
volhovM: Which branch is it?

infra/Pos/Reporting/Methods.hs
avieth: Ok, I'll look into restoring it. I was thinking of defining a state-dump of the outbound queue, which would include all of the known peers.

infra/Pos/DHT/Workers.hs
```diff
@@ -57,6 +68,39 @@ dumpKademliaStateWorker kademliaInst = localOnNewSlotWorker True $ \slotId ->
...
         let inst = kdiHandle kademliaInst
         snapshot <- liftIO $ takeSnapshot inst
-        liftIO . BSL.writeFile dumpFile . BSL.fromStrict $ Store.encode snapshot
+        case dumpFile of
+            Just fp -> liftIO . BSL.writeFile fp . BSL.fromStrict $ Store.encode snapshot
+            Nothing -> return ()
   where
     isTimeToDump slotId = flattenSlotId slotId `mod` kademliaDumpInterval == 0
+
+kademliaSubscriptionWorker
```
avieth: Will do

stack.yaml
avieth: I hope so! I needed some changes to the kademlia library, but I don't have write access to `serokell/kademlia`. I'll make a pull request over there so we can go back to the serokell variant.

stack.yaml
avieth: Not master. But it'll be merged to master soon.

stack.yaml
avieth: Actually, I think I can do it with `serokell/kademlia/master` after all

stack.yaml
volhovM: It's alright. I believe now that the policy is: all dependencies of csl master should be at master too. So as this PR is not pointing on master, it's not an issue.

gromakovsky: No, the policy is that all dependencies should be in protected branches (master or release branch).

gromakovsky: Also, @volhovM, this PR is to feature branch, so it's not important now.

volhovM: That's what i've just said :\

gromakovsky: Github doesn't update comments automatically, so I missed it

infra/Pos/Reporting/Methods.hs
avieth: Done using
`formatKnownPeers :: MonadKnownPeers m => (forall a . Format r a -> a) -> m r`

edsko:
@avieth I suspect that the easiest way to rebase this on 1116 is simply to cherry-pick your commit into that branch.

infra/Pos/Communication/Types/Protocol.hs
```diff
@@ -290,3 +291,11 @@ instance Monad m => Monoid (MkListeners m) where
...
       where
         act vI pD = (++) (mkListeners a vI pD) (mkListeners b vI pD)
+
+-- | 'Subscribe' message
+--
+-- This can be used by behind-NAT nodes to subscribe to the 'OutboundQueue'
+-- of a relay node. The node will remain subscribed for the duration of the
+-- conversation.
+data MsgSubscribe = MsgSubscribe
+    deriving (Generic, Show, Eq)
```
avieth: Moved this over to `infra/` because the Kademlia worker will use it.

infra/Pos/DHT/Real/Real.hs
```diff
@@ -133,33 +147,26 @@ withKademliaLogger
...
-       , Mockable Throw m
-       , Eq (Promise m (Maybe ()))
-       , WithLogger m
-       , Bi DHTData
-       , Bi DHTKey
-       )
-    => KademliaDHTInstance
-    -> m [NetworkAddress]
+    :: KademliaDHTInstance
+    -> STM [NetworkAddress]
```
avieth: Moved this to `STM` so that the subscription worker can use it in a transaction. It makes that whole concurrency dance a lot easier to manage.

infra/Pos/DHT/Real/Real.hs
```diff
@@ -210,11 +215,11 @@ kademliaJoinNetwork
...
-kademliaJoinNetwork inst nodes =
-    waitAnyUnexceptional (map (kademliaJoinNetwork' inst) nodes) >>= handleRes
-  where
-    handleRes (Just _) = pure ()
-    handleRes _        = throw AllPeersUnavailable
+kademliaJoinNetwork inst (node : nodes) = do
+    outcome <- try (kademliaJoinNetwork' inst node)
+    case outcome of
+        Left (e :: DHTException) -> kademliaJoinNetwork inst nodes
+        Right _ -> return ()
```
avieth: I think it's better this way: no sense trying to join the network at multiple points concurrently. In fact, I suspect it could lead to a false positive identifier clash.

avieth:
I've just pushed a commit which gets the `./scripts/launch/demo.sh` script going. It chooses a "transitional" kademlia approach (see the `run/topology{0,1,2}.yaml` and `run/kademlia{0,1,2}.yaml` files). Logs show that the kademlia worker finds the peers and puts them into the outbound queue, but also that the queue can't find suitable peers to send the announce block message to. Similar story for some `MsgMPC` messages. Unfortunately, the outbound queue doesn't log *why* they aren't suitable. I'll try to add that feature and hopefully get some more intel.

@edsko @dcoutts

edsko:
I'm seeing the same thing in the benchmark. I pushed a commit to time warp which adds some additional logging information; now we get entries in the log such as

```
^[[92m[node.worker.ssc:DEBUG:ThreadId 5342] ^[[0m[2017-07-25 14:00:34 CEST] Rejected alternative NodeId 127.0.0.1:3002:0 as it has a recent failure
```

And indeed, a bit earlier in the log we see

```
^[[93m[node:WARNING:ThreadId 5383] ^[[0m[2017-07-25 14:00:23 CEST] NodeName "node0": sending MsgMPC (OriginForward (NodeId ae4bc5d5-3063-47c6-810e-5dc93d987907)) to NodeId 127.0.0.1:3001:0 failed with NoParse
```

edsko:
@avieth So Duncan and I figured this out a bit more.. but not all the way. Let's sync up when you're around.

infra/Pos/DHT/Workers.hs
```diff
@@ -57,6 +88,154 @@ dumpKademliaStateWorker kademliaInst = localOnNewSlotWorker True $ \slotId ->
...
+    subThread :: SendActions m -> STM.TVar (Set NodeId) -> STM.TMVar NodeId -> m ()
+    subThread sendActions subscribedVar peerVar = do
+        logNotice $ sformat ("This is a subscription thread")
+        action
+      where
+        action = do
+            peer <- liftIO . atomically $ do
+                -- Block until we get a NodeId to subscribe to.
+                it <- STM.readTMVar peerVar
+                return it
```
avieth: remove redundant return

avieth:
Merging (into #1116) because this branch contains some very important fixes, in particular to the configuration/parameters and the decision regarding addressability of the tcp transport. @edsko @dcoutts

PR #1202, opened 2017-07-21
Feature/csl 1263 add benchmarking config

kantp:
This pull request is another step towards running the benchmarks on the current master. It does the following things:

- add a section in `core/constants.yaml` for the benchmarks
- allows choosing this configuration in `dev-mode`
- adds some logging to the transaction generator.

### DIFF
--- a/core/cardano-sl-core.cabal
+++ b/core/cardano-sl-core.cabal
@@ -23,6 +23,12 @@ flag dev-mode
 
   description: Switch between Development and Production mode
 
+flag dev-custom-config
+  default:     False
+  manual:      True
+
+  description: Enable this to allow setting CONFIG manually, as in --ghc-options=-DCONFIG=benchmark
+
 flag embed-config
   default:     False
   manual:      True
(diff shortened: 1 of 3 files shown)


src/node/Params.hs
```diff
@@ -100,6 +101,7 @@ getNodeParams args@Args {..} systemStart = do
             then genesisUtxo Nothing (devAddrDistr devStakeDistr)
             else genesisUtxo (Just genesisProdBootStakeholders)
                              genesisProdAddrDistribution
+    logInfo $ sformat ("npCustomUtxo: "%shown) npCustomUtxo
```
gromakovsky: 1. There is `utxoF` in `Pos.Txp.Toil.Types` which formats it better than `shown`.
2. Sometimes this utxo can be very big and I don't want it to take hundreds or thousands of lines in the output. I would print it only if it is relatively small (at most 25 for example). If it's big, I would print only the size of it. If it's really needed, it's better to dump it to a file.

kantp:
Thanks. I've removed the possibly large logging from the node again, and only left the log of the size of the `npCustomUtxo` in `cardano-wallet`.

PR #1204, opened 2017-07-21
linking: Don't use --ghc-option=-optl-lm, use `extra-libraries: m` instead

nh2:
Fixes

  ld.gold: fatal error: cannot mix -r with dynamic object

error for dynamic linking.

See https://github.com/NixOS/nixpkgs/pull/24692#issuecomment-316856374

Tested:

* [x] `nix-build -A cardano-sl`
* [x] `nix-build -A cardano-sl-tools`

### DIFF
--- a/cardano-sl.cabal
+++ b/cardano-sl.cabal
@@ -524,6 +524,8 @@ executable cardano-node
   if os(linux)
     ghc-options:       -optl-fuse-ld=gold
     ld-options:        -fuse-ld=gold
+  -- ld.gold needs these explicitly
+  extra-libraries:     m
 
   default-extensions:   DeriveDataTypeable
                         DeriveGeneric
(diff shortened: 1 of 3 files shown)


domenkozar:
Since https://github.com/NixOS/nixpkgs/pull/27584 will hit channels in a few days I'd rather wait for the patch and remove our workarounds completely.

domenkozar:
Upstream fix is merged, for profiling in the future we just need to bump nixpkgs and add the profiling flags (by default off). @nh2 so for this PR, let's remove the `-m` option and bump nixpkgs in `nixpkgs-src.json` to latest updated master branch (to support macOS).

nh2:
I think this is done by now using the upstream nixpkgs fix.

PR #1205, opened 2017-07-21
[CSM-380] Autofill endpoint descriptions from special instances

Martoon-00:
This makes compiler to ensure that each endpoint has swagger documentation.
It also adds docs to endpoints which previously lack it (`UpdateWallet`, `TxFee`, `Import`/`ExportBackupJSON`), and bring existing ones up to date.


src/Pos/Wallet/Web/Api.hs
```diff
@@ -68,13 +73,13 @@ import           Pos.Types                  (Coin, SoftwareVersion)
...
                                              DCQueryParam, ModifiesApiRes (..),
                                              ReportDecodeError (..), VerbMod)
-import           Pos.Wallet.Web.ClientTypes (Addr, CAccount, CAccountId, CAccountInit, CCoin,
-                                             CAccountMeta, CAddress, CElectronCrashReport,
-                                             CId, CInitialized, CPaperVendWalletRedeem,
-                                             CPassPhrase, CProfile, CTx, CTxId, CTxMeta,
-                                             CUpdateInfo, CWallet, CWalletInit,
-                                             CWalletMeta, CWalletRedeem, SyncProgress,
-                                             Wal)
+import           Pos.Wallet.Web.ClientTypes (Addr, CAccount, CAccountId, CAccountInit,
```
Martoon-00: Looks like extra diff

Martoon-00: @Pastafarianist I'm now using `.stylish-haskell.yaml` from `serokell-util`, but diff remains.
Problem with previous version was in `CCoin` being at wrong place

PR #1208, opened 2017-07-22
[CSL-1341] Move tx checks

pva701:


pva701:
@gromakovsky you can review, I am going to add `GenesisStakeholders`

pva701:
this PR should be merged after 
https://github.com/input-output-hk/cardano-sl/pull/1184
because it contains ^

pva701:
@gromakovsky can you review?

gromakovsky:
I am going to review #1184 first.

txp/Pos/Txp/Toil/Types.hs
```diff
@@ -55,6 +61,13 @@ import qualified Pos.Util.Modifier          as MM
 -- output) pairs.
 type Utxo = Map TxIn TxOutAux
 
+-- | Convert 'Utxo' to 'StakesMap'.
+utxoToStakes :: Utxo -> StakesMap
```
gromakovsky: Why is it in `Types`?

txp/Pos/Txp/Toil/Types.hs
```diff
@@ -64,7 +77,21 @@ utxoF :: Format r (Utxo -> r)
 utxoF = later formatUtxo
 
 -- | Wrapper for genesis utxo.
-newtype GenesisUtxo = GenesisUtxo { unGenesisUtxo :: Utxo }
+newtype GenesisUtxo = GenesisUtxo
+    { unGenesisUtxo :: Utxo
+    }
+
+data GenesisTxpContext = UnsafeGenesisTxpContext
```
gromakovsky: Add a comment.

txp/Pos/Txp/Toil/Types.hs
gromakovsky: Hm, because it's needed for `mkGenesisTxpContext`, okay.

txp/Pos/Txp/Toil/Types.hs
```diff
@@ -64,7 +77,21 @@ utxoF :: Format r (Utxo -> r)
...
+    { _gtcUtxo         :: !GenesisUtxo
+    , _gtcStakeholders :: !GenesisStakeholders
+    }
+
+makeLenses ''GenesisTxpContext
+
+mkGenesisTxpContext :: Utxo -> GenesisTxpContext
+mkGenesisTxpContext utxo = UnsafeGenesisTxpContext
+    (GenesisUtxo utxo)
+    (GenesisStakeholders . getKeys . utxoToStakes $ utxo)
```
gromakovsky: Actually it's inefficient, but I don't mind because it should be done only on start.

txp/Pos/Txp/Toil/Logic.hs
```diff
@@ -81,116 +103,154 @@ rollbackToil txun = do
...
-        -- 1. Reject all new-coming transactions (b/c we can't calculate
-        --    fee for them)
-        -- 2. Use latest policy of known type
-        -- 3. Discard the check
-        -- Implementation-wise, the 1st option corresponds to throwing an
-        -- error here (reject), the 3rd option -- doing nothing (accept), and
-        -- the 2nd option would require some engineering feats to
-        -- retrieve previous 'TxFeePolicy' and check against it.
-        return ()
+verifyTxFeePolicy (TxFee txFee) policy txSize = do
```
gromakovsky: Absolutely useless change.

txp/Pos/Txp/Toil/Logic.hs
```diff
@@ -17,33 +21,51 @@ module Pos.Txp.Toil.Logic
...
 
 ----------------------------------------------------------------------------
 -- Global
 ----------------------------------------------------------------------------
 
-type GlobalToilMode m = ( MonadUtxo m
-                        , MonadBalances m
-                        , MonadToilEnv m
-                        , WithLogger m)
+type GlobalApplyToilMode ctx m =
```
gromakovsky: `ctx` is useless here.

txp/Pos/Txp/Logic/Local.hs
```diff
@@ -79,68 +74,48 @@ txProcessTransaction itw@(txId, txAux) = do
...
-       , MonadTxpMem () ctx m
-       , WithLogger m
-       )
+txNormalize
+    :: ( TxpLocalWorkMode ctx m
+       , MonadSlots m)
     => m ()
 txNormalize = do
     utxoTip <- GS.getTip
+    epoch <- maybe (throwM ToilUnknownCurEpoch) (pure . siEpoch) =<< getCurrentSlot
```
gromakovsky: No, it's broken. When we load blocks after long offline, we don't know the current slot. This function will throw and something will go wrong.
The proper solution is to just clean mempool if we don't know the current slot.

txp/Pos/Txp/Logic/Global.hs
```diff
@@ -45,13 +48,14 @@ txpGlobalSettings =
...
 verifyBlocks
-    :: forall m.
-       TxpGlobalVerifyMode m
+    :: forall ctx m.
+       TxpGlobalVerifyMode ctx m
     => Bool -> OldestFirst NE TxpBlock -> m (OldestFirst NE TxpUndo)
-verifyBlocks verifyAllIsKnown newChain =
-    fst <$> runToilAction @_ @() (mapM verifyDo newChain)
+verifyBlocks verifyAllIsKnown newChain = do
+    epoch <- siEpoch <$> (note ToilUnknownCurEpoch =<< getCurrentSlot)
```
gromakovsky: What is it? It's completely broken. You should take epoch from header.

gromakovsky:
#1184 is merged. Rebase please.

txp/Pos/Txp/Logic/Global.hs
pva701: ok. don't we want to reflect in the name of function the fact that only blocks from same epoch can be passed? or this invariant is true for every block processing function?

pva701:
@gromakovsky rebased

txp/Pos/Txp/Logic/Global.hs
gromakovsky: > ok. don't we want to reflect in the name of function the fact that only blocks from same epoch can be passed?

Well, I think it would be useful, but it's minor.

> this invariant is true for every block processing function?

Tt is true for every function except all rollback functions and `verifyAndApplyBlocks`.

txp/Pos/Txp/Settings/Global.hs
```diff
@@ -37,7 +37,6 @@ type TxpCommonMode ctx m =
 type TxpGlobalVerifyMode ctx m =
     ( TxpCommonMode ctx m
     , MonadError ToilVerFailure m
-    , MonadSlots m  -- TODO: I don't like it (@gromak)
```
gromakovsky: Did you test whether `WITH_EXPLORER` build still works after this change? I suspect it doesn't.

db/Pos/DB/Class.hs
```diff
@@ -183,13 +187,14 @@ gsMaxProposalSize = bvdMaxProposalSize <$> gsAdoptedBVData
 gsUnlockStakeEpoch :: MonadGState m => m EpochIndex
 gsUnlockStakeEpoch = bvdUnlockStakeEpoch <$> gsAdoptedBVData
 
+-- | Checks if provided epoch is in the bootstrap era (pure version)
+isBootstrapEraPure :: EpochIndex -> EpochIndex -> Bool
+isBootstrapEraPure unlockEpoch curEpoch =
+    not isDevelopment && isBootstrapEra unlockEpoch curEpoch
```
gromakovsky: It's absolutely bullshit, because there is `isBootstrapEra` (and `not isDevelopment` should be ~removed~ there), but I am going to do CSL-1292 today, so I will fix it later.

db/Pos/DB/Class.hs
```diff
@@ -183,13 +187,14 @@ gsMaxProposalSize = bvdMaxProposalSize <$> gsAdoptedBVData
...
+
 -- | Checks if provided epoch is in the bootstrap era.
 gsIsBootstrapEra :: MonadGState m => EpochIndex -> m Bool
-gsIsBootstrapEra epoch =
-    if isDevelopment then pure False
-    else do
-        unlockStakeEpoch <- gsUnlockStakeEpoch
-        return $ isBootstrapEra unlockStakeEpoch epoch
+gsIsBootstrapEra epoch = flip isBootstrapEraPure epoch <$> gsUnlockStakeEpoch
```
gromakovsky: I like `do` version slightly more, because it makes it harder to confuse the order of arguments. But it's minor.

txp/Pos/Txp/Settings/Global.hs
pva701: Lol, in the master there isn't this constraint and it works (I suppose).
I just returned this constraint to `master` state, why wouldn't it work?

src/Pos/Client/Txp/History.hs
```diff
@@ -219,6 +219,7 @@ type TxHistoryEnv ctx m =
     , MonadSlots m
     , MonadReader ctx m
     , HasLens GenesisUtxo ctx GenesisUtxo
```
gromakovsky: Is this lens still needed, btw? Can you check?

txp/Pos/Txp/Settings/Global.hs
pva701: ok, I got it. will try now

src/Pos/Block/Logic/Internal.hs
```diff
@@ -10,8 +10,10 @@ module Pos.Block.Logic.Internal
          -- * Constraints
          MonadBlockBase
        , MonadBlockVerify
+       , MonadBlockApplyDB
        , MonadBlockApply
 
+       , applyBlocksDbUnsafe
```
gromakovsky: I totally don't like new names in this module. They are absolutely confusing. I would export `applyBlocksUnsafe` which doesn't do normalization and would define `normalizeBlocks` in `VAR` and use it in `verifyAndApplyBlocks` in the end. It should also resolve CSL-544.

src/Pos/Context/Functions.hs
```diff
@@ -4,6 +4,7 @@ module Pos.Context.Functions
        (
          -- * Genesis
          GenesisUtxo(..)
+       , mkGenesisTxpContext
```
gromakovsky: I don't think it should be here.

src/Pos/Generator/Block/Mode.hs
```diff
@@ -123,7 +124,7 @@ data BlockGenContext = BlockGenContext
     , bgcSystemStart       :: !Timestamp
     , bgcParams            :: !BlockGenParams
     , bgcDelegation        :: !DelegationVar
-    , bgcGenesisUtxo       :: !GenesisUtxo
+    , bgcGenesisTxpContext :: !GenesisTxpContext
```
gromakovsky: I think it's wrong, it should have only `GenesisStakeholders`. I don't see why `GenesisUtxo` is needed. Moreover, it seems dangerous to have it in the context, because someone can accidentally use it, but it's wrong.
Maybe I am missing smth tho.

src/Pos/Generator/Block/Mode.hs
```diff
@@ -182,7 +183,7 @@ mkBlockGenContext bgcParams@BlockGenParams{..} = do
         return BlockGenContext {..}
   where
     -- Genesis utxo is needed only for boot era stakeholders
-    bgcGenesisUtxo =
+    bgcGenesisTxpContext =
```
gromakovsky: We both know that it's bullshit, can you remove it? `GenesisStakeholders` should come from `MonadBlockGen` instead.

src/Pos/WorkMode/Class.hs
```diff
@@ -93,6 +94,7 @@ type WorkMode ssc ctx m
       , HasLens SecurityParams ctx SecurityParams
       , HasLens TxpGlobalSettings ctx TxpGlobalSettings
       , HasLens GenesisUtxo ctx GenesisUtxo
+      , HasLens GenesisStakeholders ctx GenesisStakeholders
```
gromakovsky: I think it's better to have `HasLens GenesisTxpContext` instead, but it's minor.

gromakovsky: (instead of these two)

src/Pos/Launcher/Param.hs
```diff
@@ -61,7 +62,7 @@ data NodeParams = NodeParams
     , npSecretKey      :: !SecretKey            -- ^ Primary secret key of node
     , npUserSecret     :: !UserSecret           -- ^ All node secret keys
     , npBaseParams     :: !BaseParams           -- ^ See 'BaseParams'
-    , npGenesisUtxo    :: !GenesisUtxo          -- ^ Predefined genesis utxo
+    , npGenesisTxpCont :: !GenesisTxpContext    -- ^ Predefined genesis context related to txp data.
```
gromakovsky: We use `ctx` as a short version of `context`, don't use `cont`, it's not conventional.

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -136,7 +137,7 @@ instance Buildable TestParams where
             _tpStakeDistribution
             _tpStartTime
       where
-        utxo = _tpGenUtxo & \(GenesisUtxo u) -> u
+        utxo = _tpGenTxpContext & \g -> unGenesisUtxo (g ^. gtcUtxo)
```
gromakovsky: Without pattern matching such construct doesn't make any sense. You can write `unGenesisUtxo (_tpGenTxpContext ^. gtcUtxo)`.

src/Pos/Context/Functions.hs
pva701: otherwise, I would be forced to add `cardano-sl-txp` to block-gen or something like that.
this functions more or less related to `Context`

gromakovsky: > I would be forced to add cardano-sl-txp to block-gen or something like that.

You are saying it like it's something bad.

src/Pos/Generator/Block/Mode.hs
pva701: I am convinced that there are places where `GenesisUtxo` is needed, for example, `Tracking.hs`

gromakovsky: It doesn't seem right to use `Tracking.hs` from block generator.

pva701: didn't notice that it's block-gen, ok, I can remove. but it can be needed in the future

src/Pos/Context/Functions.hs
pva701: @gromakovsky can you explain, why `Context` can re-export `GenesisUtxo`, but `Functions` can't re-export `mkGenesisTxpContext`?

gromakovsky: I think `Context` re-exports `GenesisUtxo` because it was initially defined there and then someone didn't want to change imports in many places.
Anyway, that's not the main reason I rejected this PR, if it was the only one I would approve because it's not so important.

src/Pos/Block/Logic/Internal.hs
pva701: what should functions call which called `verifyBlocksUnsafe` earlier?

src/Pos/Generator/Block/Mode.hs
gromakovsky: It's better to remove. If later someone thinks they need it, they will have to think. And maybe it will turn out that it's not needed.

src/Pos/Block/Logic/Internal.hs
gromakovsky: What is `verifyBlocksUnsafe`?

gromakovsky: My point is that we should separate `applyBlocksUnsafe` and `normalize` and let the caller decide whether they need to call both or only one. Because actually there is only one place where `normalize` should be done, in other places it's only needed to apply.

core/Pos/Util/Util.hs
```diff
@@ -325,7 +325,7 @@ type HasLens' s a = HasLens a s a
 
 -- | Version of 'lensOf' which is used when lens is to the same type
 -- as the tag.
-lensOf' :: forall s a. HasLens' s a => Lens' s a
+lensOf' :: forall a s. HasLens' s a => Lens' s a
```
volhovM: Such a change.

src/Pos/Block/Logic/Internal.hs
```diff
@@ -79,17 +82,29 @@ type MonadBlockBase ssc ctx m
        -- Needed by some components.
        , MonadGState m
        -- This constraints define block components' global logic.
-       , HasLens TxpGlobalSettings ctx TxpGlobalSettings
        , HasLens LrcContext ctx LrcContext
+       , HasLens TxpGlobalSettings ctx TxpGlobalSettings
        , SscGStateClass ssc
+       , HasLens GenesisStakeholders ctx GenesisStakeholders
+       , MonadDelegation ctx m -- TODO: split dlgApplyBlocks blocks
```
volhovM: Wat?

src/Pos/Block/Logic/Internal.hs
```diff
@@ -79,17 +82,29 @@ type MonadBlockBase ssc ctx m
...
 type MonadBlockApply ssc ctx m
+     = ( MonadBlockApplyDB ssc ctx m
+       -- Needed for normalization.
+       , MonadTxpMem TxpExtra_TMP ctx m
+       , SscLocalDataClass ssc
+       , HasLens UpdateContext ctx UpdateContext
+       )
+
+-- | Set of constraints necessary to apply or rollback blocks at high-level (without normalization).
+type MonadBlockApplyDB ssc ctx m
```
volhovM: Reverse definition order is posos. I'd swap `MonadBlockApplyDB` and `MonadBlockApply`. And rename, yes.

src/Pos/Context/Functions.hs
volhovM: Looks weird indeed. `Context` reexports `GenesisUtxo` because it was the part of the context. And it is. I guess.

src/Pos/Explorer/Txp/Local.hs
```diff
@@ -130,9 +136,9 @@ eTxProcessTransaction itw@(txId, TxAux {taTx = UnsafeTx {..}}) = do
                      flip evalUtxoStateT resolved $
                      flip runReaderT eet $
                      runExplorerReaderWrapper $
+                     flip runReaderT genStks $
```
volhovM: Eeh, i'd prefer a specific reader environment with one datatype instead of two `ReaderT` layers.

txp/Pos/Txp/Logic/Local.hs
```diff
@@ -77,68 +75,49 @@ txProcessTransaction itw@(txId, txAux) = do
...
+       , MonadSlots m)
     => m ()
-txNormalize = do
-    utxoTip <- GS.getTip
-    localTxs <- getLocalTxs
-    ToilModifier {..} <-
-        runDBToil $ execToilTLocal mempty def mempty $ normalizeToil localTxs
-    setTxpLocalData "txNormalize" (_tmUtxo, _tmMemPool, _tmUndos, utxoTip, _tmExtra)
+txNormalize = getCurrentSlot >>= \case
+    Nothing -> clearTxpMemPool "txNormalize "
```
volhovM: What does this mean?

txp/Pos/Txp/Toil/Failure.hs
```diff
@@ -42,6 +42,7 @@ data ToilVerFailure
                           , tifSize   :: !Byte }
     | ToilUnknownAttributes !ByteString
     | ToilBootDifferentStake !TxOutDistribution
+    | ToilUnknownCurEpoch
```
volhovM: Isn't there `ToilSlotUnknown` for this already?

src/Pos/Block/Logic/Internal.hs
gromakovsky: I'd rename it back to `MonadBlockApply` and delete current `MonadBlockApply`.

src/Pos/Explorer/Txp/Local.hs
gromakovsky: Also nobody knows if this code compiles :)

txp/Pos/Txp/Logic/Local.hs
gromakovsky: It means that we delete all transactions from the mempool if we don't know the current slot. Which is fine because it means we are lost. What I don't like is that we don't update tip. It's bad @pva701

src/Pos/Context/Functions.hs
pva701: >Context reexports GenesisUtxo because it was the part of the context
sound like the most strange thing I have ever heard, lets reexprot all datatypes from `Context.hs` then. gromak's motivation sounds more or less reasonable.

`mkGenesisTxpContext` creates the part of context, why not reexprot?

core/Pos/Util/Util.hs
pva701: lol, it's not strange.
it's for convenient to avoid such code `lensOf' @_ @GenesisUtxo` and write instead `lensOf' @GenesisUtxo`

src/Pos/Block/Logic/Internal.hs
pva701: @volhovM delegation applies blocks and normalises in memory state at the same time.
I propose to split it

src/Pos/Explorer/Txp/Local.hs
pva701: @volhovM I tried to do it, but there is not `HasLens` for first and second elements of pair

txp/Pos/Txp/Toil/Failure.hs
pva701: make sense

src/Pos/Client/Txp/History.hs
pva701: still needed

src/Pos/Block/Logic/Internal.hs
volhovM: For what reason?

pva701: because we introduce `normalize` (normalisation of mempool as separate function), it's not part of `applyBlocksUnsafe` anymore

pva701: but I see the comment, that it's not so easy to normalise dlg, I will remove my comment

volhovM: My point was:
1. Normalization while doing application goes super natural for dlg
2. Maybe there are reasons to separate them though i don't see them (might have missed the discussion).

src/Pos/Explorer/Txp/Local.hs
gromakovsky: 🤦‍♂️ 
The proper way is to create a special type and define `Has` for it.

gromakovsky: This way is ugly though 😞

pva701: >The proper way is to create a special type and define Has for it.
делать мне больше нечего, чтобы на каждый чих, ради того что кому-то там два ридера не нравится создавать дататайп :peka:

src/Pos/Block/Logic/Internal.hs
gromakovsky: One reason is that we need to do normalization only once when we apply a big sequence of blocks, but currently we can do it multiple times. It's not very important for delegation I think.

pva701:
@gromakovsky review plz

src/Pos/Generator/Block/Mode.hs
gromakovsky: Remove bullshit please 😔

pva701:
@gromakovsky let's remove bullshit in next issues

pva701:
@gromakovsky can you re-review again?

db/Pos/DB/Class.hs
```diff
@@ -187,7 +190,7 @@ gsUnlockStakeEpoch = bvdUnlockStakeEpoch <$> gsAdoptedBVData
 gsIsBootstrapEra :: MonadGState m => EpochIndex -> m Bool
 gsIsBootstrapEra epoch = do
     unlockStakeEpoch <- gsUnlockStakeEpoch
-    return $ isBootstrapEra unlockStakeEpoch epoch
+    pure $ isBootstrapEra unlockStakeEpoch epoch
```
gromakovsky: Such a change.

gromakovsky:
Did you test compilation with explorer?

db/Pos/DB/Class.hs
pva701: pure <3

pva701:
I did, it's ok

src/Pos/Explorer/Txp/Local.hs
```diff
@@ -147,22 +153,19 @@ eTxProcessTransaction itw@(txId, TxAux {taTx = UnsafeTx {..}}) = do
...
-       , MonadGState m
-       , MonadTxpMem ExplorerExtra ctx m
-       , WithLogger m
+       ( ETxpLocalWorkMode ctx m
+       , MonadSlots m
        )
     => m ()
 eTxNormalize = do
     utxoTip <- GS.getTip
+    epoch <- maybe (throwM ToilSlotUnknown) (pure . siEpoch) =<< getCurrentSlot
```
gromakovsky: Wait, what is it? Why is it different from `txNormalize`? It should be the same.

src/node/Params.hs
```diff
@@ -94,9 +96,11 @@ getNodeParams args@Args {..} systemStart = do
...
                 (CLI.richPoorDistr commonArgs)
                 (CLI.expDistr commonArgs)
-    let npGenesisUtxo
-            | isDevelopment = genesisUtxo Nothing (devAddrDistr devStakeDistr)
-            | otherwise = genesisUtxoProduction
+    let npGenesisTxpCtx = mkGenesisTxpContext $
+            if isDevelopment
+            then genesisUtxo Nothing (devAddrDistr devStakeDistr)
+            else genesisUtxo (Just genesisProdBootStakeholders)
+                                genesisProdAddrDistribution
```
gromakovsky: Why did you change `genesisUtxoProduction` to `genesisUtxo (Just genesisProdBootStakeholders) genesisProdAddrDistribution`?

src/Pos/Explorer/Txp/Local.hs
pva701: FUCK this FUCKING boilerplate, It's so tedious to manage same code in the fucking two places.

domenkozar:
@pva701 can you rebase on top of master?

PR #1210, opened 2017-07-23
[CSL-1398] Cache getPskFromThisEpoch in dlg db

volhovM:


src/Pos/Delegation/Class.hs
```diff
@@ -56,9 +54,6 @@ data DelegationWrap = DelegationWrap
       -- a good approximation.
     , _dwEpochId           :: EpochIndex
       -- ^ Epoch index 'DelegationWrap' is correct in relation to.
-    , _dwThisEpochPosted   :: HashSet PublicKey
```
gromakovsky: Why did you remove it from the context? It doesn't take much space and it's faster to take this data from memory than from DB.

src/Pos/Delegation/DB.hs
```diff
@@ -132,6 +154,20 @@ instance DBIteratorClass DlgTransRevIter where
...
+-- Iterator over the "this epoch posted" set.
+data ThisEpochPostedIter
+
+instance DBIteratorClass ThisEpochPostedIter where
+    type IterKey ThisEpochPostedIter = StakeholderId
+    type IterValue ThisEpochPostedIter = ()
+    iterKeyPrefix = iterPostedThisEpochPrefix
+
+-- | Get all keys of thisEpochPosted set.
+getThisEpochPostedKeys :: MonadDBRead m => m [StakeholderId]
```
gromakovsky: It would be better to return `HashSet` I think. It's more convenient because I don't think someone may need this data in other form than a set. And its more optimal, because you can construct a `HashSet` without constructing intermediate list.
Perhaps it's not really important.

src/Pos/Delegation/Types.hs
```diff
@@ -58,8 +59,24 @@ mkDlgPayload proxySKs = do
...
     wrongPSKs = filter (not . verifyPsk) proxySKs
 
--- | PSKs we've overwritten/deleted.
-type DlgUndo = [ProxySKHeavy]
+-- | Undo for the delegation component.
+data DlgUndo = DlgUndo
+    { duPsks            :: [ProxySKHeavy]
+      -- ^ PSKs we've modified when applying the block (by deleting or
+      -- overwriting).
+    , duPrevEpochPosted :: HashSet StakeholderId
```
gromakovsky: Why not strict?

src/Pos/Delegation/Class.hs
volhovM: Because there's no need for it. Previously we were checking if PSK is legal using this set before putting it to mempool. Now we just do a single read from DB. I think it can be even faster.

src/Pos/Delegation/DB.hs
volhovM: I think you're right. On the other hand `consume` is so convenient. :(

src/Pos/Delegation/Class.hs
gromakovsky: I don't think a single read from DB can be faster. But it also shouldn't be much slower.
What can be slow is to read all ids in `dlgVerifyBlocks` in tests. Profiling will answer.

src/Pos/Delegation/Types.hs
volhovM: Will do!

PR #1211, opened 2017-07-23
[CSL-1414] Write a test using BlockEvent generator and interpreter

int-index:


core/constants.yaml
```diff
@@ -165,7 +165,7 @@ qanet_tns: &qanet_tns
 travis: &travis
   <<: *qanet_tns
 
-  k: 3  # it's needed for tests
+  k: 300  # it's needed for tests
```
gromakovsky: I hope it's temporary.

godtossing/Pos/Ssc/GodTossing/DB.hs
```diff
@@ -35,12 +37,11 @@ gtGlobalStateToBatch :: GtGlobalState -> GtOp
...
--- Commented due to CSL-1113, maybe uncomment when we will use store serialization.
-  --   whenNothingM_ (gsGetBi @_ @GtGlobalState gtKey) $
-  --       gsPutBi gtKey (def {_gsVssCertificates = vcd})
-  -- where
-  --   vcd = VCD.fromList . toList $ certs
+initGtDB certs =
+    whenNothingM_ (gsGetBi @_ @GtGlobalState gtKey) $
+        gsPutBi gtKey (def {_gsVssCertificates = vcd})
+  where
+    vcd = VCD.fromList . toList $ certs
```
gromakovsky: 1. Given that this code is touched, I would remove `certs` parameter, because it's always constant.
2. It should be checked for performance regression. It was commented out for a reason and maybe it's not ok to just uncomment it.

src/Pos/GState/BlockExtra.hs
```diff
@@ -169,6 +169,7 @@ initGStateBlockExtra :: MonadDB m => HeaderHash -> m ()
 initGStateBlockExtra firstGenesisHash = do
     gsPutBi (mainChainKey firstGenesisHash) ()
     gsPutBi (forwardLinkKey genesisHash) firstGenesisHash
+    gsPutBi lastSlotsKey noLastBlkSlots
```
gromakovsky: With this change I would also modify `getLastSlots` to throw `DBMalformed` if this key is not in DB instead of returning `noLastBlkSlots`.

godtossing/Pos/Ssc/GodTossing/DB.hs
gromakovsky: Also now it makes sense to remove `getGtGlobalStateMaybe`. But only if we can really afford this change.

core/constants.yaml
int-index: How else would the introduced test work? The amount of generated blocks cannot exceed `k`, and `3` is too little. I'll need at least 10 or 20 here for the test to make sense.

godtossing/Pos/Ssc/GodTossing/DB.hs
int-index: > Given that this code is touched, I would remove certs parameter, because it's always constant.

Alright.

> It should be checked for performance regression. It was commented out for a reason and maybe it's not ok to just uncomment it.

I remember when it happened and I even had a fix for it (there's a closed PR somewhere) which was subsumed by the effort of migration to `store`.

gromakovsky: It would be good to know whether this change will be acceptable with `cborg`.

src/Pos/GState/BlockExtra.hs
int-index: Done

core/constants.yaml
int-index: I decreased it to `35`, but it's temporary and I will set it back to `3` in a different task.

src/Pos/Block/Logic/Creation.hs
```diff
@@ -309,7 +309,7 @@ applyCreatedBlock pske createdBlock = applyCreatedBlockDo False createdBlock
     applyCreatedBlockDo :: Bool -> MainBlock ssc -> m (MainBlock ssc)
     applyCreatedBlockDo isFallback blockToApply =
         verifyBlocksPrefix (one (Right blockToApply)) >>= \case
-            Left reason
+            Left (fromString . displayException -> reason)
```
gromakovsky: Why do you use it instead of `pretty`?

src/Pos/Block/Error.hs
```diff
@@ -24,11 +26,44 @@ data RollbackException = RollbackTipMismatch HeaderHash HeaderHash
...
+    = VerifyBlocksError Text
+    deriving (Show)
+
+instance Exception VerifyBlocksException where
+    displayException = toString . renderVerifyBlocksException
+
+instance Buildable VerifyBlocksException where
+    build = Data.Text.Buildable.build . renderVerifyBlocksException
+
+renderVerifyBlocksException :: VerifyBlocksException -> Text
```
gromakovsky: Why do you define `Buildable` via conversion to `Text` instead of defining conversion to `Text` via `Buildable`? The latter is more reasonable and efficient.

db/Pos/DB/Pure.hs
```diff
@@ -142,3 +147,71 @@ dbWriteBatchPureDefault (tagToLens -> l) batchOps =
 
 atomicModifyIORefPure :: (MonadIO m) => (a -> a) -> IORef a -> m ()
 atomicModifyIORefPure foo = flip atomicModifyIORef $ \a -> (foo a, ())
+
+data ByteStringDiff = ByteStringDiff
```
volhovM: I know it's almost obvious what it does but the policy is to document everything exported if possible.

db/Pos/DB/Pure.hs
```diff
@@ -142,3 +147,71 @@ dbWriteBatchPureDefault (tagToLens -> l) batchOps =
...
+    { _bsdLeft  :: ByteString
+    , _bsdRight :: ByteString
+    } deriving (Show)
+
+bsDiff :: ByteString -> ByteString -> Maybe ByteStringDiff
+bsDiff b1 b2
+    | b1 == b2 = Nothing
+    | otherwise = Just $ ByteStringDiff b1 b2
+
+data MapDiff k vDiff = MapDiff
```
volhovM: And here it's definitely needed to clarify what `vDiff` does. Docs!

godtossing/Pos/Ssc/GodTossing/DB.hs
```diff
@@ -2,45 +2,41 @@
...
-initGtDB :: MonadDB m => VssCertificatesMap -> m ()
-initGtDB _ = pass
--- Commented due to CSL-1113, maybe uncomment when we will use store serialization.
-  --   whenNothingM_ (gsGetBi @_ @GtGlobalState gtKey) $
-  --       gsPutBi gtKey (def {_gsVssCertificates = vcd})
-  -- where
-  --   vcd = VCD.fromList . toList $ certs
+initGtDB :: MonadDB m => m ()
+initGtDB =
+    whenNothingM_ (gsGetBi @_ @GtGlobalState gtKey) $
```
volhovM: I think it is incorrect. See `isInitialized` and `initKey` in `Pos.DB.GState.Common`. It's main usage is `prepareGStateDB` in`Pos.GState.GState`. I don't know what is `gtKey` but my guess is that it should be used here.

src/Pos/Block/Error.hs
```diff
@@ -24,11 +26,44 @@ data RollbackException = RollbackTipMismatch HeaderHash HeaderHash
...
         tipMismatchMsg "rollback" storedTip attemptedTip
 
 instance Exception RollbackException where
     displayException = toString . renderRollbackException
 
 instance Buildable RollbackException where
     build = Data.Text.Buildable.build . renderRollbackException
+
+data ApplyBlocksException
+    = ApplyBlocksTipMismatch Text HeaderHash HeaderHash
```
volhovM: Which is expected, which is real? Do i understand correctly that `ApplyBlocksError` captures everything we can't predict? D o c s.

src/Pos/Block/Error.hs
```diff
@@ -24,11 +26,44 @@ data RollbackException = RollbackTipMismatch HeaderHash HeaderHash
...
+    ApplyBlocksTipMismatch s tip attemptedTip ->
+        tipMismatchMsg s tip attemptedTip
+    ApplyBlocksVerifyFailure e -> renderVerifyBlocksException e
+    ApplyBlocksError e -> e
+
+instance Exception ApplyBlocksException where
+    displayException = toString . renderApplyBlocksException
+
+instance Buildable ApplyBlocksException where
+    build = Data.Text.Buildable.build . renderApplyBlocksException
```
volhovM: Why not use qualified import `as B` or `as Bulidable` instead?

src/Pos/Block/Slog/Logic.hs
```diff
@@ -283,8 +283,21 @@ slogRollbackBlocks blunds = do
...
+    -- 'blkSecurityParam' slots. 'lastSlotsToPrepend' are slots for
+    -- main blocks which are 'blkSecurityParam' far from the blocks we
+    -- want to rollback.  Concatenation of these lists contains slots
+    -- of some sequence of blocks. The last block in this sequence is
+    -- the last block we rollback. The length of this sequence is
+    -- 'min(total, n + blkSecurityParam)', where 'n' is the number of
+    -- main blocks we want to rollback and 'total' is the total number
+    -- of main blocks in our chain. So the final step is to drop last
+    -- 'n' slots from this list.
     updateLastSlots lastSlots =
```
volhovM: I would also love to see the type declaration here.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -34,6 +40,8 @@ spec = describe "Block.Logic.VAR" $ modifyMaxSuccess (min 12) $ do
     describe "verifyBlocksPrefix" verifyBlocksPrefixSpec
     describe "verifyAndApplyBlocks" verifyAndApplyBlocksSpec
     describe "applyBlocks" applyBlocksSpec
+    describe "Block.Event" $ do
```
volhovM: I wonder why didn't hlint say anything about redundant `do` here.

int-index: This is supposed to be extended anyway, so this `do` will be necessary soon.

godtossing/Pos/Ssc/GodTossing/DB.hs
int-index: I simply uncommented this code. What's your proposed fix, exactly?

volhovM: Take a look on how `GState` is initialized. There's a single key for it (`initKey`), so all sub-dbs initialization do not need to have their init keys. I think that you should delete `whenNothingM_`.

int-index: I don't know this code well enough to judge, perhaps you should make a separate issue about it.

gromakovsky: AFAIK you should simply delete this `whenNothingM_` but it will be better if @volhovM comments.

test/Test/Pos/Block/Logic/VarSpec.hs
gromakovsky: Because it's disabled in hlint. Humans know better. We have many such `do`s.

In general I don't like redundant `do`s, but in `hspec` it's quite natural.

src/Pos/Block/Error.hs
```diff
@@ -3,32 +3,82 @@
...
+    bprint
+        ("Can't "%stext%" block because of tip mismatch (stored is "
+         %shortHashF%", attempted is "%shortHashF%")")
+        action storedTip attemptedTip
+
 data RollbackException = RollbackTipMismatch HeaderHash HeaderHash
     deriving (Show)
 
-renderRollbackException :: RollbackException -> Text
+renderRollbackException :: RollbackException -> Builder
```
gromakovsky: I don't know what's the purpose of moving it into a separate function, but I don't mind.

PR #1214, opened 2017-07-24
CSL-1292 Enable bootstrap era in development mode 

gromakovsky:
I checked that payments API works in postman (transaction is added to blocks) and that tests pass (CI checks it too).
Need to check lwallet (first need to figure out how to do it). I don't see a way I could break it, but I will check.
Also I had to comment out some code after I realized that CSL-1351 wasn't implemented yet.

### DIFF
--- a/lwallet/Pos/Wallet/Light/Mode.hs
+++ b/lwallet/Pos/Wallet/Light/Mode.hs
@@ -24,10 +24,15 @@ import           Pos.Block.BListener              (MonadBListener (..), onApplyB
 import           Pos.Client.Txp.Balances          (MonadBalances (..))
 import           Pos.Client.Txp.History           (MonadTxHistory (..))
 import           Pos.Communication.Types.Protocol (NodeId)
+import           Pos.Core                         (SlotId (..))
 import           Pos.DB                           (MonadGState (..))
 import           Pos.Discovery                    (MonadDiscovery (..))
 import           Pos.Reporting.MemState           (ReportingContext)
+import           Pos.Slotting                     (MonadSlots (..),
+                                                   currentTimeSlottingSimple)
+import           Pos.Slotting.MemState            (MonadSlotsData (..))
 import           Pos.Ssc.GodTossing               (SscGodTossing)
+import           Pos.Txp                          (GenesisUtxo)
 import           Pos.Util.JsonLog                 (HasJsonLogConfig (..), JsonLogConfig,
                                                    jsonLogDefault)
 import           Pos.Util.LoggerName              (HasLoggerName' (..),
@@ -56,6 +61,7 @@ data LightWalletContext = LightWalletContext
     , lwcDiscoveryPeers   :: !(Set NodeId)
     , lwcJsonLogConfig    :: !JsonLogConfig
     , lwcLoggerName       :: !LoggerName
+    , lwcGenesisUtxo      :: !GenesisUtxo
     }
 
 makeLensesWith postfixLFields ''LightWalletContext
@@ -64,6 +70,9 @@ type LightWalletMode = Mtl.ReaderT LightWalletContext Production
 
 instance HasUserSecret LightWalletContext where
     userSecret = lwcKeyData_L
+
+instance HasLens GenesisUtxo LightWalletContext GenesisUtxo where
+    lensOf = lwcGenesisUtxo_L
 
 instance HasLens WalletState LightWalletContext WalletState where
     lensOf = lwcWalletState_L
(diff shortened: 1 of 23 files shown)


gromakovsky:
Okay, I tested lwallet, found a bug, fixed, force-pushed. Ready for a final review.

src/Pos/Genesis.hs
```diff
@@ -109,35 +109,73 @@ stakeDistribution ts@RichPoorStakes {..} =
...
+--
+-- TODO [CSL-1351] This documentation will become correct later.
+genesisUtxo ::
+       Maybe (HashMap StakeholderId Word16) -> [AddrDistribution] -> GenesisUtxo
+genesisUtxo bootStakeholdersMaybe ad
+    -- TODO [CSL-1351] Uncomment ↓.
+    --- | null bootStakeholders =
+    ---     error "genesisUtxo: no stakeholders for the bootstrap era"
+    -- TODO [CSL-1351] Delete (it's here to please hlint) ↓.
+    | ((*) @Int) 2 3 == 7 = error "hlint — makaka"
```
volhovM: What was the error?

src/Pos/Genesis.hs
```diff
@@ -109,35 +109,73 @@ stakeDistribution ts@RichPoorStakes {..} =
...
-        in zipWith utxoEntry addr distrs
+    -- This type is drunk.
+    somethingComplicated :: [([Address], [Coin])]
+    somethingComplicated = map (second stakeDistribution) ad
+    balances :: [(Address, Coin)]
+    balances = concatMap (uncurry zip) somethingComplicated
+    utxoEntry (addr, coin) =
+        ( TxIn (unsafeHash addr) 0
+-- Note: it's quite bad because HD wallets become stakeholders and each
+-- transaction will give them quite a lot of stake.
```
volhovM: I don't understand anything here. This comment doesn't make sense to me :(
Could you please rephrase it somehow?

src/Pos/Genesis.hs
```diff
@@ -109,35 +109,73 @@ stakeDistribution ts@RichPoorStakes {..} =
...
+-- But CSL-1351 is planned for this sprint, so I suppose it's fine (it's
+-- unlikely to cause problems anyway).
+-- TODO [CSL-1351] Delete ↓.
+        , TxOutAux (TxOut addr coin) (outDistr coin))
+    -- Empty distribution for PubKey address means that the owner of
+    -- this address will have the stake.
```
volhovM: Nope, it's not true. Default for empty list is "all stake to output owner". cc @neongreen

lwallet/Pos/Wallet/Light/Mode.hs
```diff
@@ -101,6 +110,20 @@ instance MonadUpdates LightWalletMode where
     waitForUpdate = error "notImplemented"
     applyLastUpdate = pure ()
 
+-- FIXME: Dummy instance for lite-wallet.
```
gromakovsky: There story is that light wallet requires some constraints which are not used and there many such instances in this file.

src/Pos/Genesis.hs
gromakovsky: There error is useless guard. It should be removed when commented code it uncommented (and `| null bootStakeholders` should take place.

src/Pos/Genesis.hs
gromakovsky: Does it differ from what's written in the comment? In any case, it should be removed within CSL-1351.

src/Pos/Genesis.hs
gromakovsky: I don't want to rephrase it in the code, because I expect you to do CSL-1351 soon. I can clarify here.
Genesis addresses in dev mode consist of normal `PubKey` addresses and HD addresses (see `devAddrDistr`). HD addresses shouldn't have much stake and currently there is this:
```
    hdwSize = 2 -- should be positive
    -- 200 coins split among hdwSize users. Should be small sum enough
    -- to avoid making wallets slot leaders.
    hdwDistr = FlatStakes (fromIntegral hdwSize) (mkCoin 200)
```

So there are two addresses each holding 100 coins. It's expected to be much smaller than the total stake in the system.
It's fine initially, but each transaction will give these HD addresses same stake as other genesis addresses. It's quite bad, but not critical.
And it will be solved by weights (HD addresses will have small weights which is already implemented).

volhovM: >but each transaction will give these HD addresses same stake 
Why do you think so? There is a _hardcoded_ list of boot stakeholders and hd keys are not included.

gromakovsky: This list of boot stakeholders is currently computed from utxo (`utxoToStakes`). Later it will be impossible because stakeholders will have weights. When it happens, one will need to uncomment the code below.
In any case, **this** code determines which stakeholders become bootstrap stakeholders. And currently it puts all genesis addresses. Currently they have the same weight (to be more precise, they have no weight, because it's not used), but different stake (same as balance). After weights are used, there won't be a problem at all and will not be different from what we have now (it's already not different unless you send a lot of money in transactions).

gromakovsky:
MacOS build `travis-ci/push` timed out 3 times in a row, `travis-ci/pr` passed. Everything else passes too. I think it's ok to merge it.

PR #1215, opened 2017-07-25
CSL-1277 Make wallet.sh almost usable

gromakovsky:
You can use wallet.sh and it will generate a proper command now.
However, I don't know how to properly execute it from bash.
So current approach is to copy it and execute manually.


neongreen:
seems that nobody will review it after all :(

PR #1218, opened 2017-07-25
[CSL-1409] Override txdistr (boot era) in redeemtx creation

volhovM:
This adds conditional `txDistr` in boot epoch for redeeming trasactions.

### DIFF
--- a/src/Pos/Communication/Tx.hs
+++ b/src/Pos/Communication/Tx.hs
@@ -9,6 +9,7 @@ module Pos.Communication.Tx
        , sendTxOuts
        ) where
 
+import           Control.Monad.Except       (runExceptT)
 import           Formatting                 (build, sformat, (%))
 import           Mockable                   (MonadMockable, mapConcurrently)
 import           System.Wlog                (logInfo)
@@ -18,7 +19,8 @@ import           Pos.Binary                 ()
 import           Pos.Client.Txp.Balances    (MonadBalances (..), getOwnUtxo)
 import           Pos.Client.Txp.History     (MonadTxHistory (..))
 import           Pos.Client.Txp.Util        (TxCreateMode, TxError (..), createMTx,
-                                             createRedemptionTx, createTx)
+                                             createRedemptionTx, createTx,
+                                             overrideTxDistrBoot)
 import           Pos.Communication.Methods  (sendTx)
 import           Pos.Communication.Protocol (NodeId, OutSpecs, SendActions)
 import           Pos.Communication.Specs    (createOutSpecs)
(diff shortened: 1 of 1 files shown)


gromakovsky:
Fix issue ID please. Also CI didn't pass, btw.

src/Pos/Communication/Tx.hs
```diff
@@ -96,12 +100,11 @@ submitRedemptionTx sendActions rsk na output = do
...
     let addCoin c = unsafeAddCoin c . txOutValue . toaOut
         redeemBalance = foldl' addCoin (mkCoin 0) utxo
-        txouts = one $
+        txOuts0 = one $
             TxOutAux {toaOut = TxOut output redeemBalance, toaDistr = []}
-    when (redeemBalance == mkCoin 0) $
-        throwM . TxError $ "Redeem balance is 0"
-    txw <- eitherToThrow TxError $
-           createRedemptionTx utxo rsk txouts
+    txOuts <- either (throwM . TxError) pure =<< runExceptT (overrideTxDistrBoot txOuts0)
```
gromakovsky: :dimq: added `Pos.Util.Util.eitherToThrow` recently, you can use it here.

src/Pos/Communication/Tx.hs
```diff
@@ -96,12 +100,11 @@ submitRedemptionTx sendActions rsk na output = do
...
+    when (redeemBalance == mkCoin 0) $ throwM . TxError $ "Redeem balance is 0"
+    txw <- eitherToThrow TxError $ createRedemptionTx utxo rsk txOuts
```
gromakovsky: Heh, the funny thing is that it's actually used here, but not used above.

volhovM:
I didn't test it. Probably @Martoon-00 can. CI now passes (except for macos which times out) and i made all needed changes.

Martoon-00:
Going to do now

Martoon-00:
@volhovM Now redeeming hangs for me.
Even in case when I provide outdated certificate, which ended up with `Redeem balance is 0` error previously.
I tested with not all blocks downloaded though, can recheck some later if it matters

volhovM:
@Martoon-00 it will hang because `getCurrentSlotBlocking` blocks. In order to send redeem transaction you now must know current `SlotId` ~ be synchronized with the chain. It's needed to actually determine whether we are or not in the boot era.

Martoon-00:
Thanks, once all blocks were downloaded redeeming started to create transaction with money.

If I reuse certificate or use outdated one I get
> Cannot send redemption transaction: Transaction creation error (Can't spend 0 coin(s) coins: amount is too small for boot  era and can't be distributed among genStakeholders)
though, thus check here https://github.com/input-output-hk/cardano-sl/pull/1218/files#diff-94ca39c8ae8c35d9a124e1b3895400d4R103 seems redundant now - I leave it on your judgement

gromakovsky:
I think those two lines should be swapped for more clear error message, good catch.

volhovM:
Lines swapped, let's wait for CI and merge then.

PR #1220, opened 2017-07-26
Print commit when creating genesis

neongreen:
Please?

### DIFF
--- a/scripts/generate/genesis.sh
+++ b/scripts/generate/genesis.sh
@@ -29,9 +29,12 @@ fi
 
 F=100 # fake avvm keys
 
+# print commit
+git show HEAD --oneline | tee $outputDir/genesisCreation.log
+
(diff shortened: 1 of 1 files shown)


scripts/generate/genesis.sh
```diff
@@ -29,9 +29,12 @@ fi
 
 F=100 # fake avvm keys
 
+# print commit
+git show HEAD --oneline | tee $outputDir/genesisCreation.log
```
gromakovsky: You added `-a` below, but didn't use it here. Looks strange. Is it intentional?

neongreen: Yes, `-a` is “append”.

gromakovsky:
I couldn't find any mentions of this file in CI files and scripts. I see no way for it to break CI.
Given that CI is overloaded and we need to merge many PRs and that I can't stop appveyor builds, I find it better to merge this PR.

PR #1223, opened 2017-07-26
[CSM-387] Set syncTip up to date in `newWallet`

Martoon-00:


flyingleafe:
We should wait for https://github.com/input-output-hk/cardano-sl/pull/1226 to be merged into `cardano-sl-0.5` first, then `merge -s` `cardano-sl-0.5` here, and then merge it into master, to preserve correct commit history.

gromakovsky:
Good!

PR #1230, opened 2017-07-27
Tests csl1396 epochorslot

rockbmb:


core/Pos/Arbitrary/Core.hs
```diff
@@ -285,6 +286,30 @@ newtype SafeWord = SafeWord
 instance Arbitrary SafeWord where
     arbitrary = SafeWord . Types.getCoinPortion <$> arbitrary
 
+-- | A wrapper over 'Int'. When converted to 'EpochOrSlot' via 'fromEnum', using this
```
gromakovsky: Why do you call it a wrapper over `Int` if it's a wrapper over `EpochOrSlot`? 🤔

core/Pos/Arbitrary/Core.hs
```diff
@@ -285,6 +286,30 @@ newtype SafeWord = SafeWord
...
+-- type ensures there's an exception.
+newtype EoSToIntOverflow = EoSToIntOverflow
+    { getInt :: Types.EpochOrSlot
```
gromakovsky: I find this `int` quite confusing, especially because it returns `EpochOrSlot`.

core/Pos/Core/Slotting.hs
```diff
@@ -90,33 +90,42 @@ instance HasEpochIndex EpochOrSlot where
...
+              | otherwise -> fromIntegral res
     fromEnum (EpochOrSlot (Right SlotId {..})) =
-        fromEnum (EpochOrSlot (Left siEpoch)) +
-        fromIntegral (getSlotIndex siSlot) +
-        1
+        let res = toInteger (fromEnum (EpochOrSlot (Left siEpoch))) +
+                  toInteger (getSlotIndex siSlot) +
+                  1
+            maxIntAsInteger = toInteger (maxBound :: Int)
+        in if | res > maxIntAsInteger -> error "fromEnum @EpochOrSlot"
```
gromakovsky: It would be good to describe what exactly caused the error, not only the function name (e. g. smth like `fromEnum @EpochOrSlot: overflow`).

test/Test/Pos/Types/SlottingSpec.hs
```diff
@@ -4,26 +4,72 @@ module Test.Pos.Types.SlottingSpec
...
 
 flattenThenUnflatten :: SlotId -> Property
 flattenThenUnflatten si = si === unflattenSlotId (flattenSlotId si)
+
+predThenSucc :: EpochOrSlot -> Property
+predThenSucc eos = eos > minBound ==> succ (pred eos) === eos
+
+predToMinBound :: Expectation
+predToMinBound =
+    shouldThrowException pred anyErrorCall (EpochOrSlot $ Left (EpochIndex 0))
```
gromakovsky: Why not use `minBound`?

test/Test/Pos/Types/SlottingSpec.hs
```diff
@@ -4,26 +4,72 @@ module Test.Pos.Types.SlottingSpec
...
+succThenPred eos = eos < maxBound ==> pred (succ eos) === eos
+
+succToMaxBound :: Expectation
+succToMaxBound = shouldThrowException succ anyErrorCall (maxBound :: EpochOrSlot)
+
+toFromEnum :: NonNegative Int -> Property
+toFromEnum (getNonNegative -> int) = fromEnum (toEnum @EpochOrSlot int) === int
+
+fromToEnum :: EpochOrSlot -> Property
+fromToEnum = toEnum . fromEnum .=. identity
```
gromakovsky: Hm, it actually doesn't always hold because if `EpochOrSlot` is e. g. `maxBound` it will throw an exception. I guess the reason it works now is that in `instance Arbitrary EpochIndex` we always generate reasonable epochs (at most 5 * 10¹²).  Maybe it makes sense to handle this case separately (just to make sure this test won't break if we increase `maxReasonableEpoch` or change `Arbitrary` instance), but I don't insist.

core/Pos/Arbitrary/Core.hs
rockbmb: It used to be a wrapper over `Int'` for about a minute before I realized it had to be a `EpochOrSlot`.

PR #1232, opened 2017-07-27
[CSL-1435] Disable VarSpec in appveyor

gromakovsky:
We need to temporarily disable `Block.Logic.Var` tests in appveyor because they are very slow, it leads to timeouts, we can't merge new PRs and the proper solution is not easy.


gromakovsky:
I failed.

gromakovsky:
Well, it really skips these tests on my machine, but doesn't skip on Windows, I will make another attempt.

gromakovsky:
Okay, both appveyor builds passed and I don't want to wait for travis because this commit changes only appveyor config on which travis CI must not depend.

PR #1233, opened 2017-07-27
[CSM-391] Throw error on sending money to redeem address

Martoon-00:
Testing results:
* Money sending on pubkey addresses still works
* Money sending to random redeem addresses generated in ghci fails
* Same thing for txFee endpoint

### DIFF
--- a/core/Pos/Core/Types.hs
+++ b/core/Pos/Core/Types.hs
@@ -7,6 +7,7 @@ module Pos.Core.Types
        (
        -- * Address
          Address (..)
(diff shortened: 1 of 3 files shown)


core/Pos/Core/Types.hs
```diff
@@ -7,6 +7,7 @@ module Pos.Core.Types
        (
        -- * Address
          Address (..)
+       , _RedeemAddress
```
flyingleafe: You could import all prisms altogether here, though I don't mind.

PR #1235, opened 2017-07-27
[WIP] [DO NOT MERGE YET] [CSL-1436] Do not override any constants in CI modes

gromakovsky:
I fixed the problem that Travis CI was using different `k`. However, for tests we still need a small `k`. So we need to find a solution for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ cache:
 before_install:
 - echo $TRAVIS_OS_NAME
 - export CSL_SYSTEM_TAG=$TRAVIS_OS_NAME
-- export DCONFIG=travis
+- export DCONFIG=ci_wallet
 - mkdir -p ~/.local/bin s3
 - export PATH=$HOME/.local/bin:$PATH
 - sudo mount -o remount,exec,size=4G,mode=755 /run/user || true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
     CSL_SYSTEM_TAG: "win64"
     CACHE_DIR: "c:\\c"
     WORK_DIR: "c:\\w"
-    DCONFIG: "qanet_tns_wallet"
+    DCONFIG: "ci_wallet"
 
 init:
   - SET BRANCH_CACHE_DIR=%CACHE_DIR%\%APPVEYOR_REPO_BRANCH:/=_%
--- a/core/constants.yaml
+++ b/core/constants.yaml
@@ -162,10 +162,13 @@ qanet_tns: &qanet_tns
   protocolMagic: 50987900
   genesisBinSuffix: qa
 
-ci: &ci  # this mode is used by CI
+# This mode is used by CI.
+# It's only an alias for one of the existing modes.
+# It exists for convenience.
+# Don't override any values! It must be consistent with the mode used
+# by cluster!
+ci: &ci
   <<: *qanet_tns
-
-  k: 35  # it's needed for tests
 
 benchmark: &benchmark
   <<: *testnet
--- a/core/constants.yaml
+++ b/core/constants.yaml
@@ -162,7 +162,7 @@ qanet_tns: &qanet_tns
   protocolMagic: 50987900
   genesisBinSuffix: qa
 
-travis: &travis
+ci: &ci  # this mode is used by CI
   <<: *qanet_tns
 
   k: 35  # it's needed for tests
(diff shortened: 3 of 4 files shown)


gromakovsky:
Another TODO: enable `VarSpec` in appveyor.

.travis.yml
```diff
@@ -26,7 +26,7 @@ cache:
 before_install:
 - echo $TRAVIS_OS_NAME
 - export CSL_SYSTEM_TAG=$TRAVIS_OS_NAME
-- export DCONFIG=travis
+- export DCONFIG=ci_wallet
```
domenkozar: This should be `testnet_ci_wallet` so we can implement DEVOPS-226

gromakovsky: No, the idea is that it's always `ci_wallet`. If we want to change it from `qanet` to `testnet` (or whatever else) it should be done in `constants.yaml`.
`ci` is just an alias for one of existing modes.

appveyor.yml
```diff
@@ -11,7 +11,7 @@ environment:
     CSL_SYSTEM_TAG: "win64"
     CACHE_DIR: "c:\\c"
     WORK_DIR: "c:\\w"
-    DCONFIG: "qanet_tns_wallet"
+    DCONFIG: "ci_wallet"
```
jmitchell: As Domen points out for travis.yml we need DCONFIG to be prefixed with either `testnet` or `mainnet`.

gromakovsky: Will this prefix be stripped further? There is no `testnet_ci_wallet` in `constants.yaml`.

domenkozar:
I propose we come up with constants naming convention. It is really hard to currently follow on the ad-hoc naming.

Proposal:

`<mainnet/testnet>_<staging/production/dev/qanet>_<wallet/fullnode>`

Examples:

- for testnet staging we'd have `testnet_staging_wallet` and `testnet_staging_fullnode`
- for mainnet prod we'd have `mainnet_prod_wallet` and `mainnet_prod_wallet`

Is there any other factor than the three above for constants use?

gromakovsky:
`ci`/`travis` mode will be removed within this issue (CSL-1436). Hopefully it will be finished tomorrow.

Regarding everything else: I don't object, but I wasn't following these naming conventions, so it's better to get @georgeee's or @fersel's opinion.

georgeee:
<mainnet/testnet>_<staging/production/dev/qanet>_<wallet/fullnode>

@domenkozar 
Distinction `wallet/fullnode` looks ok.
I am not convinced what `mainnet_staging` would be.

As well as `testnet_(dev/qanet)`

What make sense for me:
* mainnet_<wallet/fullnode>
* testnet_production_<wallet/fullnode>
* testnet_staging_<wallet/fullnode>
* qanet_<wallet/fullnode>
* qanet_upd_<wallet/fullnode>

domenkozar:
What is upd? `mainnet` will also have staging/production, where staging will be internal testing.

georgeee:
@domenkozar `qanet_upd` is config we use to launch cluster for convenient testing of update system (it has 22.5m epoch)

domenkozar:
In that case I recommend to treat that as special internal testing. So `testnet_upd_<wallet/fullnode>`

fersel:
I party agree with both of you, in my understadning we should have the following:
* mainnet_production_<wallet/fullnode>
* mainnet_staging_<wallet/fullnode>
* testnet_production_<wallet/fullnode>
* testnet_staging_<wallet/fullnode>
* qanet_<wallet/fullnode>

I'm not sure whether we need to agree on any specific names such as `qanet_upd_<wallet/fullnode>`, but I'm fine with having one as well.

domenkozar:
If I understand correctly there is no reason to differentiate between `testnet_staging_<wallet/fullnode>` and `qanet_staging_<wallet/fullnode>`. Current difference is only in protocolMagic, but it could be just the same? Then people could test qanet with our binaries and then devops can launch a proper staging using the same wallet.

I do understand that `upd` has different constants, that's why I'd say it's a special internal testing set.

domenkozar:
@fersel explains that `qanet` has different node size. Thus let's do:

- testnet_production_<wallet/full/small>
- testnet_internal_<wallet/full/small>

And then the same thing for `s/testnet/mainnet/`. 

Where small = 4 and full = 14. What do you guys think?

fersel:
I don't think we need all that. "Small" (4 nodes) cluster deployment is only required for the `qanet` (or `testnet_internal` if we call it that) as I don't think we would need to deploy `testnet_production` with 4 nodes (that doesn't make sense after all).

So my understanding and suggestion is: we just agree on that `testnet` prefix means "full amount of nodes" while `qanet` prefix means "reduced amount of nodes" (4). Of course we'll still have `testnet_production` and `testnet_staging` separation. (So I'm basically referring to the same idea I expressed in my [previous comment](https://github.com/input-output-hk/cardano-sl/pull/1235#issuecomment-320654378) as I just see it the most convenient; that said, I'm not really strict about this).

PR #1236, opened 2017-07-27
[CSL-1203] Test delegation with distribution

naryl:
A few changes to wallet and demo launch scripts.
Integrating `readline` for proper lwallet CLI
Introducing `send-distr` command for transactions with non-default stake distribution.


georgeee:
@naryl need to:
* fix hlints
* merge master to your branch
* get macos to build green

PR #1237, opened 2017-07-27
[CSM-352] [CSM-343] Redeem tx misbehaving

Martoon-00:
_Tested version in cardano-0.5, but not here yet_


flyingleafe:
@Martoon-00 don't forget to merge `cardano-sl-0.5` with `-s ours` into this PR after #1238 is merged.

PR #1238, opened 2017-07-27
[CSM-352] Rewrite `mkCTxs` without `isRedemptionTx` flag

Martoon-00:
This is crucial fix since it fixes CSM-392 as well

### DIFF
--- a/src/Pos/Wallet/Web/ClientTypes.hs
+++ b/src/Pos/Wallet/Web/ClientTypes.hs
@@ -209,14 +209,10 @@ mkCTxs
     -> [CWAddressMeta]    -- ^ Addresses of wallet
     -> Either Text CTxs
 mkCTxs diff THEntry {..} meta wAddrMetas = do
-    ctInputAddrsNe <-
-        nonEmpty ctInputAddrs
-        `whenNothing` throwError "No input addresses in tx!"
-    let isLocalAddr = isTxLocalAddress wAddrMetas ctInputAddrsNe
-        isLocalTxOutput = isLocalAddr . addressToCId . txOutAddress
+    let isOurTxOutput = flip S.member wAddrsSet . addressToCId . txOutAddress
 
-        ownInputs = filter isLocalTxOutput inputs
-        ownOutputs = filter isLocalTxOutput outputs
+        ownInputs = filter isOurTxOutput inputs
+        ownOutputs = filter isOurTxOutput outputs
 
     when (null ownInputs && null ownOutputs) $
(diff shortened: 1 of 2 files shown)


Martoon-00:
I broke redeeming with this commit, will fix in several hours

src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -207,52 +207,43 @@ mkCTxs
...
-                mbSignificantAddrs
-            return CTx {ctIsOutgoing = isOutgoing, ..}
-        -- Output addresses whose presence makes us display the transaction
-        -- (incoming half, i.e. one with 'isOutgoing' set to @false@).
-        ctSignificantOutputAddrs =
-            ctOutputAddrs &
-            if allOutputsBelongToUs then identity else filter (not . isLocalAddr)
-        ctsOutgoing = mkCTx True $ Just ctInputAddrs
-        ctsIncoming = mkCTx False $ if isRedemptionTx then Nothing else Just ctSignificantOutputAddrs
+
```
Martoon-00: @flyingleafe I'm so glad you dealt with this mess :relaxed:

src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -207,52 +207,43 @@ mkCTxs
...
+        isIncoming = incomingMoney >= outgoingMoney
+
+        ctAmount = mkCCoin . unsafeIntegerToCoin $
+            if | isOutgoing && isIncoming -> outgoingMoney
+               | isOutgoing -> outgoingMoney - incomingMoney
+               | isIncoming -> incomingMoney - outgoingMoney
+
+        mkCTx ctIsOutgoing cond = guard cond >> pure CTx {..}
+        ctsOutgoing = mkCTx True isOutgoing
+        ctsIncoming = mkCTx False isIncoming
```
Martoon-00: In case of redeeming transaction only outgoing part remains, which redeeming expects incoming one and thus fails

Martoon-00:
I tested following scenarios
1. Redeem
2. Send money withing account (w/ and w/o remainder) - this was required because `CTx` forming logic has changed
2. Send money to foreign wallet (w/ and w/o remainder)
I checked result of `newPayment` and `redeemAda` endpoints, and after each transaction I checked results of `getAccounts (all)` and `getTxHistory` (for each related account) as well.

What differs:
* Now transaction from account `A` to `A` is displayed with amount equal to sum of balances of input addresses upon the moment transaction was performed (previously we displayed `0`)

Everything else works as before

PR #1240, opened 2017-07-28
Reorganize .gitignore

Pastafarianist:


.gitignore
```diff
@@ -1,24 +1,8 @@
...
-*.dyn_hi
-*.prof
-*.aux
-*.hp
-*.tix
 .virtualenv
 .hsenv
 .hpc
-.Werror
 *.svg
```
volhovM: Should be removed together with ipython-related things.

.gitignore
```diff
@@ -27,11 +11,7 @@ TAGS
 *~
```
volhovM: Section: emacs files.

.gitignore
```diff
@@ -10,68 +9,63 @@ cabal-dev
...
+
+# Cabal
+cabal-dev
 .cabal-sandbox/
 cabal.sandbox.config
 cabal.config
-GTAGS
-GRTAGS
-GPATH
-TAGS
```
gromakovsky: Why did you delete e. g. `TAGS`? IIRC there is some software which dumps this file and perhaps some people use it.

.gitignore
```diff
@@ -10,68 +9,63 @@ cabal-dev
...
 wallet-db/
-zhogovo/
-smartgen-secret.sk
-rocks-smartwallet/
-cardano-keys/
-bla/
-testnetStakes/
-*.key.lock
-node_modules/*
 db-abc/
```
gromakovsky: It's obsolete I suppose. Grep finds `abc` in scripts only in `scripts/avvm-files/utxo-dump-last-new.json` and `scripts/clean/db.sh`.

PR #1241, opened 2017-07-28
[CSL-1439] Use optparse-appl. command mode, implement gen-avvm-seeds

volhovM:


tools/src/keygen/KeygenOptions.hs
```diff
@@ -67,19 +77,43 @@ data FakeAvvmOptions = FakeAvvmOptions
...
+keygenCommandParser :: Parser KeygenCommand
+keygenCommandParser =
+    subparser $ mconcat $
+    [ command "rearrange"
+      (infoH rearrangeMask (progDesc "Rearrange keyfiles"))
+    , command "dump-dev-keys"
+      (infoH dumpKeys (progDesc "Dump CSL dev-mode keys"))
+    , command "generate-avvm-seeds"
+      (infoH (fmap DumpAvvmSeeds dumpAvvmSeedsParser)
+            (progDesc "Generate avvm seeds with public keys"))
```
denisshevchenko: `progDesc` is a full sentence, so I think you should add a dot at the end of it. ;-) And in all such cases too.

scripts/generate/genesis.sh
```diff
@@ -32,7 +32,7 @@ F=100 # fake avvm keys
 # print commit
 git show HEAD --oneline | tee $outputDir/genesisCreation.log
 
-keygenCmd="stack exec cardano-keygen -- --genesis-dir $outputDir -m $M -n $N --richmen-share 0.94 --testnet-stake 19072918462000000 --utxo-file $utxo_file --blacklisted $blacklisted --fake-avvm-entries $F"
+keygenCmd="stack exec cardano-keygen -- generate-genesis --genesis-dir $outputDir -m $M -n $N --richmen-share 0.94 --testnet-stake 19072918462000000 --utxo-file $utxo_file --blacklisted $blacklisted --fake-avvm-entries $F"
```
neongreen: > stack exec -- cardano-keygen OPTS 

AFAIK the order doesn't matter.

> Why generate-genesis without --?

Probably because it's a command, not option (haven't looked at the code yet, though).

PR #1243, opened 2017-07-28
[CSM-389] Modify syncTip under blkSemaphore

Martoon-00:
To avoid race conditions


Martoon-00:
@flyingleafe @fersel I tested it, `newWallet` works fine

PR #1245, opened 2017-07-30
[CSL-1428] Basic LRC tests

gromakovsky:
It's ready for review. I think I provided enough comments, but for more details you can read CSL-1428.


test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -0,0 +1,264 @@
...
+        lift $ mapM resolve richmen
+    checkPoor ::
+           (Index m ~ StakeholderId, At m) => GroupId -> m -> BlockProperty ()
+    checkPoor i richmen = do
+        let __unit = () -- workaround for hindent
+        -- It's safe because there must be 4 stakeholders by construction.
+        poorGuy <- unsafeHead <$> relevantStakeholders i
+        poorGuyStake <- lift $ fromMaybe minBound <$> GS.getRealStake poorGuy
+        totalStake <- lift GS.getRealTotalStake
+        unless (isNothing $ richmen ^. at poorGuy) $
```
volhovM: `unless (isNothing X)` is heavier than `when (isJust X)` imo.

test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -0,0 +1,264 @@
...
+            Lrc.followTheSatoshi Lrc.genesisSeed (HM.toList stableStakes)
+    unless (expectedLeadersUtxo /= leaders1) $
+        stopProperty "expectedLeadersUtxo /= leaders1"
+    unless (expectedLeadersStakes /= leaders1) $
+        stopProperty "expectedLeadersStakes /= leaders1"
+    checkRichmen
+
+checkRichmen :: BlockProperty ()
+-- Here we check richmen.  The order must be the same as the one
+-- in 'allRichmenComponents'. Unfortunately, I don't know how to
```
volhovM: Why is this obligatory?

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -103,19 +105,19 @@ import           Test.Pos.Block.Logic.Emulation (Emulation (..), runEmulation, s
...
+    , _tpAllSecrets         :: !AllSecrets
     -- ^ Secret keys corresponding to 'PubKeyAddress'es from
     -- genesis 'Utxo'.
     -- They are stored in map (with 'StakeholderId' as key) to make it easy
     -- to find 'SecretKey' corresponding to given 'StakeholderId'.
     -- In tests we often want to have inverse of 'hash' and 'toPublic'.
-    , _tpStakeDistribution :: !StakeDistribution
-    -- ^ Stake distribution which was used to generate genesis utxo.
-    -- It's primarily needed to see which distribution was used (e. g.
+    , _tpStakeDistributions :: ![StakeDistribution]
```
gromakovsky: Because in `Test.Pos.Lrc.WorkerSpec.genTestParams` I use multiple distributions. And it was convenient for debugging.

test/Test/Pos/Lrc/WorkerSpec.hs
gromakovsky: Because when we generate genesis, we use first 4 secret keys for the first group (dedicate to the first richmen component), next 4 secret keys for the second group, etc.

PR #1249, opened 2017-07-31
[CSL-1351] Boot stakeholders weights

volhovM:
This adds weights to boot stakeholders.


core/Pos/Core/Genesis/Types.hs
```diff
@@ -37,17 +46,35 @@ data StakeDistribution
...
+    | ExponentialStakes !Word -- ^ Numbers of participants
+                        !Coin -- ^ Minimal coin
     -- | Custom balances list.
     | CustomStakes [Coin]
     deriving (Show, Eq, Generic)
 
+-- | Map each participant coin in the distribution.
+stakeDistrMapCoin :: (Coin -> Coin) -> StakeDistribution -> StakeDistribution
+stakeDistrMapCoin f d = case d of
+    (FlatStakes n c)       -> FlatStakes n (applyN n c)
```
gromakovsky: I don't like it. It works only if `f(c / n) = f^n(c) / n` for all `c, n`. Which is true if `f` adds a constant, but is wrong if `f` is e. g. `unsafeMulCoin 3`.

core/Pos/Core/Genesis/Types.hs
```diff
@@ -57,24 +84,72 @@ getTotalStake (BitcoinStakes _ st) = st
...
+bootDustThreshold (GenesisWStakeholders bootHolders) =
+    -- it's safe to use it here because weights are word16 and should
+    -- be really low in production, so this sum is not going to be
+    -- even more than 10-15 coins.
+    unsafeIntegerToCoin . sum $ map fromIntegral $ HM.elems bootHolders
+
+-- | Checks whether txOutDistribution matches the set of weighted boot
+-- stakeholders. Notice: it doesn't use actual txdistr type because
+-- it's defined in txp module above.
+bootRelatedDistr :: GenesisWStakeholders -> [(StakeholderId, Coin)] -> Bool
```
gromakovsky: Why is it defined here instead of txp?

core/Pos/Core/Genesis/Types.hs
```diff
@@ -57,24 +84,72 @@ getTotalStake (BitcoinStakes _ st) = st
...
+    -- All addresses in txDistr are from bootHolders
+    all (`HM.member` bootHolders) addrs &&
+    -- Every boot stakeholder is mentioned in txOutDistr
+    all (`elem` addrs) (HM.keys bootHolders) &&
+    -- Every stakeholder gets his divisor. It's safe to use ! here
+    -- because we're already sure that bootHolders ~ addrs is
+    -- bijective.
+    all (\(a,c) -> c >= (minimumPerStakeholder HM.! a)) txOutDistr
+  where
+    addrs = map fst txOutDistr
```
gromakovsky: Why is it called `addrs` if it has type `[StakeholderId]`? I expect a variable called `addrs` to have type `[Address]`.

core/Pos/Core/Genesis/Types.hs
```diff
@@ -57,24 +84,72 @@ getTotalStake (BitcoinStakes _ st) = st
...
+
+-- | Checks whether txOutDistribution matches the set of weighted boot
+-- stakeholders. Notice: it doesn't use actual txdistr type because
+-- it's defined in txp module above.
+bootRelatedDistr :: GenesisWStakeholders -> [(StakeholderId, Coin)] -> Bool
+bootRelatedDistr (GenesisWStakeholders bootHolders) txOutDistr =
+    -- All addresses in txDistr are from bootHolders
+    all (`HM.member` bootHolders) addrs &&
+    -- Every boot stakeholder is mentioned in txOutDistr
+    all (`elem` addrs) (HM.keys bootHolders) &&
```
gromakovsky: It's inefficient and this inefficiency can be fixed in 2 lines.

src/Pos/Generator/Block/Mode.hs
```diff
@@ -181,18 +179,6 @@ mkBlockGenContext bgcParams@BlockGenParams{..} = do
...
-    bgcGenStakeholders =
-        let addrs =
-                -- So we take three stakeholders in boot era.
-                take 3 $
-                map (makePubKeyAddress . toPublic) . toList $
-                view (bgpSecrets . asSecretKeys) bgcParams
-            utxoTxHash = unsafeHash ("randomutxotx" :: Text)
-            txIns = map (TxIn utxoTxHash) [0..fromIntegral (length addrs) - 1]
-            txOuts = map (\addr -> TxOutAux (TxOut addr (mkCoin 10000)) []) addrs
-        in (mkGenesisTxpContext $ GenesisUtxo $ M.fromList $ txIns `zip` txOuts) ^. gtcStakeholders
```
gromakovsky: Wow, you removed bullshit, cool! 🎉

src/Pos/Genesis.hs
```diff
@@ -96,8 +127,8 @@ stakeDistribution (BitcoinStakes stakeholders coins) =
   where
     normalize x =
         x `unsafeMulCoin` coinToInteger (coins `divCoin` (1000 :: Int))
-stakeDistribution (ExponentialStakes n) =
-    [mkCoin (2 ^ i) | i <- [n - 1, n - 2 .. 0]]
+stakeDistribution (ExponentialStakes n (fromIntegral . unsafeGetCoin -> mc)) =
+    reverse $ map mkCoin $ take (fromIntegral n) $ iterate (*2) mc
```
gromakovsky: You can do less conversions if you use `unsafeMulCoin`.

src/Pos/Genesis.hs
```diff
@@ -136,49 +159,36 @@ genesisUtxo bootStakeholdersMaybe ad
     balances = concatMap (uncurry zip) somethingComplicated
     utxoEntry (addr, coin) =
         ( TxIn (unsafeHash addr) 0
--- Note: it's quite bad because HD wallets become stakeholders and each
--- transaction will give them quite a lot of stake.
--- But CSL-1351 is planned for this sprint, so I suppose it's fine (it's
--- unlikely to cause problems anyway).
--- TODO [CSL-1351] Delete ↓.
         , TxOutAux (TxOut addr coin) (outDistr coin))
     -- Empty distribution for PubKey address means that the owner of
```
gromakovsky: Remove this comment too.

src/Pos/Launcher/Param.hs
```diff
@@ -62,17 +62,17 @@ data NodeParams = NodeParams
...
     , npSecurityParams :: !SecurityParams       -- ^ Params for "Pos.Security"
-    , npUseNTP         :: !Bool                 -- ^ Whether to use synchronisation with NTP servers.
+    , npUseNTP         :: !Bool                 -- ^ Whether to use synchronisation with NTP servers
     , npNetwork        :: !NetworkParams        -- ^ Network parameters
-    , npEnableMetrics  :: !Bool                 -- ^ Gather runtime statistics.
-    , npEkgParams      :: !(Maybe EkgParams)    -- ^ EKG statistics monitoring.
-    , npStatsdParams   :: !(Maybe StatsdParams) -- ^ statsd statistics backend.
+    , npEnableMetrics  :: !Bool                 -- ^ Gather runtime statistics
+    , npEkgParams      :: !(Maybe EkgParams)    -- ^ EKG statistics monitoring
+    , npStatsdParams   :: !(Maybe StatsdParams) -- ^ statsd statistics backend
```
gromakovsky: Very useful changes.

src/Pos/Lrc/DB/Richmen.hs
```diff
@@ -69,7 +68,7 @@ computeInitial initialDistr proxy =
         richmenType
   where
     richmenType
-        | rcConsiderDelegated proxy = RTDelegation genesisDelegation
+        | rcConsiderDelegated proxy = RTDelegation mempty
```
gromakovsky: 😒

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -136,22 +138,22 @@ instance Buildable TestParams where
...
-genSuitableStakeDistribution :: Word -> Gen StakeDistribution
-genSuitableStakeDistribution stakeholdersNum =
-    oneof [genFlat{-, genBitcoin-}, pure (ExponentialStakes stakeholdersNum)]
+genSuitableStakeDistribution :: Coin -> Word -> Gen StakeDistribution
+genSuitableStakeDistribution dustThd stakeholdersNum =
+    -- Minimum coin in distribution should be equal or greater than
+    -- boot dust threshold.
+    stakeDistrMapCoin (`unsafeAddCoin` dustThd) <$>
+    oneof [ genFlat {-, genBitcoin-} -- is broken
+          , pure (ExponentialStakes stakeholdersNum (mkCoin 0))]
```
gromakovsky: How does `ExponentialStakes` work with `mkCoin 0`? IIRC, there is `iterate (* 2)`, so everyone should get 0 😕.

src/Pos/Genesis.hs
```diff
@@ -136,49 +159,36 @@ genesisUtxo bootStakeholdersMaybe ad
...
     -- this address will have the stake.
-    outDistr coin = maybe [] (flip genesisSplitBoot coin) bootStakeholdersMaybe
--- TODO [CSL-1351] Uncomment ↓.
---         , TxOutAux (TxOut addr coin) (genesisSplitBoot bootStakeholders coin))
---     bootStakeholdersCalculated = balancesToStakeholders balances
---     bootStakeholders =
---         fromMaybe bootStakeholdersCalculated bootStakeholdersMaybe
-
--- balancesToStakeholders :: [(Address, Coin)] -> HashMap StakeholderId Word16
--- balancesToStakeholders balances = foldr step mempty balances
```
gromakovsky: I don't get it, where is this logic? Currently you assign genesis stakeholders weights regardless of their balances. It's bad.

volhovM: I don't understand this. Can you explain why is it bad? These concepts seem orthogonal to me.

test/Test/Pos/Block/Logic/Mode.hs
volhovM: IT will work because of `stakeDistrAddCoin`.

src/Pos/Genesis.hs
gromakovsky: All stakeholders have equal weight ⇒ `genesisSplitBoot` gives them equal stakes. But we want stake to depend on `StakeDistribution`. Even if you explicitly give them correct stakes but also give them equal weights, it's still bad because further transactions will destroy this distribution.
I think my solution was the most reasonable, it solves this problem and doesn't require a caller to generate genesis stakeholders manually.

volhovM: >But we want stake to depend on StakeDistribution
No we don't. Why would we?

gromakovsky: > No we don't. Why would we? What's the point?

1. Because that's how it works now and you shouldn't change this behavior unless you have a really good reason for it. Currently I don't see a good reason, I even implemented necessary logic.
2. It's counter-intuitive to use, for example, `ExponentialStakes` and see that all nodes have equal chance to become a slot leader. Really, the main idea of all these stake distributions was to affect stakes! For instance, `BitcoinDistribution` was exactly for this reason. `RichPoor` is also for stakes (we don't care who is rich and who is poor with respect to balances).
3. **Most important actually.** If the previous two points are not enough, I can give an example: people are using `RichPoorStakes` for benchmarks (in dev mode). There are a lot of poor nodes (50000) and some rich nodes. If all of them have equal stake, there will be no richmen, which is not what we want in benchmarks.

gromakovsky: > Transactions never shift the distribution while in boot era. By design.

I mean that you can set initial stakes with respect to `StakeDistribution`, but set equal weights for all genesis stakeholders. In this case initially stakes will be correct, but transactions will shift it to equal stakes (which is not what we want). Don't do it.

core/Pos/Core/Genesis/Types.hs
gromakovsky: I am still interested in this question.

core/Pos/Arbitrary/Core.hs
```diff
@@ -443,13 +442,14 @@ instance Arbitrary G.GenesisCoreData where
                          <$> arbitrary
                          <*> pure (wordILen - a)
                          <*> arbitrary
-                , pure $ G.ExponentialStakes wordILen
+                , pure $ G.ExponentialStakes wordILen (Types.mkCoin 0)
```
gromakovsky: How is it supposed to work, btw? It will lead to nobody having stake 😕

core/Pos/Arbitrary/Core.hs
```diff
@@ -464,7 +464,7 @@ instance Arbitrary G.StakeDistribution where
            sdPoor <- choose (0, 20)
            sdPoorStake <- Types.mkCoin <$> choose (1000, 50000)
            return G.RichPoorStakes{..}
-      , G.ExponentialStakes <$> choose (0, 20)
+      , G.ExponentialStakes <$> choose (0, 20) <*> pure (Types.mkCoin 0)
```
gromakovsky: This `mkCoin 0` confuses me too.

core/Pos/Core/Genesis/Types.hs
```diff
@@ -57,24 +67,71 @@ getTotalStake (BitcoinStakes _ st) = st
...
+-- | Wrapper around weighted stakeholders map to be used in genesis
+-- core data.
+newtype GenesisWStakeholders = GenesisWStakeholders
+    { getGenesisWStakeholders :: HashMap StakeholderId Word16
+    } deriving (Show, Eq)
+
+instance Buildable GenesisWStakeholders where
+    build (GenesisWStakeholders m) =
+        bprint ("GenesisWStakeholders: "%listJson)
+               (map (sformat pairF) $ HM.toList m)
```
gromakovsky: FYI: there is `mapJson`.

src/Pos/Genesis.hs
```diff
@@ -111,74 +144,79 @@ stakeDistribution ts@RichPoorStakes {..} =
...
---     step _                            = identity
+    genesisSplitBoot' x c =
+        either (\e -> error $ "genesisUtxo can't split: " <> show e <>
+                              ", genesis utxo " <> show ad)
+               identity
+               (genesisSplitBoot x c)
+    outDistr = genesisSplitBoot' gws
 
--- | 'GenesisUtxo' used in production.
-genesisUtxoProduction :: GenesisUtxo
```
gromakovsky: Why did you delete it, btw?

src/Pos/Genesis.hs
```diff
@@ -243,12 +282,15 @@ genesisDevHdwAccountKeyDatas =
...
-devAddrDistr distr =
-    [ (mainAddrs, distr)        -- Addresses from passed stake
-    , (hdwAddresses, hdwDistr)  -- HDW addresses for testing
-    ]
+devAddrDistr :: StakeDistribution -> ([AddrDistribution], GenesisWStakeholders)
+devAddrDistr distr = (aDistr, gws)
   where
+    gws = genWStakeholders [(mainAddrs, distr)]
+--        GenesisWStakeholders $ HM.fromList $
+--          map (first addressHash) $ take distrSize genesisDevPublicKeys `zip` [1..]
```
gromakovsky: Delete this commented out stuff maybe?

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -138,36 +141,67 @@ instance Buildable TestParams where
...
         let invSecretsMap = mkInvSecretsMap secretKeysList
         let _tpAllSecrets = AllSecrets invSecretsMap
-        stakeDistribution <-
-            genSuitableStakeDistribution (fromIntegral $ length invSecretsMap)
         let addresses =
                 map (makePubKeyAddress . toPublic) (toList invSecretsMap)
-        let utxo = genesisUtxo Nothing [(addresses, stakeDistribution)]
-        let _tpGenTxpContext = mkGenesisTxpContext utxo
-        return TestParams {_tpStakeDistributions = one stakeDistribution, ..}
+        let genStakeholders =
```
gromakovsky: Why don't we use `genWStakeholders` here? It would add more versatility to tests.

gromakovsky: For example, the inconsistency between LRC and computeSharesDistr was found only because we were using `ExpontentialStakes` distribution.

test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -92,8 +94,12 @@ genTestParams = do
             (genAddressesAndDistrs totalStakeGroup (toList invSecretsMap))
             (enumerate allRichmenComponents)
     let _tpStakeDistributions = snd <$> addressesAndDistrs
-    let utxo = genesisUtxo Nothing addressesAndDistrs
-    let _tpGenTxpContext = mkGenesisTxpContext utxo
+    let genStakeholders =
+            GenesisWStakeholders $
+            HM.fromList $ map ((,1) . addressHash . toPublic) $
+            take 4 $ secretKeys
```
gromakovsky: What is it? How does it even work? I will check, I don't understand how tests can pass.

In any case, you should use `genWStakeholders` stakeholders here. Also I would wrap this combination of `genesisUtxo` and `genWStakeholders` into a single function because it seems quite useful.

gromakovsky: Oh well, tests failed, that's good. 😌

txp/Pos/Txp/Toil/Logic.hs
```diff
@@ -178,28 +176,20 @@ verifyGState curEpoch txAux txFee = do
...
     :: forall ctx m .
        ( MonadError ToilVerFailure m
-       , HasLens' ctx GenesisStakeholders
+       , HasLens' ctx GenesisWStakeholders
        , MonadReader ctx m)
     => EpochIndex -> EpochIndex -> TxAux -> m ()
 verifyBootEra curEpoch unlockEpoch txAux = do
-    let bootEra = isBootstrapEra curEpoch unlockEpoch
-    bootHolders <- views (lensOf' @GenesisStakeholders) unGenesisStakeholders
+    let bootEra = isBootstrapEra unlockEpoch curEpoch
```
gromakovsky: Actually it would be better to propagate `MonadGState` here and use `gsIsBootstrapEra` here which takes only one argument, it would allow to avoid this bug. But this issue is not about it of course.

core/Pos/Core/Genesis/Types.hs
volhovM: I didn't plan to answer it. Let's say "i don't see how it would fit there logically".

gromakovsky: It's just a consensus rule from txp. Just like any other consensus rule from there. And it's used only in one place. It fits there logically quite nicely, especially if you use actual txdistr type.

core/Pos/Arbitrary/Core.hs
volhovM: Bug, thanks.

txp/Pos/Txp/Toil/Logic.hs
gromakovsky: That's what I mean: https://github.com/input-output-hk/cardano-sl/pull/1298/files

core/Pos/Arbitrary/Core.hs
```diff
@@ -464,7 +464,7 @@ instance Arbitrary G.StakeDistribution where
            sdPoor <- choose (0, 20)
            sdPoorStake <- Types.mkCoin <$> choose (1000, 50000)
            return G.RichPoorStakes{..}
-      , G.ExponentialStakes <$> choose (0, 20)
+      , G.ExponentialStakes <$> choose (0, 20) <*> pure (Types.mkCoin 1024)
```
gromakovsky: Why is it 1024 and not something arbitrary?

core/Pos/Arbitrary/Core.hs
```diff
@@ -443,13 +442,14 @@ instance Arbitrary G.GenesisCoreData where
                          <$> arbitrary
                          <*> pure (wordILen - a)
                          <*> arbitrary
-                , pure $ G.ExponentialStakes wordILen
+                , pure $ G.ExponentialStakes wordILen (Types.mkCoin 1024)
```
gromakovsky: Why is it 1024 and not something arbitrary?

src/Pos/Launcher/Scenario.hs
```diff
@@ -66,8 +66,10 @@ runNode' plugins' = ActionSpec $ \vI sendActions -> do
                         ", address: "%build%
                         ", pk hash: "%build) pk addr pkHash
 
-    genesisStakeholders <- genesisStakeholdersM
-    logInfo $ sformat ("Genesis stakeholders: " %listJson) genesisStakeholders
+    genesisStakeholders <- view (lensOf @GenesisWStakeholders)
+    logInfo $ sformat ("Dust threshold: "%build)
+        (bootDustThreshold genesisStakeholders)
+    logInfo $ sformat ("Genesis stakeholders: " %build) genesisStakeholders
```
gromakovsky: I propose to print genesis stakeholders only if there are at most, say, 10 stakeholders. Otherwise only print number of stakeholders or something.

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -138,36 +138,59 @@ instance Buildable TestParams where
...
+--     r@RichPoorStakes{..}   -> r { sdRichStake = addMul sdRichStake sdRichmen
+--                                 , sdPoorStake = addMul sdPoorStake sdPoor
+--                                 }
+--     -- there's no way to make the fair mapping of exp stakes since
+--     -- it's defined above and 'ExponentialStakes' doesn't contain all
+--     -- the data needed to describe distr in full.
+--     ExponentialStakes n mc -> ExponentialStakes n (a `unsafeAddCoin` mc)
+--   where
+--     addMul :: Coin -> Word -> Coin
+--     addMul c n = c `unsafeAddCoin` (a `unsafeMulCoin` n)
```
gromakovsky: This code is commented out, is that ok?

test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -138,36 +138,59 @@ instance Buildable TestParams where
...
+
 -- More distributions can be added if we want (e. g. RichPoor).
 genSuitableStakeDistribution :: Word -> Gen StakeDistribution
 genSuitableStakeDistribution stakeholdersNum =
-    oneof [genFlat{-, genBitcoin-}, pure (ExponentialStakes stakeholdersNum)]
+    oneof [ genFlat
+          {-, genBitcoin-} -- is broken
+          , pure (ExponentialStakes stakeholdersNum (mkCoin 8200))
```
gromakovsky: Why 8200 (not e. g. `choose (12, 8765)`)?

test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -81,46 +84,49 @@ genTestParams = do
     secretKeys <- nonrepeating stakeholdersNum
     let invSecretsMap = mkInvSecretsMap secretKeys
     let _tpAllSecrets = AllSecrets invSecretsMap
+    r <- choose (1000::Word64, 10000)
```
gromakovsky: What is `r`? Why is it called `r`?

test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -81,46 +84,49 @@ genTestParams = do
...
     -- Total stake inside one group.
-    totalStakeGroup <-
-        (`divCoin` groupsNumber) . max minTotalStake <$> arbitrary
+    let totalStakeGroup = (`unsafeMulCoin` baseN) $ mkCoin r
```
gromakovsky: Looks ugly, why not `mkCoin r BACKTICKunsafeMulCoinBACKTICK baceN`?

test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -81,46 +84,49 @@ genTestParams = do
...
+            replicateM 2 $
+            (`unsafeMulCoin` (baseN :: Int)) . mkCoin <$>
+            choose (poorStakeN, k1)
         let richStake3 =
-                foldl'
-                    unsafeSubCoin
-                    totalStakeGroup
-                    [poorStake, richStake1, richStake2]
+                totalStakeGroup `unsafeSubCoin`
+                (foldr1 unsafeAddCoin [poorStake,richStake1,richStake2])
```
gromakovsky: How is it better than my version?

test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -81,46 +84,49 @@ genTestParams = do
...
+        mapM (genAddressesAndDistrs r totalStakeGroup (toList invSecretsMap))
+             (enumerate allRichmenComponents)
     let _tpStakeDistributions = snd <$> addressesAndDistrs
-    let utxo = genesisUtxo Nothing addressesAndDistrs
-    let _tpGenTxpContext = mkGenesisTxpContext utxo
+    let _tpGenesisContext = genesisContextImplicit addressesAndDistrs
     return TestParams {..}
   where
+    baseN :: Integral i => i
+    baseN = 1000000
```
gromakovsky: It's not clear what `baseN` is. Should be commented.

core/Pos/Core/Genesis.hs
```diff
@@ -116,34 +115,49 @@ generateHdwGenesisSecretKey =
...
+    | otherwise = Right $ bootHolders `zip` stakeCoins
   where
+    weights :: [Word64]
+    weights = map fromIntegral $ HM.elems bootWHolders
+
+    weightsSum :: Word64
+    weightsSum = sum weights
+
+    bootHolders :: [StakeholderId]
+    bootHolders = HM.keys bootWHolders
```
int-index: or just use `Map` instead of `HashMap` to make this easier to reason about?

src/Pos/Client/Txp/Util.hs
```diff
@@ -29,11 +29,17 @@ import qualified Data.Map             as M
 import qualified Data.Text.Buildable
 import qualified Data.Vector          as V
 import           Ether.Internal       (HasLens (..))
-import           Formatting           (bprint, build, sformat, stext, (%))
+import           Formatting           (bprint, stext, (%))
 import           Universum
 
 import           Pos.Binary           ()
-import           Pos.Context          (GenesisStakeholders, genesisStakeholdersM)
+-- <<<<<<< HEAD
```
rockbmb: Please don't forget to remove this before the PR is merged.

src/Pos/Genesis.hs
```diff
@@ -111,74 +147,90 @@ stakeDistribution ts@RichPoorStakes {..} =
...
---
--- TODO [CSL-1351] This documentation will become correct later.
-genesisUtxo ::
-       Maybe (HashMap StakeholderId Word16) -> [AddrDistribution] -> GenesisUtxo
-genesisUtxo bootStakeholdersMaybe ad
-    -- TODO [CSL-1351] Uncomment ↓.
-    --- | null bootStakeholders =
-    ---     error "genesisUtxo: no stakeholders for the bootstrap era"
-    -- TODO [CSL-1351] Delete (it's here to please hlint) ↓.
-    | ((*) @Int) 2 3 == 7 = error "hlint — makaka"
```
rockbmb: :joy:

test/Test/Pos/Block/Logic/Mode.hs
volhovM: Should be removed, thanks!

test/Test/Pos/Block/Logic/Mode.hs
volhovM: Same reason i've described earlier. It's `8192 + eps` :peka:

test/Test/Pos/Lrc/WorkerSpec.hs
volhovM: Definitions from my notebook :unamused:

test/Test/Pos/Lrc/WorkerSpec.hs
volhovM: ``mkCoin  r `unsafeMulCoin` baseN``? :peka: 
thanks

src/Pos/Genesis.hs
```diff
@@ -221,7 +273,8 @@ devStakesDistr Nothing Nothing (Just (richs, poors, coins, richShare)) Nothing =
         then error "Impossible to make RichPoorStakes with given parameters."
         else identity
 devStakesDistr Nothing Nothing Nothing (Just n) =
-    ExponentialStakes (fromIntegral n)
+    -- minimum stake allows to have max 11 weighted stakeholders in boot era
+    ExponentialStakes (fromIntegral n) (mkCoin 2048)
```
rockbmb: Again, maybe better to have this as a top-level value in this module in case it's used again.

test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -81,46 +84,49 @@ genTestParams = do
...
-        let poorStake = thresholdCoin `unsafeSubCoin` mkCoin 1
-        -- Let's add small stake to two richmen just for fun.
-        let genSmallRichStake =
-                unsafeAddCoin thresholdCoin . mkCoin <$> choose (0, 10)
-        richStake1 <- genSmallRichStake
-        richStake2 <- genSmallRichStake
+        let thresholdN = unsafeGetCoin thresholdCoin
+            -- m1
+        let poorStakeN = thresholdN `div` baseN
+            -- P1
```
rockbmb: `m1, P1`, maybe a very short explanation would be nice.

core/Pos/Arbitrary/Core.hs
gromakovsky: I still don't understand anything. Why can't we use `choose (1, 1024)` instead of `1024`, for example?
I know that it's unlikely to matter, but who knows, maybe tests will always pass with 1024 and fail with 1023 (or 42).

volhovM: Because low values impose implicit restrictions on stakes list length.

core/Pos/Arbitrary/Core.hs
volhovM: Exponential stakes are cancer and should be removed at all :|

gromakovsky: No, they are not. They found a bug few weeks ago.

core/Pos/Arbitrary/Core.hs
gromakovsky: 1. Which restrictions? What will go wrong if the value is 3?
2. Why not `choose (1000, 2000)` then?

volhovM: Please read my first comment in this discussion. It answers both of your questions. In particular:

>We also require that no genesis balance is lower than sum of weights. 

If you set to `3` or to `1000` genesisUtxo won't be created because there's a balance that is less than `bootDustThd`.

test/Test/Pos/Lrc/WorkerSpec.hs
gromakovsky: How to write it?

avieth:
Running the demo launch script with rich/poor fails with

```
cardano-node-simple: generateWStakeholders can't convert: too big 1650000
```

The distribution is

```
--rich-poor-distr "(3,50000,6000000000,0.99)"
```

Does this need to change? I've blindly tried some changes but usually end up with another error `genesisUtxo can't split`.

PR #1250, opened 2017-08-01
[CSL-721] Write some clarifying comments about 'AsBinary'

gromakovsky:


core/Pos/Binary/Crypto.hs
```diff
@@ -87,12 +88,23 @@ instance Bi SecretSharingExtra where
...
+-- deserialization is length check. As long as length matches our
+-- expectations, the decoding succeeds (look at 'Binary' instances in
+-- 'pvss') which in turn means that we can use 'fromBinary' and be
+-- quite sure it will succeed. That's why it's important to check
+-- length here (this check is cheap, so it's good to do it as soon as
+-- possible). The only exception is 'SecretSharingExtra' (because we
+-- don't know its length in advance). Currently we check that
+-- 'SecretSharingExtra' can be parsed in 'verifyCommitment' function.
+--
+-- FIXME [CSL-1412] Check length here (the note above is currently incorrect)!
```
adinapoli-iohk: @gromakovsky Just as an heads-up, I'm planning to tackle https://issues.serokell.io/issue/CSL-1412 today. It would be awesome if you could chime in to that ticket and review the solution I'm proposing. Up to you then if you want to include this comment in this PR (I don't have a strong opinion).

gromakovsky: Well, it depends on whose PR is merged first. If yours, I will remove this line. If mine, you will need to remove it. I just don't want to include a comment which is not correct right now, but will become correct very soon without such note.

adinapoli-iohk: Fair point. Go for it then 😉

gromakovsky:
Well, I was more lucky, builds in my PR passed and in yours they timed out :)

PR #1252, opened 2017-08-01
Add vim tags to gitignore

int-index:


int-index:
Who deleted it? :unamused:

gromakovsky:
That guy.

PR #1253, opened 2017-08-01
[CSL-1448] Remove `Unused` constructor from deriveSimpleBi

adinapoli-iohk:
See https://issues.serokell.io/issue/CSL-1448 for context.

After consulting with @neongreen , we opted for removing `Unused` and its logic from the `Binary.Class.TH` module, as it's incomplete and we can always retrieve it back from `git` history when and if we decide to resume working on it.


adinapoli-iohk:
Kicked off the CI server once again, finger crossed!

PR #1256, opened 2017-08-02
[CSL-1247] Put Explorer into CSL source tree for consistency

ksaric:
https://issues.serokell.io/issue/CSL-1247

A new PR, started here - https://github.com/input-output-hk/cardano-sl/pull/1149

Used `nix-shell` on `cardano-sl` root to check if the build instructions are good.
Added to the current `.travis.yml`, currently, the output we get fails:
```
- pushd explorer
# I'm guessing we are missing something like "nix-build -A cardano-sl-explorer ..."
- pushd frontend
- "./scripts/build.sh"
- nix-shell --run "yarn test"
- popd
- "./update_explorer_web_api_docs.sh"
- popd
```


ksaric:
Other than this problem with space - https://travis-ci.org/input-output-hk/cardano-sl/jobs/260985531#L7738, time limit for MacOs - https://travis-ci.org/input-output-hk/cardano-sl/jobs/260985532#L6657 and required bump for explorer, seems to be ok.

I need help to bump _explorer_ since there are a lot of changes (I already spent a couple of hours on this, and the number of changes is big) - we should merge this ASAP, since `master` is changing every 100500 milliseconds.
We can fix any details when we have it in the repo and create branches for it like we had so far - `explorer-0.5`, `explorer-master` - or similar.

@domenkozar @gromakovsky @akegalj

stack.yaml
```diff
@@ -131,3 +142,11 @@ ghc-options:
   cardano-sl:            -Werror
   cardano-sl-tools:      -Werror
   cardano-sl-lwallet:    -Werror
+  cardano-sl-explorer:   -Werror
+
+# Added for explorer
+flags:
+  cardano-sl:
+    with-explorer: True
```
gromakovsky: This is wrong. If we want to build explorer version by default, then we should change default value of `with-explorer` in `cardano-sl.cabal`. But I think we shouldn't build explorer version by default. It should be handled by a script which builds explorer.

gromakovsky:
Did you use `git subtree` in this PR too?

explorer/CODEOWNERS
```diff
@@ -0,0 +1,13 @@
+# Unless a later match takes precedence, @akegalj will be requested
```
gromakovsky: Does it work as expected? For example does `src` mean `explorer/src`? Is this file considered at all (given that it's no longer in the root)?

gromakovsky: Well, I guess it should work, but it would be good to know for sure.

src/Pos/Util/JsonLog.hs
```diff
@@ -177,3 +177,6 @@ jsonLogDefault x = do
     JL.jsonLogDefault jlc x
 
 deriving instance CanJsonLog (t m) => CanJsonLog (TaggedTrans tag t m)
+
+-- Required for @Explorer@ @BListener@ redirect
+deriving instance CanJsonLog m => CanJsonLog (TaggedTrans tag IdentityT m)
```
gromakovsky: It looks quite strange. Doesn't it conflict with the previous line? Anyway, if it compiles, I don't mind leaving it here.

ksaric:
> Did you use git subtree in this PR too?

Yes. It's one of the reasons why I would prefer to merge this quickly, otherwise, we might have issues with history.

stack.yaml
ksaric: AFAIU, if we want Stack to compile _explorer_, which is something that we do want - we have a single `stack.yaml` and need to compile all the dependencies in one go.

> It should be handled by a script which builds explorer.

But where would that script be? Seems kind of counter-intuitive that we have a build that builds `cardano-sl` and then we have a script that builds `explorer`. 
We do want to catch any local errors in `explorer` when building, right?

explorer/CODEOWNERS
ksaric: We'll find out eventually :smile:

src/Pos/Util/JsonLog.hs
ksaric: Yeah, seems the previous more general instance doesn't quite work. I added the additional one since we need that one for _explorer_. I guess I could remove the previous instance, and yes, I don't know why it doesn't work.

stack.yaml
akegalj: @ksaric
Do you want me to help adding explorer into cardano-sl.sh script?

ksaric: @akegalj No, I would prefer if we all agreed on what is it that we need to do here first, seems we each have a different point of view.
I still think that @domenkozar wanted to have a _explorer_ build in Stack, since we asked me to merge the two Stack files. Maybe he can clarify?

gromakovsky: > AFAIU, if we want Stack to compile explorer, which is something that we do want - we have a single stack.yaml and need to compile all the dependencies in one go.

Yes. Once again, if you want to change a default value of `with-explorer`, it should be done in `cardano-sl.cabal`, because it's simpler (otherwise it's not obvious which one will be used by default: the one from cabal file or the one from stack.yaml). That's the first point.
The second point is that I don't like changing the default value. The build without explorer should be used more often (we need to build explorer less often than cluster nodes or wallets). If one wants to build explorer and doesn't enable `with-explorer`, the build will fail which is good because we will know about misconfiguration. However, if one wants to build a regular node and enables `with-explorer` the build will pass, but the node will have explorer part (which is not critical, but is still bad, because it adds some overhead and extra possibility of a bug). Also I don't think it's a good idea to change the default behavior unless there is a very good reason for it (currently I don't see one).

> But where would that script be? 

There are two scripts: one for developers and one for CI. For developers I think you should just modify `scripts/build/cardano-sl.sh` as I already said. Regarding CI, I am not sure what our current plan is, but in the simplest case you can just pass `with-explorer` flag from there.

ksaric: Removing these flags results in (gold) linker issues that we had with _explorer_. Removing gold linker from _explorer_ is an option, too, I guess.

ksaric: If we do want to compile without _explorer_ and have it as a submodule, we need to remove the linker flags here - https://github.com/input-output-hk/cardano-sl/blob/6c8d85aea1dc4f980876668452130b38a3b2d5b3/cardano-sl.cabal#L450-L454 which would remove gold-linker optimizations from `cardano-sl`. The alternative? Switch the flag to true in `cardano-sl.cabal`, remove from `stack.yaml`.

ksaric: Under `stack.yaml` packages I added `explorer` - which pulls the `cardano-sl-explorer.cabal` and everything in it. That means it tries to build `explorer` by default, and if we disable the flag `with-explorer` then it still builds but it raises the gold-linker issue we are familiar with.

gromakovsky: 1. We want to be able to compile with explorer and without explorer.
2. Building `cardano-sl-explorer` without passing `with-explorer` flag to `cardano-sl` is wrong, don't do it. Build script should detect whether we want to build explorer and enable `with-explorer` flag if we do and disable it if we don't. Regarding CI: I don't know which approach will be used, it should be discussed with devops, ideally we should build both version.
3. `with-explorer` flag is a dirty and annoying hack which shouldn't exist, but nobody knows when we will find time to remove it. Maybe it's time to do it.
4. Regarding default value: personally I don't really care as long as `build/cardano-sl.sh` has the desired behavior. For example, if I use `scripts/build/cardano-sl.sh sl` it should build without explorer (as it's done now). If I use `scripts/build/cardano-sl.sh` without any arguments, it means that I want to build everything, so it should figure out that I want explorer and `with-explorer` should be enabled.

P. S. Gold linker issues are irrelevant, they are already solved. If you build explorer you must enable `with-explorer` for `cardano-sl`. In this case everything should work.

ksaric:
@gromakovsky Ok, so we currently can't disable that stupid _explorer_ flag due to linker issues. We have to turn it on by default, which I did while removing it from `stack.yaml`.

Since that is how it's going to work (for now), your second review point works already - _explorer_ is being built with the "normal" script. If you pass `--explorer`, nothing changes and I would leave that as is since it's probably we are going to fix that sometime in the future (there is a patch on GHC https://ghc.haskell.org/trac/ghc/ticket/13810). 

We need to notify everyone that they are building `explorer` with `cardano-sl` now as well.

I know it's not great, but it's not terrible too - it could have been worse.

stack.yaml
ksaric: 1. Yes, that's the current issue.
2. The things is, there is no point of passing `with-explorer` since we need it in both cases - it's due to gold linker issues. As you wrote yourself, we must enable `with-explorer` for `cardano-sl`.
3. I can do it, but it's connected to the next point.
4. Ok, I can fix this, but I want to be sure we understand each other. We have three code chunks that we can group for _explorer_.
a. the base `cardano-sl` code.
b. the base `cardano-sl` code related to `explorer`, which we enable/disable using the stupid flag.
c. the actual `explorer` code. This is being built on its own since it's in `stack.yaml` `packages`

The issues seem to come from b. but we are actually already building _explorer_ due to c. So, what I can see as a solution for 3. is to remove the flag entirely.

We can't build this project with `explorer` inside it if we pass `False` as `with-explorer`. If I return the flag `with-explorer` to _False_.
When I don't pass a flag:
```    /home/ksaric/.nix-profile/bin/ld.gold: --hash-size=31: unknown option
    /home/ksaric/.nix-profile/bin/ld.gold: use the --help option for usage information
    collect2: error: ld returned 1 exit status
    `gcc' failed in phase `Linker'. (Exit code: 1)
```
When I pass a flag with `stack build --fast -j4 --flag cardano-sl:with-explorer`, all good.

gromakovsky: 2. No, it's wrong. I even checked it by myself. I cloned your branch, changed default value of `with-explorer` to `False`, used `scripts/build/cardano-sl.sh sl` and it worked fine.

I don't mind if explorer is enabled by default, but I want some changes in build scripts. They are very similar to what I described here: https://serokell.slack.com/archives/G3K2W68AK/p1502200697021298?thread_ts=1502193530.814070&cid=G3K2W68AK
1. You should consider wallet in build script as it's done here: https://github.com/input-output-hk/cardano-sl/pull/1274/files#diff-b4002f72f6a664eaf30722a486b80570 (clean should clean it, it should be possible to build exactly explorer, etc.).
2. `--explorer` should be changed to `--no-explorer` because explorer will be enabled by default. If explorer is built, there should be an error message if `--no-explorer` is passed. But of explorer is not built, `--no-explorer` means that `cardano-sl:-with-explorer` should be used (i. e. `with-explorer` should be disabled).

After that my only concern will be CI which should be discussed with devops I think.

ksaric: @gromakovsky I was talking about using `stack build ...` as a way to build the project. Can you try using that? Without that, I can't use `ghcid` and no `stack ghci`. I'm not sure about other things like editor plugins/build tools and such, but could very well be affected. 
_And that's because using `stack build ...` we implicitly include _explorer_ since it's in `packages`._

I also tried setting `with-explorer` to `False`, used `./scripts/build/cardano-sl.sh sl`. That went fine and compiled. That's not the issue. The issue is **building both projects** without setting the flag to true.

How are we going to build `explorer` with that flag set to `False`? And I don't mean the b. part of the `explorer`. I mean the c. part of the `explorer`.

ksaric:
As a note, we run into disc space issues when building using nix - https://travis-ci.org/input-output-hk/cardano-sl/jobs/262304063#L7720
On the other hand - https://travis-ci.org/input-output-hk/cardano-sl/jobs/262304071 :smiley:

stack.yaml
gromakovsky: > I was talking about using stack build ... as a way to build the project. 

It doesn't matter `scripts/build/cardano-sl.sh` is just a wrapper over `stack build`.

> And that's because using stack build ... we implicitly include explorer since it's in packages.

It depends on which target you pass to `stack build`. If you use `stack build cardano-sl` it won't built explorer.

> The issue is building both projects without setting the flag to true.
> How are we going to build explorer with that flag set to False?

Once again, you should never build explorer without this flag. And I described it above. If one passes `--no-explorer` to build script, but wants to build explorer, it should be an error printed to stderr.

gromakovsky:
Btw, you should also update some other scripts (e. g. `grep.sh`).

stack.yaml
ksaric: @gromakovsky And you want me to add a flag for not building _explorer_?

So do you want me to change `scripts/build/cardano-sl.sh` so it contains `explorer` build by default? If I add `--no-explorer` it should not build parts b. and c., and the flag can be set to `False`, right? And you want me to add `explorer` c. part build into the script?

- [ ] Did I get this finally?

gromakovsky: > And you want me to add a flag for not building explorer?

Not for building _explorer_ but for building `cardano-sl` with `with-explorer` disabled. I. e. not building parts b. and c. But it should work only if we build explorer, otherwise the error message should be printed!

[Here](https://github.com/input-output-hk/cardano-sl/blob/49e3805ae41aacc8ed664ab3817549fb0f63e7c6/scripts/build/cardano-sl.sh#L254) you know what you are going to build. If explorer is one of the targets, `--no-explorer` is prohibited. If it's not, then `--no-explorer` is fine.

> So do you want me to change scripts/build/cardano-sl.sh so it contains explorer build by default?

Without arguments it should be everything one-by-one. `sl+` argument should also build explorer. `explorer` argument should build only explorer (and its deps).
`exporer` should be treated similar to `lwallet` and `tools`.

ksaric:
@gromakovsky Ok, so I fixed things as you suggested. I still don't like that explorer flag and find it annoying. It could be removed and some (unnecessary/extra) cognitive workload can be removed with it.

gromakovsky:
Nobody likes this flag, but it's not easy to remove it.

scripts/build/cardano-sl.sh
```diff
@@ -245,27 +254,40 @@ elif [[ $spec_prj == "sl" ]]; then
   to_build="cardano-sl"
 elif [[ $spec_prj == "lwallet" ]]; then
   to_build="cardano-sl-lwallet"
+elif [[ $spec_prj == "explorer" ]]; then
+  to_build="cardano-sl-explorer"
 elif [[ $spec_prj == "sl+" ]]; then
   to_build="cardano-sl cardano-sl-lwallet cardano-sl-tools"
```
gromakovsky: `cardano-sl-explorer` should be here.

ksaric: Look at https://github.com/input-output-hk/cardano-sl/pull/1256/files/c6f49b61a617d39aaba9728d45ccd2ec674bfe76#diff-b4002f72f6a664eaf30722a486b80570R267

scripts/build/cardano-sl.sh
```diff
@@ -245,27 +254,40 @@ elif [[ $spec_prj == "sl" ]]; then
...
 else
   to_build="cardano-sl-$spec_prj"
 fi
 
+# If the use didn't specify that he doesn't want explicitly to exclude Explorer, build it
```
gromakovsky: No, it shouldn't be here. If I pass only `sl` as argument, I want to build parts (a) and (b), but not (c). But with this logic also (c) will be built.

ksaric:
@gromakovsky Well, if I move things as you pointed out, then we are back with linker issues. 
For example, what's the result of `./scripts/build/cardano-sl.sh --no-explorer`?
It enables c part (the actual _explorer_) but disables the flag. Can I just leave it as is? It's good now.

gromakovsky:
> For example, what's the result of ./scripts/build/cardano-sl.sh --no-explorer?

The result is:
```
echo "You can't build output with explorer and not use explorer! Invalid flag '--no-explorer'."
exit
```

ksaric:
@gromakovsky Ok, fixed.

ksaric:
@gromakovsky @domenkozar @akegalj 
It's done. Thanks guys for your help!

:rocket:

ksaric:
BTW, it would be useful if we could branch out from master to have a stable version like we had before.
So branches like `sl-explorer-0.5`, `sl-explorer-master`, or something similar.

PR #1257, opened 2017-08-02
[CSL-1430] Make generate-key and read-key

naryl:
It can generate a primary key and display information about a key. Please check if that's the required format:
```
$ stack exec -- cardano-keygen read-key --path something.key
[keygen:INFO:ThreadId 4] Processing command
[keygen:INFO:ThreadId 4] Primary: pub:60ced4ca; address hash: 4d2f0122
[keygen:INFO:ThreadId 4] No wallet set
[keygen:INFO:ThreadId 4] Keys: 
[keygen:INFO:ThreadId 4] Vss PK: PublicKey Point 99593608784699692652544980421310389590078448882169371488268607087261604356744 92169266176351874303430525274811642215745492274510204282518003720616292214609
```
I didn't find a better way to display PVSS's keys.


naryl:
No. Will do before merging.

georgeee:
`add-key` command from wallet CLI will help you!

tools/src/keygen/Main.hs
```diff
@@ -146,6 +152,37 @@ getAvvmGenesis AvvmStakeOptions {..} = do
 rearrange :: (MonadIO m, MonadFail m, WithLogger m) => FilePath -> m ()
 rearrange msk = mapM_ rearrangeKeyfile =<< liftIO (glob msk)
 
+genPrimaryKey :: (MonadIO m, MonadFail m, WithLogger m) => FilePath -> m ()
+genPrimaryKey path = generateKeyfile True Nothing path >> pure ()
```
volhovM: `void $ generateKeyfile ...`
And yes, some `logInfo "successfully generated X"` would be good.

tools/src/keygen/Main.hs
```diff
@@ -146,6 +152,37 @@ getAvvmGenesis AvvmStakeOptions {..} = do
...
+                           map (showKeyWithAddressHash . decryptESK) $
+                           view usKeys us)
+    logInfo $ maybe "No vss"
+                    (("Vss PK: " <>) . showPvssKey) $
+                    view usVss us
+
+showKeyWithAddressHash :: SecretKey -> Text
+showKeyWithAddressHash sk = sformat (build%"; address hash: "%build) pk ah
+  where
+    pk = toPublic sk
```
volhovM: In general there's no sense to have `pk` and `ah` here because they are trivial and only used once, but it's minor.

naryl: pk is used twice. First for output, then to get ah.

PR #1192, opened 2017-07-19
[DEVOPS-203] Convert travis jobs to use Nix instead of stack

cleverca22:

PR #1258, opened 2017-08-02
[CSLD-119] Enable swagger & client doc autobuild by travis

Martoon-00:
My only purpose is to update https://cardanodocs.com/technical/wallet/api/
And this PR seems to be the easiest way to do that

### DIFF
--- a/scripts/ci/travis.sh
+++ b/scripts/ci/travis.sh
@@ -70,11 +70,11 @@ for trgt in $targets; do
 
 done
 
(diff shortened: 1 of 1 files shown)


domenkozar:
Afaik haddock is not generated because it took too long. This will need to wait on https://github.com/input-output-hk/cardano-sl/pull/1192

Martoon-00:
I omitted haddock and remained executables-based documentation builders, hope them to be very fast

Martoon-00:
Okay, I've just found out that travis job limit is already fully depleted, chances that this will pass are phantom

PR #1260, opened 2017-08-03
[CSL-217 (CSM-397)] [CSM-374] Web stuff

Martoon-00:
CSL-217: added a flag into constants, which enables/disables requests logging. Tested w and w/o this flag set.
Currently flag is always on, because we have no mainnet mode.

CSM-374: Added type with TLS parameters

### DIFF
--- a/src/node/NodeOptions.hs
+++ b/src/node/NodeOptions.hs
@@ -18,8 +18,8 @@ import           Options.Applicative.Simple   (Parser, auto, execParser, footerD
                                                value)
 import           Prelude                      (show)
 import           Serokell.Util.OptParse       (fromParsec)
-import           Text.PrettyPrint.ANSI.Leijen (Doc)
 import qualified Text.Parsec.Char             as P
+import           Text.PrettyPrint.ANSI.Leijen (Doc)
 import           Universum                    hiding (show)
 
 import           Paths_cardano_sl             (version)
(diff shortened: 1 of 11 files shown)


src/node/NodeOptions.hs
```diff
@@ -222,6 +207,25 @@ addrNodeOption =
         metavar "HOST:PORT" <>
         help "Identifier of a node in a Kademlia network"
 
+tlsParamsOption :: Parser TlsParams
```
gromakovsky: It should be moved to `Pos.CLI` and used in `lwallet` too.

Martoon-00: Light wallet seems not to use tls

gromakovsky: I see this https://github.com/input-output-hk/cardano-sl/blob/master/lwallet/WalletOptions.hs#L97

Martoon-00: Thanks, I missed that

Martoon-00: @akegalj Looks like TLS params are added to light wallet options, but never used. Is it intentional?

Martoon-00:
What frustrates me, is that we can easily make TLS disable-able, but it's not clear how should that look from executable parameters point of view.
I would like to enable tls if all the parameters are given, disable of none given, and throw an error otherwise. But given library doesn't provide such option (or I failed to find it).

Well, maybe that's actually not a good idea anyway, since it allows an adversary to replace launch params and thus disable TLS

akegalj:
> Well, maybe that's actually not a good idea anyway, since it allows an adversary to replace launch params and thus disable TLS
yes, I would rather have it enabled by default and not have this option (even in dev mode)

src/node/NodeOptions.hs
akegalj: > Looks like TLS params are added to light wallet options, but never used. Is it intentional?

I was not sure how light wallet will be used. As it is not used by Daedalus atm I think we can remove it. In this case light wallet should not be installed on clients machine when daedalus is installed with installers. If it is installed, some users might want to use this instead of official Daedalus (ie, for hacking around) and they might get screwed if TLS is not on there as adversary can read the traffic there - and make some actions ? (not sure about this one)

So in the lack of understanding how light wallet works, what is its aim, and will it be installed on users machine or not I have left TLS on.

akegalj: @Martoon-00 how is light wallet working ? Do we have some http endpoints open where light wallet can do some actions without TLS on ? If that is the case I would turn TLS on there

Martoon-00: @akegalj I also have little understanding of what's going on there, but after looking at this module I got quite sure what lwallet doesn't provide any endpoints:
https://github.com/input-output-hk/cardano-sl/blob/master/lwallet/Pos/Wallet/Light/Web/Server.hs

akegalj: @Martoon-00 aha, light wallet doesn't have any implementation as of yet (at least from the code you have pasted). That being said, we can then ignore TLS there

PR #1261, opened 2017-08-03
[CSL-1444] Fix 'isInitialized' check in Issuers part of LRC DB

gromakovsky:
Apart from this fix I also performed a small refactoring during initial investigation.


pva701:
Did you launch node with existing DB? does it work now?

gromakovsky:
Yes, yes.

pva701:
Segodnya u kogo-to budet klaster cc @fersel

PR #1264, opened 2017-08-04
[CSL-1453] Make benchmarks compile conditionally

rockbmb:


appveyor.yml
```diff
@@ -99,6 +99,7 @@ test_script:
       --flag cardano-sl:with-wallet
       --flag cardano-sl-core:-asserts
       --flag cardano-sl-core:-dev-mode
+      --flag cardano-sl:bench
```
adinapoli-iohk: Genuine question: how come you choose to kept building the benches on Windows? Isn't less build time always a good time for CI? 😉

cardano-sl.cabal
```diff
@@ -30,6 +30,12 @@ Flag with-explorer
 
   description: Build with explorer
 
+flag bench
```
georgeee: So `flag` was actually needed, just not passing `--bench` wasn't sufficient to not build benchmark dependencies?

appveyor.yml
georgeee: I believe we had some problems with appveyor with benchs off. But @rockbmb, please try disabling it, maybe problems don't exist any more.

gromakovsky: But, the advantage of building benchmarks on Windows is that we will ensure they still compile. So if they don't harm much for Windows, why not enable them?
Anyway, our main bottleneck is Travis build times, not appveyor (in 99% cases appveyor finishes before).

cardano-sl.cabal
gromakovsky: `criterion` is still being built. And @rockbmb reports that it's because `criterion` is a transitive dependency of something else. So maybe stack actually can infer that it doesn't need to build benchmarks dependencies and we don't need this flag (which is extra complication).

adinapoli-iohk:
> So if they don't harm much for Windows, why not enable them?

My 2 cents: we should really have a separate Travis task dedicated to do just that (maybe regardless to the environment, ideally `nix`) so that we can keep each task relatively self-contained and fast to build 😉

domenkozar:
> My 2 cents: we should really have a separate Travis task dedicated to do just that (maybe regardless to the environment, ideally nix) so that we can keep each task relatively self-contained and fast to build :wink:

The problem is, it takes 40min to compile cardano-sl and then 5min to run the benchmarks. If you split that in two, you'll have almost double compilation times. Then we'd need to increase our workers for appveyor.

adinapoli-iohk:
> The problem is, it takes 40min to compile cardano-sl and then 5min to run the benchmarks.

Well, we wouldn't need to run the benches (not yet, at least), but ensure the code didn't rot and we can still compile it, yes. But I thought that with `nix` we were able to share dependencies (aka caching of those) between tasks of the same job? 😄

cardano-sl.cabal
georgeee: I checked latest logs from CI: no more criterion. But first PR against time-warp-nt should be merged, then I still want it to be tried whether we can get rid of this `flag bench`.  Maybe after all changes we can simply not pass `--bench` to build.

georgeee:
Also, @rockbmb, please add `--bench --no-run-benchmarks` (or whatever name these flags have) to `travis.sh`, to be run only for *linux* (not *osx*!)

stack.yaml
```diff
@@ -52,7 +52,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: e1bc23f12c388301ceae4de53fe2d513057b62ec # master
+    commit: 02e537e97ecb0ce384970cb2e7f13e4a5c09fc74 # csl1453-benches
```
gromakovsky: I've merged your PR, so please fix it to point to master.

cardano-sl.cabal
gromakovsky: I pushed a branch without this flag, the build hasn't started yet but will start sooner or later.
It's here https://travis-ci.org/input-output-hk/cardano-sl/builds/261750906

gromakovsky: https://travis-ci.org/input-output-hk/cardano-sl/builds/261770076

PR #1265, opened 2017-08-04
Use master time-warp-nt revision

avieth:
#1116 used a feature branch, but we just brought that into master.

It's not tip of master, as tip of master has breaking changes [CSL-849]

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -52,7 +52,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: c901b25bdfcc121cf5157952f7cc3f80f46e902a #feature/multiple-peers-buckets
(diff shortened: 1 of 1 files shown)


avieth:
Appveyor failed because of a timeout trying to download 7z

stack.yaml
```diff
@@ -52,7 +52,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: c901b25bdfcc121cf5157952f7cc3f80f46e902a #feature/multiple-peers-buckets
+    commit: e1bc23f12c388301ceae4de53fe2d513057b62ec
```
gromakovsky: Can you also add a comment that it's master?

(this way you will also restart appveyor)

PR #1267, opened 2017-08-04
derive CanJsonLog (TaggedTrans tag t m) instance

avieth:
It used to be derived in time-warp-nt, but was removed along with the
dependency on ether.


ksaric:
Mac build failing (linker issues?), restarted, but I think it's good enough.

PR #1268, opened 2017-08-04
[CSL-1453] Remove optparse-{simple,text} & scotty from cardano

adinapoli-iohk:
This juicy PR removes 3 deps and their direct "unique deps". In particular, we are dropping:

* `nats`
* `regex-compat`
* `scotty`
* `optparse-simple`
* `either`
* `gitrev`
* `optparse-text`

This required depending on the latest version of the `cardano-report-server`. Last but not least, I had to remove `turtle` (again) as it was used by the `cardano-postmortem` tool.


neongreen:
I updated YT.

PR #1270, opened 2017-08-04
[CSM-378] Move tx fees logic to `Pos.Client.Txp`

flyingleafe:
Transaction creation logic has been simplified, and tx fees logic (together with boot distribution logic) was moved inside `create*Tx` functions.

There's some ugly decisions in this PR which made it possible to change logic without profound refactoring. Though I find these decisions ugly, I'd better not include such refactoring in this particular PR and implement it separately - because we need to start testing this logic (including making unit tests, CSM-399) asap.

Also, changing fees constants is not present in this PR, because it's present in #1269.

This PR has WIP label because it needs to be tested properly before merging (including Daedalus acceptance tests). PR is made in advance to give more time for review.


src/Pos/Client/Txp/Addresses.hs
```diff
@@ -0,0 +1,14 @@
...
+       ) where
+
+import           Universum
+
+import           Pos.Types (Address)
+
+-- | A class which have the method to generate a new address
+class Monad m => MonadAddresses m where
+    type AddrData m :: *
+    getNewAddress :: AddrData m -> m Address
```
flyingleafe: I really don't like this. I'd rather have `getNewAddress :: m Address`, but the only way for implementing such instance for real wallet which I could think of is having some `newtype UglyWrapper m = UglyWrapper (ReaderT addrData m) deriving (<thousands of monads>)` and implementing an instance for it. Though we used to do such things, now it's a no-way, as I understand. Is there a better way to implement such a typeclass?

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -586,27 +548,20 @@ getMoneySourceWallet (AddressMoneySource addrId) = cwamWId addrId
...
-        InternalError . sformat ("UnknownFeePolicy, tag: "%build)
+getMoneySourceUtxo :: WalletWebMode m => MoneySource -> m Utxo
+getMoneySourceUtxo =
+    getMoneySourceAddresses >=>
+    mapM (decodeCIdOrFail . cwamWId) >=>
+    getOwnUtxos
+
+-- [CSM-237] It should be moved to `Pos.Wallet.Web.Mode`, but
+-- to make it possible all this mess should be neatly separated
+-- to modules and refactored
```
flyingleafe: P.S: it's because of imports which are really hard to be made non-circular in this situation.

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +260,184 @@ prepareInpsOuts utxo addrs outputs = do
...
+        -- Fee depends on size of tx in bytes, sign of a tx has the fixed size
+        -- so we can use arbitrary signer.
+        srcAddrs = NE.map fst trInputs
+        (_, fakeSK) = deterministicKeyGen "patakbardaqskovoroda228pva1488kk"
+        hdwSigners = NE.zip (NE.repeat $ fakeSigner fakeSK) srcAddrs
+    in makeMPubKeyTxAddrs hdwSigners trInputs txOutsWithRem
+
+    -- buildDistribution :: TxRaw -> Text
+    -- buildDistribution TxRaw{..} =
+    --     let entries =
```
flyingleafe: This function was used only for logging in the old code. Now the code, identical to old one, is run in environment without `WithLogger`. Is it better to add `WithLogger` to `TxDistrMode` or can we simply abandon this log line?

lwallet/Main.hs
```diff
@@ -217,9 +218,10 @@ runCmd sendActions (SendToAllGenesis duration conc delay_ sendMode tpsSentFile)
                 | otherwise = (atomically $ tryReadTQueue txQueue) >>= \case
                       Just (key, txOut, neighbours) -> do
                           utxo <- getOwnUtxo $ makePubKeyAddress $ safeToPublic (fakeSigner key)
-                          tx <- createTx utxo (fakeSigner key) (one (TxOutAux txOut []))
+                          tx <- createTx utxo (fakeSigner key) (one (TxOutAux txOut [])) (makePubKeyAddress $ toPublic key)
                           case tx of
-                              Left err -> addTxFailed tpsMVar >> logError (sformat ("Error: "%stext%" while trying to send to "%shown) err neighbours)
+                              Left (TxError err) -> addTxFailed tpsMVar >>
+                                  logError (sformat ("Error: "%stext%" while trying to send to "%shown) err neighbours)
```
Martoon-00: I think it will look much nicer within `do` block

src/Pos/Client/Txp/Util.hs
```diff
@@ -174,27 +220,35 @@ makeRedemptionTx rsk txInputs = makeAbstractTx mkWit (map ((), ) txInputs)
...
+        evalStateT (pickInputs []) (totalMoneyWithFee, sortedUnspent)
     case nonEmpty futxo of
-        Nothing       -> throwError "Failed to prepare inputs!"
-        Just inputsNE -> pure (map formTxInputs inputsNE, outputs)
+        Nothing       -> throwTxError "Failed to prepare inputs!"
+        Just inputsNE -> do
+            let totalTxAmount = unsafeSumTxOuts $ map snd inputsNE
+                trInputs = map formTxInputs inputsNE
+                trOutputs = outputs
+                trRemaining = totalTxAmount `unsafeSubCoin` totalMoneyWithFee
```
Martoon-00: Should fee go to to change address?

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +260,184 @@ prepareInpsOuts utxo addrs outputs = do
...
-    runExceptT $ do
-        properOutputs <- overrideTxDistrBoot outputs
-        uncurry (makePubKeyTx ss) <$>
-            prepareInpOuts
-                utxo
-                (makePubKeyAddress $ safeToPublic ss)
-                properOutputs
+    -> AddrData m
+    -> m (Either TxError TxAux)
+createTx utxo ss outputs addrData = runExceptT $ do
```
Martoon-00: `createTx`, `createMTx` and some functions below got almost identical, it's worth extracting some logic

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +260,184 @@ prepareInpsOuts utxo addrs outputs = do
...
+                 createFakeTxFromRawTx txRaw
+        if expectedFee == txFee
+            then pure txRaw
+            else stabilizeTxFeeDo (attempt - 1) txFee
+
+txToLinearFee
+    :: MonadError TxError m
+    => TxSizeLinear -> TxAux -> m TxFee
+txToLinearFee linearPolicy =
+    either throwError pure .
```
Martoon-00: You just added that cool `eitherToThrow` :crying_cat_face:

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +260,184 @@ prepareInpsOuts utxo addrs outputs = do
...
+  where
+    invalidFee reason = TxError ("Invalid fee: " <> reason)
+
+createFakeTxFromRawTx :: TxRaw -> TxAux
+createFakeTxFromRawTx TxRaw{..} =
+    let fakeAddr = txOutAddress . toaOut . NE.head $ trOutputs
+        fakeOutMB
+            | trRemaining == mkCoin 0 = Nothing
+            | otherwise = Just $ TxOutAux (TxOut fakeAddr trRemaining) []
+        txOutsWithRem = maybe trOutputs (\remTx -> remTx :| toList trOutputs) fakeOutMB
```
Martoon-00: Also nice place for your new function

src/Pos/Client/Txp/Util.hs
Martoon-00: I think that `WithLogger` should be supposed by any ~large constraint we use, so if this is the only obstacle then why not to use this function.
On the other hand, this case seems too minor to log, it may be even better not to litter logs with this :man_shrugging:

src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -578,19 +574,3 @@ decryptAccount (hdPass, wCId) addr@(PubKeyAddress _ (Attributes (AddrPkAttrs (Ju
     guard $ length derPath == 2
     pure $ CWAddressMeta wCId (derPath !! 0) (derPath !! 1) (addressToCId addr)
 decryptAccount _ _ = Nothing
-
--- TODO [CSM-237] Move to somewhere (duplicate getWalletsAddrMetas from Methods.hs)
-getWalletAddrMetasDB
```
Martoon-00: Woow :fist_raised:

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -586,27 +548,20 @@ getMoneySourceWallet (AddressMoneySource addrId) = cwamWId addrId
...
-                calculateTxSizeLinear linearPolicy .
-                biSize @TxAux $ txAux
-  where
-    invalidFee reason = InternalError ("Invalid fee: " <> reason)
-    unknownFeePolicy =
-        InternalError . sformat ("UnknownFeePolicy, tag: "%build)
+getMoneySourceUtxo :: WalletWebMode m => MoneySource -> m Utxo
+getMoneySourceUtxo =
+    getMoneySourceAddresses >=>
+    mapM (decodeCIdOrFail . cwamWId) >=>
```
Martoon-00: You meant `cwamId`?

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -615,235 +570,47 @@ sendMoney
...
+                       _txOutputs tx
+            srcWallet = getMoneySourceWallet moneySource
+        logInfo $
+            sformat ("Successfully spent money from "%
+                     listF ", " addressF % " addresses on " %
+                     listF ", " addressF)
+            (toList srcAddrs)
+            dstAddrs
+
+        -- TODO: this should be removed in production
```
Martoon-00: Ahaha, I wonder how long these todos are going to hang there further

src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -615,235 +570,47 @@ sendMoney
...
+    srcAddrs <- forM addrMetas $ decodeCIdOrFail . cwamId
+
+    withSafeSigners passphrase sks $ \ss -> do
+        let hdwSigner = NE.zip ss srcAddrs
+        relatedAccount <- getSomeMoneySourceAccount moneySource
+        outputs <- coinDistrToOutputs dstDistr
+        TxAux {taTx = tx} <- rewrapTxError "Cannot send transaction" $
+            submitMTx enqueueMsg hdwSigner outputs (relatedAccount, passphrase)
+
+        ownUtxo <- getOwnUtxos $ toList srcAddrs
```
Martoon-00: Aren't you trying to find just spent `TxIn`s in utxo?

lwallet/Main.hs
```diff
@@ -145,20 +145,21 @@ addTxSubmit mvar = modifySharedAtomic mvar (\(TxCount submitted failed sending)
...
-    etx <-
-        withSafeSigner (skeys !! idx) (pure emptyPassphrase) $ \mss ->
-        runEitherT $ do
-            ss <- mss `whenNothing` throwError "Invalid passphrase"
-            lift $ submitTx
-                (immediateConcurrentConversations sendActions na)
-                ss
-                (map (flip TxOutAux []) outputs)
+    let skey = skeys !! idx
+        curAddr = makePubKeyAddress $ encToPublic skey
```
Martoon-00: Ah, here we use root key as address directly, so nevermind.
Unless heavy and light wallets share the same key storage - then everything is sad, and it is for a long time already

src/Pos/Client/Txp/Util.hs
flyingleafe: @Martoon-00 Fee is not going anywhere, it just gets lost (if it will go to a change address, it wouldn't be _spent_, and it wouldn't effectively be a _fee_

src/Pos/Client/Txp/Util.hs
flyingleafe: Read again: here we don't do `throwM`, but `throwError`.

src/Pos/Client/Txp/Util.hs
flyingleafe: which function? you mean `mkOutputsWithRem`?
but it's monadic, and here we have pure context. Though it may be a good idea to abstract a pure part away.

src/Pos/Wallet/Web/Server/Methods.hs
flyingleafe: Hmm, yeah, I think you're right (gosh, these fields look so identical, it's really error-prone)

src/Pos/Wallet/Web/Server/Methods.hs
flyingleafe: Here I'm trying to find `TxOut`s corresponding to spent `TxIn`s, which are required for making `THEntry`

src/Pos/Wallet/Web/Server/Methods.hs
flyingleafe: Tbh I don't even remember what exactly it's about...

lwallet/Main.hs
flyingleafe: Yeah, if we need to migrate light wallet to hd addresses then much more job should be done, I suppose.

src/Pos/Client/Txp/Util.hs
Martoon-00: Aha, I see now

src/Pos/Client/Txp/Util.hs
Martoon-00: Ah, rly

src/Pos/Wallet/Web/Server/Methods.hs
Martoon-00: Why do you do it just after you created transaction which seem to remove inputs you seek?

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +261,208 @@ prepareInpsOuts utxo addrs outputs = do
...
+-----------------------------------------------------------------------------
+
+prepareTxWithFee
+    :: TxDistrMode ctx m
+    => Utxo
+    -> TxOutputs
+    -> ExceptT TxError m TxRaw
+prepareTxWithFee utxo outputs = do
+    feePolicy <- bvdTxFeePolicy <$> lift gsAdoptedBVData
+    case feePolicy of
```
volhovM: I advice to use lambdacase here.

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +261,208 @@ prepareInpsOuts utxo addrs outputs = do
...
+
+-- Read-only function.
+computeTxFee
+    :: TxDistrMode ctx m
+    => Utxo
+    -> TxOutputs
+    -> ExceptT TxError m TxFee
+computeTxFee utxo outputs = do
+    feePolicy <- bvdTxFeePolicy <$> gsAdoptedBVData
+    case feePolicy of
```
volhovM: Lambdacase, plus seems like a copy-paste of `prepareTxWithFee`.

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +261,208 @@ prepareInpsOuts utxo addrs outputs = do
...
+    either throwError pure .
+    bimap invalidFee TxFee .
+    integerToCoin .
+    ceiling .
+    calculateTxSizeLinear linearPolicy .
+    biSize @TxAux
+  where
+    invalidFee reason = TxError ("Invalid fee: " <> reason)
+
+createFakeTxFromRawTx :: TxRaw -> TxAux
```
volhovM: Comment.

src/Pos/Client/Txp/Util.hs
```diff
@@ -174,27 +221,35 @@ makeRedemptionTx rsk txInputs = makeAbstractTx mkWit (map ((), ) txInputs)
...
+        Just inputsNE -> do
+            let totalTxAmount = unsafeSumTxOuts $ map snd inputsNE
+                trInputs = map formTxInputs inputsNE
+                trOutputs = outputs
+                trRemaining = totalTxAmount `unsafeSubCoin` totalMoneyWithFee
+            pure TxRaw {..}
   where
-    totalMoney = unsafeIntegerToCoin $ sumCoins $ map (txOutValue . toaOut) outputs
-    allUnspent = M.toList $ filterUtxoByAddrs (toList addrs) utxo
+    unsafeSumTxOuts = unsafeIntegerToCoin . sumCoins . map (txOutValue . toaOut)
```
flyingleafe: Because one cannot send _that_ much money in one transaction?

lwallet/Main.hs
```diff
@@ -145,23 +145,24 @@ addTxSubmit mvar = modifySharedAtomic mvar (\(TxCount submitted failed sending)
 addTxFailed :: Mockable SharedAtomic m => SharedAtomicT m TxCount -> m ()
 addTxFailed mvar = modifySharedAtomic mvar (\(TxCount submitted failed sending) -> return (TxCount submitted (failed + 1) sending, ()))
 
-runCmd :: forall ssc ctx m. MonadWallet ssc ctx m => SendActions m -> Command -> CmdCtx -> m ()
+runCmd :: SendActions LightWalletMode -> Command -> CmdCtx -> LightWalletMode ()
```
gromakovsky: 👍

lwallet/Main.hs
```diff
@@ -145,23 +145,24 @@ addTxSubmit mvar = modifySharedAtomic mvar (\(TxCount submitted failed sending)
...
-        withSafeSigner (skeys !! idx) (pure emptyPassphrase) $ \mss ->
-        runEitherT $ do
-            ss <- mss `whenNothing` throwError "Invalid passphrase"
-            lift $ submitTx
-                (immediateConcurrentConversations sendActions na)
-                ss
-                (map (flip TxOutAux []) outputs)
+    let skey = skeys !! idx
+        curAddr = makePubKeyAddress $ encToPublic skey
+    etx <- withSafeSigner skey (pure emptyPassphrase) $ \mss -> runEitherT $ do
```
gromakovsky: `runEitherT`, omg (I know it was here already).

src/Pos/Generator/Block/Payload.hs
```diff
@@ -196,7 +197,7 @@ genTxPayload = do
                     splitCoins outputsN (unsafeIntegerToCoin inputsSum)
                 let txOuts = NE.fromList $ zipWith TxOut outputAddrs coins
                 let txOutAuxsPre = map (\o -> TxOutAux o []) txOuts
-                either (lift . throwM . BGFailedToCreate) pure =<<
+                either (lift . throwM . BGFailedToCreate . unTxError) pure =<<
```
gromakovsky: I think it's more correct to attach `TxError` to `BGFailedToCreate`. If later we extend `TxError`, we will want this information in block generator too.

src/Pos/Client/Txp/Util.hs
```diff
@@ -126,25 +157,27 @@ overrideTxOutDistrBoot c oldDistr = do
...
+               throwTxError $
                sformat ("Can't spend "%build%" coins: amount is too small for boot "%
                         " era and can't be distributed among genStakeholders") c
           -- TODO CSL-1351 boot stakeholders' weights are not used
           pure $ genesisSplitBoot (HM.fromList $ map (,1) genStakeholders) c
 
 -- | Same as 'overrideTxOutDistrBoot' but changes 'TxOutputs' all at once
-overrideTxDistrBoot ::
-       (TxCreateMode ctx m) => TxOutputs -> ExceptT Text m TxOutputs
+overrideTxDistrBoot
```
gromakovsky: I think ideally `overrideTxDistrBoot` functions shouldn't be exported, because for me the look like an internal detail of other tx creation functions. All tx creation functions should use it. And it's quite useless separately.

src/Pos/Client/Txp/Util.hs
gromakovsky: It's still not clear to me, because I don't know this part well. What if someone calls `createRedemptionTx` and passes a list of outputs with too large sum of coins? It's wrong to do it, but if it happens, it makes more sense to throw `TxError` than to use `error` (which is unpredictable and dangerous).

pva701:
Padajji moe review plz

src/Pos/Client/Txp/Addresses.hs
```diff
@@ -0,0 +1,17 @@
...
+module Pos.Client.Txp.Addresses
+       ( MonadAddresses (..)
+       ) where
+
+import           Universum
+
+import           Pos.Types (Address)
+
+-- | A class which have the method to generate a new address
+class Monad m => MonadAddresses m where
```
pva701: :+1:

pva701: I dreamt about such type class

src/Pos/Client/Txp/Util.hs
```diff
@@ -174,27 +221,35 @@ makeRedemptionTx rsk txInputs = makeAbstractTx mkWit (map ((), ) txInputs)
...
+                trOutputs = outputs
+                trRemaining = totalTxAmount `unsafeSubCoin` totalMoneyWithFee
+            pure TxRaw {..}
   where
-    totalMoney = unsafeIntegerToCoin $ sumCoins $ map (txOutValue . toaOut) outputs
-    allUnspent = M.toList $ filterUtxoByAddrs (toList addrs) utxo
+    unsafeSumTxOuts = unsafeIntegerToCoin . sumCoins . map (txOutValue . toaOut)
+    totalMoney = unsafeSumTxOuts outputs
+    totalMoneyWithFee = totalMoney `unsafeAddCoin` fee
+    allUnspent = M.toList utxo
```
pva701: I can't drop my comment on `pickInputs` line, so it's here.

It a little bit differs from `selectSrcAddress` (in the previous code): there was a case when `moneyLeft` is less than current input and we tried to find an input with the same amount to set `remain = 0` (and don't use an extra input). you can see it in the last if-branch of `selectSrcAddressesDo`

I would prefer to have this logic here. what do you think?

pva701: and it's subjective, but IMHO a function w/ several named parameters looks better than tuple-state and access to them via `_1`, `_2`... 
because non-familar reader is forced to see what this tuple means every time when he sees `_1`, `_2`, `_3`

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +261,215 @@ prepareInpsOuts utxo addrs outputs = do
...
+    -> Utxo
+    -> TxOutputs
+    -> m TxRaw
+stabilizeTxFee linearPolicy utxo outputs =
+    stabilizeTxFeeDo 5 (TxFee $ mkCoin 0)
+  where
+    stabilizeTxFeeDo :: Int -> TxFee -> m TxRaw
+    stabilizeTxFeeDo 0 _ = throwTxError "Couldn't stabilize tx fee after 5 attempts"
+    stabilizeTxFeeDo attempt expectedFee = do
+        txRaw <- prepareTxRaw utxo outputs expectedFee
```
pva701: `stabilizeTxFeeDo` uses `prepareTxRaw`, but `prepareTxRaw` isn't in `Fee logic` section.
it's hard to review and hard to follow the flow of logic of code when you read and modify this code.

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +261,215 @@ prepareInpsOuts utxo addrs outputs = do
...
+            | otherwise = Just $ TxOutAux (TxOut fakeAddr trRemaining) []
+        txOutsWithRem = maybe trOutputs (\remTx -> remTx :| toList trOutputs) fakeOutMB
+
+        -- We create fake signers instead of safe signers,
+        -- because safe signer requires passphrase
+        -- but we don't want to reveal our passphrase to compute fee.
+        -- Fee depends on size of tx in bytes, sign of a tx has the fixed size
+        -- so we can use arbitrary signer.
+        srcAddrs = NE.map (txOutAddress . fst) trInputs
+        (_, fakeSK) = deterministicKeyGen "patakbardaqskovoroda228pva1488kk"
```
pva701: it's so cute you perpetuated creator of this code in this string <3

src/Pos/Client/Txp/Util.hs
```diff
@@ -206,60 +261,215 @@ prepareInpsOuts utxo addrs outputs = do
...
+    => Utxo
+    -> TxOutputs
+    -> ExceptT TxError m TxRaw
+prepareTxWithFee utxo outputs = withLinearFeePolicy $ \linearPolicy ->
+    stabilizeTxFee linearPolicy utxo outputs
+
+-- | Compute, how much fees we should pay to send money to given
+-- outputs
+computeTxFee
+    :: TxDistrMode ctx m
```
pva701: it's too general constraint for this fucntion. AFAIU it would be enough `MonadError`

pva701: ah ok, I forgot we have to get current fee policy from `GState`

src/Pos/Client/Txp/Util.hs
flyingleafe: @pva701 but `prepareTxRaw` is _not_ fee logic, it's basic logic of choosing the inputs of transaction, and it's used even when no fees are supposed to be added (like in redemption txs). I doubt that adding it to `Fee logic` section makes sense.

src/Pos/Client/Txp/Util.hs
flyingleafe: @pva701 regarding logic with trying to find the _exact_ same output: is it really worth it? I doubt that cases when we have some output with _exactly_ the same amount of money as what is remained to send are common - I bet they are quite rare. We'll only win one transaction output in such rare cases, which is not much in terms of bytes.

If we _really_ want to optimize it properly, then we need to implement something smarter (like reduction to maxflow), because taking biggest output first is not optimal indeed.

pva701: ok, I realized if we even add this logic it won't prevent all cases taking of extra input.
so in general case, we take a minimum number of inputs or one more input, it is not worth it.
lets zab'em

flyingleafe:
It works; tested manually and via Daedalus acceptance tests.
(Daedalus acceptance tests don't pass though, but only because they consider outgoing transaction amount invalid, if it's displayed with added transaction fees. Issue is reported to frontend team)

PR #1274, opened 2017-08-04
[CSM-376] Extract `cardano-sl-wallet`

sectore:
to have most of wallet sources in `cardano-sl-wallet` project

How to test?
1. `scripts/build/cardano-sl.sh wallet` to build wallet only
2. `scripts/build/cardano-sl.sh` to build everything


cardano-sl.cabal
```diff
@@ -43,10 +31,12 @@ library
                         Pos.Block.BListener
                         Pos.Block.Core
                         Pos.Block.Logic
+                        Pos.Block.Logic.Util
                         Pos.Block.Network
                         Pos.Block.Pure
                         Pos.Block.RetrievalQueue
                         Pos.Block.Slog
+                        Pos.Block.Slog.Types
```
gromakovsky: To be honest, I wouldn't expose `Pos.Block.Logic.Util` and `Pos.Block.Slog.Types`, they are available from `Pos.Block.Logic` and `Pos.Block.Slog` respectively.

cardano-sl.cabal
```diff
@@ -115,8 +113,26 @@ library
...
+                        Pos.Wallet.Web.ClientTypes
+                        Pos.Wallet.Web.Error
+                        Pos.Wallet.Web.Error.Types
+                        Pos.Wallet.Web.Error.Util
                         Pos.Wallet.Web.Secret
+                        Pos.Wallet.Web.State
+                        Pos.Wallet.Web.State.Acidic
+                        Pos.Wallet.Web.State.State
+                        Pos.Wallet.Web.State.Storage
+                        Pos.Wallet.Web.Util
```
gromakovsky: Why are these `Pos.Wallet.*` modules still in `cardano-sl`? Does something else (not wallet) depend on it?

tools/cardano-sl-tools.cabal
```diff
@@ -140,13 +128,7 @@ executable cardano-wallet-hs2purs
...
   ghc-options: -pgmP cpphs -optP --cpp
 
-  if flag(with-wallet) && flag(with-web)
-    buildable:         True
-  else
-    buildable:         False
-
-  if flag(with-web)
-    cpp-options: -DWITH_WEB
+  buildable:         False
```
gromakovsky: This is wrong, we need to build `cardano-wallet-hs2purs`. I propose to move this executable to `cardano-sl-wallet`.

tools/cardano-sl-tools.cabal
```diff
@@ -363,8 +344,7 @@ executable cardano-keygen
 
   if flag(for-installer)
     buildable: False
-
-  if !flag(with-wallet)
+  else
     buildable: False
```
gromakovsky: This is wrong. I don't understand why there was `with-wallet` check. `buildable: False` should be set only if `for-installer` is enabled.

tools/cardano-sl-tools.cabal
```diff
@@ -261,8 +243,7 @@ executable cardano-swagger
 
   if flag(for-installer)
     buildable: False
-
-  if !flag(with-wallet) || !flag(with-web)
+  else
     buildable: False
```
gromakovsky: This is also wrong, we need to be able to build keygen. You can either move it to `cardano-sl-wallet` or make `cardano-sl-tools` depend on `cardano-sl-wallet` (which is fine).

cardano-sl.cabal
sectore: @gromakovsky `Pos.Client.Txp.Balances` depends on `Pos.Wallet.Web.State` (see https://github.com/input-output-hk/cardano-sl/blob/5b7acc1d68ad72b5f31acb6e5709eb4c070af20e/src/Pos/Client/Txp/Balances.hs#L27-L28). And with this all others of `Pos.Wallet.Web.State.*` are imported by `Pos.Wallet.Web.State`

tools/cardano-sl-tools.cabal
sectore: @gromakovsky Why keygen? This changes are related to `cardano-swagger` executable and not to `cardano-dht-keygen`.

tools/cardano-sl-tools.cabal
sectore: @gromakovsky I will set `buildable: True`. Let's move `cardano-wallet-hs2purs` with another `PR`, because it is hard to have this PR here in common with `master` (which changes frequently).

cardano-sl.cabal
gromakovsky: It's fine then, we will sort it out after this PR.

tools/cardano-sl-tools.cabal
gromakovsky: Ah, I misread it. Then `cardano-swagger` should be moved to `cardano-sl-wallet`. Maybe in another PR. But the point that at least one CI should check that `cardano-swagger` compiles.

cardano-sl.cabal
sectore: Ok, I will do it.

tools/cardano-sl-tools.cabal
sectore: @gromakovsky Finally I have moved `cardano-wallet-hs2purs` into `cardano-sl-wallet` project. Thx again for pointing it out!

tools/cardano-sl-tools.cabal
sectore: Done.

cardano-sl.cabal
```diff
@@ -12,17 +12,11 @@ build-type:          Simple
 extra-source-files:  README.md
 cabal-version:       >=1.10
 
-Flag with-web
-  default:     True
-  manual:      True
-
-  description: Build with web part.
-
 Flag with-wallet
```
gromakovsky: Why does this flag still exist? Is it used only to determine linker or for something else?

src/Pos/Client/CLI/Secrets.hs
```diff
@@ -14,11 +14,13 @@ import           Pos.Genesis         (genesisDevSecretKeys)
 import           Pos.Ssc.GodTossing  (genesisDevVssKeyPairs)
 import           Pos.Util.UserSecret (UserSecret, usPrimKey, usVss, writeUserSecret)
 
-import           NodeOptions         (Args (..))
+import           Pos.Client.CLI.NodeOptions         (SimpleNodeArgs (..))
 
+-- TODO (jk): Don't use Args, we do need just `devSpendingGenesisI`
+-- So we can use `userSecretWithGenesisKey` for simple + wallet node
```
flyingleafe: It seems like you already came up with a solution which doesn't require this, so this TODO is obsolete.

cardano-sl.cabal
sectore: @gromakovsky It is needed here https://github.com/input-output-hk/cardano-sl/blob/cb7d38f37c211592f0a38cb1c93096df2ba10108/cardano-sl.cabal#L429 to avoid gold linker issue as described here https://ghc.haskell.org/trac/ghc/ticket/13810

src/Pos/Client/CLI/Secrets.hs
sectore: That's right, thx!

tools/cardano-sl-tools.cabal
sectore: @gromakovsky Done. `cardano-swagger` has been moved into `cardano-sl-wallet`.

cardano-sl.cabal
Martoon-00: It seems to me that `Pos.Client.Txp.*` should also go to wallet project (eventually)

gromakovsky: No, because it can also be used (and it is used) by e. g. `cardano-sl-lwallet`.

wallet/node/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+    putText $ "Running using " <> show (CLI.sscAlgo commonArgs)
+    putText $ "Wallet is enabled!"
+
+    let vssSK = fromJust $ npUserSecret currentParams ^. usVss
+    let gtParams = gtSscParams snArgs vssSK
+
+    let sscParams :: Either (SscParams SscNistBeacon) (SscParams SscGodTossing)
+        sscParams = bool (Left ()) (Right gtParams) (CLI.sscAlgo commonArgs == GodTossingAlgo)
+    case sscParams of
+        (Left _)     -> logError "Wallet does not support NIST beacon!"
```
gromakovsky: It would be good to remove ability to run wallet with nist beacon completely (remove `sscAlgo` option from wallet). It can be done after this PR is merged though.

src/Pos/Client/CLI/Util.hs
```diff
@@ -0,0 +1,14 @@
+module Pos.Client.CLI.Util
+       ( printFlags
+       ) where
+
+import           Universum
+
+import           Pos.Constants       (isDevelopment)
+import           Pos.Util            (inAssertMode)
+printFlags :: IO ()
```
gromakovsky: Absence of empty line after imports section looks quite unpleasant.

wallet/node/Main.hs
gromakovsky: Also there are some other options which make sense for simple node, but don't make much sense for wallet, e. g. malicious activity emulation options. But again, I don't mind if this is refactored later.

sectore: @gromakovsky Could you or anybody else create an YT issue for this ^, please? I'm might be not the right person for writing this issue due the lack of knowledge... Thx!

src/Pos/Client/CLI/Util.hs
sectore: @gromakovsky Agree! And thanks for fixing typo with https://github.com/input-output-hk/cardano-sl/commit/9449b72caab8d2006baa62c4edb0cd9a381bf4a4

wallet/node/Main.hs
gromakovsky: https://issues.serokell.io/issue/CSM-409

PR #1277, opened 2017-08-05
Use containers 0.5.7.1

arybczak:


neongreen:
@arybczak could you elaborate a bit?

arybczak:
@neongreen See https://issues.serokell.io/issue/CSL-1468

cleverca22:
stack2nix is unable to change the version of the containers package, so all builds done by nix must be compatible with version 0.5.7, and 0.5.8 isn't available

PR #1278, opened 2017-08-05
[CSM-401] Fix wallet password deserialization

Pastafarianist:
The initial idea was to get rid of `instance Bi PassPhrase` altogether and move its code into `FromCType` / `ToCType`. However, this instance is used in several other places. In all of these cases the only reason for why `Bi PassPhrase` is needed is that `hash` is called on passphrases, and internally `hash` calls `serialize'`. So, in fact, only `encode` is ever used. However, I don't know how to get rid of `instance Bi PassPhrase` without splitting `Bi` into `encode` and `decode` and changing the constraint on the current `hash` function to only require `encode` to be implemented.

There is some (albeit small) code duplication. I was thinking whether it could be extracted and put into separate functions, but I'm not sure where I could put it. An obvious candidate is `Pos.Binary.Crypto`, however currently it doesn't export any identifiers, and I'm not sure whether it is a good idea to start doing that.

An alternative solution would be to fix `daedalus-bridge` to send the following to the backend:

    passphrase & hash & cbor.serialize & base16

(currently it skips the `cbor.serialize` step and dumps hash as raw bytes, hence the original error).

### DIFF
--- a/src/Pos/Wallet/Web/ClientTypes.hs
+++ b/src/Pos/Wallet/Web/ClientTypes.hs
@@ -56,6 +56,8 @@ import           Universum
 
 import           Control.Arrow             ((&&&))
 import           Control.Monad.Error.Class (throwError)
+import qualified Data.ByteArray            as ByteArray
+import qualified Data.ByteString           as BS
 import           Data.Default              (Default, def)
 import           Data.Hashable             (Hashable (..))
 import qualified Data.Set                  as S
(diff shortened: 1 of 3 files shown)


src/Pos/Wallet/Web/ClientTypes.hs
```diff
@@ -248,12 +249,22 @@ instance Show CPassPhrase where
...
+    decodeCType (CPassPhrase text) = do
+        bs <- Base16.decode text
+        let bl = BS.length bs
+        -- Currently passphrase may be either 32-byte long or empty (for
+        -- unencrypted keys).
+        if bl == 0 || bl == passphraseLength
+            then pure $ ByteArray.convert bs
+            else fail . toString $ sformat
+                 ("Expected password length 0 or "%int%", not "%int)
+                 passphraseLength bl
```
flyingleafe: I really don't like such copypaste, but since we cannot simply re-use `Binary` instance here, I think I can live with it (until next serialization refactoring though =\)

flyingleafe:
Tested, it works

PR #1279, opened 2017-08-06
[CSL-1451] Check bounds in 'mkCoin'

pva701:


core/Pos/Core/Types.hs
```diff
@@ -328,7 +328,9 @@ maxCoinVal = 45000000000000000
 -- FIXME: This operation is unsafe because it doesn't check 'maxCoinVal'.
```
gromakovsky: You forgot to delete this FIXME.

core/Pos/Core/Types.hs
```diff
@@ -328,7 +328,9 @@ maxCoinVal = 45000000000000000
...
 -- | Make Coin from Word64.
 mkCoin :: Word64 -> Coin
-mkCoin = Coin
+mkCoin c
+    | c <= maxCoinVal = Coin c
+    | otherwise       = error "mkCoin: coin is too large"
```
gromakovsky: It would be good to also attach the value of `c` to the message.

pva701: @gromakovsky done

test/Test/Pos/Types/CoinSpec.hs
```diff
@@ -135,9 +132,6 @@ overflowIntegerCausesError =
 coinToIntegralToCoin :: C.Coin -> Property
 coinToIntegralToCoin = C.mkCoin . C.unsafeGetCoin .=. identity
 
-wordToCoinToWord :: Word64 -> Property
```
gromakovsky: I think @rockbmb will be unhappy with this change. You could rewrite this property using `==>`. Also it would be good to add a test that `mkCoin tooBigNumber` fails.

rockbmb: This test isn't all too important but it's useful to know turning a `Word64` into a coin and then using `unsafeGetCoin` doesn't change anything.

pva701:
@gromakovsky fixed

PR #1280, opened 2017-08-07
Csl1453 unused deps string qq

rockbmb:


neongreen:
`neat-interpolation` has actual interpolation (which we use e.g. in `Pos.Script.Examples`). `string-qq` doesn't do interpolation, it just gives you a bit less verbose multiline strings.

adinapoli-iohk:
@rockbmb Alas we got some conflicts -- mind tackling them so we can merge? 😉

PR #1282, opened 2017-08-07
[CSL-1417] Modify block event type

int-index:
This PR introduces the tree-based block event generator and a DSL for using it, making it possible to model forks.

Features include:

* Snapshot management (manual and semi-automatic)
* Building the blockchain tree from paths
* Generator of diverse apply/rollback sequences with forks
* Pretty-printing block event sequences with paths or block hashes
* Reasonably efficient implementation that takes advantage of laziness
* <s>Superb path naming scheme: `kek/rekt/mems`.</s>

### DIFF
--- a/src/Pos/Generator/BlockEvent.hs
+++ b/src/Pos/Generator/BlockEvent.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DeriveFoldable      #-}
 {-# LANGUAGE DeriveFunctor       #-}
+{-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Pos.Generator.BlockEvent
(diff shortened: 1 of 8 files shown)


src/Pos/Generator/BlockEvent.hs
```diff
@@ -117,31 +272,60 @@ instance IsBlockEventFailure (BlockEventRollback' blund) where
...
+    deriving (Show)
+
+instance Buildable SnapshotOperation where
+      build = bprint shown
+
 data BlockEvent' blund
-    = BlkEvApply (BlockEventApply' blund)
-    | BlkEvRollback (BlockEventRollback' blund)
-    deriving (Functor, Foldable)
+      = BlkEvApply (BlockEventApply' blund)
```
int-index: Emacs has its own opinion on indentation :(

src/Pos/Generator/BlockEvent.hs
```diff
@@ -18,49 +19,201 @@ module Pos.Generator.BlockEvent
...
+            , _bgpInplaceDB   = True
+            }
+    -- Partial pattern-matching is safe because we specify
+    -- blockCount = 1 in the generation parameters.
+    OldestFirst [block] <- genBlocks blockGenParams
+    forestBlocks <- genBlocksInForest secrets blockchainForest
+    return $ BlockchainTree block forestBlocks
+
+-- Precondition: paths in the structure are non-empty.
+genBlocksInStructure ::
```
int-index: It takes a structure with paths, generates a suitable tree and returns the same structure but with blunds. It's obvious from the types, why even read the impl?

neongreen: to check that there is nothing wrong with the implementation

int-index: Alright, let's look at it step-by-step.

1. The lookup function for annotations:
    ```
    getAnnotation path =
                Map.findWithDefault BlockDescDefault path annotations
    ```
    If the user had some preferences for how the block at `path` should be generated (there's an annotation), we'll return it. Otherwise, we return `BlockDescDefault` (the user doesn't care how the particular block is generated).

2. Collect all paths with annotations.
   ```        paths = foldMapOf folded
                  (\path -> Map.singleton path (getAnnotation path))
                  s
   ```
   Here we traverse the structure the first time, finding all paths and desired generation strategies (`BlockDesc`) for these paths. The result is `Map Path BlockDesc` — a set of paths, and a block description for each.

3. Assemble the paths into a tree.
    ```
    descForest = buildBlockchainForest BlockDescDefault paths
    ```
    The result is a tree where we have a node for each path, and the nodes contain `BlockDesc` associated with these paths. One case to be handled is when we have, say, a path like `kek/mems`, but not just `kek`. In this case, what do we put into the node at path `kek`? This is why we have defaulting again, passing `BlockDescDefault` to fill these holes. By construction, if we later use one of the paths in `paths` to find a node in `descForest`, the lookup will succeed.

5. Generate blocks in the tree.
    ```
    blockForest <- genBlocksInForest secrets descForest
    ```
    The `genBlocksInForest` function traverses the tree and generates a block for each block description. 

6. A function to get the block in the tree. Now that we have a tree with the actual blocks, we flatten it back to get the map from paths to blocks. (Before this, we had a map from paths to block descriptions). By construction, this map contains all paths we had before, so it's just a lookup.
    ```
            getBlock path = Map.findWithDefault
                (error "genBlocksInStructure: impossible happened")
                path
                (flattenBlockchainForest' blockForest)
    ```
7. Replace the paths with actual blocks in the original structure:
    ```
    fmap getBlock s
    ```

Hope this helps.

int-index: To be honest, I think the code is rather straightforward if you know what it's supposed to do, but I couldn't figure out how to describe the general idea without pictures, so the end result is that there are no comments at all.

gromakovsky: I think it would be helpful to write some types explicitly. For instance, the type of `paths`.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +339,49 @@ newtype Chance = Chance {getChance :: Rational}
...
-            return (Nothing, gbes)
-        GbesStartingState ->
-            genBlockStartingState failure rollback
-        GbesBlockchain blockchain ->
-            genBlockInBlockchain blockchain failure rollback
-    put gbes' $> mBlockEvent
+newtype CheckCount = CheckCount Word
+    deriving (Eq, Ord, Show, Num)
+
+blkEvTip :: BlockEvent -> Maybe HeaderHash
```
gromakovsky: Add a comment that this function returns the tip which is expected after the event is applied. Because it can also return the tip which is expected before the event.

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +339,49 @@ newtype Chance = Chance {getChance :: Rational}
...
+        (ev:evs) -> case blkEvTip ev of
+            Nothing -> go hhStatusSoFar (ev:revNewBs) evs
+            Just tipHh -> case Map.lookup tipHh hhStatusSoFar of
+                Nothing ->
+                    let
+                        snapSave = BlkEvSnap (SnapshotSave (hhSnapshotId tipHh))
+                        needSnapSave =
+                            -- We tie the know here to determine whether to actually emit
+                            -- the command to save the snapshot.
+                            Map.findWithDefault 0 tipHh hhStatusEnd > 0
```
gromakovsky: is it intentional that `hhStatusEnd` is used here? Shouldn't it be `hhStatusSoFar`?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +339,49 @@ newtype Chance = Chance {getChance :: Rational}
...
+                        appendSnapSave = if needSnapSave then (snapSave:) else identity
+                    in
+                        go (Map.insert tipHh 0 hhStatusSoFar) (appendSnapSave $ ev:revNewBs) evs
```
gromakovsky: Should we insert `0` even if `needSnapSave` is `False`? Is there some magic related to the usage of `hhStatusEnd` above?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +339,49 @@ newtype Chance = Chance {getChance :: Rational}
...
+    go hhStatusSoFar revNewBs = \case
+        [] -> (hhStatusSoFar, revNewBs)
+        (ev:evs) -> case blkEvTip ev of
+            Nothing -> go hhStatusSoFar (ev:revNewBs) evs
+            Just tipHh -> case Map.lookup tipHh hhStatusSoFar of
+                Nothing ->
+                    let
+                        snapSave = BlkEvSnap (SnapshotSave (hhSnapshotId tipHh))
+                        needSnapSave =
+                            -- We tie the know here to determine whether to actually emit
```
gromakovsky: What is _the know_? It's a verb, isn't it? Is it a typo?

src/Pos/Generator/BlockEvent.hs
gromakovsky: I see there is `Map.lookup tipHh hhStatusSoFar` above, so `hhStatusSoFar` doesn't make sense here.

Can you explain how it works then?

src/Pos/Generator/BlockEvent.hs
int-index: Right, it's "the knot".

src/Pos/Generator/BlockEvent.hs
int-index: Yes, we tie the knot, it's 100% intentional and tested. The idea is that we need to save the snapshot only if we later use it, and we know whether we use it only ofter traversing the entire structure. So we use this result "from the future" via laziness, and it doesn't go into a loop because there's no data dependency.

src/Pos/Generator/BlockEvent.hs
int-index: Done, added type signatures for all intermediate results.

pva701: @int-index anyway, can you write some clarified comment? what annotation is?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +344,50 @@ newtype Chance = Chance {getChance :: Rational}
...
+-- The tip after the block event. 'Nothing' when the event doesn't affect the tip.
+blkEvTip :: BlockEvent -> Maybe HeaderHash
+blkEvTip = \case
+    BlkEvApply bea -> Just $
+        (headerHash . NE.head . getNewestFirst . toNewestFirst . view beaInput) bea
+    BlkEvRollback ber -> Just $
+        (headerHash . view prevBlockL . NE.head . getOldestFirst . toOldestFirst . view berInput) ber
+    _ -> Nothing
+
+-- | Empty: hash snapshot unavailable
```
pva701: do I understand correctly it's about `CheckCount` value? can you clarify?

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +344,50 @@ newtype Chance = Chance {getChance :: Rational}
...
-
-To avoid dealing with this, we append one more block. The downside is that
-appending this block may cause the sequence to be longer than 'blockIndexMax'.
-
--}
+    checkCount = sum (hhStatusEnd :: HhStatusMap)
+    (hhStatusEnd, revBs') = go Map.empty [] bs
+    bs' = reverse revBs'
+    -- NB. 'go' is tail-recursive.
+    go hhStatusSoFar revNewBs = \case
```
pva701: it would be nice if you write signature of this function

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +344,50 @@ newtype Chance = Chance {getChance :: Rational}
...
+        [] -> (hhStatusSoFar, revNewBs)
+        (ev:evs) -> case blkEvTip ev of
```
pva701: it's minor, but you can write:
```
go hhStatusSoFar revNewBs bs
    | [] <- bs = ...
    | (ev: evs) <- bs, 
      Nothing <- blkvEvTip ev = ...
    | (ev: evs) <- bs, 
      Just tipHh <- blkvEvTip ev, 
      Just k <- Map.lookup tipHh hhStatusSoFar = ...
    | (ev: evs) <- bs, 
      Just tipHh <- blkvEvTip ev, 
      Nothing <- Map.lookup tipHh hhStatusSoFar =
```
it would decrease indentation :triangular_ruler:

src/Pos/Generator/BlockEvent.hs
```diff
@@ -155,360 +344,50 @@ newtype Chance = Chance {getChance :: Rational}
...
+            Nothing -> go hhStatusSoFar (ev:revNewBs) evs
+            Just tipHh -> case Map.lookup tipHh hhStatusSoFar of
+                Nothing ->
+                    let
+                        snapSave = BlkEvSnap (SnapshotSave (hhSnapshotId tipHh))
+                        needSnapSave =
+                            -- We tie the knot here to determine whether to actually emit
+                            -- the command to save the snapshot.
+                            Map.findWithDefault 0 tipHh hhStatusEnd > 0
```
pva701: @int-index I don't get why you won't get a cyclic definition here....
you call `go` to init `hhStatusEnd` but you also try to use `hhStatusEnd` during the initialization of `hhStatusEnd`

pva701: AFAIU this code, it's not a problem to save a snapshot regardless number of usages in the end of execution.

src/Pos/Generator/BlockEvent/DSL.hs
```diff
@@ -0,0 +1,139 @@
...
+    BlockEventGenT g m () ->
+    RandT g m (Map Path BlockDesc, BlockScenario' Path)
+runBlockEventGenT' m = do
+    let
+        initialBlockEventGenState = BlockEventGenState
+            { _begsEvents = NewestFirst []
+            , _begsAnnotations = Map.empty
+            }
+    begs <-
+        let reorder (((), g), begs) = (begs, g)
```
pva701: nice construction

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -168,29 +184,102 @@ blockEventSuccessSpec = do
...
+-}
+
+genSuccessWithForks :: forall g m. (RandomGen g, Monad m) => BlockEventGenT g m ()
+genSuccessWithForks = do
+      emitBlockApply BlockApplySuccess $ pathSequence mempty ["0"]
+      generateFork "0" []
+      emitBlockApply BlockApplySuccess $ pathSequence "0" ["1", "2"]
+      generateFork ("0" <> "1" <> "2") []
+  where
+    generateFork ::
```
pva701: I hope this logic is correct. it's so gribnaya that I can't force myself to delve into it.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -168,29 +184,102 @@ blockEventSuccessSpec = do
...
+                        relPaths <- OldestFirst <$> replicateM advance generateRelativePath1
+                        whenJust (nonEmptyOldestFirst relPaths) $ \relPaths' -> do
+                            let
+                                curPath = maybe basePath NE.head $ nonEmpty (getNewestFirst rollbackFork)
+                                paths = pathSequence curPath relPaths'
+                            emitBlockApply BlockApplySuccess paths
+                            generateFork basePath (over _NewestFirst toList (toNewestFirst paths) <> rollbackFork)
+    generateRelativePath1 :: BlockEventGenT g m Path
+    generateRelativePath1 =
+        uniform (["rekt", "kek", "mems", "peka"] :: NE Path)
```
pva701: what does `reckt` mean?

int-index: It's a phonetic spelling of "wrecked". Get rekt.

src/Pos/Generator/BlockEvent.hs
int-index: Yes, it's about the result of `lookup`. `Nothing` means that a snapshot unavailable. "hash snapshot" just means that it's a db snapshot indexed by the hash of the tip, I should find a better name for it.

src/Pos/Generator/BlockEvent.hs
int-index: I don't like that you repeat code, i.e. you write `Map.lookup tipHh hhStatusSoFar`.

Indentation is ok, it shows structure. The reason you don't like it is that 4 spaces is too much, with 2 spaces it would look just fine, but codestyle is codestyle.

src/Pos/Generator/BlockEvent.hs
int-index: >  I don't get why you wouldn't get a cyclic definition here....

Because the resulting `hhStatusEnd` is not used to act on `hhStatusSoFar`, it only affects `revBs'`.

int-index: > it's not a problem to save a snapshot regardless number of usages in the end of execution

I think if I save too much snapshots, the memory consumption during tests will be more than necessary. It's not good to retain the entire DB in memory if it won't be used later.

src/Pos/Generator/BlockEvent.hs
pva701: pls, write that "hash snapshot" is "tip" or something like that b/c when I read the first time I thought it's hash of db state (yep, I am idiot...) or smth like that...

src/Pos/Generator/BlockEvent.hs
pva701: >Because the resulting hhStatusEnd is not used to act on hhStatusSoFar, it only affects revBs'.
oh, vot eto conechno lol kek cheburek. I hope nobody will shoot in the foot in this code.

test/Test/Pos/Block/Logic/VarSpec.hs
gromakovsky: Brutal. Savage. Rekt.

PR #1283, opened 2017-08-07
[CSL-1447] Restore CRC for Address serialisation

adinapoli-iohk:
This PR fixes [CSL-1447](https://issues.serokell.io/issue/CSL-1447) by restoring the CRC check inside the `decode` function in the `Bi Address` instance.

The only thing I had to watch out was to avoid the infinite loop which I've explained in the comment. I learnt the hard way why the pilgrim fathers of Cardano originally separated the `Address` encoding in the [previous serialisation revision](https://github.com/input-output-hk/cardano-sl/blob/1a345036d95647aa7ea00d182013ee822995e577/core/Pos/Binary/Core/Address.hs#L48) 😂  😂

### DIFF
--- a/core/Pos/Binary/Core/Address.hs
+++ b/core/Pos/Binary/Core/Address.hs
@@ -3,6 +3,8 @@ module Pos.Binary.Core.Address () where
 
 import           Universum
 
+import           Codec.CBOR.Encoding (Encoding)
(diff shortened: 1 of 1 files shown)


neongreen:
The optional Appveyor build timed out, merging anyway.

PR #1275, opened 2017-08-04
[CSL-1442/CSL-1443] Improve `computeSharesDistr` and tests. Improve `Arbitrary ValidRichmenStakes`. 

pva701:
* Arbitrary instance for `ValidRichmenStake` is rewritten.
* `isFairDistr` uses knapsack algorithm instead of brute force to provide opportunity to run big tests (n > 20)
* `computeSharesDistr` uses knapsack algorithm to provide strict guarantees on a difference between real and generated distribution.

I would like to implement in this issue (or maybe in next issues):
* add optimization in knapsack part: if we achieve inaccuracy `sharesDistrInaccuracy` on some step of algorithm then break and return it
* I thought of the worst case of the test for `computeSharesDistr` and got some results, I would like to add some "clever" tests about it.

PR #1288, opened 2017-08-07
[CSL-849] remove MessageName, MessagePart

avieth:
Bumped the time-warp-nt revision to one which doesn't have `MessageName`. It's replaced with `MessageCode ~ Word16`. This means the `MessagePart` motif that was used to get a whack of `Message` instances for free no longer applies, so that was removed. I could have updated it to use some similar scheme, like using the top 8 bits for the relay message constructor and the bottom eight for the particular content type. But I think it's better to have the message codes stated explicitly. I don't believe all of the codes listed in `src/Pos/Communication/Message.hs` are actually necessary. They were needed in order to compile, but surely some of the relay message instances don't need to be there.

So the length-prefixed `MessageName` is no longer used, which is the point of the ticket CSL-849, but `VerInfo` still contains arbitrarily-long `HashMap`s so the problem isn't yet solved.


neongreen:
I think that the [`HandlerSpec` encoding table](https://github.com/input-output-hk/cardano-sl/pull/1288/files#diff-5a22153eaebb1927ba5f9b5b1df6104fR63) should be removed completely, as it's a) outdated and b) not necessary since the current encoding is much simpler than what we had before.

neongreen:
(Otherwise, LGTM.)

neongreen:
```
[ 70 of 205] Compiling Pos.Communication.Message ( src\Pos\Communication\Message.hs, .w\dist\ca59d0ab\build\Pos\Communication\Message.o )
C:\w\src\Pos\Communication\Message.hs:87:10: error:
    * No instance for (Buildable (InvOrData key TxMsgContents))
        arising from a use of `Node.Message.Class.$dmformatMessage'
    * In the expression:
        Node.Message.Class.$dmformatMessage @InvOrData key TxMsgContents
      In an equation for `formatMessage':
          formatMessage
            = Node.Message.Class.$dmformatMessage @InvOrData key TxMsgContents
      In the instance declaration for
        `Message (InvOrData key TxMsgContents)'
--  While building package cardano-sl-0.5.1
```

avieth:
https://github.com/input-output-hk/cardano-sl/pull/1288/commits/736b4318fb2f598a886c4468f96e48f37d047b40

Includes a `formatMessage` definition for `TxMsgContents` that I had forgotten. Strange that my local build, and one of the travis builds, passed.

avieth:
CI has failed because of this test: `Richmen with stake distribution [0.24, 0.25, 0.26, 0.29] to test validity of @computeSharesDistr`. I've seen this before, but it's nothing to do with these changes.

gromakovsky:
@avieth I've restarted the build. This failure in tests should fully disappear after #1275 is polished and merged.

gromakovsky:
There is a conflict now.

adinapoli-iohk:
I have no idea what happened here -- if I click on the `Details` link Travis shows everything as green, but yet we get this (stale?) yellow dot here 🙉

gromakovsky:
Strange, I restarted one build, maybe it will help.

avieth:
No clue why one of the travis builds failed. Can't see an error in the logs.

adinapoli-iohk:
If I've untangled everything corretly I swear I have seen a timeout lurking at some point in the logs 🙊

neongreen:
@avieth there's a conflict already :(

avieth:
> there's a conflict already :(

I did the stack2nix thing but while it was going there must have been a commit of `pkgs/default.nix` to master. Rebased and ran it again and now CI is going.

PR #1290, opened 2017-08-07
[CSL-1475] Fix balances update in explorer

gromakovsky:
I've fixed one bug, but it's still WIP because tests still fail.
I don't understand why they fail, will continue tomorrow. For more details see the issue description (https://issues.serokell.io/issue/CSL-1475).


gromakovsky:
This fix was wrong and doesn't fix anything. Tests will be fixed in #1309.

PR #1291, opened 2017-08-07
[CSL-1472] Fix lwallet sendToAll: send change in transaction

georgeee:


gromakovsky:
@kantp is it with `DCONFIG=benchmark`?

gromakovsky:
The error message means that transaction's distribution doesn't mention all bootstrap stakeholders. This check is done only in bootstrap era and in benchmarks mode bootstrap era is disabled. So it's very strange. I will check the code, maybe there is a stupid bug somewhere.

gromakovsky:
@kantp ah, I just realized that #1249 is not merged yet. In that PR this bug is fixed.
That's the fix: https://github.com/input-output-hk/cardano-sl/pull/1249/files#diff-c9f5c3810acacb6501e720a2da519ab9R214 (swap arguments in `isBootstrapEra` call).
I will push this fix into a separate branch because it's quite crucial.

gromakovsky:
@kantp should be fixed in #1298.

gromakovsky:
@kantp unfortunately tests fail in that branch, but you can use that version already. I think that PR will be closed eventually.

kantp:
Thanks, @gromakovsky!

PR #1206, opened 2017-07-21
[CSL-1068] Migrate to newer Plutus

neongreen:
- different TxSigData
- different multisig scheme

PR #1294, opened 2017-08-08
[CSL-1068] Use new Plutus

neongreen:
This one should actually work probably.


gromakovsky:
Can we close #1206?

neongreen:
@gromakovsky done.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -210,17 +209,16 @@ verifyInputs VTxContext {..} resolvedInputs TxAux {..} =
         UnknownWitnessType t _ -> checkUnknownAddressType t addr
 
     -- the second argument here includes local context, can be used for scripts
```
gromakovsky: It's the first argument now.

neongreen: Okay, I actualized it.

PR #1295, opened 2017-08-08
[CSL-1453] Drop wreq from the stack

adinapoli-iohk:
It turned out to be quite easy to do, as we were using `wreq` in just one module. I have then run my `dependencies.sh` script to indeed prove `wreq` is now gone:

```
☁  iohk  cat deps_master.txt | grep wreq | wc -l
       0
```

### DIFF
--- a/infra/Pos/Reporting/Methods.hs
+++ b/infra/Pos/Reporting/Methods.hs
@@ -15,38 +15,44 @@ module Pos.Reporting.Methods
 
 import           Universum
 
-import           Control.Exception        (ErrorCall (..), SomeException)
-import           Control.Lens             (each, to)
-import           Control.Monad.Catch      (try)
-import           Data.Aeson               (encode)
-import           Data.Bits                (Bits (..))
-import qualified Data.HashMap.Strict      as HM
-import           Data.List                (isSuffixOf)
-import qualified Data.List.NonEmpty       as NE
-import qualified Data.Text.IO             as TIO
-import           Data.Time.Clock          (getCurrentTime)
-import           Formatting               (sformat, shown, stext, (%))
-import           Network.Info             (IPv4 (..), getNetworkInterfaces, ipv4)
-import           Network.Wreq             (partFile, partLBS, post)
-import           Pos.ReportServer.Report  (ReportInfo (..), ReportType (..))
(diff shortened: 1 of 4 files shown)


infra/cardano-sl-infra.cabal
```diff
@@ -126,6 +126,8 @@ library
                       , formatting
                       , generic-arbitrary
                       , hashable
+                      , http-client
```
adinapoli-iohk: We depend from `http-client` and `http-client-tls` anyway, so this doesn't add any extra dep, in fact.

infra/Pos/Reporting/Methods.hs
```diff
@@ -232,17 +238,25 @@ sendReport logFiles rawLogs reportType appName reportServerUri = do
...
         let pathsPart = map partFile' existingFiles
         let payloadPart =
-                partLBS "payload"
+                Form.partLBS "payload"
                 (encode $ reportInfo curTime $ existingFiles ++ memlogFiles)
-        e <- try $ liftIO $ post (reportServerUri <//> "report") $
-             payloadPart : (memlogPart ++ pathsPart)
+        -- If performance will ever be a concern, moving to a global manager
+        -- should help a lot.
+        reportManager <- liftIO (newManager tlsManagerSettings)
```
adinapoli-iohk: Note how this is not any worse than what `wreq` does, which is creating a new `Manager` everytime, unless there is one passed externally.

infra/Pos/Reporting/Methods.hs
```diff
@@ -232,17 +238,25 @@ sendReport logFiles rawLogs reportType appName reportServerUri = do
     withSystemTempFile "main.log" $ \tempFp tempHandle -> liftIO $ do
         for_ rawLogs $ TIO.hPutStrLn tempHandle
         hClose tempHandle
+        rq0 <- liftIO (parseUrlThrow $ reportServerUri <//> "report")
```
gromakovsky: Hm, it seems all these `liftIO`s inside `do` are redundant because there is `liftIO` outside (3 lines above).

adinapoli-iohk: Ok, well spotted, you are right!

PR #1303, opened 2017-08-09
[DEVOPS-203] Be able to compile stack2nix on macOS

domenkozar:


adinapoli-iohk:
Happy to report this did the trick for me when invoking `./pkgs/generate.sh` on darwin 😉

PR #1305, opened 2017-08-09
[CSL-1487] Tentative fix to disable the post-mortem tool

adinapoli-iohk:
See https://issues.serokell.io/issue/CSL-1487 for the full context.

This patch stops building the `cardano-post-mortem` tool by default, unless we are on Linux (CIwise).

@domenkozar @deepfire I've modified `default.nix` to explicitly pass the flag (as otherwise the `post-mortem-tool` would never build) but considering my limited knowledge of nix and the devops infrastructure, I've probably got it wrong 😉


domenkozar:
If you want to build it always with Nix, then it looks ok : ) I'd add a comment why this is enabled for Nix and not by default for future reference.

adinapoli-iohk:
> If you want to build it always with Nix, then it looks ok

I would say "yes", this is what we want! (But @gromakovsky can prove me wrong 😉  )

>  I'd add a comment why this is enabled for Nix and not by default for future reference.

Good idea! Will do as soon as I finish testing your PR which hopefully will fix the nix/darwin shenanigans 😁

gromakovsky:
We want to build it only for linux.

adinapoli-iohk:
I guess my nix-fu ends here, and I feel that nowadays the notion of "nix" and "linux" is actually more subtle than that as we use nix for both Darwin & Linux. Question being: how can I selectively enable the flag for Linux only?

domenkozar:
`configureFlags = optionals pkgs.stdenv.isLinux [ "-fwith-post-mortem" ];`

adinapoli-iohk:
> LGTM, but also please don't forget to put issue id into commit messages (then commits for an issue can be found from youtrack).

Ah, missed that requirement somehow! Will try to, in the future. Thanks for the heads up!

adinapoli-iohk:
Don't know what happened with the CI, I've tried restarting the build.

domenkozar:
@adinapoli-iohk rebase on master and run `./pkgs/generate.sh` and commit `pkgs/default.nix`

adinapoli-iohk:
Will do shortly. The only reason why I didn't do that is because I've got confused by the CI output. Usually we would get an explicit error message from the check script asking us to run `pkgs/generate.sh`, which looks like is not what happened here. Go figure!

adinapoli-iohk:
Uhm, weird. I can see how Travis is failing as it seems to be trying to satisfy those dependencies but they cannot (because nix thinks are not in the cabal manifest):

```
Aug 11 08:22:59 configureFlags: --verbose --prefix=/nix/store/p6lsgyxcxc7shg6r5vzw64d3vksqvwc6-cardano-sl-tools-0.5.1 --libdir=$prefix/lib/$compiler --libsubdir=$pkgid --with-gcc=gcc --package-db=/run/user/2000/nix-build-cardano-sl-tools-0.5.1.drv-0/package.conf.d --ghc-option=-optl=-Wl,-rpath=/nix/store/p6lsgyxcxc7shg6r5vzw64d3vksqvwc6-cardano-sl-tools-0.5.1/lib/ghc-8.0.2/cardano-sl-tools-0.5.1 --ghc-option=-j2 --disable-split-objs --disable-library-profiling --disable-profiling --enable-shared --disable-coverage --enable-library-vanilla --enable-executable-dynamic --enable-tests --ghc-option=-split-sections -fwith-post-mortem
Aug 11 08:23:01 Configuring cardano-sl-tools-0.5.1...
Aug 11 08:23:01 Setup: Encountered missing dependencies:
Aug 11 08:23:01 Chart -any,
Aug 11 08:23:01 Chart-diagrams -any,
Aug 11 08:23:01 fgl -any,
Aug 11 08:23:01 foldl -any,
Aug 11 08:23:01 graphviz -any,
Aug 11 08:23:01 pipes -any,
Aug 11 08:23:01 pipes-bytestring -any,
Aug 11 08:23:01 pipes-interleave -any,
Aug 11 08:23:01 pipes-safe -any
Aug 11 08:23:01 builder for ‘/nix/store/q7nyk8jjsnazvkbd6r16agyz7mlm21jd-cardano-sl-tools-0.5.1.drv’ failed with exit code 1
Aug 11 08:23:01 error: build of ‘/nix/store/q7nyk8jjsnazvkbd6r16agyz7mlm21jd-cardano-sl-tools-0.5.1.drv’ failed
```

@domenkozar any chance we need to extend our `pkgs/generate.sh` script once again so that even if we have a `build-depends` entry tucked into a conditional flag, the dependency resolution will do the right thing if such flag is enabled? (I hope I did explain myself clearly enough! If not, perhaps the [relevant part of the cabal manifest](https://github.com/input-output-hk/cardano-sl/blob/4476d55093e8d43215ae454273639208e1626033/tools/cardano-sl-tools.cabal#L580-L606) might help).

domenkozar:
The gist of the problem is that while generating package, the flag is not passed in :)

So the goal is to not build the tool while developing, but only build it on CI with Linux?

domenkozar:
A bit :) You're generating packages with the flag off, then trying to compile with the flag on :)

I don't see any good way to workaround this now, except for just enabling the post-mortem for linux only by default.

adinapoli-iohk:
@domenkozar It seems like I have reached an impasse on this PR as well 😱 

Travis is failing with all sort of creative errors, this being one:

```
Aug 22 12:44:40 error: anonymous function at /home/travis/build/input-output-hk/cardano-sl/pkgs/default.nix:1537:39 called without required argument ‘Chart’, at /nix/store/8h2rvq50a25gs1iasaycjlxksrqq0hbj-a905b7cd0c2dc0714195a50bf176cd8e4593502d.tar.gz/pkgs/development/haskell-modules/make-package-set.nix:66:27
Aug 22 12:44:40 (use ‘--show-trace’ to show detailed location information)
```

Any clue? 😞

adinapoli-iohk:
I have tried generating `pkgs/default.nix` from a linux box, let's see what the CI has to say about it.

adinapoli-iohk:
Making a fresh start out of this, as it's not going anywhere.

PR #1308, opened 2017-08-09
[CSL-1488] Remove custom `Small` types

rockbmb:


gromakovsky:
Recently I found `Small` in QuickCheck but with different semantics (it generates small numbers): https://www.stackage.org/haddock/lts-8.13/QuickCheck-2.9.2/Test-QuickCheck-Modifiers.html#t:Small
So it seems we need to come up with a different name to avoid possible confusion.
For example, `SmallSize`, but if you have better name in mind, go ahead.

PR #1316, opened 2017-08-10
[CSM-402] Repetive inputs

Martoon-00:
Filters duplicates in tx inputs.
Also includes small cosmetic changes.

I can't say it passed acceptance tests, but at least all tests which passed before this PR are ok now as well (list of failures: https://pastebin.com/TP5Dyb3s, I suppose problems may be caused by dev-mode usage actually).


src/Pos/Client/Txp/History.hs
```diff
@@ -86,13 +86,16 @@ getSenders UnsafeTx {..} = do
 data TxHistoryEntry = THEntry
     { _thTxId        :: !TxId
     , _thTx          :: !Tx
-    , _thInputs      :: ![TxOut]
     , _thDifficulty  :: !(Maybe ChainDifficulty)
-    , _thInputAddrs  :: ![Address]  -- TODO: remove in favor of _thInputs
+    , _thInputs      :: ![TxOut]
```
flyingleafe: :heart:

src/Pos/Wallet/Web/BListener.hs
```diff
@@ -84,7 +84,6 @@ onApplyTracking
     )
     => OldestFirst NE (Blund ssc) -> m SomeBatchOp
 onApplyTracking blunds = setLogger $ do
-    logWarning "This is ololo!"
```
flyingleafe: lold hardly

PR #1309, opened 2017-08-09
[CSL-1475] Fix tests in explorer mode

gromakovsky:
Tests were failing for a variety of reasons:
1. No `WITH_EXPLORER` in tests.
2. Generator and tests didn't respect `WITH_EXPLORER`.
3. `eRollbackToil` was processing txs in topological order, but it should do it in reverse order.

Apart from that:
1. Some refactoring was performed to simplify some code and make debugging easier.
2. Disabled tests were enabled in CI.
3. `sanityCheckBalances` was added to check explorer DB (only balances).
4. `sanityCheckDB` now happens after DB is modified.

PR #1320, opened 2017-08-10
[CSL-1247] Simplify CI for explorer

domenkozar:


domenkozar:
@ksaric FYI this will have failing test on linux until https://github.com/input-output-hk/cardano-sl/pull/1309 is fixed

ksaric:
I know, thanks, I'm looking at that second issue with the frontend. It appeared for several times now, I have to figure out what's wrong with that, even if this magically works after switching to NPM (I doubt it).

scripts/ci/travis.sh
```diff
@@ -67,35 +67,24 @@ done
...
 
+# Generate daedalus-bridge
 pushd daedalus
   nix-shell --run "npm install && npm run build:prod"
   echo $TRAVIS_BUILD_NUMBER > build-id
   cp ../log-config-prod.yaml .
 popd
 
+# Generate explorer frontend
+EXPLORER_EXECUTABLE=$(pwd)/cardano-sl-explorer-static.root/bin/cardano-explorer-hs2purs ./explorer/frontend/scripts/build.sh
```
ksaric: :confused: Wait. What? It's in a single line. And also, why did you put `$(pwd)` in it? Wasn't `.` enough?

domenkozar: What's wrong with a single line? :) Could also export in separate line if you'd like.

The reason for pwd is that the script maniuplates pwd, so it has to be absolute otherwise resolving relative path would be hard

ksaric: Nothing wrong with a single line, I'm just used to having one line reserved for a variable I guess. Also, I didn't know that this way you could pass a variable to the script, I thought you always needed to _export_ it first. Nice.

PR #1321, opened 2017-08-10
Merge cardano-sl-0.5 to master

Martoon-00:
This includes changes made in #1316


src/Pos/Client/Txp/History.hs
```diff
@@ -92,13 +92,16 @@ getSenders UnsafeTx {..} = do
 data TxHistoryEntry = THEntry
     { _thTxId        :: !TxId
     , _thTx          :: !Tx
-    , _thInputs      :: ![TxOut]
     , _thDifficulty  :: !(Maybe ChainDifficulty)
-    , _thInputAddrs  :: ![Address]  -- TODO: remove in favor of _thInputs
```
akegalj: is frontend UI using this field ? If yes, we should notify them how to get Addresses without that field @Martoon-00

Martoon-00: This is fully internal type. Client api operates with `CTx` instead, which remained unchanged

gromakovsky:
@Martoon-00 there are conflicts, resolve them please. Also please request review more intrusively.

Martoon-00:
❤️

PR #1322, opened 2017-08-10
[CSL-1360] Test applying fork to blockchain

int-index:


src/Pos/Block/Slog/Logic.hs
```diff
@@ -257,7 +257,7 @@ slogRollbackBlocks blunds = do
     maxSeenDifficulty <- GS.getMaxSeenDifficulty
     resultingDifficulty <-
         maybe 0 (view difficultyL) <$>
-        blkGetHeader @ssc (NE.head (getNewestFirst blunds) ^. prevBlockL)
+        blkGetHeader @ssc (NE.head (getOldestFirst . toOldestFirst $ blunds) ^. prevBlockL)
```
gromakovsky: _mdaush_

test/Test/Pos/Block/Logic/Event.hs
```diff
@@ -77,15 +73,47 @@ runBlockEvent ::
...
+        BlockRollbackFailure _ -> BlockEventSuccess (IsExpected False)
+    onFailure (e :: SomeException) = case ev ^. berOutValid of
+        BlockRollbackSuccess -> BlockEventFailure (IsExpected False) e
+        BlockRollbackFailure brf ->
+            let
+                isExpected = case brf of
+                    BlkRbSecurityLimitExceeded
+                        | Just cfe <- fromException e
+                        , CardanoFatalError msg <- cfe
+                        , "security risk" `T.isInfixOf` msg ->
```
gromakovsky: Ah, I really don't like when someone parses error messages to determine the type of error :/
I find it much better to represent different types of errors with different constructors.
For example, it can be another constructor of `CardanoFatalError`.

test/Test/Pos/Block/Logic/Event.hs
```diff
@@ -77,15 +73,47 @@ runBlockEvent ::
...
+                          True
+                        | otherwise ->
+                          False
+            in
+                BlockEventFailure (IsExpected isExpected) e
+
 runBlockEvent (BlkEvSnap ev) =
-   (BlockEventSuccess <$ runSnapshotOperation ev)
-      `catch` (return . BlockEventDbChanged)
+  (onSuccess <$ runSnapshotOperation ev)
```
gromakovsky: Minor: 2 spaces, should be 4.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -283,3 +296,63 @@ verifyBlockScenarioResult = \case
...
+-- Forks
+----------------------------------------------------------------------------
+
+singleForkSpec :: ForkDepth -> Spec
+singleForkSpec fd = do
+    prop singleForkDesc (singleForkProp fd)
+  where
+    singleForkDesc =
+      "a blockchain of length q=0..(9.5*k) blocks can switch to a fork " <>
+      "of length j>i with a common prefix i, d=q-i"
```
gromakovsky: What is `d`?

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -283,3 +296,63 @@ verifyBlockScenarioResult = \case
...
+    runBlockScenarioAndVerify scenario
+
+data ForkDepth = ForkShort | ForkMedium | ForkDeep
+
+genSingleFork :: forall g m. (RandomGen g, Monad m) => ForkDepth -> BlockEventGenT g m ()
+genSingleFork fd = do
+    let k = fromIntegral blkSecurityParam
+    d <- getRandomR $ case fd of
+        ForkShort  -> (1, k - 1)
+        ForkMedium -> (max 1 (k-2), k+2)
```
gromakovsky: It would be good to write type of `k` explicitly, because I have no idea whether it's signed or not and if it's not, then it won't work well for `k = 1` (which is not a big problem, but still).

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -283,3 +296,63 @@ verifyBlockScenarioResult = \case
...
+        ForkDeep   -> (k+1, div (k*3) 2)
```
gromakovsky: In `ForkDeep` and `ForkShort` case the first value will be greater than the second one if `k = `, but perhaps we don't care.

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -283,3 +296,63 @@ verifyBlockScenarioResult = \case
...
+        -- max index of the common prefix. i>0 because d<q
+        i = q-d
+    -- fork blockchain max index j>i. the upper bound is arbitrary.
+    -- dj=j-i
+    dj <- getRandomR (1, d*2)
+    -- now we can generate paths
+    --
+    -- B0 - B1 - B2 - B3 - B4 - B5 - B6 - B7
+    --              \
+    --                C3 - C4 - C5 - C6
```
gromakovsky: By this moment it's hard to keep track of what is `d`, `dj`, `i`, `q`, etc. If you somehow illustrate them with this picture (at least write their values for this case), it will make it easier to understand what's going on.

test/Test/Pos/Block/Logic/Event.hs
int-index: I didn't enjoy doing it either, but I'm not looking for an overhaul of error handling (at least within this task).

int-index:
I'm going to proceed in #1351

test/Test/Pos/Block/Logic/VarSpec.hs
int-index: Actually, I was going to write their values below this illustration but forgot to do it.

PR #1329, opened 2017-08-11
[CSL-1203] Test basic delegation with tx distribution 

naryl:
Patch to allow lwallet make transactions with custom stake distribution.

### DIFF
--- a/lwallet/Command.hs
+++ b/lwallet/Command.hs
@@ -23,9 +23,10 @@ import           Pos.Binary                 (deserialize')
 import           Pos.Core.Types             (ScriptVersion)
 import           Pos.Core.Version           (parseBlockVersion, parseSoftwareVersion)
 import           Pos.Crypto                 (Hash, PublicKey, decodeHash)
-import           Pos.Txp                    (TxOut (..))
+import           Pos.Txp                    (TxOut (..), TxOutDistribution)
 import           Pos.Types                  (Address (..), BlockVersion, Coin, EpochIndex,
-                                             SoftwareVersion, decodeTextAddress, mkCoin)
+                                             SoftwareVersion, StakeholderId,
+                                             decodeTextAddress, mkCoin)
 import           Pos.Update                 (SystemTag, UpId, mkSystemTag)
 
 -- | Specify how transactions are sent to the network during benchmarks using 'SendToAllGenesis'.
@@ -38,6 +39,7 @@ data SendMode =
 data Command
     = Balance Address
     | Send Int (NonEmpty TxOut)
+    | SendDistr Int TxOut TxOutDistribution
     | SendToAllGenesis !Int !Int !Int !SendMode !FilePath
       -- ^ In-order: number of txs to send per-thread, number of threads, and
       -- delay (milliseconds) between sends.
@@ -111,6 +113,20 @@ switch = lexeme $ positive $> True <|>
     positive = text "+" <|> text "y" <|> text "yes"
     negative = text "-" <|> text "n" <|> text "no"
 
+stakeholderId :: Parser StakeholderId
+stakeholderId = lexeme $ do
+    str <- many1 alphaNum
+    case decodeTextAddress (toText str) of
+        Left err                            -> fail (toString err)
+        Right (PubKeyAddress addrKeyHash _) -> pure addrKeyHash
+        Right _                             -> fail "Wrong address type"
+
+txoutDistrSingle :: Parser (StakeholderId, Coin)
+txoutDistrSingle = (,) <$> stakeholderId <*> coin
+
+txoutDistribution :: Parser TxOutDistribution
+txoutDistribution = (many txoutDistrSingle)
+
 balance :: Parser Command
 balance = Balance <$> address
 
@@ -131,6 +147,9 @@ addKeyFromFile = AddKeyFromFile <$> lexeme (many1 anyChar)
 
 send :: Parser Command
 send = Send <$> num <*> (NE.fromList <$> many1 txout)
+
+sendDistr :: Parser Command
+sendDistr = SendDistr <$> num <*> txout <*> txoutDistribution
 
 sendMode :: Parser SendMode
 sendMode = lexeme $ text "neighbours" $> SendNeighbours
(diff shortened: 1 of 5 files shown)


lwallet/Command.hs
```diff
@@ -111,6 +113,20 @@ switch = lexeme $ positive $> True <|>
     positive = text "+" <|> text "y" <|> text "yes"
     negative = text "-" <|> text "n" <|> text "no"
 
+stakeholderId :: Parser StakeholderId
+stakeholderId = lexeme $ do
+    str <- many1 alphaNum
+    case decodeTextAddress (toText str) of
```
gromakovsky: This is wrong, why do you parse `StakeholderId` from `Address`? There is `Pos.Crypto.Hashing.decodeAbstractHash` to parse hash.
The main reason why it's wrong is that soon it will be impossible to extract `StakeholderId` from `Address`.

lwallet/Main.hs
```diff
@@ -166,6 +167,19 @@ runCmd sendActions (Send idx outputs) CmdCtx{na} = do
...
+        withSafeSigner (skeys !! idx) (pure emptyPassphrase) $ \mss ->
+        runEitherT $ do
+            ss <- mss `whenNothing` throwError "Invalid passphrase"
+            lift $ submitTxNoOverride
+                (immediateConcurrentConversations sendActions na)
+                ss
+                (one $ TxOutAux output distr)
+    case etx of
+        Left err -> putText $ sformat ("Error: "%stext) err
+        Right tx -> putText $ sformat ("Submitted transaction: "%txaF) tx
```
gromakovsky: http://i.imgur.com/ApNwUtZ.png
I don't find such copy-paste acceptable. Also it would be good to get rid of `EitherT`, because we use `ExceptT`.

scripts/launch/wallet.sh
```diff
@@ -9,7 +9,10 @@ if [[ "$SSC_ALGO" != "" ]]; then
...
 fi
 
+system_start=100500 # random value, not used, but mandatory
+
 $(find_binary cardano-wallet) $(peer_config 0) $(logs smartwallet$i.log) \
                               --db-path "run/wallet-db" --rebuild-db \
                               --flat-distr "(3, 100000)" $ssc_algo "$@" \
-                              --system-start 100500  # random value, not used, but mandatory
+                              --system-start $system_start \
+                              repl
```
gromakovsky: Why `repl`? I am quite surprised that repl works. And I have never used it, because I find it more convenient to pass commands as arguments.
Anyway, I think `wallet.sh` should support both modes, just like `cardano-wallet` does (so if you need repl, you can use `wallet.sh repl`).
You can see how I modified it in #1215.

src/Pos/Communication/Tx.hs
```diff
@@ -71,6 +72,19 @@ submitMTx enqueue hdwSigner outputs = do
...
     submitAndSave enqueue txw
 
+-- | Construct Tx using secret key and given list of desired outputs without overriding TxDistr
+submitTxNoOverride
+    :: TxMode ssc ctx m
+    => EnqueueMsg m
+    -> SafeSigner
+    -> NonEmpty TxOutAux
+    -> m TxAux
+submitTxNoOverride enqueue ss outputs = do
```
gromakovsky: Copy-paste.

src/Pos/Client/Txp/Util.hs
```diff
@@ -225,6 +226,22 @@ createMTx utxo hwdSigners outputs =
         HM.lookup (AddressIA addr) signers
 
 -- | Make a multi-transaction using given secret key and info for
+-- outputs. Don't override TxDistr
+createTxNoOverride ::
```
gromakovsky: What's the purpose of this function? It can create invalid transactions. Do I understand correctly that the only purpose is to test that invalid transaction is rejected? If yes, it should be documented.

src/Pos/Communication/Tx.hs
naryl: It's just for testing too. I don't think it's a good idea to mix working and testing control flows.

So I'll just document it.

lwallet/Command.hs
gromakovsky: @naryl see #1340, you can reuse it.

naryl: There's a bigger problem then. `TxOutDistribution` is a `[(StakeholderId, Coin)]`. How can the lwallet user get `StakeholderId`s? Currently the only thing they have is `listaddr` which lists addresses and PKs.

gromakovsky: `StakeholderId` is `addressHash pk`. The easiest solution is to list not only addresses and PKs, but also `StakeholderId`s. Or you can add another command for it.

naryl: I'm trying to write with `hashHexF` ... `addressHash pk`, then read with `decodeAbstractHash`. It results in error:
```
 StakeholderId: d708352cbdad7e4e579df1d6e3482ac002121fcd2f085790f6b1da1a
> send-distr 0 jbRJqzVmBoUNqAQBykQsAeFC12GEa4ysPqiGidKSrNzU5shUzxdPcVMRhWPN 33333 6056d5a781311a76d5cd7b4f43823de687bb0cba5c4482bf4579fc9d 33333
cardano-wallet: decode hash error: decodeFull failed for AbstractHash Blake2b_224 PublicKey: DeserialiseFailure 0 "expected bytes"
  error, called at src/Debug.hs:43:11 in universum-0.4.3-5s0jbvCPrRj9dDakAh9rIJ:Debug
```

gromakovsky: Oh, `decodeAbstractHash` is just wrong :(
I will fix it and commit to the branch used in #1340.

gromakovsky: @naryl fixed in  #1340, waiting for review from @volhovM.

gromakovsky: @naryl you may have a problem, because some code uses `Parsec Text`, but other code uses `Parsec String`.
I am currently getting rid of `Parsec String`.
Unfortunately, I was in hurry and didn't do it in a separate commit :(
If this PR can wait, it's better to wait until #1342 is ready and merged.

naryl: I don't think anyone other than me needs it, so I copypasted your parser changing String for Text for now, will remove the copy and use your function before merging.

gromakovsky: Thanks!

gromakovsky: @naryl my PR was merged.

PR #1332, opened 2017-08-12
[CSL-1457] Use genesisProdBootStakeholders in block-gen

int-index:


tools/src/block-gen/Main.hs
```diff
@@ -51,14 +52,17 @@ main = flip catch catchEx $ do
 
     let nodes = length invSecretsMap
     let flatDistr = FlatStakes (fromIntegral nodes) (mkCoin $ fromIntegral nodes)
-    let bootStakeholders =
-            GenesisWStakeholders $ HM.fromList $
-            zip (HM.keys $ unInvSecretsMap invSecretsMap) (repeat 1)
+    let bootStakeholders
+            | isDevelopment =
+                GenesisWStakeholders $ HM.fromList $
```
volhovM: I think you might want to use `generateWStakeholders` instead for dev mode. Does `block-gen` only uses flat stakes for it?

pva701: it does.

pva701: I and @georgeee decided not to add other distribution and just created tasks for it. we hoped it will be added when it's needed

PR #1333, opened 2017-08-12
[CSM-405] Sort transactions on timestamp if known

Martoon-00:
In tracking, when get transactions from new block, in general we have no idea about order of those transactions. Only wallet db may know that, since it stores tx timestamps - I use them to achieve correct order.

_I didn't run DAT because they work unevenly (_

### DIFF
--- a/core/Pos/Core/Timestamp.hs
+++ b/core/Pos/Core/Timestamp.hs
@@ -6,13 +6,16 @@ module Pos.Core.Timestamp
        , getCurrentTimestamp
        , diffTimestamp
        , addMicrosecondsToTimestamp
+
+       -- * Convertions
+       , timestampToPosix
        ) where
 
 import           Universum
 
 import           Data.Text.Buildable   (Buildable)
 import qualified Data.Text.Buildable   as Buildable
-import           Data.Time.Clock.POSIX (getPOSIXTime)
+import           Data.Time.Clock.POSIX (POSIXTime, getPOSIXTime)
 import           Data.Time.Units       (Microsecond)
 import           Formatting            (Format, build)
 import qualified Prelude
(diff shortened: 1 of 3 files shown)


src/Pos/Wallet/Web/Tracking.hs
```diff
@@ -417,8 +418,16 @@ applyModifierToWallet wid newTip CAccModifier{..} = do
     mapM_ (WS.addCustomAddress ChangeAddr . fst) (MM.insertions camChange)
     WS.getWalletUtxo >>= WS.setWalletUtxo . MM.modifyMap camUtxo
     oldCachedHist <- fromMaybe [] <$> WS.getHistoryCache wid
-    WS.updateHistoryCache wid $ DL.toList camAddedHistory <> oldCachedHist
+    sortedAddedHistory <- sortTxs (DL.toList camAddedHistory)
```
flyingleafe: Actually, we have idea about transaction order in block - they are sorted _topologically_, which is more substantial than sorting by timestamps (in my opinion). Of course, in cases when transactions are topologically incomparable, they may permute, and in these cases sorting them by timestamp makes sense.

But here you sort all transactions by their timestamps, and these transactions are very likely to not have a timestamp record in metadata yet. You should at least use `_thTimestamp` field of `TxHistoryEntry` in case if there's no `TxMeta` record found - in this case you at least won't accidentally place transactions from old blocks, for which there's no metadata for some reason (it's very possible, afaik: when restoring history from the blockchain in the first time there's no metadata for transactions).

The best solution would require somehow maintaining info about partial order after topsort, but this seems really complicated, so please just consider `_thTimestamp` here and it should be somewhat fine.

core/Pos/Core/Timestamp.hs
```diff
@@ -53,3 +56,6 @@ diffTimestamp t1 t2 = getTimestamp t1 - getTimestamp t2
 
 addMicrosecondsToTimestamp :: Microsecond -> Timestamp -> Timestamp
 addMicrosecondsToTimestamp m t = Timestamp { getTimestamp = (getTimestamp t) + m }
+
+timestampToPosix :: Timestamp -> POSIXTime
+timestampToPosix (Timestamp ts) = fromIntegral ts / 1000000
```
Martoon-00: As far as I understand, in `cardano-sl-0.5` there is no explorer in this project yet, but I can replace that function on merging this PR to `master`

PR #1334, opened 2017-08-13
Update some scripts

gromakovsky:
I updated some scripts, they were obsolete.


scripts/launch/qa.sh
```diff
@@ -5,7 +5,6 @@ stack exec -- cardano-node \
     --log-config scripts/log-templates/log-config-qa.yaml \
     --logs-prefix "logs/qanet" \
     --db-path db-qanet \
-    --wallet \
```
pva701: why?

gromakovsky: Because `--wallet` option was removed.

PR #1335, opened 2017-08-13
[DEVOPS-50] Nixos tests

cleverca22:


tests/make-genesis.nix
```diff
@@ -0,0 +1,6 @@
+{ cardano-sl-tools, runCommand }:
+
+runCommand "genesis" { buildInputs = [ cardano-sl-tools ]; } ''
```
domenkozar: This could now call `./scripts/generate/genesis.sh` since it has nix support. It's important to the the genesis script anyway :)

domenkozar:
```
in job ‘tests.simpleNodeNixOps.x86_64-linux’:
file ‘nixops/nix/options.nix’ was not found in the Nix search path (add it using $NIX_PATH or -I), at [1m/nix/store/7vx0zx8bnck08rkc2wmpbnqxiv90m7iz-git-export/tests/simple-node-nixops.nix[0m:13:84
```

domenkozar:
afaik cardano-node already produces JSON logs in /var/lib/cardano-node/

cleverca22:
`Invalid option `--testnet-stake'`

something has changed in master, i'll need to update the genesis generation

domenkozar:
@cleverca22 https://github.com/input-output-hk/cardano-sl/commit/657063362d#diff-f246ff0f5accac42081c32efcc2bb2e0R62

tests/simple-node-nixops.nix
```diff
@@ -0,0 +1,26 @@
...
+      services.cardano-node = {
+        autoStart = true;
+        initialPeers = [];
+      };
+    };
+  };
+  testScript = ''
+    startAll
+    $machine->waitForUnit("cardano-node.service");
+    # TODO, implement sd_notify?
```
domenkozar: FYI: code now calls Systemd.notifyReady to get rid of this, although it should be a separatge issue outside of this PR.

default.nix
```diff
@@ -45,6 +46,27 @@ let
           "--ghc-options=-DCONFIG=${dconfig}"
           "--ghc-options=-DGITREV=${gitrev}"
         ];
+      } // optionalAttrs (genesis != null) {
```
domenkozar: FYI: this will be simplified with https://github.com/input-output-hk/cardano-sl/pull/1463

domenkozar:
@kantp @adinapoli-iohk @edsko could you review the haskell part o the post-mortem?

tools/src/post-mortem/Main.hs
```diff
@@ -18,7 +21,7 @@ main = do
             err $ "sample probability: " ++ show sampleProb
             err ""
             xs <- forM logDirs $ flip processLogDirOverview sampleProb
-            chart xs "times.png"
+            chart xs "times.svg"
```
adinapoli-iohk: What's the story behind the change of this extension?

tools/src/post-mortem/JSONLog.hs
```diff
@@ -52,7 +53,7 @@ instance Ord IndexedJLTimedEvent where
 runParseLogs :: FilePath -> (Producer IndexedJLTimedEvent IO () -> IO r) -> IO r
 runParseLogs logDir f = do
     xs <- jsonLogs logDir
-    runWithFiles xs ReadMode $ \ys -> f $ interleave (map (uncurry producer) ys)
+    runWithFiles (assert (xs /= []) xs) ReadMode $ \ys -> f $ interleave (map (uncurry producer) ys)
```
adinapoli-iohk: If you want to ensure you have a non-empty list, you might want to consider `Data.List.NonEmpty` if possible here 😉

tools/src/post-mortem/JSONLog.hs
```diff
@@ -39,7 +40,7 @@ data IndexedJLTimedEvent = IndexedJLTimedEvent
     { ijlNode      :: !NodeIndex
     , ijlTimestamp :: !Timestamp
     , ijlEvent     :: !JLEvent
-    }
+    } deriving (Show)
```
adinapoli-iohk: Parentheses around `Show` are not needed for a single deriving.

tools/src/post-mortem/Main.hs
```diff
@@ -36,12 +39,37 @@ main = do
             err $ "wait window: " ++ show waitWindow
             err ""
             for_ logDirs $ processLogDirThroughput txWindow waitWindow
+        TestPostProcess logDir -> do
+            postProcessLogs logDir
+            err "done"
```
adinapoli-iohk: I find a bit strange signalling the fact you are "done" by using `err`. What does `err` do? Cannot we simply print using `putText` or similar? Ah, I think `err` might be simply logging to `stderr`, is that right? In which case I think it's fine 😉

tools/src/post-mortem/Options.hs
```diff
@@ -54,12 +55,16 @@ throughputOptions = Throughput <$> (argument auto
                                     <> help "directories containing the log files"
                                     )))
 
+testPostProcess :: Parser Options
+testPostProcess = TestPostProcess <$> (argument str (metavar "LOGDIR" <> help "directory containing the log files"))
```
adinapoli-iohk: I might be wrong (playing this in my head), but cannot you use `strOption` ( or `strArgument`) here directly?

tools/src/post-mortem/Statistics/HydraMetrics.hs
```diff
@@ -0,0 +1,58 @@
...
+findBlockChainState = Fold combine (ChainStateInternal (0,0) "error" mempty) finish
+  where
+    finish :: ChainStateInternal -> ChainState
+    finish internal = ChainState internal chainLength (((fromIntegral slot) + 1) + ((fromIntegral epoch) * epochLength))
+      where
+        chainLength = measureLength $ M.lookup (topMostBlock internal) (blocks internal)
+        (epoch, slot) = topMostSlot internal
+        epochLength = 1080 -- TODO, look it up somehow
+    measureLength :: Maybe BlockWrapper -> Integer
+    measureLength (Just block) = 1 + measureLength (parent block)
```
adinapoli-iohk: For very big chains, you might want to go with a tail-recursive version of this function, where you would accumulate the counter in an aux variable and force the sum with a `BangPattern` to avoid building too many thunks.

tools/src/post-mortem/Main.hs
cleverca22: it outputs an svg image, but calls it `.png`, and my image viewer refuses to open such mis-named files

tools/src/post-mortem/Main.hs
cleverca22: yeah, all of the existing code uses err, which is just an alias for print to stderr

tests/simple-node-nixops.nix
cleverca22: oops, it is using notify now, i'll fix that

tools/src/post-mortem/Options.hs
cleverca22: was using all of the other options in this file as an example, havent seen this library before

tools/src/post-mortem/Statistics/HydraMetrics.hs
cleverca22: yeah, i can see that helping, though right now the chain doesnt go over 30 blocks, i'll add a todo

cleverca22:
this PR is currently on-hold and will be rewritten later to fix all of the conflicts

adinapoli-iohk:
@cleverca22 What should we be doing with this bad boy? Close or revive? 😁

PR #1337, opened 2017-08-14
CSE-198 Add size, median size for transaction.

ksaric:
https://issues.serokell.io/issue/CSE-198

I can't run _explorer_ in any mode, needed for testing this.


ksaric:
Seems `CODEOWNERS` and `PULL_REQUEST_TEMPLATE.md` are being ignored when not on project root. I will remove them.

PR #1338, opened 2017-08-14
[CSL-1458] Allow txs with outputs less then bootDustThd

volhovM:
* `genesisSplitCoin` doesn't fail when asked to split coin less then `bootDustThd`
* Transaction outputs wth value less then `bootDustThd` are now considered valid.

### DIFF
--- a/core/Pos/Core/Genesis.hs
+++ b/core/Pos/Core/Genesis.hs
@@ -127,17 +127,18 @@ genesisSplitBoot ::
        GenesisWStakeholders -> Coin -> [(StakeholderId, Coin)]
 genesisSplitBoot g@(GenesisWStakeholders bootWHolders) c
     | c < bootDustThreshold g =
-          snd $
-          foldr (\(s,w) r@(totalSum, res) ->
-                   if totalSum >= cval then r
-                   else let w' = (fromIntegral w :: Word64)
-                            toInclude = fromIntegral $ min w' (totalSum - w')
-                        in ( totalSum + toInclude
-                           , (s, mkCoin toInclude):res))
-                (0::Word64,[])
-                (HM.toList bootWHolders)
-    | otherwise = bootHolders `zip` stakeCoins
+          snd $ foldr foldrFunc (0::Word64,[]) (HM.toList bootWHolders)
+    | otherwise =
+          bootHolders `zip` stakeCoins
   where
+    foldrFunc (s,w) r@(totalSum, res) = case compare totalSum cval of
+        EQ -> r
+        GT -> error "genesisSplitBoot: totalSum > cval can't happen"
+        LT -> let w' = (fromIntegral w :: Word64)
+                  toInclude = bool w' (cval - totalSum) (totalSum + w' > cval)
+              in (totalSum + toInclude
+                 ,(s, mkCoin toInclude):res)
+
     weights :: [Word64]
     weights = map fromIntegral $ HM.elems bootWHolders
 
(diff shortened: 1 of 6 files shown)


gromakovsky:
Did you test tx sending with small output?

volhovM:
@gromakovsky no, but it's covered by tests.

gromakovsky:
Can you test it either from lwallet or web api?

core/Pos/Core/Genesis.hs
```diff
@@ -119,19 +119,24 @@ generateHdwGenesisSecretKey =
...
-               "threshold is "%build%" in respect to boot stakeholders set "%build)
-              c
-              (bootDustThreshold g)
-              g
-    | otherwise = Right $ bootHolders `zip` stakeCoins
+          snd $
+          foldr (\(s,w) r@(totalSum, res) ->
+                   if totalSum >= cval then r
+                   else let w' = (fromIntegral w :: Word64)
+                            toInclude = fromIntegral $ min w' (totalSum - w')
```
gromakovsky: 1. `fromIntegral` is probably redundant here.
2. I don't like that you abuse signed overflow here.
3. I don't understand how it works. Suppose stakeholders have weights 7, 2, 3. `cval` is 10.
First step: min 7 (-7) = 7 (overflow abuse).
Second step: min 2 5 = 2
Third step: min 3 6 = 3.

7 + 2 + 3 = 12 > 10.
Where am I wrong?

txp/Pos/Txp/Toil/Logic.hs
```diff
@@ -180,14 +182,19 @@ verifyGState curEpoch txAux txFee = do
...
-    getKeys bootHolders == HS.fromList stakeholders &&
-    -- Every stakeholder gets his divisor. It's safe to use ! here
-    -- because we're already sure that bootHolders ~ addrs is
-    -- bijective.
-    all (\(a,c) -> c >= (minimumPerStakeholder HM.! a)) txOutDistr
+bootRelatedDistr g@(GenesisWStakeholders bootHolders) txOutDistr
+    -- We allow small outputs to attribute stake to whoever because
+    -- it's not likely they will shift overall stake distribution
+    -- much.
+    | coinSum < unsafeGetCoin (bootDustThreshold g) = True
```
gromakovsky: For me it makes more sense to only allow genesis stakeholders in distribution (but not necessary all genesis stakeholders).

core/Pos/Core/Genesis.hs
```diff
@@ -119,20 +119,26 @@ generateHdwGenesisSecretKey =
...
-    | otherwise = Right $ bootHolders `zip` stakeCoins
+          snd $ foldr foldrFunc (0::Word64,[]) (HM.toList bootWHolders)
+    | otherwise =
+          bootHolders `zip` stakeCoins
   where
+    foldrFunc (s,w) r@(totalSum, res) = case compare totalSum cval of
+        EQ -> r
+        GT -> error "genesisSplitBoot: totalSum > cval can't happen"
+        LT -> let w' = (fromIntegral w :: Word64)
+                  toInclude = bool w' (totalSum - w') (totalSum + w' > cval)
```
gromakovsky: I still don't understand this logic.
Suppose we process weights 7, 3, 2. 7 will have 7, 3 will have 3, 2 will have 8. :wtf:
Perhaps `totalSum - w'` should be replaced with `cval - totalSum`?
But then there is another problem: 2 will have 0, which is bad. So you should check whether `toInclude` is 0.
Anyway, this code is drunk, I don't get it.

volhovM: I'm hopeless. Yes, I actually intended to write `cval - totalSum`. If we process `7 3 2` with `cval = 10` then `compare totalSum cval` will evaluate to `EQ` and we'll return `r`.

PR #1340, opened 2017-08-14
CSL-1499 Rework StakeholderId parsing AND CSL-1504 decodeAbstractHash is broken 

gromakovsky:


gromakovsky:
MacOS build timed out too many times in a row.

PR #1298, opened 2017-08-08
[CSL-1472] Fix isBootstrapEra check in txp

gromakovsky:
The arguments were swapped. I slightly refactored it to make such bug impossible.

PR #1341, opened 2017-08-14
[CSL-1436] Find better solution for overriding `k` for tests in CI

pva701:


src/node/Main.hs
```diff
@@ -170,6 +161,9 @@ action args@Args {..} = do
 #endif
     case (userWantsWallet, sscParams) of
         (False, Left par)  -> actionWithoutWallet @SscNistBeacon par currentParams
-        (False, Right par) -> actionWithoutWallet @SscGodTossing par currentParams
+        (False, Right par) ->
+            -- TODO: What the heck?!
+            -- @actionWithoutWallet@ requires HasCoreConstants, why?
```
pva701: @int-index can you help me to understand reason of such behaviour?

core/Pos/Core/Context.hs
```diff
@@ -9,9 +24,57 @@ module Pos.Core.Context
...
+-- | Core constants. They should be really constant and never change.
+data CoreConstants = CoreConstants
+    { _ccBlkSecurityParam :: !BlockCount
+    }
+
+makeLenses ''CoreConstants
+
+type HasCoreConstants = Given CoreConstants
+
+giveConsts :: BlockCount -> (HasCoreConstants => r) -> r
```
gromakovsky: Why does this function take `BlockCount` instead of `CoreConstants`? Later `CoreConstants` will be extended.

test/Test/Pos/CborSpec.hs
```diff
@@ -302,7 +303,7 @@ testAgainstFile name x expected =
               Right actual -> x `shouldBe` actual
 
 spec :: Spec
-spec = describe "Cbor.Bi instances" $ do
+spec = giveTestsConsts $ describe "Cbor.Bi instances" $ do
```
gromakovsky: I think it's better to use real constants in this particular module.

gromakovsky: Actually I think it's better to use test constants only in modules where running time depends on `k`.

gromakovsky:
@pva701 I fixed my commit and force-pushed.

core/Pos/Core/Context.hs
pva701: >Later CoreConstants will be extended.

this. It will take several constants correspond to `CoreConstants`

gromakovsky: Several constants as arguments? What's the purpose of it? It's less convenient and more error-prone. For example, if some arguments are `Num`s you can easily confuse them if you pass numeric literals.

pva701: lol, the same situation if it takes `CoreConstants`. you might easily confuse them when pass to `CoreConstants` constructor.

gromakovsky: No, `CoreConstants` has records.

pva701: wtf? you can consider this function as `give . CoreConstants`. constructor and this function are absolutely same.
if you meant 
```
field1 <- fromIntegral a
field2 <- fromIntegral b
...
```
you can also confuse records

pva701:
>My commit was spoiled, I will probably fix it by myself.

where?

test/Test/Pos/CborSpec.hs
pva701: what do you mean? do you suggest using `staticBlk...` everywhere besides `VarSpec` (and maybe some other tests)?

core/Pos/Core/Context.hs
pva701: >Several constants as arguments? What's the purpose of it? It's less convenient and more error-prone. 

we are using `mkSafeConstructor...` everywhere, it's yet another such place.

gromakovsky:
> where?

I fixed it already, I don't think it makes sense to search for it now.

test/Test/Pos/CborSpec.hs
gromakovsky: Yes, static constants everywhere except `VarSpec` and LRC worker spec.

core/Pos/Core/Context.hs
gromakovsky: > we are using mkSafeConstructor... everywhere, it's yet another such place.

We use it only for types with internal invariants. We don't have any invariants for core constants.

I can write
```
myConstants = CoreConstants { _ccBlkSecurityParam = 5, _ccAnotherNumber = 10 }
```
and even if someone changes the order of this values in `CoreConstants`, the code will remain correct.

pva701:
>I fixed it already, I don't think it makes sense to search for it now.

what about disses and debate in comments? does it make sense, doesn't it?

core/Pos/Core/Context.hs
pva701: яннп, you might reorder 10 and 5 and get the same troubles as with the current approach. what are we talking about?

gromakovsky:
https://github.com/input-output-hk/cardano-sl/commit/5148e697a3ff5e5669af6b6533fe632bbc4648ae#commitcomment-23642175

gromakovsky:
This is my very first version of that commit: https://github.com/input-output-hk/cardano-sl/commit/897984505bea860ee566c48595a38c75546c6f01

core/Pos/Core/Context.hs
gromakovsky: When I write `giveConsts 5 10` I have no idea what is 5 and what is 10.
When I write `CoreConstants { _ccBlkSecurityParam = 5, _ccAnotherNumber = 10 }` it's clear what is 5 and what is 10. And it's resistant to order changes.

pva701: lol, you are saying like someone who doesn't know our code will write this.
you can look at the definition of `giveConsts` and realize what it's 5 and 10

gromakovsky: You could look at the definition of `isBootstrapEra` and realize which argument is current epoch and which one is unlock epoch.
But you didn't do it and introduced a bug (https://github.com/input-output-hk/cardano-sl/pull/1298).
There are 100500 examples when someone confuses order of arguments.

pva701: Lol, it's meh. let's use data types everywhere where we can reorder arguments, you suggest to solve a minor problem again. it doesn't save us as `BlockAmount` and similar labuda. 
I don't see sense in this change, if you want you can do it.

gromakovsky: http://junior3d.ru/texture/%D0%9A%D0%BE%D0%B6%D0%B0/%D0%9C%D0%B5%D1%85/%D0%BC%D0%B5%D1%85_12.jpg

gromakovsky: We already have a data type. Functions with 100500 arguments are ugly and sooner or later this function will start having 100500 arguments and someone will have to refactor it. It's better to do it now than to postpone it. Also there is no advantage in your approach.

int-index: I agree with @gromakovsky here. Records are better than multiple arguments, because they're a cheap way to get keyword-arguments in Haskell (rather than positional arguments), which are more readable and less bug-prone.

int-index: Also I don't see the purpose of having `giveConsts` rather than using `give` directly.

pva701: @int-index you are forced to add `reflection` in your cabal if you don't have `giveConsts`

godtossing/Pos/Binary/GodTossing/Types.hs
```diff
@@ -4,35 +4,47 @@ module Pos.Binary.GodTossing.Types () where
...
-        Field [| lastKnownEoS :: EpochOrSlot                       |],
-        Field [| certs        :: VssCertificatesMap                |],
-        Field [| whenInsMap   :: HashMap StakeholderId EpochOrSlot |],
-        Field [| whenInsSet   :: Set (EpochOrSlot, StakeholderId)  |],
-        Field [| whenExpire   :: Set (EpochOrSlot, StakeholderId)  |],
-        Field [| expiredCerts :: Set (EpochOrSlot, (StakeholderId,
-                                                      EpochOrSlot,
-                                                      VssCertificate)) |]
-    ]]
+instance HasCoreConstants => Bi VssCertData where
```
int-index: Why remove the deriving?

pva701: how to add `HasCoreConstants` in TH instance without any addition efforts?

src/node/Main.hs
```diff
@@ -73,7 +73,10 @@ action args@SimpleNodeArgs {..} = do
         sscParams = bool (Left ()) (Right gtParams) (CLI.sscAlgo commonArgs == GodTossingAlgo)
     case (sscParams) of
         (Left par)  -> actionWithoutWallet @SscNistBeacon par currentParams
-        (Right par) -> actionWithoutWallet @SscGodTossing par currentParams
+        (Right par) ->
+            -- TODO: What the heck?!
+            -- @actionWithoutWallet@ requires HasCoreConstants, why?
+            giveStaticConsts (actionWithoutWallet @SscGodTossing par currentParams)
```
int-index: Better figure this out. Apparently `(SscConstraint SscGodTossing, SecurityWorkersClass SscGodTossing)` implies `HasCoreConstants`.

tools/src/analyzer/Main.hs
```diff
@@ -36,7 +37,7 @@ main = do
 
     case txFile of
         Nothing   -> pure ()
-        Just file -> analyzeVerifyTimes file confirmationParam logs
+        Just file -> giveStaticConsts $ analyzeVerifyTimes file confirmationParam logs
```
int-index: Preferably, use `give` as close to the top level as possible (in this case, `main = giveStaticConsts $ do`), because any time you do it in a nested expression, I start to worry it might be a second `give`.

tools/src/block-gen/Main.hs
```diff
@@ -76,7 +77,7 @@ main = flip catch catchEx $ do
                 True
                 bootStakeholders
     seed <- maybe randomIO pure bgoSeed
-    bracket (openNodeDBs (not bgoAppend) bgoPath) closeNodeDBs $ \db ->
+    bracket (openNodeDBs (not bgoAppend) bgoPath) closeNodeDBs $ \db -> giveStaticConsts $
```
int-index: move this `give` higher

wallet/node/Main.hs
```diff
@@ -9,40 +9,44 @@ module Main
...
 
-import           NodeOptions (WalletNodeArgs(..), WalletArgs(..), getWalletNodeOptions)
+import           NodeOptions                (WalletArgs (..), WalletNodeArgs (..),
+                                             getWalletNodeOptions)
 
 actionWithWallet :: SscParams SscGodTossing -> NodeParams -> WalletArgs -> Production ()
 actionWithWallet sscParams nodeParams wArgs@WalletArgs {..} =
     bracketWalletWebDB walletDbPath walletRebuildDb $ \db ->
-        bracketWalletWS $ \conn -> do
+        bracketWalletWS $ \conn -> giveStaticConsts $
```
int-index: Move it up!

wallet/src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -179,14 +179,14 @@ type CachedCAccModifier = CAccModifier
 -- | Evaluates `txMempoolToModifier` and provides result as a parameter
 -- to given function.
 fixingCachedAccModifier
-    :: (WalletWebMode m, MonadKeySearch key m)
+    :: (WalletWebMode m, MonadKeySearch key m, HasCoreConstants)
```
int-index: You added `HasCoreConstants` here *and* to `WalletWebMode`? Why?

pva701: I forgot to remove it here (I started to add `HasCoreConstants` and realised it would be a problem)

godtossing/Pos/Binary/GodTossing/Types.hs
int-index: Modify the TH generator to take additional constraints as input

pva701:
>Give me a good reason to have both!

how do you suggest to get constants? from a command line? or just hardcode them in the `node/Main.hs` (and similar top level files)?

int-index:
> how do you suggest to get constants? from a command line? or just hardcode them in the node/Main.hs (and similar top level files)?

What is different between `CoreConfig` and `CoreConstants`? How do you decide what data goes where?

I see two cases here:

1. we want it to be static (compile-time). In this case, read from a file safely (using TH), not with `unsafePerformIO`
2. we want it to be dynamic (run-time). In this case, read from wherever you want in `main` and then pass via `reflection`.

pva701:
`CoreConstants` - constants of run-time
`CoreConfig` - constants of compile-time

So we are going to add field corresponding to `CoreConfig` every time when some compile-time constant becomes run-time constant.


>1. we want it to be static (compile-time). In this case, read from a file safely (using TH), not with unsafePerformIO

we use `unsafePerformIO` in dev mode and Th in prod mode, AFAIK

>2. we want it to be dynamic (run-time). In this case, read from wherever you want in main and then pass via reflection.

we read from config and pass them. I don't like to hard code them. Pass run-time constants as command-line sound more or less reasonable for me. but it's not necessary now, AFAIK.

int-index:
> CoreConstants - constants of run-time
> CoreConfig - constants of compile-time

Yes, exactly, I don't like this separation. In `dev` mode all constants are run-time, and then we have tests where we might want to specify different constants in different parts of the test suite.

Just dump everything in `CoreConfig`, pass it via `reflection`, and then get it at run-time from somewhere (from the command-line, from a config-file, or from a TH-embedded file for distributed binaries).

src/node/Main.hs
pva701: seems it's false, I see error
```
    /home/pva701/github/cardano-sl/src/Pos/Launcher/Launcher.hs:35:32: error:
        • Could not deduce (Data.Reflection.Given
                              Pos.Core.Context.CoreConstants)
```
where constraints are `(SscConstraint ssc, SecurityWorkersClass ssc)`

pva701: anyway I am going to move all `give` to a top level, so it won't matter

int-index: That's what I'm saying. The constraint are `(SscConstraint ssc, SecurityWorkersClass ssc)` and `ssc ~ SscGodTossing`, so we can conclude that the compiler thinks that `HasCoreConstants` is somehow necessary to get `(SscConstraint SscGodTossing, SecurityWorkersClass SscGodTossing)`.

core/Pos/Core/Context.hs
int-index: Re-export it

src/node/Main.hs
pva701: oh, ok. got it. mda

godtossing/Pos/Binary/GodTossing/Types.hs
pva701: mmmmm, nice plan.
what about other constraints? with non empty type-parametes.
I will try now, but it will look like kostil for `HasCoreConstants`

gromakovsky:
> Just dump everything in CoreConfig, pass it via reflection, and then get it at run-time from somewhere (from the command-line, from a config-file, or from a TH-embedded file for distributed binaries).

1. I am 99% sure it will take significantly more time, but this task is very important and urgent and the current solution solves the main problem.
2. It will require adding `HasCoreConfig` constraint in more places, even where we don't need to be able to change anything in runtime.
3. Sometimes it can be convenient to know for sure that some value is 100% constant and that you don't need to search for `give` usages to know actual value.

Anyway, the main point is (1), currently the main goal is to be able to modify `k` in tests and this PR solves it.

pva701:
sounds reasonable. @gromakovsky , che skajesh?

gromakovsky:
Created https://issues.serokell.io/issue/CSL-1506 for future.

pva701:
@gromakovsky done

core/constants.yaml
```diff
@@ -167,11 +167,6 @@ qanet_tns: &qanet_tns
   protocolMagic: 50987900
   genesisBinSuffix: qa
 
-travis: &travis
```
volhovM: :cool:

pva701:
I would like to add TH instance to get rid of `HasCoreConstants` for `Bi` instances

pva701:
`Cxt` is https://hackage.haskell.org/package/template-haskell-2.12.0.0/docs/Language-Haskell-TH.html#t:Cxt

pva701:
Cxt ~ Context, AFAIU

PR #1344, opened 2017-08-15
[CSL-1485] Don't allow deleting/adding an unspent output twice

neongreen:
Caveats:

* I haven't tried actually sending transactions :confused:
* When used with MapModifier, it doesn't actually check that outputs are/aren't in the DB that the MapModifier will be applied to.


core/Pos/Util/Modifier.hs
```diff
@@ -141,9 +145,13 @@ insertionsMap = HM.mapMaybe identity . getMapModifier
...
 insertions = HM.toList . insertionsMap
 
+-- | Get all deletions from 'MapModifier' as a 'HashSet'.
+deletionsSet :: MapModifier k v -> HashSet k
+deletionsSet = getKeys . HM.filter isNothing . getMapModifier
+
 -- | Get all deletions from 'MapModifier'.
 deletions :: MapModifier k v -> [k]
-deletions = HM.keys . HM.filter isNothing . getMapModifier
+deletions = HS.toList . deletionsSet
```
gromakovsky: It's a bit less efficient, because it involves conversion to set and then to list, but ok.

txp/Pos/Txp/Toil/Trans.hs
```diff
@@ -42,8 +45,16 @@ instance MonadUtxoRead m => MonadUtxoRead (ToilT __ m) where
...
+    utxoPut id aux = ether $ do
+        utxo <- use tmUtxo
+        when (id `HM.member` MM.insertionsMap utxo) $
+            error ("utxoPut@ToilT: "+|id|+" is already in utxo")
+        tmUtxo %= MM.insert id aux
+    utxoDel id = ether $ do
+        utxo <- use tmUtxo
+        when (id `HS.member` MM.deletionsSet utxo) $
+            error ("utxoDel@ToilT: "+|id|+" is already deleted from utxo")
+        tmUtxo %= MM.delete id
```
gromakovsky: Isn't it better to use `utxoGet` to detect whether we have input in underlying storage?
In this case you can define this logic in general way and leave old functions in class.
E. g. have `utxoPutUnchecked` and `utxoDelUnchecked` inside the class and `utxoPut` and `utxoDel` outside.

gromakovsky: @neongreen good evening!

georgeee:
Not required for mainnet

This PR should introduce only sanity checks for transaction processing

Which are desirable, but not perfectly necessary

txp/Pos/Txp/Toil/Trans.hs
neongreen: :+1:

neongreen:
auxx doesn't work reliably, but I saw in logs that a transaction got sent successfully (two times).

PR #1345, opened 2017-08-15
Use the core-relay setup in local benchmark script

kantp:
This allows to run local benchmarks involving core, relay, and edge nodes.


core/constants.yaml
```diff
@@ -176,7 +176,7 @@ benchmark: &benchmark
   <<: *testnet
 
   genesisN: 50000
-  k: 6
+  k: 2
```
avieth: Is this an important change? Just curious.

kantp: No, it's not important.

The reason I made that I want to be more sensitive to bugs that happen at the transition from one epoch to the next.

scripts/bench/topology/topology0.yaml
```diff
@@ -2,18 +2,24 @@ nodes:
...
     addr: 127.0.0.1
     port: 3001
   node2:
     type: core
     region: undefined
-    static-routes: [[node0], [node1]]
+    static-routes: [[node0], [node1], [node4]]
     addr: 127.0.0.1
     port: 3002
+  node4:
```
avieth: No `node3`? Was this intentional? Not a big deal in any case.

scripts/launch/demo.sh
```diff
@@ -97,15 +102,13 @@ while [[ $i -lt $panesCnt ]]; do
...
-  pane="${window}.$i"
-
   if [[ $i -lt $n ]]; then
     tmux send-keys "$(node_cmd $i "$stats" "$stake_distr" "$wallet_args" "$system_start" "$config_dir" $exec_name) --no-ntp" C-m
   else
     # Number of transactions to send per-thread: 300
     # Concurrency (number of threads sending transactions); $CONC
     # Delay between sends on each thread: 500 milliseconds
-    tmux send-keys -t ${pane} "sleep 40s && $(bench_cmd $i "$stake_distr" "$system_start" 300 $CONC 500 neighbours)" C-m
+    tmux send-keys "sleep 40s && $(bench_cmd $i "$stake_distr" "$system_start" 3000 $CONC 500 neighbours)" C-m
```
avieth: Should make the number of transactions per thread (3000) come from a script parameter or environment variable.

scripts/bench/topology/topology0.yaml
kantp: node3 is an edge node that will subscribe to the relay, so it's not mentioned in the topology file.

scripts/launch/demo.sh
kantp: I agree, we should do that.

PR #1348, opened 2017-08-15
Delete run-smart-generator.sh script

gromakovsky:
We don't have smart-generator, so I suppose this script is obsolete.


volhovM:
Makes sense.

PR #1349, opened 2017-08-15
[CSL-1502] Change genesis stakeholder weight from Word16 to Word32

gromakovsky:
It is needed to be able to use rich-poor distribution with 50000 poor nodes.

P. S. I haven't tested it yet, going to do it after it builds.

### DIFF
--- a/core/Pos/Arbitrary/Core.hs
+++ b/core/Pos/Arbitrary/Core.hs
@@ -443,7 +443,8 @@ instance Arbitrary G.GenesisCoreData where
                 , G.CustomStakes <$> vector innerLen
                 ]
         stakeDistrs <- vectorOf outerLen distributionGen
-        hashmapOfHolders <- arbitrary :: Gen (HashMap Types.StakeholderId Word16)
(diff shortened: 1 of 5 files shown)


gromakovsky:
I tested it and it doesn't work well :(
Node consume 100% CPU and memory usage constantly grows until I stop nodes.

avieth:
> I tested it and it doesn't work well :(
> Node consume 100% CPU and memory usage constantly grows until I stop nodes.

Interesting. How could this patch possibly cause such an issue? FWIW I did basically the same patch locally, but with `Word64` instead of `Word32`, but I wasn't able to use it in a real-world setting because the boot dust threshold was too high.

gromakovsky:
@avieth 

> Interesting. How could this patch possibly cause such an issue? 

Probably that's because `genesisSplitBoot` generates too big distributions which results into quadratic size utxo which is too big for 50000 nodes.

> I wasn't able to use it in a real-world setting because the boot dust threshold was too high.

dust threshold is deleted in #1338 and my branch is based on that branch.

PR #1351, opened 2017-08-16
[CSL-1360] [CSL-1496] [CSL-1497] [CSL-1510] [CSL-1525] Tests for apply/rollback

int-index:
Utilities added:

* [x] Display block scenario context in case of failure
* [x] Marker for broken tests

Tests added:

* [x] Fork McForkyface (short, medium, deep)
* [x] Apply blocks across epochs (genesis-only and with main blocks)

Bug fixes:

* [x] The security check in `slogRollbackBlocks` was using an incorrect ordering (NewestFirst instead of OldestFirst).
* [x] The security check in `slogRollbackBlocks` made LRC broken (now it's disabled in LRC but enabled everywhere else)
* [x] Fix the treatment of slots in BlockScenario interpreter


test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -265,24 +271,130 @@ blockPropertyScenarioGen m = do
...
+    dj <- getRandomR (1, d*2)
+    -- now we can generate paths:
+    --
+    -- B0 - B1 - B2 - B3 - B4 - B5 - B6 - B7
+    --              \
+    --                C3 - C4 - C5 - C6
+    --
+    -- in this example, q=7, d=5, i=2, dj=4
+    let
+        nonEmptyCuz r [] = error ("Requirement failed: " <> r)
```
rockbmb: `nonEmptyCuz`, what does `Cuz` mean here?

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -265,24 +271,130 @@ blockPropertyScenarioGen m = do
...
+    let expectSuccess = d <= k
+    -- original blockchain max index q<(9.5*k)
+    q <- getRandomR (d+1, 9 * k + div k 2)
+    let
+        -- max index of the common prefix. i>0 because d<q
+        i = q-d
+    -- fork blockchain max index j>i. the upper bound is arbitrary.
+    -- dj=j-i
+    dj <- getRandomR (1, d*2)
+    -- now we can generate paths:
```
rockbmb: Nice description!

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -265,24 +271,130 @@ blockPropertyScenarioGen m = do
...
+genSingleFork :: forall g m. (RandomGen g, Monad m) => ForkDepth -> BlockEventGenT g m ()
+genSingleFork fd = do
+    let k = fromIntegral blkSecurityParam :: Int
+    -- 'd' is how deeply in the chain the fork starts. In other words, it's how many
+    -- blocks we're going to rollback (therefore must be >1).
+    d <- getRandomR $ case fd of
+        ForkShort  -> (1, if k > 1 then k-1 else 1)
+        ForkMedium -> (if k > 2 then k - 2 else 1, k+2)
+        ForkDeep   -> (k+1, div (k*3) 2 + 1)
+    -- the depth must be <=k for a successful rollback.
```
int-index: it must be `<=k` for a *successful* rollback, we're testing failing ones as well.

test/Test/Pos/Block/Logic/VarSpec.hs
int-index: Ah, this is actually horrible, it means "because"

test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -265,24 +271,130 @@ blockPropertyScenarioGen m = do
...
             " relative to the " <> show snapId <> " snapshot:\n" <>
-            show dbDiff
+            show dbDiff <> "\n" <>
+            pretty bsec
+
+----------------------------------------------------------------------------
+-- Multi-epoch
+----------------------------------------------------------------------------
+
+-- Input: the amount of blocks after crossing.
```
int-index: crossing the epoch boundary

PR #1353, opened 2017-08-16
[DEVOPS-265] Don't build explorer on macOS

domenkozar:
We have no macOS developers in explorer team (at the moment).

cc @ksaric @akegalj @sectore please confirm if so :)


ksaric:
@domenkozar I confirm.

PR #1355, opened 2017-08-16
[CSL-1508] retry joining kademlia indefinitely

avieth:
For P2P nodes. The node will not try to do any business logic until it
joins the network, so a misconfiguration of kademlia will result in
repeated log warnings about failure to join.

### DIFF
--- a/infra/Pos/DHT/Real/Real.hs
+++ b/infra/Pos/DHT/Real/Real.hs
@@ -3,6 +3,8 @@
 
 module Pos.DHT.Real.Real
        ( kademliaJoinNetwork
+       , kademliaJoinNetworkNoThrow
+       , kademliaJoinNetworkRetry
        , K.lookupNode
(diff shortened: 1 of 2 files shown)


avieth:
I noticed that loading SSC state takes quite a while. The kademlia and network-transport servers are up while that's happening. Is this / could this be a problem?

avieth:
Test failure in one of the travis builds

```
Aug 16 17:41:10   1) Test.Pos.Lrc.Worker.Lrc.Worker.lrcSingleShotNoLock Computes richmen correctly according to the stake distribution right before the '8 * k'-th slot.
```

gromakovsky:
@avieth it's https://issues.serokell.io/issue/CSL-1501, I restarted both failed builds (one was timeouted).

volhovM:
I confirm this fixes demo in master. Yay.

PR #1357, opened 2017-08-16
[CSL-1502] Match on stake distr in generateWStakeholders

volhovM:
Now boot stakeholders generation (for dev/tests/bench) matches on real distribution. Which in turn gives possibility to set `genesisStakeholders` to rich in `richPoor` distribution. Before this, implementation gave stake to everybody thus producing huge stakeholders weights and `bootDustThd` which led to a bunch of problems.
I also removed support for bitcoin distribution as it's not used by anyone and can be emulated by `CustomStakes`.

### DIFF
--- a/lwallet/Main.hs
+++ b/lwallet/Main.hs
@@ -426,7 +426,6 @@ main = giveStaticConsts $ do
     let devStakeDistr =
             devStakesDistr
                 (CLI.flatDistr woCommonArgs)
-                (CLI.bitcoinDistr woCommonArgs)
                 (CLI.richPoorDistr woCommonArgs)
                 (CLI.expDistr woCommonArgs)
     let wpGenesisContext =
--- a/lwallet/Main.hs
+++ b/lwallet/Main.hs
@@ -62,12 +62,11 @@ import           Pos.Crypto                       (Hash, SecretKey, SignTag (Sig
                                                    safeToPublic, toPublic, unsafeHash,
                                                    withSafeSigner)
 import           Pos.Data.Attributes              (mkAttributes)
-import           Pos.Genesis                      (GenesisContext (..),
-                                                   StakeDistribution (..), devAddrDistr,
-                                                   devStakesDistr,
+import           Pos.Genesis                      (StakeDistribution (..), devAddrDistr,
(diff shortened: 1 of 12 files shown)


src/Pos/Genesis.hs
```diff
@@ -173,39 +136,65 @@ genesisUtxo gws@(GenesisWStakeholders bootStakeholders) ad
...
+        error $ "genesisContextImplicit: distributions have " <>
+                "common keys which is forbidden " <>
+                pretty a <> ", " <> pretty b
+    genStakeholders =
+        GenesisWStakeholders $
+        foldr1 mergeStakeholders $
+        map (getGenesisWStakeholders . generateWStakeholders) addrDistr
     utxo = genesisUtxo genStakeholders addrDistr
 
--- | Generate weighted stakeholders using passed address distribution.
```
gromakovsky: Why did you delete this line?

kantp:
I just tried to merge this on the current master, and run the benchmarks. The good news is, using the `rich-poor-distribution` no longer causes the nodes to crash on startup. However, the transaction generator is not able to send transactions. I get the error
```Error: Not enough money to send (need 1 coin(s) coins more) while trying to send to [NodeId 127.0.0.1:3004:0]```
indicating that there is not enough

**Comment by volhovM** [comment]
@kantp i was testing richpoor with demo dumping utxo and i manually confirmed that poor users have their coins.

File: lwallet/Main.hs [hunk 2]
```diff
@@ -185,28 +183,21 @@ runCmd sendActions (SendToAllGenesis duration conc delay_ sendMode tpsSentFile)
...
-                    balance `subCoin` mkCoin 1
-                txOut2 = TxOut {
-                    txOutAddress = makePubKeyAddress (toPublic key),
-                    txOutValue = val2
-                    }
                 stakeholderId = addressHash (toPublic key)
                 toDistr val = [(stakeholderId, val)]
-                txOuts = TxOutAux txOut1 (toDistr val1)
-                    :| [TxOutAux txOut2 (toDistr val2)]
+                txOuts = TxOutAux txOut1 (toDistr val1) :| []
```
gromakovsky: What happened here?

PR #1358, opened 2017-08-16
demo.sh: start system 30 seconds in the future

kantp:
This prevents nodes from trying to communicate with each other before
all the nodes have started up.

### DIFF
--- a/scripts/launch/demo.sh
+++ b/scripts/launch/demo.sh
@@ -61,7 +61,7 @@ fi
 # and start processing the first slot.
 if [ -z "$system_start" ]
   then
-    system_start=$((`date +%s` + 1))
+    system_start=$((`date +%s` + 30))
 fi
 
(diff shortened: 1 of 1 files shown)


gromakovsky:
Is it really necessary to wait for so long? What takes so much time? Perhaps there is a bigger problem if initialization takes so much time.
Isn't e. g. 10 seconds enough?
I guess it will be quite annoying to wait for 30 seconds when you want to launch demo.

kantp:
@gromakovsky The 30 seconds are needed. For a five node setup and 30 seconds delay, the last node only has to work 2 seconds for the system start.

We might want to investigate why the startup time is this high separately, but I think at the moment, the 30 seconds are needed.

kantp:
In fact, I just had a run where 30 seconds was apparently not enough.

PR #1359, opened 2017-08-16
[CSL-1509] Move 'cardano-sl' into its subfolder

rockbmb:


domenkozar:
@rockbmb why was this closed? :)

PR #1360, opened 2017-08-16
[CSL-1509] Move 'cardano-sl' into its subfolder

rockbmb:


cardano-sl/cardano-sl.cabal
```diff
@@ -3,13 +3,13 @@ version:             0.5.1
 synopsis:            Cardano SL main implementation
 description:         Please see README.md
 license:             MIT
-license-file:        LICENSE
+license-file:        ../LICENSE
```
domenkozar: You'll need to move files to subfolder as well and then symlink them at the top-level, otherwise Nix won't work properlly

stack.yaml
```diff
@@ -7,7 +7,7 @@ flags:
 extra-package-dbs: []
 
 packages:
-- '.'
+- cardano-sl
```
domenkozar: Afaik @georgeee proposed `node` as a name

rockbmb: I used another name because at first I had some issues that I incorrectly thought were related to it being `node`, this was a placeholder.

rockbmb:
@domenkozar Working on it, should be done today, still.

rockbmb:
@gromakovsky Oh, you're right. Will fix.

PR #1365, opened 2017-08-17
[CSM-256, CSM-427] Resubmit txs

Martoon-00:
This adds to wallet a possibility to resubmit failed transactions.

Briefly what's been done:
* Info about _pending transaction_ is stored in db
* Wallet Blistener detects when pending transaction appears in (or vanishes from) blockchain
* Special resubmission worker tracks for transactions not in blockchain and resubmits them with exponential backoff; it also finds transactions which are deep enough to be declared persistent.
* Added info about transaction state to `CTx` (type for transaction info returned to frontend)
* Wallet endpoints which make transactions (`newPayment` and `redeem`) now report transaction creation as successful if there is a chance to eventually submit it.
* In transaction submission peers' responses are tracked. Success for any peer may lead transaction to be considered created even if local application failed. Except that, on peer ack resubmission delay is increased dramatically.

Tested:
* Normal work
* When tx submission artificially throws reclaimable error
* Networking errors were all suppressed, submitting to unavailable nodes ended ok or hung everything :+1: 
* DAT passed, except two failures which happen is `master` as well


wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,174 @@
...
+  where
+    tryLater = pass
+    discard  = do
+        casPtxCondition ptxTxId PtxApplying (PtxWon'tApply $ sformat build e)
+        logInfo $ sformat ("Transaction "%build%" was canceled") ptxTxId
+
+filterApplicablePtxs
+    :: MonadPendings m
+    => SlotId -> [PendingTx] -> m [PendingTx]
+filterApplicablePtxs curSlot ptxs = do
```
Martoon-00: Am I right that I have to take block semaphore here? Otherwise I can catch a moment when some conflicting transaction is removed from mempool but not added to blocks yet.
If I have to, won't it be too expensive?

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,174 @@
...
+-- if needed.
+processPtxs
+    :: MonadPendings m
+    => SendActions m -> SlotId -> [PendingTx] -> m ()
+processPtxs sendActions curSlot ptxs = do
+    mapM_ (processPtxInUpperBlocks curSlot) ptxs
+    toResubmit <- filterApplicablePtxs curSlot $
+                  filter ((PtxApplying ==) . ptxCond) ptxs
+    logInfo $ sformat fmt (map ptxTxId toResubmit)
+    resubmitPtxsDuringSlot sendActions ptxs
```
flyingleafe: I bet you wanted to resubmit only `toResubmit` here, not all `ptxs`

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,174 @@
...
+            ptxs
+
+    processPtxs sendActions curSlot selectedPtxs
+  where
+    wHash PendingTx{..} = WithHash (taTx ptxTxAux) ptxTxId
+    evalPtxsPerSlotLimit = do
+        slotDuration <- getLastKnownSlotDuration
+        return $ fromIntegral $
+            convertUnit slotDuration `div` ptxsResubmitionPeriod
+    -- FIXME [CSM-256]: move to constants?
```
flyingleafe: Yes, please

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
flyingleafe: TL;DR: simply get rid of this function.

First, this code seems simply incorrect, because you provide `mempty` as `UtxoModifier` to `runToilTLocal` down there, which means that `processTx` won't consider valid txOuts from mempool. Use `txProcessTransaction` instead.

Second - I don't believe that we really need to use block semaphore here. It seems like in theory such a conflict can appear, but it applies to basic transaction sending too, and we don't use block semaphore when we send a transaction, and we're doing it just fine so far.

Third, I actually think that you might not need to apply transactions to mempool before resubmitting it at all, because `submitAndSave` applies them anyway, and I don't believe that resending non-applicable transaction to your neighbors _once_ (because after failed resubmittion it would be deleted from the pool) is too much overhead. Moreover, you'll receive `ToilKnown` every time you try to `submitAndSave` a transaction which is already in mempool, and then (for some reason) try to resubmit it later (judging by the code above). It doesn't make sense to me.

wallet/src/Pos/Wallet/Web/Pending.hs
```diff
@@ -0,0 +1,11 @@
+-- | Pending transactions.
+
+module Pos.Wallet.Web.Pending
+    ( module Pos.Wallet.Web.Pending.Types
+    , module Pos.Wallet.Web.Pending.Worker
+    , module Pos.Wallet.Web.Pending.Util
+    ) where
```
Martoon-00: These 3 files are splitted among subprojects :/

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -358,13 +364,16 @@ trackingApplyTxs (getEncInfo -> encInfo) allAddresses getDiff getTs txs =
 
             usedAddrs = map cwamId ownOutAddrMetas
             changeAddrs = evalChange allAddresses (map cwamId ownInpAddrMetas) usedAddrs
+
+            mSlotId = getPtxBlkInfo blkHeader
```
Martoon-00: Yep, thanks. Same with comment above

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
Martoon-00: Right, mad bug :unamused:

core/constants.yaml
```diff
@@ -54,6 +54,9 @@ dev: &dev
   # Delegation
   lightDlgConfirmationTimeout: 500
 
+  # Transaction resubmition
+  pendingTxResubmissionPeriod: 10 # seconds
+  pendingTxAttemptsLimit: 5
```
flyingleafe: As I said in PM, we decided to get rid of this limit. Reason: if the transaction has managed to escape to outer network, we cannot longer predict what is going to happen in its lifetime - probably, after 5 attempts, some node is going to include it in block, and its "expired" status is going to be invalid. True tx expiration is possible only if we include expiry slot into core structure of transaction, but we're not going to do it for now.

node/src/Pos/Wallet/Web/Pending/Types.hs
```diff
@@ -0,0 +1,62 @@
...
+-- | Required information about block where given pending transaction is sited
+type PtxBlockInfo = SlotId
+
+-- | Current state of pending transaction.
+--
+-- Once transaction is created, it should be assigned 'PtxApplying' condition
+-- in order to be tracked by resubmitter.
+--
+-- When wallet tracker notices a block with given transaction, it switches
+-- condition to 'PtxInUpperBlocks' providing needed information about that block.
```
pva701: why `Upper`? I couldn't get it, got eventually. IHMO it should be named `Newest` AFAIK

also `PtxApplying` doesn't reflect it's not in the blockchain. but it's subjective. I don't mind to use this name.

node/src/Pos/Wallet/Web/Pending/Types.hs
```diff
@@ -0,0 +1,62 @@
...
+-- Transactions in this state are periodically checked by resubmitter for
+-- the possibility (and need) to submit them again.
+--
+-- Resubmitter also checks whether transaction is deep enough in blockchain to
+-- be moved to 'PtxPersisted' state.
+--
+-- If transaction is ever noticed to conflict with transactions existing in
```
pva701: when can it happen? is this status for just in case?

pva701: ok, I got it. it's needed when utxo state is changed

node/src/Pos/Wallet/Web/Pending/Types.hs
```diff
@@ -0,0 +1,62 @@
...
+    = PtxApplying                    -- ^ Is waiting to be applyed
+    | PtxInUpperBlocks PtxBlockInfo  -- ^ Recently appeared in block.
+    | PtxPersisted                   -- ^ Transaction is ~guaranteed to remain in blockchain
+                                     --   (with up to *high* assurance level)
+    | PtxWontApply Text              -- ^ Can't be applyed and requires user's
+                                     --   input to reform tx
+    deriving (Eq, Show)
+
+-- | All info kept about pending transaction
+data PendingTx = PendingTx
```
pva701: not strict fields :thinking:

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,167 @@
...
+    mdepth <- getWalletAssuredDepth _ptxWallet
+    if | PtxInUpperBlocks slotId <- _ptxCond,
+         Just depth <- mdepth,
+         longAgo depth slotId -> do
+             void $ casPtxCondition _ptxWallet _ptxTxId _ptxCond PtxPersisted
+             logInfo $ sformat ("Transaction "%build%" got persistent") _ptxTxId
+       | otherwise -> pass
+  where
+     longAgo depth (flattenSlotId -> ptxSlot) =
+         ptxSlot + getBlockCount depth < flattenSlotId curSlot
```
pva701: it's wrong to measure a difference in the slots, not in the blocks, AFAIU. you have assurance guarantee in blocks, not slots.

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,167 @@
...
+        let newCond = PtxWontApply (sformat build e)
+        void $ casPtxCondition _ptxWallet _ptxTxId PtxApplying newCond
+        logInfo $ sformat ("Pending transaction "%build%" was canceled")
+                  _ptxTxId
+
+processPtxInUpperBlocks :: MonadPendings m => SlotId -> PendingTx -> m ()
+processPtxInUpperBlocks curSlot PendingTx{..} = do
+    mdepth <- getWalletAssuredDepth _ptxWallet
+    if | PtxInUpperBlocks slotId <- _ptxCond,
+         Just depth <- mdepth,
```
pva701: btw what is `mdepth`?

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,167 @@
...
+    interval <- evalSubmitDelay (length ptxs)
+    forM_ ptxs $ \ptx -> do
+        delay interval
+        fork $ resubmitTx sendActions ptx
+  where
+    submitionEta = 5 :: Second
+    evalSubmitDelay toResubmitNum = do
+        slotDuration <- getLastKnownSlotDuration
+        let checkPeriod = max 0 $ slotDuration - convertUnit submitionEta
+        return (checkPeriod `div` fromIntegral toResubmitNum)
```
pva701: taks, what about case when `checkPeriod` < `toResubmitNum`

pva701: should we take minimum `min toResubmitNum checkPeriod`

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,167 @@
...
+processPtxsToResubmit sendActions ptxs = do
+    ptxsPerSlotLimit <- evalPtxsPerSlotLimit
+    let toResubmit =
+            take ptxsPerSlotLimit $
+            filter ((PtxApplying ==) . _ptxCond)
+            ptxs
+    logInfo $ sformat fmt (map _ptxTxId toResubmit)
+    resubmitPtxsDuringSlot sendActions toResubmit
+  where
+    fmt = "Transactions to resubmit on current slot: "%listJson
```
pva701: I don't get this logic at all. are you going to resubmit one tx in `pendingTxResubmitionPeriod`? as I see `processPtxsToResubmit` doesn't follow this strange rule

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,167 @@
...
+                  _ptxTxId
+
+processPtxInUpperBlocks :: MonadPendings m => SlotId -> PendingTx -> m ()
+processPtxInUpperBlocks curSlot PendingTx{..} = do
+    mdepth <- getWalletAssuredDepth _ptxWallet
+    if | PtxInUpperBlocks slotId <- _ptxCond,
+         Just depth <- mdepth,
+         longAgo depth slotId -> do
+             void $ casPtxCondition _ptxWallet _ptxTxId _ptxCond PtxPersisted
+             logInfo $ sformat ("Transaction "%build%" got persistent") _ptxTxId
```
pva701: are we going to store all persistent txs for the whole period of life of a node?

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,167 @@
...
+    :: MonadPendings m
+    => SendActions m -> SlotId -> [PendingTx] -> m ()
+processPtxs sendActions curSlot ptxs = do
+    mapM_ (processPtxInUpperBlocks curSlot) ptxs
+    processPtxsToResubmit sendActions ptxs
+
+processPtxsOnSlot
+    :: MonadPendings m
+    => SendActions m -> SlotId -> m ()
+processPtxsOnSlot sendActions curSlot = do
```
pva701: tbh I don't like we store all transactions which we couldn't resubmit. I would like to set some TTL in slots. if `curSlot - ptxSlot > this ttl` we just drop this tx

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -0,0 +1,167 @@
...
+    wHash PendingTx{..} = WithHash (taTx _ptxTxAux) _ptxTxId
+
+-- | On each slot this takes several pending transactions and resubmits them if
+-- needed and possible.
+startPendingTxsResubmitter
+    :: MonadPendings m
+    => SendActions m -> m ()
+startPendingTxsResubmitter sa =
+    void . fork . setLogger $
+    onNewSlot False (processPtxsOnSlot sa)
```
pva701: hm, I supposed `onNewSlot` start new thread by himself

wallet/src/Pos/Wallet/Web/Methods/Txp.hs
```diff
@@ -0,0 +1,74 @@
...
+    => NonEmpty (CId Addr, Coin)
+    -> m (NonEmpty TxOutAux)
+coinDistrToOutputs distr = do
+    addrs <- mapM decodeCTypeOrFail cAddrs
+    pure $ NE.zipWith mkTxOut addrs coins
+  where
+    (cAddrs, coins) = NE.unzip distr
+    mkTxOut addr coin = TxOutAux (TxOut addr coin) []
+
+-- | Like 'submitAndSaveTx', but suppresses errors which can gone
```
Martoon-00: gramota

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
Martoon-00: In theory, db can have no information about this pending tx, so `Maybe BlockCount` is here

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
Martoon-00: If `checkPeriod` < `toResubmitNum`, then I have to send everything immediately, that's ok.
`min toResubmitNum checkPeriod` doesn't make sense because one is measured in seconds, while another has no measure

Martoon-00: Maybe it is worth converting everything to `Microseconds`, then `checkPeriod` will be ~always greater `toResubmitNum`

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
Martoon-00: Something like that. I wanted to limit average pending txs submission rate here, but in `processPtxsToResubmit` actually submit using that special timing rule.
Rename `pendingTxResubmitionPeriod` somehow?

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
Martoon-00: This will be reflected on information which transaction history returns.
We store whole transaction history anyway, which is not smaller

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
Martoon-00: Same like with canceled transactions, I would likely cleanup them if we didn't display `PtxCondition` in frontend

node/src/Pos/Wallet/Web/Pending/Types.hs
Martoon-00: Which name would you suggest instead of `PtxApplying`?

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
Martoon-00: Wallet stuff starter blocks if I remove this `fork`

node/src/Pos/Communication/Tx.hs
```diff
@@ -47,28 +50,29 @@ type TxMode ssc ctx m
...
-submitAndSave
+submitAndSaveTx
     :: TxMode ssc ctx m
     => EnqueueMsg m -> TxAux -> m TxAux
-submitAndSave enqueue txAux@TxAux {..} = do
+submitAndSaveTx enqueue txAux@TxAux {..} = do
     let txId = hash taTx
+    whenM (liftIO $ (> 0) <$> randomRIO (0 :: Int, 6)) $ do
+        logInfo "Failing sending"
+        throwM ToilKnown
```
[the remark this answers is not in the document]
Martoon-00: :man_facepalming:

node/src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -375,6 +435,11 @@ deriveSafeCopySimple 0 'base ''CTxMeta
 deriveSafeCopySimple 0 'base ''CUpdateInfo
 deriveSafeCopySimple 0 'base ''AddressLookupMode
 deriveSafeCopySimple 0 'base ''CustomAddressType
+deriveSafeCopySimple 0 'base ''TxAux
```
gromakovsky: Why is it here and not in `Pos.SafeCopy`? I mean `TxAux`, I dunno about other types, I guess they should be here, but for `TxAux` it seems more reasonable to put it into `Pos.SafeCopy`. Or move `Pos.SafeCopy` somewhere into wallet namespace.

Martoon-00: Thanks, I'll fix this along with changes requested by @pva701

PR #1368, opened 2017-08-18
[CSM-403] [CSM-408] Backup to JSON only one wallet

Martoon-00:
CSM-403: changed Import/ExportBackupJSON endpoints so that they work with one wallet.
Tested with deleting wallet and importing it back, balances and transaction history remain.

CSM-408: removed useless endpoint


Martoon-00:
Still need to update client-api :/

darko-mijic:
Is this ready to be merged?

PR #1373, opened 2017-08-18
[CSL-1328] Use GITREV env variable to specify commit of build

georgeee:


default.nix
```diff
@@ -4,6 +4,7 @@ in
 { system ? builtins.currentSystem
 , config ? {}
 , dconfig ? "testnet_staging"
+, gitrev ? "unknown"
```
domenkozar: This needs to be a bit different for Hydra, let me find docs.

cleverca22: https://github.com/NixOS/hydra/blob/7e9b6acbce9e1faf25c1f43e441490a5167174af/release.nix#L1

https://hydra.iohk.io/jobset/serokell/cardano-sl#tabs-configuration

an attrset called `cardano` will be passed into `release.nix`, containing `outPath`, `revCount`, and `rev`, `release.nix` then needs to pass that to `default.nix`

georgeee: @cleverca22 @domenkozar can I ask you guys to commit this change to branch? In principle I only need `GITREV` env variable appropriately set

domenkozar: @cleverca22 fixed it

PR #1376, opened 2017-08-19
[CSL-1461] Redeem transactions don't need fees

int-index:


txp/Pos/Txp/Toil/Logic.hs
```diff
@@ -161,9 +162,16 @@ verifyAndApplyTx curEpoch verifyVersions tx@(_, txAux) = do
   where
     ctx = Utxo.VTxContext verifyVersions
 
+isRedeemTx :: MonadUtxoRead m => TxAux -> m Bool
+isRedeemTx txAux = do
+    resolvedOuts <- traverse utxoGet $ (view txInputs . taTx) txAux
+    let inputAddresses = fmap (txOutAddress . toaOut) . catMaybes . toList $ resolvedOuts
+    return $ any isRedeemAddress inputAddresses
```
georgeee: I would set it more restrictive, in particular replace `any` with `all`

PR #1380, opened 2017-08-20
[CSL-1509] Update scripts

rockbmb:


gromakovsky:
One of the builds already failed with `curl: (22) The requested URL returned error: 502 Bad Gateway`. I see no reason in waiting for CI to pass, because these scripts are not used in CI.
Merging.

PR #1383, opened 2017-08-21
[CSL-1497] Fix security check in rollback and disable it in LRC

int-index:


node/src/Pos/Lrc/Worker.hs
```diff
@@ -197,7 +197,7 @@ lrcDo epoch consumers tip = tip <$ do
     whileAfterCrucial b = getEpochOrSlot b > crucial
     crucial = EpochOrSlot $ Right $ crucialSlot epoch
     withBlocksRolledBack blunds =
-        bracket_ (rollbackBlocksUnsafe blunds)
+        bracket_ (rollbackBlocksUnsafe (BypassSecurityCheck True) blunds)
```
pva701: it would be nice if you wrote some clarify comment why `True` here

int-index: Because LRC can rollback more than `k` blocks :man_shrugging: 

I don't know why it can do it, so my comment would be hardly of any help. If someone more knowledgeable added a comment, it'd be great.

pva701: The reason is: we might call LRC rollback during `applyWithRollback`, which himself might rollback blocks and both these rollbacks (LRC and `applyWithRollback`) are summarized and give too deeper rollback.

rollback in the `applyWithRollback` might not be deeper than `k`, but LRC rollback we create artificially. To avoid deep artificial rollback we ignore invariant in the LRC.

pva701:
Issue's description suggests to resolve this issue in another way, tbh I like your way more than in the issue's description. but gromak might diss you.

node/src/Pos/Lrc/Worker.hs
gromakovsky: @pva701 you are overcomplicating. LRC almost always rolls back more than `k` blocks. The only exception is when chain quality is exactly 0.5 (in this case it rolls back exactly `k` blocks).

PR #1387, opened 2017-08-21
[CSL-1510] Make tests deterministic

int-index:


core/Pos/Arbitrary/Core.hs
```diff
@@ -468,7 +467,7 @@ instance Arbitrary G.GenesisCoreData where
         let chop _ [] = []
             chop n l = taken : chop n dropped
               where (taken, dropped) = splitAt n l
-        allAddrs <- fmap makePubKeyAddress <$> nonrepeating (outerLen * innerLen)
+        allAddrs <- fmap makePubKeyAddress <$> vector (outerLen * innerLen)
```
volhovM: Why is this safe? `GenesisCoreData` shouldn't have repeating addresses inside b/c of `mkGenesisCoreData` restrictions.

core/Pos/Arbitrary/Crypto.hs
```diff
@@ -62,61 +49,33 @@ instance Arbitrary SignTag where
...
--- Repeat the same for ADA redemption keys
-redemptionKeys :: [(RedeemPublicKey, RedeemSecretKey)]
-redemptionKeys = unsafeMakePool "[generating redemption keys for tests..]" 50 redeemKeyGen
+    arbitrary = snd . deterministicKeyGen <$> arbitrarySizedS 32
 
 instance Arbitrary RedeemPublicKey where
-    arbitrary = fst <$> elements redemptionKeys
+    arbitrary = fst . fromMaybe err . redeemDeterministicKeyGen <$>
+                arbitrarySizedS 32
+      where err = error "arbitrary@RedeemPublicKey: keygen failed"
```
volhovM: Shouldn't you `throw` instead?

core/Pos/Util/Util.hs
```diff
@@ -189,6 +191,18 @@ instance ToJSON Byte where
 instance IsString s => MonadFail (Either s) where
     fail = Left . fromString
 
+instance Rand.MonadRandom m => Rand.MonadRandom (ReaderT r m) where
```
volhovM: Nice. Though i think there was a similar instance in `test/Test/Pos/Block/...` somewhere. Can you check it?

node/test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -84,7 +83,7 @@ genTestParams :: Gen TestParams
 genTestParams = do
     let _tpStartTime = 0
     let stakeholdersNum = 4 * groupsNumber
-    secretKeys <- nonrepeating stakeholdersNum
+    secretKeys <- vector stakeholdersNum
```
volhovM: Why duplicates here are valid?

neongreen:
@volhovM 

> I do not undrestand why replacing nonrepeating with vector works in almost all the places you've used it.

Because previously we had a pool of only 50 keys, and so if you tried to do `replicateM 100 arbitrary` you would get a lot of duplicates. Now calls to `arbitrary` generate a completely new key every time – this applies to `PublicKey`, `SecretKey`, `RedeemPublicKey`, `RedeemSecretKey`, `VssKeyPair`, `VssPublicKey` and so on.

neongreen:
Let's go through all uses of `nonrepeating` that I deleted:

> ```haskell
> allAddrs <- fmap makePubKeyAddress <$> nonrepeating (outerLen * innerLen)
> ```
  
Here `makePubKeyAddress` takes a `PublicKey` so it'll always be unique.

> ```haskell
> secretKeys <- nonrepeating stakeholdersNum
> ```

Here `mkAllSecretsSimple` takes a list of `SecretKey`s, which are always unique.

> ```haskell
> keys           <- generate $ nonrepeating n
> vssKeys        <- generate $ nonrepeating n
> ```

Ditto.

> ```haskell
> let sks@[sk1, sk2, sk3,  sk4] = runGen $ nonrepeating 4
> ```

Ditto.

core/Pos/Arbitrary/Crypto.hs
```diff
@@ -62,61 +49,33 @@ instance Arbitrary SignTag where
...
--- 'arbitrary' and then use 'Pos.Crypto.toPublic' to get the corresponding
--- public key.
-
-keys :: [(PublicKey, SecretKey)]
-keys = unsafeMakePool "[generating keys for tests...]" 50 keyGen
-{-# NOINLINE keys #-}
-
 instance Arbitrary PublicKey where
-    arbitrary = fst <$> elements keys
+    arbitrary = fst . deterministicKeyGen <$> arbitrarySizedS 32
```
flyingleafe: I'm afraid that there was a reason for using Secure Random™ here. Though if @neongreen approved, it must be alright.

neongreen: @flyingleafe it's actually my commit, @int-index just cherry-picked it :trollface:

neongreen: The reason for using Secure Random™ was that it seemed better to test *exactly* the functions that are used for generating keys (e.g. what if `keyGen` puts additional constraints on the seed and just doing `arbitrarySizedS 32` isn't enough?). I still think that it's better, but having deterministic tests is better still.

core/Pos/Arbitrary/Crypto.hs
neongreen: Why? Afaik there's no difference between `error` and `throw` in the `Gen` monad.

core/Pos/Util/Util.hs
neongreen: I checked and it doesn't look like there's one.

neongreen:
I tried running tests locally – they finish in around 24 minutes on my machine, but no single test takes more than three (or five?) minutes. Yet, Travis says “No output has been received in the last 10m0s” and kills tests.

My guesses:

1. Some particular test (e.g. the dreaded “stakeholder with low stake will be chosen seldom”) takes more than 10m on Travis?
2. Tests used to take less than 10 minutes and something is wrong with the way Travis handles terminal output so it only sees the output when the tests finish?
3. I'm just unlucky and Travis reliably stumbles upon a seed which actually makes the tests hang.

neongreen:
Sigh.

Can we make Travis output test names as it executes them?

gromakovsky:
> Can we make Travis output test names as it executes them?

There was a poll in Serokell Slack, and results were almost equal, so I decided not to raise this topic further.

gromakovsky:
Personally I would like to have the old behaviour, but I didn't find enough supporters.

neongreen:
After https://github.com/input-output-hk/cardano-sl/pull/1405 is merged it'd be nice if someone could rebase this (because then I'll find out which test fails).

default.nix
```diff
@@ -23,16 +23,20 @@ let
...
+          minutes=0
+          limit=30               # 30 minutes for tests
+          export TESTS_RUNNING="true"
+          while [ "$TESTS_RUNNING" = "true" ]; do
+              echo -n -e " \b"
+              if [ $minutes == $limit ]; then break; fi
+              minutes=$((minutes+1))
+              sleep 60
+          done &
         '';
```
jmitchell: Before merging, verify the output of failing tests is printed. Prior to the `tail -F` approach test results, including failure details, were being written to a log file whose contents weren't printed.

neongreen: @jmitchell 

> verify the output of failing tests is printed

At least locally it does.

gromakovsky:
See #1654.

PR #1390, opened 2017-08-21
[CSL-1526] Support unknown transaction inputs

pva701:


txp/Pos/Binary/Txp/Core.hs
```diff
@@ -18,11 +18,30 @@ import qualified Pos.Txp.Core.Types as T
...
+
 instance Bi T.TxIn where
-  encode txIn = encodeListLen 2 <> encode (T.txInHash txIn) <> encode (T.txInIndex txIn)
-  decode = do
-    enforceSize "TxIn" 2
-    T.TxIn <$> decode <*> decode
+    encode (T.TxInUtxo utxoTxIn) =
+        encodeListLen 2 <>
+        encode (0 :: Word8) <>
+        encode (serialize' utxoTxIn)
```
gromakovsky: It looks wrong. Why did you use `serialize'` here? This seems to be a holywar question, but if you really want this overhead, you should fix `decode` I suppose.
I might be wrong though, the tests will show.

txp/Pos/Txp/Core/Tx.hs
```diff
@@ -41,6 +41,10 @@ topsortTxs toTx input =
     dup a = (a,a)
     txHashes :: HashMap (Hash Tx) a
     txHashes = HM.fromList $ map (over _1 (whHash . toTx) . dup) input
+    txByInput :: TxIn -> Maybe a
+    txByInput (TxInUtxo x) = HM.lookup (txInHash x) txHashes
+    txByInput _ = Nothing
```
gromakovsky: @volhovM review `topsortTxs` changes please.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -186,7 +186,7 @@ verifyInputs VTxContext {..} resolvedInputs TxAux {..} = do
         -> (TxIn, TxOutAux) -- ^ Input and corresponding output data
         -> TxInWitness
         -> m ()
-    inputPredicates i (txIn@TxIn{..}, toa@(TxOutAux txOut@TxOut{..} _)) witness = do
+    inputPredicates i (txIn, toa@(TxOutAux txOut@TxOut{..} _)) witness = do
```
gromakovsky: I don't understand, where is the logic we were discussing?
If there are inputs with unknown types, our behaviour depends on `vtcVerifyAllIsKnown`:
1. If it's `True`, we must report an error that input type is unknown (it should be checked explicitly to have clear error messages).
2. If it's `False`, we should ignore most of the checks (maybe all checks). For instance, we should ignore sum check, etc.

txp/Pos/Binary/Txp/Core.hs
pva701: I just copied from `Bi TxInWitness`

gromakovsky: I don't like how it's done for `TxInWitness`, but it's subjective and it seems that majority likes it, so I don't mind. But there is a bug, note `deserialize'` in `decode @TxInWitness`.

txp/Pos/Binary/Txp/Core.hs
```diff
@@ -18,11 +18,30 @@ import qualified Pos.Txp.Core.Types as T
...
+    encode (T.TxInUnknown tag bs) =
+        encodeListLen 2 <>
+        encode tag <>
+        encode bs
```
georgeee: What exatcly will be done here?
I mean is it just bytestring would be printed or bytestring, prefixed by size? Or how?

georgeee: Ok, I see it's addressed in #1394

pva701:
@gromakovsky re-review plz, I don't want to write tests, because there is some mess :buэ:

gromakovsky:
@pva701 which mess, I don't get it? Maybe there is mess somewhere, but not in serialization tests. Adding a serialization test is 1 line. I won't approve without serialization tests.

node/cardano-sl.cabal
```diff
@@ -19,7 +19,7 @@ Flag with-wallet
   description: Build with wallet
 
 Flag with-explorer
-  default:     True
+  default:     False
```
gromakovsky: You need to justify it.

node/src/Pos/Explorer/Txp/Toil/Logic.hs
```diff
@@ -161,12 +161,12 @@ delTxExtraWithHistory id addrs = do
...
-getTxRelatedAddrs TxAux {taTx = UnsafeTx {..}} undo =
-    map txOutAddress _txOutputs `unionNE` map (txOutAddress . toaOut) undo
+getTxRelatedAddrs TxAux {taTx = UnsafeTx {..}} (catMaybes . toList -> undo) =
+    map txOutAddress _txOutputs `unionNEnList` map (txOutAddress . toaOut) undo
   where
     toSet = HS.fromList . toList
-    -- Safe here, because union of non-empty sets can't be empty.
-    unionNE lhs rhs = NE.fromList $ toList $ HS.union (toSet lhs) (toSet rhs)
+    -- Safe here, because union of non-empty and maybe empty sets can't be empty.
+    unionNEnList lhs rhs = NE.fromList $ toList $ HS.union (toSet lhs) (HS.fromList rhs)
```
gromakovsky: Add a type of this function please, it's a bit hard to read.

node/test/Test/Pos/Lrc/FollowTheSatoshiSpec.hs
```diff
@@ -93,7 +93,7 @@ instance Arbitrary StakeAndHolder where
             nAdr = S.size setUtxo
             values = scanl1 unsafeAddCoin $ replicate nAdr coins
             utxoList =
-                (zipWith TxIn (replicate nAdr txId) [0 .. fromIntegral nAdr]) `zip`
+                (zipWith (\txi id -> TxInUtxo txi id) (replicate nAdr txId) [0 .. fromIntegral nAdr]) `zip`
```
gromakovsky: `\txi id -> TxInUtxo txi id` doesn't look better than `TxInUtxo`.

node/src/Pos/Explorer/Txp/Local.hs
```diff
@@ -164,7 +164,7 @@ eTxProcessTransaction itw@(txId, TxAux {taTx = UnsafeTx {..}}) = do
                 action = eProcessTx curEpoch tx (TxExtra Nothing curTime txUndo)
                 -- NE.fromList is safe here, because if `resolved` is empty, `processTx`
                 -- wouldn't save extra value, thus wouldn't reduce it to NF
```
gromakovsky: AFAIU, now we also rely on that `verifyAllIsKnown` is `True` when we verify local transactions. It would be good to write about it I think.

txp/Pos/Txp/Core/Tx.hs
```diff
@@ -66,7 +70,7 @@ topsortTxs toTx input =
             tsVisited %= HS.insert txHash
             let visitedNew = HS.insert txHash visitedThis
                 dependsUnfiltered =
-                    nub $ mapMaybe (\x -> HM.lookup (txInHash x) txHashes)
+                    nub $ mapMaybe txByInput
```
gromakovsky: @volhovM please review `topsortTx` changes (maybe you're going to review full PR, I dunno).

txp/Pos/Txp/Core/Types.hs
```diff
@@ -142,21 +143,29 @@ instance NFData TxDistribution
...
+    }
+    | TxInUnknown !Word8 !ByteString
+    deriving (Eq, Ord, Generic, Show, Typeable)
 
 instance Hashable TxIn
 
 instance Buildable TxIn where
-    build TxIn {..} = bprint ("TxIn "%shortHashF%" #"%int) txInHash txInIndex
+    build TxInUtxo {..} = bprint ("TxInUtxo "%shortHashF%" #"%int) txInHash txInIndex
+    build (TxInUnknown tag bs) = bprint ("TxInUnknown "%build%" "%base16F) tag bs
```
gromakovsky: It would be more safe to use `int` instead of `build`, but it's not important.

txp/Pos/Txp/Core/Types.hs
```diff
@@ -338,7 +347,7 @@ mkTxPayload txws = do
 ----------------------------------------------------------------------------
 
 -- | Particular undo needed for transactions
```
gromakovsky: It would be good to write what exactly this `TxUndo` means (had to be written long ago) and why it can contain `Nothing`s.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -250,5 +266,9 @@ rollbackTxUtxo
...
     let tx@UnsafeTx {..} = taTx txAux
     let txid = hash tx
-    mapM_ utxoDel $ take (length _txOutputs) $ map (TxIn txid) [0..]
-    mapM_ (uncurry utxoPut) $ NE.zip _txInputs undo
+    mapM_ utxoDel $ take (length _txOutputs) $ map (TxInUtxo txid) [0..]
+    mapM_ (uncurry utxoPut) $ concatMap knownInputAndUndo $ toList $ NE.zip _txInputs undo
+  where
+    knownInputAndUndo (_, Nothing)         = []
+    knownInputAndUndo (TxInUnknown _ _, _) = []
+    knownInputAndUndo (inp, Just u)        = [(inp, u)]
```
gromakovsky: To be honest, returning `Maybe` and using `mapMaybe` looks more conventional to me, but it's not important.
I think it would good to also yell loudly if we encounter `Nothing` and not `TxInUnknown` or `Just` and `TxInUnknown`.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -76,16 +76,30 @@ verifyTxUtxo
...
-    when vtcVerifyAllIsKnown $ verifyAttributesAreKnown _txAttributes
-    return (map snd resolvedInputs, txFee)
+    let unknownTxInMB = find (isTxInUnknown . snd) $ zip [0..] (toList _txInputs)
+    case (vtcVerifyAllIsKnown, unknownTxInMB) of
+        (True, Just (inpId, txIn)) -> throwError $
+            ToilUnknownInput inpId txIn
+        (False, Just _) -> do
+            -- Case when at least one input isn't known
+            minimalReasonableChecks
+            resolvedInputs <- mapM (runMaybeT . resolveInput) _txInputs
```
gromakovsky: I don't understand how this `runMaybeT` can help us. I guess you should right something like `fmap rightToMaybe . runExceptT . resolveInput`.

pva701:
>Adding a serialization test is 1 line.

serialization tests for what? I didn't add new datatypes
@gromakovsky

node/test/Test/Pos/Lrc/FollowTheSatoshiSpec.hs
pva701: i am narkoman

gromakovsky:
Oh, you added it in the first version, but now there is indeed no new types. Ok then.

node/src/Pos/Explorer/Txp/Local.hs
pva701: don't get it. we relied on this earlier, I didn't change anything related to policy. or you just want to clarify it?

txp/Pos/Txp/Toil/Utxo/Functions.hs
pva701: lol, I just want to get list of `Maybe` instead of list of `Either` @gromakovsky

txp/Pos/Txp/Toil/Utxo/Functions.hs
pva701: @gromakovsky do you want `error`?
tbh it's so dangerous. a user might update node and load block from db and decode txin which was unknown earlier but there isn't still undo for this txin.

txp/Pos/Txp/Toil/Utxo/Functions.hs
gromakovsky: So what? Again, how does `runMaybeT` help us here?
```
f :: MonadError () m => Bool -> m ()
f True = pure ()
f False = throwError ()

skovoroda :: (MonadBardaq m, MonadError () m) => m ()
skovoroda = do
  kek :: Maybe () <- runMaybeT (f False)
  print kek
  return ()

g :: MonadBardaq m => m ()
g = do
  print =<< runExceptT skovoroda
```

I expect `g` to print `Left ()`. But that's not what you want here.

txp/Pos/Txp/Toil/Utxo/Functions.hs
gromakovsky: Well, it would be good to at least `logWarning` then, but it looks complicated, because we don't have `WithLogger` here (unlike in update system), so ok, let's leave it how it is.

node/src/Pos/Explorer/Txp/Local.hs
gromakovsky: Previously we didn't rely on it that much. Now it's possible that there won't be any resolved input, but tx will be considered valid. But it's only possible if `verifyAllIsKnown` which is not the case for local tx processing. If someone changes it, this code can be broken.

txp/Pos/Txp/Core/Tx.hs
pva701: @volhovM I am sure everything is fine here, but @volhovM posmoti zbl, a ne :loh: stav' v slack

pva701:
@gromakovsky I vsё

txp/Pos/Txp/Core/Types.hs
```diff
@@ -142,21 +143,29 @@ instance NFData TxDistribution
...
+-- | Transaction arbitrary input.
+data TxIn
+    = TxInUtxo
     { -- | Which transaction's output is used
       txInHash  :: !TxId
       -- | Index of the output in transaction's outputs
     , txInIndex :: !Word32
-    } deriving (Eq, Ord, Show, Generic, Typeable)
+    }
+    | TxInUnknown !Word8 !ByteString
```
gromakovsky: Would be good to document it, btw.

txp/Pos/Binary/Txp/Core.hs
```diff
@@ -19,10 +19,20 @@ import qualified Pos.Txp.Core.Types as T
...
 
 instance Bi T.TxIn where
-  encode txIn = encodeListLen 2 <> encode (T.txInHash txIn) <> encode (T.txInIndex txIn)
-  decode = do
-    enforceSize "TxIn" 2
-    T.TxIn <$> decode <*> decode
+    encode T.TxInUtxo{..} =
+        encodeListLen 2 <>
+        encode (0 :: Word8) <>
+        encode (serialize' (txInHash, txInIndex))
```
volhovM: Is using `serialize'` inside `encode` legal at all? Why not `encodeListLen 3 <> ...`  instead?

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -76,16 +76,30 @@ verifyTxUtxo
...
 verifyTxUtxo ctx@VTxContext {..} ta@(TxAux UnsafeTx {..} witnesses _) = do
-    verifyConsistency _txInputs witnesses
-    verResToMonadError (ToilInvalidOutputs . formatFirstError) $
-        verifyOutputs ctx ta
-    resolvedInputs <- mapM resolveInput _txInputs
-    txFee <- verifySums resolvedInputs _txOutputs
-    verifyInputs ctx resolvedInputs ta
-    when vtcVerifyAllIsKnown $ verifyAttributesAreKnown _txAttributes
-    return (map snd resolvedInputs, txFee)
+    let unknownTxInMB = find (isTxInUnknown . snd) $ zip [0..] (toList _txInputs)
```
volhovM: > non-infix `zip`

Wat? :p

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -76,16 +76,30 @@ verifyTxUtxo
     :: (MonadUtxoRead m, MonadError ToilVerFailure m)
     => VTxContext
     -> TxAux
-    -> m (TxUndo, TxFee)
+    -> m (TxUndo, Maybe TxFee)
```
volhovM: `verifyTxUtxo` documentation doesn't reflect this change.

txp/Pos/Txp/Core/Types.hs
```diff
@@ -338,7 +347,9 @@ mkTxPayload txws = do
 ----------------------------------------------------------------------------
 
 -- | Particular undo needed for transactions
-type TxUndo = NonEmpty TxOutAux
+-- Just means we know transaction input, hence know TxOutAux corresponding to it,
+-- Nothing otherwise.
+type TxUndo = NonEmpty (Maybe TxOutAux)
```
gromakovsky: Yes, newtype would be better, I agree.

PR #1391, opened 2017-08-21
[CSL-1360] Test applying fork to blockchain

int-index:


node/test/Test/Pos/Block/Logic/VarSpec.hs
```diff
@@ -288,3 +303,97 @@ verifyBlockScenarioResult = \case
...
+    -- fork blockchain max index j>i. the upper bound is arbitrary.
+    -- dj=j-i
+    dj <- getRandomR (1, d*2)
+    -- now we can generate paths:
+    --
+    -- B0 - B1 - B2 - B3 - B4 - B5 - B6 - B7
+    --              \
+    --                C3 - C4 - C5 - C6
+    --
+    -- in this example, q=7, d=5, i=2, dj=4
```
gromakovsky: It would be good to also write that `B0` is 0-th genesis block which shouldn't be generated. I didn't understand what was going on, wrote a big wall of text and only then realized it.

int-index:
Tests in their present form have caught three bugs (two in the actual code and one in the testing framework). I propose we merge them as is to avoid regressions and then we can discuss further improvement.

node/test/Test/Pos/Block/Logic/VarSpec.hs
int-index: Is it? Why do you think so?

gromakovsky:
Dunno, ok, let's merge.

gromakovsky:
I am not closing the issue, because I still have some questions. You can close it though.

PR #1392, opened 2017-08-21
CSL-1489 Couple address and attributes, pt. 2

gromakovsky:
~The main part is done. Now there are 11 `undefined`s in places where addresses are created (client code and tests, also block generation). I need to define them. It might be hard somewhere, so I don't know how much time it will take.~

**Update**: it's done and waiting for review and full testing on cluster.


gromakovsky:
**The problems so far:**

1. `instance Buildable WalletUserSecret where` in `Pos.Wallet.Web.Secret` wants to convert a `PublicKey` to an `Address`.
Planned fix: print two addresses, one with bootstrap era distribution, another one with single key distribution. Better solution is to use reflection, but it seems to be unnecessary.

2. `encToCId :: EncryptedSecretKey -> CId w` is defined in `Pos.Wallet.Web.ClientTypes.Functions` and it also wants to convert a `PublicKey` to an `Address`.
Planned fix: pass boolean flag whether we are in bootstrap era. There are many usages of this function, so I would like someone familiar with that code to think whether there will be fundamental problems first.

3. Block payload generator needs to created public key addresses. Should be easy to fix.
Planned fix: need to understand that code better first, but I expect it to be relatively easy.

4. `submitTx` wants to know our own utxos. However, now it's impossible to know whether an `Address` belongs to us. Someone can send coins to our public key and set arbitrary distribution and we won't know this address belongs to us. I am not sure how crucial this problem is. Probably not much. We can think that someone just throws away their money when they do it.
Planned fix: get our utxo for addresses with bootstrap era distribution and single key distribution. If someone sends us coins with multi-key distribution, we won't use these outputs, but this feature is not used by our software anyway, so it should be fine for now.

5. `Test.Pos.Lrc.WorkerSpec` also wants to create addresses. It should be even easier to fix than (3).
Planned fix: just put addresses with single key distribution in those tests. It's a bit hacky, but should work for those tests.

These are all problems in `cardano-sl`. There are more `undefined`s in higher-level projects (e. g. wallet). I will get back to them after I deal with 3, 4, 5 from the list above.

node/src/Pos/Wallet/Web/ClientTypes/Functions.hs
```diff
@@ -54,7 +54,7 @@ cIdToAddress :: CId w -> Either Text Address
 cIdToAddress (CId (CHash h)) = decodeTextAddress h
 
 encToCId :: EncryptedSecretKey -> CId w
-encToCId = addressToCId . makePubKeyAddress . encToPublic
+encToCId = addressToCId . makePubKeyAddress undefined . encToPublic
```
flyingleafe: I think that passing `IsBootstrapEra` as an argument here is the simplest solution.

node/src/Pos/Wallet/Web/Secret.hs
```diff
@@ -34,7 +35,7 @@ instance Buildable WalletUserSecret where
     build WalletUserSecret{..} =
         bprint ("{ root = "%addressF%", set name = "%build%
                 ", wallets = "%pairsF%", accounts = "%pairsF%" }")
-        (makePubKeyAddress $ encToPublic _wusRootKey)
+        (makePubKeyAddress undefined $ encToPublic _wusRootKey)
```
flyingleafe: I think it's safe to provide `IsBootstrapEra False` instead of `undefined` here, because root key addresses are never actually used to store and send money, and are used only for identifying particular HD address tree. Actually, I think that there's no significant difference between `IsBootstrapEra False` and `IsBootstrapEra True` here, but it's better to make some helper (like `makeRootPubKeyAddress`) to avoid confusion.

wallet/src/Pos/Wallet/Web/Account.hs
```diff
@@ -157,7 +157,7 @@ deriveAccountSK
 deriveAccountSK passphrase AccountId{..} accIndex = do
     key <- getSKById aiWId
     maybeThrow badPass $
-        deriveLvl2KeyPair passphrase key aiIndex accIndex
+        deriveLvl2KeyPair undefined passphrase key aiIndex accIndex
```
flyingleafe: Probably we may add `(MonadSlots, MonadGState)` to `AccountMode` to be able to provide `IsBootstrapEra` here.

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -457,7 +457,7 @@ getEncInfo :: EncryptedSecretKey -> (HDPassphrase, CId Wal)
 getEncInfo encSK = do
     let pubKey = encToPublic encSK
     let hdPass = deriveHDPassphrase pubKey
-    let wCId = addressToCId $ makePubKeyAddress pubKey
+    let wCId = addressToCId $ makePubKeyAddress undefined pubKey
```
flyingleafe: Same as with `encToCId` here.

node/src/Pos/Wallet/Web/Secret.hs
gromakovsky: Well, if there is no difference, I would rather use bootstrap era distribution, because it takes less space. Regarding helper: got it, will look into it.

node/src/Pos/Wallet/Web/ClientTypes/Functions.hs
gromakovsky: Do you mean `IsBootstrapEraAddr True`? That's what I am going to do, also will add a TODO comment.

wallet/src/Pos/Wallet/Web/Account.hs
gromakovsky: Won't I spoil something if I make it blocking (using `getCurrentSlotBlocking`)?

gromakovsky:
@georgeee ok, should I deploy a cluster by myself or ask someone to do it?
If the former case, do we have any instructions how to do it?
In the latter case, who should I ask?

gromakovsky:
Also I need to regenerate genesises.

gromakovsky:
Or maybe not. I only changed `MultiKeyDistribution` which we don't use. I think old genesis should be compatible. Anyway, testing will show.

georgeee:
@gromakovsky is everything ready to be tested? Currently I'm not aware we have evrything ready to launch cluster (some stuff is being handled by Networking team)

gromakovsky:
@georgeee I tested it locally: launched demo and sent a tx using `lwallet`. I found a bug, fixed it, now it works fine (I can send a tx).
Tests still fail because there is a bug related to script transactions. I am going to fix it, but I don't think anybody will test scripts, so I don't think it blocks testing. Anyway, I hope to fix it soon.

gromakovsky:
@georgeee update: I fixed script example and now tests pass. I think everything is done in this issue.

gromakovsky:
>  (also contact Philipp to properly benchmark it)

Update: I contacted Philipp, ran benchmarks and Philipp confirmed everything is ok.
So we only need to test on cluster.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -112,7 +112,7 @@ instance HasCoreConstants => MonadAddresses Pos.Wallet.Web.Mode.WalletWebMode wh
     type AddrData Pos.Wallet.Web.Mode.WalletWebMode = (AccountId, PassPhrase)
     getNewAddress (accId, passphrase) = do
         clientAddress <- L.newAddress RandomSeed passphrase accId
-        (, Nothing) <$> decodeCTypeOrFail (cadId clientAddress)
+        decodeCTypeOrFail (cadId clientAddress)
```
volhovM: I don't understand this.

lwallet/Pos/Wallet/Light/Mode.hs
```diff
@@ -135,4 +136,8 @@ instance HasCoreConstants => MonadTxHistory LightWalletSscType LightWalletMode w
 
 instance MonadAddresses LightWalletMode where
     type AddrData LightWalletMode = PublicKey
-    getNewAddress pk = pure (makePubKeyAddress pk, Just $ addressHash pk)
+    getNewAddress pk = do
+        -- Light wallet doesn't know current slot, so let's assume
+        -- it's 0-th epoch. It's enough for our current needs.
+        ibea <- IsBootstrapEraAddr <$> gsIsBootstrapEra 0
```
gromakovsky: Currently this switch is achieved by setting unlock epoch to 0 in benchmarks mode and it works fine.

lwallet/Main.hs
```diff
@@ -183,15 +183,16 @@ runCmd sendActions (SendToAllGenesis duration conc delay_ sendMode tpsSentFile)
         txQueue <- atomically $ newTQueue
         -- prepare a queue with all transactions
         logInfo $ sformat ("Found "%shown%" keys in the genesis block.") (length keysToSend)
+        -- Light wallet doesn't know current slot, so let's assume
+        -- it's 0-th epoch. It's enough for our current needs.
+        ibea <- IsBootstrapEraAddr <$> gsIsBootstrapEra 0
```
gromakovsky: Light wallet is such a bardaq 🙄

node/src/Pos/Launcher/Scenario.hs
```diff
@@ -72,11 +72,9 @@ runNode' NodeResources {..} workers' plugins' = ActionSpec $ \vI sendActions ->
     nodeStartMsg
     inAssertMode $ logInfo "Assert mode on"
     pk <- getOurPublicKey
-    addr <- getOurPubKeyAddress
     let pkHash = addressHash pk
-    logInfoS $ sformat ("My public key is: "%build%
-                        ", address: "%build%
-                        ", pk hash: "%build) pk addr pkHash
+    logInfoS $ sformat ("My public key is: "%build%", pk hash: "%build)
```
gromakovsky: https://github.com/input-output-hk/cardano-sl/commit/0c375b03eb0058bbd4d6cfee21b6f6b25b5545c6

I also don't want to print address depending on whether we are in bootstrap era, because it requires blocking. The only thing I can do is to print two addresses.

txp/Pos/Txp/Core/Types.hs
```diff
@@ -183,15 +157,12 @@ instance NFData TxOut
 -- | Transaction output and auxilary data corresponding to it.
 -- [CSL-366] Add more data.
 data TxOutAux = TxOutAux
```
gromakovsky: Because:
> [CSL-366] Add more data.

I suppose we will do it in this life.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
gromakovsky: In CSL-1483 I introduced a hack: `getNewAddresses` started returning not only an address, but also an optional `StakeholderId` who receives the stake. Now it's useless, because stake distribution is a part of address.

txp/Pos/Txp/Core/Core.hs
```diff
@@ -50,13 +77,70 @@ mkTxProof UnsafeTxPayload {..} =
...
+-- This function splits coin into chunks of @weightsSum@ and
+-- distributes them between bootstrap era stakeholders. Remainder is
+-- assigned randomly (among bootstrap era stakeholders) based on hash of the
+-- coin.
+--
+-- If coin is lower than 'bootDustThreshold' then this function
+-- distributes coins among first stakeholders in the list according to
+-- their weights. The list is ordered by 'StakeholderId's.
+bootstrapEraDistr ::
+       GenesisWStakeholders -> Coin -> [(StakeholderId, Coin)]
```
neongreen: 7-space indentation strikes again

txp/Pos/Txp/Core/Core.hs
```diff
@@ -50,13 +77,70 @@ mkTxProof UnsafeTxPayload {..} =
...
+
+----------------------------------------------------------------------------
+-- Details
+----------------------------------------------------------------------------
+
+-- Compute bootstrap era distribution.
+--
+-- This function splits coin into chunks of @weightsSum@ and
+-- distributes them between bootstrap era stakeholders. Remainder is
+-- assigned randomly (among bootstrap era stakeholders) based on hash of the
```
neongreen: I guess that distributing the remainder among the stakeholders (instead of giving it all to one person) would be nicer, but I don't know whether it's important enough to bother implementing it.

txp/Pos/Txp/Core/Core.hs
```diff
@@ -50,13 +77,70 @@ mkTxProof UnsafeTxPayload {..} =
...
+-- their weights. The list is ordered by 'StakeholderId's.
+bootstrapEraDistr ::
+       GenesisWStakeholders -> Coin -> [(StakeholderId, Coin)]
+bootstrapEraDistr g@(GenesisWStakeholders bootWHolders) c
+    | c < bootDustThreshold g =
+          snd $ foldr foldrFunc (0::Word64,[]) (M.toList bootWHolders)
+    | otherwise =
+          bootHolders `zip` stakeCoins
+  where
+    foldrFunc (s,w) r@(totalSum, res) = case compare totalSum cval of
```
neongreen: something very difficult for 2am :( maybe a more detailed comment would help

txp/Pos/Txp/Core/Types.hs
```diff
@@ -289,58 +256,32 @@ instance NFData TxProof
...
--- Currently there are two invariants:
--- • number of txs must be same as number of witnesses and
---   number of distributions;
--- • each distribution must have the same number of elements as
---   number of outputs in corresponding transaction.
-mkTxPayload
-    :: (Bi Tx, MonadFail m)
-    => [TxAux] -> m TxPayload
+-- Currently there is only one invariant:
+-- • number of txs must be same as number of witnesses.
```
neongreen: The invariant seems to be correct by construction (each tx is paired with a witness), so this comment isn't needed.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -157,23 +153,7 @@ verifyOutputs VTxContext {..} (TxAux UnsafeTx {..} _ distrs) =
           , sformat ("output #"%int%" sends money to a redeem address ("
                     %addressF%"), this is prohibited") i addr
           )
-        ] ++ checkDist i d txOutValue
-    checkDist i d txOutValue =
```
neongreen: Where do these checks live now, if anywhere?

txp/Pos/Script/Examples.hs
```diff
@@ -124,17 +124,16 @@ goodStdlibRedeemer = fromE $ parseRedeemer [text|
...
 multisigValidator :: Int -> [StakeholderId] -> Script
 multisigValidator n ids = fromE $ parseValidator [text|
     validator : List (Maybe (Pair ByteString ByteString)) -> Comp Unit {
         validator sigs = verifyMultiSig
             ${shownN} ${shownIds}
             (!concatenate ${shownTag}
-                (!concatenate
-                    (!concatenate #825820 (txhash))
-                    (!concatenate #5820 (txdistrhash))))
+                (!concatenate #5820 (txhash)
```
neongreen: LGTM.

node/src/Pos/Launcher/Scenario.hs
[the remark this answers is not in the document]
volhovM: Alright, maybe not that important.

core/Pos/Core/Address.hs
```diff
@@ -164,38 +166,60 @@ makeAddress spendingData attributesUnwrapped =
...
+    -> HDAddressPayload    -- ^ Derivation path
     -> PublicKey
     -> Address
-makePubKeyHdwAddress path = makePubKeyAddressImpl (Just path)
+makePubKeyHdwAddress ibe path = makePubKeyAddressImpl (Just path) ibe
 
-makePubKeyAddressImpl :: Maybe HDAddressPayload -> PublicKey -> Address
-makePubKeyAddressImpl path key = makeAddress spendingData attrs
+makePubKeyAddressImpl ::
+       Maybe HDAddressPayload -> IsBootstrapEraAddr -> PublicKey -> Address
```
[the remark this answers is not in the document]
volhovM: @neongreen you've just described my whole work experience.

node/src/Pos/AllSecrets.hs
```diff
@@ -85,7 +85,7 @@ instance Buildable AllSecrets where
             (unInvAddrSpendingData _asSpendingData)
 
 -- | Make simple 'AllSecrets' assuming that only public key addresses
--- exist in the system.
+-- with bootstrap era distribution exist in the system.
 mkAllSecretsSimple :: [SecretKey] -> AllSecrets
```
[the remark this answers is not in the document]
gromakovsky: To be fair, I don't get what you mean. Which predicate are you talking about?

core/Pos/Core/Genesis/Types.hs
```diff
@@ -113,13 +111,13 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
+    unsafeIntegerToCoin . sum $ map fromIntegral $ toList bootHolders
 
 -- | Safe constructor for 'GenesisCoreData'. Throws error if something
 -- goes wrong.
 mkGenesisCoreData ::
        [AddrDistribution]
-    -> HashMap StakeholderId Word16
+    -> Map StakeholderId Word16
     -> Either String GenesisCoreData
 mkGenesisCoreData distribution bootStakeholders = do
```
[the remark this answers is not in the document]
gromakovsky: Ok, I will add this check just in case, but at the same time I admit there may be cases when someone may want to have different distribution in genesis just to test something.
Anyway, if someone needs it, they will comment this check out or come up with something better.

core/Pos/Binary/Core/Address.hs
```diff
@@ -85,14 +86,14 @@ instance Bi AddrStakeDistribution where
         \case
             BootstrapEraDistr -> encodeListLen 0
```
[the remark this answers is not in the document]
gromakovsky: Economy!

core/Pos/Core/Types.hs
```diff
@@ -149,19 +155,39 @@ data AddrType
...
--- TODO [CSL-1489] This stake distribution is currently unused!
-
 -- | Stake distribution associated with an address.
 data AddrStakeDistribution
     = BootstrapEraDistr
     -- ^ Stake distribution for bootstrap era.
     | SingleKeyDistr !StakeholderId
     -- ^ Stake distribution stating that all stake should go to the given stakeholder.
-    | MultiKeyDistr ![(StakeholderId, Coin)]
+    | UnsafeMultiKeyDistr !(Map StakeholderId CoinPortion)
```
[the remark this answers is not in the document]
gromakovsky: It was discussed in https://issues.serokell.io/issue/CSL-1483
It's primary for UX (so that you don't need to regenerate an address when you only change output value and want to keep same portions).

txp/Pos/Txp/Toil/Utxo/Functions.hs
gromakovsky: In `mkMultiKeyDistr`.

txp/Pos/Txp/Core/Types.hs
gromakovsky: That's the invariant of `TxPayload` type (and that's written in the comment) and it doesn't hold by construction. This constructor enforces it.

txp/Pos/Txp/Core/Core.hs
gromakovsky: It's Misha's code, I didn't change the logic, you missed all the action (he failed to write it properly so many times :trollface:).

core/Pos/Core/Address.hs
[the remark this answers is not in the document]
gromakovsky: I dunno why `hindent` behaves this way, to be fair.

gromakovsky:
@volhovM @neongreen I think I addressed all your comments.

node/src/Pos/AllSecrets.hs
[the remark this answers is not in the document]
volhovM: NVM, bred kakoi-to napisal.

georgeee:
Benchmarked locally. Most transactions got in. Runbench script is now not configured with proper rate limitting, hence it's hard to argue on whether PR introduced some regressions. But this is unlikely (address format change either won't work for any tx or will work)

PR #1396, opened 2017-08-22
[CSL-1531] Recovery ~ too much blocks

volhovM:


src/Pos/Block/Network/Retrieval.hs
```diff
@@ -122,9 +123,10 @@ retrievalWorkerImpl sendActions =
         updateRecoveryHeader nodeId header
         endedRecoveryVar <- newEmptyMVar
         let cont (headers :: NewestFirst NE (BlockHeader ssc)) =
-                let firstHeader = headers ^. _Wrapped . _neLast
+                let oldestHeader = headers ^. _Wrapped . _neLast
```
int-index: Can you avoid `_Wrapped` please? The issue is that if the type changes to `OldestFirst`, the logic will break, but the code will compile. There's a `_NewestFirst` iso for this.

volhovM: Alright.

core/constants.yaml
```diff
@@ -31,8 +31,8 @@ dev: &dev
   kademliaDumpInterval: 4
   neighboursSendThreshold: 2
   networkDiameter: 3
-  networkConnectionTimeout: 2000
-  networkReceiveTimeout: 5000
+  networkConnectionTimeout: 2000 # 2 sec
+  networkReceiveTimeout: 20000 # 20 sec
```
int-index: Why did you increase the timeout from 5s to 20s? ot baldy?

src/Pos/Block/Network/Logic.hs
```diff
@@ -270,16 +270,19 @@ requestHeaders
...
     inRecovery <- recoveryInProgress
     logDebug $ sformat ("requestHeaders: inRecovery = "%shown) inRecovery
     flip (maybe onNothing) mHeaders $ \(MsgHeaders headers) -> do
         logDebug $ sformat
-            ("requestHeaders: withConnection: received "%listJson%
-             " from nodeId "%build%" of total size "%builder)
-            (map headerHash headers) nodeId (unitBuilder $ biSize headers)
+            ("requestHeaders: received "%int%" headers of total size "%builder%
+             " from nodeId "%build%": "%listJson)
+            (headers ^. _Wrapped . to NE.length)
```
int-index: Minor: I think you could've used `Foldable.length` here and avoid lenses/unwrapping.

src/Pos/Block/Network/Retrieval.hs
```diff
@@ -122,9 +123,10 @@ retrievalWorkerImpl sendActions =
...
+                    newestHeader = headers ^. _Wrapped . _neHead
                 in handleCHsValid sendActions endedRecoveryVar nodeId
-                                  firstHeader (headerHash header)
+                                  oldestHeader (headerHash newestHeader)
```
int-index: :+1:

PR #1217, opened 2017-07-25
CSM-337 Extrapolate timestamps in case of slot duration change.

ksaric:
https://issues.serokell.io/issue/CSM-337

Most of the work was done by @ajnsit - https://github.com/input-output-hk/cardano-sl/pull/1107
There are some things I need to fix (imports), comments, but this should be it.

I found this code kind of confusing, it seems to be scattered everywhere.

@gromakovsky Can you take a look at this? The most confusing is the fact we track two epoch - the last one and the penult one. I don't even know why, this might be wrong, depending on how we track that - https://github.com/input-output-hk/cardano-sl/compare/ksaric/CSM-337?expand=1#diff-9750ad2f43ce45f68873f4b73aa3c651R87

PR #1399, opened 2017-08-22
CSL-1537 Protocol stops after epoch #1.

ksaric:
https://issues.serokell.io/issue/CSL-1537

@pva701 Already warned me about this in the review, but I decided to obey the tests - https://github.com/input-output-hk/cardano-sl/pull/1217#discussion_r131148737

I can't find any mention of the error in issue - `grep -r --exclude-dir={.stack-work,.git} "epoch is surprisingly unknown" /home/ksaric/projects/curry/cardano-sl/logs/2017-08-23_001513`

And using `./scripts/launch/demo-with-wallet-api.sh` I can reach `11th slot of 8th epoch`. I'll leave it until morning, but I think this is good.

### DIFF
--- a/infra/Pos/Slotting/Impl/Simple.hs
+++ b/infra/Pos/Slotting/Impl/Simple.hs
@@ -63,11 +63,11 @@ getCurrentSlotBlockingSimple
     => SimpleSlottingVar
     -> m SlotId
 getCurrentSlotBlockingSimple var = do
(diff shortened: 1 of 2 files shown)


kantp:
I can confirm that this solves CSL-1537 for my local cluster.

ksaric:
Phew, that was close. I reached `17th slot of 210th epoch` when left overnight with 3 nodes so I think we are good.

PR #1400, opened 2017-08-23
Merge cardano-sl-0.5 to master

Martoon-00:
Changes from #1333 + using `timestampToPosix` in explorer


core/Pos/Core/Timestamp.hs
```diff
@@ -59,6 +61,9 @@ diffTimestamp t1 t2 = getTimestamp t1 - getTimestamp t2
 addMicrosecondsToTimestamp :: Microsecond -> Timestamp -> Timestamp
 addMicrosecondsToTimestamp m t = Timestamp { getTimestamp = (getTimestamp t) + m }
 
+timestampToPosix :: Timestamp -> POSIXTime
+timestampToPosix (Timestamp ts) = fromIntegral ts / 1000000
```
ksaric: You could add this to the `TimeUnit` typeclass. Ideally, the signature would be something like `toUnitInteger :: (Coercible b Integer, TimeUnit b) => a -> b -> Integer`, but you can make it more specific with the default implementation.

Martoon-00: How can I add function to typeclass?

ksaric: https://github.com/acw/time-units/blob/master/Data/Time/Units.hs

Or you can "extend" the typeclass by creating a custom one - `TimeUnitCustom` or something and add that function to it and place it locally in `cardano-sl` module.

Martoon-00: Hm, it may work.
But `POSIXTime` is not integer actually. And it's not clear which type to use instead of `Integer` not to loose precision accidentally (in general case)

gromakovsky:
Needs to be merged again :/

Martoon-00:
Merged under #1365

PR #1402, opened 2017-08-23
Fix client api build if purecsript is already installed

akegalj:
If user has installed wrong version of purescript prior solution won't work as old compiler would be used. This removes locally installed purescript compiler and installs correct compiler

### DIFF
--- a/daedalus/package.json
+++ b/daedalus/package.json
@@ -5,7 +5,7 @@
   "main": "output/Daedalus.ClientApi/index.js",
   "scripts": {
     "start": "npm run build:dev",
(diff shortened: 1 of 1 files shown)


akegalj:
> Does adding purescript@0.11.5 to devDependencies not work? For me npm uninstall purescript is not a good idea.

@sectore can you elaborate? Why do you think its not a good idea. 

If devDependecies fix the problem then lets go that route

domenkozar:
This won't work, as on Nix the purescript is provided by Nix so it shouldn't be uninstalled

gromakovsky:
Let's close it then? @akegalj

PR #1404, opened 2017-08-23
[CSL-1494] Support Tag 24 for embedded CBOR

adinapoli-iohk:
See https://issues.serokell.io/issue/CSL-1494.

This PR **does not** touch the `Bi Address` instance, as that needs to be reworked anyway, and doing that as part of this PR would have been a bit bulkier, so I have decided to postpone it in a separate PR.


adinapoli-iohk:
As part of [this PR](https://github.com/input-output-hk/cardano-sl/pull/1394/commits/8126ff66f76c9a66cb8f6a890b4c9a7025a554fd) I have changed the tag-24-related encoding & decoding functions (after feedback from Duncan), so I guess I'll wait a little bit for the API to consolidate before backporting it here.

adinapoli-iohk:
Ok, I have amended the API accordingly, removed the `U` vs `U24` distinction in `CborSpec` and added more documentation to the `extensionProperty` in `CborSpec`.

PR #1405, opened 2017-08-23
[DEVOPS-274] Pass --format=progress to hspec

jmitchell:
This may fix the suppressed test output issues we've had in Travis.

Don't merge until we confirm CI passes.


jmitchell:
I've reproduced locally even with this change. Cancelled the queued Travis builds and will investigate further.

PR #1406, opened 2017-08-23
[DEVOPS-274] Ensure Travis outputs detailed test report

jmitchell:
Replaces #1405, which didn't work.


domenkozar:
I wonder why this is happening with Nix but not stack. I'm not very much of favor in bashism kicking in, are we sure we can't fix this using cabal+hspec?

domenkozar:
According to https://github.com/haskell/cabal/issues/3024#issuecomment-169504671 let's try `testTarget = "--show-details=direct"`

jmitchell:
After reproducing the issue locally I tried `testTarget = "--show-details=direct"`. Although `cabal test --help` says

> 'direct': send results of test cases in real time; no log file

the output was still suppressed and sent to the log file. I also tried `--log=/dev/stdout`, but this results in a permissions error. In conjunction with these approaches I was also passing `--format=progress` to hspec [through various means](https://hspec.github.io/options.html).

Anyway, Domen and I already agreed to proceed with this change and make an issue to diagnose these problems.

domenkozar:
@dcoutts might have an opinion about this :)

jmitchell:
CI failed due to stale stack2nix output. The fix is [in master](https://github.com/input-output-hk/cardano-sl/commit/924278fa44c4d4903273e1b5af4ae63dda16918f), and master has since been merged here.

neongreen:
macOS builds have timed out.

PR #1407, opened 2017-08-23
[CSM-422] Store wallet backup with unencrypted secret key

Martoon-00:
`exportBackup` endpoint now accepts password to decode encrypted secret key. `importBackup` endpoint accepts password as well - one which is assigned to wallet on import.

Tested with postman and daedalus-bridge, including cases with passphrases.


darko-mijic:
Let's wait with merging this one until we agree on how to encrypt the whole exported file.

flyingleafe:
As we agreed on finishing this only after mainnet, I'll close it now (gonna be reopened later)

PR #1409, opened 2017-08-24
--bench-mode

gromakovsky:


scripts/build/cardano-sl.sh
```diff
@@ -194,8 +221,20 @@ if [[ $for_installer == true ]]; then
   commonargs="$commonargs --flag cardano-sl-tools:for-installer"
 fi
 
-# CONFIG = dev, prod, or wallet
-dconfig=dev
+if [[ $asserts == false ]]; then
+  commonargs="$commonargs --flag cardano-sl-core:-asserts"
```
volhovM: Do we have assertions in core only? 🤔

gromakovsky: Yes. We have `inAssertMode` define there and the rest uses this function.

gromakovsky:
CI failures are not related to this PR.

PR #1410, opened 2017-08-24
[CSL-1493] Initial attempt to remove unsafety from deserialisation

adinapoli-iohk:
See https://issues.serokell.io/issue/CSL-1493 for the bigger context.

This PR does essentially two things:

* It renames `deserialize` (and its strict version) to `unsafeDeserialize`, to make more explicit this is a pure function which might fail at runtime. I still think it's useful for situation like tests, and currently we cannot completely eradicate it from `Bi` instances (see below);

* It introduces a `deserialize` function which runs inside a `Decoder`, so that we can rely on a properly-defined `fail` instance. It seemed more conservative than having this function run in an arbitrary Monad `m`. It also saves a bit of typing and if we really want to have a variant which fails gracefully there is `decodeFull`.

**One subtle change I have operated is to require `deserialize` to consume its input in full**. I'm on the fence whether this is what we want or not, but sounds like for the `Bi` instances this is always almost what we want (that's not the case for other places). If somebody could advise what's the best course of action here, it would be great (I'm not also fond of the proliferation of all this different decoding functions each with different semantic, maybe later in the sprint we can review the situation holistically).

There are some things I would like precious feedback from @gromakovsky . In particular:

* There are a bunch of pure functions inside the DB code which were using `deserialize` before. What to do with those?

* I couldn't remove `deserialize` from `instance Bi (Attributes AddrAttributes)` as that uses the `decodeAttributes` function, which takes an "updater" function which is pure and currently there are calls to `deserialize` happening there. We should decide what we want to do with that.

Let me know if you think this is a good start!


core/Pos/Binary/Core/Address.hs
```diff
@@ -138,9 +138,14 @@ instance Bi (Attributes AddrAttributes) where
             , aaStakeDistribution = BootstrapEraDistr
             }
         go n v acc =
+            -- [CSL-1493] Should we:
+            -- 1. fail explicitly here (by using risky unsafe functions)
+            -- 2. Yield `Nothing` in case we cannot deserialize `v`
+            -- 3. Make `go` and thus `decodeAttributes` take an impure `updater`
+            --    function (which runs in the `Decoder` monad).
```
gromakovsky: (1) is bad for reasons I described in the issue.
(2) is bad because I think we should distinguish cases when we don't want to decode something and we have malformed data.
(3) seems to be good, I don't see disadvantages.

node/src/Pos/DB/Block.hs
```diff
@@ -287,8 +287,9 @@ type MonadBlockDBWrite ssc m
 -- Pure implementation
 ----------------------------------------------------------------------------
 
+-- [CSL-1493] Should we make this function return an `Either e (Block ssc, Undo)`?
```
gromakovsky: In this particular case `unsafeDeserialize'` is not so critical, because we are decoding data from our DB which we control by ourselves. But I would like to avoid it anyway. We have `DBMalformed` exception for the cases when our DB is malformed (so we don't rely that it's always valid and handle it gracefully).
`decodeOrFailPureDB` is used only in `dbGetBlundPure` which has (indirectly `MonadThrow` constraint), so it can be thrown from there (using `throwM`).
tl;dr: yes, I would like to see `Either` here (or `MonadThrow` constraint).

core/Pos/Binary/Core/Address.hs
adinapoli-iohk: Great! Will go with (3) then and report back if there are any unforeseen hiccups.

gromakovsky:
@adinapoli-iohk I pushed regenerated `pkgs/default.nix` to this branch.

adinapoli-iohk:
❤️

adinapoli-iohk:
Argh! The CI spotted an oversight I've made:

```
Aug 24 17:26:19 
Aug 24 17:26:19 Command.hs:124:5: error:
Aug 24 17:26:19     • Couldn't match type ‘cborg-0.1.1.0:Codec.CBOR.Decoding.Decoder
Aug 24 17:26:19                              s0 a0’
Aug 24 17:26:19                      with ‘PublicKey’
Aug 24 17:26:19       Expected type: Text.Parsec.Prim.ParsecT Text () Identity PublicKey
Aug 24 17:26:19         Actual type: Text.Parsec.Prim.ParsecT
Aug 24 17:26:19                        Text () Identity (cborg-0.1.1.0:Codec.CBOR.Decoding.Decoder s0 a0)
Aug 24 17:26:19     • In a stmt of a 'do' block: pure (deserialize' bs)
Aug 24 17:26:19       In the expression:
Aug 24 17:26:19         do { token <- some $ noneOf " ";
Aug 24 17:26:19              bs <- maybe (fail "Incorrect base58") pure
Aug 24 17:26:19                    $ decodeBase58 bitcoinAlphabet (encodeUtf8 $ toText token);
Aug 24 17:26:19              pure (deserialize' bs) }
Aug 24 17:26:19       In an equation for ‘base58PkParser’:
Aug 24 17:26:19           base58PkParser
Aug 24 17:26:19             = do { token <- some $ noneOf " ";
Aug 24 17:26:19                    bs <- maybe (fail "Incorrect base58") pure
Aug 24 17:26:19                          $ decodeBase58 bitcoinAlphabet (encodeUtf8 $ toText token);
Aug 24 17:26:19                    pure (deserialize' bs) }
Aug 24 17:26:19 builder for ‘/nix/store/ndg246mm8fjcx08gah13mnf6c4lmvvxy-cardano-sl-lwallet-0.5.1.drv’ failed with exit code 1
Aug 24 17:26:19 error: build of ‘/nix/store/ndg246mm8fjcx08gah13mnf6c4lmvvxy-cardano-sl-lwallet-0.5.1.drv’ failed
```

Fixing it as we speak..

gromakovsky:
I took a look at the new change, looks good to me.

adinapoli-iohk:
This needs some love to resolve the conflicts, will do later today.

PR #1412, opened 2017-08-24
[CSE-198, CSE-206] 

ksaric:
Trying to figure out why and how https://issues.serokell.io/issue/CSE-206 is causing issues in https://issues.serokell.io/issue/CSE-198 with `{"Left":"decodeFull failed for AbstractHash Blake2b_256 BlockHeaderStub: DeserialiseFailure 0 \"expected bytes\""}`


adinapoli-iohk:
@ksaric I don't know if it's related or it can be useful, but I remember about changing something in that area as part of my PR to restore the `label` mechanism for `Bi` instances. In particular, see this line:

https://github.com/input-output-hk/cardano-sl/pull/1289/files#diff-f70d7ade7a372136e0bc2ea8ab665decR71

I had to give `blake2b` a concrete type `()` as otherwise GHC wasn't able to infer correctly what `b` was. The rationale on why I did that change is that `b` seems to be a phantom type of `AbstractHash` anyway, not used in this particular invocation, but I'm flagging this up in case it's indeed the culprit, even at the cost of being called out for a silly mistake 😉

ksaric:
Thanks for the advice @adinapoli-iohk.
However, it seems that your example is moving from a specific type to a more general one (`toHashSeed :: BackupPhrase -> Either Text ByteString`), while I have the problem with the other way around. Your advice was already suggested by @neongreen and it was very helpful - https://github.com/input-output-hk/cardano-sl/pull/1412/files#diff-05213c918dc0ba7afe5b9532b718cebaL335
I will probably switch it back to `()` if I figure out what's wrong with this.

Switching to `deserialize' $ hashBinary` I get `uncaught exception: DeserialiseFailure (DeserialiseFailure 0 "expected list len")`. 
It does seem (from both the error messages) as though the binary format is unacceptable in this format, but I would have to dig that out to find out why.

adinapoli-iohk:
> Switching to deserialize' $ hashBinary I get uncaught exception: DeserialiseFailure (DeserialiseFailure 0 "expected list len").
It does seem (from both the error messages) as though the binary format is unacceptable in this format, but I would have to dig that out to find out why.

Yes, it looks like it's expecting the input to `deserialize'` to have been encoded this way:

```
encode yy = encodeListLen xx <> ....
```

We use such `encodeListLen` pattern very extensively throughout Cardano, both for our domain data types and for tuples as well (indeed, a tuple `(a,b)` is encoded as a list `[a,b]` and thus the encoding would look like `encodeListLen 2 <> encode a <> encode b`.

ksaric:
As @neongreen suggested, I used `readMaybe` to directly read the binary content back. Added tests.
Results in _explorer_ working as before.

fixed_hash_bug1
fixed_hash_bug2
fixed_hash_bug3

ksaric:
Actually, that was CSE-206 :man_facepalming:

ksaric:
# Tested using

## Import wallet
```
curl -X POST \
  --cacert /home/ksaric/projects/curry/cardano-sl/scripts/tls-files/ca.crt https://localhost:8090/api/wallets/keys \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -d '"/home/ksaric/projects/curry/cardano-sl/keys/1.key"'

curl -X POST \
  --cacert /home/ksaric/projects/curry/cardano-sl/scripts/tls-files/ca.crt https://localhost:8090/api/wallets/keys \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -d '"/home/ksaric/projects/curry/cardano-sl/keys/2.key"'
```

## Create account(s)
```
curl -X POST \
  --cacert /home/ksaric/projects/curry/cardano-sl/scripts/tls-files/ca.crt https://localhost:8090/api/accounts \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -H 'postman-token: a36f7315-b1ec-094b-89f9-27f2e8348134' \
  -d '{
   "caInitMeta": {
      "caName":"Lolol"
   },
   
   "caInitWId": "LaVWe9CiVirQLVR8fgxBYDuM7EbiH4i2kDmtqiM8SeKZ42g4idhR"
  
}'
```

## Send money 

Change the recipient address to the one generated above.

```
curl -X POST \
  --cacert /home/ksaric/projects/curry/cardano-sl/scripts/tls-files/ca.crt https://localhost:8090/api/txs/payments/LaVWX1Lfkqgzb6sDLnUvyxUYunUvBmYspQH4cnLV2UtYartph9kL@2147483648/SfpjaPo632UY5uUpYLFaMruhrQENqFfdCUjcSU4N1yTfREfuXgBkLbLS2ksoisXtNxCRcGembkdULE8cCKPD3h1bcwhXaewuN/3 \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json'
```

ksaric:
This PR solves two issues:
- https://issues.serokell.io/issue/CSE-206
- https://issues.serokell.io/issue/CSE-198

PR #1415, opened 2017-08-24
[CSL-1530] Repair blockgen

volhovM:
This fixes bugs and bad design decisions that led to unusable block-gen.

### DIFF
--- a/node/src/Pos/Generator/Block/Payload.hs
+++ b/node/src/Pos/Generator/Block/Payload.hs
@@ -18,7 +18,6 @@ import           Data.List                  (notElem, (!!))
 import qualified Data.List.NonEmpty         as NE
 import qualified Data.Map                   as M
 import qualified Data.Vector                as V
-import           Ether.Internal             (HasLens (..))
 import           Formatting                 (build, sformat, (%))
 import           System.Random              (RandomGen (..))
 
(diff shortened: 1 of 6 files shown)


node/src/Pos/Generator/BlockEvent.hs
```diff
@@ -185,6 +185,7 @@ genBlocksInTree secrets bootStakeholders blockchainTree = do
             , _bgpBlockCount      = 1
             , _bgpTxGenParams     = txGenParams
             , _bgpInplaceDB       = True
+            , _bgpSkipNoKey       = False
             }
     -- Partial pattern-matching is safe because we specify
     -- blockCount = 1 in the generation parameters.
```
int-index: I already wrote a rant that you've broken this logic (because I've spotted `catMaybes` in `genBlocks`), but now I see that `_bgpSkipNoKey = False` means that it's still valid. Please modify the comment to mention that now we rely on both `blockCount = 1` and `skipNoKey = False`.

volhovM: Where did you write it?

int-index: I wrote it in this review but deleted when I realized that the code is correct.

node/src/Pos/Generator/Block/Payload.hs
```diff
@@ -191,35 +194,35 @@ genTxPayload = do
...
-                txOutAuxsPre <- mapM txOutToOutAux txOuts
+
+                let txOutAuxsPre :: NonEmpty TxOutAux
+                    txOutAuxsPre =
+                        NE.fromList $
+                        zipWith (\pk c -> let distr = one (addressHash pk, c)
+                                              txOut = TxOut (makePubKeyAddress pk) c
+                                          in TxOutAux txOut distr)
+                                outputPks
+                                coins
```
gromakovsky: I guess someone will have fun time merging this stuff with #1416 😉

volhovM:
I tested it, yes, works for me.

PR #1419, opened 2017-08-25
[CSL-1549] policy update, time-warp-nt revision bump

avieth:
This is to use time-warp-nt at https://github.com/serokell/time-warp-nt/pull/96

CI on that pull request keeps timing out. Will go to a master revision as soon as that's sorted.


avieth:
Time limit exceeded on OSX

PR #1420, opened 2017-08-25
CSE-159 Since not showing on most pages.

ksaric:
https://issues.serokell.io/issue/CSE-159

The issue was fixed by https://issues.serokell.io/issue/CSM-337, what we need is to return the UI part to show the "Since" field.

I was unable to test this using `master` because of the error that appears there on the third node (when running, explorer or wallet):
```
cardano-node: too few bytes
From:   Pos.Txp.Core.Types.TxIn:
        demandInput
```

So I used `explorer-mock` to run it, seems to work fine. 
I do need to test this using some mode - _dev_, _qa_ or _staging_.
Returned CSS lost rules as commented by @sectore. Otherwise, I, the CSSHero, would be unable to finish this task.


ksaric:
After `./scripts/clean/db.sh` running `master` seems to work.
After a small fix, we have:

explorer_slots_time

So I think we are good :champagne:

ksaric:
Just Mac build fails :arrow_up:

ksaric:
@sectore Tested, seems to work as it worked before.

PR #1421, opened 2017-08-25
[CSL-1540] Make Topology Yaml parser less lenient

adinapoli-iohk:
See here for the full context: https://issues.serokell.io/issue/CSL-1540

With the new code, now trying to use a hostname instead of an IP address in the `addr` field would result in:

screen shot 2017-08-25 at 14 39 21

Conversely, trying to use an IP address in the `host` field...

screen shot 2017-08-25 at 14 39 37


adinapoli-iohk:
> Looks ok to me. I take it you tested it and the readMaybe stuff works?

I tested everything manually (see screenshots) by tweaking the `topology.yaml` file for a node and observe the failure (or lack thereof) 😉

PR #1423, opened 2017-08-25
Removed deprected node-type argument

avieth:
It's inferred from the topology file.


avieth:
I noticed in iohk-ops that the `--node-type` option is passed. No need, it's never used. Sorry for the confusion. @cleverca22 @deepfire @domenkozar

PR #1426, opened 2017-08-26
[CSL-1561] Add --postboot to lwallet

volhovM:


gromakovsky:
What is it? Why is it needed?

gromakovsky:
> CSL-1559 Make locking in delegation consistent with other code

volhovM:
Fair.

PR #1427, opened 2017-08-26
appveyor:  try to reduce noise

deepfire:
This follows best practices[1] on making `stack setup` stop killing everyone's machines on Appveyor.

-- 
1. https://github.com/commercialhaskell/stack/blob/master/appveyor.yml#L20


rockbmb:
The log is much more readable now and doesn't freeze my browser anymore.

PR #1429, opened 2017-08-27
[CSL-1545] Disable light delegation endpoints

volhovM:
cc @georgeee


gromakovsky:
Please don't restart builds blindly, they contains failed tests, they are referenced in https://issues.serokell.io/issue/CSL-1566.

PR #1430, opened 2017-08-27
CSL-1480 Sort out Bi tests

volhovM:
This PR mostly reduces duplicate tests by moving things out of `CborSpec` to appropriate `BinarySpec`s.


node/test/Test/Pos/Block/Identity/BinarySpec.hs
```diff
@@ -66,8 +65,9 @@ spec = giveStaticConsts $ describe "Block types" $ do
                 describe "Body" $ do
                     binaryTest @(BT.Body (BT.MainBlockchain SscNistBeacon))
                     binaryTest @(BT.Body (BT.MainBlockchain SscGodTossing))
-                binaryTest @BT.MainExtraHeaderData
-                binaryTest @BT.MainExtraBodyData
                 describe "MainToSign" $ do
                     binaryTest @(BT.MainToSign SscNistBeacon)
```
gromakovsky: I have a feeling that we can delete `SscNistBeacon` tests, but it might be subjective and not for this PR :)

node/test/Test/Pos/CborSpec.hs
```diff
@@ -347,197 +304,51 @@ testAgainstFile name x expected =
...
+            testAUnit
+            testANewtype
+            binaryTest @ARecord
+            binaryTest @AUnit
+            binaryTest @ANewtype
+        describe "Lib/core instances" $ do
+            binaryTest @(Attributes X1)
+            binaryTest @(Attributes X2)
+            binaryTest @MessageCode
+            binaryTest @SlottingData
```
gromakovsky: I see some inconsistency in having it here instead of e. g. `Test.Pos.Slotting.BinarySpec`, but it's not critical to me.

node/test/Test/Pos/CryptoSpec.hs
```diff
@@ -111,13 +102,16 @@ spec = describe "Crypto" $ do
                 binaryTest @Crypto.EncShare
                 binaryTest @Crypto.SecretProof
                 binaryTest @Crypto.SecretSharingExtra
+                binaryTest @Crypto.HDAddressPayload
+                binaryTest @(Crypto.AbstractHash SHA256 U)
```
gromakovsky: Are we using SHA256 anywhere?

node/test/Test/Pos/Explorer/Identity/BinarySpec.hs
```diff
@@ -6,13 +6,14 @@ module Test.Pos.Explorer.Identity.BinarySpec
...
-import           Test.Pos.Util  (binaryTest)
+import           Pos.Explorer          (TxExtra)
+import           Test.Pos.Util         (binaryTest)
 
 
 spec :: Spec
 spec = describe "Explorer types" $ do
     describe "Bi instances" $ do
-        binaryTest @TxExtra
+        modifyMaxSuccess (const 100) $ binaryTest @TxExtra
```
gromakovsky: I guess 100 is the default and I don't see a good reason why we need to modify max success for this particular type (is it structurally big?), so it looks a bit confusing to me.

node/test/Test/Pos/Txp/Identity/BinarySpec.hs
```diff
@@ -8,39 +8,42 @@ import           Universum
...
+            binaryTest @T.TxOut
+            binaryTest @T.TxOutAux
+            binaryTest @T.Tx
+            binaryTest @T.TxInWitness
+            modifyMaxSuccess (const 100) $ binaryTest @T.TxDistribution
+            binaryTest @T.TxSigData
+            modifyMaxSuccess (const 100) $ binaryTest @T.TxAux
+            binaryTest @T.TxProof
+            binaryTest @(SmallGenerator T.TxPayload)
+            modifyMaxSuccess (const 10) $ binaryTest @T.TxPayload
```
gromakovsky: I see two conceptually same tests, I think only one of them should survive.

node/test/Test/Pos/Types/Identity/BinarySpec.hs
```diff
@@ -49,3 +54,6 @@ spec = giveStaticConsts $ describe "Types" $ do
             binaryTest @T.ApplicationName
             binaryTest @T.SoftwareVersion
             binaryTest @T.BlockVersion
+        describe "Util/Data" $ do
+            binaryTest @(NewestFirst NE U)
+            binaryTest @(OldestFirst NE U)
```
gromakovsky: Perhaps it's better to rename this module to `Test.Pos.Core.BinarySpec` and move these tests to something like `Test.Pos.Util.ChronoSpec`.

node/test/Test/Pos/CborSpec.hs
volhovM: Oh, yeah, i didn't find a proper place for them. And eventually forgot to think again.

node/test/Test/Pos/CryptoSpec.hs
volhovM: Valid concern. It was in `CborSpec` so i just moved it here. Probably should be removed.

gromakovsky:
FYI: conflicts arrived.

adinapoli-iohk:
> Personally I liked having binary tests in one module more, but I don't have good points in favour of it and it's minor

Probably the boat already left the dock, but I liked having all the tests in one place as well. My point for doing that is that it makes extremely easy for somebody which wants to contribute to the serialisation layer to hack only on 1 file, instead of having to navigate the codebase searching for the appropriate place to add the test. It adds unnecessary overhead, in my opinion, but I'm happy to stick to the majority of consensuses here 😉

volhovM:
I agree with the suggestion to move all binary-related identity tests to the single module. Though I'd prefer to do it in a separate PR since reviewing different refactorings in the single PR is hard (whenever you just move the code around you should check that it's not changed, you know what i mean).

gromakovsky:
I am merging it, because it's useful anyway (e. g. it deletes duplicated tests).
Let's hope that someone will do CSL-1570 at some point and will move tests to `CBorSpec`.

PR #1433, opened 2017-08-27
[CSM-399] Test createTx

Pastafarianist:
TODO:

- [ ] Get rid of hardcoded address seeds
- [ ] Discuss how to get rid of code repetition (_a lot_ in this PR is based on `Test.Pos.Block.Logic.Mode`)
- [x] Discuss whether this PR is large enough to be merged as-is, or
- [x] Write more tests (currently there are only two)

### DIFF
--- a/node/src/Pos/Client/Txp/Util.hs
+++ b/node/src/Pos/Client/Txp/Util.hs
@@ -24,6 +24,8 @@ module Pos.Client.Txp.Util
 
        -- * Additional datatypes
        , TxError (..)
+       , TxOutputs
+       , TxWithSpendings
        ) where
 
 import           Universum
@@ -36,7 +38,7 @@ import qualified Data.List.NonEmpty       as NE
 import qualified Data.Map                 as M
 import qualified Data.Text.Buildable
 import qualified Data.Vector              as V
-import           Formatting               (bprint, build, sformat, stext, (%))
+import           Formatting               (bprint, build, sformat, shown, stext, (%))
 
 import           Pos.Binary               (biSize)
 import           Pos.Client.Txp.Addresses (MonadAddresses (..))
@@ -62,6 +64,10 @@ type TxInputs = NonEmpty TxIn
 type TxOwnedInputs owner = NonEmpty (owner, TxIn)
 type TxOutputs = NonEmpty TxOutAux
 type TxWithSpendings = (TxAux, NonEmpty TxOut)
+
+instance Buildable TxWithSpendings where
+    build (txAux, neTxOut) =
+        bprint ("("%shown%", "%shown%")") txAux neTxOut
 
 -- This datatype corresponds to raw transaction.
 data TxRaw = TxRaw
@@ -89,7 +95,7 @@ instance Buildable TxError where
     build (FailedToStabilize iters) =
         bprint ("Transaction creation error: failed to stabilize fee after "%build%" iterations") iters
     build (OutputIsRedeem addr) =
-        bprint ("Destination address "%build%" is a redemption address") addr
+        bprint ("Output address "%build%" is a redemption address") addr
     build RedemptionDepleted =
         bprint "Redemption address balance is 0"
     build (GeneralTxError msg) =
--- a/node/src/Pos/Client/Txp/Util.hs
+++ b/node/src/Pos/Client/Txp/Util.hs
@@ -38,7 +38,8 @@ import qualified Data.List.NonEmpty       as NE
 import qualified Data.Map                 as M
 import qualified Data.Text.Buildable
 import qualified Data.Vector              as V
-import           Formatting               (bprint, build, sformat, shown, stext, (%))
+import           Formatting               (bprint, build, sformat, stext, (%))
+import           Serokell.Util            (listJson)
 
 import           Pos.Binary               (biSize)
 import           Pos.Client.Txp.Addresses (MonadAddresses (..))
@@ -67,7 +68,7 @@ type TxWithSpendings = (TxAux, NonEmpty TxOut)
 
 instance Buildable TxWithSpendings where
     build (txAux, neTxOut) =
-        bprint ("("%shown%", "%shown%")") txAux neTxOut
+        bprint ("("%build%", "%listJson%")") txAux neTxOut
 
 -- This datatype corresponds to raw transaction.
 data TxRaw = TxRaw
--- a/node/test/Test/Pos/Block/Logic/Mode.hs
+++ b/node/test/Test/Pos/Block/Logic/Mode.hs
@@ -19,6 +19,7 @@ module Test.Pos.Block.Logic.Mode
 
        , BlockProperty
        , blockPropertyToProperty
+       , genSuitableStakeDistribution
        ) where
 
 import           Universum
@@ -110,7 +111,6 @@ import           Test.Pos.Block.Logic.Emulation (Emulation (..), runEmulation, s
 -- before testing starts.
 data TestParams = TestParams
     { _tpGenesisContext     :: !GenesisContext
-    -- ^ Genesis context.
     , _tpAllSecrets         :: !AllSecrets
     -- ^ Secret keys corresponding to 'PubKeyAddress'es from
     -- genesis 'Utxo'.
@@ -166,20 +166,20 @@ instance Arbitrary TestParams where
             toList @(NonEmpty SecretKey) <$>
              -- might have repetitions
             (arbitrary `suchThat` (\l -> length l < 15))
-        let _tpStartTime = fromMicroseconds 0
         let invSecretsMap = mkInvSecretsMap secretKeysList
         let publicKeys = map toPublic (toList invSecretsMap)
         let addresses = map makePubKeyAddressBoot publicKeys
         let invAddrSpendingData =
                 mkInvAddrSpendingData $
                 addresses `zip` (map PubKeyASD publicKeys)
-        let _tpAllSecrets = AllSecrets invSecretsMap invAddrSpendingData
         stakeDistribution <-
             genSuitableStakeDistribution (fromIntegral $ length invSecretsMap)
         let addrDistribution = [(addresses, stakeDistribution)]
         let _tpGenesisContext =
                 genesisContextImplicit invAddrSpendingData addrDistribution
+        let _tpAllSecrets = AllSecrets invSecretsMap invAddrSpendingData
         let _tpStakeDistributions = one stakeDistribution
+        let _tpStartTime = fromMicroseconds 0
         return TestParams {..}
 
 ----------------------------------------------------------------------------
--- a/node/test/Test/Pos/Client/Txp/Mode.hs
+++ b/node/test/Test/Pos/Client/Txp/Mode.hs
@@ -1,0 +1,291 @@
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Execution mode for tx creation tests.
+
+module Test.Pos.Client.Txp.Mode
+       ( TxpTestProperty
+       , TxpTestMode
+       ) where
+
+import           Universum
+
+import           Control.Lens                   (lens, makeClassy, makeLensesWith)
+import qualified Data.ByteString                as BS
+import           Data.Time.Units                (Microsecond, fromMicroseconds)
+import           Ether.Internal                 (HasLens (..))
+import           Mockable                       (Production, currentTime, runProduction)
+import           Test.QuickCheck                (Arbitrary (..), Gen, Property,
+                                                 Testable (..), forAll, ioProperty,
+                                                 suchThat)
+import           Test.QuickCheck.Monadic        (PropertyM, monadic)
+
+import           Pos.AllSecrets                 (mkInvAddrSpendingData, mkInvSecretsMap)
+import           Pos.Block.Core                 (Block, BlockHeader)
+import           Pos.Block.Slog                 (mkSlogGState)
+import           Pos.Block.Types                (Undo)
+import           Pos.Client.Txp.Addresses       (MonadAddresses (..))
+import           Pos.Client.Txp.Util            (TxCreateMode)
+import           Pos.Core                       (AddrSpendingData (..), HasCoreConstants,
+                                                 IsHeader, SlotId, Timestamp (..),
+                                                 makePubKeyAddressBoot)
+import           Pos.Crypto                     (SecretKey, toPublic)
+import           Pos.DB                         (MonadGState (..))
+import qualified Pos.DB                         as DB
+import qualified Pos.DB.Block                   as DB
+import           Pos.DB.DB                      as DB
+import           Pos.DB.DB                      (initNodeDBs)
+import           Pos.Genesis                    (GenesisContext (..), GenesisUtxo,
+                                                 GenesisWStakeholders,
+                                                 genesisContextImplicit, gtcUtxo,
+                                                 gtcWStakeholders)
+import qualified Pos.GState                     as GS
+import           Pos.Lrc                        (LrcContext (..), mkLrcSyncData)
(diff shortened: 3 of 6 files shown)


node/test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -166,20 +166,20 @@ instance Arbitrary TestParams where
...
                 addresses `zip` (map PubKeyASD publicKeys)
-        let _tpAllSecrets = AllSecrets invSecretsMap invAddrSpendingData
         stakeDistribution <-
             genSuitableStakeDistribution (fromIntegral $ length invSecretsMap)
         let addrDistribution = [(addresses, stakeDistribution)]
         let _tpGenesisContext =
                 genesisContextImplicit invAddrSpendingData addrDistribution
+        let _tpAllSecrets = AllSecrets invSecretsMap invAddrSpendingData
         let _tpStakeDistributions = one stakeDistribution
+        let _tpStartTime = fromMicroseconds 0
```
flyingleafe: These changes seem pointless.

node/test/Test/Pos/Client/Txp/Mode.hs
```diff
@@ -0,0 +1,288 @@
...
+-- An object of this type has instance Arbitrary and is used as the source
+-- of randomness in the tests.
+data TxpTestParams = TxpTestParams
+    { _ttpGenesisContext :: !GenesisContext
+    , _ttpStartTime      :: !Microsecond
+    } deriving (Show)
+
+makeClassy ''TxpTestParams
+
+instance Arbitrary TxpTestParams where
```
flyingleafe: This is almost perfect copypaste of `TestParams`, which you actually never use in tests - because you generate a new utxo every time manually, and you don't need information about genesis stakeholders at all. It seems to me that this may be safely removed completely - correct me if I'm wrong.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -0,0 +1,111 @@
...
+        "An attempt to create a tx for 1 coin when you have 1 coin fails because of the fee"
+
+createMTxWorksWhenWeAreRichSpec :: HasCoreConstants => TxpTestProperty ()
+createMTxWorksWhenWeAreRichSpec = do
+    txOrError <- createMTx cmpUtxo cmpSigners cmpOutputs cmpAddrData
+    case txOrError of
+        Left err -> stopProperty $ sformat ("Failed to create tx: "%build) err
+        Right tx -> ensureTxMakesSense tx cmpTxOutAuxInput cmpTxOutAuxOutput
+  where
+    CreateMTxParams {..} = make1to1Params 1000000 1
```
flyingleafe: And _here_ (as well as in other similar cases) you may use `arbitrary` instances to test a range of acceptable parameters instead of just one.

node/test/Test/Pos/Block/Logic/Mode.hs
Pastafarianist: This is done to align the ordering of the fields in `TestParams` with their initialization. A purely cosmetic change.

node/test/Test/Pos/Client/Txp/Mode.hs
Pastafarianist: Actually, I do use it: it is required by `initNodeDBs` in `initTxpTestContext`. In turn, `initNodeDBs` is required to have the `MonadGState` context.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
Pastafarianist: Working on it.

node/test/Test/Pos/Client/Txp/Mode.hs
```diff
@@ -0,0 +1,291 @@
...
+    , ttcSSlottingVar :: !SimpleSlottingVar
+    , ttcSlotId       :: !(Maybe SlotId)
+    -- ^ If this value is 'Just' we will return it as the current
+    -- slot. Otherwise simple slotting is used.
+    , ttcParams       :: !TxpTestParams
+    }
+
+makeLensesWith postfixLFields ''TxpTestContext
+
+instance HasCoreConstants => TxCreateMode TxpTestMode
```
gromakovsky: It seems this line doesn't do anything, because `TxCreateMode` is a just an alias for other constraints. I didn't even know it's possible to write such thing.
Anyway, I think it's ok to have it here, because it is helpful when code stops compiling due to a missing instance.

node/test/Test/Pos/Client/Txp/Mode.hs
```diff
@@ -0,0 +1,291 @@
...
+            Just slot -> pure (Just slot)
+    getCurrentSlotBlocking =
+        view ttcSlotId_L >>= \case
+            Nothing -> Slot.getCurrentSlotBlockingSimple =<< view ttcSSlottingVar_L
+            Just slot -> pure slot
+    getCurrentSlotInaccurate = do
+        view ttcSlotId_L >>= \case
+            Nothing -> Slot.getCurrentSlotInaccurateSimple =<< view ttcSSlottingVar_L
+            Just slot -> pure slot
+    -- FIXME: this has already caused a bug with Explorer (CSE-203).
```
gromakovsky: Note that CSE-203 was fixed and `currentTimeSlottingSimple` should be ok now. I don't think FIXME still makes sense here.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -0,0 +1,246 @@
...
+    prop manyAddressesTo1Desc manyAddressesTo1Spec
+    prop manyAddressesToManyDesc manyAddressesToManySpec
+    prop redemptionDesc redemptionSpec
+    prop txWithRedeemOutputFailsDesc txWithRedeemOutputFailsSpec
+  where
+    createMTxWorksWhenWeAreRichDesc =
+        "Transaction is created successfully when we have 1 input with 1M coins " <>
+        "and 1 output with 1 coin"
+    stabilizationDoesNotFailDesc =
+        "The message \"Couldn't stabilize tx fee after 5 attempts\" is not thrown " <>
```
gromakovsky: I don't think it's a good idea to use error message instead of exception type here. I would write that `FailedToStabilize` is not thrown.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -0,0 +1,246 @@
...
+                    "Available: "%shown%"\n"%
+                    "Used: "%shown
+                    ) txOutAvailable txOutUsed
+  where
+    txOutAvailable = map toaOut $ M.elems utxo
+    txOutUsed = NE.toList neTxOut
+
+unsafeIntegerToTxId :: Integer -> TxId
+unsafeIntegerToTxId n =
+    unsafeFromJust $ rightToMaybe $ decodeHash $
```
gromakovsky: I would recommend using `leftToPanic "something"` instead of `unsafeFromJust . rightToMaybe` to have more descriptive error message when impossible happens.

node/test/Test/Pos/Client/Txp/Mode.hs
```diff
@@ -0,0 +1,291 @@
...
+    dbGet = DB.dbGetPureDefault
+    dbIterSource = DB.dbIterSourcePureDefault
+
+instance MonadAddresses TxpTestMode where
+    type AddrData TxpTestMode = ()
+    getNewAddress _ = pure address
+      where
+        -- seed for address generation is a ByteString with 32 255's
+        seed = BS.replicate seedSize (255 :: Word8)
+        (_, address) = generateAddressAux seed
```
volhovM: Wait, so we do not own the related secret key for the change address? Don't we want to use the change later? How does it work?

node/test/Test/Pos/Client/Txp/Util.hs
```diff
@@ -0,0 +1,40 @@
...
+
+generateAddressAux :: ByteString -> (SecretKey, Address)
+generateAddressAux seed =
+    if length seed /= seedSize then
+        error $ sformat ("Internal error: seed size must be exactly "%build%" bytes") seedSize
+    else
+        let (pk, sk) = deterministicKeyGen seed
+        in (sk, makePubKeyAddressBoot pk)
+
+generateRedeemAddressAux :: ByteString -> (RedeemSecretKey, Address)
```
volhovM: Copy-paste. Can you merge it with `generateAddressAux`? You don't even specify the exact function in error string.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -0,0 +1,246 @@
...
+    utxo = one (TxInUtxo (unsafeIntegerToTxId 0) 0, txOutAuxInput)
+    signers = one (fakeSigner sk, addr)
+    outputs = one txOutAuxOutput
+    addrData = ()
+
+----------------------------------------------------------------------------
+-- Helpers
+----------------------------------------------------------------------------
+
+data CreateMTxParams = CreateMTxParams
```
volhovM: Would be good to document (fields too).

node/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -0,0 +1,246 @@
...
+    CreateMTxParams {..} = makeManyAddressesToManyParams 1 200000 1 1
+
+feeIsNonzeroSpec :: HasCoreConstants => TxpTestProperty ()
+feeIsNonzeroSpec = do
+    txOrError <- createMTx cmpUtxo cmpSigners cmpOutputs cmpAddrData
+    case txOrError of
+        Left (NotEnoughMoney _) -> return ()
+        Left err -> stopProperty $ pretty err
+        Right _ -> stopProperty $
+            sformat ("Transaction was created even though there were "%
```
volhovM: Why do you use sformat without any formatters inside? :confused:

node/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -0,0 +1,246 @@
...
+manyAddressesTo1Spec :: HasCoreConstants => TxpTestProperty ()
+manyAddressesTo1Spec = do
+    txOrError <- createMTx cmpUtxo cmpSigners cmpOutputs cmpAddrData
+    case txOrError of
+        Left err -> stopProperty $ pretty err
+        Right tx -> ensureTxMakesSense tx cmpUtxo cmpOutputs
+  where
+    CreateMTxParams {..} = makeManyAddressesToManyParams 10 100000 1 1
+
+manyAddressesToManySpec :: HasCoreConstants => TxpTestProperty ()
```
volhovM: Can be merged with `manyAddressesTo1Spec`.

node/src/Pos/Client/Txp/Util.hs
```diff
@@ -63,6 +65,10 @@ type TxOwnedInputs owner = NonEmpty (owner, TxIn)
 type TxOutputs = NonEmpty TxOutAux
 type TxWithSpendings = (TxAux, NonEmpty TxOut)
 
+instance Buildable TxWithSpendings where
+    build (txAux, neTxOut) =
+        bprint ("("%shown%", "%shown%")") txAux neTxOut
```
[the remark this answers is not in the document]
Pastafarianist: I didn't know about `listJson`. Fixed.

node/test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -110,7 +111,6 @@ import           Test.Pos.Block.Logic.Emulation (Emulation (..), runEmulation, s
 -- before testing starts.
 data TestParams = TestParams
     { _tpGenesisContext     :: !GenesisContext
-    -- ^ Genesis context.
```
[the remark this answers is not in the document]
Pastafarianist: It does not convey any useful information. Are you sure?

node/test/Test/Pos/Client/Txp/Mode.hs
Pastafarianist: It doesn't, but it acts somewhat like an assertion: it makes sure that `TxpTestMode` actually implements everything from `TxCreateMode`.

node/test/Test/Pos/Client/Txp/Mode.hs
Pastafarianist: The fix for CSE-203 was exactly to replace `currentTimeSlottingSimple` with `Timestamp 0`.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
Pastafarianist: This is a leftover from before the `TxError` refactoring. Fixed.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
Pastafarianist: Didn't know about that helper. Fixed.

node/test/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -19,6 +19,7 @@ module Test.Pos.Block.Logic.Mode
 
        , BlockProperty
        , blockPropertyToProperty
+       , genSuitableStakeDistribution
```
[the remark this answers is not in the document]
Pastafarianist: It is not a reexport, it's actually defined right there.

node/test/Test/Pos/Client/Txp/Mode.hs
Pastafarianist: The change is never used in the tests.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
Pastafarianist: Fixed.

node/test/Test/Pos/Client/Txp/Util.hs
Pastafarianist: I can factor out the common part. Cannot merge the two functions because the types differ.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
Pastafarianist: Not sure how. Can you elaborate?

node/test/Test/Pos/Client/Txp/Util.hs
volhovM: I meant exactly factoring common code out, yes.

node/test/Test/Pos/Client/Txp/UtilSpec.hs
volhovM: Two functions _only_ differ in the third argument of `makeManyAddressesToManyParams`.

node/test/Test/Pos/Client/Txp/Mode.hs
gromakovsky: The fix for CSE-203 was not to use current time as timestamp in block processing, because it doesn't make sense.

node/test/Test/Pos/Client/Txp/Mode.hs
gromakovsky: > Wait, so we do not own the related secret key for the change address?

Actually we do own, it's generated deterministically from constant seed.
But such kind of ownage is quite bad 😝 
Anyway, it doesn't matter if the change isn't used.

node/test/Test/Pos/Block/Logic/Mode.hs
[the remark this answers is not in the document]
gromakovsky: Yes, because later someone will write `TODO: comment` comment, then there will be wars, etc.
When there is a comment, nobody is concerned about adding it :)

PR #1434, opened 2017-08-27
[CSL-1550] Add --behavior

neongreen:
Note that the first part of CSL-1550 was (improperly) merged in a different branch: https://github.com/input-output-hk/cardano-sl/pull/1130/commits/ea0f06c6009896ebfc75c4ee68e9f122bd67b1e8.


core/Pos/Aeson/Crypto.hs
```diff
@@ -0,0 +1,21 @@
...
+
+import           Crypto.Hash   (HashAlgorithm)
+import           Data.Aeson    (FromJSON (..), ToJSON (..))
+import           Formatting    (sformat)
+
+import           Pos.Crypto    (AbstractHash, PublicKey, fullPublicKeyF, hashHexF)
+import           Pos.Util.Util (parseJSONWithRead)
+
+instance ToJSON (AbstractHash algo a) where
+    toJSON = toJSON . sformat hashHexF
```
georgeee: why not `show` here (for implementations to be consistent)?

neongreen: This isn't my code, I just moved it.

core/cardano-sl-core.cabal
```diff
@@ -150,32 +143,42 @@ library
...
                      , pvss
                      , quickcheck-instances
                      , random
+                     , reflection
+                     , resourcet
                      , safecopy
                      , semigroups
                      , serokell-util
                      , stm
+                     , stm
```
domenkozar: double?

neongreen: Bad merge, will fix it.

PR #1435, opened 2017-08-28
[CSL-1562] Improve 'Poll' tests

rockbmb:


node/test/Test/Pos/Update/PollSpec.hs
```diff
@@ -152,3 +162,64 @@ applyActions ps actionList =
...
+    mempty
+    mempty
+    mempty
+    mempty
+    mempty
+    mempty
+    mempty
+
+perform :: [PollAction] -> Poll.PurePoll ()
+perform = foldl (>>) (return ()) . map actionToMonad
```
gromakovsky: Isn't `foldl (>>) (return ())` the same as `sequence`?

update/Pos/Update/Poll/Pure.hs
```diff
@@ -119,6 +119,7 @@ instance Bi UpdateProposal => MonadPoll PurePoll where
           appName = svAppName . upSoftwareVersion $ uProp
     deactivateProposal ui =
         PurePoll $ (Poll.psActiveProposals . at ui .= Nothing) >>
-                   (Poll.psActivePropsIdx . mapped . at ui .= Nothing)
+                   (Poll.psActivePropsIdx . mapped . at ui .= Nothing) >>
+                   (Poll.psActivePropsIdx %= HM.filter (not . null))
```
gromakovsky: Perhaps it would be more reasonable to use `do`-notation here instead of two sequenced `>>`s.

node/test/Test/Pos/Update/PollSpec.hs
```diff
@@ -152,3 +162,73 @@ applyActions ps actionList =
...
+    mempty
+    mempty
+    mempty
+    mempty
+    mempty
+    mempty
+
+-- | Apply a sequence of 'PollAction's from left to right.
+perform :: [PollAction] -> Poll.PurePoll ()
+perform = sequence . map actionToMonad
```
Martoon-00: And now it's the same as `traverse_` )

node/test/Test/Pos/Update/PollSpec.hs
rockbmb: http://hackage.haskell.org/package/base-4.10.0.0/docs/src/Data.Foldable.html#sequence_

```
sequence_ :: (Foldable t, Monad m) => t (m a) -> m ()
sequence_ = foldr (>>) (return ())
```

I need it to be a `foldl` here. I can just reverse the list, but I'd rather not overengineer this.

PR #1437, opened 2017-08-28
CSL-1529  Change the way data is updated in SlottingVar 

gromakovsky:
This PR does two things:
1. It introduces `LockedTVar` type which can be used to synchronize in-memory data and data stored in DB. For more context this issue description and CSL-1560. However, it's currently not used, because I decided to leave its usages for CSL-1560 and because it's not really necessary for CSL-1529.
2. It moves slotting data update from Slog to Update and changes the way it's done. I left a comment which explains why desynchronization of in-memory data and data from DB is ok. Apart from desynchronization problem (which would be substantial if we updated memory state after updating DB), it solves two other problems:
• previously data was read from DB and put into memory state before it was written to DB. It somehow worked because we apply genesis block separately and update slotting data only when apply genesis block.
• In rollback we were using `putEpochSlottingDataM` despite the fact that it can delete some slotting data, not add it.

### DIFF
--- a/node/src/Pos/Block/Slog/Logic.hs
+++ b/node/src/Pos/Block/Slog/Logic.hs
@@ -25,7 +25,6 @@ import           Universum
 import           Control.Lens           (_Wrapped)
 import           Control.Monad.Except   (MonadError (throwError))
 import qualified Data.List.NonEmpty     as NE
-import qualified Data.Map.Strict        as M
 import           Formatting             (build, sformat, (%))
 import           Serokell.Util          (Color (Red), colorize)
 import           Serokell.Util.Verify   (formatAllErrors, verResToMonadError)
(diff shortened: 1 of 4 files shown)


gromakovsky:
Launched demo just to check slotting data is updated. Stopped on block for `slot: 1st slot of 4th epoch`. No problems with slotting.

node/src/Pos/Block/Slog/Logic.hs
```diff
@@ -328,9 +326,3 @@ slogCommon
     -> m ()
 slogCommon newLastSlots = do
     slogPutLastSlots newLastSlots
```
rockbmb: What happened to this function? Does it require no further changes?

gromakovsky: This logic was moved into `Pos.Update.Logic.Global`.

PR #1440, opened 2017-08-28
[CSL-1517] Check upper bound in decode@Coin

gromakovsky:


core/Pos/Binary/Core/Coin.hs
```diff
@@ -22,4 +22,11 @@ import           Pos.Core.Types   (Coin, mkCoin, unsafeGetCoin)
 
 instance Bi Coin where
     encode = encode . unsafeGetCoin
-    decode = mkCoin <$> decode
+    decode =
+        decode >>= \case
+            number
+                | number > unsafeGetCoin maxBound ->
```
rockbmb: The error check being here and not in `encode` means we'd rather throw an error when decoding than when encoding, right?

gromakovsky: No, because you can't create `Coin` with invalid value (see `mkCoin` constructor). Here we just make sure that `mkCoin` won't fail.

PR #1441, opened 2017-08-28
[CSL-823] Rework security workers

gromakovsky:


godtossing/Pos/Ssc/GodTossing/Workers.hs
```diff
@@ -419,3 +431,46 @@ waitUntilSend msgTag epoch slMultiplier = do
...
+checkForIgnoredCommitmentsWorker
+    :: forall ctx m.
+       SscMode SscGodTossing ctx m
+    => (WorkerSpec m, OutSpecs)
+checkForIgnoredCommitmentsWorker = localWorker $ do
+    counter <- newTVarIO 0
+    void $ onNewSlot True (checkForIgnoredCommitmentsWorkerImpl counter)
+
+-- This worker checks whether our commitments appear in blocks. This
+-- check is done only if we actually should participate in
```
neongreen: The comment should say what happens if this check fails (i.e. just emits a warning).

godtossing/Pos/Ssc/GodTossing/Workers.hs
```diff
@@ -419,3 +431,46 @@ waitUntilSend msgTag epoch slMultiplier = do
...
+-- GodTossing. It's triggered if there are
+-- 'mdNoCommitmentsEpochThreshold' consequent epochs during which we
+-- had to participate in GodTossing, but our commitment didn't appear
+-- in blocks.
+checkForIgnoredCommitmentsWorkerImpl
+    :: forall ctx m. (SscMode SscGodTossing ctx m)
+    => TVar Word -> SlotId -> m ()
```
neongreen: And here it'd be nice to have a haddock saying what this counter counts.

godtossing/Pos/Ssc/GodTossing/Workers.hs
```diff
@@ -419,3 +431,46 @@ waitUntilSend msgTag epoch slMultiplier = do
...
+checkForIgnoredCommitmentsWorkerImpl counter SlotId {..}
+    -- It's enough to do this check once per epoch near the end of the epoch.
+    | getSlotIndex siSlot /= 9 * fromIntegral blkSecurityParam = pass
```
neongreen: I think that `when (getSlotIndex siSlot == 9 * fromIntegral blkSecurityParam) $ ...` is a bit nicer, but whatever.

godtossing/Pos/Ssc/GodTossing/Workers.hs
```diff
@@ -419,3 +431,46 @@ waitUntilSend msgTag epoch slMultiplier = do
...
+            ourId <- getOurStakeholderId
+            globalCommitments <-
+                getCommitmentsMap . view gsCommitments <$> gtGetGlobalState
+            unless (ourId `HM.member` globalCommitments) $ do
+                -- `modifyTVar'` returns (), hence not used
+                newCounterValue <-
+                    atomically $ do
+                        !x <- succ <$> readTVar counter
+                        x <$ writeTVar counter x
+                when (newCounterValue > mdNoCommitmentsEpochThreshold) $ do
```
neongreen: Shouldn't you reset this counter at some point?

godtossing/Pos/Ssc/GodTossing/Workers.hs
gromakovsky: It's a private function of this module, so I think it's not necessary, regular comment should be enough (will add).

godtossing/Pos/Ssc/GodTossing/Workers.hs
gromakovsky: Ok.

godtossing/Pos/Ssc/GodTossing/Workers.hs
gromakovsky: I should.

gromakovsky:
@neongreen it's ready for re-review.

PR #1442, opened 2017-08-28
[CSL-1565] Refactor TxError

Pastafarianist:
This is needed for CSM-399.

### DIFF
--- a/node/src/Pos/Client/Txp/Util.hs
+++ b/node/src/Pos/Client/Txp/Util.hs
@@ -74,19 +74,27 @@ data TxRaw = TxRaw
     -- ^ Remaining money
     }
 
-data TxError
-    = TxError { unTxError :: !Text }
+data TxError =
+      NotEnoughMoney !Coin    -- ^ Parameter: how much more money is needed
+    | FailedToStabilize !Int  -- ^ Parameter: how many attempts were performed
+    | OutputIsRedeem !Address -- ^ One of the tx outputs is a redemption address
+    | RedemptionDepleted      -- ^ Redemption address has already been used
+    | GeneralTxError !Text    -- ^ Parameter: description of the problem
     deriving (Show, Generic)
 
 instance Exception TxError
 
 instance Buildable TxError where
-    build (TxError msg) = bprint ("Transaction creation error ("%stext%")") msg
-
-throwTxError
-    :: MonadError TxError m
-    => Text -> m a
-throwTxError = throwError . TxError
+    build (NotEnoughMoney coin) =
+        bprint ("Transaction creation error: not enough money, need "%build%" more") coin
+    build (FailedToStabilize iters) =
+        bprint ("Transaction creation error: failed to stabilize fee after "%build%" iterations") iters
+    build (OutputIsRedeem addr) =
+        bprint ("Destination address "%build%" is a redemption address") addr
+    build RedemptionDepleted =
+        bprint "Redemption address balance is 0"
+    build (GeneralTxError msg) =
+        bprint ("Transaction creation error: "%stext) msg
 
 -----------------------------------------------------------------------------
 -- Tx creation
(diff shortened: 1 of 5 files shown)


node/src/Pos/Client/Txp/Util.hs
```diff
@@ -74,19 +74,27 @@ data TxRaw = TxRaw
     -- ^ Remaining money
     }
 
-data TxError
-    = TxError { unTxError :: !Text }
+data TxError =
+      NotEnoughMoney !Coin    -- ^ Parameter: how much more money is needed
+    | FailedToStabilize !Int  -- ^ Parameter: how many attempts were performed
+    | OutputIsRedeem !Address
+    | RedemptionDepleted
```
volhovM: Add comments to these two constructors too. And remove "Parameter", explanation is clear even without it.

node/src/Pos/Client/Txp/Util.hs
```diff
@@ -74,19 +74,27 @@ data TxRaw = TxRaw
...
-    => Text -> m a
-throwTxError = throwError . TxError
+    build (NotEnoughMoney coin) =
+        bprint ("Transaction creation error: not enough money, need "%build%" more") coin
+    build (FailedToStabilize iters) =
+        bprint ("Transaction creation error: failed to stabilize fee after "%build%" iterations") iters
+    build (OutputIsRedeem addr) =
+        bprint ("Destination address "%build%" is a redeem address") addr
+    build RedemptionDepleted =
+        bprint "Redeem balance is 0"
```
volhovM: Eh, what does it meen? "Redeem address balance" or what?

gromakovsky:
Please ping me once it stops being WIP.

wallet/src/Pos/Wallet/Web/Methods/Txp.hs
```diff
@@ -31,8 +31,12 @@ rewrapTxError
     :: forall m a. MonadCatch m
     => Text -> m a -> m a
 rewrapTxError prefix =
-    rewrapToWalletError (\SomeException{} -> True) (InternalError . sbuild) .
-    rewrapToWalletError (\TxError{} -> True) (RequestError . sbuild)
+    rewrapToWalletError (\SomeException{}     -> True) (InternalError . sbuild) .
+    rewrapToWalletError (\NotEnoughMoney{}    -> True) (RequestError . sbuild) .
+    rewrapToWalletError (\FailedToStabilize{} -> True) (RequestError . sbuild) .
```
Martoon-00: It doesn't work like you may expect.
E.g. `Nothing & \Just 5 -> ()` would throw error.
Here you have to specify exception type without mentioning constructors

Pastafarianist: @Martoon-00 how would you suggest doing that?

node/src/Pos/Generator/Block/Payload.hs
```diff
@@ -212,7 +211,7 @@ genTxPayload = do
 
         eTx <- lift . lift $
             createGenericTx makeTestTx ownUtxo txOutAuxs changeAddrData
-        (txAux, _) <- either (throwM . BGFailedToCreate . unTxError) pure eTx
+        (txAux, _) <- either (throwM . BGFailedToCreate . pretty) pure eTx
```
gromakovsky: I think it would be better to put `TxError` into `BGFailedToCreate` instead, but it's minor.

wallet/src/Pos/Wallet/Web/Methods/Txp.hs
Martoon-00: Something like
`rewrapToWalletError @TxError (\(e :: TxError) -> True) (...)`
or perhaps more prettier
`rewrapToWalletError @TxError (const True) (...)`

node/src/Pos/Client/Txp/Util.hs
Pastafarianist: @volhovM I agree that the word "parameter" in this context is ugly, but I don't think it will be clearer without it. Do you have any better suggestions?

node/src/Pos/Client/Txp/Util.hs
Pastafarianist: Fixed.

node/src/Pos/Client/Txp/Util.hs
volhovM: It's perfectly clear without it.

node/src/Pos/Generator/Block/Payload.hs
Pastafarianist: There are two instances where `BGFailedToCreate` is used, so I think it will be easier to leave this as-is now.

gromakovsky:
@Martoon-00 can you approve if everything is ok now?

PR #1443, opened 2017-08-28
scripts/generate/genesis.sh:  some more order in the business

deepfire:
This allows better automation in `iohk-ops`.  Been tested in the last week's three deployments.


scripts/generate/genesis.sh
```diff
@@ -20,26 +43,54 @@ mkdir $outputDir
...
-if [[ "$N" == "" ]]; then 
+if [[ "$N" == "" ]]; then
     echo "You didn't set N (poor keys amount)";
     exit 1
 fi
 
-F=100 # fake avvm keys
+case "${BUILD_MODE}" in
+nix )
+        IOHKOPS_NIX=${IOHKOPS_DIR}/default.nix
```
domenkozar: This should now use expressions from the cardano-sl/default.nix

domenkozar:
This also depends on DEVOPS-255 (for example there is no more qanet)

georgeee:
Plus regenerated genesises on top of current master

domenkozar:
https://hydra.iohk.io/eval/62859

PR #1444, opened 2017-08-29
[CSL-1572] Add tests to 'PureToss'

rockbmb:


node/test/Test/Pos/Ssc/GodTossing/Toss/PureSpec.hs
```diff
@@ -0,0 +1,100 @@
...
+actionToMonad (DelCommitment sid)  = Toss.delCommitment sid
+actionToMonad (DelOpening sid)     = Toss.delOpening sid
+actionToMonad (DelShares sid)      = Toss.delShares sid
+actionToMonad (SetEpochOrSlot eos) = Toss.setEpochOrSlot eos
+
+emptyTossSt :: Toss.GtGlobalState
+emptyTossSt = def
+
+perform :: HasCoreConstants => [TossAction] -> Toss.PureToss ()
+perform = foldl (>>) (return ()) . map actionToMonad
```
neongreen: Isn't this `mapM actionToMonad`?

node/test/Test/Pos/Ssc/GodTossing/Toss/PureSpec.hs
```diff
@@ -0,0 +1,100 @@
...
+
+instance HasCoreConstants => Arbitrary TossAction where
+    arbitrary = genericArbitrary
+    shrink = genericShrink
+
+actionToMonad :: Toss.MonadToss m => TossAction -> m ()
+actionToMonad (PutCommitment sc)   = Toss.putCommitment sc
+actionToMonad (PutOpening sid o)   = Toss.putOpening sid o
+actionToMonad (PutShares sid ism)  = Toss.putShares sid ism
+actionToMonad (PutCertificate v)  = Toss.putCertificate v
```
neongreen: One more space

node/test/Test/Pos/Ssc/GodTossing/Toss/PureSpec.hs
```diff
@@ -0,0 +1,100 @@
...
+
+emptyTossSt :: Toss.GtGlobalState
+emptyTossSt = def
+
+perform :: HasCoreConstants => [TossAction] -> Toss.PureToss ()
+perform = foldl (>>) (return ()) . map actionToMonad
+
+type TossTestInfo = (Word64, Word64, Word64, Word64, Word64)
+
+(==^) :: HasCoreConstants => [TossAction] -> [TossAction] -> TossTestInfo -> Property
```
neongreen: Write a haddock so that people would be able to understand what it does without having to look at the code.

node/test/Test/Pos/Ssc/GodTossing/Toss/PureSpec.hs
georgeee: :D

node/test/Test/Pos/Ssc/GodTossing/Toss/PureSpec.hs
rockbmb: I was going to refactor this and add a typeclass to unify these tests and `PurePoll` ones, and as such the comments would be the same, but seeing as it might take a while until that happens, I added the comments you mentioned.

node/test/Test/Pos/Ssc/GodTossing/Toss/PureSpec.hs
rockbmb: `mapM_`, forgot the regular version does not discard results.

gromakovsky:
@rockbmb I want to do something with this PR. Can you squash your commits (because 3 commits basically fix the first one) and rebase on master?

rockbmb:
@gromakovsky Oh, I thought this email was about the `rocksdb-haskell` PR when I saw it on my phone.
Yes, I can.

rockbmb:
@gromakovsky Done 👍
Also, let me know if what you're doing is something I can help with.

rockbmb:
@gromakovsky it appears adding this set of tests causes Travis to hang. In any case, I would need to temporarily disable them, were this PR to be merged immediately, because they aren't entirely correct (there is an explanation why here: https://github.com/input-output-hk/cardano-sl/blob/master/lib/test/Test/Pos/Update/PollSpec.hs#L208)

I'll fix this problem, and then I'll deactivate them for now.

gromakovsky:
@rockbmb if you're talking about
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

then it's not because of your tests, we have this problem in other branches, too.

rockbmb:
@gromakovsky if that's the case, then I won't disable them, and tests are fixed now.

PR #1449, opened 2017-08-29
[CSL-1573] Change deployment formats in constants.yaml

volhovM:


georgeee:
Update CI configs

volhovM:
I think i should also regenerate genesis to make things better, but i'm not sure i know the default param to launch `generate/genesis.sh`.

scripts/generate/genesis.sh
```diff
@@ -2,9 +2,13 @@
 
 # This script generates genesis files.
 # Launch it from the project root directory.
-# It will create a directory with all needed data inside. It will not
-# copy generated genesis files into the project, this is to be done
-# manually.
+#
+# It will create a directory with all needed data inside. 
+#
+# (!) Make sure you compile project with a appropriate mode
```
gromakovsky: an

core/constants.yaml
```diff
@@ -114,95 +130,98 @@ dev: &dev
...
+  networkWaitLogInterval: 32 # 32s
 
-  genesisBinSuffix: tn
-
+  protocolMagic: 60987900
+  genesisBinSuffix: tn 
 
-testnet_staging: &testnet_staging
-  <<: *testnet
+testnet_staging_full: &testnet_staging_full
```
domenkozar: @deepfire FYI

PR #1451, opened 2017-08-29
[CSL-1340] Refine reporting

gromakovsky:
It is WIP, but almost done. I decided to drop few lines about this PR while I remember them very well.
This work was partially performed by @georgeee, first three commits are by him. The rest are mine.
Overview of things done:
1. Electron crash endpoint was deleted by @georgeee, ask him about details if you don't like it.
2. Functions in `Pos.Reporting` were refactored, now it provides simpler interface, basically 3 `report*` functions (where `*` is severity). Also some lower-level functions are exported to be used by launcher.
3. `REPORT:XX` comments were added. You can grep for `REPORT` to see all places where we report something.
4. Rollback is reported depending on depth (deep rollback is considered critical, short rollback is not).
5. I was mostly reviewing code and thinking (also discussed some things in best practice channel and with Misha), during that time I made some small changes.

TODO
1. I want to add few more reporting usages. See [this comment](https://issues.serokell.io/issue/CSL-1340#comment=96-6908).
2. `reportNode` shouldn't catch async exceptions. I will wait a little, maybe my PR to `universum` will be merged, then I will switch to new version (which should also improve few more places), otherwise I will change that place.
3. We still need to do something with `reportingFatal`. The idea is that it should catch some types of errors everywhere. We need to use it wherever we fork a thread, because exception handler doesn't apply to forked threads. It is used in many places, but most likely not in all places and somewhere can be used incorrectly.
I think the proper solution is to get rid of `fork` usages in code (as was suggested by FPC) and then use `reportingFatal` in only one place. Note that this work was started [already](https://github.com/input-output-hk/cardano-sl/pull/1163).
Alternatively we can insert `reportingFatal` in proper places, but it won't be maintained and will be spoiled eventually.
Anyway, I don't want to do it in this PR, I will create another issue about it. This PR already contains useful things.


ssc/Pos/Ssc/Class/Helpers.hs
```diff
@@ -23,3 +23,5 @@ class Ssc ssc => SscHelpersClass ssc where
     sscStripPayload :: Byte -> SscPayload ssc -> Maybe (SscPayload ssc)
     -- | Returns default payload for the given local slot index.
     sscDefaultPayload :: LocalSlotIndex -> (SscPayload ssc)
+    -- | Returns 'True' if the error must be reported.
+    sscIsCriticalError :: SscVerifyError ssc -> Bool
```
volhovM: Super cancer :(
I'd better have some default typeclass for "reportable" errors rather than this particular method in... helpers.

Anyway.

PR #1452, opened 2017-08-29
[CSL-1580] Run demo with 4 nodes by default

int-index:


scripts/launch/demo.sh
```diff
@@ -17,7 +17,7 @@ if [[ ! -d "$base/../../.stack-work" ]]; then
 fi
 
 # Define the default amount of nodes to run
-DEFAULT_NODES_N=3
+DEFAULT_NODES_N=4
```
volhovM: can you throw an error if `N /= 4` ? Just in case, because it doesn't work otherwise.

volhovM:
Only script-related changes, merging.

scripts/launch/demo.sh
gromakovsky: What exactly doesn't work?

scripts/launch/demo.sh
```diff
@@ -17,14 +17,19 @@ if [[ ! -d "$base/../../.stack-work" ]]; then
...
 
 # Detect node number or use default
 n=$1
 if [[ "$n" == "" ]]; then
   n=$DEFAULT_NODES_N
 fi
 
+if [[ "$n" != 4 ]]; then
+  echo "Currently we don't support any amount of nodes in demo that's not 4. See CSL-1580"
```
gromakovsky: Weird :/

scripts/launch/demo.sh
```diff
@@ -17,14 +17,19 @@ if [[ ! -d "$base/../../.stack-work" ]]; then
...
+  exit
```
gromakovsky: Revert it please. It shouldn't exit in this case.

volhovM: ok

volhovM: `76874b40f4ff723cc5274fe88a2acd0ba7cb877f`

PR #1453, opened 2017-08-29
[CSL-1203] Add the addr-distr command to lwallet

int-index:


lwallet/Command.hs
```diff
@@ -156,6 +163,33 @@ proposeUpdate =
     lexeme parseSoftwareVersion <*>
     many1 parseProposeUpdateSystem
 
+stakeholderIdP :: Parser StakeholderId
+stakeholderIdP = hash
```
gromakovsky: `stakeholderIdParser` exists in `Pos.Client.CLI.Util`.
I don't like the idea of having multiple ways to parse it. Even if you don't like that definition, let's agree on one of them somehow.

lwallet/Command.hs
```diff
@@ -156,6 +163,33 @@ proposeUpdate =
...
+
+coinPortionP :: Parser CoinPortion
+coinPortionP = do
+    (token, modifier) <- anyText <&> \s -> case Text.stripSuffix "%" s of
+        Nothing -> (s, identity)
+        Just s' -> (s', (/100))
+    case readMaybe (toString token) of
+        Just (modifier -> a) | a >= 0, a <= 1 -> return $ unsafeCoinPortionFromDouble a
```
gromakovsky: It would be good to write explicit type at least in one place to make it clearer which numeric type is used. There are many possibilities, any of them should be ok (type application, inline type like `token :: Double`, separate type declaration).

lwallet/Command.hs
```diff
@@ -156,6 +163,33 @@ proposeUpdate =
...
+
+addrStakeDistrP :: Parser AddrStakeDistribution
+addrStakeDistrP =
+    BootstrapEraDistr <$ text "boot" <|>
+    multiKeyDistrP
+  where
+    multiKeyDistrP = do
+        parts <- many1 ((,) <$> stakeholderIdP <* text ":" <*> coinPortionP)
+        case parts of
+            [(sId, coinPortion)] | coinPortion == unsafeCoinPortionFromDouble 1 ->
```
gromakovsky: `unsafeCoinPortionFromDouble 1` is `maxBound`, so it might be better to use it here to avoid a partial function.

lwallet/Main.hs
```diff
@@ -315,8 +325,11 @@ runCmd _ ListAddresses _ = do
    putText "Available addresses:"
    for_ (zip [0 :: Int ..] addrs) $ \(i, pk) -> do
        addr <- makePubKeyAddressLWallet pk
-       putText $ sformat ("    #"%int%":   "%build%" (PK: "%stext%")")
-                    i addr (toBase58Text pk)
+       putText $ sformat ("    #"%int%":   addr:      "%build%"\n"%
+                          "          pk base58: "%stext%"\n"%
+                          "          pk hex:    "%fullPublicKeyHexF%"\n"%
+                          "          pk hash:   "%hashHexF)
```
gromakovsky: Would be good to also write that it's called stakeholder id (`pk hash (stakeholder id): bla`).

lwallet/Main.hs
```diff
@@ -341,6 +354,10 @@ runCmd _ (AddKeyFromPool i) CmdCtx{..} = do
 runCmd _ (AddKeyFromFile f) _ = do
     secret <- readUserSecret f
     mapM_ addSecretKey $ secret ^. usKeys
+runCmd _ (AddrDistr pk asd) _ = do
+    putText $ sformat build addr
```
gromakovsky: `sformat build` is `pretty`.

lwallet/Main.hs
```diff
@@ -128,6 +131,13 @@ Avaliable commands:
                                      e is current epoch.
    add-key-pool <N>               -- add key from intial pool
    add-key <file>                 -- add key from file
+
+   addr-distr <N> boot
+   addr-distr <N> [<M>:<coinPortion>]+
+                                  -- address of pk <N> (encoded in base58) with provided distribution, where <M> is
+                                  -- stakeholder id (pk hash), and the coin portion can be a coefficient in [0..1]
+                                  -- or a percentage (ex. 42%)
```
gromakovsky: It's not clear what this command does. Neither from name nor from description. I had no idea until I read the code with its behaviour. Now I see that it **prints** the address and that's it.
Other commands' names are verbs, so it's clearer what they do.

lwallet/Main.hs
int-index: At first I did it, but it breaks the alignment.

lwallet/Command.hs
int-index: The definition there won't work in my case because it will consume the colon which I use as a separator.

PR #1454, opened 2017-08-30
CSE-203 Use SlotId as timestamp for transactions. CSL + Explorer comm…

ksaric:
https://issues.serokell.io/issue/CSE-203
https://issues.serokell.io/issue/CSE-197

Fixes the two issues above. Tested with _dev_ mode.


node/src/Pos/Explorer/Txp/Global.hs
```diff
@@ -52,14 +53,16 @@ extraOps (ExplorerExtra em (HM.toList -> histories) balances) =
...
 applyBlund
     :: (MonadSlots ctx m, EGlobalApplyToilMode ctx m)
     => TxpBlund
     -> m ()
 applyBlund blund = do
-    curTime <- currentTimeSlotting
-    uncurry (eApplyToil curTime) $ blundToAuxNUndoWHash blund
+    -- First get the current @SlotId@ so we can calculate the time.
+    -- Then get when that @SlotId@ started and use that as a time for @Tx@.
+    mTxTimestamp <- getCurrentSlotBlocking >>= getSlotStart
```
gromakovsky: No, it's wrong. You should take a slot from `TxpBlund`.
```
-- [CSL-1156] Maybe find better approach (at least wrap into normal types).
type TxpBlock = Either (Some IsGenesisHeader) (Some IsMainHeader, TxPayload)
type TxpBlund = (TxpBlock, TxpUndo)
```

`TxpBlund` has `TxpBlock` inside. If it's `Left`, it's a genesis block which doesn't contain transactions. It doesn't have a slot, only epoch, but you can use e. g. `SlotId epoch minBound`. If it's `Right`, you can use `headerSlotL` lens.

node/src/Pos/Explorer/Txp/Local.hs
```diff
@@ -126,7 +129,10 @@ eTxProcessTransactionNoLock itw@(txId, txAux) = runExceptT $ do
             M.fromList $
             catMaybes $
             toList $ NE.zipWith (liftM2 (,) . Just) _txInputs resolvedOuts
-    curTime <- currentTimeSlotting
+    -- First get the current @SlotId@ so we can calculate the time.
+    -- Then get when that @SlotId@ started and use that as a time for @Tx@.
+    mTxTimestamp <- getCurrentSlotBlocking >>= getSlotStart
```
gromakovsky: You can avoid using `getCurrentSlotBlocking` here. This function shouldn't block if current slot is unknown. The reason it will work is that there is
```
    epoch <- siEpoch <$> (note ToilSlotUnknown =<< getCurrentSlot)
```
above. So if we don't know current slot, it will finish earlier. I would recommend removing this line (which makes impression that this line can block) and write:
```
  slot <- (note ToilSlotUnknown =<< getCurrentSlot)
  let epoch = siEpoch slot
  mTxTimestamp <- getSlotStart slot
```
It can also be slightly more efficient.

node/src/Pos/Explorer/Txp/Global.hs
ksaric: I didn't even know that a transaction has a slot :grin: 
Ok, thanks, will fix.

node/src/Pos/Explorer/Txp/Local.hs
ksaric: Ok, will fix, thanks.

node/src/Pos/Explorer/Txp/Global.hs
gromakovsky: No, not transaction, it's a block.
`TxpBlock` is just like `Block`, but contains reduced set of information (only what is needed for tx processing).

ksaric:
@gromakovsky Changing `currentTimeSlotting = currentTimeSlottingSimple` and running tests with my previous changes seems to fail:
```
    Failures:
    
      test/Test/Pos/Block/Logic/VarSpec.hs:314: 
      1) Test.Pos.Block.Logic.Var.Block.Logic.VAR.Block.Event, Apply through epoch, apply a sequence of blocks that spans through epochs (additional blocks after crossing: 0)
           uncaught exception: CardanoException (Cardano fatal error: getCurrentSlotBlocking is used in generator when slot is unknown) (after 1 test)
           TestParams {
             utxo = {TxInUtxo 482515c0 #0: {txout = TxOut 24666136666890730 coin(s) -> LaVWd6GZcFCNMKensCLiB6kiBcoJKfkMhtSyKLwLMeZ1PRVTpeYQ}}
             secrets: AllSecrets {
             secret keys: 1 items
             stakeholders: [85097b10]
             inverse spending data: {LaVWd6GZcFCNMKensCLiB6kiBcoJKfkMhtSyKLwLMeZ1PRVTpeYQ: PubKeyASD pub:9fea6627}
           }
           
             stake distributions: [FlatStakes 1 (Coin {getCoin = 24666136666890727})]
             start time: 0mcs
           }
    
      test/Test/Pos/Block/Logic/VarSpec.hs:314: 
      2) Test.Pos.Block.Logic.Var.Block.Logic.VAR.Block.Event, Apply through epoch, apply a sequence of blocks that spans through epochs (additional blocks after crossing: 4)
           uncaught exception: CardanoException (Cardano fatal error: getCurrentSlotBlocking is used in generator when slot is unknown) (after 1 test)
           TestParams {
             utxo = {TxInUtxo 482515c0 #0: {txout = TxOut 24666136666890730 coin(s) -> LaVWd6GZcFCNMKensCLiB6kiBcoJKfkMhtSyKLwLMeZ1PRVTpeYQ}}
             secrets: AllSecrets {
             secret keys: 1 items
             stakeholders: [85097b10]
             inverse spending data: {LaVWd6GZcFCNMKensCLiB6kiBcoJKfkMhtSyKLwLMeZ1PRVTpeYQ: PubKeyASD pub:9fea6627}
           }
           
             stake distributions: [FlatStakes 1 (Coin {getCoin = 24666136666890727})]
             start time: 0mcs
           }
    
    Randomized with seed 1769033029
    
    Finished in 1632.3634 seconds
    578 examples, 2 failures, 31 pending


```

Fixing the things you requested and then running tests seems to work. Tried it three times. 
BTW, we actually don't have a `minBound` for `SlotId`. I think we could add it, I don't know how to calculate it from the constants (I see slot duration, but I don't see any slot number in epoch/epoch length). For now, I just used a 0.

node/src/Pos/Explorer/Txp/Global.hs
```diff
@@ -52,14 +55,33 @@ extraOps (ExplorerExtra em (HM.toList -> histories) balances) =
...
+    -- use e. g. SlotId epoch minBound. If it's Right, you can use headerSlotL lens.
+    --
+    -- type TxpBlund = (TxpBlock, TxpUndo)
+    -- type TxpBlock = Either (Some IsGenesisHeader) (Some IsMainHeader, TxPayload)
+
+    let txpBlock = txpBlund ^. _1
+    let slotId   = case txpBlock of
+            Left gensisBlock -> SlotId
+                                  { siEpoch = gensisBlock ^. epochIndexL
+                                  , siSlot  = UnsafeLocalSlotIndex 0
```
gromakovsky: > BTW, we actually don't have a minBound for SlotId.

@ksaric we have `minBound` for `LocalSlotIndex`: https://github.com/input-output-hk/cardano-sl/blob/ee50e1d41cd99082dbf3e030e24745e74eefa66e/core/Pos/Core/Slotting.hs#L154

It's better to use it here instead of `UnsafeLocalSlotIndex`, because `UnsafeLocalSlotIndex` is, well, unsafe.
Of course we know that it's safe here, because its argument is always 0, but anyway, when I see a function prefixed with `Unsafe`, I need to think whether it's safe.

ksaric: Ah, I see, that's what you meant. :man_facepalming: 
Fixed.

node/test/Test/Pos/Network/Windows/DnsDomainsSpec.hs
```diff
@@ -16,19 +16,19 @@
 -- | Specification of 'Pos.Network.Windows.DnsDomains'
 
 module Test.Pos.Network.Windows.DnsDomainsSpec
```
gromakovsky: I want @rockbmb or @adinapoli-iohk to review changes in this module.
Or just revert these changes, I don't understand why this PR touches this module.

adinapoli-iohk: @gromakovsky from a quick eyeball it seems like they are all cosmetic changes to please whatever linter was in scope when writing this code 😉 . At glance, the changes:

* Moves the `if !defined(POSIX)` upper to include modules that are used only when running the spec on windows (otherwise a "redundant import" message would be shown);

* Moves the position of the `prop_GetWindowsDefDnsServer`;

I cannot check this on my machine as I don't use Windows, but surely AppVeyor should reject this code if we made a mistake, so it's 👍  on my part.

gromakovsky: Ok, @ksaric, then please just check that these tests are indeed launched by appveyor and were not accidentally disabled.

ksaric: @gromakovsky They were launched, and that's why they failed, that's why I updated the test. 
I hope it works this time.

PR #1456, opened 2017-08-30
Update script for connecting to staging

flyingleafe:
As file `run/topology.yaml` file is under gitignore, I'll past it here for convenience:

```
wallet:
  relays: [[{"addr": "52.59.160.100", "port": 3000}, {"addr": "52.58.254.151", "port": 3000}]]
```


scripts/launch/staging.sh
```diff
@@ -1,10 +1,14 @@
...
 
 stack exec -- cardano-node \
-    --system-start 1499246772 \
+    --system-start 1504019546 \
     --log-config log-config-prod.yaml \
     --logs-prefix "logs/qanet" \
+    --keyfile secret.key \
     --db-path db-qanet \
     --wallet-db-path wdb-qanet \
-    --no-ntp \
```
gromakovsky: Can you justify why you decided to enable NTP? I recall it was quite unreliable.

gromakovsky: Is no-ntp the default now? 🤔

PR #1460, opened 2017-08-30
[CSM-415] Add 'ctIsLocal' flag

Martoon-00:
Added a flag to `CTx` which indicates, whether given transaction is local i.e. transfers money within same wallet


flyingleafe:
Please check that tx history is working both in dev mode and staging. Also run DAT in both cases, but bear in mind that there might still be some inconsistencies in DAT with tx fees amount (it was going to be fixed, but I'm not sure if it's merged yet)

Martoon-00:
Tested manually in dev mod and staging, but failed to launch DAT

PR #1464, opened 2017-08-31
[CSL-1585] Namespace cardano EKG metrics

adinapoli-iohk:
This PR correctly namespace IOHK's metrics by the "cardano" namespace, for which I have added a smart "constructor" inside a new module called `Pos.System.Metrics.Constants` within `core`.

Note this PR is not complete as we need to modify `timewarp-nt` so that `registerQueueMetrics` can take an optional namespace (will do this in parallel):

https://github.com/serokell/time-warp-nt/blob/40f4235cda65c746c6edc2a3455242d96d7175bf/src/Network/Broadcast/OutboundQueue.hs#L607

Last but not least, now the time-based queues are parametrised by their unit of measures, microseconds in `txp` case. (maybe one of the devops like @deepfire or @domenkozar  could tell us if `statsd` will choke if non-ASCII chars are passed to it). This is the result so far:

screen shot 2017-08-31 at 12 42 10

### DIFF
--- a/core/Pos/System/Metrics/Constants.hs
+++ b/core/Pos/System/Metrics/Constants.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Pos.System.Metrics.Constants (
-    withCardanoNamespace
+      cardanoNamespace
+    , withCardanoNamespace
                                         ) where
 
 import           Universum
--- a/core/Pos/System/Metrics/Constants.hs
+++ b/core/Pos/System/Metrics/Constants.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE OverloadedStrings #-}
 module Pos.System.Metrics.Constants (
       cardanoNamespace
     , withCardanoNamespace
--- a/core/Pos/System/Metrics/Constants.hs
+++ b/core/Pos/System/Metrics/Constants.hs
@@ -1,0 +1,12 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Pos.System.Metrics.Constants (
+    withCardanoNamespace
+                                        ) where
+
+import           Universum
+
+cardanoNamespace :: Text
+cardanoNamespace = "cardano"
+
+withCardanoNamespace :: Text -> Text
+withCardanoNamespace label = cardanoNamespace <> "." <> label
--- a/infra/Pos/Network/Types.hs
+++ b/infra/Pos/Network/Types.hs
@@ -1,5 +1,6 @@
+{-# LANGUAGE CPP                       #-}
 {-# LANGUAGE ExistentialQuantification #-}
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedStrings         #-}
 
 #if !defined(mingw32_HOST_OS)
 #define POSIX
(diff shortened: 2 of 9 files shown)


adinapoli-iohk:
Now we have namespaced metrics for everything IOHK related, with units of measure:

screen shot 2017-08-31 at 15 53 04

domenkozar:
I suggest we use ASCII just to be sure - it's quicker than verifying this :)

adinapoli-iohk:
> I suggest we use ASCII just to be sure - it's quicker than verifying this :)

Haha, I was sure you guys were going to prefer the safe route 😁 . I think we are fine though.

StatsD spec explicitly says tags can be UTF-8:

https://github.com/b/statsd_spec#metric-types--formats

Th DD agent should support utf-8 as well:

https://github.com/DataDog/datadogpy/pull/152

But if you guys don't feel confident anyway I'm happy to go with ASCII 😀

domenkozar:
We can either take the risk of figuring out only on next deploy or spend a few minutes taking the safe route :) Both are fine with me though - my experience with python shows that few programmers write code that works well with non-ASCII encodings.

adinapoli-iohk:
> We can either take the risk of figuring out only on next deploy or spend a few minutes taking the safe route :) 

I won't be the guy responsible for degrading the velocity of the project 😀  Safe route will be! Fix incoming.

dcoutts:
> This is currently using a non-master, non-upstream revision of timewarp-nt as my outstanding PR needs to be merged first.

The TW PR is merged, you can update.

core/Pos/System/Metrics/Constants.hs
```diff
@@ -0,0 +1,12 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Pos.System.Metrics.Constants (
+    withCardanoNamespace
+                                        ) where
```
dcoutts: Is this spacing in our style guide? :-)

core/Pos/System/Metrics/Constants.hs
```diff
@@ -0,0 +1,13 @@
+{-# LANGUAGE OverloadedStrings #-}
```
gromakovsky: This extension (among [many others](https://github.com/input-output-hk/cardano-sl/blob/master/core/cardano-sl-core.cabal#L189)) is enabled by default.

adinapoli-iohk: Ah, I must have been taken in by flycheck or something 😉

node/src/Pos/Launcher/Runner.hs
```diff
@@ -122,8 +124,8 @@ runRealModeDo NodeResources {..} outSpecs action =
         case npEnableMetrics of
             False -> return Nothing
             True  -> Just <$> do
-                ekgStore' <- setupMonitor
-                    (runProduction . runToProd JsonLogDisabled oq) node' nrEkgStore
+                let ekgStore' = nrEkgStore
+                registerMetrics (Just cardanoNamespace) (runProduction . runToProd JsonLogDisabled oq) node' ekgStore'
```
dcoutts: You don't need `ekgStore'` any more. It was only there because the old `setupMonitor` returned a possibly modified store, and no it doesn't, so there's no need to name the result.

infra/Pos/Network/Types.hs
```diff
@@ -1,5 +1,6 @@
+{-# LANGUAGE CPP                       #-}
 {-# LANGUAGE ExistentialQuantification #-}
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedStrings         #-}
```
gromakovsky: In `infra` (and I guess in all other packages) this extension is enabled by default too.

node/src/Pos/Util/Monitor.hs
```diff
@@ -0,0 +1,47 @@
+{-# OPTIONS_GHC -fno-warn-name-shadowing #-}
```
gromakovsky: I find it much better to avoid name shadowing than to suppress warnings. If there are fundamental reasons why it can't be avoided, it should be justified in comments.

node/src/Pos/Util/Monitor.hs
```diff
@@ -0,0 +1,47 @@
...
+      startMonitor
+    , stopMonitor
+
+    ) where
+
+import           Control.Concurrent           (killThread)
+import           Control.Monad.IO.Class
+import           Mockable.Class
+import qualified Mockable.Metrics             as Metrics
+import           Node
```
gromakovsky: Imports should be qualified or contain explicit import list, according to [style-guide](https://github.com/input-output-hk/cardano-sl-style-guides/blob/master/haskell-style-guide.md#imports).

PR #1467, opened 2017-08-31
[CSLD-108] How to connect node to the cluster

denisshevchenko:


deepfire:
@denisshevchenko, @domenkozar, this raises an important question -- do we want to support the `stack` way of building `cardano-sl` for the external community?

This is an important decision with far-reaching consequences, AFAICT.

domenkozar:
Deveopers use stack, I think that's fine.

denisshevchenko:
@domenkozar Agree. This script is for developers, so `stack exec` looks good here.

gromakovsky:
These files are ignored by CI ⇒ merging.

PR #1468, opened 2017-08-31
[CSL-1542] Remove unused imports warning

rockbmb:


node/test/Test/Pos/Network/Windows/DnsDomainsSpec.hs
```diff
@@ -5,6 +5,8 @@
 #define POSIX
 #endif
 
+{-# OPTIONS_GHC -Wno-unused-imports   #-}
```
gromakovsky: Can you also write a comment explaining why this warning is disabled?

PR #1470, opened 2017-09-01
[CSL-1559] Refactor locking in delegation

volhovM:
This uses newly introduced global `StateLock` instead of old (dlg-specific) synchronization mechanism.

### DIFF
--- a/node/src/Pos/Delegation/Logic/Common.hs
+++ b/node/src/Pos/Delegation/Logic/Common.hs
@@ -29,8 +29,8 @@ import           Formatting           (bprint, build, sformat, stext, (%))
 
 import           Pos.Constants        (dlgCacheParam, lightDlgConfirmationTimeout,
                                        messageCacheTimeout)
-import           Pos.Core             (ProxySKHeavy, StakeholderId, addressHash,
-                                       epochIndexL)
+import           Pos.Core             (HasCoreConstants, ProxySKHeavy, StakeholderId,
+                                       addressHash, headerHash)
 import           Pos.Crypto           (ProxySecretKey (..), PublicKey)
 import           Pos.DB               (DBError (DBMalformed), MonadDBRead)
 import qualified Pos.DB.Block         as DB
@@ -69,8 +69,6 @@ instance B.Buildable DelegationError where
 -- | Convenient monad to work in 'DelegationWrap' state context.
 type DelegationStateAction = State DelegationWrap
 
--- Misha knows better probably.
--- {-# ANN runDelegationStateAction ("HLint: ignore Use fmap" :: Text) #-}
 -- | Executes atomic action on delegation variable.
 runDelegationStateAction
     :: (MonadIO m, MonadMask m, MonadDelegation ctx m)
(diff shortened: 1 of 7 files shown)


node/src/Pos/Delegation/Logic/Common.hs
```diff
@@ -68,22 +67,24 @@ instance B.Buildable DelegationError where
...
 -- | Convenient monad to work in 'DelegationWrap' state context.
-type DelegationStateAction m = StateT DelegationWrap m
+type DelegationStateAction = State DelegationWrap
 
 -- Misha knows better probably.
-{-# ANN runDelegationStateAction ("HLint: ignore Use fmap" :: Text) #-}
--- | Effectively takes a lock on ProxyCaches mvar in NodeContext and
--- allows you to run some computation producing updated ProxyCaches
--- and return value. Will put MVar back on exception.
+-- {-# ANN runDelegationStateAction ("HLint: ignore Use fmap" :: Text) #-}
```
gromakovsky: Perhaps no longer needed, but I am not sure.

node/src/Pos/Delegation/Logic/Common.hs
```diff
@@ -101,17 +102,17 @@ invalidateProxyCaches curTime = do
 -- * Sets '_dwEpochId' to epoch of tip.
 -- * Initializes mempools/LRU caches.
 mkDelegationVar ::
-       forall ssc m. (MonadIO m, DB.MonadBlockDB ssc m)
+       forall ssc m. (MonadIO m, DB.MonadBlockDB ssc m, HasCoreConstants)
     => m DelegationVar
 mkDelegationVar = do
     tip <- DB.getTipHeader @ssc
-    RWV.new
+    liftIO $ newTVarIO
```
gromakovsky: `liftIO` is most likely redundant.

node/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -556,23 +560,33 @@ dlgRollbackBlocks blunds = do
...
     tip <- DB.getTipHeader @ssc
     oldPool <- runDelegationStateAction $ do
-        dwEpochId .= (tip ^. epochIndexL)
         pool <- uses dwProxySKPool toList
         dwProxySKPool .= mempty
         pure pool
-    forM_ oldPool (processProxySKHeavy @ssc)
+    forM_ oldPool $ \psk -> do
+        res <- processProxySKHeavyInternal @ssc psk
+        -- should we throw error here also?
```
gromakovsky: No, we shouldn't throw error. Why do we even log it as error? If we have a PSK in mempool and it becomes invalid after rollback, it's absolutely legal situation, we can ignore it. That's what normalization does.

node/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -556,23 +560,33 @@ dlgRollbackBlocks blunds = do
...
+        when (res /= PHAdded) $ logError $
+            sformat ("dlgNormalizeOnRollback: got psk verdict "%shown%" for psk "%build)
+                    res
+                    psk
+    runDelegationStateAction $ dwTip .= headerHash tip
```
gromakovsky: All this logic looks incorrect. When we call `dlgNormalizeOnRollback`, we have already updated tip in DB, but haven't updated it in mempool. So I expect `processProxySKHeavyInternal` to always fail due to tip mismatch.
You should update mempool and set tip inside single atomic STM action, otherwise it can break invariants.

volhovM:
@gromakovsky I've changed the last commit and fixed errors you pointed on. Please re-review!

PR #1471, opened 2017-09-01
[DEVOPS-302] release.nix: simplify and allow setting the platform

domenkozar:
- platform is now an attribute suffix
- explorer and report server are now built only for linux


domenkozar:
https://hydra.iohk.io/jobset/serokell/cardano-pr-1471#tabs-jobs

PR #1472, opened 2017-09-01
[CSL-1579] Revise 'reportingFatal'

gromakovsky:
The idea behind `reportingFatal` is that we want to catch some types of exceptions every time they happen and send to report server. I did few changes:
1. I removed `reportingFatal` from functions like block creation which don't have any forking or exception handling inside and aren't special at all with respect to thread management and exception handling. Having this logic there isn't useful at all, it doesn't help us to achieve the main goal: catch fatal errors everywhere.
2. After (1) all usages of `reportingFatal` started having the following pattern:
```
reportingFatal action `catchAll` handler
```
So we have a handler inside `reportingFatal` which rethrows an exception and this exception is then caught by `handler` (which catches everything).
I find it more reasonable to have one exception handler and use `reportOrLog` helper there. That's what I did.
3. I also fixed two workers which didn't restart themselves on failure. It's subjective whether they had to, but I guess they did. There is a delay before they are restarted, because restarting right after an exception is probably useless.
4. I updated `universum` version, so now most of the code which catches `SomeException` doesn't catch asynchronous exceptions, so `catchAll` is quite safe.
There is a caveat though: `Production` from `node-sketch` uses implementation from `base`, so it catches async exceptions as well.

P. S. I haven't yet dealt with listeners and I decided not to do it in this PR. Will do in subsequent one.

### DIFF
--- a/infra/Pos/Slotting/Util.hs
+++ b/infra/Pos/Slotting/Util.hs
@@ -24,20 +24,19 @@ module Pos.Slotting.Util
 import           Universum
 
 import           Data.Time.Units        (Millisecond)
-import           Formatting             (build, int, sformat, shown, (%))
+import           Formatting             (int, sformat, shown, (%))
 import           Mockable               (Delay, Fork, Mockable, delay, fork)
 import           Serokell.Util          (sec)
-import           System.Wlog            (WithLogger, logDebug, logError, logInfo,
-                                         logNotice, modifyLoggerName)
+import           System.Wlog            (WithLogger, logDebug, logInfo, logNotice,
+                                         modifyLoggerName)
 
 import           Pos.Core               (FlatSlotId, HasCoreConstants, LocalSlotIndex,
                                          SlotId (..), Timestamp (..), flattenSlotId,
                                          slotIdF)
-import           Pos.Exception          (CardanoException)
 import           Pos.KnownPeers         (MonadFormatPeers)
 import           Pos.Recovery.Info      (MonadRecoveryInfo (recoveryInProgress))
 import           Pos.Reporting.MemState (HasReportingContext)
-import           Pos.Reporting.Methods  (reportError, reportingFatal)
+import           Pos.Reporting.Methods  (reportOrLogE)
 import           Pos.Shutdown           (HasShutdownContext, runIfNotShutdown)
 import           Pos.Slotting.Class     (MonadSlots (..))
 import           Pos.Slotting.Error     (SlottingError (..))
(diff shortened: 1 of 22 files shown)


stack.yaml
```diff
@@ -44,11 +44,11 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/input-output-hk/cardano-report-server.git
-    commit: c2af07ab7d627556ed3f6185b062e4cd1fb5ad26 #master
+    commit: fa30523135060288abbf998408171b21f87f7247 # todo
```
gromakovsky: https://github.com/input-output-hk/cardano-report-server/pull/4

stack.yaml
```diff
@@ -44,11 +44,11 @@ packages:
...
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: 51ce7d0f54c6232fb0c27f2edbe25a338535d02d #master
+    commit: a88f92b1cda47b4959d7edb08f4d170c61e70ed9 # todo
```
gromakovsky: https://github.com/serokell/time-warp-nt/pull/101

cleverca22:
@gromakovsky 
https://travis-ci.org/input-output-hk/cardano-sl/jobs/270888162#L2159
hydra and nix tests will keep failing until the stack2nix data is updated and added to the PR

infra/Pos/Slotting/Util.hs
```diff
@@ -124,20 +123,20 @@ onNewSlotImpl
     :: forall ctx m. OnNewSlot ctx m
     => Bool -> Bool -> (SlotId -> m ()) -> m ()
 onNewSlotImpl withLogging startImmediately action =
-    reportingFatal impl `catch` workerHandler
+    impl `catch` workerHandler
   where
     impl = onNewSlotDo withLogging Nothing startImmediately actionWithCatch
+    -- [CSL-1578] FIXME: it is awful!!!
```
volhovM: Sorry, what exactly is awful here?

node/src/Pos/Security/Workers.hs
```diff
@@ -105,7 +105,8 @@ checkForReceivedBlocksWorkerImpl SendActions {..} = afterDelay $ do
             "by ourselves"
         reportEclipse
     repeatOnInterval delF action = runIfNotShutdown $ do
-        () <- action
+        -- REPORT:ERROR 'reportOrLogE' in block retrieval worker.
+        () <$ action `catchAny` \e -> reportOrLogE "Security worker" e
```
volhovM: Why not `` action `catchAny` reportOrLogE "Security worker"``?

infra/Pos/Slotting/Util.hs
gromakovsky: Hm, probably I overreacted :D
The bad think is `fork` which is applied to each invocation of action (on each tick) which forces us to catch exceptions from action separately. [This comment](https://github.com/input-output-hk/cardano-sl/pull/1163#issuecomment-318136810) provides detailed overview of consequences of having `fork` there.
Action itself should determine whether it wants to be fork (one can pass `fork action` instead of `action` if it's necessary).
Probably catching exceptions from `action` separately is not that bad. So ok, I will remove this comment and will use `reportOrLogE` here (especially because nobody knows when CSL-1578 will be done).

node/src/Pos/Security/Workers.hs
gromakovsky: Dunno, probably I modified this code many times locally. Anyway, I think it's not essential.

PR #1474, opened 2017-09-02
[CSL-1551] Add genesis heavyweight delegation

gromakovsky:
1. `GenesisDelegation` type with some invariants was introduced and added to `GenesisCoreData` as well as `GenesisContext`.
2. Delegation part of GState DB is now initialized according to `GenesisDelegation`.
3. `keygen` currently always generates empty genesis delegation. We need to perform a bigger refactoring to support this feature, so it's left for CSL-1596.

P. S. I also need to regenerate genesis, will do it.
**Update**: regenerated.


core/Pos/Arbitrary/Core.hs
```diff
@@ -523,9 +534,11 @@ instance Arbitrary G.GenesisCoreData where
                 ]
         stakeDistrs <- vectorOf outerLen distributionGen
         hashmapOfHolders <- arbitrary :: Gen (Map Types.StakeholderId Word16)
+        delegation <- arbitrary
```
volhovM: Maybe delegation issuers should be related to addresses of genesis core data?

core/Pos/Core/Genesis/Types.hs
```diff
@@ -96,13 +103,50 @@ instance Buildable GenesisWStakeholders where
...
+-- 2. PSKs must be consistent with keys in the map, i. e. issuer's ID must be
+--    equal to the key in the map.
+-- 3. Delegates can't be issuers, i. e. transitive delegation is not supported.
+--    It's not needed in genesis, it can always be reduced.
+newtype GenesisDelegation = UnsafeGenesisDelegation
+    { unGenesisDelegation :: HashMap StakeholderId ProxySKHeavy
+    } deriving (Show, Eq)
+
+-- | Empty 'GenesisDelegation'.
+noGenesisDelegation :: GenesisDelegation
```
volhovM: I'd go with `deriving Monoid` instead.

core/Pos/Arbitrary/Core.hs
gromakovsky: 🤷‍♂️

core/Pos/Core/Genesis/Types.hs
gromakovsky: If I have `GenesisDelegation` where `a` delegates to `b` and `GenesisDelegation` where `b` delegates to `c`, their `mappend` will be illegal.

PR #1478, opened 2017-09-03
[CSL-1563] Move 'gitRev' out of 'cardano-sl'

rockbmb:


core/Pos/Core/Constants/Raw.hs
```diff
@@ -200,6 +201,15 @@ isDevelopment = True
 isDevelopment = False
 #endif
 
+#define QUOTED(x) "/**/x/**/"
+
+gitRev :: Text
```
volhovM: How is this going to work? You've moved the function that depends on a environment variable from `node` to `core` which doesn't solve the problem: changing git revision outside will change it and lead to recompilation of core and all dependent modules.

scripts/build/cardano-sl.sh
```diff
@@ -327,19 +327,25 @@ echo "'explorer' flag: $explorer"
...
+  sbuild="stack build --ghc-options=\"$ghc_opts_2\" $commonargs $norun --dependencies-only $args $prj"
   echo -e "$sbuild\n"
   eval $sbuild
 
   if [[ $no_code == true ]]; then
-    ghc_opts_2="$ghc_opts -fwrite-interface -fno-code"
+    ghc_opts_3="$ghc_opts2 -fwrite-interface -fno-code"
   else
-    ghc_opts_2="$ghc_opts"
+    ghc_opts_3="$ghc_opts2"
```
volhovM: Why did you switch to `ght_opts_N`? I personally find it less convenient than overriding `ghc_opts`.

gromakovsky:
@volhovM see [my comment](https://issues.serokell.io/issue/CSL-1563#comment=96-7017). Moving `gitRev` into executables won't solve the problem, because we will still need to pass `DGITREV=bla` to `cardano-sl` which will recompile whole `cardano-sl` whenever something changes. There may be a way to pass CPP definition only to executables, but I am not aware of it.
If we pass `DGITREV` only to `core`, `stack` will rebuild everything only when something changes in `core` which happens substantially less often than in all packages.
The problem with this solution is that `GITREV` may easily point to wrong commit. Note, however, that this problem exists anyway. In current master if you change revision but don't change anything in `cardano-sl`, it won't rebuild anything.
I don't know what is the ideal solution, we should probably ask in #haskell-best-practice channel.

rockbmb:
@gromakovsky have we decided how to move forward with this?

gromakovsky:
@rockbmb yes, see [here](https://issues.serokell.io/issue/CSL-1563#comment=96-7109).

gromakovsky:
I also asked a question in #sl-devops: https://input-output-rnd.slack.com/archives/G3TLHGM18/p1504948435000038

gromakovsky:
See #1647.

PR #1480, opened 2017-09-03
[CSL-1583] Add logging for update worker, minor code refactoring

georgeee:


node/src/Pos/Update/Download.hs
```diff
@@ -53,22 +53,34 @@ downloadUpdate :: forall ctx m . UpdateMode ctx m => ConfirmedProposalState -> m
...
+           else
+              logDebug $ sformat ("Update downloading already in progress, download state: "%listJson)
+                                 downSet
+
   where
+    clear downSetVar up = do
+        downSet <- atomically $ do
+            modifyTVar' downSetVar (HS.delete cpsUpdateProposal)
+            readTVar downSetVar
+        logDebug $ sformat ("Update "%build%" downloaded, download state: "%listJson)
```
gromakovsky: 1. It seems that this code will be called when `downloadUpdateDo` throws an exception, so this debug messages is a bit misleading.
2. Passing `UpdateProposal` to the functions inside `where` is redundant in this case, because it's always `cpsUpdateProposal` and it's available to this scope. Or if you really want to pass it, it makes more sense to use `up` instead of `cpsUpdateProposal` two lines above.

Anyway, I am going to approve, because apparently I need to touch this code in CSL-1592 anyway.

PR #1481, opened 2017-09-03
[CSM-437, CSM-366] Don't leave money on addresses with revealed public keys

flyingleafe:
These changes in transaction creation logic are intended to prevent users from accidentally having money on addresses with already revealed public keys, which is not safe in post-quantum environment.

This doesn't prevent accidentally having money on used address completely, of course - e. g. it doesn't prevent somebody from sending some money to one of users' used addresses, but this makes sure that every single transaction takes all the money from any given address, without leaving some money intact.


node/src/Pos/Client/Txp/Util.hs
```diff
@@ -196,9 +208,35 @@ makeRedemptionTx rsk txInputs = makeAbstractTx mkWit (map ((), ) txInputs)
...
+sumTxOutCoins :: NonEmpty TxOutAux -> Integer
+sumTxOutCoins = sumCoins . map (txOutValue . toaOut)
+
+-- | Group unspent outputs by addresses
+groupUtxo :: Utxo -> [UtxoGroup]
+groupUtxo utxo =
+    map mkUtxoGroup preUtxoGroups
+  where
+    futxo = M.toList utxo
+    comparingAddrs = (== EQ) ... comparing (txOutAddress . toaOut . snd)
```
Martoon-00: Why not `(==) 'on' txOutAddress . toaOut . snd`

node/src/Pos/Client/Txp/Util.hs
```diff
@@ -196,9 +208,35 @@ makeRedemptionTx rsk txInputs = makeAbstractTx mkWit (map ((), ) txInputs)
...
+    preUtxoGroups = NE.groupBy comparingAddrs futxo
```
Martoon-00: `groupBy` works for quadratic time

node/src/Pos/Client/Txp/Util.hs
```diff
@@ -234,24 +272,29 @@ prepareTxRaw utxo outputs (TxFee fee) = do
             pure TxRaw {..}
   where
     sumTxOuts = either (throwError . GeneralTxError) pure .
-        integerToCoin . sumCoins . map (txOutValue . toaOut)
-    allUnspent = M.toList utxo
-    sortedUnspent =
-        sortOn (Down . txOutValue . toaOut . snd) allUnspent
+        integerToCoin . sumTxOutCoins
+    gUtxo = groupUtxo utxo
+    isOutputAddr addr = any ((addr ==) . txOutAddress . toaOut) outputs
```
Martoon-00: And here

PR #1482, opened 2017-09-03
[CSL-1584] Make tx payload in blockgen configurable

georgeee:


core/constants.yaml
```diff
@@ -6,7 +6,8 @@
 #   * testnet_public_{full,wallet}: final version for users, big k
 #   * testnet_staging_{full,wallet}: QA, smaller k
 # * Dev (4 nodes):
-#   * devnet_longep_{full,wallet}: development purposes, big k
+#   * devnet_midep_{full,wallet}: development purposes, k as on staging
+#   * devnet_longep_{full,wallet}: development purposes, k as on mainnet
```
volhovM: Lacks description of benchmarks and update system configs.

PR #1483, opened 2017-09-04
[CSM-435] Wallet ready flag

ajnsit:
This is WIP because I am still refining the endpoints for wallets that are not ready

### DIFF
--- a/node/src/Pos/Wallet/Web/State/Storage.hs
+++ b/node/src/Pos/Wallet/Web/State/Storage.hs
@@ -17,6 +17,7 @@ module Pos.Wallet.Web.State.Storage
        , getAccountMeta
        , getWalletMetas
        , getWalletMeta
+       , getWalletMetaIncludeUnready
        , getWalletPassLU
        , getWalletSyncTip
        , getWalletAddresses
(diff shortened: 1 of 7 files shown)


node/src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -208,7 +213,7 @@ getWalletSyncTip cWalId = preview (wsWalletInfos . ix cWalId . wiSyncTip)
 
 getWalletAddresses :: Query [CId Wal]
 getWalletAddresses =
-    map fst . sortOn (view wiCreationTime . snd) . HM.toList <$>
+    map fst . sortOn (view wiCreationTime . snd) . filter (view wiIsReady . snd) . HM.toList <$>
```
flyingleafe: I think that it would be nice to not only add this check here, but also throw an error with according message in `getWallet`, when given wallet isn't ready yet.

ajnsit: Done!

akegalj:
@ajnsit you didn't expose endpoints to client-api (purescript code) ?

PR #1484, opened 2017-09-04
Cardano sl 0.6 staging

georgeee:


georgeee:
Shouldn't be merged.
Only `scripts/generate/genesis.sh` should be copied to master

PR #1485, opened 2017-09-04
[CSL-1592] First step towards reworking software update downloading

gromakovsky:
1. Previously 'checkForUpdate' was downloading all new updates for our
    application one-by-one. It was useless because we only need to the latest one.
    It's also useless because after we download one update, we won't download
    anymore until the application is restarted.
    Now 'checkForUpdate' downloads only the newest update.
    
    2. It's no longer possible to download multiple updates concurrently.
    Previously it was possible, but it didn't make any sense. After one
    download is started, another one will not start at all (if the first one
    has already created a file) or will do some stuff and fail slightly later
    (again, due to check that the file with an update doesn't exist) or in the
    case there will be a race condition between two downloads with unpredictable
    results.
    
    3. If we have already downloaded an update, we won't download anymore.
    Previously the logic was similar, but the check was done using 'doesFileExist'
    which was probably less reliable.
    
    4. Copy-pasted 'updateTriggerWorker' was moved into 'Pos.Update.Worker'.
    
    There are few more things to improve, see CSL-1602.

### DIFF
--- a/node/src/Pos/Update/Download.hs
+++ b/node/src/Pos/Update/Download.hs
@@ -4,16 +4,15 @@
 
 module Pos.Update.Download
        ( downloadUpdate
-       , downloadHash
        ) where
 
-import           Control.Concurrent.STM  (modifyTVar')
+import           Universum
+
 import           Control.Lens            (views)
 import           Control.Monad.Except    (ExceptT (..), throwError)
 import qualified Data.ByteArray          as BA
 import qualified Data.ByteString.Lazy    as BSL
 import qualified Data.HashMap.Strict     as HM
-import qualified Data.HashSet            as HS
 import           Ether.Internal          (HasLens (..))
 import           Formatting              (build, sformat, stext, (%))
 import           Network.HTTP.Client     (Manager, newManager)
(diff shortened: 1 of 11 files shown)


node/src/Pos/Update/Download.hs
```diff
@@ -22,80 +21,74 @@ import           Network.HTTP.Simple     (getResponseBody, getResponseStatus,
...
-                                 downSet
-
+downloadUpdate cps = do
+    downloadLock <- ucDownloadLock <$> view (lensOf @UpdateContext)
+    withMVar downloadLock $ \() -> do
+        downloadedUpdateMVar <-
+            ucDownloadedUpdate <$> view (lensOf @UpdateContext)
+        tryTakeMVar downloadedUpdateMVar >>= \case
+            Nothing -> downloadUpdateDo cps
+            Just existingCPS -> onAlreadyDownloaded existingCPS
```
georgeee: Why we take `MVar` here, log something and don't put it back? Perhaps Tvar should be used

node/src/Pos/Update/Worker.hs
```diff
@@ -28,12 +35,42 @@ usWorkers =
...
+        let newestCPS =
+                maximumBy (comparing cpsToNumericVersion) confirmedProposals
+        logInfo $
+            sformat
+                ("There are new confirmed update proposals for our application: "
+                 %listJsonIndent 2%
+                 "\n The newest one is: "%build%" and we want to download it")
+                (cpsUpdateProposal <$> confirmedProposals)
+                (cpsUpdateProposal newestCPS)
+        downloadUpdate newestCPS
```
georgeee: Checks like `isVersionAppropriate` and for `This update is not for our system` should be done here, before picking up maximum

node/src/Pos/Update/Download.hs
gromakovsky: Fix already.

node/src/Pos/Update/Worker.hs
gromakovsky: I think it doesn't matter, because it's only a sanity check which must always pass (by contract of `getConfirmedProposals`).
There advantages and disadvantages of doing it earlier and later.
If we do it earlier, we detect error in code earlier.
If we do it later, we cover more logic and will detect if something is broken somewhere in between this function and the place where we check it.
In theory we can do this check in 100500 places or can never do it. I don't think it's essential enough to really care about it.

georgeee: @gromakovsky but what will happen if there are proposal A which has installer for Windows and proposal B for MacOS. And B hash highest version than A. And then widnows can't be updated to A, though it seems beneficial and good thing to do

gromakovsky: Ah, you meant different check, ok, got it.

gromakovsky: @georgeee ok, I changed `getAllConfirmedProposals` so that it returns only those proposals which have data for our `SystemTag`.

node/src/Pos/Update/Download.hs
georgeee: Right

update/Pos/Update/Constants.hs
```diff
@@ -82,6 +88,20 @@ curSoftwareVersion :: SoftwareVersion
 curSoftwareVersion = SoftwareVersion ourAppName
                                      (ccApplicationVersion updateConstants)
 
+ourSystemTag :: SystemTag
+ourSystemTag = $(do
+    mbTag <- runIO (lookupEnv "CSL_SYSTEM_TAG")
```
domenkozar: Could we here just take different tags per OS instead of using bash env? It's easier for now as we don't really need to change this.

georgeee: Determining os if harder then just setting it from build script I guess

PR #1487, opened 2017-09-04
[CSM-438] Rewrite stabilize fee algorithm

Martoon-00:
Use better iterative algorithm to settle transaction, which may find suboptimal fee, but always converges.
Algorithm is still to be improved later, most likely after mainnet.

Tested manually in dev mode


node/src/Pos/Communication/Tx.hs
```diff
@@ -49,10 +48,10 @@ type TxMode ssc m
       , TxCreateMode m
       )
 
-submitAndSaveTx
+submitAndSave
```
Martoon-00: This is an extra change I made in CSM-256 and now name of this function was confusing (`submitTx` called `submitAndSaveTx`)

PR #1490, opened 2017-09-04
[CSL-619] Delete broken 'MVar' functions

gromakovsky:
These functions are bad, because they don't take asynchronous exceptions
into account. Read carefully and you'll see it.
Also they are not used at all.
If someone wants these functions, it's usually better to use STM instead.


gromakovsky:
Unicode handling by git on Windows = 🤦‍♂️

PR #1493, opened 2017-09-04
[CSL-1583] Test US for master

georgeee:


node/src/Pos/Launcher/Scenario.hs
```diff
@@ -146,8 +146,11 @@ runNode nr (plugins, plOuts) =
 nodeStartMsg :: WithLogger m => m ()
 nodeStartMsg = logInfo msg
   where
-    msg = sformat ("Application: " %build% ", last known block version " %build)
-                   Const.curSoftwareVersion Const.lastKnownBlockVersion
+    msg = sformat ("Application: " %build% ", last known block version "
+                    %build% ", systemTag: " %build)
+                   Const.appSystemTag
```
gromakovsky: This should be the third one, not the first one.

node/src/Pos/Update/Download.hs
```diff
@@ -9,6 +9,7 @@ module Pos.Update.Download
 import           Universum
 
 import           Control.Lens            (views)
+import           Control.Monad.Catch     (handleAll)
```
gromakovsky: Don't use this module use [safe functions](https://www.stackage.org/haddock/lts-9.2/safe-exceptions-0.1.6.0/Control-Exception-Safe.html#v:handleAny).

node/src/Pos/Update/Download.hs
```diff
@@ -106,36 +110,43 @@ downloadUpdateDo cps@ConfirmedProposalState {..} = do
...
-        logInfo "Update was downloaded"
+        logInfo $ "Update was downloaded, saved to " <> show updPath
         downloadedMVar <- views (lensOf @UpdateContext) ucDownloadedUpdate
         putMVar downloadedMVar cps
         logInfo "Update MVar filled, wallet is notified"
 
-    whenLeft res logWarning
+    whenLeft res logDownloadError
   where
+    handleErr = return . Left . show
```
gromakovsky: If you catch `SomeException`, you should use `reportOrLog` function in your handler to report fatal errors.

node/src/Pos/Update/Download.hs
```diff
@@ -106,36 +110,43 @@ downloadUpdateDo cps@ConfirmedProposalState {..} = do
...
     isVersionAppropriate :: SoftwareVersion -> Bool
     isVersionAppropriate ver = svAppName ver == svAppName curSoftwareVersion
         && svNumber ver > svNumber curSoftwareVersion
 
-
 -- Download a file by its hash.
 --
 -- Tries all servers in turn, fails if none of them work.
-downloadHash :: [Text] -> Hash LByteString -> IO (Either Text LByteString)
+downloadHash :: UpdateMode ctx m => [Text] -> Hash LByteString -> m (Either Text LByteString)
```
gromakovsky: This constraint is too much for this function (it only needs `MonadIO` and `WithLogger`), but it's minor, so I don't care much.

PR #1494, opened 2017-09-04
[CSL-1600] Use macOS provided security binary

georgeee:


domenkozar:
it overrides `postPatch` to do nothing instead of patching the source.

PR #1495, opened 2017-09-04
[CSL-1578]  Finish efforts on concurrency code refinement 

gromakovsky:
It's basically #1163 ported to master and also removal of `fork` from `onNewSlot` as was discussed in that PR.

### DIFF
--- a/infra/Pos/Shutdown/Logic.hs
+++ b/infra/Pos/Shutdown/Logic.hs
@@ -1,27 +1,15 @@
 module Pos.Shutdown.Logic
-       ( runIfNotShutdown
-       , triggerShutdown
-       , waitForWorkers
+       ( triggerShutdown
+       , waitForShutdown
        ) where
 
 import           Universum
 
-import           Control.Concurrent.STM (readTBQueue, readTVar, writeTBQueue, writeTVar)
-import           System.Wlog            (WithLogger, logDebug, logInfo)
+import           Control.Concurrent.STM (check, readTVar, writeTVar)
+import           System.Wlog            (WithLogger, logInfo)
 
 import           Pos.Shutdown.Class     (HasShutdownContext (..))
-import           Pos.Shutdown.Types     (shdnIsTriggered, shdnNotifyQueue)
-
-runIfNotShutdown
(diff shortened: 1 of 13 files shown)


gromakovsky:
I wrote some text in CSL-1606.

node/src/Pos/Block/Worker.hs
```diff
@@ -83,14 +83,19 @@ blkWorkers =
...
+-- TODO [CSL-1606] Using 'fork' here is quite bad, it's a temporary solution.
 blkOnNewSlot :: WorkMode ssc ctx m => (WorkerSpec m, OutSpecs)
 blkOnNewSlot =
     onNewSlotWorker True announceBlockOuts $ \slotId sendActions ->
         recoveryCommGuard $
         () <$
         metricWorker slotId `concurrently`
-        blockCreator slotId sendActions
+        void
+            (fork $
```
int-index: You added a fork in one place and added it in another?

gromakovsky: Yes, but only for this particular worker. For all other `onNewSlot` there is no `fork` because it's not needed. I would like to get rid of this one as well, but in another issue.
See this commit's message: https://github.com/input-output-hk/cardano-sl/commit/7f426d0bdbfdfdb8be92620950b614a85c6339ad
Also see CSL-1606 for broader explanation of `trickier`.

infra/Pos/Shutdown/Logic.hs
```diff
@@ -30,10 +18,12 @@ triggerShutdown = do
...
-waitForWorkers 0 = pass
-waitForWorkers n = do
-    view (shutdownContext . shdnNotifyQueue) >>= atomically . readTBQueue
-    waitForWorkers (n - 1)
+-- | Wait for the shutdown var to be true.
+waitForShutdown
+  :: (MonadIO m, MonadReader ctx m, WithLogger m, HasShutdownContext ctx)
+  => m ()
+waitForShutdown = do
+  v <- view (shutdownContext . shdnIsTriggered)
```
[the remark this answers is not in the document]
gromakovsky: It's not my code, so sorry for that 😿

PR #1501, opened 2017-09-05
[CSL-1373] Tool to analyse a blockchain

adinapoli-iohk:
The long-awaited blockchain tool is here! 🎉 

Hopefully it will be useful to the core team even in this first incarnation. Demo time, first!

## Displaying stats in 'Table' mode
screen shot 2017-09-05 at 17 38 58

## Displaying stats in 'CSV' mode
screen shot 2017-09-05 at 17 39 16

## Display stats in 'Human' mode
screen shot 2017-09-05 at 17 39 36

We support a number of basic-but-useful options:

```
Cardano SL blockchain generator

Usage: cardano-blockchain-analyser --db FILEPATH ([-k] | [-m] | [-g] | [-b])
                                   [--print-mode [human|csv|table]]
                                   [-i|--incremental]
  Analyse a blockchain and spit out useful metrics.

Available options:
  -h,--help                Show this help text
  --db FILEPATH            Location of the database where the blockchain is
                           stored.
  -k                       Display block counts in kilobytes (KB).
  -m                       Display block counts in megabytes (MB).
  -g                       Display block counts in gigabytes (GB).
  -b                       Display block counts in bytes (B).
  --print-mode [human|csv|table]
                           Select the desidered rendering mode, one between
                           'human', 'csv' or 'table'.
  -i,--incremental         Run in incremental mode. In this mode, table output
                           will be disabled and rendered as a .csv, as is not
                           possible to generate nice-looking tables whilst
                           reading the blockchain one block at time. You almost
                           always want to be using this mode for huge
                           blockchains, as it's much more memory efficient.

Command example:

  cardano-blockchain-analyser --db /path/to/existing/db
```

## Caveats

In order for the tool to be able to lookup blocks & headers is **required** to build it with the same `productionMode` used to build/forge the input database.


int-index:
Can you also remove `scripts/analyze/blocks.sh` as it's obsoleted by this new tool?

georgeee:
Neat! Screenshots look promising!

tools/src/blockchain-analyser/Main.hs
```diff
@@ -0,0 +1,98 @@
...
+-- | Tries to fetch an `Undo` for the given `Block`.
+fetchUndo :: HasCoreConstants => Block SscGodTossing -> BlockchainInspector (Maybe Undo)
+fetchUndo block = DB.blkGetUndo @SscGodTossing (blockHeaderHash $ getBlockHeader block)
+
+-- | Analyse the blockchain lazily by rendering all the blocks at once, loading the whole
+-- blockchain into memory. This mode generates very nice-looking tables, but using it for
+-- big DBs might not be feasible.
+analyseBlockchainLazily :: HasCoreConstants => CLIOptions -> HeaderHash -> BlockchainInspector ()
+analyseBlockchainLazily cli currentTip = do
+    allBlocks <- go currentTip mempty
```
gromakovsky: There is `loadBlundsFromTipWhile` in `Pos.DB.DB` which does the same thing (except that it returns `NewestFirst`).

tools/src/blockchain-analyser/Rendering.hs
```diff
@@ -0,0 +1,197 @@
+
+module Rendering ( render
+                 , renderBlock
+                 , renderBlocks
+                 , renderHeader
+                 ) where
```
gromakovsky: The way we format export lists is defined [here](https://github.com/input-output-hk/cardano-sl-style-guides/blob/master/haskell-style-guide.md#export-lists).
I know it's minor, but I want it to be consistent with other code. Note that you module `Options` follows that rule, so even in this PR formatting is different. When it's different, some people will change it, it will lead to unnecessary extra diff and noise, sometimes even conflicts.

tools/src/blockchain-analyser/Main.hs
```diff
@@ -0,0 +1,98 @@
...
+-- | Calculate the size of each folder.
+dbSizes :: FilePath -> IO [DBFolderStat]
+dbSizes root = do
+    parents <- ls root
+    forM (root : parents) $ \f -> (toText f,) <$> du_s f
+
+-- | Analyse the blockchain, printing useful statistics.
+analyseBlockchain :: HasCoreConstants => CLIOptions -> HeaderHash -> BlockchainInspector ()
+analyseBlockchain cli tip =
+    if incremental cli then do liftIO $ putText (renderHeader cli)
```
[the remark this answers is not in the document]
adinapoli-iohk: Oh, neat, I assumed it was not 😉

tools/src/blockchain-analyser/Types.hs
```diff
@@ -0,0 +1,84 @@
...
+import           Pos.DB.Sum           (DBSum (..))
+import           Pos.Ssc.GodTossing   (SscGodTossing)
+import           Pos.Util.Util        (postfixLFields)
+
+type DBFolderStat = (Text, Integer)
+
+-- | Enough context for analysing blocks.
+data BlockchainInspectorContext = BlockchainInspectorContext
+    { tbgcDBSum   :: DB.DBSum
+    , tbgcNodeDBs :: DB.NodeDBs
```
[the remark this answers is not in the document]
adinapoli-iohk: I think you are right -- Initially I wasn't aware of the difference between the different modes so I wanted to have `DBSum` at hand in order to have a bunch of instances defined, but I don't think this is necessary anymore.

tools/src/blockchain-analyser/Rendering.hs
```diff
@@ -0,0 +1,197 @@
...
+toTableRow :: HasCoreConstants => UOM -> (Block SscGodTossing, Maybe Undo) -> [Text]
+toTableRow uom (block, mbUndo) =
+    let blockHeader   = getBlockHeader block
+        previousBlock = sformat build (prevBlock block)
+        blockHash     = blockHeaderHash blockHeader
+        epoch         = sformat build (getEpochIndex (getEpoch blockHeader))
+        blockType     = either (const "GENESIS") (const "MAIN") block
+        slot          = maybe mempty (sformat build . getSlotIndex . siSlot) (getSlot blockHeader)
+        leader        = maybe mempty (sformat build) (getLeader blockHeader)
+        txCount       = sformat build (length (getTxs block))
```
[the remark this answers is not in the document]
adinapoli-iohk: Neat!

tools/src/blockchain-analyser/Main.hs
adinapoli-iohk: I've tried to use this one, but the `condition` of `loadBlundsFromTipWhile` is a pure function whereas my terminating condition is whether I can fetch the next block or not, for which I need to run a monadic computation in the DB monad. Dunno if there is some other trick I can play, but otherwise I guess we'll have to live with the current implementation 😞

adinapoli-iohk:
@gromakovsky Ok, I think I have addressed most of the feedback (module the one on `allBlock` which I think we might need to stick to the status quo). Let me know what you think!

tools/src/blockchain-analyser/Main.hs
gromakovsky: @adinapoli-iohk just pass `const True`, it will load all blocks and will stop when it reaches the deepest genesis block. That's exactly what you need, AFAIU. A small advantage of using `loadBlundsFromTipWhile` is that there will be `DBMalformed` exception when there is no previous block in DB for any block different from the deepest genesis block (which indeed shouldn't happen for valid DB), while your version will ignore it silently.

tools/src/blockchain-analyser/Types.hs
```diff
@@ -0,0 +1,74 @@
...
+instance
+    HasCoreConstants =>
+    MonadBlockDBGeneric (BlockHeader SscGodTossing) (Block SscGodTossing) Undo BlockchainInspector
+  where
+    dbGetBlock = BDB.dbGetBlockDefault @SscGodTossing
+    dbGetUndo = BDB.dbGetUndoDefault @SscGodTossing
+    dbGetHeader = BDB.dbGetHeaderDefault @SscGodTossing
+
+prevBlock :: HasCoreConstants => Block SscGodTossing -> HeaderHash
+prevBlock = either (view prevBlockL) (view prevBlockL)
```
gromakovsky: Actually there is
```
instance (HasPrevBlock s, HasPrevBlock s') =>
         HasPrevBlock (Either s s') where
    prevBlockL = choosing prevBlockL prevBlockL
```
so you can write `view prevBlockL` (no `either`).

tools/src/blockchain-analyser/Types.hs
```diff
@@ -0,0 +1,74 @@
...
+import           Pos.DB               (MonadBlockDBGeneric (..), MonadDBRead (..))
+import qualified Pos.DB               as DB
+import qualified Pos.DB.Block         as BDB
+import           Pos.Ssc.GodTossing   (SscGodTossing)
+import           Pos.Util.Util        (postfixLFields)
+
+type DBFolderStat = (Text, Integer)
+
+-- | Enough context for analysing blocks.
+data BlockchainInspectorContext = BlockchainInspectorContext { tbgcNodeDBs :: DB.NodeDBs }
```
gromakovsky: I just noticed: `tbgc` prefix looks strange, usually we use abbreviation of type name as a prefix (so `bic` in this case).

adinapoli-iohk:
@gromakovsky ok, it should all be addressed now! Note how I refrained from using `Blund` inside `Rendering` so the `analyseBlockchainLazily` still needs to fetch each `Undo` manually and the lookup can fail with `Nothing`. We could in principle use one of the function which makes the tool blowup with a `MalformedDB` exception, but I think that for tools like this it's nice to have a bit of slack and be more lenient than usual.

PR #1503, opened 2017-09-05
[CSM-442] Improve tx creation tests

Pastafarianist:
- [x] Get rid of `PureDB`
- [x] Use Arbitrary instances for address generation
- [x] Provide modifiable `BlockVersionData`
- [x] Provide `BlockVersionData` via `ReaderT` instead of `StateT` and use `local` to change it


core/Pos/Crypto/SafeSigning.hs
```diff
@@ -139,9 +139,10 @@ safeDeterministicKeyGen
 safeDeterministicKeyGen seed pp =
     bimap PublicKey (mkEncSecret pp) (safeCreateKeypairFromSeed seed pp)
 
--- | SafeSigner datatype to encapsulate sensible data
+-- | SafeSigner datatype to encapsulate sensitive data
 data SafeSigner = SafeSigner EncryptedSecretKey PassPhrase
                 | FakeSigner SecretKey
+                deriving Show
```
flyingleafe: what did you need it for?
it's dangerous to have `Show` instance for sensitive data like this, because it can be accidentally shown in some logs or wherever, which is bad

flyingleafe: Ah, never mind: `Show` instances for `EncryptedSecretKey` and `PassPhrase` are safe, it's alright.

Martoon-00: +1
I guess that `Show` instance is required by QuickCheck for automatic generation

Pastafarianist: Exactly. It is required by `forAllM`.

PR #1505, opened 2017-09-05
[CSL-1613] Perform a check for genesis block creation outside the sta…

int-index:
…te lock

### DIFF
--- a/node/src/Pos/Block/Logic/Creation.hs
+++ b/node/src/Pos/Block/Logic/Creation.hs
@@ -41,7 +41,6 @@ import           Pos.Core                   (Blockchain (..), EpochIndex,
                                              epochSlots, flattenSlotId, getEpochOrSlot,
                                              headerHash)
 import           Pos.Crypto                 (SecretKey)
-import           Pos.DB                     (DBError (..))
 import qualified Pos.DB.Block               as DB
 import qualified Pos.DB.DB                  as DB
 import           Pos.Delegation             (DelegationVar, DlgPayload (getDlgPayload),
@@ -64,7 +63,7 @@ import           Pos.Update.Core            (UpdatePayload (..))
 import qualified Pos.Update.DB              as UDB
 import           Pos.Update.Logic           (clearUSMemPool, usCanCreateBlock,
                                              usPreparePayload)
-import           Pos.Util                   (maybeThrow, _neHead)
+import           Pos.Util                   (_neHead)
 import           Pos.Util.Util              (HasLens (..), HasLens', leftToPanic)
 import           Pos.WorkMode.Class         (TxpExtra_TMP)
 
@@ -123,11 +122,16 @@ createGenesisBlockAndApply epoch =
         Left UnknownBlocksForLrc ->
             Nothing <$ logInfo "createGenesisBlock: not enough blocks for LRC"
         Left err -> throwM err
-        Right leaders ->
-            modifyStateLock
-                HighPriority
-                "createGenesisBlockAndApply"
-                (createGenesisBlockDo epoch leaders)
+        Right leaders -> do
+            tipHeader <- DB.getTipHeader
+            needGen <-
+                -- preliminary check outside the lock,
+                -- must be repeated inside the lock
+                needCreateGenesisBlock epoch tipHeader
+            if needGen
+                then modifyStateLock HighPriority "createGenesisBlockAndApply"
+                     (\_ -> createGenesisBlockDo epoch leaders)
+                else return Nothing
 
 createGenesisBlockDo
     :: forall ssc ctx m.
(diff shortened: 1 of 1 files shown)


node/src/Pos/DB/Block.hs
```diff
@@ -113,6 +114,13 @@ blkGetHeader
     => HeaderHash -> m (Maybe (BlockHeader ssc))
 blkGetHeader = dbGetBi BlockIndexDB . blockIndexKey
 
+blkGetHeaderOrThrow
+    :: (HasCoreConstants, SscHelpersClass ssc, MonadDBRead m)
+    => HeaderHash -> m (BlockHeader ssc)
+blkGetHeaderOrThrow = maybeThrow (DBMalformed msg) <=< blkGetHeader
+  where
+    msg = "blkGetHeaderOrThrow: expected the header to be present in the DB, but it isn't"
```
gromakovsky: It would be good to also put header's hash into error message (you won't be able to use cool `<=<`, but that's because life is hard).

node/src/Pos/Block/Logic/Creation.hs
```diff
@@ -138,23 +142,12 @@ createGenesisBlockDo
     -> HeaderHash
     -> m (HeaderHash, Maybe (GenesisBlock ssc))
 createGenesisBlockDo epoch leaders tip = do
-    let noHeaderMsg =
-            "There is no header is DB corresponding to tip from semaphore"
-    tipHeader <- maybeThrow (DBMalformed noHeaderMsg) =<< DB.blkGetHeader tip
+    tipHeader <- DB.blkGetHeaderOrThrow tip
```
gromakovsky: Actually there is `getTipHeader` in `Pos.DB.DB`.

node/src/Pos/Block/Logic/Creation.hs
```diff
@@ -123,11 +122,16 @@ createGenesisBlockAndApply epoch =
         Left UnknownBlocksForLrc ->
             Nothing <$ logInfo "createGenesisBlock: not enough blocks for LRC"
         Left err -> throwM err
-        Right leaders ->
-            modifyStateLock
-                HighPriority
-                "createGenesisBlockAndApply"
-                (createGenesisBlockDo epoch leaders)
+        Right leaders -> do
+            tipHeader <- DB.blkGetHeaderOrThrow =<< DB.getTip
```
gromakovsky: See below, there is `getTipHeader`.

node/src/Pos/Block/Logic/Creation.hs
int-index: Isn't it better to use the tip provided by the lock here?

gromakovsky: No, tip provided by the lock should never be used and will be removed. It's not guaranteed to be the same as the one in DB and DB is the right source of truth.

Even if later we introduce a way to safely store tip in-memory, I still find it better to have a helper which reads this tip from some mutable variable and fetches header from DB. But now it's not the case, so just use `DB.getTipHeader`.

int-index: Alright, I changed the code so it doesn't use the tip from the lock.

node/src/Pos/Block/Logic/Creation.hs
int-index: Done.

node/src/Pos/DB/Block.hs
int-index: Now that you told me about `getTipHeader`, I removed `blkGetHeaderOrThrow` (it's not needed in this PR anymore).

node/src/Pos/Block/Logic/Creation.hs
```diff
@@ -123,38 +122,31 @@ createGenesisBlockAndApply epoch =
...
-    let noHeaderMsg =
-            "There is no header is DB corresponding to tip from semaphore"
-    tipHeader <- maybeThrow (DBMalformed noHeaderMsg) =<< DB.blkGetHeader tip
+createGenesisBlockDo epoch leaders = do
+    tipHeader <- DB.getTipHeader
     logDebug $ sformat msgTryingFmt epoch tipHeader
-    shouldCreate tipHeader >>= \case
-        False -> (tip, Nothing) <$ logShouldNot
+    needCreateGenesisBlock epoch tipHeader >>= \case
+        False -> (BC.blockHeaderHash tipHeader, Nothing) <$ logShouldNot
```
gromakovsky: There is polymorphic `headerHash`, but I think both options are fine here.

adinapoli-iohk:
No clue about what's going on with Travis, it seems like it's all green (if you open the linked Travis Job) but still we have one yellow dot of doom for this one.

gromakovsky:
I restarted hlint job, let's see if it works.

gromakovsky:
It helped.

georgeee:
@int-index can you please explain what can be a possible effect of including/not including this change into mainnet release?

int-index:
If we don't include this changeset, there will be more blocking between block creation and transaction processing. It's a performance concern.

georgeee:
@int-index how much more locking? Taking lock once in 20s?

int-index:
I haven't done any measurements, but this code used to take a lock at the beginning of every slot. Now it does so only when a genesis block must be actually created.

PR #1506, opened 2017-09-05
[CSL-1614] Update versions in .cabal files

gromakovsky:


gromakovsky:
Only one MacOS build hasn't passed yet. MacOS is infinitely overloaded, so I think it's ok to merge right. It's a very small change anyway.

PR #1508, opened 2017-09-05
[CSL-803] Connect lwallet to node DB

int-index:


int-index:
It's not possible for lwallet and node to use the DB simultaneously:

image

So I'm going to implement two modes of operation for `lwallet`: with a DB and without a DB.

gromakovsky:
Okay, @int-index, I am merging this one as we agreed.

gromakovsky:
@int-index look https://issues.serokell.io/issue/CSL-1620

PR #1509, opened 2017-09-06
[CSL-1612] Add --seed for the genesis generator

neongreen:
and regenerate all genesises with `--seed 0`


core/Pos/Crypto/Random.hs
```diff
@@ -29,6 +36,35 @@ newtype SecureRandom a = SecureRandom {runSecureRandom :: IO a}
 instance MonadRandom SecureRandom where
     getRandomBytes n = SecureRandom (ByteArray.convert <$> secureRandomBS n)
 
+-- | A random generator with global state. Unlike 'SecureRandom', you can
+-- set the seed that will be used for 'GlobalRandom', or let it use OpenSSL.
+newtype GlobalRandom a = GlobalRandom {runGlobalRandom :: IO a}
```
gromakovsky: I don't understand, why do you need this hack? There is `deterministic` function for deterministic generation.

neongreen: I could use `deterministic` if the whole genesis-generating function was in `MonadRandom` and didn't use IO. That's not the case, and removing IO from those functions was too much effort.

Alternatively, I could pass the seed to all functions that use randomness, and use `deterministic seed` there. However, this is unacceptable because it's basically [this](https://xkcd.com/221/), just a bit more elaborate. To make it acceptable, I'd need to additionally thread the state of the generator so that different states would be used during different invocations of functions that use randomness.

gromakovsky: Okay, so tl;dr: global variable is used because it's faster to implement.

PR #1513, opened 2017-09-06
[CSL-1618] Run LRC before creating genesis block

gromakovsky:
Nobody will run it for us, we should do it by ourselves.

I also removed redundant 'waitLrc' in one place.
'lrcActionOnEpochReason' does 'waitLrc' internally.

We don't need to catch 'UnknownBlocksForLrc' because 'shouldCreate'
guarantees it.


gromakovsky:
@fersel asked me to merge it before Travis CI passes (which will happen on Saturday probably) because this fix is absolutely critical.

PR #1514, opened 2017-09-07
Merge CSL-1618 to master (from cardano-sl-0.6)

gromakovsky:


node/src/Pos/Block/Logic/Creation.hs
```diff
@@ -155,7 +149,14 @@ createGenesisBlockDo epoch leaders tip = do
         | otherwise =
             (chainQualityThreshold @Double <=) <$>
             calcChainQualityM (flattenSlotId $ SlotId epoch minBound)
+    -- We need to run LRC here to make 'verifyBlocksPrefix' not hang.
+    -- It's important to do it after taking 'StateLock'.
+    -- Note that it shouldn't fail, because 'shouldCreate' guarantees that we
+    -- have enough blocks for LRC.
     actuallyCreate tipHeader = do
+        lrcSingleShot epoch
```
pva701: tbh I don't understand why it correctly worked earlier. what did you change in logic?

gromakovsky: Why do you think it worked? It didn't. That's what CSL-1618 was all about.
If by _earlier_ you mean more than 2 days earlier, then it worked because we also had `lrcOnNewSlot` which was periodically launching LRC.

PR #1515, opened 2017-09-07
Patch genesis keys

volhovM:
I've added newlines and then introduced 3 more keys as dictated my monetary police.


gromakovsky:
Are you sure master is the right place for this PR?

PR #1516, opened 2017-09-07
Volhovm/genesis keys patch/0.6

volhovM:


gromakovsky:
Also make a PR to master please (`-s ours`, but probably it doesn't matter).

PR #1523, opened 2017-09-07
Regenerate genesis with right math

neongreen:


neongreen:
don't merge yet, I want to test it

PR #1524, opened 2017-09-07
CSL-1624 Always fail in compile-time if CSL_SYSTEM_TAG isn't set

gromakovsky:


gromakovsky:
In `travis-ci/pr` all builds succeeded. In `travis-ci/push` MacOS build hasn't started yet, even though 1.5 days passed. I think `/pr` build is much more important and the fact that it passed (and 3 builds in `/push` and appveyor build) is quite enough to merge it in this case.

PR #1526, opened 2017-09-08
[CSE-190] Explorer backend test coverage

ksaric:
https://issues.serokell.io/issue/CSE-190

Tests for the _explorer_.


explorer/src/Pos/Explorer/ExtraContext.hs
```diff
@@ -41,4 +41,4 @@ makeExtraCtx :: HasConfiguration => ExtraContext
 makeExtraCtx =
     let addressCoinPairs = utxoToAddressCoinPairs $ unGenesisUtxo genesisUtxo
         redeemOnly = filter (isRedeemAddress . fst) addressCoinPairs
-    in ExtraContext $ V.fromList redeemOnly
+    in ExtraContext $ V.fromList redeemOnly
```
Pastafarianist: Please avoid removing newlines at ends of files in particular, and such hidden changes in large PRs in general.

ksaric:
We can (finally?) mock out the external functions (easily) - see `HasExplorerCSLInterface` which is using `ExplorerMockMode` passed through `ExtraContextT`.
Now all that's missing is to add functional tests for benchmarking.

explorer/src/Pos/Explorer/ExplorerMode.hs
```diff
@@ -0,0 +1,331 @@
...
+instance HasConfigurations => MonadGState ExplorerTestMode where
+    gsAdoptedBVData = DB.gsAdoptedBVDataDefault
+
+instance (HasConfigurations, MonadSlotsData ctx ExplorerTestMode)
+      => MonadSlots ctx ExplorerTestMode
+  where
+    getCurrentSlot = do
+        view etcSlotId_L >>= \case
+            Nothing -> Slot.getCurrentSlotSimple =<< view etcSSlottingVar_L
+            Just slot -> pure (Just slot)
```
akegalj: `maybe` or some alternative other then case look like a more readable alternative (to me at least, its subjective). Minor - and irrelevant

explorer/src/Pos/Explorer/TestUtil.hs
```diff
@@ -0,0 +1,240 @@
...
+produceBlocksByBlockNumberAndSlots
+    :: forall m. (HasConfiguration, HasUpdateConfiguration, MonadIO m, Monad m)
+    => BlockNumber
+    -> SlotsPerEpoch
+    -> SlotLeaders
+    -> [SecretKey]
+    -> m [Block]
+produceBlocksByBlockNumberAndSlots blockNumber slotsNumber producedSlotLeaders secretKeys = do
+
+    -- This is just plain wrong and we need to check for it.
```
akegalj: Should we create an issue and/or add TODO/FIXME tag ?

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -221,51 +229,65 @@ getBlocksPage mPageNumber mPageSize = do
...
+    -- Get total blocks in the blockchain. Get the blocks total using this mode.
     blocksTotal <- toInteger <$> getBlocksTotal
 
-    -- Get total pages from the blocks. And we want the page
-    -- with the example, the page size 10,
-    -- to start with 10 + 1 == 11, not with 10 since with
-    -- 10 we'll have an empty page.
-    let totalPages = (blocksTotal - 1) `div` pageSize
+    -- Make sure the parameters are valid.
+    when (blocksTotal < 1) $
```
akegalj: As you said, this should probably go into different PR.

explorer/test/Test/Pos/Explorer/Web/ServerSpec.hs
```diff
@@ -1,17 +1,264 @@
...
+    describe "pureGetBlocksTotal"
+    $ modifyMaxSuccess (const 200) $ do
+        prop "created blocks means block size > 0" $ emptyBlk $ \blk0 -> leftToCounter blk0 $ \blk ->
+            let mainBlock   = Right blk
+                blocksTotal = pureGetBlocksTotal mainBlock
+            in counterexample ("Total blocks sizes: " <> show blocksTotal <>
+                               "\n\nBlock: " <> show blk) $
+                 blocksTotal > 0
+
+-- | A spec with the simple test that two equal algorithms should work the same.
```
akegalj: Description is too general. When I read it I thought that its testing equality of any two given algorithms.

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
```diff
@@ -0,0 +1,215 @@
...
+            emmGetBlundFromHH      = \_ -> pure $ Just $ genTestArgs ^. gtaBlundsFromHeaderHashes,
+            emmGetSlotStart        = \_ -> pure $ Just $ genTestArgs ^. gtaSlotStart,
+            emmGetLeadersFromEpoch = \_ -> pure $ Just $ genTestArgs ^. gtaSlotLeaders
+        }
+
+    -- The extra context so we can mock the functions.
+    let extraContext :: ExtraContext
+        extraContext = makeMockExtraCtx mode
+
+    runExplorerTestMode testParams extraContext $ getBlocksPage pageNumber (Just 10)
```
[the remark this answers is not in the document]
ksaric: 10 is the number of pages, it's what we use everywhere - when we modify that in other places, we can change it here, too.

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
```diff
@@ -0,0 +1,215 @@
...
+  where
+    -- | Configuration.
+    getBlocksPageConfig :: Config
+    getBlocksPageConfig = defaultConfig
+        { reportFile = Just "bench/results/ServerBackend.html"
+        }
+
+    get100BlocksTotalMock :: Benchmark
+    get100BlocksTotalMock = withDefConfigurations $
+        env (genArgs 100 50 2) $ bench "Get total blocks" . whnf getBlocksTotalMockBench
```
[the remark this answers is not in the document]
ksaric: No, they cannot be arbitrary. We want predictable and reproducible benchmarks, not random benchmarks.

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
```diff
@@ -0,0 +1,215 @@
...
+-- | Space @getBlocksPage@.
+runSpaceBenchmark :: IO ()
+runSpaceBenchmark = mapM_ (mainWith =<<)
+    [ get100BlocksTotalMock
+    , get100BlocksPageBench
+    , get1000BlocksPageBench
+    , get10000BlocksPageBench
+    ]
+  where
+    get100BlocksTotalMock :: IO (Weigh ())
```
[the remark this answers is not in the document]
ksaric: The best I could find for tracking memory usage - https://hackage.haskell.org/package/weigh
Other than profiling, but AFAIK, we can't do that from benchmarks (well, everything could be done, but not easily).

explorer/src/Pos/Explorer/TestUtil.hs
ksaric: I'm still working on the PR, but I think I just forgot to remove that line. Thanks for the reminder.

explorer/test/Test/Pos/Explorer/Web/ServerSpec.hs
ksaric: Well, it is. But that just proves that both are correct (or equally wrong). Will improve.

ksaric:
> I believe explorer/bench/results/ServerBackend.html file is autogenerated? Do we really want to commit it to repository ?

Yes. That's the CPU benchmark result. The idea is that we should be able to track the performance easily. If you change something, when you commit it, you can see the change. This way we can track what (or who) changed the performance of specific endpoints.

> @ksaric how can we compile/run these benchmarks and tests? (I am a bit out of loop)

`stack bench cardano-sl-explorer`

But as mentioned before, these tests are currently more a skeleton since they are using pure functions lifted in the required context.
"Now all that's missing is to add functional tests for benchmarking." - we need to add tests that use `ExplorerTestMode` since that can give a more realistic performance (ideally, we would need to use a "real" database, currently we are using memory, even in these functional tests).

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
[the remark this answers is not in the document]
sectore: Constant of `pageSize` is defined here https://github.com/input-output-hk/cardano-sl/blob/master/explorer/src/Pos/Explorer/DB.hs#L94-L95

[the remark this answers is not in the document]
ksaric: Yes, it's the merge to master from *yesterday*.

explorer/src/Pos/Explorer/ExtraContext.hs
```diff
@@ -37,8 +86,91 @@ instance Monad m => HasGenesisRedeemAddressInfo (ExtraContextT m) where
...
-        redeemOnly = filter (isRedeemAddress . fst) addressCoinPairs
-    in ExtraContext $ V.fromList redeemOnly
+-------------------------------------------------------------------------------------
+-- Explorer mock mode
+--
+-- The simple data structure that encapsulates functions that use CSL. We want to "cut"
+-- them out of the picture in order to be able to mock them.
+-------------------------------------------------------------------------------------
+
+data ExplorerMockMode = ExplorerMockMode
```
sectore: Nice idea to have this ^! How about to re-name it to `ExplorerMockableMode` (note `able`)? It's a tiny change of the name, but with this we can say: Hey, we are using this guy not just in production, but also we are `able` to mock it for tests if needed. With `ExplorerMockMode` (w/o `\able\`) it sounds like we have something to use for mocking only (for me at least).

explorer/src/Pos/Explorer/ExtraContext.hs
```diff
@@ -37,8 +86,91 @@ instance Monad m => HasGenesisRedeemAddressInfo (ExtraContextT m) where
...
+    getSlotStartCSLI :: SlotId -> m (Maybe Timestamp)
+    getLeadersFromEpochCSLI :: EpochIndex -> m (Maybe SlotLeaders)
+
+-- | The instance for external CSL functions.
+instance (Monad m, MonadBlockDB m, MonadDBRead m, MonadSlotsData ctx m) =>
+    HasExplorerCSLInterface (ExtraContextT m) where
+
+    getTipBlockCSLI = do
+        extraCtx <- Ether.ask @ExtraContext
+        let explorerMockMode = ecExplorerMockMode extraCtx
```
sectore: Maybe we could extract these two lines above to get `ExplorerMockMode` into a function, because they are used in all following functions `get...`. Just to be more DRY.

ksaric: We could do that, but it's just two lines and it's pretty simple to understand as-is. If we add another function things will become a bit more unreadable.
But yes, good catch, I was considering it :smile:

explorer/src/Pos/Explorer/ExtraContext.hs
ksaric: Makes sense. I will change it.

explorer/src/Pos/Explorer/TestUtil.hs
```diff
@@ -0,0 +1,240 @@
...
+
+instance HasConfiguration => Arbitrary DlgUndo where
+    arbitrary = genericArbitrary
+
+instance HasConfiguration => Arbitrary Undo where
+    arbitrary = genericArbitrary
+
+----------------------------------------------------------------
+-- Utility
+-- TODO(ks): Extract this in some common PureBlockTest module in src/?
```
sectore: It would be better to extract these function into `explorer/test/...` (not `explorer/src`), isn't? As I can see `basicBlockGenericUnsafe`, `basicBlock` etc. are used in tests only.

ksaric: Do have in mind some of these functions will probably be used in `bench/` which doesn't have access to `test`.
The idea behind the comment is to move this module somewhere globally, on the project level (like `lib/`), so we can reuse it from any module (and add other useful functions from any module).

explorer/src/Pos/Explorer/TestUtil.hs
```diff
@@ -0,0 +1,240 @@
...
+    -> BlockHeader
+    -> [TxAux]
+    -> ProxySKBlockInfo
+    -> SlotId
+    -> DlgPayload
+    -> SscPayload
+    -> UpdatePayload
+    -> SecretKey
+    -> Either Text MainBlock
+producePureBlock limit prev txs psk slot dlgPay sscPay usPay sk =
```
sectore: Wow - we have 9 parameters here. This function seems to be a good candidate to win a :medal_sports: in our sports category `heavy-weight-parameter-function` (or maybe:1st_place_medal:)  :smile:

ksaric: Actually, this is an existing function that I copied, but forgot from where :thinking: 
It's this reason that I believe we should have a commonplace module for all the functions we use in tests.

explorer/src/Pos/Explorer/TestUtil.hs
```diff
@@ -0,0 +1,240 @@
...
+
+        epochBlocks :: [MainBlock]
+        epochBlocks =
+            -- TODO(ks): Not correct, but I will fix it later.
+            generateBlocks getPrevBlockHeader <$> blockNumbers
+          where
+            blockNumbers :: [Word]
+            blockNumbers = [1..slotsPerEpoch']
+
+            -- type BlockHeader ssc = Either (GenesisBlockHeader ssc) (MainBlockHeader ssc)
```
sectore: `ssc` has been removed from `GenesisBlockHeader` and `MainBlockHeader`. Or just remove this comment, it's hard to keep types in a comment up-to-date.

ksaric: Yes, nice catch.

explorer/src/Pos/Explorer/TestUtil.hs
```diff
@@ -0,0 +1,240 @@
...
+
+                currentSecretKey :: SecretKey
+                currentSecretKey = secretKeys Prelude.!! integralBlockNumber
+
+                integralBlockNumber :: Integral a => a
+                integralBlockNumber = fromIntegral blockNumber'
+
+-- type SlotLeaders   = NonEmpty StakeholderId
+-- type StakeholderId = AbstractHash Blake2b_224 PublicKey
+
```
sectore: Pls remove these commented types ^ .

ksaric: Seems I already did, will push soon.

explorer/src/Pos/Explorer/Web/ClientTypes.hs
```diff
@@ -100,9 +101,11 @@ import           Servant.API                      (FromHttpApiData (..))
...
 -- The following types explain the situation better:
 --
--- type AddressHash = AbstractHash Blake2b_224
--- type Hash        = AbstractHash Blake2b_256
--- type TxId        = Hash Tx
+-- type AddressHash   = AbstractHash Blake2b_224
+-- type Hash          = AbstractHash Blake2b_256
+--
+-- type TxId          = Hash Tx               = AbstractHash Blake2b_256 Tx
+-- type StakeholderId = AddressHash PublicKey = AbstractHash Blake2b_224 PublicKey
```
sectore: How would you update this comment with these comment if the original types are changed by anybody else?

ksaric: Manually. This is the simplest thing I could do to explain to myself and others what are these things, and that is pretty valuable.
I personally don't think it's self-explanatory and these comments do simplify reading the code.

ksaric:
I was trying to remove `GeneratedTestArguments` and use `ExplorerMock(able)Mode` instead, but it seems that we cannot derive a `Generic` instance from it (because of it's complex/rich structure).
Also, we need to take a look at `ExplorerTestMode` if we want to extend it and use it in benchmarking.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -13,6 +16,15 @@ module Pos.Explorer.Web.Server
        , explorerApp
        , explorerHandlers
 
+       -- pure functions
+       , pureGetBlocksTotal
+       , pureGetBlocksPagesTotal
+
```
sectore: I know that naming might be hard. But I'm sure there are better ways to name pure function without `pure...`.

ksaric: Why? It's the most simple and straightforward thing to write, so why not use that? Note that the original function names are taken, and these are used in tests, so I don't see why it shouldn't be named like this.
But if you have better ideas how to express that a function is pure without using the word pure...
The alternative would be to use something like `getBlocksTotal` for a pure function and `getBlocksTotalM` for a "monadic" one. But then we would need to change the other functions as well.

explorer/src/Pos/Explorer/Web/ClientTypes.hs
sectore: I see your point and the comment is a nice documentation. However, nobody will care about this comment if - for example - `StakeholderId` has been changed and the type in this comment might be deprecated then. Just because `StakeholderId` is defined in `Pos.Core.Address` and not here in `Pos/Explorer/Web/ClientTypes.hs`. In such a case the comment will not just be useless but also wrong. We have already this case with deprecated `ssc` types. Maybe you could add some fictional type to explain it and link to the modules of these types to have real examples. Just an idea.

ksaric: > Maybe you could add some fictional type to explain it and link to the modules of these types to have real examples. Just an idea.

It's a good idea. I will think how I to use types to show this information. 
If you want them removed, I can remove them, but I would wait until we see a single one of them is wrong (which I don't think will happen soon).

explorer/src/Pos/Explorer/Web/Server.hs
sectore: > Why?
Because we have tons of pure functions which are not prefixed with `pure`. So why should we do this with few of them? You could use a name which described what the function does. For example `roundBlocksPagesTotal` instead of `pureGetBlocksPagesTotal`.

explorer/test/Test/Pos/Explorer/Web/ClientTypesSpec.hs
```diff
@@ -103,9 +106,79 @@ unitTests = do
 -- Quickcheck tests
 ----------------------------------------------------------------------------
 
+-- type AddressHash   = AbstractHash Blake2b_224
+-- type Hash          = AbstractHash Blake2b_256
+-- type TxId          = Hash Tx = AbstractHash Blake2b_256 Tx
+-- type StakeholderId = AddressHash PublicKey = AbstractHash Blake2b_224 PublicKey
```
sectore: Let's remove these ^ commented type information. Our editors can do this work for us :)

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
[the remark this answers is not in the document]
sectore: I'm not sure if we really have to put generated file `ServerBackend.html` into repository. It will be changed on my machine if I run `stack bench cardano-sl-explorer`. I guess this will be happened anytime if anyone run benchmarks on its machine.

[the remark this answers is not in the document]
ksaric: Notice the comment here - https://github.com/input-output-hk/cardano-sl/pull/1526#issuecomment-339933853

PR #1527, opened 2017-09-08
[CSL-1621] Rework lwallet

gromakovsky:
1. Light wallet was renamed to Aux, because it's not really a light wallet, it's a set of auxiliary stuff.
2. `Main` was splint into multiple modules.
3. Execution mode was basically reworked from scratch. Now it's almost the same as `RealMode`, but also has `CmdCtx` inside `ReaderT` environment and instances of some classes from `Client` modules (e. g. `MonadAddresses`).

Note that functionality didn't really change, even CLI remained almost the same as it was (you can see that modifications in scripts are really small).
I checked benchmarks and they work fine.


rubbish/src/Command/Parser.hs
```diff
@@ -1,94 +1,54 @@
...
-many1Till p end = (:) <$> p <*> manyTill p end
+-- many1Till :: Parser a -> Parser b -> Parser [a]
+-- many1Till p end = (:) <$> p <*> manyTill p end
 
 anyText :: Parser Text
 anyText = lexeme $ fmap toText $ manyTill anyChar (void (try space) <|> try eof)
 
 filePath :: Parser FilePath
-filePath = many1Till (alphaNum <|> oneOf "_.") (void (try space) <|> try eof)
+filePath = lexeme (many1 anyChar)
```
int-index: Mda, now this thing will consume everything up to `eof`.

rubbish/src/Command/Parser.hs
```diff
@@ -1,94 +1,54 @@
...
 lexeme :: Parser a -> Parser a
 lexeme p = spaces *> p >>= \x -> spaces $> x
 
 text :: String -> Parser Text
 text = lexeme . fmap toText . string
 
-many1Till :: Parser a -> Parser b -> Parser [a]
-many1Till p end = (:) <$> p <*> manyTill p end
+-- many1Till :: Parser a -> Parser b -> Parser [a]
+-- many1Till p end = (:) <$> p <*> manyTill p end
```
int-index: Commented out code.

rubbish/src/Command/Parser.hs
int-index: The proper way to implement it is to add escaping for whitespace in filenames.

rubbish/src/Mode.hs
```diff
@@ -0,0 +1,217 @@
...
+    dbGetHeader = dbGetHeaderSscDefault @RubbishSscType
+
+instance MonadGState RubbishMode where
+    gsAdoptedBVData = gsAdoptedBVDataDefault
+
+instance HasCoreConstants => MonadBListener RubbishMode where
+    onApplyBlocks = onApplyBlocksStub
+    onRollbackBlocks = onRollbackBlocksStub
+
+-- FIXME: I preserved the old behavior, but it most likely should be
```
int-index: It should! Please do it.

rubbish/src/Mode.hs
```diff
@@ -0,0 +1,217 @@
...
+instance HasSlogGState RubbishContext where
+    slogGState = rcRealModeContext_L . slogGState
+
+instance HasJsonLogConfig RubbishContext where
+    jsonLogConfig = rcRealModeContext_L . jsonLogConfig
+
+instance (HasCoreConstants, MonadSlotsData ctx RubbishMode)
+      => MonadSlots ctx RubbishMode
+  where
+    getCurrentSlot = getCurrentSlotSum
```
int-index: To be honest, I don't like what's going in this module. It'll be hard to guarantee consistency with `RealMode` this way, but I plan for `cardano-node-cli` (aka rubbish) to be a drop-in replacement for a node, used in testing and benchmarks. What I propose to do instead is to add a natural transformation:

```
inRealMode  :: RealMode SscGodTossing a -> RubbishMode a
inRealMode = withReaderT rcRealModeContext
```

and then define methods like this:

```
getCurrentSlotBlocking = inRealMode getCurrentSlotBlocking
```

It's possible to add more logic like this:

```
getCurrentSlotBlocking = do
    additionalWork
    inRealMode getCurrentSlotBlocking
```

but you'll know that the behavior of the RealMode is preserved.

rubbish/src/Mode.hs
int-index: Basically, I forgot to do it in CSL-803

rubbish/src/Pos/Rubbish/Hacks.hs
```diff
@@ -12,10 +12,10 @@ import           Pos.DB     (MonadGState, gsIsBootstrapEra)
 
 -- | In order to create an 'Address' from a 'PublicKey' we need to
 -- choose suitable stake distribution. We want to pick it based on
--- whether we are currently in bootstrap era.  Light wallet doesn't
--- know current slot, so let's assume it's 0-th epoch. It's enough for
--- our current needs.
-makePubKeyAddressLWallet :: MonadGState m => PublicKey -> m Address
-makePubKeyAddressLWallet pk = do
+-- whether we are currently in bootstrap era.  Rubbisy doesn't know
```
int-index: typo

int-index: And if you know how to fix that we use 0th epoch, perhaps do.

rubbish/src/Mode.hs
gromakovsky: Well, I would do it if it was trivial, but it's not so trivial, because AFAIU there were two implementations so far: this one with genesis utxo and the one in wallet (which uses wallet-db).
I don't want this thing to know about wallet db and I don't want to do something something non-(absolutely)trivial that isn't really for this issue (it was for CSL-803 most likely).
So I would like to do it in another issue.

rubbish/src/Command/Parser.hs
gromakovsky: Yes, but maybe someone will need it. There is also commented out `PublicKey` parser which isn't used.

rubbish/src/Pos/Rubbish/Hacks.hs
gromakovsky: Same as in https://github.com/input-output-hk/cardano-sl/pull/1527#discussion_r137869555
I just don't want this PR to become bigger if something can be done later (that wasn't spoiled by me).

rubbish/src/Command/Parser.hs
gromakovsky: I am wondering why nobody has written it yet 😒

rubbish/src/Mode.hs
gromakovsky: Created https://issues.serokell.io/issue/CSL-1635

rubbish/src/Pos/Rubbish/Hacks.hs
gromakovsky: Created https://issues.serokell.io/issue/CSL-1636

int-index:
You forgot to `git mv scripts/launch/{rubbish,aux}.sh`

jagajaga:
Due to windows limitations, not Linux.

gromakovsky:
If we want to use `cardano-auxx` in mainnet (I suppose we do, see CSLREQ-210, CSL-1553), then I would say it's necessary for mainnet.

PR #1529, opened 2017-09-08
[CSL-1598] Fix kademlia/topology/address/listener CLI mess

volhovM:
* Moved `tcpAddr` to `NetworkConfig`.
* Removed `TransportConfig` as it was holding `tcpAddr` only.
* Changed binding CLI logic:
   * Bind address is unchanged
   * External address is taken from cli or taken as bind address if absent (no need to specify same address twice explicitly)
   * If kademlia external address is not equal to CLI external address, exception is thrown
* Removed a bunch of useless parameters in explorer.
* Formatted `Pos.Network` according to the style guide.

### DIFF
--- a/explorer/src/explorer/ExplorerOptions.hs
+++ b/explorer/src/explorer/ExplorerOptions.hs
@@ -16,50 +16,32 @@ import           Options.Applicative.Simple (Parser, auto, help, long, metavar, 
                                              showDefault, simpleOptions, strOption,
                                              switch, value)
 import           Prelude                    (show)
-import           Serokell.Util.OptParse     (fromParsec)
 
 import           Paths_cardano_sl_explorer  (version)
 import qualified Pos.Client.CLI             as CLI
-import           Pos.DHT.Model              (DHTKey)
-import           Pos.DHT.Real.CLI           (dhtExplicitInitialOption, dhtKeyOption,
-                                             dhtNetworkAddressOption, dhtPeersFileOption)
 import           Pos.Network.CLI            (NetworkConfigOpts, networkConfigOption)
 import           Pos.Statistics             (EkgParams, StatsdParams, ekgParamsOption,
                                              statsdParamsOption)
 import           Pos.Util.BackupPhrase      (BackupPhrase, backupPhraseWordsNum)
-import           Pos.Util.TimeWarp          (NetworkAddress, addrParser)
 
 data Args = Args
-    { dbPath             :: !FilePath
-    , rebuildDB          :: !Bool
-    , keyfilePath        :: !FilePath
-    , backupPhrase       :: !(Maybe BackupPhrase)
-    , dhtKey             :: !(Maybe DHTKey)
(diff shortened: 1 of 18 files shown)


volhovM:
I need to rebase and test it first.

explorer/src/explorer/ExplorerOptions.hs
```diff
@@ -16,51 +16,32 @@ import           Options.Applicative.Simple (Parser, auto, help, long, metavar,
...
 import           Pos.Network.CLI            (NetworkConfigOpts, networkConfigOption)
-import           Pos.DHT.Model              (DHTKey)
-import           Pos.DHT.Real.CLI           (dhtExplicitInitialOption, dhtKeyOption,
-                                             dhtNetworkAddressOption, dhtPeersFileOption)
 import           Pos.Statistics             (EkgParams, StatsdParams, ekgParamsOption,
                                              statsdParamsOption)
 import           Pos.Util.BackupPhrase      (BackupPhrase, backupPhraseWordsNum)
-import           Pos.Util.TimeWarp          (NetworkAddress, addrParser)
 
 data Args = Args
```
gromakovsky: There is [this issue](https://issues.serokell.io/issue/CSL-1534), btw.

gromakovsky: It should just use `CommonNodeArgs` I believe, but ok, let's leave it for future.

infra/Pos/Network/CLI.hs
```diff
@@ -141,51 +193,50 @@ monitorStaticConfig :: forall m. (
                     -> Peers NodeId    -- ^ Initial value
                     -> m T.StaticPeers
 monitorStaticConfig cfg@NetworkConfigOpts{..} origMetadata initPeers = do
-    events :: Chan (MonitorEvent m) <- liftIO $ newChan
-
-    let loop :: Peers NodeId
```
gromakovsky: ~Useless~ Very subjective change, should be avoided. May lead to extra conflicts, make reviewers spend their time, etc. and doesn't make life much better.

infra/Pos/Network/CLI.hs
```diff
@@ -194,67 +245,92 @@ monitorStaticConfig cfg@NetworkConfigOpts{..} origMetadata initPeers = do
...
+intNetworkConfigOpts ::
+       forall m.
+       ( WithLogger m
+       , MonadIO m
+       , Mockable Fork m
+       , Mockable Catch m
+       , Mockable Throw m
+       )
+    => NetworkConfigOpts
+    -> m (T.NetworkConfig DHT.KademliaParams)
```
gromakovsky: This change is really useless, you should be more tolerant to formatting style you don't like.

volhovM: Yeah, maybe.

volhovM:
@gromakovsky First commit of this PR is mostly cosmetic, you can review other commits explicitly, they are smaller.

PR #1530, opened 2017-09-08
[CSL-1617] Use JSON for genesis files

pva701:
WIP


core/Pos/Core/Genesis/Types.hs
```diff
@@ -146,6 +168,145 @@ data GenesisCoreData = UnsafeGenesisCoreData
...
+integerToJSON :: Word64 -> JSValue
+integerToJSON = JSString . show
+
+coinToJSON :: Coin -> JSValue
+coinToJSON = word64ToJSON . unsafeGetCoin
+
+doubleToJSON :: Double -> JSValue
+doubleToJSON = JSString . show
+
+instance Monad m => ToJSON m Word16 where
```
georgeee: Minor TODO: move cananonical json instances to separate module

core/Pos/Core/Genesis/Types.hs
```diff
@@ -157,8 +318,8 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
 
 -- | Safe constructor for 'GenesisCoreData'. Throws error if something
 -- goes wrong.
-mkGenesisCoreData ::
-       [AddrDistribution]
+mkGenesisCoreData
+    :: [AddrDistribution]
```
georgeee: `AddrDistribution` is confusing, use different type!

core/Pos/Core/Genesis/Parser.hs
```diff
@@ -2,24 +2,29 @@
...
 -- time. Doesn't use TH with lift because it's difficult to provide 'Lift'
 -- instance to 'GenesisCoreData'
-compileGenCoreData :: GenesisCoreData
-compileGenCoreData =
+compileGenSpec :: GenesisSpec
+compileGenSpec =
+    --let file = $(embedFile =<< makeRelativeToProject ("init-data-" <> genesisBinSuffix <> ".json"))
     let file = $(embedFile =<< makeRelativeToProject ("genesis-core-" <> genesisBinSuffix <> ".bin"))
-    in case decodeFull file of
+    in case parseCanonicalJSON (BSL.fromStrict file) of
```
avieth: We *could* be more lenient here and parse typical JSON. Canonical JSON is only needed when verifying the genesis in production: the canonical rendering of the parsed thing must hash to some known value, but it doesn't matter if the input JSON has extra whitespace or keys out of alphanumeric order etc.

pva701: @avieth I know, we decided to compute hash of canonical json instead of input json today, however, I wrote this code earlier

core/Pos/Core/Genesis/Types.hs
```diff
@@ -155,37 +204,41 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
+    pure $ UnsafeGenesisSpec avvmDistr delega specType
+
+data ProtocolConstants = ProtocolConstants
+    {
+      -- | Security parameter from paper
+      pcK             :: !Int
+      -- | Magic constant for separating real/testnet
+    , pcProtocolMagic :: !Int32
+      -- | Start time of network (in @Production@ running mode). If set to
+      -- zero, then running time is 2 minutes after build.
```
avieth: Misplaced comment

core/Pos/Core/Genesis/Types.hs
```diff
@@ -155,37 +204,41 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
+    , pcVssMaxTTL     :: !Word64
+      -- | VSS certificates min timeout to live (number of epochs)
+    , pcVssMinTTL     :: !Word64
+      -- | Threshold of epochs for malicious activity detection
```
avieth: Misplaced comment

core/Pos/Core/Genesis/Types.hs
```diff
@@ -155,37 +204,41 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
+    }
+data GenesisCanonical = GenesisCanonical
+    { gcBootStakeholders :: !GenesisWStakeholders
+    , gcHeavyDelegation  :: !GenesisDelegation
+    , gcStartTime        :: !(Maybe Timestamp)
```
avieth: It's ok to have no start time?

core/Pos/Core/Genesis/Types.hs
```diff
@@ -155,37 +204,41 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
+      -- | Magic constant for separating real/testnet
+    , pcProtocolMagic :: !Int32
+      -- | Start time of network (in @Production@ running mode). If set to
+      -- zero, then running time is 2 minutes after build.
+    , pcVssMaxTTL     :: !Word64
+      -- | VSS certificates min timeout to live (number of epochs)
+    , pcVssMinTTL     :: !Word64
+      -- | Threshold of epochs for malicious activity detection
+    }
+data GenesisCanonical = GenesisCanonical
```
avieth: I'm confused by this name. A Haskell datatype is by nature canonical so the `Canonical` suffix is redundant. It's the JSON representation which may or may not be canonical. Could we rename this to `Genesis`?

core/Pos/Core/Genesis/Types.hs
```diff
@@ -135,16 +142,58 @@ mkGenesisDelegation psks = do
...
+    -- weights (@[(A, 5), (B, 2), (C, 3)]@). Setting this
+    -- overrides default settings for boot stakeholders (e.g. rich
+    -- in testnet stakes).
+    , tiDistribution    :: !TestnetDistribution
+    , tiSeed            :: !Integer
+      -- ^ Seed to use (when no seed is provided, a secure random generator
+      -- is used)
+    }
+    | MainnetInitializer {
+      miStartTime        :: !Timestamp
```
avieth: No start time needed for the testnet initializer?

core/Pos/Core/Genesis/Types.hs
```diff
@@ -135,16 +142,58 @@ mkGenesisDelegation psks = do
...
     return $ UnsafeGenesisDelegation resMap
 
+data TestBalanceOptions = TestBalanceOptions
+    { tsoPoors        :: Word
+    , tsoRichmen      :: Word
+    , tsoRichmenShare :: Double
+    , tsoTotalBalance :: Word64
+    } deriving (Show)
+
+data FakeAvvmOptions = FakeAvvmOptions
```
avieth: In what sense is this "fake"? Could we call it `TestnetAvvmOptions` instead?

core/Pos/Core/Genesis/Types.hs
```diff
@@ -135,16 +142,58 @@ mkGenesisDelegation psks = do
...
+      tiTestBalance     :: !TestBalanceOptions
+    , tiFakeAvvmBalance :: !FakeAvvmOptions
+    -- ^ Explicit bootstrap era stakeholders, list of addresses with
+    -- weights (@[(A, 5), (B, 2), (C, 3)]@). Setting this
+    -- overrides default settings for boot stakeholders (e.g. rich
+    -- in testnet stakes).
+    , tiDistribution    :: !TestnetDistribution
+    , tiSeed            :: !Integer
+      -- ^ Seed to use (when no seed is provided, a secure random generator
+      -- is used)
```
avieth: Misleading comment? It looks like a seed is always provided.

core/Pos/Core/Genesis/Types.hs
avieth: Ah, in dev mode I guess. Would be better in my opinion to have `gcStartTime :: !Timestamp` and in dev mode use the command-line argument to fill this in.

avieth:
I'm shy about diving in and doing work on this because I don't think I completely understand the design w.r.t. canonical JSON. In this PR there's *canonical* `ToJSON` instances for genesis-related stuff, but is this what we want? Why not typical aeson `ToJSON` and `FromJSON` instances? Then we can get the canonical JSON by defining some `Aeson.Value -> Maybe Canonical.JSValue`, but still freely parse and print non-canonically, which is more human-readable.

I also don't appreciate why we're using a hash of the canonical JSON rather than a hash of the genesis block itself. It does have a binary representation just like any other block, right?

core/Pos/Core/Genesis/Types.hs
georgeee: Right, to be passed in runtime

core/Pos/Core/Genesis/Types.hs
georgeee: `GenesisData`?

core/Pos/Core/Genesis/Types.hs
georgeee: Right, for testnet it's more convenient to pass start time in runtime

core/Pos/Core/Genesis/Types.hs
georgeee: In a sense generated certificates aren't real

core/Pos/Core/Genesis/Types.hs
pva701: tbh I don't like `GenesisData`, a lot of datatypes can be named `GenesisData` and it doesn't reflect a sense of this datatype.

core/Pos/Core/Genesis/JSON.hs
```diff
@@ -0,0 +1,169 @@
+-- |
```
gromakovsky: 🏳️‍🌈

core/Pos/Util/Util.hs
```diff
@@ -577,6 +579,12 @@ withTempFile tmpDir template action =
      ignoringIOErrors :: MC.MonadCatch m => m () -> m ()
      ignoringIOErrors ioe = ioe `MC.catch` (\e -> const (return ()) (e :: Prelude.IOError))
 
+applyPattern :: Show a => FilePath -> a -> FilePath
```
gromakovsky: Why is it here and not documented? This function seems to do something weird which is used only by keygen, so it should probably be somewhere near keygen.
If you really think it's general enough, it should be at least documented.

core/Pos/Core/Genesis/Types.hs
avieth: Would it not be OK to use the same type for "real" and "fake"? They have the same form, no? Within the program the testnet and mainnet avvm certs are equally real.

core/Pos/Core/Genesis/Types.hs
avieth: Conceptually the testnet initializer still needs a start time so we should put it in that constructor as well. What will we do with the command-line option for start time in dev mode? Use it to construct the initializer I'd imagine.

core/Pos/Core/Genesis/Types.hs
avieth: How about `GenesisBlockData`?

In `node/src/Pos//Genesis.hs` there is `data GenesisContext`. Is this `GenesisCanonical` supposed to replace it?

core/Pos/Core/Genesis/Types.hs
avieth: In any case, there's always a start time, so this ought to be `gcStartTime :: !Timestamp`.

core/Pos/Core/Genesis/Types.hs
```diff
@@ -155,37 +204,41 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
-        , gcdHeavyDelegation = delega
-        }
+    pure $ UnsafeGenesisSpec avvmDistr delega specType
+
+data ProtocolConstants = ProtocolConstants
+    {
+      -- | Security parameter from paper
+      pcK             :: !Int
+      -- | Magic constant for separating real/testnet
+    , pcProtocolMagic :: !Int32
```
avieth: We'll run into trouble with this if we insist on treating it like a top-level constant via `newInitFuture`. Protocol magic is defined here in a module which depends upon `Pos.Crypto.SignTag`, but then `Pos.Crypto.SignTag` makes use of this faux top-level constant and so we get a cycle.

The natural way of fixing this is to have `signTag` take the protocol magic as a parameter.

avieth: Breaking it with an `hs-boot` for `Pos.Core.Constants.Raw` should be ok I guess. The `hs-boot` will define `protocolMagic` only.

gromakovsky:
I am closing this one, see #1546. This PR still contains valuable comments which should be checked before finishing #1546.

PR #1535, opened 2017-09-09
[CSM-407] Refactor wallet

Martoon-00:
Part of CSM-407, what's done:

1. Reverted `ClientTypes.Instances` -> `ClientTypes.Functions` dependency, it allows to get rid of functions like `txIdToCTxId` which are equivalent to `encodeCType`
2. *Reordered* types in `ClientTypes.Types`
3. Some other small changes where I was dissatisfied with my code/comments
4. Move `instance Buildable`s for ClientTypes directly to type declarations
5. Removed extra `TODO`s (I was conservative here)


node/src/Pos/Wallet/Web/ClientTypes/Types.hs
```diff
@@ -171,36 +181,38 @@ data CAccountMeta = CAccountMeta
...
+-- Wallet structure - requests
+----------------------------------------------------------------------------
 
--- | Query data for account creation
-data CAccountInit = CAccountInit
-    { caInitMeta :: !CAccountMeta
-    , caInitWId  :: !(CId Wal)
-    } deriving (Show, Generic)
+-- | Query data for wallet creation
+data CWalletInit = CWalletInit
```
akegalj: @Martoon-00 we should be careful not to break Daedalus UI. This change won't break it atm because they are using client-api to create this object and make a request. 

But this will be breaking change for exchanges, who are using direct requests to the backend. 

Now, if we want to commit to this change we should add information about api changes to the release document so that other clients can react properly

Martoon-00: I haven't change any type actually, only reordered them

akegalj: Ok, wanted do double check with you. github actually confused me with diff ordering

flyingleafe:
This seems really nice, but we need approval from @georgeee / @dcoutts on this stage.

akegalj:
@flyingleafe 
we agreed with @dcoutts not to merge this before the launch. I will set WIP flag till post mainnet

gromakovsky:
@flyingleafe @Martoon-00 it seems we are now allowed to merge it to master.

Martoon-00:
Thanks for reminder, I'm going to resolve conflicts today

PR #1536, opened 2017-09-10
[CSM-446] WIP Take pending txs into account when evaluating balances

Martoon-00:
If transaction was failed to get applied but still can be applied later automatically (the only case I see is mempool overflow though), we still want it to influence on balances

_To be tested_

### DIFF
--- a/node/src/Pos/Client/Txp/Balances.hs
+++ b/node/src/Pos/Client/Txp/Balances.hs
@@ -2,6 +2,7 @@
 
 module Pos.Client.Txp.Balances
        ( MonadBalances(..)
+       , BalancesError(..)
        , getOwnUtxo
        , getBalanceFromUtxo
        , getOwnUtxosDefault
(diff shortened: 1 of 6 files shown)


node/src/Pos/Client/Txp/Balances.hs
```diff
@@ -53,17 +58,31 @@ type BalancesEnv ext ctx m =
...
+    tryTopsortTxs txs = fromMaybe txs $ topsortTxs identity txs
+
 getOwnUtxosDefault :: BalancesEnv ext ctx m => [Address] -> m Utxo
 getOwnUtxosDefault addrs = do
     let (redeemAddrs, commonAddrs) = partition isRedeemAddress addrs
 
-    updates <- getUtxoModifier
+    mempoolUpdates <- getUtxoModifier
+    recentTxs <- getRecentTxs
+    let applyUpdates = alaf Endo foldMap applyTxToUtxoWarilyPure recentTxs
```
flyingleafe: I strongly believe that there was a less confusing way to do it (e. g. _foldl'_?...)

node/src/Pos/Client/Txp/Balances.hs
```diff
@@ -53,17 +58,31 @@ type BalancesEnv ext ctx m =
...
 
+-- TODO [CSM-407] Move it somewhere :/
+-- | Get active pending transactions which are *not* in blockchain yet.
+getRecentTxs :: WebWalletModeDB ctx m => m [WithHash Tx]
+getRecentTxs = tryTopsortTxs . mapMaybe extract <$> WS.getPendingTxs
+  where
+    extract PendingTx{..} = do
+        let PtxApplying th = _ptxCond
+        Just $ WithHash (_thTx th) _ptxTxId
+    tryTopsortTxs txs = fromMaybe txs $ topsortTxs identity txs
```
flyingleafe: I think that it's not okay to just silently return unsorted tx list if we couldn't topsort list of transactions.
I'm not sure how exactly `topsortTxs` works now, but I believe that it should support transactions which depend on same `TxIn`s appearing in the list (which might perfectly be the case for pending pool). If it's so, then we clearly should throw an error if topsort fails - because it means that transactions in pending pool form a loop, which is nonsense.
If `topsortTxs` doesn't support such transactions (which is not likely), then I think it itself should be upgraded to support them.

flyingleafe:
After discussion with @Martoon-00 it became apparent that such changes can possibly lead to problems with creating invalid transactions. To make it right, we need to separate current valid balance from balance changes in pending transactions. This should better be done after mainnet release in course of CSM-50

PR #1537, opened 2017-09-10
Update launch scripts

deepfire:


jagajaga:
@deepfire copy-paste is not good.

PR #1538, opened 2017-09-10
[CSL-1635] Read own UTXO from DB with streaming in 'auxx'

int-index:


georgeee:
Agree, let's not affect benchmarks

gromakovsky:
Tbh, I have no idea what to do with this PR now. Let's wait until things settle down.

georgeee:
Required for delegation testing (before proper lwallet is established)

Not necessary to be merged to mainnet and maybe even `master` (eventually)

int-index:
Good idea, I'll make the necessary changes after I deal with Nix failing (the script is taking its time).

PR #1540, opened 2017-09-10
[CSL-1152] verifyHeaders check more predicates

volhovM:
Pre-story. Say you have headers `[h1,h2]` (`h2` is more difficult than `h1`) and you want to verify their signatures. Let `h1` be signed with `BlockSignature`: we match stakeholder id with related field in the header and check signature. And we're done. Then assume that `h2` is signed with `BlockPSignatureHeavy`, and `d` has the right to issue block according to `psk`. But then also assume that `psk` got into blockchain in block having header `h1`. So it is literally impossible to verify block signature of more than one header at once given using current delegation scheme.

What this PR does is:
* Delegation logic got abstracted (added some methods to `Cede`, removed duplicated code).
* As a result, found one missing check in the old mempool checker.
* `classifyNewHeader` now checks proxy signatures when header is a direct continuation of our chain
* `verifyHeaders` now accepts `Maybe Stakeholders` as a parameter so stakeholder field in headers can be (and is now) matched against data derived from fts.

### DIFF
--- a/node/src/Pos/Block/Logic/Header.hs
+++ b/node/src/Pos/Block/Logic/Header.hs
@@ -199,6 +199,11 @@ classifyHeaders inRecovery headers = do
                  ("Newest header is from slot "%build%", but current slot"%
                   " is "%build%" (and we're not in recovery mode)")
                  (newestHeader ^. epochOrSlotG) currentSlot
+         -- This check doesn't normally fail. RetrievalWorker
+         -- calculates lrc every time before calling this function so
+         -- it only fails when oldest header is from the next epoch e'
+         -- and somehow lrc didn't calculate data for e' knowing the
+         -- last header from e.
        | isNothing leaders ->
              pure $ CHsUseless $
              "Don't know leaders for oldest header epoch " <> pretty headersValid
--- a/node/src/Pos/Block/Logic/Header.hs
+++ b/node/src/Pos/Block/Logic/Header.hs
@@ -39,6 +39,7 @@ import           Pos.Crypto                (hash)
 import           Pos.DB                    (MonadDBRead)
 import qualified Pos.DB.Block              as DB
 import qualified Pos.DB.DB                 as DB
+import           Pos.Delegation.Cede       (dlgVerifyHeader, runDBCede)
 import qualified Pos.GState                as GS
 import           Pos.Slotting.Class        (MonadSlots (getCurrentSlot))
 import           Pos.Ssc.Class             (SscHelpersClass)
@@ -87,20 +88,19 @@ classifyNewHeader (Right header) = do
     let tip = headerHash tipHeader
     -- First of all we check whether header is from current slot and
     -- ignore it if it's not.
-    pure $ if
-        -- Checks on slots
+    if -- Checks on slots
         | maybe False (newHeaderSlot >) curSlot ->
-            CHUseless $ sformat
+            pure $ CHUseless $ sformat
                ("header is for future slot: our is "%build%
                 ", header's is "%build)
                curSlot newHeaderSlot
         | newHeaderEoS <= tipEoS ->
-            CHUseless $ sformat
+            pure $ CHUseless $ sformat
                ("header's slot "%build%
                 " is less or equal than our tip's slot "%build)
                newHeaderEoS tipEoS
         -- If header's parent is our tip, we verify it against tip's header.
-        | tip == header ^. prevBlockL ->
+        | tip == header ^. prevBlockL -> do
             let vhp =
                     VerifyHeaderParams
                     { vhpPrevHeader = Just tipHeader
@@ -118,17 +118,21 @@ classifyNewHeader (Right header) = do
                     , vhpMaxSize = Nothing
                     , vhpVerifyNoUnknown = False
                     }
-                verRes = verifyHeader vhp (Right header)
-            in case verRes of
-                   VerSuccess        -> CHContinues
-                   VerFailure errors -> mkCHRinvalid errors
+            let verRes = verifyHeader vhp (Right header)
+            dlgHeaderValid <- runDBCede $ runExceptT $ dlgVerifyHeader header
+            -- this case should be implemented using exceptions/monaderror (maybe..?)
+            pure $ case (verRes,dlgHeaderValid) of
+                   (VerSuccess,Right ())        -> CHContinues
+                   (VerSuccess,Left e)          -> mkCHRinvalid [e]
+                   (VerFailure errors,Left e)   -> mkCHRinvalid (e:errors)
+                   (VerFailure errors,Right ()) -> mkCHRinvalid errors
         -- If header's parent is not our tip, we check whether it's
         -- more difficult than our main chain.
-        | tipHeader ^. difficultyL < header ^. difficultyL -> CHAlternative
+        | tipHeader ^. difficultyL < header ^. difficultyL -> pure CHAlternative
         -- If header can't continue main chain and is not more
         -- difficult than main chain, it's useless.
         | otherwise ->
-            CHUseless $
+            pure $ CHUseless $
             "header doesn't continue main chain and is not more difficult"
 
 -- | Result of multiple headers classification.
--- a/node/src/Pos/Block/Logic/Header.hs
+++ b/node/src/Pos/Block/Logic/Header.hs
@@ -28,13 +28,13 @@ import           Pos.Block.Logic.Util      (lcaWithMainChain)
 import           Pos.Block.Pure            (VerifyHeaderParams (..), verifyHeader,
                                             verifyHeaders)
 import           Pos.Constants             (genesisHash, recoveryHeadersMessage)
-import           Pos.Context               (lrcActionOnEpochReason)
 import           Pos.Core                  (BlockCount, EpochOrSlot (..),
                                             HasCoreConstants, HeaderHash, SlotId (..),
-                                            blkSecurityParam, difficultyL, epochIndexL,
-                                            epochOrSlotG, getChainDifficulty,
-                                            getEpochOrSlot, headerHash, headerHashG,
-                                            headerSlotL, prevBlockL)
+                                            blkSecurityParam, bvdMaxHeaderSize,
+                                            difficultyL, epochIndexL, epochOrSlotG,
+                                            getChainDifficulty, getEpochOrSlot,
+                                            headerHash, headerHashG, headerSlotL,
+                                            prevBlockL)
 import           Pos.Crypto                (hash)
 import           Pos.DB                    (MonadDBRead)
 import qualified Pos.DB.Block              as DB
@@ -83,7 +83,7 @@ classifyNewHeader
     => BlockHeader ssc -> m ClassifyHeaderRes
 -- Genesis headers seem useless, we can create them by ourselves.
 classifyNewHeader (Left _) = pure $ CHUseless "genesis header is useless"
-classifyNewHeader (Right header) = do
+classifyNewHeader (Right header) = fmap (either identity identity) <$> runExceptT $ do
     curSlot <- getCurrentSlot
     tipHeader <- DB.getTipHeader @ssc
     let tipEoS = getEpochOrSlot tipHeader
@@ -91,26 +91,26 @@ classifyNewHeader (Right header) = do
     let newHeaderSlot = header ^. headerSlotL
     let newHeaderEpoch = header ^. epochIndexL
     let tip = headerHash tipHeader
-    leaders <-
-        lrcActionOnEpochReason
-            newHeaderEpoch
-            "classifyNewHeader"
-            LrcDB.getLeaders
+    maxBlockHeaderSize <- bvdMaxHeaderSize . snd <$> GS.getAdoptedBVFull
     -- First of all we check whether header is from current slot and
     -- ignore it if it's not.
-    if -- Checks on slots
-        | maybe False (newHeaderSlot >) curSlot ->
-            pure $ CHUseless $ sformat
-               ("header is for future slot: our is "%build%
-                ", header's is "%build)
-               curSlot newHeaderSlot
-        | newHeaderEoS <= tipEoS ->
-            pure $ CHUseless $ sformat
-               ("header's slot "%build%
-                " is less or equal than our tip's slot "%build)
-               newHeaderEoS tipEoS
+    when (maybe False (newHeaderSlot >) curSlot) $
+        throwError $
+        CHUseless $ sformat
+            ("header is for future slot: our is "%build%
+             ", header's is "%build)
+            curSlot newHeaderSlot
+    when (newHeaderEoS <= tipEoS) $
+        throwError $
+        CHUseless $ sformat
+            ("header's slot "%build%
+             " is less or equal than our tip's slot "%build)
+            newHeaderEoS tipEoS
         -- If header's parent is our tip, we verify it against tip's header.
-        | tip == header ^. prevBlockL -> do
+    if | tip == header ^. prevBlockL -> do
+            leaders <-
+                maybe (throwError $ CHUseless "Can't get leaders") pure =<<
+                lift (LrcDB.getLeaders newHeaderEpoch)
             let vhp =
                     VerifyHeaderParams
                     { vhpPrevHeader = Just tipHeader
@@ -123,18 +123,17 @@ classifyNewHeader (Right header) = do
                     -- this decision.
                     , vhpCurrentSlot = Nothing
                     , vhpLeaders = Just leaders
-                    -- can't know in advance since this can change over epoch
-                    , vhpMaxSize = Nothing
+                    , vhpMaxSize = Just maxBlockHeaderSize
                     , vhpVerifyNoUnknown = False
                     }
-            let verRes = verifyHeader vhp (Right header)
+            case verifyHeader vhp (Right header) of
+                VerFailure errors -> throwError $ mkCHRinvalid errors
+                _                 -> pass
+
             dlgHeaderValid <- runDBCede $ runExceptT $ dlgVerifyHeader header
-            -- this case should be implemented using exceptions/monaderror (maybe..?)
-            pure $ case (verRes,dlgHeaderValid) of
-                   (VerSuccess,Right ())        -> CHContinues
-                   (VerSuccess,Left e)          -> mkCHRinvalid [e]
-                   (VerFailure errors,Left e)   -> mkCHRinvalid (e:errors)
-                   (VerFailure errors,Right ()) -> mkCHRinvalid errors
+            whenLeft dlgHeaderValid $ throwError . CHInvalid
+
+            pure CHContinues
         -- If header's parent is not our tip, we check whether it's
         -- more difficult than our main chain.
         | tipHeader ^. difficultyL < header ^. difficultyL -> pure CHAlternative
@@ -165,9 +164,9 @@ data ClassifyHeadersRes ssc
 classifyHeaders ::
        forall ctx ssc m.
        ( DB.MonadBlockDB ssc m
+       , MonadCatch m
        , HasLens' ctx LrcContext
        , MonadSlots ctx m
-       , MonadCatch m
        , WithLogger m
        , HasCoreConstants
        )
@@ -178,25 +177,16 @@ classifyHeaders inRecovery headers = do
     tipHeader <- DB.getTipHeader @ssc
     let tip = headerHash tipHeader
     haveOldestParent <- isJust <$> DB.blkGetHeader @ssc oldestParentHash
-    leaders <-
-        lrcActionOnEpochReason
-            oldestHeaderEpoch
-            "classifyHeaders"
-            LrcDB.getLeaders
+    leaders <- LrcDB.getLeaders oldestHeaderEpoch
     let headersValid =
             isVerSuccess $
-            verifyHeaders (Just leaders) (headers & _NewestFirst %~ toList)
+            verifyHeaders leaders (headers & _NewestFirst %~ toList)
     mbCurrentSlot <- getCurrentSlot
     let newestHeaderConvertedSlot =
             case newestHeader ^. epochOrSlotG of
                 EpochOrSlot (Left e)  -> SlotId e minBound
                 EpochOrSlot (Right s) -> s
-    if | not headersValid ->
-             pure $ CHsInvalid "Header chain is invalid"
-       | not haveOldestParent ->
-             pure $ CHsInvalid
-                 "Didn't manage to find block corresponding to parent \
-                 \of oldest element in chain (should be one of checkpoints)"
+    if
        | newestHash == headerHash tip ->
              pure $ CHsUseless "Newest hash is the same as our tip"
        | newestHeader ^. difficultyL <= tipHeader ^. difficultyL ->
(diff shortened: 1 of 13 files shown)


node/src/Pos/Delegation/Cede/Logic.hs
```diff
@@ -103,3 +111,115 @@ dlgReachesIssuance i d psk = reach i
...
+
+-- | Verifies a header from delegation perspective (signature checks).
+dlgVerifyHeader ::
+       (MonadCedeRead m)
+    => MainBlockHeader ssc
+    -> ExceptT Text m ()
+dlgVerifyHeader h = do
+    -- Issuer didn't delegate the right to issue to elseone.
+    let issuer = h ^. mainHeaderLeaderKey
+    let sig = h ^. gbhConsensus ^. mcdSignature
```
gromakovsky: @neongreen said _mda_ when I used two `^.`.

node/src/Pos/Delegation/Cede/Logic.hs
```diff
@@ -103,3 +111,115 @@ dlgReachesIssuance i d psk = reach i
...
+        _ -> pass
+
+-- | Wrapper that turns on/off check of cycle creation in psk
+-- verification.
+newtype CheckForCycle = CheckForCycle Bool
+
+-- | Verify consistent heavy PSK.
+dlgVerifyPskHeavy ::
+       (MonadCedeRead m)
+    => HashSet StakeholderId
```
gromakovsky: There's `RichmenSet` alias in `Pos.Lrc.Types`.

node/src/Pos/Delegation/Logic/Mempool.hs
```diff
@@ -189,64 +188,48 @@ processProxySKHeavyInternal psk = do
...
-        in evalMapCede cedeModifier $ detectCycleOnAddition psk
+
+    -- This is inefficient. Consider supporting this map
+    -- in-memory or changing mempool key to stakeholderId.
+    let _cmPskMods = HM.fromList $
+            map (bimap addressHash pskToDlgEdgeAction) $
+            HM.toList cyclePool
+        -- Not used since we can't have more than one psk per issuer
+        -- in mempool and "has posted this epoch" is fully backed up
+        -- by the database.
```
gromakovsky: I am not sure it's ok. What we if have a PSK from `x` and we are processing another PSK from `x`? I might be missing something, but it seems that verification will be considered successful, but PSK won't be inserted. Which is not too bad, but it will, for example, make this PSK relayed to others while it probably shouldn't. Am I right?

node/src/Pos/Delegation/Cede/Logic.hs
neongreen: That's true. @volhovM _mda_

node/src/Pos/Delegation/Cede/Class.hs
```diff
@@ -19,16 +19,33 @@ import           Pos.Delegation.Cede.Types (DlgEdgeAction (..))
 import           Pos.Types                 (ProxySKHeavy, StakeholderId, addressHash)
 
 
--- | This monad includes psk reading capabilities.
+-- | This monad abstracts data needed for verifying headers/blocks
+-- from the standpoint of delegatino component.
```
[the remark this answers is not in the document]
neongreen: I really like `delegatino`

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -164,8 +172,15 @@ classifyHeaders inRecovery headers = do
...
-                       verifyHeaders (headers & _Wrapped %~ toList)
+    leaders <-
+        lrcActionOnEpochReason
+            newestHeaderEpoch
+            (sformat ("classifyHeaders: there are no leaders for epoch " %build)
+                     newestHeaderEpoch)
+            LrcDB.getLeaders
+    let headersValid =
+            isVerSuccess $
+            verifyHeaders (Just leaders) (headers & _Wrapped %~ toList)
```
gromakovsky: 1. Is it ok that we don't use `dlgVerifyHeader` in this function?
2. Is it guaranteed that all headers are from the same epoch? If not, then passing only one `SlotLeaders` argument can't work I suppose.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -164,8 +172,15 @@ classifyHeaders inRecovery headers = do
     tipHeader <- DB.getTipHeader @ssc
     let tip = headerHash tipHeader
     haveOldestParent <- isJust <$> DB.blkGetHeader @ssc oldestParentHash
-    let headersValid = isVerSuccess $
-                       verifyHeaders (headers & _Wrapped %~ toList)
+    leaders <-
+        lrcActionOnEpochReason
+            newestHeaderEpoch
+            (sformat ("classifyHeaders: there are no leaders for epoch " %build)
+                     newestHeaderEpoch)
```
gromakovsky: Putting information about epoch into reason is probably 100% useless because `LrcDataUnknown` has `EpochIndex` inside and is formatted this way: 
```
    build (LrcDataUnknown epoch reason) =
        bprint
            ("LRC data isn't presented for epoch #"%int%
             " so raise the exception with reason: "%stext)
            epoch reason
```

node/src/Pos/Block/Logic/Header.hs
gromakovsky: I see that `verifyHeaders` will use given leaders for given blocks until it reaches `GenesisBlock`. But you pass `newestHeader ^. epochIndexL`. Probably you should at least change it to `oldestHeader ^. epochIndexL`. But it's probably better to pass leaders for all epochs.
P. S. Maybe I just don't know something and blocks are guaranteed to be from the same epoch here.

node/src/Pos/Delegation/Logic/Mempool.hs
volhovM: No-no, you're missing the point. This comment says that having an extra layer of modifications over database is redudantant. 

Say that `x` has posted in this epoch before current slot. Now `x` tries to add one more cert. Checking function will understand that he did post cert using DB only. Same applies if `x` didn't post the cert. Mempool corresponds to the single block modifications and single block can't have duplicates.

node/src/Pos/Block/Logic/Header.hs
volhovM: Thanks, will fix.

node/src/Pos/Block/Logic/Header.hs
volhovM: 1. It's okay, since we anyway can't verify proxy signatures if we have >1 header. See PR comment.
2. No, it is not guaranteed. Yes, there is no total safety here (and can't be).

`newest ~ oldest` is a bug, thank you. There is no point in passing "leaders for all epochs" because it's not possible (I'm not sure though, maybe it will work for [9k,10k] period, when do we compute lrc?) to know leaders of future epoch in advance. Note: `classifyHeaders` is used in recovery mode now (since default mode is incremental and we check only one header).

node/src/Pos/Delegation/Cede/Class.hs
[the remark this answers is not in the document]
volhovM: I will leave it like this then :p

node/src/Pos/Delegation/Logic/Mempool.hs
gromakovsky: Can you clarify this check:
>         existsSameMempool <- uses dwProxySKPool $ \m -> HM.lookup iPk m == Just psk

Why do we check equality here? What if someone whose PSK is in mempool sends another PSK? Will we accept it? Doesn't it make possible for someone to spam with heavyweight PSKs which will be relayed again and again?

node/src/Pos/Block/Logic/Header.hs
gromakovsky: > There is no point in passing "leaders for all epochs" because it's not possible

On, well, yes. We should also ensure that lrc will be used before you call `lrcActionOnEpochReason`, otherwise it will just hang. I will check it.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -164,8 +178,14 @@ classifyHeaders inRecovery headers = do
     tipHeader <- DB.getTipHeader @ssc
     let tip = headerHash tipHeader
     haveOldestParent <- isJust <$> DB.blkGetHeader @ssc oldestParentHash
-    let headersValid = isVerSuccess $
-                       verifyHeaders (headers & _Wrapped %~ toList)
+    leaders <-
```
gromakovsky: Is it possible that we are classifying this headers before we verified genesis block for `oldestHeaderEpoch`? I can't proof that it's not. And if it's possible, then this line can hang, because it waits until LRC for `oldestHeaderEpoch` is finished.
If we don't hold `StateLock` here (I hope we don't) then I would recommend using `lrcSingleShot oldestHeaderEpoch` before calling `lrcActionOnEpochReason`. It should be harmless.

node/src/Pos/Delegation/Logic/Mempool.hs
gromakovsky: @volhovM ↑

volhovM: >What if someone whose PSK is in mempool sends another PSK? Will we accept it?

We will and it will overwrite previous one with the same issuerPk.

>Doesn't it make possible for someone to spam with heavyweight PSKs which will be relayed again and again?

We have caching for it. Also "someone" should have enough stake.

gromakovsky: Well, AFAIU, cache doesn't matter, because it will lead to big number of messages anyway.

Regarding «have enough stake»: well, an adversary with 5% of stake can already spam the network by sending PSKs again and again from multiple machines. I suppose it's possible to make so big traffic of PSKs that sending txs will become barely possible. I don't think it's good.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -112,23 +122,25 @@ classifyNewHeader (Right header) = do
...
+                _                 -> pass
+
+            dlgHeaderValid <- runDBCede $ runExceptT $ dlgVerifyHeader header
+            whenLeft dlgHeaderValid $ throwError . CHInvalid
+
+            pure CHContinues
         -- If header's parent is not our tip, we check whether it's
         -- more difficult than our main chain.
-        | tipHeader ^. difficultyL < header ^. difficultyL -> CHAlternative
+        | tipHeader ^. difficultyL < header ^. difficultyL -> pure CHAlternative
```
gromakovsky: So???
Once again, suppose someone who is not a leader sends us a more difficult block which isn't a continuation of our chain. It will trigger recovery which is bad.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -189,11 +199,22 @@ classifyHeaders inRecovery headers = do
                  ("Newest header is from slot "%build%", but current slot"%
                   " is "%build%" (and we're not in recovery mode)")
                  (newestHeader ^. epochOrSlotG) currentSlot
+       | isNothing leaders ->
```
gromakovsky: You should write a comment explaining why it usually can't happen.

node/src/Pos/Block/Network/Logic.hs
```diff
@@ -309,10 +310,20 @@ handleRequestedHeaders
...
         newestHash = headerHash newestHeader
-        oldestHash = headerHash $ headers ^. _Wrapped . _neLast
+        oldestHash = headerHash oldestHeader
+        oldestEpoch = oldestHeader ^. epochIndexL
+    -- Try to calculate LRC for the oldest header epoch. If we're in
+    -- recovery and oldest header is from the next epoch, no lrc will
+    -- be automatically calculated as all workers are locked in
+    -- recovery mode. So we should try to do it manually.
+    logDebug "handleREquesteHeaders: LRC started"
+    -- TODO EXCEPTION HANDLING HERE
```
gromakovsky: https://youtu.be/ZrNLv76sCJQ

georgeee:
Can be introduced later via software update (no protocol changes)

See: https://issues.serokell.io/issue/CSL-1152#comment=96-7358

node/src/Pos/Block/Logic/Header.hs
volhovM: No, it will not trigger the recovery. It will add a certain task to retrievalworker with this `h`. We will only enter the recovery after we get first k+ headers that lead us from our tip to `h`.

node/src/Pos/Block/Network/Logic.hs
```diff
@@ -309,10 +310,19 @@ handleRequestedHeaders
...
         newestHash = headerHash newestHeader
-        oldestHash = headerHash $ headers ^. _Wrapped . _neLast
+        oldestHash = headerHash oldestHeader
+        oldestEpoch = oldestHeader ^. epochIndexL
+    -- Try to calculate LRC for the oldest header epoch. If we're in
+    -- recovery and oldest header is from the next epoch, no lrc will
+    -- be automatically calculated as all workers are locked in
+    -- recovery mode. So we should try to do it manually.
+    logDebug "handleREquesteHeaders: LRC started"
+    void $ tryAny $ withStateLockNoMetrics LowPriority $ const $ lrcSingleShot oldestEpoch
```
gromakovsky: So where is exception handling? Apparently you just deleted the comment.
1. I believe it's ok to ignore `UnknownBlocksForLrc` here, but it's the only exception we can ignore here. All other exceptions should be at least logged (or maybe not caught at all).
2. Everytime you use `tryAny` or any other function which catches `ErrorCall` or `CardanoFatalError` you should use `reportOrLog` inside exception handler.

gromakovsky: And, to be honest, I would change this code to explicitly check our tip's slot and call LRC only when we know it will succeed (i. e. our tip's slot is from `oldestEpoch - 1` and it's after crucial slot). Also it makes sense to check chain quality, because see CSL-1660.
This way you don't need to take locks (and note that it will take some time to release it, because LRC loads up to `2k` blocks from the DB) and use exceptions for control flow.

node/src/Pos/Block/Logic/Header.hs
gromakovsky: Well, I still think this check should be done, but I don't mind if it's done in a separate issue.

volhovM: I see no need for this change. I've already explained why the "problem" you're trying to fix is not crucial.

node/src/Pos/Block/Network/Logic.hs
volhovM: @gromakovsky I've implement conditional calculation, though i didn't understand what did you mean by "check chain quality explicitly". I will amend if you decide that it's still needed.

node/src/Pos/Block/Network/Logic.hs
```diff
@@ -348,6 +342,28 @@ handleRequestedHeaders cont inRecovery headers = do
...
+            let handler UnknownBlocksForLrc =
+                    logWarning $
+                    "handleRequestedHeaders: tried lrcSingleShot, " <>
+                    "got UnknownBlocksForLrc"
+                handler e                   = throwM e
+            withStateLockNoMetrics LowPriority $ const $
+                lrcSingleShot oldestEpoch `catch` handler
+            logDebug "handleRequesteHeaders: LRC ended"
+        else logDebug $ sformat ("handleRequestHeaders: oldest header epoch ("%build%
+                                 ") /= tip epoch ("%build%" + 1)")
```
gromakovsky: This message looks a bit confusing. I mean `tip epoch (tipEpoch + 1)`.

node/src/Pos/Block/Network/Logic.hs
```diff
@@ -348,6 +342,28 @@ handleRequestedHeaders cont inRecovery headers = do
...
+    tryCalculateLrc = do
+        tipEpoch <- view epochIndexL <$> DB.getTipHeader @ssc
+        if oldestEpoch == tipEpoch + 1 then do
+            logDebug "handleRequesteHeaders: LRC started"
+            let handler UnknownBlocksForLrc =
+                    logWarning $
+                    "handleRequestedHeaders: tried lrcSingleShot, " <>
+                    "got UnknownBlocksForLrc"
+                handler e                   = throwM e
+            withStateLockNoMetrics LowPriority $ const $
```
gromakovsky: IMO, it should be `HighPriority`, because it's block processing.

node/src/Pos/Block/Network/Logic.hs
gromakovsky: @volhovM the thing is that now you still can get `UnknownBlocksForLrc` because LRC checks that there is at least one block after `crucialSlot`. So if tip is from slot `(e, 0)` and you receive a header from slot `(e + 1, 0)`, LRC will fail. So you should check that there is at least one block after crucial slot.
And according to CSL-1660, you should do a stricter check that there are at least `k` blocks. This check can be done without loading blocks, because we store slots of last `k` blocks. But I don't mind if you check presence of only 1 block and add a TODO with `CSL-1660` tag (also write about it in CSL-1660).

node/src/Pos/Block/Network/Logic.hs
volhovM: Fair point.

node/src/Pos/Block/Network/Logic.hs
volhovM: I've fixed both by amending to the last commit.

node/src/Pos/Block/Network/Logic.hs
```diff
@@ -338,6 +342,28 @@ handleRequestedHeaders cont inRecovery headers = do
...
   where
+    newestHeader = headers ^. _NewestFirst . _neHead
+    oldestHeader = headers ^. _NewestFirst . _neLast
+    newestHash = headerHash newestHeader
+    oldestHash = headerHash oldestHeader
+    oldestEpoch = oldestHeader ^. epochIndexL
+
+    tryCalculateLrc = do
+        tipEpoch <- view epochIndexL <$> DB.getTipHeader @ssc
+        if oldestEpoch == tipEpoch + 1 then do
```
gromakovsky: @volhovM it seems that my comment is still correct: https://github.com/input-output-hk/cardano-sl/pull/1540#discussion_r139650150

If tip is before crucial slot, LRC will fail and we will log a warning, but the situation is quite normal.
Why not check that tip's slot is after crucial slot? And add reference to CSL-1660.

node/src/Pos/Block/Network/Logic.hs
volhovM: @gromakovsky please check it again, i've added it.

node/src/Pos/Block/Network/Logic.hs
```diff
@@ -348,6 +344,41 @@ handleRequestedHeaders cont inRecovery headers = do
...
+        let tipEpoch = tip ^. epochIndexL
+        let differentEpochs = oldestEpoch == tipEpoch + 1
+        -- CSL-1660 We should also check that there's k blocks after
+        -- crucial slot.
+        let tipAfterCrucial = case unEpochOrSlot tipEpochOrSlot of
+                -- we assume differentEpochs is true, so this is
+                -- before zero's slot of epoch e, while new header has
+                -- epoch e+1.
+                Left _   -> False
+                Right si -> siSlot si > siSlot (crucialSlot $ siEpoch si)
```
gromakovsky: It seems more natural to pass `siEpoch si + 1`, but it doesn't matter here, because you compare `siSlot`s eventually.

node/src/Pos/Delegation/Cede/Holders.hs
```diff
@@ -59,12 +60,18 @@ evalMapCede = flip Ether.evalLazyStateT
...
+        ether $ use (cmHasPostedThisEpoch . at sId) >>= \case
+            Nothing                -> lift $ DB.isIssuerPostedThisEpoch sId
+            Just v                 -> pure v
 
 instance MonadDBRead m => MonadCede (MapCede m) where
     modPsk eAction = do
         let issuer = dlgEdgeActionIssuer eAction
-        ether $ identity %= HM.insert issuer eAction
+        ether $ cmPskMods %= HM.insert issuer eAction
+    addThisEpochPosted sId = ether $ cmHasPostedThisEpoch %= HM.insert sId True
```
pva701: why didn't you use `Set` instead of `Map` of Bool?

node/src/Pos/Delegation/Logic/Mempool.hs
```diff
@@ -206,10 +207,11 @@ processProxySKHeavyInternal psk = do
...
         -- This is inefficient. Consider supporting this map
         -- in-memory or changing mempool key to stakeholderId.
-        let cedeModifier =
-                HM.fromList $
+        let _cmPskMods = HM.fromList $
                 map (bimap addressHash pskToDlgEdgeAction) $
                 HM.toList cyclePool
+            _cmHasPostedThisEpoch = mempty -- not used
+            cedeModifier = CedeModifier {..}
         in evalMapCede cedeModifier $ detectCycleOnAddition psk
```
pva701: AFAIU you can use `CedeModifier {..}` directly here

node/src/Pos/Delegation/Cede/Holders.hs
```diff
@@ -59,12 +60,18 @@ evalMapCede = flip Ether.evalLazyStateT
 
 instance MonadDBRead m => MonadCedeRead (MapCede m) where
```
pva701: I've already asked why constraint isn't `MonadCedeRead m`, IIRC you agreed it was better to use `MonadCedeRead`

node/src/Pos/Delegation/Cede/Holders.hs
pva701: ok I got it, IMHO it would be better to use `MapModifier Stakeholer ()` b/c such construction is used for similar things and hence it's more intuitive, but it's subjective.

node/src/Pos/Delegation/Cede/Holders.hs
volhovM: Can't/don't remember me saying that. I either implement what you say and do not implement `MonadDBRead (MayCede m)` constraint or vice versa. I chose option 1. They are more or less the same imho.

volhovM:
Github shows 2 "in progress" builds on appveyor, but in fact they are green if you check them on the appveyor site directly. So I'm merging this one.
@pva701 thanks for your comments.

node/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -344,26 +335,24 @@ dlgVerifyBlocks ::
...
         HashSet StakeholderId ->
         Block ssc ->
-        ExceptT Text (MapCede (StateT DlgVerState m)) DlgUndo
+        ExceptT Text (MapCede m) DlgUndo
     verifyBlock _ (Left genesisBlk) = do
-        prevThisEpochPosted <- use dvCurEpoch
-        dvCurEpoch .= HS.empty
         let blkEpoch = genesisBlk ^. epochIndexL
+        prevThisEpochPosted <- getAllPostedThisEpoch
+        mapM_ delThisEpochPosted prevThisEpochPosted
```
pva701: can you explain this code? do you want to set to an empty set of `posted this epoch certs`? why?

PR #1542, opened 2017-09-11
[CSL-1332] Don't allow certificates with duplicate VSS keys

neongreen:


neongreen:
Ok, fixing tests.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
```diff
@@ -244,11 +248,31 @@ getCertId = addressHash . vcSigningKey
 
 -- | VssCertificatesMap contains all valid certificates collected
 -- during some period of time.
-type VssCertificatesMap = HashMap StakeholderId VssCertificate
+newtype VssCertificatesMap = VssCertificatesMap
```
gromakovsky: Our convention is to either not export constructor or call it with `Unsafe` prefix when it has some internal invariants. In this particular case we want `addressHash (vcKey v) == k` for each `(k, v)` in map.

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
```diff
@@ -244,11 +248,31 @@ getCertId = addressHash . vcSigningKey
...
+insertVss c (VssCertificatesMap m) =
+    VssCertificatesMap (HM.insert (getCertId c) c m)
+
+deleteVss :: StakeholderId -> VssCertificatesMap -> VssCertificatesMap
+deleteVss id (VssCertificatesMap m) = VssCertificatesMap (HM.delete id m)
+
+-- | Safe constructor of 'VssCertificatesMap'.
+--
+-- (Well, almost safe: it doesn't check that all certificates have different
+-- 'vcVssKey's).
```
gromakovsky: 😔

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
```diff
@@ -468,10 +473,22 @@ checkCertificatesPayload
...
+        getVssCertificates
+    exceptGuardSnd CertificateDuplicateVssKey
+        (not . (`HS.member` existingVssKeys) . vcVssKey)
+        (HM.toList (getVssCertificatesMap certs))
+    let newVssKeys =
+            HS.fromList . map vcVssKey . toList . getVssCertificatesMap $
+            certs
+    exceptGuardSnd CertificateDuplicateVssKey
+        (not . (`HS.member` newVssKeys) . vcVssKey)
+        (HM.toList (getVssCertificatesMap certs))
```
gromakovsky: This check seems to always fail.
Also it's probably useless, because for block processing it's ensured by deserialization and for data processing it's ensured by the check above.
But I guess it shouldn't be expensive at all, so I don't mind having it here (but it should be fixed, I am 99% sure).

neongreen:
Commits that should be reviewed particularly carefully:

* [[CSL-1332] Generate valid 'VssCertificatesMap's in Arbitrary](https://github.com/input-output-hk/cardano-sl/pull/1542/commits/ee7bec5ddc79e81235c94a9a58b06ea273aa2c9a)
* [[CSL-1332] Fix a wrong check in 'checkCertificatesPayload'](https://github.com/input-output-hk/cardano-sl/pull/1542/commits/f63a841fe04a9d6d5d3fedb7f9f1c88b6a48e7fe)
* [[CSL-1332] Fix tests that were broken after adding 'insertVss'](https://github.com/input-output-hk/cardano-sl/pull/1542/commits/7bc431f9003b6cd29afe90d2a45b73ce98157df3)
* [[CSL-1332] Fix some tests by deleting certs even from improper VCD](https://github.com/input-output-hk/cardano-sl/pull/1542/commits/8b75fef2346f16fb6709ace86d244a9817b2b4d7)

core/Pos/Util/Util.hs
```diff
@@ -62,6 +63,13 @@ module Pos.Util.Util
        -- * Aeson
        , parseJSONWithRead
 
+       -- * QuickCheck
+       , qcIsLeft
+       , qcIsRight
+       , qcElem
+       , qcNotElem
+       , qcFail
```
gromakovsky: AFAIK we put such stuff into `Test.Pos.Util`.

core/Pos/Util/Util.hs
```diff
@@ -109,6 +117,8 @@ import           Data.HashMap.Strict            (HashMap)
 import qualified Data.HashMap.Strict            as HM
 import           Data.HashSet                   (fromMap)
 import           Data.List                      (last)
+import           Data.Map                       (Map)
```
gromakovsky: AFAIK it's exported from `Universum`.

godtossing/Pos/Arbitrary/Ssc/GodTossing.hs
```diff
@@ -151,14 +153,11 @@ instance Arbitrary GtPayload where
...
         makeSmall $
         oneof
-            [ CommitmentsPayload <$> arbitrary <*> genVssCerts
-            , OpeningsPayload <$> arbitrary <*> genVssCerts
-            , SharesPayload <$> arbitrary <*> genVssCerts
-            , CertificatesPayload <$> genVssCerts
+            [ CommitmentsPayload <$> arbitrary <*> arbitrary
+            , OpeningsPayload <$> arbitrary <*> arbitrary
+            , SharesPayload <$> arbitrary <*> arbitrary
+            , CertificatesPayload <$> arbitrary
```
gromakovsky: 👍

godtossing/Pos/Ssc/GodTossing/Core/Core.hs
```diff
@@ -318,22 +311,26 @@ mkGtProof payload =
...
         OpeningsPayload <$> openings <*> certs
     (SharesPayload shares0 certs0) -> do
-        let certs = stripHashMap limCerts certs0
+        let certs = stripVss limCerts certs0
         let shares = stripHashMap (lim - biSize certs) shares0
         SharesPayload <$> shares <*> certs
   where
     limCerts = lim `div` 3 -- certificates are 1/3 less important than everything else
                            -- this is a random choice in fact
+    stripVss l = fmap UnsafeVssCertificatesMap .
```
gromakovsky: I believe it would be good to write a comment here explaining why it's safe (it's quite obvious, but still).

godtossing/Pos/Ssc/GodTossing/Core/Types.hs
gromakovsky: Perhaps making it safe **and** total is too much for now, but I would make it partial then and use `error` if invariant is violated. I believe it's safer. And this check should be cheap, because we won't have many participants in GodTossing anyway.

georgeee:
This can be added post-mainnet:

For initial months we'll control whole network and thus we ensure same there won't be more than one *active* vss cert with same vss key:
  a) Genesis block contains distinct vss keys in vss certs
  b) Node will renew these certs when they expire. Same vss key will be used, but it will be put in place of already expired vss cert
  c) Only core nodes are eligible to commit vss certs

For Shelley we sure must implement this issue. Earlier is better so we double-ensure same vss keys won''t be committed.

Note, have more than one *active* vss cert with same vss key will make SSC break.

*Inclusion in Shelley*

Given previous blockchain wasn't subject to having same vss keys in active vss certs, update prior to Shelley is easy:

1. Bump block version in update proposal
2. Include this fix in this software update

godtossing/Pos/Binary/GodTossing/Core.hs
```diff
@@ -163,3 +99,35 @@ decodeCommitments = do
     unless (allDistinct (map (view _1) comms :: [PublicKey])) $
         fail "decodeCommitments: two commitments have the same signing key"
     pure (mkCommitmentsMap comms)
+
+----------------------------------------------------------------------------
+-- TH-generated instances go to the end of the file
+----------------------------------------------------------------------------
+
+deriveSimpleBi ''GtPayload [
```
pva701: :facepalm:

pva701: tbh I forgot it can generate instances for datatype for several constructors, thx for reminder

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
```diff
@@ -461,6 +461,10 @@ checkSharesPayload epoch shares = do
 -- For certificates we check that
 --   * certificate hasn't been sent already
 --   * certificate is generated by richman
+--   * there is no existing certificate with the same VSS key
+--   * the map doesn't contain certificates with the same VSS key (this
```
pva701: thb it's unclear which certificates you talking about, it would be better to write explicitly.

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
pva701: as said Vanya it seems to always fail, moreover I don't see a necessity to check this statement for each map separately, b/c AFAIU map can't contain duplicated elements, however, they can contain duplicates in union

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
```diff
@@ -483,12 +484,17 @@ checkCertificatesPayload epoch certs = do
...
-    let newVssKeys =
-            HS.fromList . map vcVssKey . toList . getVssCertificatesMap $
-            certs
-    exceptGuardSnd CertificateDuplicateVssKey
-        (not . (`HS.member` newVssKeys) . vcVssKey)
-        (HM.toList (getVssCertificatesMap certs))
+    -- find pubkeys corresponding to vss keys that are duplicated
+    let isDupGroup (_:_:_) = True
+        isDupGroup _       = False
+    let keysWithDups =
```
pva701: it seems this check won't work b/c two *different* stakeholder can't have two same certificates because vss certificate includes _public key_ and default `Eq` instance for `VssCertificate` will say vss certificate have different public keys, hence they aren't equal.
so we should rewrite `Eq` for `VssCertificate` or something like that.

godtossing/Pos/Ssc/GodTossing/VssCertData.hs
```diff
@@ -82,15 +82,21 @@ lookupExpiryEpoch id mp = vcExpiryEpoch <$> lookup id mp
...
-delete id mp@VssCertData{..}
-    | Just (ins, expiry) <- lookupEoSes id mp = VssCertData
-          lastKnownEoS
-          (deleteVss id certs)
-          (HM.delete id whenInsMap)
-          (S.delete (ins, id) whenInsSet)
-          (S.delete (expiry, id) whenExpire)
-          expiredCerts
-    | otherwise = mp
+delete id mp@VssCertData{..} =
```
pva701: it looks ok.

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
pva701: ok, I am wrong

neongreen:
@gromakovsky @pva701 rerereview please.

gromakovsky:
@neongreen synchronize with master and I will rereview.

neongreen:
@gromakovsky done.

gromakovsky:
> The Travis CI build failed

core/Pos/Core/Vss.hs
```diff
@@ -71,22 +86,75 @@ checkCertSign VssCertificate {..} =
...
+-- with duplicate signing keys or with duplicate 'vcVssKey's.
+mkVssCertificatesMap
+    :: MonadFail m
+    => [VssCertificate] -> m VssCertificatesMap
+mkVssCertificatesMap certs = do
+    unless (allDistinct (map vcSigningKey certs)) $
+        fail "mkVssCertificatesMap: two certs have the same signing key"
+    unless (allDistinct (map vcVssKey certs)) $
+        fail "mkVssCertificatesMap: two certs have the same VSS key"
+    pure $ UnsafeVssCertificatesMap (HM.fromList (map toCertPair certs))
```
gromakovsky: I would write `pure $ mkVssCertificatesMapLossy certs` here to slightly decrease code duplication (no need to define `toCertPair` in two places).

core/Pos/Core/Vss.hs
```diff
@@ -71,22 +86,75 @@ checkCertSign VssCertificate {..} =
...
+    map toCertPair . nubOrdOn vcVssKey
   where
     toCertPair vc = (getCertId vc, vc)
 
+-- | A map with a single certificate is always valid so this function is
+-- safe to use in case you have one certificate and want to create a map
+-- from it.
+mkVssCertificatesMapSingleton :: VssCertificate -> VssCertificatesMap
+mkVssCertificatesMapSingleton vc =
+    UnsafeVssCertificatesMap $ HM.singleton (getCertId vc) vc
```
gromakovsky: `mkVssCertificatesMapLossy [vc]` or `mkVssCertificatesMapLossy . one` might be a bit better, but it's subjective and minor.

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
```diff
@@ -472,10 +481,27 @@ checkCertificatesPayload
...
     richmenSet <- getKeys <$> (note (NoRichmen epoch) =<< getRichmen epoch)
     exceptGuardM CertificateAlreadySent
-        (notM hasCertificateToss) (HM.keys certs)
+        (notM hasCertificateToss) (HM.keys (getVssCertificatesMap certs))
     exceptGuardSnd CertificateNotRichmen
         ((`HS.member` richmenSet) . addressHash . vcSigningKey)
-        (HM.toList certs)
+        (HM.toList (getVssCertificatesMap certs))
+    existingVssKeys <-
+        HS.fromList . map vcVssKey . toList . getVssCertificatesMap <$>
```
gromakovsky: `getVssCertificatesMap` is probably useless, because `toList` is defined for `VssCertificatesMap` too (AFAIK), but it's not really essential.

godtossing/Pos/Ssc/GodTossing/Toss/Base.hs
```diff
@@ -464,6 +468,11 @@ checkSharesPayload epoch shares = do
 -- For certificates we check that
 --   * certificate hasn't been sent already
 --   * certificate is generated by richman
+--   * there is no existing certificate with the same VSS key in
+--     MonadToss's state
+--   * all certificates in the 'VssCertificatesMap' have distinct VSS keys
+--      (this check is useless for certificate maps obtained through
+--      deserialization, but not useless in other cases)
```
gromakovsky: Do we still need this check here, given that it's now invariant of `VssCertificatesMap` type?

gromakovsky: Shouldn't be needed.

godtossing/Pos/Ssc/GodTossing/Toss/Trans.hs
```diff
@@ -76,7 +75,7 @@ instance MonadToss m =>
     putShares id sh =
         ether $ tmShares . at id .= Just sh
     putCertificate cert =
-        ether $ tmCertificates %= HM.insert (getCertId cert) cert
+        ether $ tmCertificates %= fst . insertVss cert
```
gromakovsky: It seems to be ok to discard the second part of result here, because it should be always empty (ensured by `checkCertificatesPayload`). It would be good to write a brief comment explaining it.

core/Pos/Core/Vss/Types.hs
```diff
@@ -60,4 +62,12 @@ instance Hashable VssCertificate where
...
 -- | VssCertificatesMap contains all valid certificates collected
 -- during some period of time.
-type VssCertificatesMap = HashMap StakeholderId VssCertificate
+--
+-- Invariants:
+--   * stakeholder ids correspond to 'vcSigningKey's of associated certs
+--   * no two certs have the same 'vcVssKey'
+newtype VssCertificatesMap = UnsafeVssCertificatesMap
+    { getVssCertificatesMap :: HashMap StakeholderId VssCertificate }
+    deriving (Eq, Show, Generic, NFData, Monoid, Container, NontrivialContainer)
```
gromakovsky: `Monoid` looks wrong to me, `mappend` is dangerous, because it can break one invariant.

core/Pos/Core/Vss.hs
neongreen: I don't want to do this because `nubOrdOn` isn't cheap and here we can avoid it.

neongreen: Okay, I will define `toCertPair` as a top-level function because it can be used in three places.

gromakovsky: Deal.

core/Pos/Core/Vss/Types.hs
neongreen: I'm going to fix `Monoid` instead of removing it entirely, because we use `<>` in some places.

neongreen:
@gromakovsky don-er

PR #1546, opened 2017-09-12
[CSL-1617] Use JSON for genesis files

gromakovsky:
This PR is basically a continuation of #1530 ported to newest master.


core/Pos/Aeson/Core.hs
```diff
@@ -0,0 +1,62 @@
...
+module Pos.Aeson.Core
+       (
+       ) where
+
+import           Universum
+
+import           Data.Aeson       (ToJSON (toJSON), FromJSON (..))
+import           Data.Aeson.TH    (defaultOptions, deriveToJSON, deriveJSON, deriveFromJSON)
+import           Data.Time.Units  (Millisecond, Microsecond)
+import           Serokell.Util.Base64  (JsonByteString (..))
```
gromakovsky: Why Base64? I believe we agreed to use Base16 :/

core/Pos/Core/Genesis/Parser.hs
```diff
@@ -40,9 +42,9 @@ allGenesisSpecs =
                 [|(suff, path, $(embedFile path))|]
             )
     in HM.fromList $ bins <&> \(suff, path, file) ->
-           case error "TODO: parse JSON here" file of
+           case decodeEither file of
```
gromakovsky: There is one more `error` in this module where `decodeEither` should be used.

core/Pos/Aeson/Core.hs
```diff
@@ -0,0 +1,62 @@
...
+
+instance FromJSON SharedSeed where
+    parseJSON v = SharedSeed <$> parseJSON v
+
+instance FromJSON (AsBinary w) where
+    parseJSON v = AsBinary <$> parseJSON v
+
+deriving instance ToJSON SlotCount
+
+-- NOTE: some of these types are used on frontend (PureScript).
```
gromakovsky: Oh, so we are sharing these instances with PureScript, it's sad :(

core/Pos/Aeson/Crypto.hs
```diff
@@ -17,5 +22,26 @@ instance ToJSON (AbstractHash algo a) where
 instance HashAlgorithm algo => FromJSON (AbstractHash algo a) where
     parseJSON = parseJSONWithRead
 
+instance (HashAlgorithm algo, FromJSON (AbstractHash algo a))
+         => FromJSONKey (AbstractHash algo a) where
+    fromJSONKey = FromJSONKeyTextParser parser
+      where
+        parser = either (fail . T.unpack) pure . decodeAbstractHash
```
gromakovsky: There is `eitherToFail`.

core/Pos/Aeson/Genesis.hs
```diff
@@ -0,0 +1,60 @@
...
+instance FromJSON AvvmEntry where
+    parseJSON = withObject "avvmEntry" $ \o -> do
+        aeCoin <- (* (1000000 :: Integer)) <$> o .: "coin"
+        (addrText :: Text) <- o .: "address"
+        aePublicKey <- fromAvvmPk addrText
+        return AvvmEntry{..}
+
+instance FromJSON AvvmData
+
+instance FromJSON (ProxySecretKey EpochIndex)
```
gromakovsky: Why is it here and not in `Pos.Crypto`?

core/Pos/Crypto/Signing.hs
```diff
@@ -170,6 +171,12 @@ fullSignatureHexF :: Format r (Signature a -> r)
 fullSignatureHexF = later $ \(Signature x) ->
     B16.formatBase16 . CC.unXSignature $ x
 
+-- | Parse 'Signature' from base16 encoded string.
+parseFullSignature :: Text -> Maybe (Signature a)
```
gromakovsky: Why does it return `Maybe` instead of `Either` with error message? It doesn't make sense to me.

core/Pos/Crypto/Signing.hs
```diff
@@ -232,6 +239,17 @@ verifyProxyCert issuerPk (PublicKey delegatePk) o (ProxyCert sig) =
         (mconcat ["00", CC.unXPub delegatePk, Bi.serialize' o])
         (Signature sig)
 
+-- | Formatter for 'ProxyCert' to show it in hex.
+fullProxyCertHexF :: Format r (ProxyCert a -> r)
+fullProxyCertHexF = later $ \(ProxyCert x) ->
+    B16.formatBase16 . CC.unXSignature $ x
+
+-- | Parse 'ProxyCert' from base16 encoded string.
+parseFullProxyCert :: Text -> Maybe (ProxyCert a)
```
gromakovsky: Same here, `Maybe` is strange.

core/Pos/Aeson/Core.hs
```diff
@@ -0,0 +1,62 @@
...
+deriveFromJSON defaultOptions ''EpochIndex
+deriveFromJSON defaultOptions ''Millisecond
+deriveFromJSON defaultOptions ''Microsecond
+deriveFromJSON defaultOptions ''Timestamp
+deriveFromJSON defaultOptions ''CoinPortion
+deriveFromJSON defaultOptions ''SoftforkRule
+deriveFromJSON defaultOptions ''Coeff
+deriveFromJSON defaultOptions ''TxSizeLinear
+deriveFromJSON defaultOptions ''TxFeePolicy
+deriveFromJSON defaultOptions ''BlockVersionData
```
gromakovsky: Where did it come from actually? I see you added this stuff, but didn't remove it anywhere. It's suspicious.

core/Pos/Aeson/Core.hs
akegalj: @gromakovsky client-api (ie, purescript) will be soon (in a month) be removed and deprecated

gromakovsky: @akegalj does it apply to explorer too? Does explorer use these instances or its own wrappers?

akegalj: ah right. Not sure atm - I believe they are used. @sectore @ksaric do you know are these instances used by Explorer UI ?

Long term plan is also rewriting Explorer UI in javascript, but I don't think this will happen in 2017. (maybe around December)

ksaric: @akegalj AFAIK the Aeson instances for `explorer` are here - https://github.com/input-output-hk/cardano-sl/blob/master/explorer/src/Pos/Explorer/Aeson/ClientTypes.hs
The client types are here - https://github.com/input-output-hk/cardano-sl/blob/master/explorer/src/Pos/Explorer/Web/ClientTypes.hs

akegalj: @gromakovsky If it is problematic that they are shared with purescript, we can duplicate those instances for both Explorer and Middleware . Then they won't be shared.

core/Pos/Aeson/Core.hs
pva701: done.

core/Pos/Aeson/Genesis.hs
pva701: because pos.crypto will import `pos.aeson.core`, it will cause cycl dep

core/Pos/Crypto/Signing.hs
pva701: lol, I just copied parser for public key

core/Pos/Aeson/Core.hs
pva701: it's used for bytesting b/c there is convenient `JsonByteString` for base64 and I decided it doesn't matter we use `base16` or `base64` but using this wrapper is less error-prone

core/Pos/Crypto/Signing.hs
pva701: will fix them

core/Pos/Aeson/Genesis.hs
gromakovsky: But why do you need to import whole `Pos.Crypto` and not only parts of it?
Anyway, the fact that `Pos.Crypto` imports `Pos.Aeson.Core` is quite strange and I think it's wrong.

core/Pos/Crypto/Signing.hs
gromakovsky: Parser for public key was strange too.

core/Pos/Aeson/Core.hs
```diff
@@ -0,0 +1,61 @@
...
+
+instance FromJSON (AsBinary w) where
+    parseJSON v = AsBinary . getJsonByteString <$> parseJSON v
+
+deriving instance ToJSON SlotCount
+
+instance FromJSON CoinPortion where
+    parseJSON v = unsafeCoinPortionFromDouble <$> parseJSON v
+
+deriveJSON defaultOptions ''Coin
```
gromakovsky: It should also be below `NOTE` (it's used by frontend).
Also I will be glad if you rebase and make all aeson instances changes in one commit, because it's hard to understand what happened with them in this PR.

explorer/src/explorer/Params.hs
```diff
@@ -69,15 +67,7 @@ getNodeParams args@Args {..} systemStart = do
...
 
-    let devBalanceDistr =
-            devBalancesDistr
-                (CLI.flatDistr commonArgs)
-                (CLI.richPoorDistr commonArgs)
-                (CLI.expDistr commonArgs)
-
-    let npGenesisCtx
-            | isDevelopment = devGenesisContext devBalanceDistr
-            | otherwise = genesisContextProduction
```
avieth: A mistake to remove this? How do we control when a development genesis context is used?

ksaric: We don't need this anymore, I have a new change (https://issues.serokell.io/issue/CSE-217) that I'll create soon that uses `Pos.Client.CLI.getNodeParams`.

pva701: we will generate a development genesis context like genesis for testnet from seed, hence we won't consider development mode separately and will remove all predefined dev fields

core/Pos/Core/Genesis/Types.hs
```diff
@@ -155,37 +161,153 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
+
+-- | Genesis data contains all data which determines consensus
+-- rules. It must be same for all nodes. It's used to initialize
+-- global state, slotting, etc.
+data GenesisData = GenesisData
+    { gdBootStakeholders :: !GenesisWStakeholders
+    , gdHeavyDelegation  :: !GenesisDelegation
+    , gdStartTime        :: !Timestamp
+    , gdVssCerts         :: !VssCertificatesMap
+    , gdNonAvvmBalances  :: !(HashMap Address Coin)
```
avieth: This seems to be the only thing that `GenesisData` has which `GenesisSpec` with a `MainnetInitializer` does not have. Should such a `HashMap Address Coin` also go in `MainnetInitializer`?

gromakovsky: No, in mainnet only AVVM users have coins.

core/Pos/Core/Genesis/Testnet.hs
```diff
@@ -0,0 +1,217 @@
+-- | Generation of genesis data for testnet.
+
+module Pos.Core.Genesis.Testnet
```
avieth: Rename to `Pos.Core.Genesis.Generate` or similar? It's not just about testnet, since mainnet data can be generated too.

gromakovsky: Right, it's planned and it's exactly the name I proposed in our private chat.

core/Pos/Core/Genesis/Testnet.hs
```diff
@@ -0,0 +1,217 @@
...
+    -- ^ Address distribution for non avvm addresses
+    , ggdBootStakeholders :: !GenesisWStakeholders
+    -- ^ Set of boot stakeholders (richmen addresses or custom addresses)
+    , ggdGtData           :: !VssCertificatesMap
+    -- ^ Genesis vss data (vss certs of richmen)
+    , ggdSecretKeys       :: !(Maybe [(SecretKey, EncryptedSecretKey, VssKeyPair)])
+    -- ^ Secret keys for non avvm addresses
+    , ggdFakeAvvmSeeds    :: !(Maybe [ByteString])
+    -- ^ Fake avvm seeds (needed only for testnet)
+    }
```
avieth: There's non avvm distribution, fake avvm seeds. What about real avvm stuff?

pva701: @avieth we don't generate real avvm stuff, it's known, but we don't know real avvm seeds, only redeem public keys

core/Pos/Core/Genesis/Testnet.hs
avieth: That's what I thought after looking at the exports. But after reading through the file I wonder is it really useful for generating mainnet data? I see mentions only of fake avvm, but surely in mainnet there's no "fake" data.

pva701: Actually, we don't generate mainnet data AFAIU, but we use it for dev mode too, so it's reasonable.

core/Pos/Core/Genesis/Testnet.hs
avieth: Ok. So this `GeneratedGenesisData` is useful / relevant only for testnet after all, right? We would never use this to construct a "real" genesis with "real" avvm data?

core/Pos/Core/Genesis/Testnet.hs
```diff
@@ -0,0 +1,217 @@
...
+    -- Calculate actual balances
+    desiredRichBalance = getShare tboRichmenShare testBalance
+    oneRichmanBalance = desiredRichBalance `div` richs +
+        if desiredRichBalance `mod` richs > 0 then 1 else 0
+    realRichBalance = oneRichmanBalance * richs
+    poorsBalance = testBalance - realRichBalance
+    onePoorBalance = if poors == 0 then 0 else poorsBalance `div` poors
+    realPoorBalance = onePoorBalance * poors
+
+    mpcBalance = getShare (coinPortionToDouble Const.genesisMpcThd) testBalance
```
avieth: `genesisMpcThd` isn't actually a constant, it's a configuration which depends upon a genesis which isn't known until run time (made clear in https://github.com/input-output-hk/cardano-sl/pull/1550).

So this is puzzling: in order to `genTestnetDistribution`, we need to have some genesis value. But presumably the testnet distribution is going to be used to define the genesis, right?

avieth: i.e. there may be a cycle lurking here which would cause looping: `genTestnetDistribution` will hang on the `newInitFuture` `MVar` which it will attempt to fill after generating a genesis from that distribution. In #1550 this would be eliminated at compile time due to an unsatisfied constraint.

gromakovsky: @avieth this module generates `GenesisData` (actually part of it), while `genesisMpcThd` is known from `GenesisSpec`. So there is now loop here.
`GenesisSpec` is initialized by `applyConfigInfo`.

avieth: I see. Good!
So do I understand correctly that there must always be a `GenesisData`, which is the single source of truth for the application, and there are precisely two ways of getting it:

```
-- Parse it from the canonical JSON
mainNetGenesisData :: FilePath -> IO (Either Error GenesisData)
-- Generate it from a spec
testNetGenesisData :: GenesisSpec -> Either Error GenesisData
```

?

gromakovsky: More or less yes.

For now we just want to support testnet way when we have `genesis-spec.yaml` (or, if we take your branch's approach, just some file on disk) with serialized `GenesisSpec`, parse it as soon as possible after application start and then use `generateTestnetOrMainnetData` to generate `GenesisData`. We are focused on making it work in all cases, it's CSLREQ-213.

After that we are planning two support two modes: passing either encoded `GenesisSpec` or encoded `GenesisData`. It's described in CSLREQ-214. There is some outdated info there, specifically, the fields from (1) shouldn't be optional, they are part of `GenesisSpec`. In `constants.yaml` we are going to have `genesis: genesis-mainnet.json` or `  genesisSpec: genesis-spec-testnet.yaml`.

It was our plan so far. I know that in your branch there is only one configuration file, so probably some adjustments are necessary.

core/Pos/Core/Genesis/Canonical.hs
```diff
@@ -0,0 +1,435 @@
...
+            ]
+
+instance Monad m => ToJSON m GenesisAvvmBalances where
+    toJSON = toJSON . getGenesisAvvmBalances
+
+instance Monad m => ToJSON m SharedSeed where
+    toJSON (SharedSeed seed) = pure . JSString . formatToString base16F $ seed
+
+instance Monad m => ToJSON m BlockVersionData where
+    toJSON (BlockVersionData scriptVersion slotDuration maxBlockSize maxHeaderSize maxTxSize maxProposalSize mpcThd heavyDelThd updateVoteThd updateProposalThd updateImplicit softforkRule txFeePolicy unlockStakeEpoch) =
```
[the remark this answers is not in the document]
pva701: Lol, Vanya wrote it not to forget field when it would be added: compiler will say you forgot field in constructor and you will be forced to fix this instance. if you use wildcard you probably will forget to fix this instance

core/Pos/Core/Genesis/Parser.hs
```diff
@@ -1,73 +1,72 @@
...
-genCoreData :: GenesisCoreData
-setGenCoreData :: GenesisCoreData -> IO ()
-(genCoreData, setGenCoreData) =
-    unsafePerformIO (newInitFuture "genCoreData")
-{-# NOINLINE genCoreData #-}
-{-# NOINLINE setGenCoreData #-}
+genesisSpec :: GenesisSpec
+setGenesisSpec :: GenesisSpec -> IO ()
+(genesisSpec, setGenesisSpec) =
+    unsafePerformIO (newInitFuture "genesisSpec")
```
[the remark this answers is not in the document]
pva701: see here
https://github.com/input-output-hk/cardano-sl/pull/1550

core/Pos/Core/Genesis/Types.hs
```diff
@@ -155,37 +161,155 @@ bootDustThreshold (GenesisWStakeholders bootHolders) =
...
+      -- | Magic constant for separating real/testnet.
+    , pcProtocolMagic :: !Int32
+      -- | VSS certificates max timeout to live (number of epochs).
+    , pcVssMaxTTL     :: !Word32
+      -- | VSS certificates min timeout to live (number of epochs).
+    , pcVssMinTTL     :: !Word32
+    } deriving (Show, Eq, Generic)
+
+-- | Specification how to generate full genesis data.
+data GenesisSpec = UnsafeGenesisSpec
```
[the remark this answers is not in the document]
pva701: it named `GenesisData` (suppose you meant this datatype)

core/Pos/Core/Genesis/Generate.hs
```diff
@@ -0,0 +1,217 @@
...
+import           Pos.Core.Vss               (VssCertificate, VssCertificatesMap,
+                                             mkVssCertificate)
+import           Pos.Crypto                 (EncryptedSecretKey, RedeemPublicKey,
+                                             SecretKey, VssKeyPair, deterministic,
+                                             emptyPassphrase, keyGen, randomNumberInRange,
+                                             redeemDeterministicKeyGen, safeKeyGen,
+                                             toPublic, toVssPublicKey, vssKeyGen)
+
+-- | Data generated by @genTestnetOrMainnetData@ using genesis-spec.
+data GeneratedGenesisData = GeneratedGenesisData
```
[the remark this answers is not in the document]
pva701: because avvm data isn't _generated_

core/genesis-spec-bn.yaml
```diff
@@ -0,0 +1,46 @@
+initializer:
```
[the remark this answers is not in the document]
pva701: `bn` stands for `benchmark`

lrc/Pos/Lrc/Genesis.hs
```diff
@@ -2,11 +2,13 @@ module Pos.Lrc.Genesis
...
+import           Pos.Core.Genesis.Parser (genesisSpec)
+import           Pos.Core.Genesis.Types  (gsFtsSeed)
+import           Pos.Core.Types          (SharedSeed (..))
 
 -- | Seed that will be used for the 0th epoch. We must hardcode a seed
 -- because we need to somehow determine leaders for the first ever epoch
 -- (stakes are hardcoded as well so we can run FTS on those stakes using this
 -- seed).
 genesisSeed :: SharedSeed
-genesisSeed = SharedSeed "vasa opasa skovoroda Ggurda boroda provoda"
```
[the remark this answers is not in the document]
pva701: now, it didn't rip. `ftsSeed: "c2tvdm9yb2RhIEdndXJkYSBib3JvZGEgcHJvdm9kYSA="` this string in base64 equals this legendary seed. @gromakovsky controlled it would be left in the code. he encoded it in `base16`, `base64`, `base58` and etc for all possible cases. I hope he wrote a test which validates this seed equals this string.

core/Pos/Core/Genesis/Canonical.hs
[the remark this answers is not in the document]
gromakovsky: Yes, exactly. I actually could have used `RecordWildCards` here, because we also have tests for this stuff, but it's safer to have extra protection apart from tests.

PR #1549, opened 2017-09-12
[CSE-192] Speed up genesis address info

Pastafarianist:
This PR adds a new in-memory cache for storing data that is required by Explorer's "genesis info" page. A follow-up to CSE-192 is CSE-229 where we should store the number of currently redeemed addresses in Explorer DB.


ksaric:
Interesting.
Going to reuse your `ExtraContext`, just what I was looking for :smiling_imp:

explorer/src/Pos/Explorer/Web/ExtraContext.hs
```diff
@@ -0,0 +1,48 @@
...
+import qualified Data.Vector   as V
+import qualified Ether
+
+import           Pos.Context   (NodeContext)
+import           Pos.Core      (Address, Coin, isRedeemAddress)
+import           Pos.Genesis   (GenesisUtxo (..))
+import           Pos.Txp       (utxoToAddressCoinPairs)
+import           Pos.Util.Util (lensOf)
+
+type ExtraContextT m = Ether.ReaderT' ExtraContext m
```
flyingleafe: Why a regular `ReaderT` isn't sufficient here? 🤔

Pastafarianist: Because it is used in the context of

```haskell
type ExplorerProd = ExtraContextT (ExplorerBListener (RealMode SscGodTossing))
```

and

```haskell
type RealMode ssc = Mtl.ReaderT (RealModeContext ssc) Production
```

so there would be a conflict between two `ReaderT`s.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -526,21 +516,21 @@ getGenesisSummary
     :: ExplorerMode ctx m
     => m CGenesisSummary
 getGenesisSummary = do
-    redeemAddressCoinPairs <- getRedeemAddressCoinPairs
-    cgsNumRedeemed <- length <$> filterM (uncurry isAddressRedeemed) redeemAddressCoinPairs
-    pure CGenesisSummary {cgsNumTotal = length redeemAddressCoinPairs, ..}
+    grai <- getGenesisRedeemAddressInfo
+    cgsNumRedeemed <- V.length <$> V.filterM (uncurry isAddressRedeemed) grai
```
gromakovsky: `length` in `Universum` is polymorphic, btw (maybe you know it already and prefer to be more explicit).

explorer/src/Pos/Explorer/Web/Transform.hs
```diff
@@ -3,36 +3,48 @@
...
+import           Pos.Explorer.Socket.App   (NotifierSettings, notifierApp)
+import           Pos.Explorer.Web.Server   (explorerApp, explorerHandlers,
+                                            explorerServeImpl)
 
 -----------------------------------------------------------------
 -- Transformation to `Handler`
 -----------------------------------------------------------------
 
-type ExplorerProd = ExplorerBListener (RealMode SscGodTossing)
+type ExplorerProd = ExtraContextT (ExplorerBListener (RealMode SscGodTossing))
```
gromakovsky: FPC's recommendation was not to use `ether`. The way to avoid `ether` here is to define `ExplorerContext` and use `ReaderT ExplorerContext Production` instead of `RealMode SscGodTossing`.
However, I don't mind using `ether`, so I don't insist on changing it. It's explorer after all :)

ksaric: :unamused:

explorer/src/Pos/Explorer/Web/Server.hs
Pastafarianist: Thank you, I didn't know that. Replaced `V.length` with `length`.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -537,74 +539,72 @@ getTxSummary cTxId = do
...
-
-    let addressCoinPairs :: [(Address, Coin)]
-        addressCoinPairs = utxoToAddressCoinPairs (unGenesisUtxo genesisUtxo)
-
-        redeemOnly :: [(Address, Coin)]
-        redeemOnly = filter (isRedeemAddress . fst) addressCoinPairs
-
-    pure redeemOnly
+data GenesisSummaryInternal = GenesisSummaryInternal
+    { gsiNumRedeemed            :: !Int
```
ksaric: It's either a `Bool` or the number of redeemed coins.
Later you refer it as `isRedeemed`.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -537,74 +539,72 @@ getTxSummary cTxId = do
...
+        => Address -> Coin -> m GenesisSummaryInternal
     getRedeemAddressInfo address initialBalance = do
         currentBalance <- fromMaybe (mkCoin 0) <$> EX.getAddrBalance address
         if currentBalance > initialBalance then
             throwM $ Internal $ sformat
                 ("Redeem address "%build%" had "%build%" at genesis, but now has "%build)
                 address initialBalance currentBalance
         else
-            let isRedeemed = currentBalance == mkCoin 0
+            let isRedeemed = if currentBalance == mkCoin 0 then 1 else 0
```
ksaric: zeroCoins = mkCoin 0? It's repeating all over the place.

explorer/src/Pos/Explorer/Web/Server.hs
Pastafarianist: I've added a comment to clarify the usage of this field.

explorer/src/Pos/Explorer/Web/Server.hs
Pastafarianist: Replaced with `minBound`.

PR #1550, opened 2017-09-13
[CSL-1617] JSON Genesis block format

avieth:
Previously there was a "raw" CoreConfig, which existed just to avoid a
cyclic dependency induced by SignTag, which uses the protocol magic
value defined by the core configuration: SignTag could safely import the
raw configuration, but the configuration with domain-specific types was
out-of-bounds because it depended on crypto stuff.

Here I've done some factoring to remove the cyclic dependencies and
give one configuration which uses domain-specific types.

I've also got rid of the 'coreConfig' top-level constant. We now use
Data.Reflection (the HasConfiguration constraint) to make the
configuration available. Instead of parsing some yaml at compile time,
we read it and parse it at run-time and use it to come up with a
'Configuration' value to reify/reflect through 'HasConfiguration'.
This is safer and simpler than the 'newInitFuture' approach.

These changes will aid in bringing in the custom genesis information.
All of the machinery by which a 'GenesisSpec' is generated can be
self-contained, completely orthogonal to the configuration system
(actually, it *must not* use any of the configuration system, because
'GenesisSpec' is a part of the configuration). Once a 'GenesisSpec' is
created, by whatever means (default hard-coded, read from file,
generated from distribution) it can be used to come up with the complete
'Configuration' value.

## Update September 14

No longer WIP. Ready to merge into `pva-and-gromak/csl1617-json-genesis`.

Configuration is never read at compile time, nor is the CSL_SYSTEM_TAG env
variable sourced at compile time. Everything is controlled by a configuration
yaml. See `conguration.yaml`, which currently has only a development setup
but will ultimately obsolete the `constants.yaml`.

A note on terminology: I refer to something which is constant for the lifetime of
the executable as a constant, whereas something which is constant for the
lifetime of the process is a configuration. So we have comparatively few
constants and comparatively many configuration values.


core/Pos/Configuration.hs
```diff
@@ -0,0 +1,392 @@
...
+
+----------------------------------------------------------------------------
+-- Config itself
+----------------------------------------------------------------------------
+
+data Configuration = Configuration
+    {
+      -- | Specifies the genesis
+      -- FIXME check for redundancies and remove them (GenesisSpec may contain
+      -- values also defined later in Configuration).
```
gromakovsky: Yes, actually many of these values are part of `GenesisSpec`.

neongreen:
I haven't reviewed particularly carefully (and indeed there are some outdated comments and questions like the one about `Pos.Binary.Core.Address` that should be answered), but I'm okay with these changes in general.

neongreen:
Though I think I'll dismiss my [✔️] because it's WIP after all.

neongreen:
> The plan is to use Data.Reflection (the HasConfiguration constraint) to make the
> configuration available. Instead of parsing some yaml at compile time,
> we'll read it and parse it at run-time

I think that we might have to keep an embedded config available anyway because @georgeee said that deploying >1 file is a pain. cc @georgeee

gromakovsky:
@avieth there was a force-push to `pva-and-gromak/csl1617-json-genesis`, please rebase.

gromakovsky:
> By the way, I think that we might have to keep an embedded config available anyway 

No, actually we don't need to embed it, because devops won't use this feature anyway (it's written in a YT comment).
A little bit more details: long-term (well, ideally long should be 1-2 days) is to embed only genesis hash optionally. All the configuration will be written in runtime and hash of `GenesisData` will be compared with embedded `genesisHash` (if it's embedded at all, note that it's optional).

core/Pos/Constants.hs
```diff
@@ -0,0 +1,37 @@
+{-# LANGUAGE CPP #-}
+
+-- | Non-configurable constants
+--   For configurable constants, see Pos.Configuration.
+
+module Pos.Constants
```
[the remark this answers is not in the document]
avieth: I'll put it back in `Pos.Core.Constants`.

avieth:
> @avieth there was a force-push to pva-and-gromak/csl1617-json-genesis, please rebase.


Done

core/Pos/Core/Address.hs
```diff
@@ -60,23 +60,22 @@ import           Serokell.Util           (mapJson)
 
 import           Pos.Binary.Class        (Bi)
 import qualified Pos.Binary.Class        as Bi
-import           Pos.Binary.Core.Address ()
```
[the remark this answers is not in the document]
avieth: Does what depend on it?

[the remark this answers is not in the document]
gromakovsky: Does `Pos.Binary.Core.Address` transitively depend on `Pos.Core.Address`? Otherwise I don't get why this import was removed.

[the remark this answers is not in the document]
avieth:
```
    Module imports form a cycle:
             module ‘Pos.Core.Genesis.Types’ (./Pos/Core/Genesis/Types.hs)
            imports ‘Pos.Core.Address’ (./Pos/Core/Address.hs)
      which imports ‘Pos.Binary.Core.Address’ (./Pos/Binary/Core/Address.hs)
      which imports ‘Pos.Binary.Core.Types’ (./Pos/Binary/Core/Types.hs)
      which imports ‘Pos.Core.Configuration’ (./Pos/Core/Configuration.hs)
      which imports ‘Pos.Core.Genesis.Types’ (./Pos/Core/Genesis/Types.hs)
```

godtossing/Pos/Ssc/GodTossing/Configuration.hs
```diff
@@ -0,0 +1,21 @@
...
+    , gtConfiguration
+    , withGtConfiguration
+    ) where
+
+import           Pos.Ssc.GodTossing.Genesis.Types
+import           Data.Reflection                  (Given (..), give)
+
+type HasGtConfiguration = Given GtConfiguration
+
+type GtConfiguration = GenesisGtData
```
gromakovsky: It will become unneeded soon, because `GenesisGtData` will be removed. We moved `VssCertificatesMap` into `cardano-sl-core`, so now this `GenesisGtData` is either part of `GenesisSpec` or can be uniquely generated from it.

avieth: Could it be that we'll need the configuration in order to produce the configuration? The protocol magic number is not known until we get the genesis spec, but it's needed  in order to make `VssCertificate`.

avieth: It might be possible to define a lazy version of `Data.Reflection.give`... `giveFix :: forall a r. (a -> a) -> (Given a => r) -> r`

gromakovsky: No, `GenesisSpec` is exactly what is stored in a yaml file. Obtaining `GenesisSpec` is simple `Data.Yaml.decodeEither`. So there shouldn't be a problem.
In some cases it stores `VssCertificate`, but in this case we just put it in some format (determined `ToJSON` instances) into yaml file.

avieth:
Some more commentary in this big commit https://github.com/input-output-hk/cardano-sl/pull/1550/commits/d5324e7046cb6df9ae83b923923d0bd39519b9e4

The executables are nearly up-to-date.

Major TODO is to finish `Pos.Launcher.Configuration.withConfigurations`: update its type to something like the `ConfigInfo` type and parse the yamls at these file paths into the actual configuration. No need for `unsafePerformIO`!

avieth:
@gromakovsky @neongreen @pva701 This is ready for merge (see updated description).

Yes it's massive. It touches basically everything because the global mutable variables from `newInitFuture` are gone, so a configuration constraint (a `Given` from `Data.Reflection`) arises everywhere that one of these configuration values is used. This is a good thing: the type actually mentions its dependencies.

Some redundancies were removed from the configuration. Check out `configuration.yaml` for an example of how it looks now. This file is parsed at run time, never at compile time, typically close to `main`.

The next step is to actually use the `GenesisData`. I see we have canonical to/from JSON instances for it. Note that these instances now depend upon the configuration (`HasConfiguration` constraint) because it includes `VssCertificates`, which need the protocol magic number for parsing. So even in mainnet, where the genesis file defines the protocol magic, the configuration has to give the protocol magic as well.

gromakovsky:
@avieth personally I really like what you've done and of course that's what we ideally want to have, but according to @georgeee we are not going to include this change into mainnet (and into master) for now, because it's not strictly necessary.
So I am not going to review it unless @georgeee or @dcoutts tell us that we are going to have this PR in mainnet. Not because I don't like this PR, but because at the current stage it's not for mainnet, so it's not worth spending time on it.

> because it includes VssCertificates, which need the protocol magic number for parsing

Oh, that's a good catch which I didn't realize earlier. So parsing `GenesisSpec` requires `GenesisSpec` 😔 
Well, it can be resolved by making JSON deserialization a bit less strict and not checking signatures there.

avieth:
> Oh, that's a good catch which I didn't realize earlier. So parsing GenesisSpec requires GenesisSpec :pensive:
> Well, it can be resolved by making JSON deserialization a bit less strict and not checking signatures there.

No it's actually not so bad! `GenesisSpec` doesn't include any vss certificates so it can be parsed without any configuration. It's `GenesisData` which needs the protocol magic. So we could parse the `GenesisSpec`, then parse the `GenesisData`, and then perhaps do some consistency checks between them,

gromakovsky:
> GenesisSpec doesn't include any vss certificates so it can be parsed without any configuration.

No, it does include vss certificates in some cases:
https://github.com/input-output-hk/cardano-sl/blob/89db5c118908a3364b1972300f3b76a611c12e87/core/Pos/Core/Genesis/Types.hs#L195
https://github.com/input-output-hk/cardano-sl/blob/89db5c118908a3364b1972300f3b76a611c12e87/core/Pos/Core/Genesis/Types.hs#L214

avieth:
> No, it does include vss certificates in some cases:

The `FromJSON VssCertificate` instance derived in `Pos.Aeson.Core` apparently doesn't need the protocol magic. This is the one used by `GenesisSpec`. The canonical `FromJson m VssCertificate` however does require protocol magic, because it uses `recreateVssCertificate`.

gromakovsky:
> The FromJSON VssCertificate instance derived in Pos.Aeson.Core apparently doesn't need the protocol magic. This is the one used by GenesisSpec. The canonical FromJson m VssCertificate however does require protocol magic, because it uses recreateVssCertificate.

Aha, yes, indeed.
Well, canonical instance can be changed not to use `recreateVssCertificate` if necessary.

gromakovsky:
@avieth btw, I wanted to check if your branch works, but it doesn't compile for me:
```
--  While building package cardano-sl-update-0.6.0 using:
      /home/gromak/.stack/setup-exe-cache/x86_64-linux-nopie/Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2 --builddir=.stack-work/dist/x86_64-linux-nopie/Cabal-1.24.2.0 build lib:cardano-sl-update --ghc-options " -ddump-hi -ddump-to-file"
    Process exited with code: ExitFailure 1
    Logs have been written to: /home/gromak/Work/cardano-sl/.stack-work/logs/cardano-sl-update-0.6.0.log

    Configuring cardano-sl-update-0.6.0...
    Preprocessing library cardano-sl-update-0.6.0...
    Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2: can't find source for
    Pos/Aeson/Update in .,
    .stack-work/dist/x86_64-linux-nopie/Cabal-1.24.2.0/build/autogen
```

avieth:
Oops forgot to add that file. Just committed and pushed.

gromakovsky:
@avieth did you try running it (simple demo or benchmarks)?
In b254169a5 demo works, but 029f62de6 breaks it. I am going to investigate.

avieth:
> @avieth did you try running it (simple demo or benchmarks)?
In b254169 demo works, but 029f62d breaks it. I am going to investigate.

I have it working locally. Neither of those commits are mine.

avieth:
> I have it working locally. Neither of those commits are mine.

I mean to say, the demo works at tip of this branch

gromakovsky:
> Neither of those commits are mine.

Yes, sure, I am just saying that it worked at some point :)

> I mean to say, the demo works at tip of this branch

Well, now that I found out why 029f62de6 doesn't work I find it a bit strange that your branch works. The reason why 029f62de6 doesn't work is that slot duration in config is written in seconds, but it's expected in milliseconds. And I see the same in your branch: https://github.com/input-output-hk/cardano-sl/blob/avieth/csl1617_config/core/genesis-spec-dn.yaml#L16

avieth:
> Well, now that I found out why 029f62d doesn't work I find it a bit strange that your branch works. The reason why 029f62d doesn't work is that slot duration in config is written in seconds, but it's expected in milliseconds. And I see the same in your branch: https://github.com/input-output-hk/cardano-sl/blob/avieth/csl1617_config/core/genesis-spec-dn.yaml#L16

Heh, I came across that issues too. This branch does not use `genesis-spec-dn.yaml`. Everything is in `configuration.yaml`. The example in this branch uses 20000 ms.

avieth:
> This branch does not use genesis-spec-dn.yaml. Everything is in configuration.yaml. 

I did this because I'm under the impression that devops prefers to have as few files as possible.

avieth:
@gromakovsky heads up: I just noticed that min and max VSS TTL are inverted (max < min) in the `genesis-spec-dn.yaml` and `genesis-spec-bn.yaml` files.

gromakovsky:
Now in `pva-and-gromak/csl1617-json-genesis` demo in dev mode works, but in benchmarks mode 
```
    transportE <-
        liftIO $ TCP.createTransport addrInfo tcpParams
```
hangs in `Pos.Launcher.Resource`.
I am trying to figure out why.

avieth:
> I am trying to figure out why.

What's the difference between benchmarks mode and dev mode? Just the choice of genesis spec yaml?

gromakovsky:
> What's the difference between benchmarks mode and dev mode? Just the choice of genesis spec yaml?

No, there is also a flag in `cardano-sl-core.cabal` which determines value of `isDevelopment` variable.

What I know so far is that `NodeParams` are not properly constructed. Adding `!` to `currentParams` in `Main.hs` makes this line hang:
>         !currentParams <- CLI.getNodeParams cArgs nArgs systemStart

gromakovsky:
> No, there is also a flag in cardano-sl-core.cabal which determines value of isDevelopment variable.

Oh, sorry, benchmarks use dev mode (cabal flag I mean).
Another different is `dev-custom-config` flag, but probably it only affects the choise of `genesis-input`.
@avieth did you try running benchmarks in your branch?

gromakovsky:
Well, the problem turned out to be that in benchmarks mode there are 50000 poor nodes and creating 50000 secret keys takes eternity.

avieth:
> @avieth did you try running benchmarks in your branch?

Building now. To be sure, you mean `scripts/bench/buildbench.sh` and `scripts/bench/runbench.sh` correct?

gromakovsky:
Yes

avieth:
> Well, the problem turned out to be that in benchmarks mode there are 50000 poor nodes and creating 50000 secret keys takes eternity.

Oh. That's a relief (sort of).

avieth:
> Well, the problem turned out to be that in benchmarks mode there are 50000 poor nodes and creating 50000 secret keys takes eternity.

Is it this slow on master too? I seem to recall it used to take about 30 seconds on my laptop but now it's much longer.

gromakovsky:
No, it's faster in master. Also it turned out that if I wait for few minutes it eventually finishes (I was afraid it could take hours or days).

avieth:
@gromakovsky @pva701 @dcoutts 

Just about everything is in place at tip of this branch (459c398bcc76cfbba514edab6eb54290ce88f4f2).
The configuration yaml genesis key can be either 'spec' or 'src'. For 'src' it'll be the canonical JSON file, which is blake2b_256 hashed and checked against `Pos.Core.Constants.genesisDataDigest`, parsed to `GenesisData`, then used to come up with a `GenesisSpec`. If it's a 'spec' then the `GenesisSpec` is parsed directly. So `GenesisSpec` is the real deal, and `GenesisData` is just one avenue to get there (the one that will be used for mainnet).

The old `genesisHash`, which is used to identify the block header with no parent and was the digest of `"patak"`, is now (rightly) a configuration rather than a constant: when using a `GenesisSpec` given directly in configuration, the hash will be that digest of `"patak"`, but when a canonical JSON is given, the `genesisHash` is the digest of the file. So it plays a double role as consistency checker and special header hash value.

core/Pos/Core/Configuration.hs
```diff
@@ -0,0 +1,442 @@
...
+    -> (HasConfiguration => m r)
+    -> m r
+withConfiguration conf@Configuration{..} mSystemStart act = case ccGenesis of
+    -- If a 'GenesisData' source file is given, we check its hash against the
+    -- built-in constant, parse it, and we're done.
+    GCSrc fp -> do
+        !bytes <- liftIO $ BS.readFile fp
+
+        let digest :: Hash.Digest Hash.Blake2b_256
+            digest = Hash.hash bytes
```
gromakovsky: You could've used `Pos.Crypto.Hashing.hash` here.

core/Pos/Core/Configuration.hs
```diff
@@ -0,0 +1,442 @@
...
+        when (digest /= genesisDataDigest) $
+            throwM $ GenesisHashMismatch (sformat shown digest) (sformat shown genesisDataDigest)
```
gromakovsky: `sformat shown` is `show`.

core/Pos/Core/Configuration.hs
```diff
@@ -0,0 +1,442 @@
...
+                                             GenesisInitializer (..), GenesisHash (..),
+                                             GenesisData, genesisSpecFromData)
+import           Pos.Crypto.Hashing         (Hash, unsafeHash, AbstractHash (..))
+
+-- | The Configuration (probably from a yaml) and the system start time (which
+-- is not necessarily read from the Configuration).
+type HasConfiguration =
+    ( Given Configuration
+      -- System start time.
+    , Given Timestamp
```
gromakovsky: I really want it to be a newtype.

core/Pos/Core/Constants.hs
```diff
@@ -1,5 +1,55 @@
...
+Just genesisDataDigest = digestFromByteString @_ @BS.ByteString bytesDigest
+  where
+    hexDigest = "0e5751c026e543b2e8ab2eb06099daa1d1e5df47778f7787faab45cdf12fe3a8"
+    (bytesDigest, _) = B16.decode hexDigest
+ 
+{-# INLINE genesisDataDigest #-}
+
+-- | @True@ if current mode is 'Development'.
+--
+-- FIXME put it in Pos.Core.Configuration and don't use CPP.
```
gromakovsky: We want to get rid of it completely.

core/Pos/Core/Genesis.hs
```diff
@@ -26,88 +18,71 @@ module Pos.Core.Genesis
...
 
 ----------------------------------------------------------------------------
 -- Data taken from 'GenesisSpec'
 ----------------------------------------------------------------------------
 
+-- FIXME avieth
+-- define these all in Pos.Core.Configuration
+--
+-- FIXME avieth
+-- why is 'Prod' present in these names?
```
gromakovsky: Removed already in our branch.

core/Pos/Core/Genesis/Canonical.hs
```diff
@@ -51,6 +53,14 @@ import           Pos.Crypto                 (ProxyCert, ProxySecretKey (..), Pub
 -- Primitive standard/3rdparty types
 ----------------------------------------------------------------------------
 
+instance (Monad m, Applicative m, MonadFail m) => ReportSchemaErrors m where
```
gromakovsky: I would make it OVERLAPPABLE.

core/Pos/Core/Configuration.hs
georgeee: No, we need serialization to mess up things here!

gromakovsky: Sorry, I meant `hashRaw`.

avieth:
I reviewed commit-by-commit from 99f1d2b38753328657ace073616dfb27e5b71c8a to 9201ba2acc46d1d3b6bf06a915c2e33b65539473 and I'm quite happy with it all. My only questions/concerns are

1. We're embedding the configuration at comile-time for tests and benchmarks. What's the benefit of doing this? Is it inconvenient to load it at runtime as we do for the main executable?
2. ~~There's a TODO in 9526e1e6cd245be7d665f9e02e34874cbec40866 but I see now that's been resolved in later commits.~~

core/Pos/Core/Genesis/Generate.hs
```diff
@@ -2,128 +2,149 @@
...
-
-    let skVssCerts = map (\(sk, _, _, vc, _) -> (sk, vc)) $ richmenList ++ poorsList
-    let richSkVssCerts = take (fromIntegral tboRichmen) skVssCerts
-    let secretKeys = map (\(sk, hdwSk, vssSk, _, _) -> (sk, hdwSk, vssSk)) $ richmenList ++ poorsList
+        (,) <$> replicateM (fromIntegral tboRichmen)
+                           (generateSecretsAndAddress Nothing tboUseHDAddresses)
+            <*> replicateM (fromIntegral tboPoors)
+                           (generateSecretsAndAddress Nothing tboUseHDAddresses)
+
+    let richSkVssCerts = map (\(sk, _, _, vc, _) -> (sk, vc)) $ richmenList
```
avieth: VSS certs for poor men not needed ever? I guess the assumption is that a generated genesis never gives any stake to poor men?

avieth:
Is there a reason for using javascript for genesis hashing and avvm-to-json. Not a big deal, just wondering if there's something more to it other than personal preference.

PR #1555, opened 2017-09-13
[CSL-1576] Improve blkOnNewSlot exception handling

volhovM:


georgeee:
This PR helps in situation when something bad (exception) happens in reporting mechanisms (error reporting, EKG) and it may make block creation not happen.

For mainnet, all blocks to be created on cluster and thus if chain quality would decrease, it will be immediately seen and action performed.

Reporting may fail due to bug and/or misconfiguration. But in general unlikely to break if worked once (on cluster launch).

So verdict: desribale, but unnecessary for mainnet

volhovM:
@gromakovsky I've separated workers.

gromakovsky:
Thank you.

PR #1560, opened 2017-09-14
[CSL-1619] Add MsgNo{Block,Headers} and related logic

volhovM:
> What does this PR introduce?

Headers/block communication didn't have the response option for fail case. Instead, server node just didn't reply. As a result, calling party couldn't distinguish between timeout and negative response which lead to ambiguity. Now server party replies with exact error if it occurs. 

> Did the protocol change

Yes. Old clients won't manage to deserialize the message (binary format changed). But from the logic standpoint changes are minor. 

> Why is it needed in the mainnet.

1. We will later may want to introduce node web interface for querying blocks and headers separately and unability to understand whether node didn't reply or failed is very important to the client.
2. Similarly, we may have problems with debugging network timeouts because in order to confirm it is a timeout we need to check server node's logs.

### DIFF
--- a/node/src/Pos/Binary/Communication.hs
+++ b/node/src/Pos/Binary/Communication.hs
@@ -40,37 +40,55 @@ deriveSimpleBi ''MsgGetBlocks [
     ]]
 
 instance (HasCoreConstants, SscHelpersClass ssc) => Bi (MsgHeaders ssc) where
-  encode (MsgHeaders b) = encode b
-  decode = MsgHeaders <$> decode
+    encode = \case
+        (MsgHeaders b) -> encodeListLen 2 <> encode (0 :: Word8) <> encode b
+        (MsgNoHeaders t) -> encodeListLen 2 <> encode (1 :: Word8) <> encode t
+    decode = do
+        enforceSize "MsgHeaders" 2
+        tag <- decode @Word8
+        case tag of
+            0 -> MsgHeaders <$> decode
+            1 -> MsgNoHeaders <$> decode
+            t -> fail $ "MsgHeaders wrong tag: " <> show t
 
 instance (HasCoreConstants, SscHelpersClass ssc) => Bi (MsgBlock ssc) where
-  encode (MsgBlock b) = encode b
-  decode = MsgBlock <$> decode
+    encode = \case
+        (MsgBlock b) -> encodeListLen 2 <> encode (0 :: Word8) <> encode b
+        (MsgNoBlock t) -> encodeListLen 2 <> encode (1 :: Word8) <> encode t
+    decode = do
+        enforceSize "MsgBlock" 2
+        tag <- decode @Word8
+        case tag of
+            0 -> MsgBlock <$> decode
+            1 -> MsgNoBlock <$> decode
+            t -> fail $ "MsgBlock wrong tag: " <> show t
 
 -- deriveSimpleBi is not happy with constructors without arguments
 -- "fake" deriving as per `MempoolMsg`.
 -- TODO: Shall we encode this as `CBOR` TkNull?
 instance Bi MsgSubscribe where
-  encode MsgSubscribe = encode (42 :: Word8)
-  decode = do
-    x <- decode @Word8
-    when (x /= 42) $ fail "wrong byte"
-    pure MsgSubscribe
+    encode MsgSubscribe = encode (42 :: Word8)
+    decode = do
+        x <- decode @Word8
+        when (x /= 42) $ fail "wrong byte"
+        pure MsgSubscribe
 
 ----------------------------------------------------------------------------
 -- Protocol version info and related
 ----------------------------------------------------------------------------
 
 instance Bi HandlerSpec where
-  encode input = case input of
-    ConvHandler mname        -> encodeListLen 2 <> encode (0 :: Word8) <> encodeKnownCborDataItem mname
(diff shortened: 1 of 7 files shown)


volhovM:
This is still WIP, do not merge!

 I was testing this change and have found a weird bug with unproper error handling. Thiss eems to be very important, i'm investigating.

georgeee:
Well, this is indeed not strictly necessary for mainnet, though I would ask @volhovM to expand on second commit (it was referred to some strange things observed with error handling, which are this things?)

Regarding inclusion post-mainnet: it will be a new endpoint if we want to change formats. But as far as I understand we were to refactor block handling in much more general way, so ... looks like there is no point merging this not now no later :(

@volhovM if there is some performance degradation or any other factor for mainnet, then please provide this commentary

volhovM:
This one is no longer WIP. So, the story: after I've added negative responses i decided to test them a little bit and discovered several bugs with a common source: exception handling in recovery worker was broken. Long story short: we expect that if failure happens inside the worker while we're in recovery, we catch in on the top level and then drop communication with recovery server node. But after #1393 was merged this predicate got broken: extra error handling (catch >> ignore) was added thus masking and hiding errors. This led to infinite loops on exceptions inside retrieval worker.

This PR together with feature fixes these bugs. And I tested negative responses also (throwing them in different places instead of responding correctly), seems to work as charm. It is absolutely crucial fix to have.

cc @georgeee

node/src/Pos/Binary/Communication.hs
```diff
@@ -40,37 +40,55 @@ deriveSimpleBi ''MsgGetBlocks [
...
+        ConvHandler mname        ->
+            encodeListLen 2 <> encode (0 :: Word8) <> encodeKnownCborDataItem mname
+        UnknownHandler word8 bs  ->
+            encodeListLen 2 <> encode word8 <> encodeUnknownCborDataItem bs
+    decode = do
+        enforceSize "HandlerSpec" 2
+        tag <- decode @Word8
+        case tag of
+            0 -> ConvHandler        <$> decodeKnownCborDataItem
+            _ -> UnknownHandler tag <$> decodeUnknownCborDataItem
```
avieth: It wasn't introduced in this change, but `decode` is not left-inverse to `encode` here: an encoded unknown handler with `0 :: Word8` will deserialize to conv handler.

node/src/Pos/Block/Network/Logic.hs
```diff
@@ -259,36 +262,47 @@ matchRequestedHeaders headers mgh@MsgGetHeaders {..} inRecovery =
...
-                handleRequestedHeaders cont inRecovery headers
-            MRUnexpected msg -> handleUnexpected headers msg
+    case mHeaders of
+        Nothing -> do
+            logWarning "requestHeaders: received Nothing as a response on MsgGetHeaders"
+            throwM $ DialogUnexpected $
+                sformat ("requestHeaders: received Nothing from "%build) nodeId
+        Just (MsgNoHeaders t) -> do
+            logWarning $ "requestHeaders: received MsgNoHeaders: " <> t
+            throwM $ DialogUnexpected $
```
avieth: Is this really unexpected/exceptional? I guess this request is only made if we've somehow discovered that this peer really does have the relevant headers. Still we might not want to throw, because it will appear as a failure to the outbound queue when in fact the peer is completely fine.

node/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -367,3 +370,23 @@ retrieveBlocks' i conv prevH endH = lift (recvLimited conv) >>= \case
...
+    => (t2 -> (t1 -> t -> NonEmpty x) -> m (Map NodeId (m b)))
+    -> t2
+    -> x
+    -> m b
+enqueueMsgSingle enqueue msg conv = do
+    results <- enqueue msg (\_ _ -> one conv) >>=
+               waitForConversations
+    case toList results of
+        [] ->      throwM $ DialogUnexpected "enqueueMsgSingle: nobody responded"
+        (_:_:_) -> throwM $ DialogUnexpected "enqueueMsgSingle: more than one node responded"
```
avieth: Misleading error messages here. It's not that nobody responded, it's that the queue decided not to contact anybody.

node/src/Pos/Block/Network/Logic.hs
volhovM: I do not understand your model. We either succeed to communicate with peer or not. If we have any problems with peer, we ignore him for a while. Recovery worker is written using `m ()` actions and every failure is represented by the exception. Exception handler in recovery worker kicks user out and finds new recovery server. Every failure to retrieve/process headers is an exception.

node/src/Pos/Block/Network/Retrieval.hs
volhovM: Fair.

volhovM: Fixed

node/src/Pos/Binary/Communication.hs
volhovM: How bad is it? Can we fix it in a separate PR?

volhovM:
@dcoutts I've left two commits only as we've agreed.

georgeee:
To be merged into master as part of cardano-sl-1.0

PR #1566, opened 2017-09-14
Bump time-warp library

kantp:
This includes two changes: a small fix to the edge node subscribing
functionality, and the function `bucketSpareCapacity`, which will be
used to determine how many nodes can still subscribe to a relay before
it is considered "full".


georgeee:

PR #1567, opened 2017-09-14
[CSL-1611] Add ThreadId logging back

ChShersh:


domenkozar:
Having a thread id in logs would help, but it's not really critical.

georgeee:

PR #1569, opened 2017-09-15
[CSL-1658] Regenerate genesis

neongreen:


gromakovsky:
Will be useless after #1625.

PR #1577, opened 2017-09-18
[CSL-1663] Depend on custom version of ekg-statsd

adinapoli-iohk:
This PR uses our custom fork on `ekg-statsd` to be able to send stats to `statsd` more frequently.

We are currently discussing with the `ekg-statsd` maintainers whether or not it would make sense to simply drop the optimisation the library is performing in sending only the diff of what's changed after every iteration, but in the meantime, using our custom version should achieve the same effect by using `defaultStatsOptions { fullFlushIterations = Just 1 }`


jmitchell:
The commit I just added should resolve the Travis stack2nix build. Previously failure: https://travis-ci.org/input-output-hk/cardano-sl/jobs/276830301

adinapoli-iohk:
Trying once again to close and reopen.

PR #1580, opened 2017-09-18
[CSL-1672] Set relay networking policy

kantp:
- increase the MaxAhead for messages to 200, to match the current
  mempool size
- set the rate limiting towards core nodes to 3 Hz

These settings led to a decent throughput of ~6 transactions per
second, while keeping the core network running stably even when the
relays were exposed to higher transaction rates.


georgeee:

PR #1584, opened 2017-09-18
[CSL-969] Use `scrypt` for passphrases hashing

Martoon-00:
We need to keep a hash of password in secret key to check whether password matches.
This PR switches from `blake256` hashing to `scrypt` with [secure-] randomly generated salt.

### DIFF
--- a/core/Pos/Crypto/SafeSigning.hs
+++ b/core/Pos/Crypto/SafeSigning.hs
@@ -46,8 +46,8 @@ import           Pos.Crypto.Signing    (ProxyCert (..), ProxySecretKey (..),
 import           Pos.Crypto.SignTag    (SignTag (SignProxySK), signTag)
 
 data EncryptedSecretKey = EncryptedSecretKey
-    { eskPayload :: !CC.XPrv
-    , eskHash    :: !S.EncryptedPass
+    { eskPayload :: !CC.XPrv          -- ^ Secret key itself
+    , eskHash    :: !S.EncryptedPass  -- ^ Hash of passphrase used for key creation
     }
 
 instance Show EncryptedSecretKey where
@@ -103,12 +103,13 @@ mkEncSecret pp payload =
 encToPublic :: EncryptedSecretKey -> PublicKey
 encToPublic (EncryptedSecretKey sk _) = PublicKey (CC.toXPub sk)
 
--- | Re-wrap unencrypted secret key as an encrypted one
+-- | Re-wrap unencrypted secret key as an encrypted one.
+-- NB: for testing purposes only
 noPassEncrypt
     :: Bi PassPhrase
     => SecretKey -> EncryptedSecretKey
 noPassEncrypt (SecretKey k) =
-    mkEncSecretWithSalt (S.mkSalt ("" :: Text)) emptyPassphrase k
+    mkEncSecretWithSalt def emptyPassphrase k
 
 checkPassMatches
     :: (Bi PassPhrase, Alternative f)
--- a/core/Pos/Crypto/SafeSigning.hs
+++ b/core/Pos/Crypto/SafeSigning.hs
@@ -75,13 +75,12 @@ instance Default PassPhrase where
     def = emptyPassphrase
 
 -- | Parameters used to evaluate hash of passphrase.
--- They influence on resulting hash length, memory and time consumption.
 passScryptParam :: S.ScryptParams
 passScryptParam =
     fromMaybe (error "Bad passphrase scrypt parameters") $
-    S.scryptParamsLen 14 8 1 hashLen  -- params recomended in documentation to scrypt
-  where
-    hashLen = 32  -- maximal passphrase length
+    S.mkScryptParams def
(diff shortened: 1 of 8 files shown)


src/Pos/Wallet/Web/Server/Methods.hs
```diff
@@ -0,0 +1,1100 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
+{-# LANGUAGE TypeOperators       #-}
+
+-- | Wallet web server.
+
+module Pos.Wallet.Web.Server.Methods
```
volhovM: This is a bug. `src` doesn't exist in master.

Martoon-00:
@volhovM Could you take another look on this?

core/Pos/Crypto/SafeSigning.hs
```diff
@@ -139,7 +168,7 @@ safeDeterministicKeyGen
     -> PassPhrase
     -> (PublicKey, EncryptedSecretKey)
 safeDeterministicKeyGen seed pp =
-    bimap PublicKey (mkEncSecret pp) (safeCreateKeypairFromSeed seed pp)
+    bimap PublicKey (mkEncSecretWithSalt (S.mkSalt seed) pp) (safeCreateKeypairFromSeed seed pp)
```
neongreen: So if the attacker knows the salt, they know the seed ⇒ they know the keypair. That seems really bad.

neongreen: Perhaps you can take a `(ByteString, ByteString)` instead.

volhovM: Definitiely, add one hash round.

neongreen: A hash will probably do too, yeah. `bimap PublicKey (mkEncSecretWithSalt (S.mkSalt (hash seed)) pp) (safeCreateKeypairFromSeed seed pp)` (it's important that we use the hash as a salt, not as a seed)

Martoon-00: Mad bug

core/Pos/Crypto/Scrypt.hs
```diff
@@ -0,0 +1,68 @@
+-- | Wrapper over scrypt library
+
+module Pos.Crypto.Scrypt
+    ( S.ScryptParams
```
volhovM: Wrong indentation, we use 7 spaces, remember?

core/Pos/Crypto/SafeSigning.hs
```diff
@@ -37,14 +39,15 @@ import           Universum
...
 import qualified Pos.Binary.Class      as Bi
 import           Pos.Crypto.Hashing    (Hash, hash)
+import qualified Pos.Crypto.Scrypt     as S
 import           Pos.Crypto.Signing    (ProxyCert (..), ProxySecretKey (..),
                                         PublicKey (..), SecretKey (..), Signature (..),
                                         sign, toPublic)
 import           Pos.Crypto.SignTag    (SignTag (SignProxySK), signTag)
 
 data EncryptedSecretKey = EncryptedSecretKey
     { eskPayload :: !CC.XPrv
```
volhovM: Though it's not mostly related to this change, can you put two documentation lines on these fields? I'm not sure i understand how this whole framework works :(
Also why field with type `EncryptedPass` is named `eskHash`? And why payload has type of a private key?

core/Pos/Crypto/SafeSigning.hs
```diff
@@ -72,33 +75,60 @@ emptyPassphrase = PassPhrase mempty
...
-
-checkPassMatches :: (Bi PassPhrase, Alternative f) => PassPhrase -> EncryptedSecretKey -> f ()
-checkPassMatches pp (EncryptedSecretKey _ pph) = guard (hash pp == pph)
-
-changeEncPassphrase
+noPassEncrypt
     :: Bi PassPhrase
+    => SecretKey -> EncryptedSecretKey
+noPassEncrypt (SecretKey k) =
+    mkEncSecretWithSalt (S.mkSalt ("" :: Text)) emptyPassphrase k
```
volhovM: This kind of salt seems to be insecure since its serialization is something really simple and predictable. I'm by no means a theoretical cryptographer, but stackoverflow suggests to use non-trivial salts. Why not put some `"kolbasa  selyodka shapka"` here instead? And tbh I would put the default salt into the separate top-level definition (even though it's only used here).

flyingleafe: @volhovM it actually doesn't matter at all which salt to use _here_, because here we provide empty password, which efficiently says "we don't want any security for this key"

volhovM:
All CIs were green before cosmetics. We believe cosmetics didn't break anything so I merged this PR (it was needed ASAP).

PR #1592, opened 2017-09-19
[CSM-460] Expose Daedalus backend IP:PORT via walletAddress flag

akegalj:
Required by Bittrex.

Issue here: https://issues.serokell.io/issue/CSM-460

### DIFF
--- a/node/src/Pos/Client/CLI/Options.hs
+++ b/node/src/Pos/Client/CLI/Options.hs
@@ -11,7 +11,7 @@ module Pos.Client.CLI.Options
        , portOption
        , timeLordOption
        , webPortOption
-       , walletPortOption
+       , daedalusWalletAddressOption
        , networkAddressOption
        , externalNetworkAddressOption
        , listenNetworkAddressOption
@@ -30,6 +30,7 @@ import           Serokell.Util                        (sec)
 import           Serokell.Util.OptParse               (fromParsec)
 
 import           Pos.Binary.Core                      ()
+import           Pos.Client.CLI.Util                  (sscAlgoParser)
 import           Pos.Communication                    (NodeId)
 import           Pos.Constants                        (isDevelopment, staticSysStart)
 import           Pos.Core                             (Timestamp (..))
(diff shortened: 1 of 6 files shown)


wallet/src/Pos/Wallet/Web/Server/Launcher.hs
```diff
@@ -46,11 +48,11 @@ import           Pos.Web                          (TlsParams, serveImpl)
 walletServeImpl
     :: MonadWalletWebMode m
     => m Application     -- ^ Application getter
-    -> Word16            -- ^ Port to listen
+    -> NetworkAddress    -- ^ IP and port to listen
```
Martoon-00: Eee, finally not just `Word16` is here

wallet/src/Pos/Wallet/Web/Server/Launcher.hs
```diff
@@ -46,11 +48,11 @@ import           Pos.Web                          (TlsParams, serveImpl)
...
     -> Maybe TlsParams
     -> m ()
-walletServeImpl app =
-    serveImpl app "127.0.0.1"
+walletServeImpl app (ip, addr) =
```
akegalj: should be `(ip, port)` -- naming is a bit misleading

node/src/Pos/Client/CLI/Options.hs
```diff
@@ -188,12 +188,16 @@ webPortOption portNum help =
         <> Opt.value portNum
         <> Opt.showDefault
 
-walletPortOption :: Word16 -> String -> Opt.Parser Word16
-walletPortOption portNum help =
-    Opt.option Opt.auto $
-        templateParser "wallet-port" "PORT" help -- "Port for wallet"
-        <> Opt.value portNum
-        <> Opt.showDefault
+daedalusWalletAddressOption :: Maybe NetworkAddress -> Opt.Parser NetworkAddress
```
Martoon-00: Why "daedalus" prefix if you make this PR not for daedalus use?

akegalj: @Martoon-00 true, right. I will remove the prefix.

wallet/src/Pos/Wallet/Web/Server/Launcher.hs
Martoon-00: Why "addr", it indicates a port, right?

akegalj:
@flyingleafe 
> I was concerned that maybe --wallet-port non-default argument is used in some script already, which could make it a breaking change, but I searched through 3 repositories (cardano-sl, daedalus, iohk-ops) and didn't find any uses. So that should be alright.

yes, luckily we were using only default options for walletPort in our use cases - so there was no need for anything custom

PR #1595, opened 2017-09-19
[DEVOPS-227] enable cardano profiling

cleverca22:
@domenkozar there is now a profiling option, and hydra will pre-build everything with and without profiling

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -5,7 +5,10 @@ in
 , config ? {}
 , dconfig ? "testnet_staging_full"
 , gitrev ? "unknown"
-, pkgs ? (import (localLib.fetchNixPkgs) { inherit system config; }) }:
+, pkgs ? (import (localLib.fetchNixPkgs) { inherit system config; })
+# profiling slows down performance by 50% so we don't enable it by default
+, enableProfiling ? false
+}:
 
 with pkgs.lib;
 with pkgs.haskell.lib;
@@ -21,7 +24,6 @@ let
         configureFlags = [
           "-f-asserts"
           "-f-dev-mode"
-          "--ghc-option=-optl-lm"
         ];
         testTarget = "--log=test.log || (sleep 10 && kill $TAILPID && false)";
         preCheck = ''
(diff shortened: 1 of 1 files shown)


deepfire:
@cleverca22, looks very nice!

default.nix
```diff
@@ -5,7 +5,9 @@ in
 , config ? {}
 , dconfig ? "testnet_staging_full"
 , gitrev ? "unknown"
-, pkgs ? (import (localLib.fetchNixPkgs) { inherit system config; }) }:
+, pkgs ? (import (localLib.fetchNixPkgs) { inherit system config; })
+, enableProfiling ? false
```
domenkozar: I'd add a comment here: `# profiling slows down performance by 50% so we don't enable it by default`

node/cardano-sl.cabal
```diff
@@ -25,6 +25,7 @@ Flag with-explorer
   description: Build with explorer
 
 library
+  extra-libraries:      m
```
domenkozar: Our used nixpkgs commit a905b7cd0c2dc0714195a50bf176cd8e4593502d includes the GHC patch in aafe3d29c1572b89aa0f4f8095ffd9a43c77229e - so I wonder why we need this?

domenkozar: Also once we figure out why we need this, it will have to be enabled only for linux

release.nix
```diff
@@ -27,6 +28,7 @@ let
     cardano-sl-explorer-static = [ "x86_64-linux" ];
     cardano-report-server-static = [ "x86_64-linux" ];
   };
-in (mergeAttrsMap (dconfig: { ${dconfig } = (withDconfig dconfig).mapTestOn platforms; }) dconfigs)
-   // ((withDconfig null).mapTestOn { stack2nix = supportedSystems; })
+in (mergeAttrsMap (dconfig: { ${dconfig } = (withDconfig dconfig false).mapTestOn platforms; }) dconfigs)
+   // ((withDconfig null false).mapTestOn { stack2nix = supportedSystems; })
    // (rlib.mapTestOn { purescript = supportedSystems; })
+   // (mergeAttrsMap (dconfig: { "${dconfig}-profiled" = (withDconfig dconfig true).mapTestOn platforms; }) dconfigs)
```
domenkozar: As discussed this needs to be rebased on top of `cardano-sl-1.0-genesis-json2` branch and then we build profiled build only of `cardano-sl-core` on Linux.

PR #1605, opened 2017-09-21
[DEVOPS-333] allow the nix expressions to work with a blank NIX_PATH

cleverca22:


domenkozar:
@dcoutts we'd like to include this one for 1.0 - it only changes the way we bootstrap the correct pinned nixpkgs for building packages, it causes no behaviour change for the CI.

PR #1606, opened 2017-09-21
[CSE-226] Add feedback about empty slot

sectore:
_Task:_
https://issues.serokell.io/issue/CSE-226

_How to test:_
1. [Build](https://github.com/input-output-hk/cardano-sl/tree/master/explorer/frontend#short-version-of-installation) `Explorer` locally and run it with its [mock server](https://github.com/input-output-hk/cardano-sl/tree/master/explorer#run-mock-server).
2. Open http://localhost:3100/epoch/1/slot/1 in your browser to get the search result page about an empty slot.


sectore:
> Optional "An error occurred,
epoch / slot cannot be found." instead of "
Epoch / slot do not exist." since the error can is more general than just searching for epoch / slot.

@ksaric Valid point! I've requested a Japanese translation for this message and will add it to this PR tomorrow.

PR #1609, opened 2017-09-21
Canonically encode/decode Timestamp as seconds

gromakovsky:


core/Pos/Core/Genesis/Canonical.hs
```diff
@@ -229,7 +230,7 @@ instance Monad m => ToJSON m GenesisData where
         mkObject
             [ ("bootStakeholders", toJSON gdBootStakeholders)
             , ("heavyDelegation", toJSON gdHeavyDelegation)
-            , ("startTime", toJSON gdStartTime)
+            -- , ("startTime", toJSON gdStartTime)
```
gromakovsky: It's accidental, will force push in a moment.

PR #1614, opened 2017-09-21
[CSL-1683] Add configuration options to launcher, propagate to node

gromakovsky:
Now cardano-launcher has options to set configuration path and key.
It also passes their values to node, unless they are explicitly specified
with '-n'.

### DIFF
--- a/tools/src/launcher/Main.hs
+++ b/tools/src/launcher/Main.hs
@@ -12,10 +12,10 @@ import           Universum
 import           Control.Concurrent           (modifyMVar_)
 import           Control.Concurrent.Async     (Async, async, cancel, poll, wait, waitAny,
                                                withAsyncWithUnmask)
-import           Data.Default                 (def)
 import           Data.List                    (isSuffixOf)
 import qualified Data.Text.IO                 as T
 import qualified Data.Text.Lazy.IO            as TL
+import           Data.Time.Units              (Second, convertUnit)
 import           Data.Version                 (showVersion)
 import           Formatting                   (format, int, shown, stext, text, (%))
 import qualified NeatInterpolation            as Q (text)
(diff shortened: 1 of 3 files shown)


tools/src/launcher/Main.hs
```diff
@@ -202,6 +208,26 @@ main = do
...
+        addConfFileOption cfoFilePath .
+        addConfKeyOption cfoKey . addSystemStartOption cfoSystemStart
+
+    addConfFileOption filePath =
+        maybeAddOption "--configuration-file" (toText filePath)
+    addConfKeyOption key = maybeAddOption "--configuration-key" key
+    addSystemStartOption =
+        maybe identity (maybeAddOption "--system-start" . timestampToText)
+
+    maybeAddOption :: Text -> Text -> [Text] -> [Text]
```
neongreen: won't it fail if somebody writes `--configuration-key=blah` instead of `--configuration-key blah`?

domenkozar:
This one is required for daedalus, since `cardano-launcher` previously hardcoded `--configuration-*` args.

tools/src/launcher/Main.hs
```diff
@@ -202,6 +208,26 @@ main = do
                     , loUpdateArchive)
                     loNodeTimeoutSec
                     loReportServer
+  where
+    addConfigurationOptions :: ConfigurationOptions -> [Text] -> [Text]
```
neongreen: This is rather mda because when somebody updates `ConfigurationOptions`, they'll likely forget to update this function. I don't know a way to make GHC warn you when you haven't used all fields from a record, so I think you should do one of the following:

* `addConfigurationOptions opt@(ConfigurationOptions _ _ _) = ...`
* move `addConfigurationOptions` to the place where `ConfigurationOptions` is defined
* add a commend to the place where `ConfigurationOptions` is defined, saying “please update blah blah when you add a field here”

tools/src/launcher/Main.hs
gromakovsky: Well, it will fail, but actually this logic wasn't even planned initially. I wanted to simply propagate all configuration options, but then decided to also check whether they are passed using `-n` argument which is slightly better, but doesn't cover all cases. I think it's ok here.

tools/src/launcher/Main.hs
gromakovsky: Right, I will use the first way, give me a minute.

tools/src/launcher/Main.hs
neongreen: Okay, yes, it seems ok.

neongreen:
nb. I haven't checked whether `pretty @Integer $ fromIntegral $ convertUnit @_ @Second ts` is okay because I've no idea how the hell timestamps are encoded (n? n*1000? n*1000000? something else?). If you're sure that it's correct then ok.

gromakovsky:
It should be `--system-start 100500` where 100500 is amount of seconds.

PR #1618, opened 2017-09-22
Dry run 2 genesis & configuration update

deepfire:
1. supply genesis deployed during the second dry-run
2. update `nodes/configuration.yaml` to point to that genesis (with hash updated)

Note, that because `nodes/configuration.yaml` was updated for mainnet, it's now inconsistent (and was inconsistent before this branch).


domenkozar:
This is needed for hash to actually match the genesis, altough we now have two of them.

PR #1622, opened 2017-09-22
[DEVOPS-352] allow enabling debug symbols for everything

cleverca22:


default.nix
```diff
@@ -7,20 +7,27 @@ in
...
 
 with pkgs.lib;
 with pkgs.haskell.lib;
 
 let
+  # TODO: DEVOPS-355
+  enableDWARFDebugging = drv: appendConfigureFlag (dontStrip drv) "--ghc-options=-g --disable-executable-stripping --disable-library-stripping";
+  dontStrip = drv: overrideCabal drv (drv: { dontStrip = true; });
+  
+  nop = args: args;
```
domenkozar: this is not needed

domenkozar:
@dcoutts this prepares a way to have DWARF debug symbols, but doesn't include them by default. The node size is 20MB without and 170MB with debug symbols. With 40 nodes we'd need to deploy 8GB which is too much.

PR #1626, opened 2017-09-22
[CSM-465] Remove obsolete update info when querying `nextUpdate`

flyingleafe:
After investigation with Ilya it turned out that `postponeUpdate` endpoint introduced in previous PRs is working as intended, the problem is that it doesn't get called by frontend for some reason. But then we came up with a solution which doesn't depend on frontend, which is filtering info about already installed updates by comparison with current software version.

PR has a WIP flag because it should be tested locally before merge (to avoid similar problems as before).


akegalj:
@flyingleafe have you been able to test it ?

PR #1629, opened 2017-09-23
[CSL-1688] Implement scripts to locally test update system

georgeee:
This PR proposes set of scripts to locally test software update:

1. Propose new version
2. Launch wallet nodes via launcher
3. See wallet node (with Daedalus) updated

### DIFF
--- a/Daedalus.sh.example
+++ b/Daedalus.sh.example
@@ -1,6 +1,0 @@
-#!/usr/bin/env sh
-
-SCRIPT=$(readlink -f $0)
-SCRIPT_PATH=$(dirname $SCRIPT)
-
-NODE_TLS_REJECT_UNAUTHORIZED=0 $SCRIPT_PATH/../daedalus/release/linux-x64/Daedalus-linux-x64/Daedalus
--- a/Daedalus.sh.example
+++ b/Daedalus.sh.example
@@ -1,0 +1,6 @@
+#!/usr/bin/env sh
+
+SCRIPT=$(readlink -f $0)
+SCRIPT_PATH=$(dirname $SCRIPT)
+
+NODE_TLS_REJECT_UNAUTHORIZED=0 $SCRIPT_PATH/../daedalus/release/linux-x64/Daedalus-linux-x64/Daedalus
--- a/auxx/src/Plugin.hs
+++ b/auxx/src/Plugin.hs
@@ -77,24 +77,26 @@ rawExec mHasAuxxMode AuxxOptions{..} mSendActions = \case
     Left WithCommandAction{..} -> do
         printAction <- getPrintAction
         printAction "... the auxx plugin is ready"
-        forever $ withCommand $ runCmd mHasAuxxMode mSendActions printAction
-    Right cmd -> runWalletCmd mHasAuxxMode mSendActions cmd
+        forever $ withCommand $ \line -> do
+            expr <- eitherToThrow $ Lang.parse line
+            value <- eitherToThrow =<< Lang.evaluate commandProcs expr
+            withValueText printAction value
+    Right cmd -> worker' runCmdOuts $ runWalletCmd cmd
+  where
+    worker' specs w =
+        worker specs $ \sa -> do
+            logInfo $ sformat ("Length of genesis utxo: " %int)
+                              (length $ unGenesisUtxo genesisUtxo)
+            w (addLogging sa)
 
-runWalletCmd ::
-       ( HasCompileInfo
-       , MonadIO m
-       , Mockable Catch m
-       , MonadThrow m
-       , CanLog m
-       , HasLoggerName m
-       , Mockable Delay m
-       )
-    => Maybe (Dict (MonadAuxxMode m))
-    -> Maybe (SendActions m)
-    -> Text
-    -> m ()
-runWalletCmd mHasAuxxMode mSendActions line = do
-    runCmd mHasAuxxMode mSendActions printAction line
+runWalletCmd :: (HasConfigurations, HasCompileInfo) => Text -> Worker AuxxMode
+runWalletCmd line sendActions = do
+    printAction "Running command: "
+    printAction $ "\t" <> line
+    let commandProcs = createCommandProcs printAction sendActions
+    expr <- eitherToThrow $ Lang.parse line
+    value <- eitherToThrow =<< Lang.evaluate commandProcs expr
+    withValueText printAction value
     printAction "Command execution finished"
     printAction " " -- for exit by SIGPIPE
     liftIO $ hFlush stdout
--- a/auxx/src/Plugin.hs
+++ b/auxx/src/Plugin.hs
@@ -14,7 +14,7 @@ import           Universum
 import           System.Exit (ExitCode (ExitSuccess))
 import           System.Posix.Process (exitImmediately)
 #endif
-import           Data.Constraint (Dict(..))
+import           Data.Constraint (Dict (..))
 import           Formatting (float, int, sformat, stext, (%))
 import           Mockable (Catch, Delay, Mockable, delay)
 import           Node.Conversation (ConversationActions (..))
@@ -23,9 +23,8 @@ import           Serokell.Util (sec)
 import           System.IO (hFlush, stdout)
 import           System.Wlog (CanLog, HasLoggerName, WithLogger, logDebug, logInfo)
 
-import           Pos.Communication (Conversation (..), OutSpecs (..), SendActions (..),
-                                    WorkerSpec, delegationRelays, relayPropagateOut, txRelays,
-                                    usRelays, worker)
+import           Pos.Communication (Conversation (..), OutSpecs (..), SendActions (..), WorkerSpec,
+                                    delegationRelays, relayPropagateOut, txRelays, usRelays, worker)
 import           Pos.Crypto (AHash (..), fullPublicKeyF, hashHexF)
 import           Pos.Launcher.Configuration (HasConfigurations)
 import           Pos.Txp (genesisUtxo, unGenesisUtxo)
@@ -38,7 +37,7 @@ import           AuxxOptions (AuxxOptions (..))
 import           Command (createCommandProcs)
 import qualified Lang
 import           Mode (MonadAuxxMode)
-import           Repl (WithCommandAction (..), PrintAction)
+import           Repl (PrintAction, WithCommandAction (..))
 
 ----------------------------------------------------------------------------
 -- Plugin implementation
@@ -77,26 +76,24 @@ rawExec mHasAuxxMode AuxxOptions{..} mSendActions = \case
     Left WithCommandAction{..} -> do
         printAction <- getPrintAction
         printAction "... the auxx plugin is ready"
-        forever $ withCommand $ \line -> do
-            expr <- eitherToThrow $ Lang.parse line
-            value <- eitherToThrow =<< Lang.evaluate commandProcs expr
-            withValueText printAction value
-    Right cmd -> worker' runCmdOuts $ runWalletCmd cmd
-  where
-    worker' specs w =
-        worker specs $ \sa -> do
-            logInfo $ sformat ("Length of genesis utxo: " %int)
-                              (length $ unGenesisUtxo genesisUtxo)
-            w (addLogging sa)
+        forever $ withCommand $ runCmd mHasAuxxMode mSendActions printAction
+    Right cmd -> runWalletCmd mHasAuxxMode mSendActions cmd
 
-runWalletCmd :: (HasConfigurations, HasCompileInfo) => Text -> Worker AuxxMode
-runWalletCmd line sendActions = do
-    printAction "Running command: "
-    printAction $ "\t" <> line
-    let commandProcs = createCommandProcs printAction sendActions
-    expr <- eitherToThrow $ Lang.parse line
-    value <- eitherToThrow =<< Lang.evaluate commandProcs expr
-    withValueText printAction value
+runWalletCmd ::
+       ( HasCompileInfo
+       , MonadIO m
+       , Mockable Catch m
+       , MonadThrow m
+       , CanLog m
+       , HasLoggerName m
+       , Mockable Delay m
+       )
+    => Maybe (Dict (MonadAuxxMode m))
+    -> Maybe (SendActions m)
+    -> Text
+    -> m ()
+runWalletCmd mHasAuxxMode mSendActions line = do
+    runCmd mHasAuxxMode mSendActions printAction line
     printAction "Command execution finished"
     printAction " " -- for exit by SIGPIPE
     liftIO $ hFlush stdout
--- a/docs/how-to/test-update-system-locally.md
+++ b/docs/how-to/test-update-system-locally.md
@@ -1,0 +1,29 @@
+# Testing update system locally
+
+## Prerequisites
+
+Additional utilities to be installed:
+
+* webfs (webfsd) server
+* uxterm to emulate UI when no `$UI` provided
+
+## With Daedalus
+
+```
+UI=Daedalus.sh ./scripts/launch/us-test.sh
+```
+
+This will launch nodes, Daedalus
+
+Print `start` in original console to propose an update.
+
+## Without Daedalus
+
+```
+./scripts/launch/us-test.sh
+```
+
+This will launch nodes, ui emulation as simple `uxterm` terminal with incitation to write `update` to emulate UI closing with exit code 20.
+
+Print `start` in original console to propose an update.
+
--- a/docs/how-to/test-update-system-locally.md
+++ b/docs/how-to/test-update-system-locally.md
@@ -1,5 +1,13 @@
 # Testing update system locally
 
+This instruction covers how to test software update locally:
+
+  1. Prepare and propose software update
+      a. Prepare alt configuration with changed application version constant
+      b. Propose updated with script that simply replaces one configuration file (in setup) with other
+  2. Launch wallet node with launcher, Daedalus (or Daedalus emulator)
+      a. When update arrives Deadalus to finish, launcher to restart Daedalus + node
+
 ## Prerequisites
 
 Additional utilities to be installed:
@@ -8,6 +16,8 @@ Additional utilities to be installed:
 * uxterm to emulate UI when no `$UI` provided
 
 ## With Daedalus
+
+Write command to launch Daedalus on your system to `Daedalus.sh`
 
 ```
 UI=Daedalus.sh ./scripts/launch/us-test.sh
@@ -23,7 +33,7 @@ Print `start` in original console to propose an update.
 ./scripts/launch/us-test.sh
 ```
 
-This will launch nodes, ui emulation as simple `uxterm` terminal with incitation to write `update` to emulate UI closing with exit code 20.
+This will launch nodes, ui emulation as simple `uxterm` terminal with invitation to write `update` to emulate UI closing with exit code 20.
 
 Print `start` in original console to propose an update.
 
--- a/docs/how-to/test-update-system-locally.md
+++ b/docs/how-to/test-update-system-locally.md
@@ -5,8 +5,8 @@ This instruction covers how to test software update locally:
   1. Prepare and propose software update
       a. Prepare alt configuration with changed application version constant
       b. Propose updated with script that simply replaces one configuration file (in setup) with other
-  2. Launch wallet node with launcher, Daedalus (or Daedalus emulator)
-      a. When update arrives Deadalus to finish, launcher to restart Daedalus + node
+  2. Launch wallet node with launcher, Daedalus (or script that serves as Daedalus emulator)
+      a. When update arrives to Deadalus to finish, launcher will restart Daedalus + node
 
 ## Prerequisites
 
@@ -17,7 +17,7 @@ Additional utilities to be installed:
 
 ## With Daedalus
 
-Write command to launch Daedalus on your system to `Daedalus.sh`
+Write command to launch Daedalus on your system to `daedalus-example.sh`
 
 ```
 UI=Daedalus.sh ./scripts/launch/us-test.sh
--- a/scripts/launch/Test.hs
+++ b/scripts/launch/Test.hs
@@ -15,6 +15,7 @@ import           Control.Monad.Except      (ExceptT (..), runExceptT)
 import           Control.Monad.Trans       (lift)
 import           Data.Aeson.Lens
 import qualified Data.List                 as L
+import qualified Data.List.NonEmpty        as NE
 import qualified Data.Map                  as M
 import           Data.Maybe                (catMaybes, fromJust, fromMaybe)
 import           Data.Monoid               ((<>))
@@ -139,32 +140,67 @@ updateTest config (Config{..}) = do
             genBlockSize <- config ^? genBlockSizeL
             sv <- config ^? svL
             (cNodes :: Int) <- round <$> config ^? richL
-            return $ (,cNodes) $ "propose-update 0 "
-              <> showF bvMajor <> "." <> showF bvMinor <> "." <> showF bvAlt
-              <> " " <> showF genScriptVersion <> " " <> showF (genSlot / 1000)
-              <> " " <> showF genBlockSize
-              <> " " <> T.unpack walletAppName <> ":" <> showF (sv + 1)
-              <> " " <> T.unpack tag <> " test_updater.sh none"
+            let bvmStr = "bvm"
+                    <> " script-version: " <> showF genScriptVersion
+                    <> " slot-duration: " <> showF (genSlot / 1000)
+                    <> " max-block-size: " <> showF genBlockSize
+                updStr = "upd-bin"
+                    <> " system: " <> show (T.unpack tag)
+                    <> " installer-path: ./test_updater.sh"
+            return $ (,cNodes) $ "propose-update i:0"
+              <> " block-version: " <> showF bvMajor <> "." <> showF bvMinor <> "." <> showF bvAlt
+              <> " software-version: ~software~" <> T.unpack walletAppName <> ":" <> showF (sv + 1)
+              <> " bvm: (" <> bvmStr <> ")"
+              <> " update: (" <> updStr <> ")"
+              <> " vote-all: false"
     Prelude.putStrLn proposeUpdStr
     Y.encodeFile "run/conf-fresh.yaml" (M.singleton cKey config'')
 
-    TIO.writeFile "test_updater.sh" $ "cp run/conf-fresh.yaml run/cofiguration.wallet.yaml"
-    let cmdPropose = L.intercalate "," $
-                  (map (\i -> "add-key run/gen-keys/keys-testnet/rich/key"<>show i<>".sk") [1..cNodes])
-                  ++ [ proposeUpdStr ]
+    TIO.writeFile "test_updater.sh" $ "cp run/conf-fresh.yaml run/configuration.wallet.yaml"
+    let cmdPropose = L.intercalate ";" $
+                  (map (\i -> "add-key primary: true file: ./run/gen-keys/generated-keys/rich/key"<>show i<>".sk") [0..cNodes-1])
+                  ++ [ "listaddr", proposeUpdStr ]
         keygenBase = "cardano-keygen --configuration-key "<>cKey<>" --configuration-file "<>cFile'<>" --system-start 0"
-        auxx cmds = "cardano-auxx --keyfile run/auxx.keys --configuration-key "
-            <>cKey<>" --configuration-file "<>cFile'<>" --system-start 0"
-            <> " --peer 127.0.0.1:3001 --db-path run/auxx-db cmd --commands \""<>T.pack cmds<>"\""
+        auxx cmds = ("cardano-auxx",)
+            [ "--keyfile", "run/auxx.keys"
+            , "--configuration-key", cKey
+            , "--configuration-file", cFile'
+            , "--system-start", "0"
+            , "--peer", "127.0.0.1:3001"
+            , "--db-path", "run/auxx-db"
+            , "--log-config", "scripts/log-templates/log-config-greppable.yaml"
+            , "cmd", "--commands", T.pack cmds
+            ]
     shells (keygenBase <> " generate-keys-by-spec --genesis-out-dir run/gen-keys") empty
     shells (keygenBase <> " rearrange --mask 'run/gen-keys/keys-testnet/rich/key*.sk'") empty
-    (exitCode, T.lines -> auxxOut) <- shellStrict (auxx cmdPropose) empty
-    let fileHash = L.last $ T.words $ head $ filter ("Read file test_updater.sh succesfuly, its hash" `T.isPrefixOf`) auxxOut
-    let updHash = L.last $ T.words $ head $ filter ("Update proposal submitted, upId" `T.isPrefixOf`) auxxOut
+    (exitCode, textToLines -> auxxOut) <- uncurry procStrict (auxx cmdPropose) empty
+    case exitCode of
+        ExitSuccess -> do
+            echo "Proposing an update succeeded, here are the logs:"
+            mapM_ (echo . mappend "\t") auxxOut
+        ExitFailure _ -> do
+            echo "Proposing an update failed, see auxx output:"
+            mapM_ (echo . mappend "\t") auxxOut
+            exit exitCode
+    let matchInAuxxOut pat = do
+            (lineToText -> t) <- NE.toList auxxOut
+            match (suffix pat) t
+        extractFromAuxxOut s pat =
+            case matchInAuxxOut pat of
+                m:_ -> return m
+                _ -> do
+                    echo $ "Could not extract " <> s <> " from auxxOut"
+                    exit $ ExitFailure 1
+    fileHash <- extractFromAuxxOut "fileHash" $
+        "Read file ./test_updater.sh succesfuly, its hash: " *>
+        (T.pack <$> some hexDigit)
+    updHash <- extractFromAuxxOut "updHash" $
+        "Update proposal submitted, upId: " *>
+        (T.pack <$> some hexDigit)
     TIO.putStrLn $ fileHash <> " " <> updHash
-    let cmdVote = L.intercalate "," $
-                  (map (\i -> "vote "<>show i<>" y "<>T.unpack updHash) [1..cNodes-1])
-    shells (auxx cmdVote) empty
+    let cmdVote = L.intercalate ";" $
+                  (map (\i -> "vote "<>show i<>" agree: true up-id: "<>T.unpack updHash) [1..cNodes-1])
+    uncurry procs (auxx cmdVote) empty
 
     shells ("mkdir run/serve-upd; mv test_updater.sh run/serve-upd/"<>fileHash) empty
 
--- a/scripts/launch/Test.hs
+++ b/scripts/launch/Test.hs
@@ -128,7 +128,7 @@ updateTest :: Y.Value -> Config -> IO ()
 updateTest config (Config{..}) = do
     let config' = config & appNameL .~ walletAppName
         config'' = config' & svL %~ (+1)
-        showF = (show :: Int -> String) . round
+        showF = T.pack . show @Int . round
         cFile' = T.pack $ F.encodeString cFile
         (proposeUpdStr, cNodes) = fromMaybe (error "Failed to form update proposal cmd") $ do
             tag <- config ^? tagL
@@ -145,22 +145,23 @@ updateTest config (Config{..}) = do
                     <> " slot-duration: " <> showF (genSlot / 1000)
                     <> " max-block-size: " <> showF genBlockSize
                 updStr = "upd-bin"
-                    <> " system: " <> show (T.unpack tag)
+                    <> " system: " <> T.pack (show tag)
                     <> " installer-path: ./test_updater.sh"
             return $ (,cNodes) $ "propose-update i:0"
               <> " block-version: " <> showF bvMajor <> "." <> showF bvMinor <> "." <> showF bvAlt
-              <> " software-version: ~software~" <> T.unpack walletAppName <> ":" <> showF (sv + 1)
+              <> " software-version: ~software~" <> walletAppName <> ":" <> showF (sv + 1)
               <> " bvm: (" <> bvmStr <> ")"
               <> " update: (" <> updStr <> ")"
-              <> " vote-all: false"
-    Prelude.putStrLn proposeUpdStr
+              <> " vote-all: true"
+    TIO.putStrLn proposeUpdStr
     Y.encodeFile "run/conf-fresh.yaml" (M.singleton cKey config'')
 
     TIO.writeFile "test_updater.sh" $ "cp run/conf-fresh.yaml run/configuration.wallet.yaml"
-    let cmdPropose = L.intercalate ";" $
-                  (map (\i -> "add-key primary: true file: ./run/gen-keys/generated-keys/rich/key"<>show i<>".sk") [0..cNodes-1])
-                  ++ [ "listaddr", proposeUpdStr ]
-        keygenBase = "cardano-keygen --configuration-key "<>cKey<>" --configuration-file "<>cFile'<>" --system-start 0"
+    let cmdPropose = T.intercalate ";"
+            [ "add-key-pool " <> T.unwords (format ("i:"%d) <$> [0..cNodes-1])
+            , "listaddr"
+            , proposeUpdStr
+            ]
         auxx cmds = ("cardano-auxx",)
             [ "--keyfile", "run/auxx.keys"
             , "--configuration-key", cKey
@@ -169,10 +170,8 @@ updateTest config (Config{..}) = do
             , "--peer", "127.0.0.1:3001"
             , "--db-path", "run/auxx-db"
             , "--log-config", "scripts/log-templates/log-config-greppable.yaml"
-            , "cmd", "--commands", T.pack cmds
+            , "cmd", "--commands", cmds
             ]
-    shells (keygenBase <> " generate-keys-by-spec --genesis-out-dir run/gen-keys") empty
-    shells (keygenBase <> " rearrange --mask 'run/gen-keys/keys-testnet/rich/key*.sk'") empty
     (exitCode, textToLines -> auxxOut) <- uncurry procStrict (auxx cmdPropose) empty
     case exitCode of
         ExitSuccess -> do
@@ -198,9 +197,6 @@ updateTest config (Config{..}) = do
         "Update proposal submitted, upId: " *>
         (T.pack <$> some hexDigit)
     TIO.putStrLn $ fileHash <> " " <> updHash
-    let cmdVote = L.intercalate ";" $
-                  (map (\i -> "vote "<>show i<>" agree: true up-id: "<>T.unpack updHash) [1..cNodes-1])
-    uncurry procs (auxx cmdVote) empty
 
     shells ("mkdir run/serve-upd; mv test_updater.sh run/serve-upd/"<>fileHash) empty
 
--- a/scripts/launch/Test.hs
+++ b/scripts/launch/Test.hs
@@ -1,5 +1,5 @@
 #!/usr/bin/env stack
--- stack runghc --package vector --package cassava --package turtle --package safe --package aeson --package yaml --package lens-aeson
+-- stack script --resolver lts-9.12
 
 {-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE OverloadedStrings   #-}
@@ -8,12 +8,13 @@
 {-# LANGUAGE TupleSections       #-}
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE ViewPatterns        #-}
+{-# LANGUAGE PackageImports      #-}
 
 import           Control.Lens
 import           Control.Monad             (forM_, void, when)
 import           Control.Monad.Except      (ExceptT (..), runExceptT)
 import           Control.Monad.Trans       (lift)
-import           Data.Aeson.Lens
+import "lens-aeson" Data.Aeson.Lens
 import qualified Data.List                 as L
 import qualified Data.List.NonEmpty        as NE
 import qualified Data.Map                  as M
@@ -194,7 +195,7 @@ updateTest config (Config{..}) = do
         "Read file ./test_updater.sh succesfuly, its hash: " *>
         (T.pack <$> some hexDigit)
     updHash <- extractFromAuxxOut "updHash" $
-        "Update proposal submitted, upId: " *>
+        "Update proposal submitted along with votes, upId: " *>
         (T.pack <$> some hexDigit)
     TIO.putStrLn $ fileHash <> " " <> updHash
 
--- a/scripts/launch/Test.hs
+++ b/scripts/launch/Test.hs
@@ -1,0 +1,181 @@
+#!/usr/bin/env stack
+-- stack runghc --package vector --package cassava --package turtle --package safe --package aeson --package yaml --package lens-aeson
+
+{-# LANGUAGE DeriveGeneric       #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections       #-}
+{-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE ViewPatterns        #-}
+
+import           Control.Lens
+import           Control.Monad             (forM_, void, when)
+import           Control.Monad.Except      (ExceptT (..), runExceptT)
+import           Control.Monad.Trans       (lift)
+import           Data.Aeson.Lens
+import qualified Data.List                 as L
+import qualified Data.Map                  as M
+import           Data.Maybe                (catMaybes, fromJust, fromMaybe)
+import           Data.Monoid               ((<>))
+import           Data.Scientific           (Scientific)
+import qualified Data.Text                 as T
+import           Data.Text.IO              as TIO
+import           Data.Text.Lazy            (fromStrict)
+import           Data.Text.Lazy.Encoding   (encodeUtf8)
+import qualified Data.Yaml                 as Y
+import           Filesystem.Path.CurrentOS (encodeString)
+import qualified Filesystem.Path.CurrentOS as F
+import           GHC.Generics
+import           Prelude                   hiding (FilePath)
+import           Text.Printf
+import           Turtle                    hiding (printf)
+
+
+-- CLI Parser
+
+data Command =
+    Update
+  | GenWalletConf !FilePath
+  deriving (Show)
+
+cmdParser :: Parser Command
+cmdParser =
+      subcommand "update" "Test software update" (pure Update)
+  <|> subcommand "gen-wallet-conf" "generate wallet configuration"
+        (GenWalletConf <$> optPath "out" 'o' "Output file")
+
+data Config = Config
+      { cFile :: FilePath
+      , cKey  :: Text
+      , cCmd  :: Command
+      }
+
+parser :: Parser Config
+parser = Config
+    <$> optPath "configuration-file" 'c' "Configuration file"
+    <*> optText "configuration-key" 'k' "Configuration key"
+    <*> cmdParser
+
+walletAppName :: Text
+walletAppName = "csl-daedalus"
+
+main :: IO ()
+main = do
+  c@Config {..} <- options "Helper CLI around NixOps to run experiments" parser
+  config <- fromMaybe (error "No config key found") . M.lookup cKey . either (error "Wrong yaml config") id
+              <$> Y.decodeFileEither (F.encodeString cFile)
+  case cCmd of
+    Update -> updateTest config c
+    GenWalletConf outFile ->
+        let config' = config & appNameL .~ walletAppName
+         in Y.encodeFile (F.encodeString outFile) (M.singleton cKey config')
+
+toObject :: Y.Value -> Maybe Y.Object
+toObject (Y.Object a) = Just a
+toObject _            = Nothing
+
+genSpecL :: AsValue t => Traversal' t Y.Value
+genSpecL = key "core"
+         . key "genesis"
+         . key "spec"
+
+genesisConstL = genSpecL
+         . key "blockVersionData"
+
+genSlotL = genesisConstL
+         . key "slotDuration"
+         . _Number
+
+genBlockSizeL = genesisConstL
+         . key "maxBlockSize"
+         . _Number
+
+genScriptVersionL = genesisConstL
+         . key "scriptVersion"
+         . _Number
+
+richL :: AsValue t => Traversal' t Scientific
+richL = genSpecL
+         . key "initializer"
+         . key "testnetInitializer"
+         . key "testBalance"
+         . key "richmen"
+         . _Number
+
+svL :: AsValue t => Traversal' t Scientific
+svL = key "update"
+         . key "applicationVersion"
+         . _Number
+
+tagL = key "update"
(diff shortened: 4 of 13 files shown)


georgeee:
> The updater has exited with ExitSuccess

This is indeed expected. This is because wallet node is started with `launcher`

scripts/launch/Test.hs
```diff
@@ -0,0 +1,217 @@
...
+            , "--configuration-key", cKey
+            , "--configuration-file", cFile'
+            , "--system-start", "0"
+            , "--peer", "127.0.0.1:3001"
+            , "--db-path", "run/auxx-db"
+            , "--log-config", "scripts/log-templates/log-config-greppable.yaml"
+            , "cmd", "--commands", T.pack cmds
+            ]
+    shells (keygenBase <> " generate-keys-by-spec --genesis-out-dir run/gen-keys") empty
+    shells (keygenBase <> " rearrange --mask 'run/gen-keys/keys-testnet/rich/key*.sk'") empty
```
gromakovsky: Rearrange is a useless hack. Why is it here?

gromakovsky:
> This is indeed expected. This is because wallet node is started with launcher

The thing is: if I run `scripts/launch/demo-with-wallet-api.sh`, there are no logs from wallet node in console. It's a breaking change (I am quite sure previously there were logs) which I would like to avoid.

Daedalus.sh.example
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
```
gromakovsky: I don't like having this file in top-level directory. Can it be moved somewhere?

tools/src/launcher/Main.hs
```diff
@@ -103,6 +105,10 @@ optionsParser = do
         short   'w' <>
         help    "An argument to be passed to the wallet." <>
         metavar "ARG"
+    loWalletLogging <- optional $ textOption $
```
int-index: use a `switch`

v0d1ch:
I implemented the logging part and I believe it works fine since I see `pid` which was not the case before.
```
Running in the client scenario
Running the updater
The updater has exited with ExitFailure 127
Starting the node
Node started
Starting the wallet
My pid 20616, PPID=20612
xterm: Xt error: Can't open display:
xterm: DISPLAY is not set
Update started
The wallet has exited with code 20
Killing the node in 5 seconds
Killing the node now
Running the updater
The updater has exited with ExitFailure 127 ...
```
 Now, next problem is, when I pass in the wallet path to `us-test.sh` manually instead of `ui-simulator.sh` I get `MissingSystemStartTime` exception. So to be clear I pass in `wallet/.stack-work/dist/x86_64-osx/Cabal-1.24.2.0/build/cardano-node/cardano-node` as the path and then the exception occurs. I could use some help with this since I am not familiar with the system yet.

Daedalus.sh.example
gromakovsky: I still don't like having this file in top-level directory. Can it be moved somewhere?

scripts/launch/demo.sh
```diff
@@ -85,6 +85,14 @@ echo "Using system start time "$system_start
 
 echo "Number of panes: $panesCnt"
 
+function prefix_args {
```
volhovm: What does this function do? I mean, I can read the code and understand each step, but... how would one even describe it? :thinking:

tools/src/launcher/Main.hs
```diff
@@ -133,6 +136,9 @@ optionsParser = do
         short   'w' <>
         help    "An argument to be passed to the wallet frontend executable." <>
         metavar "ARG"
+    loWalletLogging <- switch $
+        long    "wlogging" <>
+        help    "Logging flag for the launcher wallet."
```
volhovm: Logging flag what? :)
Does it turn on the logging? From Info to Debug? From Warning to Info? More descriptive `help` (`long` too) would be good.

tools/src/launcher/Main.hs
```diff
@@ -515,11 +523,18 @@ spawnNode (path, args, mbLogPath) = do
...
-    logNotice "Starting the wallet"
-    view _1 <$>
-        liftIO (readProcessWithExitCode path (map toString args) mempty)
+runWallet :: Bool -> (FilePath, [Text]) -> M ExitCode
+runWallet shouldLog (path, args) = do
+    putText "Starting the wallet"
+    if shouldLog then
+        liftIO $ do
+        (_, stdO, stdE, pid) <- runInteractiveProcess path (map toString args) Nothing Nothing
+        withAsync (forever $ IO.hGetLine stdO >>= IO.hPutStrLn stdout . ("[wallet] " <>)) $ \_ ->
```
volhovm: I know this was a problematic part, but:
1. Looks hacky. Have you considered using `turtle`?
2. Don't you want to at least (maybe) `cancel` these two async threads?

int-index: withAsync does the canceling

v0d1ch: Can someone just give this a go ? On my machine I get this error
```
+ rm -v './run//topology*.yaml'
rm: ./run//topology*.yaml: No such file or directory
```

v0d1ch: Also I think this is related to https://github.com/input-output-hk/cardano-sl/issues/1869

v0d1ch:
Note that in order to be able to actually test this demo.sh script needs to change a bit
```
set +eo xtrace
```
I also had to add `db-path` for wallet launcher
`launcher_args=" $launcher_args --node-timeout 5 $node_args_ --system-start 0 --wlogging  --db-path $base"`
After these changes you should be able to run either `./scripts/launch/us-test.sh` or `./scripts/launch/demo-with-wallet.sh` and get something like [this](https://imgur.com/a/2LT40) .
@georgeee Should I rebase ? I understand there is no strict policy on rebasing vs merging right ? There is a lot of code that I did not work on so I have no problem with rebasing but just a little afraid that i will screw up something else.

Daedalus.sh.example
gromakovsky: I still don't like having this file in top-level directory. Can it be moved somewhere?

@v0d1ch ?

Move it to `scripts/launch/daedalus-example.sh`, for example, and write a comment explaining what this file does.
Also make sure you won't break it (notice that it uses relative paths to the script directory, afaiu).

gromakovsky:
> I also had to add db-path for wallet launcher
launcher_args=" $launcher_args --node-timeout 5 $node_args_ --system-start 0 --wlogging --db-path $base"

1. Can you make this change in the PR as well? It's needed because the command line interface of `cardano-launcher` has been changed.
2. Why do you pass `$base`? You should pass the same path as the one passed to `cardano-node`. You can grep `db-path` in `common-functions.sh`.

tools/src/launcher/Main.hs
```diff
@@ -517,11 +532,47 @@ spawnNode (path, args, mbLogPath) = do
...
+                         , Process.std_out = Process.UseHandle logHandle
+                         , Process.std_err = Process.UseHandle logHandle
+                         }
+            system' phvar cr mempty
+        Nothing ->
+           if shouldLog then
+               liftIO $ do
+               let cr = (Process.proc (toString path) (map toString args))
+                         { Process.std_in  = Process.CreatePipe
+                         , Process.std_out = Process.NoStream
```
pva701: It seems you should have used `Inherit` instead of `NoStream` because it's case when wallet's logs must go into launcher's stdout/stderr

tools/src/launcher/Main.hs
```diff
@@ -488,13 +505,11 @@ spawnNode (path, args, mbLogPath) = do
...
             (lp,) <$> openFile lp AppendMode
         Nothing -> do
-            tempdir <- fromString <$> getTemporaryDirectory
-            -- FIXME (adinapoli): `Shell` from `turtle` was giving us no-resource-leak guarantees
-            -- via the `Managed` monad, which is something we have lost here, and we are back to manual
-            -- resource control. In this case, however, shall we really want to nuke the file? It seems
-            -- something useful to have lying around in the filesystem. We should probably close the
-            -- `Handle`, though, but if this program is short lived it won't matter anyway.
-            IO.openTempFile tempdir "cardano-node-output.log"
+            (_, stdO, stdE, pid) <- runInteractiveProcess path (map toString args) Nothing Nothing
```
pva701: dunno why you didn't use similar approach https://github.com/input-output-hk/cardano-sl/pull/1629/files#diff-6a553d1d73e7c27fa55aff67ff05a784R544.

it seems to be a duplication of code https://github.com/input-output-hk/cardano-sl/pull/1629/files#diff-6a553d1d73e7c27fa55aff67ff05a784R622

tools/src/launcher/Main.hs
```diff
@@ -359,18 +374,20 @@ clientScenario
     :: NodeDbPath
     -> Maybe FilePath                      -- ^ Logger config
     -> (FilePath, [Text], Maybe FilePath)  -- ^ Node, its args, node log
-    -> (FilePath, [Text])                  -- ^ Wallet, args
+    -> (FilePath, [Text], Maybe FilePath)
```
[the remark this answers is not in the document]
v0d1ch: I did it like this since it is defined in the same way for regular node

tools/src/launcher/Main.hs
v0d1ch: Kinda hard to guess what you mean, the links point to the same thing so I am little confused. Do you mean that `runInteractiveProcess` should of should NOT be used ?
Anyway I abstracted out the `ProcessCreate` generation so I think it looks better now ?

tools/src/launcher/Main.hs
v0d1ch: Done

v0d1ch:
@gromakovsky any idea why travis build fails ?
`'dist/test/test.log' has become inaccessible: No such file or directory`

gromakovsky:
I am working on it.

tools/src/launcher/Main.hs
pva701: The code was a little bit changed, but why don't you use `createProcInherit`?
we have other levels of logging apart from `Info` which can be in stdout, but you write all logs from stdout as `Info`. I expect inheritance will save logging level

tools/src/launcher/Main.hs
```diff
@@ -517,11 +524,63 @@ spawnNode (path, args, mbLogPath) = do
...
+        Nothing ->
+           if shouldLog then
+               liftIO $ do
+               let cr = createProcInherit path args
+               system' phvar cr mempty
+           else
+              liftIO $ do
+              (_, _, _, pid) <- runInteractiveProcess path (map toString args) Nothing Nothing
+              waitForProcess pid
+
```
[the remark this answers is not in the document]
v0d1ch: Well those three functions yield `CreateProcess` which later `system'` spawns with 
`(m, stdO, stdE, ph) <- Process.createProcess p`  - this function also takes care of terminating the process and 
prints from handle. I don't know if I can replace it since It would mean I would have to change some things and my task was just to add logging so I assumed all works as expected except logging.

tools/src/launcher/Main.hs
v0d1ch: I used inherited process handle and indeed I see more logging similar to other nodes

tools/src/launcher/Main.hs
```diff
@@ -569,6 +572,13 @@ createProcPipe path args =
         , Process.std_err = Process.CreatePipe
         }
 
+asyncLog :: Handle -> Handle -> ProcessHandle -> NodeType -> IO ExitCode
+asyncLog stdO stdE pH nt = do
+    withAsync (forever $ IO.hGetLine stdO >>= customLogger LInfo nt) $ \_ ->
+        withAsync (forever $ IO.hGetLine stdE >>= customLogger LError nt) $ \_ -> do
+        waitForProcess pH
+
+
```
v0d1ch: @pva701 I also refactored `withAsync` calls to separate function

auxx/src/Plugin.hs
```diff
@@ -24,8 +24,8 @@ import           System.IO (hFlush, stdout)
 import           System.Wlog (CanLog, HasLoggerName, WithLogger, logDebug, logInfo)
 
 import           Pos.Communication (Conversation (..), OutSpecs (..), SendActions (..),
-                                    WorkerSpec, delegationRelays, relayPropagateOut, txRelays,
-                                    usRelays, worker)
+                                   WorkerSpec, delegationRelays, relayPropagateOut,
+                                   txRelays, usRelays, worker)
```
gromakovsky: It's a strange change, maybe you misconfigured `stylish-haskell` or something 🤷‍♂️

scripts/launch/demo.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -eo xtrace
+set +eo xtrace
```
gromakovsky: What does it mean?

scripts/launch/demo.sh
gromakovsky: > understand each step

You are smart, I have no idea what's going on here even if I read the code :/

tools/src/launcher/Main.hs
```diff
@@ -133,7 +139,15 @@ optionsParser = do
...
         help    "An argument to be passed to the wallet frontend executable." <>
         metavar "ARG"
+    loWalletLogging <- switch $
+        long    "wlogging" <>
+        help    "Bool that determines if wallet should log to stdout"
 
+    loWalletLogPath <- optional $ textOption $
+        long    "wallet-log-path" <>
+        help    ("File where wallet stdout/err will be redirected " <>
+                 "(def: temp file).") <>
```
gromakovsky: Why temp file? We discarded the idea of using temp files for logging. Also you should update the description of `--node-log-path` option.

tools/src/launcher/Main.hs
```diff
@@ -488,24 +501,21 @@ spawnNode (path, args, mbLogPath) = do
...
-            -- `Handle`, though, but if this program is short lived it won't matter anyway.
-            IO.openTempFile tempdir "cardano-node-output.log"
+            let cr = createProcInherit path args
+            (_, stdO, stdE, pid) <- Process.createProcess cr
+            case stdO of
+                Just o -> do
+                    _ <- asyncLog o (fromJust stdE) pid NNode
+                    return (path, o)
+                Nothing -> do
+                    return (path, stdin)
```
gromakovsky: Sorry, but I don't understand anything now. You create a process here and then you create it again below. It doesn't make any sense to me.

tools/src/launcher/Main.hs
```diff
@@ -517,11 +527,70 @@ spawnNode (path, args, mbLogPath) = do
...
+createProcHandle path args pHandle =
+    (Process.proc (toString path) (map toString args))
+        { Process.std_in = Process.CreatePipe
+        , Process.std_out = Process.UseHandle pHandle
+        , Process.std_err = Process.UseHandle pHandle
+        }
+
+createProcPipe :: FilePath -> [Text] -> Process.CreateProcess
+createProcPipe path args =
+    (Process.proc (toString path) (map toString args))
```
gromakovsky: `FilePath` is `String`, so applying `toString` to `path` is redundant.

tools/src/launcher/Main.hs
```diff
@@ -517,11 +527,70 @@ spawnNode (path, args, mbLogPath) = do
...
+            let cr = createProcHandle path args logHandle
+            system' phvar cr mempty
+        Nothing ->
+           if shouldLog then
+               liftIO $ do
+               let cr = createProcInherit path args
+               system' phvar cr mempty
+           else
+              liftIO $ do
+              (_, _, _, pid) <- runInteractiveProcess path (map toString args) Nothing Nothing
```
gromakovsky: Using `runInteractiveProcess` here looks really strange. It will create a pipe and the out of the spawned process will be sent to this pipe. Another end of this pipe is returned by this function (actually three handles for three standard streams) and discarded.
It makes more sense to me to use `CreateProcess` with `std_out` and `std_err` set to `NoStream`.

tools/src/launcher/Main.hs
```diff
@@ -517,11 +527,70 @@ spawnNode (path, args, mbLogPath) = do
...
+
+createProcPipe :: FilePath -> [Text] -> Process.CreateProcess
+createProcPipe path args =
+    (Process.proc (toString path) (map toString args))
+        { Process.std_in = Process.CreatePipe
+        , Process.std_out = Process.CreatePipe
+        , Process.std_err = Process.CreatePipe
+        }
+
+asyncLog :: Handle -> Handle -> ProcessHandle -> NodeType -> IO ExitCode
```
gromakovsky: Are we really sure we need this function? It reads output of the spawned process and sends it to our stdout/stderr.
The only difference between this approach and inheriting stdout/stderr is that we can modify those messages. However, some messages take several lines. For example, block header takes several lines. With this approach it will look like:
```
[node:ERROR:ThreadId 4] [2017-12-07 16:55:41.20 UTC] MainBlockHeader:
[node:ERROR:ThreadId 4] [2017-12-07 16:55:41.20 UTC]     hash: 1f19aa3d840b964c55b1ac5cc3e4e19c8d3fa5b066fecedcdb2d811bb0f8d2b3
[node:ERROR:ThreadId 4] [2017-12-07 16:55:42.02 UTC]     previous block: f9f06e9ca71b5fc5c26e7c893e6d6f0011f35a262aa1f2959f05b2e3d0a808a6
…
```

Also it involves extra overhead and extra code.
I personally vote for removing it.

scripts/launch/demo.sh
```diff
@@ -109,24 +117,39 @@ while [[ $i -lt $panesCnt ]]; do
...
+        updater_file="$config_dir/updater$i.sh"
+        launcher_=$(find_binary cardano-launcher)
+        launcher_args="--node $node_ "
+        if [[ "$UI" != "" ]]; then
+          ui_args_="$(prefix_args "$UI_ARGS" "-w")"
+          launcher_args=" $launcher_args --wallet $UI $ui_args_"
+        fi
+        node_args_="$(prefix_args "$node_args" "-n")"
+        launcher_args=" $launcher_args --updater /usr/bin/env -u bash --update-archive $updater_file"
+        launcher_args=" $launcher_args --node-timeout 5 $node_args_ --system-start 0 --wlogging --db-path $run_dir/node-db$i $rts_opts $reb $no_ntp $keys_args --wallet-log-path logs/launcher/wallet-node-log.txt --node-log-path logs/launcher/node-log.txt"
```
gromakovsky: 1. What's the point in passing `--node-log-path logs/launcher/node-log.txt`? Because of that, the main problem we had during last weeks is still not solved. I launched it locally and there are still no logs in console. That's the main goal!
2. Using constant paths for logs in demo is quite bad idea, because we launch demo multiple times and want logs to be separated. See how it's done in master, logs are put into directory which contains date in its name.

auxx/src/Plugin.hs
v0d1ch: I didn't load the stylish-haskell at all. Now I have set it to re-format on save.

scripts/launch/demo.sh
v0d1ch: This fixes the error I was getting before , @volhovm helped with this comment on slack
> So there's ` gen_kademlia_topology` function in `common-functions.sh`
> IT generates topology/kademlia files, but before it does it `rm -v`s them.
> The reason why it fails for you is because `set -e` is turned on i guess.
> `rm -v` returns nonzero codes which terminates the script. Just `set +e`, call the function, do `set -e` back > later.

scripts/launch/demo.sh
v0d1ch: ¯\_(ツ)_/¯ my task was to add logging, I am not sure who wrote the function, I don't mess too much with bash scripts

tools/src/launcher/Main.hs
v0d1ch: Will do, this was copy/pasta

tools/src/launcher/Main.hs
v0d1ch: @georgeee  asked to have `[node]` and `[wallet]` prepended to log output. If you think there is better solution I will gladly implement it.

tools/src/launcher/Main.hs
v0d1ch: I refactored that

tools/src/launcher/Main.hs
v0d1ch: True this is bollocks. I changed everything.

scripts/launch/demo.sh
v0d1ch: So you are saying that console logs need to happen always and if you pass in `--node-log-path` then logs are also written to file ?

auxx/src/Plugin.hs
gromakovsky: And now something is really wrong. There should be no changes in this file at all, but you changed formatting of all importsn
Apparently your `stylish-haskell` is misconfigured. We have proper stylish-haskell config in the root of the repository. However, it requires version 0.8.1.0. Probably your stylish-haskell is older. In this case please update. Otherwise we need to figure out what's wrong with your setup.

scripts/launch/demo.sh
gromakovsky: Right. I won't even try to understand bash code, will only check that nothing is broken. And if something is broken due to bash code, I will ask George to fix it, because it's he who wrote it.

scripts/launch/demo.sh
gromakovsky: Ok, this change is no longer necessary after #2062. Please revert it.

scripts/launch/demo.sh
gromakovsky: I guess I clarified it in Slack. Let me know if something is not clear yet.

tools/src/launcher/Main.hs
```diff
@@ -258,7 +277,9 @@ main =
             & Log.lcFilePrefix .~ loLauncherLogsPrefix
             & Log.lcTree %~ case loLauncherLogsPrefix of
                   Nothing ->
-                      identity
+                      -- identity
+                      set Log.ltFiles [Log.HandlerWrap "launcher" Nothing] .
+                      set Log.ltSeverity (Just Log.Debug)
```
gromakovsky: Sorry, can you explain this change?

tools/src/launcher/Main.hs
```diff
@@ -515,11 +523,73 @@ spawnNode (path, args, mbLogPath) = do
...
+           if shouldLog then
+               liftIO $ do
+               let cr = createProc PInherit path args Nothing
+               system' phvar cr mempty NWallet
+           else
+               liftIO $ do
+               let cr = createProc PNoStream path args Nothing
+               system' phvar cr mempty NWallet
+
+createProc :: ProcessType -> FilePath -> [Text] -> (Maybe Handle) -> Process.CreateProcess
```
gromakovsky: Why do you need `ProcessType` type at all? Why not just pass `StdStream` from `System.Process`? The only different between these types is that your `PHandle` doesn't have `Handle` inside, but `UseHandle` from `StdStream` does. However, whenever you pass `PHandle`, you also pass `Just Handle`.
I really don't see any good reason to have `ProcessType` type. Just use `StdStream`.

gromakovsky: If you pass `StdStream`, this function can be written much-much shorter.
```
createProc stdStream path args = (Process.proc path (map toString args))
  { Process.std_in = Process.CreatePipe
  , Process.std_out = stdStream
  , Process.std_err = stdStream
  }
```

tools/src/launcher/Main.hs
```diff
@@ -515,11 +523,73 @@ spawnNode (path, args, mbLogPath) = do
...
             return (ph, asc)
 
-runWallet :: (FilePath, [Text]) -> M ExitCode
-runWallet (path, args) = do
-    logNotice "Starting the wallet"
-    view _1 <$>
-        liftIO (readProcessWithExitCode path (map toString args) mempty)
+runWallet :: Bool -> (FilePath, [Text], Maybe FilePath) -> M ExitCode
+runWallet shouldLog (path, args, mLogPath) = do
+    customLogger LInfo NWallet "Starting the wallet"
```
gromakovsky: Why do you print it using `customLogger` now? This message corresponds to the launcher's logic, so it should still be printed using `logNotice` (i. e. launcher's logging).

tools/src/launcher/Main.hs
```diff
@@ -96,6 +105,10 @@ data LauncherOptions = LO
 -- | The concrete monad where everything happens
 type M a = (HasConfigurations, HasCompileInfo) => Log.LoggerNameBox IO a
 
+data LType = LInfo | LNotice | LWarning | LError
```
gromakovsky: AFAIU, only only `LInfo` and `LError` are used. Do we need other two? If we want it to contain only those log types which we are going to use, then let's have only two constructors. If we want all possible log types, then why not just use `Severity` type from `log-warper`?

tools/src/launcher/Main.hs
```diff
@@ -96,6 +105,10 @@ data LauncherOptions = LO
...
+data NodeType = NWallet | NNode
```
gromakovsky: `NodeType` name looks confusing to me. This definition makes me think that `NWallet` is `cardano-node` (from `cardano-sl-wallet`) and `NNode` is `cardano-sl-node-simple` (from `cardano-sl-node`).
But `NWallet` is not a node at all, it's frontend!
Maybe call it `data Executable = EWallet | ENode`? Dunno, maybe there is better name. But wallet is not a node.

tools/src/launcher/Main.hs
```diff
@@ -559,14 +629,18 @@ system'
...
-system' phvar p sl = liftIO (do
+system' phvar p sl nt = liftIO (do
     let open = do
-            (m, _, _, ph) <- Process.createProcess p
+            (m, stdO, stdE, ph) <- Process.createProcess p
             putMVar phvar ph
             case m of
-                Just hIn -> IO.hSetBuffering hIn IO.LineBuffering
+                Just hIn -> do
+                    _ <- asyncLog (fromJust stdO) (fromJust stdE) ph nt
```
gromakovsky: Wait, AFAIU, it means that now `system'` will apply `asyncLog` logic to the spawned process every time, unconditionally. Are we sure this is what we want? What if we spawn process with `std_out = Inherit`? We will read out own output and print it our output, then read again, etc. Won't it lead to infinite loop?
Anyway, even if it won't, I don't see any reason to do it.
There is only one case when we want to apply `asyncLog` logic: when want to print output of both node and wallet.

gromakovsky: Also are you sure that `std0` and `stdE` will always be `Just`? Can you write a comment explaining why it's true. It's not obvious from this code, so it's better to have a comment explaining that.

gromakovsky: AFAIU, it only makes sense to use `asyncLog` when the process is spawned with `std_out = CreatePipe` and `std_err = CreatePipe` and maybe `UseHandle`.
If it's `Inherit`, there will be infinite loop probably. If it's `NoStream`, `std0` and `stdE` will be `Nothing` most likely.

tools/src/launcher/Main.hs
gromakovsky: Frankly, I'd like to avoid printing output of two executables to the same stream. I think it will be mess. Some messages take several lines. If outputs of two executables are combined, these messages may be mixed and it will be hard to read them.
`asyncLog` approach assumes that each message is 1 line. But it's not the case.
Moreover, after code review I see that it's easy to misuse `asyncLog`, for instance by applying to the process with inherited stdout (which leads to infinite loop).
And in fact there are no cases when we need to print output of both wallet and node to console. We only need node to be able to log to console.
It would be easier for me to review if this function didn't exist :)
But if you want to leave it, then ok, just use it properly. It should be used only when both wallet and node print to console. It means that `loWalletLogging` is `True`, `loWalletLogPath` is `Nothing` and `loNodeLogPath` is `Nothing`. Currently it's used unconditionally by `system'` which is wrong.

tools/src/launcher/Main.hs
```diff
@@ -515,11 +523,73 @@ spawnNode (path, args, mbLogPath) = do
...
+            system' phvar cr mempty NWallet
+        Nothing ->
+           if shouldLog then
+               liftIO $ do
+               let cr = createProc PInherit path args Nothing
+               system' phvar cr mempty NWallet
+           else
+               liftIO $ do
+               let cr = createProc PNoStream path args Nothing
+               system' phvar cr mempty NWallet
```
gromakovsky: It is copy-pasted, please refactor.

tools/src/launcher/Main.hs
v0d1ch: Oops ?

tools/src/launcher/Main.hs
v0d1ch: Done

tools/src/launcher/Main.hs
v0d1ch: Done

tools/src/launcher/Main.hs
v0d1ch: Done with two constructors, I don't think we need more.

tools/src/launcher/Main.hs
v0d1ch: Done

tools/src/launcher/Main.hs
v0d1ch: Right, I am checking now what constructor is used to create `StdStream`

tools/src/launcher/Main.hs
v0d1ch: Done

scripts/launch/demo.sh
v0d1ch: Done

tools/src/launcher/Main.hs
```diff
@@ -510,10 +506,10 @@ spawnNode (path, args, mbLogPath) = do
             -- fails when it prints unicode characters. In the meantime, don't print it.
             -- See DAEF-12.
             -- printf ("Redirecting node's stdout and stderr to "%fp%"\n") logPath
-            return $ createProc PHandle path args (Just logHandle)
-        Nothing -> return $ createProc PInherit path args Nothing
+            return $ createProc (Process.UseHandle logHandle) path args
```
v0d1ch: Do we even need this anymore ? Maybe I can just do `createProc Process.Inherit path args` and delete this as well as node and wallet log path fields if the nodes already print log to file by default.

auxx/src/Plugin.hs
```diff
@@ -11,34 +11,34 @@ module Plugin
...
+import           Pos.Util.TimeWarp          (jsonLog)
+import           Pos.WorkMode               (EmptyMempoolExt, RealMode, RealModeContext)
+
+import           AuxxOptions                (AuxxOptions (..))
+import           Command                    (createCommandProcs)
 import qualified Lang
-import           Mode (MonadAuxxMode)
-import           Repl (WithCommandAction (..), PrintAction)
+import           Mode                       (MonadAuxxMode)
+import           Repl                       (PrintAction, WithCommandAction (..))
```
gromakovsky: Revert changes in this file please.

tools/src/launcher/Main.hs
```diff
@@ -13,64 +13,70 @@
...
+import           Pos.Launcher                         (HasConfigurations, withConfigurations)
+import           Pos.Launcher.Configuration           (ConfigurationOptions (..))
+import           Pos.Reporting.Methods                (compressLogs, retrieveLogFiles, sendReport)
+import           Pos.ReportServer.Report              (ReportType (..))
+import           Pos.Update                           (installerHash)
+import           Pos.Update.DB.Misc                   (affirmUpdateInstalled)
+import           Pos.Util                             (HasLens (..), directory, postfixLFields,
+                                                       sleep)
+import           Pos.Util.CompileInfo                 (HasCompileInfo, retrieveCompileTimeInfo,
+                                                       withCompileInfo)
```
gromakovsky: Setup stylish-haskell properly please. I can't accept these changes in import section. See also: https://github.com/input-output-hk/cardano-sl/pull/1629#discussion_r155807686

tools/src/launcher/Main.hs
```diff
@@ -515,11 +519,47 @@ spawnNode (path, args, mbLogPath) = do
...
+            createDirectoryIfMissing True (directory lp)
+            (_, logHandle) <- (lp,) <$> openFile lp AppendMode
+            liftIO $ IO.hSetBuffering logHandle IO.LineBuffering
+            let cr = createProc (Process.UseHandle logHandle) path args
+            system' phvar cr mempty EWallet
+        Nothing ->
+           let cr = if shouldLog then
+                        Process.Inherit
+                    else Process.NoStream
+           in liftIO $ system' phvar (createProc cr path args) mempty EWallet
```
gromakovsky: `liftIO` is most likely redundant here, because there is another `liftIO` before `case mLogPath of` above. The same is true for `liftIO $ IO.hSetBuffering logHandle IO.LineBuffering`.

tools/src/launcher/Main.hs
```diff
@@ -559,16 +599,31 @@ system'
...
     -- ^ Exit code
-system' phvar p sl = liftIO (do
+system' phvar p sl nt = liftIO (do
     let open = do
-            (m, _, _, ph) <- Process.createProcess p
+            (hIn, stdO, stdE, ph) <- Process.createProcess p
             putMVar phvar ph
-            case m of
-                Just hIn -> IO.hSetBuffering hIn IO.LineBuffering
+            case [hIn,stdO, stdE] of
```
gromakovsky: It's more natural to use a tuple (`(hIn, std0, stdE)`), rather than a list.

tools/src/launcher/Main.hs
```diff
@@ -515,11 +519,47 @@ spawnNode (path, args, mbLogPath) = do
...
+
+asyncLog :: Handle -> Handle -> ProcessHandle -> Executable -> IO ExitCode
+asyncLog stdO stdE pH nt = do
+    withAsync (forever $ IO.hGetLine stdO >>= customLogger LInfo nt) $ \_ ->
+        withAsync (forever $ IO.hGetLine stdE >>= customLogger LError nt) $ \_ -> do
+        waitForProcess pH
+
+customLogger :: Log.CanLog m => LType -> Executable -> String -> m ()
+customLogger logType logName logStr = do
+    Log.usingLoggerName (Log.LoggerName logNameStr) $
```
gromakovsky: `LoggerName` has `IsString` instance (I suppose), so applying `LoggerName` constructor is likely redundant.

tools/src/launcher/Main.hs
```diff
@@ -515,11 +519,47 @@ spawnNode (path, args, mbLogPath) = do
...
+    (Process.proc path (map toString args))
+            { Process.std_in = Process.CreatePipe
+            , Process.std_out = stdStream
+            , Process.std_err = stdStream
+            }
+
+asyncLog :: Handle -> Handle -> ProcessHandle -> Executable -> IO ExitCode
+asyncLog stdO stdE pH nt = do
+    withAsync (forever $ IO.hGetLine stdO >>= customLogger LInfo nt) $ \_ ->
+        withAsync (forever $ IO.hGetLine stdE >>= customLogger LError nt) $ \_ -> do
```
gromakovsky: `Data.Text.IO` from `text` package also has `hGetLine` function which should be more efficient. In general we tend to avoid `String` unless there is no way to avoid it.

tools/src/launcher/Main.hs
```diff
@@ -515,11 +519,47 @@ spawnNode (path, args, mbLogPath) = do
...
     logNotice "Starting the wallet"
-    view _1 <$>
-        liftIO (readProcessWithExitCode path (map toString args) mempty)
+    phvar <- newEmptyMVar
+    liftIO $ case mLogPath of
+        Just lp -> do
+            createDirectoryIfMissing True (directory lp)
+            (_, logHandle) <- (lp,) <$> openFile lp AppendMode
+            liftIO $ IO.hSetBuffering logHandle IO.LineBuffering
+            let cr = createProc (Process.UseHandle logHandle) path args
```
gromakovsky:
```
            createDirectoryIfMissing True (directory lp)
            (_, logHandle) <- (lp,) <$> openFile lp AppendMode
            liftIO $ IO.hSetBuffering logHandle IO.LineBuffering
            let cr = createProc (Process.UseHandle logHandle) path args
```

These four lines are duplicated between `runWallet` and `spawnNode`.
Can you refactor please?

tools/src/launcher/Main.hs
```diff
@@ -559,16 +599,31 @@ system'
...
+                [(Just hndl), (Just o), (Just e)] -> do
+                    -- log certain kind of processes since
+                    -- inherited ones might create infinite loop
+                    case (Process.std_out p, Process.std_err p) of
+                        (Process.CreatePipe, Process.CreatePipe) -> do
+                            _ <- asyncLog o e ph nt
+                            return ()
+                        _ -> return ()
+                    case (Process.std_out p, Process.std_err p) of
+                        (Process.UseHandle _, Process.UseHandle _) -> do
```
gromakovsky: `UseHandle` means that we want to redirect all console output to a file. Previously `redirect` meant that **instead** of printing to console we were printing to a file. But after this change messages will be printed to a file **and** to console.
First, I think that more natural meaning of redirect is to print only to a file.
Second, I'd like not to change old behaviour.

tl;dr: let's use `asyncLog` only in `CreatePipe` case.

tools/src/launcher/Main.hs
```diff
@@ -442,13 +462,9 @@ runUpdater ndbp (path, args, runnerPath, mUpdateArchivePath) = do
...
 runUpdaterProc path args = liftIO $ do
-    let cr = (Process.proc (toString path) (map toString args))
-                 { Process.std_in  = Process.CreatePipe
-                 , Process.std_out = Process.CreatePipe
-                 , Process.std_err = Process.CreatePipe
-                 }
+    let cr = createProc Process.CreatePipe path args
     phvar <- newEmptyMVar
-    system' phvar cr mempty
+    system' phvar cr mempty EWallet
```
gromakovsky: Hm, updater is not wallet, it's another executable. Perhaps it makes sense to add `EUpdater`.

tools/src/launcher/Main.hs
```diff
@@ -479,33 +495,21 @@ spawnNode (path, args, mbLogPath) = do
...
-                 }
+            _ <- createDirectoryIfMissing True (directory lp)
+            logHandle <- openFile lp AppendMode
+            liftIO $ IO.hSetBuffering logHandle IO.LineBuffering
+            -- TODO (jmitchell): Find a safe, reliable way to print `logPath`. Cardano
+            -- fails when it prints unicode characters. In the meantime, don't print it.
+            -- See DAEF-12.
+            -- printf ("Redirecting node's stdout and stderr to "%fp%"\n") logPath
+            return $ createProc (Process.UseHandle logHandle) path args
+        Nothing -> return $ createProc Process.Inherit path args
```
gromakovsky: It should be `CreatePipe`, not `Inherit`, when wallet also wants to log to console.
That's actually the only case when we want to use `asyncLog`: when both executables want to log to console.
If wallet's output is redirected to a file or disabled, it should be `Inherit`.

The same is true for wallet: if node prints to console, wallet should use `CreatePipe` instead of `Inherit` (when wallet logging is enabled).

tools/src/launcher/Main.hs
```diff
@@ -515,11 +519,47 @@ spawnNode (path, args, mbLogPath) = do
...
+customLogger :: Log.CanLog m => LType -> Executable -> String -> m ()
+customLogger logType logName logStr = do
+    Log.usingLoggerName (Log.LoggerName logNameStr) $
+        case logType of
+            LInfo  -> logInfo   $ toText logStr
+            LError -> logError  $ toText logStr
+    where
+        logNameStr = case logName of
+            ENode   -> "node"
+            EWallet -> "wallet"
```
gromakovsky: Btw, I didn't manage to make this `customLogger` work. Maybe I did something wrong, I will re-test after you fix everything I pointed so far.
I suspect that's because `node` and `wallet` logger names have default severity, which is `Warning`.
And to be honest, I don't see why it's necessary to use `log-warper`'s logging here, why not just use simple `putStrLn`? You can add desired prefix using `mappend`.

tools/src/launcher/Main.hs
gromakovsky: It's better to leave this code. Node and wallet can log to file, but:
1. It's not default behaviour, it depends on logging config. We pass config which makes them print to file, but we may forget or something.
2. Not all output is done via log-warper. For instance, if an uncaught exception happens, it will be printed to `stderr` by runtime system. So the only way to make it appear in a file is to redirect `stderr` to this file.

tools/src/launcher/Main.hs
v0d1ch: I used `log-warper` since I noticed it was introduced to this file some time ago. If the logs need to be colored I will revert back.
https://imgur.com/a/kihiN

v0d1ch:
@gromakovsky I made changes per your review. I also updated my stylish-haskell so formatting should be fine now.

scripts/launch/demo.sh
```diff
@@ -113,24 +121,39 @@ while [[ $i -lt $panesCnt ]]; do
...
+        updater_file="$config_dir/updater$i.sh"
+        launcher_=$(find_binary cardano-launcher)
+        launcher_args="--node $node_ "
+        if [[ "$UI" != "" ]]; then
+          ui_args_="$(prefix_args "$UI_ARGS" "-w")"
+          launcher_args=" $launcher_args --wallet $UI $ui_args_"
+        fi
+        node_args_="$(prefix_args "$node_args" "-n")"
+        launcher_args=" $launcher_args --updater /usr/bin/env -u bash --update-archive $updater_file"
+        launcher_args=" $launcher_args --node-timeout 5 $node_args_ --system-start 0 --wlogging --db-path $run_dir/node-db$i $rts_opts $reb $no_ntp $keys_args"
```
gromakovsky: `--wlogging` should be enabled only if `--wallet` is passed to the launcher (i. e. `"$UI" != ""`).

tools/src/launcher/Main.hs
```diff
@@ -479,33 +487,18 @@ spawnNode (path, args, mbLogPath) = do
...
-                 { Process.std_in  = Process.CreatePipe
-                 , Process.std_out = Process.UseHandle logHandle
-                 , Process.std_err = Process.UseHandle logHandle
-                 }
+            createLogFileProc path args lp
+            -- TODO (jmitchell): Find a safe, reliable way to print `logPath`. Cardano
+            -- fails when it prints unicode characters. In the meantime, don't print it.
+            -- See DAEF-12.
+            -- printf ("Redirecting node's stdout and stderr to "%fp%"\n") logPath
+        Nothing -> return $ createProc Process.CreatePipe path args
```
gromakovsky: Once again, see this comment: https://github.com/input-output-hk/cardano-sl/pull/1629#discussion_r155955761

It should be `CreatePipe` when wallet logging is enabled (and not redirected) and `Inherit` otherwise. So you need to pass one more argument to this function.
The same is true for wallet.

tools/src/launcher/Main.hs
```diff
@@ -515,11 +508,49 @@ spawnNode (path, args, mbLogPath) = do
...
+
+asyncLog :: Handle -> Handle -> ProcessHandle -> Executable -> IO ExitCode
+asyncLog stdO stdE pH nt = do
+    withAsync (forever $ T.hGetLine stdO >>= customLogger nt) $ \_ ->
+        withAsync (forever $ T.hGetLine stdE >>= customLogger nt) $ \_ -> do
+        waitForProcess pH
+
+customLogger :: Executable -> Text -> IO ()
+customLogger loggerName logStr = do
+    putStrLn $ logNameStr <> logStr
```
gromakovsky: As a small nitpick: you can pass some value denoting whether `logStr` is from `stdout` or `stderr` and print to `stdout` or `stderr` here. Currently all output, even `stderr`, will be printed to `stdout`.

scripts/launch/demo.sh
v0d1ch: If I do this then run the `us-test.sh` script it fails with 
`Invalid argument `scripts/launch/ui-simulator.sh'`

tools/src/launcher/Main.hs
```diff
@@ -474,38 +483,28 @@ writeWindowsUpdaterRunner runnerPath = liftIO $ do
...
-    -- See DAEF-12.
-
-    -- printf ("Redirecting node's stdout and stderr to "%fp%"\n") logPath
-    liftIO $ IO.hSetBuffering logHandle IO.LineBuffering
-    let cr = (Process.proc (toString path) (map toString args))
-                 { Process.std_in  = Process.CreatePipe
-                 , Process.std_out = Process.UseHandle logHandle
-                 , Process.std_err = Process.UseHandle logHandle
-                 }
+            let cr = if shouldLog then
```
v0d1ch: Should I pass in the wallet log path as well and check if it `Nothing` before printing to console or wallet logging flag is enough ?

scripts/launch/demo.sh
gromakovsky: What exactly did you try to do?
I think you should define `wlogging=""` before `if [[ "$UI" != "" ]]; then` and define `wlogging=" --wlogging "` inside that `if`.
Then you can change `--wlogging` to `$wlogging` on this line.

gromakovsky: Or even better, you can add `--wlogging` on line 147:
> launcher_args=" $launcher_args --wallet $UI $ui_args_"

becomes

> launcher_args=" $launcher_args --wlogging --wallet $UI $ui_args_"

tools/src/launcher/Main.hs
```diff
@@ -515,11 +514,55 @@ spawnNode (path, args, mbLogPath) = do
...
+    (Process.proc path (map toString args))
+            { Process.std_in = Process.CreatePipe
+            , Process.std_out = stdStream
+            , Process.std_err = stdStream
+            }
+
+asyncLog :: Handle -> Handle -> ProcessHandle -> Executable -> IO ExitCode
+asyncLog stdO stdE pH nt = do
+    withAsync (forever $ T.hGetLine stdO >>= customLogger stdO nt) $ \_ ->
+        withAsync (forever $ T.hGetLine stdE >>= customLogger stdE nt) $ \_ -> do
```
gromakovsky: Wait, does it even work? AFAIU, you are passing the same stream as the one you read from (`stdE`). You should pass `stderr` of our process (https://www.stackage.org/haddock/lts-9.1/base-4.9.1.0/System-IO.html#v:stderr). The same is true for `stdout`.

tools/src/launcher/Main.hs
```diff
@@ -515,11 +514,55 @@ spawnNode (path, args, mbLogPath) = do
             logInfo "Node has started"
             return (ph, asc)
 
-runWallet :: (FilePath, [Text]) -> M ExitCode
-runWallet (path, args) = do
+runWallet
+    :: Bool                               -- ^ wallet logging
+    -> (FilePath, [Text], Maybe FilePath) -- ^ Wallet, its args, wallet log file
+    -> (FilePath, [Text], Maybe FilePath) -- ^ Node, its args, node log file
```
gromakovsky: Don't pass redundant args, it's enough to pass `Maybe FilePath`, you ignore other two values.

tools/src/launcher/Main.hs
```diff
@@ -474,38 +483,28 @@ writeWindowsUpdaterRunner runnerPath = liftIO $ do
 
 spawnNode
     :: (FilePath, [Text], Maybe FilePath)
+    -> Bool -- Wallet logging
     -> M (ProcessHandle, Async ExitCode)
-spawnNode (path, args, mbLogPath) = do
+spawnNode (path, args, mbLogPath) shouldLog = do
```
gromakovsky: `shouldLog` name is confusing. One may think it applies to node (i. e. whether node should log), but it's about wallet (whether wallet should log).

tools/src/launcher/Main.hs
gromakovsky: Yes, you should also check whether wallet's output is redirected to a file. You can pass `doesWalletLogToFile :: Bool` variable which will be defined (outside this function) as `doesWalletLogToFile = isNothing walletLogPath && walletLog`.

v0d1ch: I named it `doesWalletLogToConsole` since when the `FilePath` is `Nothing` and `walletLog` is `True` then we want to log to console

scripts/launch/demo.sh
v0d1ch: I did it also on line 147 but had the `--wallet` before  `--wlogging` flag and that is why it failed. All good now.

tools/src/launcher/Main.hs
v0d1ch: Done. It worked before but it is not the output we are trying to get.

tools/src/launcher/Main.hs
```diff
@@ -15,26 +15,27 @@ import           Universum
...
-                                                       withAsyncWithUnmask)
+                                                       withAsync, withAsyncWithUnmask)
 import           Control.Exception.Safe (tryAny)
 import           Control.Lens (makeLensesWith)
 import qualified Data.ByteString.Lazy as BS.L
 import           Data.List (isSuffixOf)
+import           Data.Maybe (isNothing)
 import qualified Data.Text.IO as T
 import           Data.Time.Units (Second, convertUnit)
+import           Data.Tuple.Select (sel3)
```
gromakovsky: It's not written anywhere, but we don't use `tuple` library. We use functionality from `lens` instead. `sel3` is `^. _3`. Both `^.` and `_3` are exported from `Universum`.

tools/src/launcher/Main.hs
gromakovsky: Oh, right, my mistake.

docs/how-to/test-update-system-locally.md
```diff
@@ -0,0 +1,39 @@
...
+      a. Prepare alt configuration with changed application version constant
+      b. Propose updated with script that simply replaces one configuration file (in setup) with other
+  2. Launch wallet node with launcher, Daedalus (or Daedalus emulator)
+      a. When update arrives Deadalus to finish, launcher to restart Daedalus + node
+
+## Prerequisites
+
+Additional utilities to be installed:
+
+* webfs (webfsd) server
```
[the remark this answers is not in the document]
volhovm: Too bad.

scripts/launch/Test.hs
```diff
@@ -0,0 +1,214 @@
...
+    updHash <- extractFromAuxxOut "updHash" $
+        "Update proposal submitted along with votes, upId: " *>
+        (T.pack <$> some hexDigit)
+    TIO.putStrLn $ fileHash <> " " <> updHash
+
+    shells ("mkdir run/serve-upd; mv test_updater.sh run/serve-upd/"<>fileHash) empty
+
+
+
+-- buildAMI :: NixOpsConfig -> IO ()
```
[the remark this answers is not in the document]
volhovm: This must be removed.

scripts/launch/Test.hs
```diff
@@ -0,0 +1,214 @@
...
+{-# LANGUAGE TupleSections       #-}
+{-# LANGUAGE TypeApplications    #-}
+{-# LANGUAGE ViewPatterns        #-}
+{-# LANGUAGE PackageImports      #-}
+
+import           Control.Lens
+import           Control.Monad             (forM_, void, when)
+import           Control.Monad.Except      (ExceptT (..), runExceptT)
+import           Control.Monad.Trans       (lift)
+import "lens-aeson" Data.Aeson.Lens
```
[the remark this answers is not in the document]
neongreen: `-XPackageImports`

docs/how-to/test-update-system-locally.md
```diff
@@ -0,0 +1,39 @@
+# Testing update system locally
+
+This instruction covers how to test software update locally:
+
+  1. Prepare and propose software update
+      a. Prepare alt configuration with changed application version constant
+      b. Propose updated with script that simply replaces one configuration file (in setup) with other
+  2. Launch wallet node with launcher, Daedalus (or Daedalus emulator)
```
neongreen: What's “Daedalus emulator”? It's just a nitpick but I'd like this sentence to be clearer

docs/how-to/test-update-system-locally.md
```diff
@@ -0,0 +1,39 @@
...
+      a. When update arrives Deadalus to finish, launcher to restart Daedalus + node
```
neongreen: Sounds like broken English.

scripts/launch/Test.hs
```diff
@@ -0,0 +1,214 @@
...
+
+bvL = key "update" . key "lastKnownBlockVersion"
+
+bvAltL = bvL . key "bvAlt" . _Number
+bvMinorL = bvL . key "bvMinor" . _Number
+bvMajorL = bvL . key "bvMajor" . _Number
+
+updateTest :: Y.Value -> Config -> IO ()
+updateTest config (Config{..}) = do
+    let config' = config & appNameL .~ walletAppName
```
neongreen:
```haskell
let config' = config & appNameL .~ walletAppName
                     & svL +~ 1
```

scripts/launch/Test.hs
```diff
@@ -0,0 +1,214 @@
...
+        ExitSuccess -> do
+            echo "Proposing an update succeeded, here are the logs:"
+            mapM_ (echo . mappend "\t") auxxOut
+        ExitFailure _ -> do
+            echo "Proposing an update failed, see auxx output:"
+            mapM_ (echo . mappend "\t") auxxOut
+            exit exitCode
+    let matchInAuxxOut pat = do
+            (lineToText -> t) <- NE.toList auxxOut
+            match (suffix pat) t
```
neongreen: I think that in this case it would've been sli-ighly better to avoid ViewPatterns.

```haskell
t <- NE.toList auxxOut
match (suffix pat) (lineToText t)
```

docs/how-to/test-update-system-locally.md
```diff
@@ -0,0 +1,39 @@
...
+## Prerequisites
+
+Additional utilities to be installed:
+
+* webfs (webfsd) server
+* uxterm to emulate UI when no `$UI` provided
+
+## With Daedalus
+
+Write command to launch Daedalus on your system to `Daedalus.sh`
```
neongreen: You renamed it so update this doc too.

tools/src/launcher/Main.hs
```diff
@@ -559,16 +603,26 @@ system'
...
     -- ^ Exit code
-system' phvar p sl = liftIO (do
+system' phvar p sl nt = liftIO (do
     let open = do
-            (m, _, _, ph) <- Process.createProcess p
+            (hIn, stdO, stdE, ph) <- Process.createProcess p
             putMVar phvar ph
-            case m of
-                Just hIn -> IO.hSetBuffering hIn IO.LineBuffering
+            case ((hIn, stdO, stdE)) of
```
neongreen: Extra parens.

tools/src/launcher/Main.hs
```diff
@@ -559,16 +603,26 @@ system'
...
+                ((Just hndl), (Just o), (Just e)) -> do
```
neongreen: And here: `(Just hndl, Just o, Just e)`.

neongreen:
^ I left a bunch of nitpicks but it's fine if you ignore them. Other than the one about updating the doc mentioning `Daedalus.sh`, please do it.

scripts/launch/Test.hs
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env stack
+-- stack script --resolver lts-9.12
```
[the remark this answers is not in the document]
v0d1ch: stack.yaml in root uses lts-9.1 so I have set it to that

docs/how-to/test-update-system-locally.md
v0d1ch: It does, I did some changes but don't really know what to write

docs/how-to/test-update-system-locally.md
v0d1ch: renamed to _script that serves as Daedalus emulator_

PR #1635, opened 2017-09-23
[CSM-450] Take mempool into account when using addresses lists

Martoon-00:
Actual set of addresses of given account is a union of addresses stored in db and related addresses participating in transactions of mempool.
Sometimes we omitted the second part, and in very rare cases it was essential.

Tested: via DAT with settings where fee tests failed


adinapoli-iohk:
@flyingleafe @Martoon-00 Is this PR still relevant?

If it is, we should merge it. If not, let's close it 😉

Martoon-00:
I guess we can close, this issue is getting resolved by @pva701 in a better way

PR #1636, opened 2017-09-23
[CSL-1687] Treat fixedTimeCQ value as seconds

gromakovsky:
Before Configuration refactoring this value was treated as seconds, but it
was broken during the refactoring.

It solves (1) and (2) from CSL-1687. I have no idea why overall chain quality is 0 there. Most likely that's just another consequence of this misconfiguration. I couldn't reproduce it locally. I can try to understand it if I have logs, but probably it's not useful to do it, it's better to check whether this fix helps (it most likely does) first.


jmitchell:
@gromakovsky, regarding overall chain quality can you check my latest update in CSL-1687?

PR #1639, opened 2017-09-24
[CSE-228] Add TotalAda endpoint

Pastafarianist:
Tested here https://issues.serokell.io/issue/CSE-228#comment=96-7725

### DIFF
--- a/explorer/src/Pos/Explorer/Web/Api.hs
+++ b/explorer/src/Pos/Explorer/Web/Api.hs
@@ -20,10 +20,11 @@ import           Universum
 
 import           Data.Proxy                   (Proxy (Proxy))
 
-import           Pos.Explorer.Web.ClientTypes (CAddress, CAddressSummary, CBlockEntry,
-                                               CBlockSummary, CGenesisAddressInfo,
-                                               CGenesisSummary, CHash, CTxBrief, CTxEntry,
-                                               CTxId, CTxSummary, Byte)
+import           Pos.Explorer.Web.ClientTypes (Byte, CAddress, CAddressSummary,
+                                               CBlockEntry, CBlockSummary, CCoin,
+                                               CGenesisAddressInfo, CGenesisSummary,
+                                               CHash, CTxBrief, CTxEntry, CTxId,
+                                               CTxSummary)
 import           Pos.Explorer.Web.Error       (ExplorerError)
 import           Pos.Types                    (EpochIndex)
 import           Servant.API                  ((:<|>), (:>), Capture, Get, JSON,
@@ -34,6 +35,10 @@ type PageNumber = Integer
 
 -- | Common prefix for all endpoints.
 type API = "api"
+
+type TotalAda = API
+    :> "totalada"
+    :> Get '[JSON] (Either ExplorerError CCoin)
 
 type BlocksPages = API
     :> "blocks"
@@ -116,7 +121,8 @@ type StatsTxs = API
 
 -- | Servant API which provides access to explorer
 type ExplorerApi =
-         BlocksPages
+         TotalAda
+    :<|> BlocksPages
     :<|> BlocksPagesTotal
     :<|> BlocksSummary
     :<|> BlocksTxs
--- a/explorer/src/Pos/Explorer/Web/Api.hs
+++ b/explorer/src/Pos/Explorer/Web/Api.hs
@@ -18,10 +18,11 @@ module Pos.Explorer.Web.Api
 
 import           Universum
 
+import           Data.Fixed                   (Micro)
 import           Data.Proxy                   (Proxy (Proxy))
 
 import           Pos.Explorer.Web.ClientTypes (Byte, CAddress, CAddressSummary,
-                                               CBlockEntry, CBlockSummary, CCoin,
+                                               CBlockEntry, CBlockSummary,
                                                CGenesisAddressInfo, CGenesisSummary,
                                                CHash, CTxBrief, CTxEntry, CTxId,
                                                CTxSummary)
@@ -37,8 +38,9 @@ type PageNumber = Integer
 type API = "api"
 
 type TotalAda = API
-    :> "totalada"
-    :> Get '[JSON] (Either ExplorerError CCoin)
+    :> "supply"
+    :> "ada"
+    :> Get '[JSON] (Either ExplorerError Micro)
 
 type BlocksPages = API
     :> "blocks"
--- a/explorer/src/Pos/Explorer/Web/Server.hs
+++ b/explorer/src/Pos/Explorer/Web/Server.hs
@@ -46,9 +46,9 @@ import           Pos.Block.Core                   (MainBlock, mainBlockSlot,
 import           Pos.Block.Types                  (Blund, Undo)
 import           Pos.Core                         (AddrType (..), Address (..), Coin,
                                                    EpochIndex, HeaderHash, Timestamp,
-                                                   difficultyL, gbHeader, gbhConsensus,
-                                                   getChainDifficulty, isRedeemAddress,
-                                                   isUnknownAddressType,
+                                                   coinToInteger, difficultyL, gbHeader,
+                                                   gbhConsensus, getChainDifficulty,
+                                                   isRedeemAddress, isUnknownAddressType,
                                                    makeRedeemAddress, mkCoin, siEpoch,
                                                    siSlot, sumCoins, timestampToPosix,
                                                    unsafeIntegerToCoin, unsafeSubCoin)
@@ -71,13 +71,13 @@ import           Pos.Explorer                     (TxExtra (..), getEpochBlocks,
                                                    getLastTransactions, getPageBlocks,
                                                    getTxExtra)
 import qualified Pos.Explorer                     as EX (getAddrBalance, getAddrHistory,
-                                                         getTxExtra)
+                                                         getTxExtra, getUtxoSum)
 import           Pos.Explorer.Aeson.ClientTypes   ()
 import           Pos.Explorer.Web.Api             (ExplorerApi, explorerApi)
 import           Pos.Explorer.Web.ClientTypes     (Byte, CAddress (..),
                                                    CAddressSummary (..),
                                                    CAddressType (..), CBlockEntry (..),
-                                                   CBlockSummary (..),
+                                                   CBlockSummary, CCoin,
                                                    CGenesisAddressInfo (..),
                                                    CGenesisSummary (..), CHash,
                                                    CTxBrief (..), CTxEntry (..),
@@ -118,6 +118,8 @@ explorerApp serv = serve explorerApi <$> serv
 
 explorerHandlers :: ExplorerMode ctx m => SendActions m -> ServerT ExplorerApi m
 explorerHandlers _sendActions =
+      apiTotalAda
+    :<|>
       apiBlocksPages
     :<|>
       apiBlocksPagesTotal
@@ -142,6 +144,7 @@ explorerHandlers _sendActions =
     :<|>
       apiStatsTxs
   where
+    apiTotalAda           = tryGetTotalAda
     apiBlocksPages        = getBlocksPagesDefault
     apiBlocksPagesTotal   = getBlocksPagesTotalDefault
     apiBlocksSummary      = catchExplorerError . getBlockSummary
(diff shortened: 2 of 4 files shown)


explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -186,6 +192,19 @@ explorerHandlers _sendActions =
...
+getTotalAda = do
+    utxoSum <- EX.getUtxoSum
+    -- getUtxoSum returns Integer, but for external consumers we want to provide Coin.
+    -- However, we don't want to return meaningless values to them, so we add truncation.
+    pure $ mkCCoin $ truncateIntegerToCoin utxoSum
+  where
+    truncateIntegerToCoin :: Integer -> Coin
+    truncateIntegerToCoin n
+        | n < 0 = mkCoin 0
+        | n > coinToInteger (maxBound :: Coin) = maxBound :: Coin
```
ksaric: This is a bit strange to me. We can return an error if the number is larger than the `maxBound`. 
A simple `throwM $ Internal "Coin amount is out of bounds!"` would be more explicit and would alert people that something is wrong.

explorer/src/Pos/Explorer/Web/Api.hs
```diff
@@ -35,6 +36,10 @@ type PageNumber = Integer
 -- | Common prefix for all endpoints.
 type API = "api"
 
+type TotalAda = API
+    :> "totalada"
```
ksaric: I would prefer to have named something more in the spirit of REST recommendations. 
So something like `currencies/ada` or `currencies?name=ada` (this is what I would prefer) or `currencies/ada(s)/total/` is more appropriate.
For example, this is a decent guide for the names - http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api

explorer/src/Pos/Explorer/Web/Server.hs
Martoon-00: I agree, it seems to be abnormal situation

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -186,6 +192,19 @@ explorerHandlers _sendActions =
 -- API Functions
 ----------------------------------------------------------------
 
+getTotalAda :: ExplorerMode ctx m => m CCoin
+getTotalAda = do
+    utxoSum <- EX.getUtxoSum
+    -- getUtxoSum returns Integer, but for external consumers we want to provide Coin.
+    -- However, we don't want to return meaningless values to them, so we add truncation.
+    pure $ mkCCoin $ truncateIntegerToCoin utxoSum
```
Martoon-00: By @Pastafarianist: `totalAda` endpoint returns lovelaces?

akegalj: good point. Do we return total ada or total lovelaces ?

explorer/src/Pos/Explorer/Web/Server.hs
akegalj: I agree. We shouldn't truncate money info. Throwing exception as @ksaric suggested seems like a right approach here

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -186,6 +193,20 @@ explorerHandlers _sendActions =
...
+getTotalAda = do
+    utxoSum <- EX.getUtxoSum
+    validateUtxoSum utxoSum
+    pure $ fromInteger utxoSum / 1e6
+  where
+    validateUtxoSum :: ExplorerMode ctx m => Integer -> m ()
+    validateUtxoSum n
+        | n < 0 = throwM $ Internal $
+            sformat ("Internal tracker of utxo sum has a negative value: "%build) n
+        | n > coinToInteger (maxBound :: Coin) = throwM $ Internal $
```
akegalj: @gromakovsky we are sure that utx sum will always fit into `Coin` , right ?

Because it seems to me that it is safer to throw exception here then to bound it.

@Pastafarianist said
> When I implemented it, I was thinking that since this was going to be displayed on CoinMarketCap, we probably wanted to make that endpoint as robust as possible — thus never return errors, but sometimes return invalid values instead.

but, as in practice, it should fit `Coin` in this corner use case (which I don't think it will happen) I would rather fail then show incorrect information about total ada.

What do you think @gromakovsky ?

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -192,18 +193,19 @@ explorerHandlers _sendActions =
...
 
-getTotalAda :: ExplorerMode ctx m => m CCoin
+getTotalAda :: ExplorerMode ctx m => m Micro
 getTotalAda = do
     utxoSum <- EX.getUtxoSum
-    -- getUtxoSum returns Integer, but for external consumers we want to provide Coin.
-    -- However, we don't want to return meaningless values to them, so we add truncation.
-    pure $ mkCCoin $ truncateIntegerToCoin utxoSum
+    validateUtxoSum utxoSum
+    pure $ fromInteger utxoSum / 1e6
```
ksaric: I have a question. Before you returned lovelaces (the smallest unit of currency), but now you aren't. 
Isn't this a `Coin`? And what is a `Micro`? Can you explain the connection to the currency/units?

explorer/src/Pos/Explorer/Web/Api.hs
```diff
@@ -37,8 +38,9 @@ type PageNumber = Integer
 type API = "api"
 
 type TotalAda = API
-    :> "totalada"
-    :> Get '[JSON] (Either ExplorerError CCoin)
+    :> "supply"
+    :> "ada"
+    :> Get '[JSON] (Either ExplorerError Micro)
```
ksaric: :neutral_face: I won't say it's great, but I'm not going to be a review Nazi either.

explorer/src/Pos/Explorer/Web/Server.hs
Martoon-00: I believe that there is a function which checks that coin is in bounds, smth like `integerToCoin`.
Please use it

explorer/src/Pos/Explorer/Web/Server.hs
Martoon-00: Now literally `ada` is returned as endpoint name suggests. `Micro` allows to keep number with resolution of `10^-6`, and that perfectly fits to represent `ada`

explorer/src/Pos/Explorer/Web/Api.hs
Martoon-00: Could someone explain me why `CCoin` was used here? I know that it is used in wallet to pass number through daedalus-bridge as string to prevent precision losses. But here we do not have bridge, right?

ksaric: Explorer frontend possibly? We do have a "bridge" for _explorer_, we are using `CCoin` on other endpoints (in the data structure, though).

explorer/src/Pos/Explorer/Web/Server.hs
gromakovsky: > we are sure that utx sum will always fit into Coin , right ?

We are sure only if we are sure that we have no bugs. So we are not sure.

> I would rather fail then show incorrect information about total ada.

But that's exactly what is done here, isn't it? `throwM Internal` looks quite reasonable.

akegalj: > But that's exactly what is done here, isn't it? throwM Internal looks quite reasonable.

right - don't know how I missed it. I thought it returns truncated Coin. Its ok. Sorry @Pastafarianist I missread the line

explorer/src/Pos/Explorer/Web/Api.hs
Martoon-00: If we have bridge for explorer, then I guess that `Mirco` is not ok here :/

akegalj: @Martoon-00 bridge is not using this endpoint (at least not atm). I am not sure how Micro is serialized. Will test it.

explorer/src/Pos/Explorer/Web/Server.hs
ksaric: Wait. Why don't we call that `Ada`? :confused:

akegalj:
To review both points from @Martoon-00 : 
> If we have bridge for explorer, then I guess that Mirco is not ok here :/ Bridge will truncate number to double. @akegalj Is that true?

atm we don't use it in the bridge. We are just exposing backend endpoint for  other to use this value.

> I believe that there is a function which checks that coin is in bounds, smth like integerToCoin

yes, there is `integerToCoin` already here https://github.com/input-output-hk/cardano-sl/blob/master/core/Pos/Core/Coin.hs#L82 which does all the checks @Pastafarianist made with his `validateUtxoSum`. 

As both of these are safe - they can be fixed later. Lets continue with merging this and fixing those minor points later. Issue to fix them here https://issues.serokell.io/issue/CSM-470

akegalj:
@Martoon-00 could you approve as both of the points seems minor and there is an issue to track them https://issues.serokell.io/issue/CSM-470

explorer/src/Pos/Explorer/Web/Api.hs
ksaric: But there is a point to what @Martoon-00 is saying. If we use this endpoint from _explorer_ do we have the corresponding client representation?

akegalj: @ksaric Micro is serializable somehow , I just didn't check it myself to see how it serializes things. The best would be just to return a number (or json alternatively).

I am currently compiling in order to check it

Martoon-00:
Yes, `CCoin` represents integer number. `Ada`s are measured in floating number with 6 digits after point, so `Micro` is used instead of `Coin`

ksaric:
:confused: 
So:
- `maxCoinVal = 45000000000000000` - lovelaces, `CCoin`
- `maxCoinVal = 45000000000.000000` - `Micro` = `Ada`

So why not `newtype CAda = CAda ...` with either `Text` or something suitable for PS and return that?
If we return `Micro`, we have to manually define the PS side (conversions).

Martoon-00:
@akegalj Just said that there is no purescript for explorer

ksaric:
@Martoon-00 He said "no bridge" because he was thinking about wallet "bridge". 
There is a "bridge" for _explorer_.

ksaric:
@Martoon-00 Sorry for the confusion. This is starting to sound more like an architectural discussion (patterns anyone?) than anything else.
There are two bridges, each bridge is for a separate frontend:
- wallet
- explorer

@akegalj Can you confirm we need a client representation for _explorer_ - the "bridge" from Haskell to Purescript?

akegalj:
> Can you confirm we need a client representation for explorer - the "bridge" from Haskell to Purescript?

@ksaric yes. Do we use Micro or this endpoint in the Explorer anywhere atm @ksaric ?

ksaric:
> Do we use Micro or this endpoint in the Explorer anywhere atm

@akegalj No, we can create the suggested newtype `CAda` to represent this.

akegalj:
> Also, we can create the suggested newtype CAda to represent this.

@ksaric yes, lets do this when this will be needed in UI. It is not needed right now.

I just checked how Micro is serialized and it gives scientific notation. 
```
curl -v http://localhost:8100/api/supply/ada
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8100 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8100 (#0)
> GET /api/supply/ada HTTP/1.1
> Host: localhost:8100
> User-Agent: curl/7.55.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
< Date: Wed, 27 Sep 2017 12:04:26 GMT
< Server: Warp/3.2.13
< Content-Type: application/json;charset=utf-8
< 
* Connection #0 to host localhost left intact
{"Right":4.499998899998801e10}
```

This looks ok to me.

Thoughts @Pastafarianist @ksaric @Martoon-00 ?

domenkozar:
@dcoutts good to go?

domenkozar:
I've opened #1664 (it needs immediate attention) and we'll be merging cardano-sl-1.0 to master, so I'm closing this one.

akegalj:
Tested here https://issues.serokell.io/issue/CSE-228#comment=96-7725

akegalj:
Changed base to cardano-sl-1.0 and reopened as it has been tested.

akegalj:
Closing due too cherry picked PR https://github.com/input-output-hk/cardano-sl/pull/1679

PR #1643, opened 2017-09-25
[CSM-452] Make all wallet code consider bootstrap era flag

ajnsit:


wallet/src/Pos/Wallet/Web/Methods/Logic.hs
```diff
@@ -157,37 +160,53 @@ getWallets = getWalletAddresses >>= mapM getWallet
...
 newAccount
     :: MonadWalletWebMode m
-    => AddrGenSeed -> PassPhrase -> CAccountInit -> m CAccount
+    => IsBootstrapEraAddr -> AddrGenSeed -> PassPhrase -> CAccountInit -> m CAccount
 newAccount = newAccountIncludeUnready False
 
+newAccountHandler
+    :: MonadWalletWebMode m
+    => AddrGenSeed -> PassPhrase -> CAccountInit -> m CAccount
+-- TODO: AJ: Hard coding isBootstrapEra as True for API calls. Check if that's okay.
```
akegalj: this looks ok for me. @flyingleafe right?

ajnsit: I kept it that way because I didn't want to change the existing behaviour of the API. But do you think it makes sense to add another address endpoint which allows controlling the value for isBootstrapEra?

wallet/src/Pos/Wallet/Web/Mode.hs
```diff
@@ -156,6 +157,16 @@ type MonadWalletWebMode m =
...
     )
 
+-- AJ: Need to use unblocking getCurrentSlot? Also using 0 as default??
+getIsBootstrapEra ::
+    ( MonadSlots ctx m
+    , MonadGState m
+    )
+    => m IsBootstrapEraAddr
+getIsBootstrapEra = do
+    e <- maybe 0 siEpoch <$> getCurrentSlot
```
flyingleafe: I'm not sure if it's okay to use 0 slot as default here. This would mean that if we couldn't get current slot in reward era for some reason, we'll create addresses with bootstrap distribution automatically, which is incorrect. I think that there's a better way to detect current era without blocking. cc @gromakovsky

wallet/src/Pos/Wallet/Web/Methods/Logic.hs
flyingleafe: No, I think that here `getBootstrapEra` should be used too. This way we won't change the behavior for bootstrap era and we'll have correct behavior when reward era comes.

wallet/src/Pos/Wallet/Web/Account.hs
```diff
@@ -62,12 +62,13 @@ getSKById wid = do
...
-    => PassPhrase
+    => IsBootstrapEraAddr
+    -> PassPhrase
     -> CWAddressMeta
     -> m EncryptedSecretKey
-getSKByAccAddr passphrase addrMeta@CWAddressMeta {..} = do
+getSKByAccAddr ibe passphrase addrMeta@CWAddressMeta {..} = do
     (addr, accKey) <-
-        deriveAccountSK passphrase (addrMetaToAccount addrMeta) cwamAccountIndex
+        deriveAccountSK ibe passphrase (addrMetaToAccount addrMeta) cwamAccountIndex
```
Martoon-00: Actually there are some places where we assume the only way to build an address by given secret key.
One of such places is just below this line.

There may be more caveats like this one. Please, try to emulate switching to another era (switching hardcoded value while nodes are off should be enough), whether, say, transaction creation will still work. I'll accept after that

adinapoli-iohk:
@flyingleafe What should we be doing with this PR?

gromakovsky:
We are not planning to work on this issue in nearest feature. This PR is already quite outdated and it will be even more outdated when we get back to this issue. Perhaps it will be easier to just redo it.
So I am closing it. In any case we can always reopen if necessary.

PR #1647, opened 2017-09-25
[CSL-1563] Use TH to fetch/print git revision

volhovM:
This PR provides a simple TH-based framework (`Pos.Util.CompileInfo`) to fetch the compile time data like git revision. Previous solution with cpp definition led to unneeded recompilations and now this problem is solved.


gromakovsky:
Also there was a request to print revision when `--version` is passed.

CODEOWNERS
```diff
@@ -41,13 +41,13 @@ core/Pos/Merkle.hs              @neongreen
 
 core/Pos/Util/Config.hs         @neongreen
 core/Pos/Util/Config/           @neongreen
-node/src/Pos/CompileConfig/     @neongreen
+lib/src/Pos/CompileConfig/     @neongreen
```
gromakovsky: Indentation :(

scripts/build/cardano-sl.sh
```diff
@@ -174,6 +174,8 @@ do
     spec_prj="wallet"
   elif [[ $var == "explorer" ]]; then
     spec_prj="explorer"
+  elif [[ $var == "node" ]]; then
+    spec_prj="node"
```
gromakovsky: You should also modify `sl+` case to build `node`.

auxx/src/Command/Tx.hs
```diff
@@ -215,6 +218,7 @@ sendTxsFromFile
        , HasUpdateConfiguration
        , HasNodeConfiguration
        , HasGtConfiguration
+       , HasCompileInfo
```
gromakovsky: It's strange that it's needed here.

lib/README.md
```diff
@@ -0,0 +1,85 @@
+<!-- THIS IS AUTOGENERATED FILE. DO NOT CHANGE IT MANUALLY! -->
```
gromakovsky: How is this file generated? Will it be correctly auto generated after being moved to `lib/`?

volhovM: cc @denisshevchenko

PR #1648, opened 2017-09-25
Set versions in .cabal files to 1.0.0

gromakovsky:


neongreen:
stack2nix fails, you probably want to synchronize with cardano-sl-1.0

PR #1651, opened 2017-09-26
[CSL-1627] Use latest version of time-warp-nt 

adinapoli-iohk:
This PR bumps `time-warp-nt` to its latest version, in order to fix:

* https://issues.serokell.io/issue/CSL-1697
* https://issues.serokell.io/issue/CSL-1652

In a nutshell, previous versions of `time-warp-nt` had a version of `removePeer` which was leaving, under certain circumstances, empty lists inside the peer routes, yielding those errors. In practice those errors were simply telling us that if an edge node disconnected we couldn't reach it anymore, but as it was being shown as an error in the logs was creating the false perception something was not working correctly, whereas it was just the manifestation of a normal behaviour of the system.

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -3043,15 +3043,11 @@ self: {
           description = "JSON encoding of ekg metrics";
           license = stdenv.lib.licenses.bsd3;
         }) {};
-      ekg-statsd = callPackage ({ base, bytestring, ekg-core, fetchgit, mkDerivation, network, stdenv, text, time, unordered-containers }:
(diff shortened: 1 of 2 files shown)


georgeee:
@adinapoli-iohk is it still actual? Can we close (discard) it?

adinapoli-iohk:
@georgeee We can close this, as the relevant changes were merged a long time ago into `time-warp-nt` and I think that at this stage we are already depending from a much more updated version of `time-warp-nt` anyway!

PR #1652, opened 2017-09-26
[CSM-469] Fix sync logic

Martoon-00:
Now we do not try to return any exact network difficulty if last header is
unknown.

Previous logic was unsafe, because under some curcimstances it may
return network difficulty as known and equal to local difficulty at the very
start of the node (before `knownLeader` gets initialized) and client would
suggest the node to be synced.

Tested: in dev mode with daedalus.
At first `networkDifficulty=null` (even if it's not the first launch of the node but with `NO_REBUILD` set) and daedalus patiently waits till sync

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -3039,15 +3039,11 @@ self: {
           description = "JSON encoding of ekg metrics";
           license = stdenv.lib.licenses.bsd3;
         }) {};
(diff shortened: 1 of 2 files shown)


Martoon-00:
Tested in mainnet as well

volhovM:
@Martoon-00 Very good. Rebase this on the top of `cardano-sl-1.0` and let's merge.

fersel:
Tested, works fine.

PR #1654, opened 2017-09-26
[CSL-1510] Make tests deterministic

neongreen:


core/Pos/Arbitrary/Crypto.hs
```diff
@@ -38,17 +37,13 @@ import           Pos.Crypto.Signing.Safe           (PassPhrase, createProxyCert,
...
-them safely requires randomness which must come from IO (we could use an
-'arbitrary' randomness generator for an 'Arbitrary' instance, but then what's
-the point of testing key generation when we use different generators in
-production and in tests?). So, we just generate lots of keys and seeds with
-'unsafePerformIO' and use them for everything.
+Generating keys takes time, so we just pregenerate lots of keys in advance
+and use them in 'Arbitrary' instances.
 -}
 
 keysToGenerate :: Int
```
rockbmb: If this number was chosen for a reason, perhaps a comment explaining why would be in order. This isn't too important.

neongreen: No reason really, just “big enough so that we wouldn't have collisions _too_ often”

PR #1659, opened 2017-09-27
[CSE-232] Allow CORS request by `cardano-mainnet`

sectore:
See https://issues.serokell.io/issue/CSE-232


ksaric:
@sectore Testnet too. 
`*` doesn't look that bad now, does it? :smile:

akegalj:
@domenkozar websocket connection is failing if we don't specify correct CORS (alternatively we can make it less restrictive - but lets not do that).

Can you confirm that the list of domain names for explorer is complete here.

domenkozar:
@akegalj looks good to me now, we should update this later on.

domenkozar:
There is now a separate PR for 1.0

PR #1661, opened 2017-09-27
[CSLD-136] Rollback and block-gen documents.

denisshevchenko:


docs/how-to/generate-blockchain.md
```diff
@@ -0,0 +1,33 @@
...
+
+It is possible situation when blockchain is not advanced for more than `k` slots.
+In this case we should generate it. Use `cardano-block-gen` command for it.
+
+## Basic Command
+
+The simplest example of the command:
+
+```
+$ stack exec -- cardano-block-gen   \
```
georgeee: Add flags `--configuration-file configuration.yaml`, `--configuration-key mainnet_full`

docs/how-to/rollback-blocks.md
```diff
@@ -0,0 +1,39 @@
+# How to Rollback Blocks
+
+It's possible when some broken data appear on the blockchain. In this case we
+have to rollback some blocks. Tool `cardano-auxx` provides such a functionality.
+
+First of all, we have to activate `repl`-mode:
+
+```
+$ stack exec -- cardano-auxx            \
```
georgeee: 1. Add flags `--configuration-file configuration.yaml`, `--configuration-key mainnet_full`
2. Provide commands to use via `cmd`, not `repl`

docs/how-to/generate-blockchain.md
```diff
@@ -0,0 +1,35 @@
...
+
+where:
+
+*  `1000` is a length of generated blockchain (number of blocks),
+*  `new-db` is a path to generated database.
+
+## Additional Parameters
+
+If the database already exists, we can use `--append` option. In this case new
+database won't be created and records will be appended to existing database.
```
gromakovsky: I don't think DevOps will ever want not to use `--append`. I would put it into command above.

docs/how-to/generate-blockchain.md
```diff
@@ -0,0 +1,35 @@
...
+
+It is possible to use a custom seed for generating blocks. Use `--seed` parameter
+to define integer number that will be used as a seed, for example `--seed 100`.
+If the value of seed is not defined, random integer number will be used instead.
+
+Since we generate blocks with random transactions it can be useful to limit the
```
gromakovsky: I don't think DevOps will ever want to generate random transactions. I would explicitly pass `--tx-count "(0,0)" `.

docs/how-to/generate-blockchain.md
```diff
@@ -0,0 +1,35 @@
...
+The simplest example of the command:
+
+```
+$ stack exec -- cardano-block-gen                \
+    --system-start 0                             \
+    --configuration-file node/configuration.yaml \
+    --configuration-key mainnet_full             \
+    --blocks 1000                                \
+    --generated-db new-db
+```
```
gromakovsky: This command won't work, because it's impossible to create valid blocks without secret keys. There is `--secret` parameter which can be used to pass secret key (multiple keys can be passed).
Also while I was reviewing it, I realized that current implementation of block-gen won't work for mainnet, because there we use heavyweight delegation. I created https://issues.serokell.io/issue/CSL-1712 to take care of it.
DevOps should pass secret keys to which bootstrap stakeholders delegated heavily.

gromakovsky: Btw, @georgeee, do DevOps have access to all secret keys? If not, we probably need to provide better tooling.

docs/how-to/rollback-blocks.md
```diff
@@ -0,0 +1,27 @@
...
+```
+$ stack exec -- cardano-auxx                      \
+    --system-start 0                              \
+    --configuration-file node/configuration.yaml  \
+    --configuration-key mainnet_full              \
+    --log-config log-config-prod.yaml             \
+    --logs-prefix "logs/mainnet-1.0"              \
+    --db-path db-mainnet-1.0                      \
+    cmd --commands="rollback 5 /tmp/cardano-sl-rollback-txs-dump"
+
```
gromakovsky: Hm, where is ``` (I might be blind)?

denisshevchenko: Sorry, fixed.

PR #1663, opened 2017-09-27
Fix space leak in Pos.Delegation.Worker

arybczak:
Details in a code comment.


adinapoli-iohk:
Oh wow, TIL indeed, great detective work, @arybczak !

gromakovsky:
It's already merged to 1.0 and master.

PR #1669, opened 2017-09-27
CSL-1710 Test block-gen and rollback tools

gromakovsky:
The task was to test block-gen and rollback tools.
Results can be found in comments to CSL-1710.
During the testing I found some bad things and fixed them. The most important fix is topological sort in `send-from-file` command.
It's probably better to review this PR per commit, all of them are rather small and self-contained.

### DIFF
--- a/node/src/Pos/Worker.hs
+++ b/node/src/Pos/Worker.hs
@@ -14,7 +14,7 @@ import           Data.Tagged             (untag)
 import           Pos.Block.Worker        (blkWorkers)
 import           Pos.Communication       (OutSpecs, Relay, WorkerSpec, localWorker,
                                           relayPropagateOut, wrapActionSpec)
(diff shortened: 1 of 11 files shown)


node/src/Pos/Worker.hs
```diff
@@ -80,6 +80,6 @@ allWorkers NodeResources {..} = mconcatPair
   where
     NodeContext {..} = nrContext
     properSlottingWorkers =
-       fst (localWorker (recoveryCommGuard "logNewSlot" logNewSlotWorker)) :
+       fst (localWorker logNewSlotWorker) :
```
volhovM: Explain this change please.

gromakovsky: I think explanation in the commit message is good enoguh: https://github.com/input-output-hk/cardano-sl/pull/1669/commits/7a9041ac91cbaf8373ab39bbe546a15ed6af5aae

PR #1671, opened 2017-09-27
[CSM-475] Sort transaction history to remove on tracked rollback

Martoon-00:
On block applying txs are sorted before being added to tx history.
On rollback we expect transactions to be removed from history in opposite
order, so transactions of rollbacked block should be sorted as well to
allow us properly drain them from the top of current history.


Martoon-00:
Will rebase to master

Martoon-00:
Was tested with many transactions and artificially made rollback, already merged to master

PR #1676, opened 2017-09-28
[CSL-1713] time-warp ack timeout configurable

avieth:
Optional `conversationEstablishTimeout` key in the node section of configuration. Default is 30 seconds. It's in milliseconds like the network connection timeout.

Really this belongs in the infra configuration, and so does the network connection timeout, but I kept the two in the same area rather than doing that refactor.

### DIFF
--- a/node/src/Pos/Configuration.hs
+++ b/node/src/Pos/Configuration.hs
@@ -71,7 +71,7 @@ data NodeConfiguration = NodeConfiguration
       -- support caching
     , ccNetworkConnectionTimeout     :: !Int
       -- ^ Network connection timeout in milliseconds
-    , ccConversationEstablishTimeout :: !(Maybe Int)
+    , ccConversationEstablishTimeout :: !Int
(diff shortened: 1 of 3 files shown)


node/src/Pos/Configuration.hs
```diff
@@ -97,6 +101,11 @@ networkDiameter = sec . ccNetworkDiameter $ nodeConfiguration
 networkConnectionTimeout :: HasNodeConfiguration => Microsecond
 networkConnectionTimeout = ms . fromIntegral . ccNetworkConnectionTimeout $ nodeConfiguration
 
+-- | Default is 30 seconds.
+conversationEstablishTimeout :: HasNodeConfiguration => Microsecond
```
volhovM: Oh, no, please, let's specify this explicitly in the configuration. Looking into the code is ~2 times psychologically harder than looking into config.yaml.

avieth: Fine by me.

PR #1678, opened 2017-09-28
Remove `instance Default Salt`

Martoon-00:


core/Pos/Crypto/Scrypt.hs
```diff
@@ -49,8 +50,8 @@ mkSalt :: Bi salt => salt -> S.Salt
 mkSalt = S.Salt . serialize'
 
 -- | Salt which can be used for hardcoded values.
-instance Default S.Salt where
-    def = mkSalt ("" :: Text)
+defSaltUnsafe :: S.Salt
```
volhovM: call it emptysalt pls

Martoon-00:
Makes sense

PR #1683, opened 2017-09-29
[CSM-477] Hide wallet/account names from logs

Martoon-00:
... till the day we allow unicode symbols in logging library.

I assume that there is no need to hide redeeming seed because if user passes invalid key with unicode chars endpoint will just throw a strange error


Martoon-00:
@dcoutts Our logging library throws error when non-ascii symbols are passed.
Something like `commitAndReleaseBuffer: invalid argument (invalid character)`.

We are going to resolve this in nice way, but this reproduces only on linux and mac so we will do it when the dust settles

PR #1686, opened 2017-09-30
[CSL-1720] Don't call BListener during lrc

pva701:


wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -503,19 +503,19 @@ getEncInfo encSK = do
...
+selectOwnAddresses encInfo getAddr =
+    mapMaybe (\a -> (a,) <$> decryptAddress encInfo (getAddr a))
 
 setLogger :: HasLoggerName m => m a -> m a
 setLogger = modifyLoggerName (<> "wallet" <> "sync")
 
-decryptAccount :: (HDPassphrase, CId Wal) -> Address -> Maybe CWAddressMeta
-decryptAccount (hdPass, wCId) addr = do
+decryptAddress :: (HDPassphrase, CId Wal) -> Address -> Maybe CWAddressMeta
+decryptAddress (hdPass, wCId) addr = do
```
gromakovsky: I don't get why these changes are in this commit.

pva701: hz, I noticed and decided to fix it

pva701:
@gromakovsky fixed tests
@neongreen review plz (at least this commit https://github.com/input-output-hk/cardano-sl/pull/1686/commits/ce1e05ccaa4bed17588033de4b1a0cd17e9b678b)

lib/src/Pos/Util/Config.hs
```diff
@@ -2,6 +2,7 @@
 
 module Pos.Util.Config
        ( embedYamlConfigCT
+       , embedYamlConfig
```
gromakovsky: What's the difference between `embedYamlConfigCT` and `embedYamlConfig`? Do we need both?

lib/src/Pos/Block/Logic/Internal.hs
```diff
@@ -153,8 +153,11 @@ normalizeMempool = do
 -- Invariant: all blocks have the same epoch.
 applyBlocksUnsafe
     :: forall ssc ctx m . MonadBlockApply ssc ctx m
-    => OldestFirst NE (Blund ssc) -> Maybe PollModifier -> m ()
-applyBlocksUnsafe blunds pModifier = do
+    => ShouldCallBListener
+    -> OldestFirst NE (Blund ssc)
+    -> Maybe PollModifier
+    -> m ()
+applyBlocksUnsafe bsc blunds pModifier = do
```
gromakovsky: bsc? Maybe `scb`?

lib/src/Pos/Util/Config.hs
pva701: I left `embededYamlConfigCT` just in case, it may be needed in future

pva701:
@gromakovsky approve?

pva701:
approve najmi

auxx/src/Command/Rollback.hs
```diff
@@ -57,7 +58,7 @@ rollbackAndDump numToRollback outFile = do
             liftIO $ BSL.writeFile outFile (serialize allTxs)
             logInfo $ sformat ("Dumped "%int%" transactions to "%string)
                       (length allTxs) (outFile)
-            rollbackBlocksUnsafe (BypassSecurityCheck True) blunds
+            rollbackBlocksUnsafe (BypassSecurityCheck True) (ShouldCallBListener False) blunds
```
gromakovsky: ???

pva701: chto ti hochech, zachem vizivat ego tut?

gromakovsky: What's the reason not to call it?

gromakovsky: Imo, the only reason not to call BListener is when we are sure we will apply the same blocks back. But here it's not the case.

pva701: ok, fixed. no hz zachem blistener should know about rollback in auxx

PR #1687, opened 2017-09-30
[CSE-241] Allow not resetting DB for getUtxoSum

Pastafarianist:


node/src/Pos/Explorer/DB.hs
```diff
@@ -98,19 +98,22 @@ getLastTransactions = gsGetBi lastTxsPrefix
...
+prepareExplorerDB = do
+    unlessM balancesInitializedM $ do
         let GenesisUtxo utxo = genesisUtxo
             addressCoinPairs = utxoToAddressCoinPairs utxo
         putGenesisBalances addressCoinPairs
-        putGenesisUtxoSum addressCoinPairs
         putInitFlag
+    unlessM utxoSumInitializedM $ do
+        -- Smooth migration for CSE-228.
+        unlessM utxoSumInitializedM $ do
```
gromakovsky: This line look redundant.

Pastafarianist: It is, thank you for spotting it.

PR #1695, opened 2017-10-01
[CSE-236] Search by time is not working

ksaric:
https://issues.serokell.io/issue/CSE-236

## About the migration changes

Before we had mappings Epoch -> [HeaderHash].
Now we have mappings (Epoch, Page) -> [HeaderHash].

The old key is "e/epoch/" <> serialize' epoch.
The new key is "e/epochs/" <> serialize' epoch <> "/" <> (serialize' $ UnsignedVarInt page).

We can summarize this migration as a function - (Epoch, [HeaderHash]) -> ((Epoch,Page), [HeaderHash]) where we switch from 21600 HeaderHashes we had per Epoch to 10 per page, per epoch.

We use Conduit since we don't want to load everything into memory at once.


ksaric:
So now we have working epoch paging and we get the epoch page number.

Tested on _mainnet_ works fine.
Missing just the frontend changes.

ksaric:
Tested on _mainnet_ with the frontend changes, all seems good.
Fast & nice.

node/src/Pos/Explorer/DB.hs
```diff
@@ -147,7 +151,8 @@ data ExplorerOp
 
     | PutPageBlocks !Page ![HeaderHash]
 
-    | PutEpochBlocks !Epoch ![HeaderHash]
+    | PutEpochBlocks !Epoch !Page ![HeaderHash]
```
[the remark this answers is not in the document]
ksaric: Yes, the database should be deleted and explorer restarted. 
The frontend is involved also, and there is no "middle ground" between migrations.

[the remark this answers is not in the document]
akegalj: > The frontend is involved also, and there is no "middle ground" between these migrations.

can you elaborate? (didn't get the point)

> Yes, the database should be deleted and explorer restarted.

I am ok for restarting the DB. Do we have to remove both nodes and Explorers DB or is removing Explorers DB sufficient ? Note that we won't be able to get away with removing DB all the time as it will increase sync time (thus increasing "downtime" )

At this moment I am not sure how Explorer will behave when its not fully synced:
 * how will main page behave ? (will it show older info or smth else)
 * how will address detail or transaction detail page behave ? (probably as address or tx can't be found)

@ksaric can you provide these details ^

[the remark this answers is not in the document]
ksaric: @akegalj I mean, the current DB format and the previous DB format are incompatible and the new format is bounded to the frontend directly - we have paging, we didn't have that before. 
This isn't just a simple DB migration, the web API has changed to reflect the changes.

We need to remove the explorer DB only. Until it's synced it will display the data that has at the moment - if it's syncing and currently has 200/2160 pages, it will display those 200 pages and all that they contain - address, transactions...

We could simply run this locally, connect explorer to _mainnet_, and then when it syncs, copy the DB to the "official" _mainnet_ server. It would need to sync only the blocks that have been created while we were transferring the DB from local to global.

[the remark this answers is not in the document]
akegalj: > This isn't just a simple DB migration, the web API has changed to reflect the changes.

I know web api has changed. Still, we will for sure modify Explorer DB scheme in the future. If its too complicated to migrate it now (because we still don't know how to properly migrate rocsdb) then we can remove the db and see how fast it will take for it to be up to date.

Can you verify this locally - it would be useful to know:
 1) how much time would it take to sync if we remove both node and explorer db
 2) how much time would it take to sync if we remove only explorer db (this would also ensure that removing only explorer db works as expected - as I never tried it myself)

can you do these tests? ^ 

@Pastafarianist would also like to hear your opinion about DB change, as you already did similar safe transition.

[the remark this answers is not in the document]
ksaric: 1. I don't see any reason to try that and explorer DB is the node DB, since explorer is a node++.
2. Around the same speed as node syncing.

I don't why these results would be interesting and why we need to spend time on them, but I'll test it if it's necessary.
Again, we could find that out when we sync the DB before replacing it, if that's the direction we are going in.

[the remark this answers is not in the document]
akegalj: > I don't why these results would be interesting and why we need to spend time on them, but I'll test it if it's necessary.

If they differ a lot, then it might be beneficial to do trick with syncing and moving the db -> which complicates things for devops_

> Again, we could find that out when we sync the DB before replacing it, if that's the direction we are going in.

as said, this complicates life of devops and before doing so it would be useful to know is it really needed. Currently blockchain is not that big and we might sacrifice a minute of downtime in order not to change the deployment (if it takes a minute - which we would need to measure)

> I don't see any reason to try that and explorer DB is the node DB, since explorer is a node++.

this implies that explorer db is the same one as nodes db ? (have to check it in code)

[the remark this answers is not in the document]
ksaric: > this implies that explorer db is the same one as nodes db ? (have to check it in code)

AFAIK explorer DB should be node DB + explorer specific part(s).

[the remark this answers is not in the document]
akegalj: @ksaric can't we extend `prepareExplorerDB` in a way that adds new `Page` to the DB ? We can iterate over epoch and calculate this `Page` and fix the DB ?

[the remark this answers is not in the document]
ksaric: @akegalj And save it where? Notice that the prefix/key is the same, just a `Page` was added - https://github.com/input-output-hk/cardano-sl/pull/1695/files#diff-9ea213700be61c330d86a770f61d6d03L261
I'm not saying it's not possible, but would be quite difficult - for example, do we fit all the current epoch blocks in memory, if we go down that route? 
I _could_ change the key/prefix, start iterating using conduit and place that in the database. 
But then we would have a check that checks that once, not a DB migration. 
It's likely to take some time, and the end result? A piece of code that's used _once_.

Pastafarianist:
@ksaric, even before I review the changes I have to say that we _must_ write a migration, because currently the node is compiled with Explorer by default — and hence all our users have a copy of Explorer DB, which will be invalidated by your changes and most likely render their installations broken.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -256,12 +262,13 @@ getBlocksPage mPageNumber pageSize = do
...
     blunds          <- forM pageBlocksHH getBlundOrThrow
-    cBlocksEntry    <- forM (rights' blunds) toBlockEntry
+    cBlocksEntry    <- forM (blundToMainBlockUndo blunds) toBlockEntry
 
     -- Return total pages and the blocks. We start from page 1.
     pure (totalPages, reverse cBlocksEntry)
   where
-    rights' x = [(mb, u) | (Right mb, u) <- x]
+    blundToMainBlockUndo :: [Blund SscGodTossing] -> [(MainBlock SscGodTossing, Undo)]
+    blundToMainBlockUndo blund = [(mainBlock, undo) | (Right mainBlock, undo) <- blund]
```
Pastafarianist: Irrelevant change. If you collect many of them, please split them off to a separate PR. A few is fine. Please rely on common sense while deciding on the borderline between "few" and "many".

ksaric:
@Pastafarianist I didn't expect that. But wait, don't we run `cardano-node-simply` only? Ah, yes, the format is incompatible any way you put it.
@akegalj I will start with this ASAP :disappointed:

ksaric:
Tested 4 times. 
CPU/memory usage seems fine.

There is a (small) issue which I don't consider critical. 
The first page of the 1. epoch has 9 instead of 10 elements. 
I'm guessing this is somehow connected to `splitEvery`, but not sure why. I presume it's because the number of blocks is uneven and it's being reflected on the first page. I think it's not ordering.

It should work fine. For example:
```
Prelude> filter ((/= 10) . length) $ splitEvery 10 [0..21599]
[]

Prelude> filter ((/= 10) . length) $ splitEvery 10 [0..21598]
[[21590,21591,21592,21593,21594,21595,21596,21597,21598]]
```

I tested it using this wonderful new recipe:
- checkout `cardano-sl-1.0`
- build it using stack
- run on *mainnet*
- let it sit for a while, say ~15 minutes (that's how much it takes to sync the *mainnet* blockchain nowadays)
- add caffeine (optional)
- go to http://localhost:8100/api/search/epoch/0?slot=1
- checkout `ksaric/CSE-236-mainnet`
- build it using stack
- run on *mainnet*
- hold your breath and cross your fingers (wearing your lucky clothes)
- go to http://localhost:8100/api/epochs/0
- admire

The whole recipe step-by-step:
```
git checkout cardano-sl-1.0
stack build --fast -j4
./scripts/clean/db.sh
./scripts/launch/connect-to-cluster/explorer-mainnet.sh
```

Then you wait and:
```
git checkout ksaric/CSE-236-mainnet
stack build --fast -j4
./scripts/clean/db.sh
./scripts/launch/connect-to-cluster/explorer-mainnet.sh
```

node/src/Pos/Explorer/DB.hs
```diff
@@ -104,10 +133,116 @@ prepareExplorerDB = do
...
+        epochPagesConduit = CL.map convertToPagedMap
+          where
+            convertToPagedMap
+                :: (Epoch, [HeaderHash])
+                -> Map EpochPagedBlocksKey [HeaderHash]
+            convertToPagedMap ehh = fromListWith (++) (convertToPaged ehh)
+
+            convertToPaged :: (Epoch, [HeaderHash]) -> [((Epoch, Page), [HeaderHash])]
+            convertToPaged (epoch, headerHashes) =
+                convertHHsToEpochPages <$> convertHHsToPages
```
Pastafarianist: Please replace `<$>` with `map` here.

node/src/Pos/Explorer/DB.hs
```diff
@@ -104,10 +133,116 @@ prepareExplorerDB = do
...
+        -- load - performance reasons.
+        epochPagesConduit
+            :: (MonadDBRead m)
+            => Conduit (Epoch, [HeaderHash]) m (Map EpochPagedBlocksKey [HeaderHash])
+        epochPagesConduit = CL.map convertToPagedMap
+          where
+            convertToPagedMap
+                :: (Epoch, [HeaderHash])
+                -> Map EpochPagedBlocksKey [HeaderHash]
+            convertToPagedMap ehh = fromListWith (++) (convertToPaged ehh)
```
Pastafarianist: Am I understanding correctly that `(Epoch, Page)` pairs must be unique in `convertToPaged ehh`?

node/src/Pos/Explorer/DB.hs
```diff
@@ -104,10 +133,116 @@ prepareExplorerDB = do
...
+              where
+                -- | Persist @Epoch@ @Page@ blocks.
+                persistEpochPageBlocks :: [ExplorerOp]
+                persistEpochPageBlocks = putKeyBlocks <$> M.toList mapEpochPagedHHs
+                  where
+                    putKeyBlocks
+                        :: (EpochPagedBlocksKey, [HeaderHash])
+                        -> ExplorerOp
+                    putKeyBlocks keyBlocks = PutEpochBlocks epoch page blocks
+                      where
```
Pastafarianist: I believe having 4-5 nested `where` clauses makes the code much less readable. Please avoid deep nesting.

node/src/Pos/Explorer/DB.hs
```diff
@@ -104,10 +133,116 @@ prepareExplorerDB = do
...
+                persistMaxPageNumbers :: [ExplorerOp]
+                persistMaxPageNumbers =
+                    [ PutEpochPages epoch maxPageNumber | (epoch, maxPageNumber) <- emp]
+                  where
+                    emp :: [EpochPagedBlocksKey]
+                    emp = findEpochMaxPages mapEpochPagedHHs
+
+        -- | Finally, just persist all the operations atomically.
+        persistExplorerOpSink :: (MonadDB m) => Sink ([ExplorerOp]) m ()
+        persistExplorerOpSink = CL.mapM_ writeBatchGState
```
Pastafarianist: This is not atomic. Essentially, you are persisting each of the `Op`s atomically, but not all of them at once.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -588,36 +595,33 @@ getGenesisPagesTotal (fromIntegral -> pageSize) = do
...
     -> m [CBlockEntry]
 epochSlotSearch epochIndex slotIndex = do
 
-    -- [CSE-236] Disable search for epoch only
-    -- TODO: Remove restriction if epoch search will be optimized
-    when (isNothing slotIndex) $
-        throwM $ Internal "We currently do not support searching for epochs only."
-
+    -- The slots start from 0 so we need to modify the calculation of the index.
+    let page = fromIntegral $ (slotIndex + 10) `div` 10
```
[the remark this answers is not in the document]
ksaric: Yes.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -588,36 +595,33 @@ getGenesisPagesTotal (fromIntegral -> pageSize) = do
...
     -- Get pages from the database
     -- TODO: Fix this Int / Integer thing once we merge repositories
```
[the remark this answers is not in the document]
ksaric: It's my comment, left there to remind me to simplify things when I find the time.

node/src/Pos/Explorer/DB.hs
ksaric: `convertToPagedMap` will return a `Map` with `(Epoch, Page)` unique, if that's what you mean.

node/src/Pos/Explorer/DB.hs
ksaric: Why? It's not like this is the only piece of code that has the more general version of the functor.
If it's because of readability, I can change it, but since the functions are (type) annotated, I think it's pretty readable.

node/src/Pos/Explorer/DB.hs
ksaric: Deep nesting is indicating local calls. I prefer this than to pollute the namespace - this is where polluting the namespace is actually a consequence of removing/flattening the clauses.

node/src/Pos/Explorer/DB.hs
ksaric: Isn't `Batch` used for "all or nothing" situations? I'll take a look at this in more detail, thanks.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -633,13 +637,69 @@ epochSlotSearch epochIndex slotIndex = do
...
+epochPageSearch
+    :: ExplorerMode ctx m
+    => EpochIndex
+    -> Maybe Int
+    -> m (Int, [CBlockEntry])
+epochPageSearch epochIndex mPage = do
+
+    -- If the user doesn't define the page, the default page is the first one.
+    let page :: Int
+        page = fromMaybe 1 mPage
```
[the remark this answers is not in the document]
ksaric: Fair enough, will move.

explorer/frontend/src/Explorer/Update.purs
```diff
@@ -633,23 +634,47 @@ update (ReceiveBlockSummary (Left error)) state =
 
 -- Epoch, slot
 
-update (RequestSearchBlocks epoch slot) state =
+update (RequestEpochPageSearch epochIndex pNumber) state =
+    { state:
+          set loading true $
+          -- Important note: Don't use `set currentBlocksResult Loading` here,
+          -- because we will lost all previous data of `currentBlocksResult` while pagenating
```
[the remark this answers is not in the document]
ksaric: Fixed.

explorer/frontend/src/Explorer/Update.purs
```diff
@@ -633,23 +634,47 @@ update (ReceiveBlockSummary (Left error)) state =
...
+          -- We are updating `blsViewLoadingPagination` instead this ^
```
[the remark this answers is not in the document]
ksaric: Fixed.

explorer/src/Pos/Explorer/Web/Api.hs
```diff
@@ -86,11 +87,16 @@ type AddressSummary = API
     :> Capture "address" CAddress
     :> Get '[JSON] (Either ExplorerError CAddressSummary)
 
+type EpochPageSearch = API
```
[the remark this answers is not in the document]
ksaric: Fixed.

node/src/Pos/Explorer/DB.hs
ksaric: It's atomic.

ksaric: Ah, I think I see what you mean. Will fix.

explorer/src/Pos/Explorer/Web/Server.hs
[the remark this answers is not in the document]
Pastafarianist: @ksaric please open an issue.

node/src/Pos/Explorer/DB.hs
Pastafarianist: I mean that the result of `convertToPaged ehh` should contain unique keys, and `fromListWith` actually never performs concatenation. Is that correct?

node/src/Pos/Explorer/DB.hs
Pastafarianist: @ksaric generalization is good when you intend to use it. In this case it's unneeded. Furthermore, `map` reminds the reader of the type of its arguments instantly, unlike `<$>`. Please don't put the extra strain of keeping track of type information on the reader and use `map`.

node/src/Pos/Explorer/DB.hs
```diff
@@ -104,10 +133,115 @@ prepareExplorerDB = do
...
+        epochPagesExplorerOpSink = CL.mapM_ persistEpochBlocks
+          where
+            -- | Persist atomically, all the operations together.
+            persistEpochBlocks
+                :: (MonadDB m)
+                => Map EpochPagedBlocksKey [HeaderHash]
+                -> m ()
+            persistEpochBlocks mapEpochPagedHHs =
+                -- Here we persist all the changes atomically.
+                writeBatchGState $ persistEpochPageBlocks ++ persistMaxPageNumbers
```
Pastafarianist: As far as I understand, you still perform all batch writes separately. However, apparently this cannot be easily fixed, unless we want to load the whole list of operations into memory. We don't support transactions in RocksDB (yet), so we can't open a transaction, and concatenating all operations to produce one huge list would 1) probably occupy a ton of memory and 2) would render Conduit unneeded. I'd recommend keeping the code as-is, but we'd probably be better off hearing some advice from @pva701 or @gromakovsky.

node/src/Pos/Explorer/DB.hs
Pastafarianist: For now, keep it as-is, but we should decide on whether this is acceptable style or not (a quick survey showed varying results) and stick to the style guide.

node/src/Pos/Explorer/DB.hs
ksaric: This is correct, will fix.

ksaric:
To be squashed, since it contains a lot of irrelevant commits.

node/src/Pos/Explorer/DB.hs
ksaric: We have atomic persistence of an *Epoch* but not the persistence of multiple *Epoch*s.
As you said yourself, we would occupy too much memory if we drop conduit, and these key-value data stores aren't that flexible (in my experience) as relational are.
If we had a relational (say Oracle) we could post it on the database (in batches) and in the end commit all of it atomically.

I have an idea how to do it in RocksDB, but it's a hack and I don't want to do it that way.
I'll write about it here so if someone wants to be a hero, he can.
I'll write about the specific case, generalize away.
In this example, we have a (new) key `"e/epochs/" <> serialize' epoch <> "/" <> encodedPage`.

We could temporarily create a new key, say `"e/posting/epochs/" <> serialize' epoch <> "/" <> encodedPage` for the persistence of the new epochs using conduit.
We then add another one, say `"e/postingStatus/epochs/" <> serialize' epoch` with a simple `Bool` that designates whether all the keys are persisted (we could simplify this, but nevermind).
In the end, we simply check whether all the keys are `True` and if yes, we either _change their keys_ or save them under the original key again, delete the temp keys we created and voila - we have a very crude hackish atomic transaction.

Again, this seems (a bit) insane and I would rather leave it as-is.

explorer/src/Pos/Explorer/Web/Server.hs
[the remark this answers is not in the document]
Pastafarianist: Can you rewrite this as `(slotIndex `div` 10) + 1` then?

node/src/Pos/Explorer/DB.hs
Pastafarianist: Whether a DB is relational or not is independent of whether it supports transactions. In fact, RocksDB [_does_ support transactions](https://github.com/facebook/rocksdb/wiki/Transactions), but AFAIU we need to use a different way of initializing the DB (called TransactionDB in RocksDB lingo) to make use of that.

explorer/src/Pos/Explorer/Web/Server.hs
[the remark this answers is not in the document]
ksaric: Done.

```
all (\(t1,t2) -> t1 == t2) $ zip [(slotIndex,((slotIndex `div` 10) + 1)) | slotIndex <- [0..23]] [(slotIndex,((slotIndex + 10) `div` 10)) | slotIndex <- [0..23]]
True
```

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -633,13 +638,65 @@ epochSlotSearch epochIndex slotIndex = do
...
+    => EpochIndex
+    -> Int
+    -> m (Int, [CBlockEntry])
+epochPageSearch epochIndex page = do
+
+    -- We want to fetch as many pages as we have in this @Epoch@.
+    epochPagesNumber <- getEpochPages epochIndex >>= maybeThrow (Internal "No epoch pages.")
+
+    -- Get pages from the database
+    -- TODO: Fix this Int / Integer thing once we merge repositories
```
Pastafarianist: The repositories of `cardano-sl` and `cardano-sl-explorer` have long been merged. Please open an issue, assign it to yourself and update this comment with the issue number. The reasoning is that currently you are the only one who has an idea of what the problem is about. For example, I don't even understand from this phrase where the `Int` / `Integer` problem arises, let alone how to fix it.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -633,13 +638,65 @@ epochSlotSearch epochIndex slotIndex = do
...
+        errMsg = sformat ("No blocks on epoch "%build%" page "%build%" found!") epoch page'
+
+    -- | Sorting.
+    sortBlocksByEpochSlots
+        :: [(Block SscGodTossing, Undo)]
+        -> [(Block SscGodTossing, Undo)]
+    sortBlocksByEpochSlots blocks = sortBy (comparing $ Down . getBlockIndex . fst) blocks
+      where
+        -- | Get the block index number. We start with the the index 1 for the
+        -- gensis block and add 1 for the main blocks since they start with 1
```
Pastafarianist: * genesis

explorer/src/Pos/Explorer/Web/Server.hs
ksaric: The issue has been opened - https://issues.serokell.io/issue/CSE-249

explorer/src/Pos/Explorer/Web/Server.hs
ksaric: Fixed.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -185,9 +191,10 @@ explorerHandlers _sendActions =
     getStatsTxsDefault page =
         catchExplorerError $ getStatsTxs page
 
-    defaultPageSize size = (fromIntegral $ fromMaybe 10 size)
-    defaultLimit limit   = (fromIntegral $ fromMaybe 10 limit)
-    defaultSkip  skip    = (fromIntegral $ fromMaybe 0  skip)
+    defaultPageSize size  = (fromIntegral $ fromMaybe 10 size)
+    defaultLimit limit    = (fromIntegral $ fromMaybe 10 limit)
+    defaultSkip skip      = (fromIntegral $ fromMaybe 0  skip)
+    defaultPage page      = (fromIntegral $ fromMaybe 1  page)
```
akegalj: Minor. Looks like a good candidate for Def typeclass.

ksaric: Yes, this was refactored and removed by @sectore in the `master` branch.

ksaric:
Before we had mappings `Epoch -> [HeaderHash]`.
Now we have mappings `(Epoch, Page) -> [HeaderHash]`.

The old key is `"e/epoch/" <> serialize' epoch`.
The new key is `"e/epochs/" <> serialize' epoch <> "/" <> (serialize' $ UnsignedVarInt page)`.

We can summarize this migration as a function - `(Epoch, [HeaderHash]) -> ((Epoch,Page), [HeaderHash])` where we switch from 21600 `HeaderHash`es we had to 10 per page, per epoch.

We use `Conduit` since we don't want to load everything into memory at once.

node/src/Pos/Explorer/DB.hs
```diff
@@ -104,10 +167,86 @@ prepareExplorerDB = do
...
         putGenesisBalances addressCoinPairs
         putInitFlag
+
     -- Smooth migration for CSE-228.
     unlessM utxoSumInitializedM $ do
         putCurrentUtxoSum
 
+    -- Smooth migration for CSE-236.
+    unlessM blockEpochPagesInitializedM $ do
+        convertOldEpochBlocksFormat
```
dcoutts: Is this really the best place to put the migration? This is in a query. What is to stop this happening multiple times concurrently? Surely it is better to do the migration as part of server startup and opening the DB and before starting the web server listening thread.

From the other comments here, I am not clear about whether this migration is enough or if devops need to do anything with clearing the DB.

node/src/Pos/Explorer/DB.hs
```diff
@@ -257,8 +414,18 @@ blockPagePrefix page = "e/page/" <> encodedPage
   where
     encodedPage = serialize' $ UnsignedVarInt page
 
-blockEpochPrefix :: Epoch -> ByteString
-blockEpochPrefix epoch = "e/epoch/" <> serialize' epoch
+-- | TODO(ks): To remove in the next version.
+previousEpochBlocksPrefix :: Epoch -> ByteString
+previousEpochBlocksPrefix epoch = "e/epoch/" <> serialize' epoch
```
dcoutts: It'd be better to call this "old" rather than "previous". Previous makes one thing of the epoch before this one, rather than an older version of the DB key.

dcoutts:
Obviously we'll also need to fix up the conflicts and make the CI pass too.

node/src/Pos/Explorer/DB.hs
Pastafarianist: @dcoutts this is not in a query. This code is within `prepareExplorerDB` which is executed at server startup

dcoutts: Sorry, I was mislead by the github diff, it looked like it was inside `getLastTransactions`. Oops! Ok, never mind about that complaint then.

ksaric: AFAIK they don't have to do anything, the migration will run only on those machines that have the "old" DB epoch format when preparing explorer database.

node/src/Pos/Explorer/DB.hs
ksaric: Ok, will fix.

ksaric:
Fixed merge conflicts. 
I run it afterward on the `mainnet`. Waited until it was synced and tested using the _explorer_ frontend, which seems good to me.

georgeee:
From @fersel:
Not yet to be merged (we have staging but not mainnet deployed, during this "window" we can't merge anything to release branch)

PR #1696, opened 2017-10-01
[CSE-240] Optimize endpoints which need slot leaders

Pastafarianist:
The point of this PR is to allow fetching slot leaders one by one instead of all at once for an epoch. According to investigation by @ksaric, this should speed up some endpoints which need slot leaders, primarily `/api/search/epoch/{epoch}`.

@ksaric points out CPU profiling of `getLeaders` here https://files.slack.com/files-pri/T09C3TRRD-F73LH8F5F/cpu2x10.png which shows high consumption of this endpoint. In DEV mode where `getLeaders` is pretty small there are no CPU/memory spikes - this gets worse with blockchain growth (when getLeaders increases). More info about getLeaders here https://issues.serokell.io/issue/CSE-219 .


lib/src/Pos/Lrc/DB/Leaders.hs
```diff
@@ -17,23 +19,39 @@ import           Universum
...
 
-putLeaders :: MonadDB m => EpochIndex -> SlotLeaders -> m ()
-putLeaders epoch = putBi (leadersKey epoch)
+-- | Put leaders for all slots in the specified epoch in the DB.
+-- The DB contains two mappings:
+-- * EpochIndex -> SlotLeaders
+-- * SlotId -> StakeholderId (added in CSE-240)
+putLeadersForEpoch :: MonadDB m => EpochIndex -> SlotLeaders -> m ()
+putLeadersForEpoch epoch leaders = do
+    putLeadersForEpochAllAtOnce epoch leaders
```
volhovM: What happens if I power off pc between these two lines?

ksaric: Good catch @volhovM. 
You can pull up those helpers below into the local function scope using _where_, change them to return the DB operation and perform `BatchOp` on the list containing both of their operations. Or `putBatchBi`.

lib/src/Pos/Lrc/DB/Leaders.hs
```diff
@@ -44,13 +62,49 @@ prepareLrcLeaders ::
...
+
+putLeadersForEpochAllAtOnce :: MonadDB m => EpochIndex -> SlotLeaders -> m ()
+putLeadersForEpochAllAtOnce epoch = putBi (leadersForEpochKey epoch)
+
+putLeadersForEpochSeparately :: MonadDB m => EpochIndex -> SlotLeaders -> m ()
+putLeadersForEpochSeparately epoch leaders =
+    putBatchBi putLeadersForEachSlotOps
+  where
+    putLeadersForEachSlotOps :: [(ByteString, StakeholderId)]
+    putLeadersForEachSlotOps = [
```
ksaric: :nauseated_face: 
Can you break this up a bit? Optional, but I would find it more readable.

lrc/Pos/Lrc/DB/Common.hs
```diff
@@ -42,6 +47,14 @@ putBi
     => ByteString -> v -> m ()
 putBi = dbPutBi LrcDB
 
+putBatchBi
```
ksaric: This seems like a good idea.

gromakovsky:
Don't we need it in 1.0?

lib/src/Pos/Lrc/Worker.hs
```diff
@@ -196,10 +198,10 @@ issuersComputationDo epochId = do
 
 leadersComputationDo :: LrcMode ssc ctx m => EpochIndex -> SharedSeed -> m ()
 leadersComputationDo epochId seed =
-    unlessM (isJust <$> getLeaders epochId) $ do
+    unlessM (isJust <$> LrcDB.getLeadersForEpoch epochId) $ do
```
gromakovsky: Here and above it makes more sense to use `getLeader (SlotId epochId minBound)`. It should be `Just` iff `LrcDB.getLeadersForEpoch epochId` is `Just`.

lib/src/Pos/Lrc/DB/Leaders.hs
gromakovsky: Actually nothing bad will happen, but I would prefer doing both lines atomically.

lib/src/Pos/Lrc/DB/Leaders.hs
```diff
@@ -44,13 +62,49 @@ prepareLrcLeaders ::
        , HasConfiguration
        )
     => m ()
-prepareLrcLeaders =
-    whenNothingM_ (getLeaders 0) $
-        putLeaders 0 genesisLeaders
+prepareLrcLeaders = do
+    whenNothingM_ (getLeadersForEpoch 0) $
```
gromakovsky: TBH, I would rework LRC DB initialization the same way as it's done for GState DB (i. e. have initialization flag), but it can be done later.

lib/src/Pos/Lrc/DB/Leaders.hs
```diff
@@ -44,13 +62,49 @@ prepareLrcLeaders ::
...
+        putLeadersForEpoch 0 genesisLeaders
+    -- Smooth migration for CSE-240.
+    initLeaders 1
+  where
+    initLeaders :: MonadDB m => EpochIndex -> m ()
+    initLeaders i = do
+        whenNothingM_ (getLeader $ SlotId i (UnsafeLocalSlotIndex 0)) $ do
```
gromakovsky: You can use `minBound` instead of `UnsafeLocalSlotIndex 0` to avoid `Unsafe` word :)

lib/src/Pos/Lrc/DB/Leaders.hs
```diff
@@ -44,13 +62,49 @@ prepareLrcLeaders ::
...
+  where
+    putLeadersForEachSlotOps :: [(ByteString, StakeholderId)]
+    putLeadersForEachSlotOps = [
+            (leaderKey $ mkSlotId epoch i, leader)
+            | (i, leader) <- zip [1..] $ toList leaders
+        ]
+    mkSlotId :: EpochIndex -> Word16 -> SlotId
+    mkSlotId epoch' slot =
+        -- Using @UnsafeLocalSlotIndex@ because we trust the callers.
+        SlotId epoch' (UnsafeLocalSlotIndex slot)
```
gromakovsky: There is `unsafeMkLocalSlotIndex` function which is a bit better because it will `error` if `slot` is out of bounds.

lib/src/Pos/Lrc/DB/Leaders.hs
```diff
@@ -44,13 +62,49 @@ prepareLrcLeaders ::
...
+putLeadersForEpochAllAtOnce epoch = putBi (leadersForEpochKey epoch)
+
+putLeadersForEpochSeparately :: MonadDB m => EpochIndex -> SlotLeaders -> m ()
+putLeadersForEpochSeparately epoch leaders =
+    putBatchBi putLeadersForEachSlotOps
+  where
+    putLeadersForEachSlotOps :: [(ByteString, StakeholderId)]
+    putLeadersForEachSlotOps = [
+            (leaderKey $ mkSlotId epoch i, leader)
+            | (i, leader) <- zip [1..] $ toList leaders
```
gromakovsky: Why do you enumerate slots from 1?

ksaric:
It should be in `cardano-sl-1.0`. 
I thought that we still have the same workflow - PR to `master`, PR to `cardano-sl-1.0`, but I'm honestly not sure anymore :man_shrugging:

gromakovsky:
PR to master should be made after PR to release branch is merged.

akegalj:
@ksaric @Pastafarianist do we have some estimates or measurements how much this PR would speed up things ?

ksaric:
@akegalj I don't have any (CPU) numbers, I vaguely remember ~7% CPU usage through time just on deserialization, but there is about ~200Mb overhead when deserializing all the slot leaders, which is a lot.

domenkozar:
1.0 is merged regularly to master, so the PR should be made against 1.0

gromakovsky:
It's hard to track two PRs, let's close this one, cardano-sl-1.0 should be merged to master after #1725 is merged, so these changes won't disappear.

PR #1697, opened 2017-10-01
[CSL-1718] Initial version of instruction for US

georgeee:


docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
+Overview
+========
+
+Software update is a protocol mechanism that allows nodes to
+automatically update their executables without changing protocol
```
deepfire: A minor nit: the binary substitution mechanism is powerful enough that any changes at all can be introduced at all, so the "without changing the protocol itself" is merely about intent, I'd say.  Perhaps that should be clarified.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+=============
+1. Cluster nodes' keys. They are crucial to have since cluster nodes
+   are the only nodes that have stake in the system so they are the
+   only nodes that can propose an update and vote for it.
+2. Access to the S3 bucket
+3. Installers corresponding to software version we're pushing to cluster
+   - Those are two executables (exe for win, pkg for mac) which are to be provided by QA (after QA procedures passed)
+   - Advice: after providing installers, ask QA to confirm hashes are same, we don't want to ship users incorrect installers to users
+4. Software version
+   - Single integer, denote application version as for installers provided
```
deepfire: I'm sorry, I'm not quite sure about the meaning here -- is it:

1. the installers themselves get a second, single-integer version attached,
2. the bundled `cardano-node` binaries have the version attached
3. something else?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+Note `--system-start 0` is perfectly valid because we don't need actual value for `cardano-auxx`.
+
+Import secret keys (only 4 of 7 needed, because we need majority):
+
+```
+stack exec cardano-auxx -- --system-start 0 --log-config
+  scripts/log-templates/log-config-qa.yaml
+  --logs-prefix "logs/auxx-update-1.0.1" --db-path auxx-update-1.0.1
+  --peer <relay dns name>:3000 cmd --commands "add-key node1.key, add-key node2.key, add-key node3.key, add-key node4.key, listaddr"
+```
```
deepfire: This combination of  "import" and "propose-update" below suggests that keys are actually copied somewhere by this procedure.

This sounds troublesome, and raises a question -- has this procedure undergone security audit?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+
+Propose an update:
+
+```
+stack exec cardano-auxx -- --system-start 0 --log-config
+  scripts/log-templates/log-config-qa.yaml
+  --logs-prefix "logs/auxx-update-1.0.1" --db-path auxx-update-1.0.1
+  --peer <relay dns name>:3000 cmd --commands "propose-update 0 0.0.0 0 20 2000000 csl-daedalus:1 win64 daedalus1.exe none macos daedalus1c.pkg none"
+```
```
deepfire: I think it would be help our confidence in using this, if the arguments to `propose-update` could be documented.

I looked at https://github.com/input-output-hk/cardano-sl/blob/master/auxx/src/Command/Types.hs#L40, but even it has a line like:
```
puIdx :: Int -- TODO: what is this? rename
```.

File: docs/how-to/push-update-on-mainnet.md (hunk 3 above)
Comment (deepfire) [reply docs/how-to/push-update-on-mainnet.md]: In case the keys are indeed copied to another location -- perhaps the update mechanism could be reimplemented, so that there is no need for that?

File: docs/how-to/push-update-on-mainnet.md [hunk 5]
```diff
@@ -0,0 +1,118 @@
...
+
+In command `propose-update`:
+
+* First argument is index of imported key you're sending update
+from. Get it from `listaddr` output -- any cluster node will
+do (`0` states for first key imported).
+* Parameters 2-5 are block version data parameters -- leave they as
```
domenkozar: leave them*

It's important to document these parameters so we know what is their importance if they need to be changed

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+stack exec cardano-auxx -- --system-start 0 --log-config
+  scripts/log-templates/log-config-qa.yaml
+  --logs-prefix "logs/auxx-update-1.0.1" --db-path auxx-update-1.0.1
+  --peer <relay dns name>:3000 cmd --commands "propose-update 0 0.0.0 0 20 2000000 csl-daedalus:1 win64 daedalus1.exe none macos daedalus1c.pkg none"
+```
+
+In command `propose-update`:
+
+* First argument is index of imported key you're sending update
+from. Get it from `listaddr` output -- any cluster node will
```
domenkozar: Shouldn't this say that the output will be given by "propose an update" step since it executes it?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+do (`0` states for first key imported).
+* Parameters 2-5 are block version data parameters -- leave they as
+they are provided in the cli example
+* `csl-daedalus:N` is software version description, you should substitute `N` with integer provided along with installers (see *Prerequisites* section, *4.* item)
+* End of command: list of triples -- update installers. Provide the path to installers
```
domenkozar: Can we describe what each field of the three means? For example `none`?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+Value `upId` is used on one of next steps, when we'll vote for update.
+
+Uploading update files
+======================
+
+Upload installers to S3 bucket:
+* `update-system-testing` if update is being performed on staging
+* `update-system` if update is being performed on live mainnet cluster
```
domenkozar: This will need further instructions from devops where they are located.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+  --peer <relay dns name>:3000 cmd --commands "vote 1 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 2 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 3 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 4 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6"
+```
+
+In `vote N y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6`:
+
+* `N` should be index of key (from `listadr`)
+* `y` is for yes
+* `4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6` is update proposal id or `upId`
+
+Votes will be sent to the network, software, update will apply soon.
```
domenkozar: How do we know it's succesfully? When is soon?

georgeee:
@domenkozar it's written with assumption you execute all steps from single machine, but could be rearranged to be executed on multiple machines

Actual restriction is to have secret key at place

docs/how-to/push-update-on-mainnet.md
volhovM: I'll update it.

docs/how-to/push-update-on-mainnet.md
volhovM: Keys are imported into the local wallet secret key file `secret.key` right in the folder you execute `auxx` in.

docs/how-to/push-update-on-mainnet.md
volhovM: Fixed

docs/how-to/push-update-on-mainnet.md
volhovM: Fixed.

docs/how-to/push-update-on-mainnet.md
volhovM: cc @georgeee

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,118 @@
...
+   - Single integer, denote application version as for installers provided
+
+Proposing the update
+====================
+
+Ensure your keys are somewhere around and named, say, `nodeX.key`,
+where `X ∈ [1..7]`. Procedure is done using `cardano-auxx`
+executable.
+
+You should change param `--peer` for all commands listed below. Also consider changing `--log-config`, `--logs-prefix`.
```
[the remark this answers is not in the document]
volhovM: Yes.

docs/how-to/push-update-on-mainnet.md
volhovM: cc @gromakovsky i don't remember when it kicks in. Next epoch?

docs/how-to/push-update-on-mainnet.md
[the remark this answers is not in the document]
domenkozar: Let's document the purpose of this flag then and just say it can be any relay part of the cluster

docs/how-to/push-update-on-mainnet.md
domenkozar: I was told that in k slots.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+4. Nodes that see an update try to download and apply it
+   automatically.
+
+Now, again, step-by-step.
+
+Prerequisites
+=============
+1. Cluster nodes' keys. They are crucial to have since cluster nodes
+   are the only nodes that have stake in the system so they are the
+   only nodes that can propose an update and vote for it.
```
jmitchell: Are all 7 core node keys required, or is a majority technically sufficient? Using all 7 for now is fine, but I'd like to understand the minimal technical requirement for the day when more nodes may be involved.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+   - Advice: after providing installers, ask QA to confirm hashes are same, we don't want to ship users incorrect installers to users
+4. Software version
+   - Single integer, denote application version as for installers provided
+
+Proposing the update
+====================
+
+Ensure your keys are somewhere around and named, say, `nodeX.key`,
+where `X ∈ [1..7]`. Procedure is done using `cardano-auxx`
+executable.
```
jmitchell: What branch of cardano-sl should be used for this? If this is subject to change, please document who DevOps needs to coordinate with to be sure we're using the right revision of `cardano-auxx`.

docs/how-to/push-update-on-mainnet.md
jmitchell: Ah, this is answered on line 37. Only a majority of the secret keys are required.

jmitchell: I suggest saying "Majority of cluster nodes' keys." here instead.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+
+You should change param `--peer` for all commands listed below. Also consider changing `--log-config`, `--logs-prefix`.
```
jmitchell: Does the `--log-config` mean we should expect some logging output related to the update proposal? If so, where will those logs be emitted, and how can their output be distinguished from existing logs?

docs/how-to/push-update-on-mainnet.md
jmitchell: @volhovM, how does the cluster authenticate the keys without their contents being copied? Is the authentication done over a secure channel?

Or maybe I have this all wrong, and should think of the node keys signing some asset to indicate their intent to propose an update.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+
+Syntax and semantics of `propose-update` command:
+
+```
+propose-update <N> <block ver> <script ver> <slot duration> <max block size> <software ver> <propose_file>?
+```
+
+* First argument is index of imported key you're sending update
+from. Step "import secret keys" has `listaddr` command in the end. Check it's output -- any cluster node will
+do (`0` states for first key imported).
```
jmitchell: Suppose I import the first 4 node keys as you demonstrate above, starting with `node1.key`. Then suppose I run `propose-update` with `0` as the first argument. Therefore the update will be sent from the node corresponding `node1.key`, right?

How/when does that first node verify that 3+ other nodes have proposed the same update?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+=============
+1. Cluster nodes' keys. They are crucial to have since cluster nodes
+   are the only nodes that have stake in the system so they are the
+   only nodes that can propose an update and vote for it.
+2. Access to the S3 bucket
+3. Installers corresponding to software version we're pushing to cluster
+   - Those are two executables (exe for win, pkg for mac) which are to be provided by QA (after QA procedures passed)
+   - Advice: after providing installers, ask QA to confirm hashes are same, we don't want to ship users incorrect installers to users
+4. Software version
+   - Single integer, denote application version as for installers provided
```
jmitchell: How is this version number determined?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+
+* First argument is index of imported key you're sending update
+from. Step "import secret keys" has `listaddr` command in the end. Check it's output -- any cluster node will
+do (`0` states for first key imported).
+* Parameters 2-5 are block version data parameters -- leave them as
+they are provided in the cli example.
+* `csl-daedalus:1` is software version description, you should substitute `1` (version) with the integer provided along with installers (see *Prerequisites* section, *4.* item)
+* End of command: list of triples -- update installers. Provide the path to installers
+without any slashes (installers should be in the same dir as auxx is
+launched in). First tuple element is platform, second is path, `none` stands for binary diff (this feature is not used for now).
```
jmitchell: Is issuing an update to only one platform (`win64` or `macos`) supported? If not, does `cardano-auxx` require one triple tuple for each?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+
+That's it. Successfull command output looks like this:
+
+```
+[smart-wallet:DEBUG] [2017-09-21 15:13:24 MSK] Proposing update...
+Read file installer062win64.exe succesfuly, its hash: 01abf1c8b881c2f8ea4d1349a700f29d4088e68dc04b6bf4680ea7e14638373e
```
jmitchell: Please update the output to indicate what hashing algorithm is used. We may need to audit the hash reported here against a downloaded update.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+Note:
+
+```
+Read file installer062win64.exe succesfuly, its hash: 01abf1c8b881c2f8ea4d1349a700f29d4088e68dc04b6bf4680ea7e14638373e
+Read file installer062macos64.pkg succesfuly, its hash: 3bc1084841fb99fff03ef92bc35eef9a80a20aeca688dfc1b50a4aa6dd6f7c73
+```
+
+These hashes should be used as names when you will upload installers on S3 buckets.
```
jmitchell: To clarify, before uploading the Windows installer from the auxx directory to the appropriate S3 bucket, it must be renamed from `installer062win64.exe` to `01abf1c8b881c2f8ea4d1349a700f29d4088e68dc04b6bf4680ea7e14638373e` with no extension or anything, correct?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+Value `upId` is used on one of next steps, when we'll vote for update.
+
+Uploading update files
+======================
+
+Upload installers to S3 bucket:
+* `update-system-testing` if update is being performed on staging
+* `update-system` if update is being performed on live mainnet cluster
```
jmitchell: What state indicates to a cluster which of these two buckets should be used? `DCONFIG`?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+stack exec cardano-auxx -- --system-start 0 --log-config
+  scripts/log-templates/log-config-qa.yaml
+  --logs-prefix "logs/auxx-update-1.0.1" --db-path auxx-update-1.0.1
+  --peer <relay dns name>:3000 cmd --commands "vote 1 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 2 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 3 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 4 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6"
+```
+
+In `vote N y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6`:
+
+* `N` should be index of key (from `listadr`)
+* `y` is for yes
```
jmitchell: Are `no` votes possible?

More generally:
1. Do proposals expire?
2. Can proposals be withdrawn?

If not, there's some risk that a very old proposed update which wasn't enacted could be inadvertently voted in later on.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+```
+stack exec cardano-auxx -- --system-start 0 --log-config
+  scripts/log-templates/log-config-qa.yaml
+  --logs-prefix "logs/auxx-update-1.0.1" --db-path auxx-update-1.0.1
+  --peer <relay dns name>:3000 cmd --commands "vote 1 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 2 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 3 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6,vote 4 y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6"
+```
+
+In `vote N y 4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6`:
+
+* `N` should be index of key (from `listadr`)
```
jmitchell: Does this mean only nodes which were involved in proposing an update allowed to vote on it?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+* `y` is for yes
+* `4c827d6fe03d4c3646ebbbc28d4e09c57690e1dcba54b9adc0050d3f76734cf6` is update proposal id or `upId`
+
+Successfull output ends in "submitted a vote". Votes will be sent to the network, software update will apply soon.
```
jmitchell: 1. What mechanism guarantees that proposed updates which haven't yet been approved by a majority can't possibly result in a software update?
2. How do we check whether a node has applied an update yet?
3. If a node fails to automatically apply an update (or fails during the update process) is there a recommended manual update procedure?
4. Can we check what ratio of the network is running the current set of updates?
5. Suppose while a node is offline multiple updates are proposed and voted in. When the node comes back online, does it apply the updates sequentially in the order they were ratified, a random one, only the latest one, or something else? Does this strategy ensure the node's final state is valid?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,120 @@
...
+Now, again, step-by-step.
+
+Prerequisites
+=============
+1. Cluster nodes' keys. They are crucial to have since cluster nodes
+   are the only nodes that have stake in the system so they are the
+   only nodes that can propose an update and vote for it.
+2. Access to the S3 bucket
+3. Installers corresponding to software version we're pushing to cluster
+   - Those are two executables (exe for win, pkg for mac) which are to be provided by QA (after QA procedures passed)
```
jmitchell: Are these executables the same sort our CI systems are creating? I imagine updates should all be signed using the same keys we currently use for both the standard Daedalus installers.

docs/how-to/push-update-on-mainnet.md
neongreen: > Are no votes possible?

Yes, `no` votes are possible with `n`.

> Do proposals expire?

Yes:

```
-- According to implicit agreement rule all proposals which were put
-- into blocks earlier than 'updateImplicit' slots before slot
-- of current block become implicitly decided (approved or rejected).
-- If proposal's total positive stake is bigger than negative, it's
-- approved. Otherwise it's rejected.
```

> Can proposals be withdrawn?

AFAIK no, but I'm not 100% sure.

docs/how-to/push-update-on-mainnet.md
neongreen: No, it means that you can only vote from addresses that are under your control. I.e. `vote 1 y <proposal ID>` on your machine and `vote 1 y <proposal ID>` on my machine will count as two different votes because we (supposedly) have different secret keys.

docs/how-to/push-update-on-mainnet.md
neongreen: Any branch that is compatible with the blockchain that you want to publish the update on (`cardano-sl-1.0` in our case).

docs/how-to/push-update-on-mainnet.md
neongreen: Yes, there will be a message saying `Proposing update...` in the console, followed by `Update proposal submitted, upId: <...>` if the submission was successful (or, well, there'll be an error message if it wasn't successful 🙂). The logs will be emitted to the folder specified by `--logs-prefix`, e.g. you can specify `--logs-prefix logs/update-mainnet` or something.

docs/how-to/push-update-on-mainnet.md
jmitchell: `node/mainnet-genesis.json` in `cardano-sl-1.0` includes:

```
        "updateImplicit": "10000",
```

If my calculations are right that means in mainnet all proposals will be implicitly accepted or rejected within ~2.3 days after they are proposed.

docs/how-to/push-update-on-mainnet.md
neongreen: > Or maybe I have this all wrong, and should think of the node keys signing some asset to indicate their intent to propose an update.

That's exactly right. The node keys sign an update proposal, and then it's broadcast to other nodes as a special message (similarly to how transactions are sent) and ultimately included into the next block.

docs/how-to/push-update-on-mainnet.md
jmitchell: For now the plan is to, from a single machine, have 4/7 core node keys involved in creating an update proposal. I'm wondering whether the remaining 3 nodes would be able to vote, even though they weren't involved in the proposal.

If I understand correctly, those remaining 3 *would* be allowed, and it could be accomplished by running the above command from any machine which has added any of the keys for those nodes. However, that's out of scope for the procedure we'll be using in the near term.

docs/how-to/push-update-on-mainnet.md
neongreen: The technical requirement is that the majority of _stake_ is required. I don't know how the stake is split between the seven secret keys, but I assume that the split is equal (I'll ask in Slack just in case).

jmitchell: Ok, got it :)

My understanding from genesis creation is stake started out evenly distributed across these core nodes. I'm not sure whether that fluctuates as transactions are made.

docs/how-to/push-update-on-mainnet.md
neongreen: By the `applicationVersion` key in `configuration.yaml`. The current version is 0, and thus any number bigger than 0 will be acceptable.

neongreen: (But ideally you should specify `applicationVersion` matching the version in `configuration.yaml` contained in the new installer)

neongreen: Upd: I'm wrong, the current version for mainnet is `1`, thus the number should be at least `2`.

docs/how-to/push-update-on-mainnet.md
neongreen: > Therefore the update will be sent from the node corresponding node1.key, right?

The update will be signed with node1.key, that's the only relevant thing (cardano-auxx doesn't send commands to a node, it _is_ a node).

docs/how-to/push-update-on-mainnet.md
jmitchell: Should each software version be `1 +` the latest applied update?

Suppose over a short period of time we need to propose multiple updates. Is the order in which they're ratified important? For instance, if version `4` is ratified before `3` will the update system apply `4` immediately?

neongreen: Upd: I'm wrong again 🤕 Some users are running `0`, some users are running `1`. Thus we'll need two updates (first we'll upgrade everyone to `1`, then we'll upgrade everyone to `2`). I or someone else will provide more details soon.

docs/how-to/push-update-on-mainnet.md
jmitchell: > it [cardano-auxx] *is* a node

Is this true even when all the commands are executed on the deployer?

docs/how-to/push-update-on-mainnet.md
neongreen: Ok, I'm told that devops should know which commit corresponds to 1, so we'll need to take an installer for v1, propose it as an update, then take an installer for v2 and propose it as well.

docs/how-to/push-update-on-mainnet.md
neongreen: > Are these executables the same sort our CI systems are creating?

AFAIK yes.

docs/how-to/push-update-on-mainnet.md
jmitchell: @domenkozar, are you clear on what rev corresponds to v1?

@neongreen, how long after v1 is ratified (majority of stake votes yes) should devops wait to propose and ratify v2 to be sure no nodes get v2 before v1? Even better, please answer questions 2-5 at the end of the PR.

docs/how-to/push-update-on-mainnet.md
neongreen: Yep, unless there's something I misunderstand about your comment. By “executed on the deployer” do you mean “cardano-auxx will be run on the deployer”? If so, then yes.

docs/how-to/push-update-on-mainnet.md
neongreen: I know for sure that the URL should be `<update server>/<hash with no extension or anything>`, so I guess that the bit about S3 is true.

docs/how-to/push-update-on-mainnet.md
neongreen: DCONFIG doesn't exist anymore. The node fetches the update from a server hardcoded in [Daedalus launcher](https://github.com/input-output-hk/daedalus/blob/master/installers/Launcher.hs#L59), but I haven't found a build of Daedalus which has `update-system-testing` so I might be wrong.

docs/how-to/push-update-on-mainnet.md
neongreen: > I'm wondering whether the remaining 3 nodes would be able to vote, even though they weren't involved in the proposal.

Yes.

docs/how-to/push-update-on-mainnet.md
neongreen: > What mechanism guarantees that proposed updates which haven't yet been approved by a majority can't possibly result in a software update?

There's no ironclad guarantee for this, of course, but we only download confirmed proposals and we only call `addConfirmedProposal` on a proposal if (according to our calculations) it has been approved by stake majority.

neongreen: > If a node fails to automatically apply an update (or fails during the update process) is there a recommended manual update procedure?

Just run the newest installer. You might also need to delete `%APPDATA%\Daedalus` and restore your wallet afterwards.

> How do we audit the update/vote information written to the blockchain?

It looks like auxx doesn't have a command like “show all available proposals”, but it should be easy to add. Otherwise, we can just look at another node's logs – when it prints information about received blocks, it should also show the votes/proposals contained in those blocks.

docs/how-to/push-update-on-mainnet.md
neongreen: > how long after v1 is ratified (majority of stake votes yes) should devops wait to propose and ratify v2 to be sure no nodes get v2 before v1?

Nodes can't get v2 before v1, as the current “confirmed version” on the blockchain is v0 (despite some nodes having v1) and the proposal with v2 would get rejected. After v1 becomes confirmed, we can push the v2 update. Unless I'm mistaken, the timeline is as follows:

* v1 is pushed and voted on
* `k` blocks later, v1 becomes confirmed and nodes start downloading it
* immediately after v1 becomes confirmed, we can push v2

docs/how-to/push-update-on-mainnet.md
neongreen: We use Blake2b_256, but data is encoded as CBOR before hashing it, so you won't be able to simply hash the file using an independent CLI tool. If needed, I can provide info on what bytestring you would need to prepend to the file to get the correct hash. Alternatively, you can use `hashFile` from module `Command.Update` in ghci.

docs/how-to/push-update-on-mainnet.md
neongreen: > does cardano-auxx require one triple tuple for each [platform]?

Yes. `propose-update ... <platform1 installerPath1 diffPath1> <platform2 installerPath2 diffPath2> ...`

> Is issuing an update to only one platform (win64 or macos) supported?

Yes.

docs/how-to/push-update-on-mainnet.md
neongreen: > **2.** How do we check whether a node has applied an update yet?

A particular node that you have access to?
* It will restart itself (which should be visible in the logs)
* After a restart it will print its version (e.g. `Application: cardano-sl:0`), which too will be visible in the logs

> **4.** Can we check what ratio of the network is running the current set of updates?

AFAIK no. We can try looking at users' logs from the report server, but I'm not sure that we have enough info about each user to avoid double-counting some of them.

neongreen: > **5.** When the node comes back online, does it apply the updates sequentially in the order they were ratified, select a random one, select only the latest one, or something else? Does this strategy ensure the node's final state is valid?

The worker which applies the last downloaded update runs concurrently with the worker which downloads updates, so it might end up choosing a random one. E.g. if there are two confirmed updates, then both of these options are possible:

* download v1, restart & apply v1, download v2, restart & apply v2
* download v1, download v2, restart & apply v2

One thing that we can't do is apply v1 after v2, so the state of the node will be sound if all of {v0 → v1, v1 → v2, v0→v2} are valid. If you don't test v0→v2, then you would be relying on luck.

jmitchell: > One thing that we can't do is apply v1 after v2

Is this scenario prevented by code? I realize it would be bad, but I'm unclear whether we need to do anything special to avoid it.

docs/how-to/push-update-on-mainnet.md
jmitchell: Suppose the latest ratified update is version `5`, and we try to:

1. propose an update tagged as version `7`,
2. have majority vote in favor of `7`, and
3. wait for version `7` to be applied.

What's the first of those 3 events to fail due to version `6` getting skipped, and how will it manifest?

If it doesn't fail on event (1), how should next update be versioned to ensure it is applied (and that `7` isn't because it may not be compatible with `6`)? Would the version after that need to be `7` or `8`?

docs/how-to/push-update-on-mainnet.md
jmitchell: > If you don't test v0→v2, then you would be relying on luck.

Can we seriously consider enforcing consecutive version updates in the code to prevent this? As the number of updates grows the number of combinations for possible update paths grows rapidly.

jmitchell: So far we've discussed applying updates using Daedalus installers, but what about the wallets exchanges are running?

docs/how-to/push-update-on-mainnet.md
deepfire: What about the version gap for the non-updated platform, @neongreen?  The scenario:

1. Suppose both platforms A and B start with version N.
2. Update round 1: Platform A is updated to N+1, platform B doesn't get an update
3. Update round 2: both platforms are updated

What are the supposed version numbers in this case? I presume N+2 and N+1.  But that means we have completely independent versioning for the platforms.

Can you foresee any problems with that? /cc @akegalj

docs/how-to/push-update-on-mainnet.md
deepfire: @neongreen, I see that you provide many answers that are not reflected in this document.

Are they available somewhere else?  If not, do you plan on recording them here (anywhere else)?

docs/how-to/push-update-on-mainnet.md
deepfire: @neongreen, this sub-thread seems to imply that it is impossible to ratify `vN+2`, without ratifying `vN+1` first (having `vN` ratified as a precondition, naturally).  Is this correct?

neongreen: It will fail at step (1) – nodes would reject the proposal and it wouldn't be included into blocks. You can verify that by grepping the first several chars of a proposal ID in papertrail:

> Oct 03 16:32:56 mainnet.ec2.p-b-4 cardano-node-start: [node.server.update:DEBUG:ThreadId 244000] [2017-10-03 13:32:55 UTC] Processing of proposal csl-daedalus:3 { block v0.0.0, UpId: 53c6d13b, { scripts v0, slot duration (mcs): 20000, block size limit: 1.907 MiB, header size limit: 1.907 MiB, tx size limit: 4 KiB, proposal size limit: 700 B, mpc threshold: 20000000000000/1000000000000000 (approx. 0.02), heavyweight delegation threshold: 300000000000/1000000000000000 (approx. 0.0003), update vote threshold: 1000000000000/1000000000000000 (approx. 0.001), update proposal threshold: 100000000000000/1000000000000000 (approx. 0.1), update implicit period (slots): 10000, softfork rule: no change, tx fee policy: no change, unlock stake epoch: no change }, tags: [patak], no attributes } with id 53c6d13b failed: proposal 53c6d13b has wrong software version for app csl-daedalus **(last known is 0, proposal contains 3)**

(Unfortunately, it looks like this message won't be logged into a locally running node. Papertrail it is.)

docs/how-to/push-update-on-mainnet.md
neongreen: If you update platform A to N+1, you won't be able to update B to N+1 later – you'll need to update both to N+2.

* You publish an update to N+1 that contains an installer for Windows. N+1 becomes a confirmed version after `k` blocks. Windows nodes download the update. macOS nodes don't download anything.
* Someone else publishes an update to N+5. It gets rejected because the current confirmed version is N+1.
* You publish an update to N+2 that contains both a Windows and a macOS installer. N+2 becomes a confirmed version after `k` blocks. Windows and macOS nodes download the update (even though for Windows it's N+1→N+2 and for macOS it's N→N+2).

neongreen: > Can you foresee any problems with that?

If the Daedalus installers don't rely on some particular old files being around – no, there shouldn't be any problems.

docs/how-to/push-update-on-mainnet.md
neongreen: I'll update the doc after I finish answering all unanswered questions.

docs/how-to/push-update-on-mainnet.md
neongreen: > Is [applying v2 after v1] prevented by code?

Yep. We know our software version and don't apply anything that isn't newer.

neongreen: AFAIK we'll just have them rebuild the node manually.

jmitchell: > don't apply anything that isn't newer.

That's not my concern. My concern is skipping versions.

jmitchell: I gather from https://github.com/input-output-hk/cardano-sl/pull/1697/files#r142403318 that proposals which skip versions are prevented.

jmitchell: Ah, but even if all ratified proposals have consecutive versions, updates may be applied by skipping versions. My earlier concern still stands.

>> If you don't test v0→v2, then you would be relying on luck.

> Can we seriously consider enforcing consecutive version updates in the code to prevent this? As the number of updates grows the number of combinations for possible update paths grows rapidly.

docs/how-to/push-update-on-mainnet.md
gromakovsky: To be more precise, in `k` **blocks** (which slightly different from **slots**) after the block with votes.

docs/how-to/push-update-on-mainnet.md
volhovM: Logging output is not crucial in fact. Everything important (commands' output) is printed using `putStrLn`, not through logger. You'll have everything on your `stdout`.

docs/how-to/push-update-on-mainnet.md
volhovM: >The update will be signed with node1.key
Exactly, we just need protocol participant with enough stake (poor can't propose updates). Proposal will be sent from `auxx` directly.

docs/how-to/push-update-on-mainnet.md
volhovM: We do not need to have update system documentation in this manual, it has another purpose. 
https://cardanodocs.com/cardano/update-mechanism/
https://cardanodocs.com/technical/updater/

docs/how-to/push-update-on-mainnet.md
volhovM: Yes, true.

docs/how-to/push-update-on-mainnet.md
volhovM: @fersel should know.

docs/how-to/push-update-on-mainnet.md
volhovM: We always download the latest update, consecutive updates are not supposed to be supported for now. We distribute installers (for now, not binary diffs) so it shouldn't be a problem.

Consecutive updates are really tedious for those who need to update for a lot of versions since every update is basically "kill node, run executable, start node over".

docs/how-to/push-update-on-mainnet.md
deepfire: > last known is 0, proposal contains 3

@neongreen, to clarify -- is "known" == "ratified" here?

docs/how-to/push-update-on-mainnet.md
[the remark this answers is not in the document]
jmitchell: bump

docs/how-to/push-update-on-mainnet.md
deepfire: @volhovM, @neongreen, sorry, but aren't these two statements in a more-or-less direct contradiction?

statement 1:
> The worker which applies the last downloaded update runs concurrently with the worker which downloads updates, so it might end up choosing a random one. E.g. if there are two confirmed updates, then both of these options are possible:
>
>    download v1, restart & apply v1, download v2, restart & apply v2
>    download v1, download v2, restart & apply v2

statement 2:

> We always download the latest update, consecutive updates are not supposed to be supported for now.

?

volhovM: https://github.com/input-output-hk/cardano-sl/blob/master/lib/src/Pos/Update/Worker.hs#L67
We download the newest one.

docs/how-to/push-update-on-mainnet.md
volhovM: @deepfire it means "adopted". Blockchain uniquely defines last software version.

domenkozar:
@volhovM will you incorporate PR comments into the doc?

docs/how-to/push-update-on-mainnet.md
domenkozar: @volhovM so I think we should document here the fact that node keys are imported into the local wallet db.

docs/how-to/push-update-on-mainnet.md
domenkozar: mainnet:

- URL: update.cardano-mainnet.iohk.io
- S3 Bucket: update.cardano-mainnet.iohk.io in mainnet AWS role

staging:

- URL: https://s3.eu-central-1.amazonaws.com/update-system-testing/
- S3 Bucket: update-system-testing

docs/how-to/push-update-on-mainnet.md
domenkozar: @volhovM there are two things to be documented here:

a) for wallets, version is determined from `applicationVersion` key in configuration.yaml

b) version number is monotonically increasing and the update system can't skip a single digit (1 -> 3)

docs/how-to/push-update-on-mainnet.md
domenkozar: What @neongreen just said contradicts the previous requrement, where version names are monotonically increasing.

So we have an issue here that should be documented: all platforms should be updated at the same time to the next version.

docs/how-to/push-update-on-mainnet.md
domenkozar: @volhovM needs a note about how hash is calculated

docs/how-to/push-update-on-mainnet.md
domenkozar: @volhovM docs to be updated with an example as Jake gave

docs/how-to/push-update-on-mainnet.md
domenkozar: Above 2 links should be inside this document at the beginning as prerequisite.

docs/how-to/push-update-on-mainnet.md
domenkozar: @denisshevchenko these questions about the update system are really useful, I suggest we incorporate them into the cardano docs.

docs/how-to/push-update-on-mainnet.md
domenkozar: @volhovM here we need a clarification that those two executables are installers provided by our CI.

domenkozar:
@denisshevchenko lots of useful docs here for the update system :)

docs/how-to/push-update-on-mainnet.md
volhovM: I'm highly confused :(

>where version names are monotonically increasing.

What? Version names for each platform will be montonically increasing, though non-consequent versions are possible.

>So we have an issue here that should be documented: all platforms should be updated at the same time to the next version.

How did you come to this statement?

gromakovsky: Application versions must be consequent. Blockchain will reject an update to version 3 if the last confirmed version is 1.
Regarding installers, nothing is checked by blockchain protocol. We can include arbitrary number of installer hashes into update proposal (well, there is upper limit on proposal size), we can use arbitrary system tags, etc. It's up to us to decide what which installers we want to include.

> So we have an issue here that should be documented: all platforms should be updated at the same time to the next version.

If you mean that we must always include all supported system tags into update proposal, then it's wrong. We can update only Windows or not update anything. And it might be useful in case we fix only Windows-specific bug.
However, it we want to update all platforms to version X, then we must include all system tags into update proposal. Updating to X more than once is impossible.

gromakovsky: It's essential to understand that application version is associated with application name (e. g. `csl-daedalus`), not with system tag (e. g. `win64`).
Application version is checked by protocol. Installer hashes are not checked in any way.

docs/how-to/push-update-on-mainnet.md
gromakovsky: @jmitchell @domenkozar as @neongreen says, hash is calculated in two steps: CBOR-encode data and then hash it. I can add a command to `cardano-sl-auxx` which computes and prints hash of an installer (and write about it to the doc). Will it be sufficient?

docs/how-to/push-update-on-mainnet.md
volhovM: Done.

docs/how-to/push-update-on-mainnet.md
volhovM: Done.

docs/how-to/push-update-on-mainnet.md
jmitchell: @gromakovsky, such a command would be very useful. Please add it and the docs here.

docs/how-to/push-update-on-mainnet.md
cleverca22: should this be `macos` or `macos64`?

gromakovsky: It should be `macos64`. Good catch, thanks, will fix in a moment. It should be the same as `systemTag` in configuration.

volhovM:
@domenkozar @jmitchell approve maybe?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,187 @@
...
+4. Software version
+   - Single integer, must be greater by 1 than the last confirmed
+     version. Initially `csl-daedalus` application has version 0. The
+     first update should have version 1, the next one 2, etc. If you
+     try to propose an update with version 3 while the last confirmed
+     one is 1, your update proposal won't be accepted by the nodes.
+   - Node knows its software version from `configuration.yaml` file,
+     `applicationVersion` field. When you propose an update to version
+     X, make sure that all attached installers use configuration with
+     `applicationVersion` set to `X`.
```
jmitchell: CSL-1758 describes a bad infinite update loop scenario if we get the versions wrong. If we conclude in that issue that no technical safeguards will be implemented, and that human verification alone is acceptable then I suggest the following:

1. Swap items 3 and 4
2. In what's currently number 3, add that getting this wrong can result in Daedalus users suffering denial of service due to repeating Daedalus updates.

volhovM: @jmitchell done. Though as fersel said in the YT issue, this scenario is mostly improbable.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,192 @@
...
+
+Uploading update files
+======================
+
+Upload installers to S3 bucket:
+* For staging:
+    * URL: https://s3.eu-central-1.amazonaws.com/update-system-testing/
+    * S3 Bucket: `update-system-testing`
+* For mainnet:
+    * URL: https://update.cardano-mainnet.iohk.io
```
cleverca22: the URL in https://github.com/input-output-hk/daedalus/blob/eb713a66eb2c0445fbe8c2faa59f0884edd83712/installers/Launcher.hs#L68 does not use TLS
and the SSL cert for that domain is currently invalid

cleverca22: the URL in https://github.com/input-output-hk/daedalus/blob/eb713a66eb2c0445fbe8c2faa59f0884edd83712/installers/Launcher.hs#L68 does not use TLS
and the SSL cert for that domain is currently invalid

cleverca22: it would help in the future to have a link to `Launcher.hs` in the docs, so devs know where the url comes into play, and if it may be different on the current branch

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,192 @@
...
+
+COMMONOPTS="--system-start 0 --configuration-file ../cardano-sl/node/configuration.yaml --configuration-key ${CONFIG_KEY}"
+
+AUXXOPTS="--log-config ../cardano-sl/scripts/log-templates/log-config-qa.yaml --logs-prefix logs/aux-update.1.0.1 --db-path aux-update-1.0.1 --peer ${RELAY_PEER}"
+```
+
+Rearranging the keys
+--------------------
+
+Apply `cardano-keygen rearrange` to the keys to make them usable for proposing an update and voting:
```
domenkozar: Shouldn't rearrange be ran after importing? If not, then I don't understand what it does.

domenkozar:
@volhovM almost there :)

docs/how-to/push-update-on-mainnet.md
georgeee: @cleverca22 is it a hard requirement? From our point of view, checking hash is well sufficient?

docs/how-to/push-update-on-mainnet.md
volhovM: No, rearrangement should be executed before importing. In fact, key won't be imported unless it's rearranged. 

User secret has these fields:
```
    { _usKeys      :: [EncryptedSecretKey]
    , _usPrimKey   :: Maybe SecretKey
    ... }
```
And by default `_usKeys` is empty. Rearrangement creates an encrypted key from primary key and puts it into `_usKeys`.

docs/how-to/push-update-on-mainnet.md
volhovM: I've updated the doc including the link, but I don't see any issue big enough to block the merge. Even if client won't be able to download from http because we serve files on https it's easy to resolve (serve on both instead of redirecting http->https and ship https: prefix in the next update).
@cleverca22

docs/how-to/push-update-on-mainnet.md
volhovM: @domenkozar ^

docs/how-to/push-update-on-mainnet.md
```diff
@@ -0,0 +1,195 @@
...
+
+Let's break down the invocation of `propose-update`. First come arguments that you almost certainly won't need to modify:
+
+* `0` is the index of key that will be used to sign the update.
+* `0.0.0` is block version to be used after the update. Currently it's `0.0.0` as well and so the update won't change the block version.
+* `0` is new script version. Again, it's the same as the current one.
+* `20` is new slot duration (in seconds). Unchanged.
+* `2000000` is new maximum block size (in bytes). Unchanged.
+
+The next argument (`csl-daedalus:1`) is software version description. You should substitute `1` (version) with the integer provided along with installers (see *Prerequisites* section, *4.* item)
```
gromakovsky: 4 or 3?

volhovM: fixed

volhovM:
@cleverca22 Can you re-review please?

PR #1701, opened 2017-10-02
[CSL-1734] Add a command to auxx for sending an update proposal

int-index:
* added a propose-unlock-stake-epoch command to auxx that sends an update proposal with a new value of unlockStakeEpoch
* added a print-bvd command auxx in order to monitor the current BlockVersionData
* fixed the name of puIdx (there was a TODO about it)

### DIFF
--- a/auxx/src/Command/Types.hs
+++ b/auxx/src/Command/Types.hs
@@ -10,14 +10,11 @@ module Command.Types
 
 import           Universum
 
-import           Serokell.Data.Memory.Units (Byte)
-
-import           Pos.Core.Types             (ScriptVersion)
-import           Pos.Crypto                 (PublicKey)
-import           Pos.Txp                    (TxOut)
-import           Pos.Types                  (AddrStakeDistribution, Address, BlockVersion,
-                                             EpochIndex, SoftwareVersion)
-import           Pos.Update                 (SystemTag, UpId)
+import           Pos.Crypto (PublicKey)
+import           Pos.Txp    (TxOut)
+import           Pos.Types  (AddrStakeDistribution, Address, BlockVersion, EpochIndex,
+                             SoftwareVersion)
+import           Pos.Update (BlockVersionModifier, SystemTag, UpId)
 
 -- | Specify how transactions are sent to the network during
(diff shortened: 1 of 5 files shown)


auxx/src/Command/Types.hs
```diff
@@ -38,7 +39,7 @@ data SendToAllGenesisParams = SendToAllGenesisParams
 
 -- | Parameters for 'ProposeUpdate' command.
 data ProposeUpdateParams = ProposeUpdateParams
-    { puIdx             :: Int -- TODO: what is this? rename
+    { puSecretKeyIdx    :: Int -- the node that creates/signs the proposal
```
gromakovsky: I wouldn't call it _the node_, it's the index of one of the secrets from our keyfile (`UserSecret`).

auxx/src/Command/Update.hs
```diff
@@ -126,13 +131,65 @@ propose sendActions ProposeUpdateParams{..} = do
...
+            , bvmMaxTxSize         = Nothing
+            , bvmMaxProposalSize   = Nothing
+            , bvmMpcThd            = Nothing
+            , bvmHeavyDelThd       = Nothing
+            , bvmUpdateVoteThd     = Nothing
+            , bvmUpdateProposalThd = Nothing
+            , bvmUpdateImplicit    = Nothing
+            , bvmSoftforkRule      = Nothing
+            , bvmTxFeePolicy       = Nothing
+            , bvmUnlockStakeEpoch  = Just puseUnlockStakeEpoch
```
gromakovsky: I would define a `BlockVersionModifier` with all fields set to `Nothing` somewhere and use it here (and probably in the command above).

int-index: As a matter of fact, I recall that it was defined somewhere. Good idea to use it.

gromakovsky:
Please don't restart builds, there have failing tests.

gromakovsky:
I rebased on the latest master to fix CI.

gromakovsky:
TODO: we will need to update documentation about proposing an update after it's merged to this branch.

PR #1702, opened 2017-10-02
[CSL-1689] Use newer log-warper which creates folders properly

neongreen:


gromakovsky:
All required checks failed and only the optional one succeeded :suspect:

gromakovsky:
I rebased.

PR #1704, opened 2017-10-03
[CSM-485] Fastened history fetching more

Martoon-00:
This includes:

1. Fetching limited history from database. Now time for history endpoint is approx. `min(totalTxsNum, limit) * 7` ms, which looks ok taking into account (if I understood it correctly) that passed limit is usually a number of 1-2 order of magnitude for bittrex and up to 1000 for daedalus.

2. History update is performed as single modifying db transaction, not as `get + set`

_Still need to test correctness of this_

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Methods/History.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/History.hs
@@ -15,7 +15,7 @@ import qualified Data.Set                   as S
 import           Data.Time.Clock.POSIX      (getPOSIXTime)
 import           Formatting                 (build, sformat, stext, (%))
 import           Serokell.Util              (listJson)
-import           System.Wlog                (WithLogger, logError, logInfo, logWarning)
+import           System.Wlog                (WithLogger, logError, logInfo)
 
 import           Pos.Aeson.ClientTypes      ()
 import           Pos.Aeson.WalletBackup     ()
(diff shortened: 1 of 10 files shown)


wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -110,7 +107,7 @@ getHistoryLimited
     -> Maybe Word
     -> m ([CTx], Word)
 getHistoryLimited mCWalId mAccId mAddrId mSkip mLimit =
-    first applySkipLimit <$> getHistory mCWalId mAccId mAddrId
+    first applySkipLimit <$> getHistory mCWalId mAccId mAddrId limit
```
Martoon-00: Oh, I wished it to be (`skip` + `limit`)

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -110,7 +107,7 @@ getHistoryLimited
     -> Maybe Word
     -> m ([CTx], Word)
 getHistoryLimited mCWalId mAccId mAddrId mSkip mLimit =
-    first applySkipLimit <$> getHistory mCWalId mAccId mAddrId
+    first applySkipLimit <$> getHistory mCWalId mAccId mAddrId (skip + limit)
```
pva701: I don't like you process even transactions which you will skip, i.e. you will try to find addresses of transactions which you will skip in future. It's double work and it might reduce performance

pva701: maybe it's enough for now, but I suggest to fix it

Martoon-00:
in favor of #1717

PR #1705, opened 2017-10-03
[CSL-1737] Run auxx with node as an option

int-index:


auxx/src/AuxxOptions.hs
```diff
@@ -34,6 +34,7 @@ data AuxxOptions = AuxxOptions
     { aoAction         :: !AuxxAction
     , aoCommonNodeArgs :: !CLI.CommonNodeArgs  -- ^ Common CLI args for nodes
     , aoPeers          :: ![NodeId]
+    , aoNodeEnabled    :: !Bool
     -- ^ Peers with which we want to communicate
     --   TODO: we also have topology, so it can be redundant.
```
gromakovsky: You spoiled haddock comments.

gromakovsky:
@int-index hlint failed.

PR #1706, opened 2017-10-03
Add wallet testing scripts

Martoon-00:


scripts/test/wallet/many-addrs-test.sh
```diff
@@ -0,0 +1,46 @@
...
+key2="$keydir/2.key"
+addrsNum=1000
+
+echo -e "\nLooking for key files $key1 and $key2"
+
+echo -e "\nResetting wallet database..."
+curl -k https://127.0.0.1:8090/api/test/reset -d ""
+
+
+echo -e "\nCreating 2 wallets..."
```
akegalj: maybe we should run two backends and send money between two nodes (launched on the same machine).

This would be more realistic test of bittrex environment then to send between two wallets within same node/wallet/wallet-db .

pva701:
I used them, extremely useful

pva701:
no approvit' ya konechno je ne budu :trollface:

gromakovsky:
Rebase on master please (to fix CI).

Martoon-00:
Right

scripts/test/wallet/many-addrs-test.sh
Martoon-00: And I don't envy that guy who will need such scenario for true similarity)

PR #1710, opened 2017-10-04
[CSL-1712] Consider delegation in block-gen

pva701:


lib/src/Pos/Generator/Block/Logic.hs
```diff
@@ -80,12 +82,14 @@ genBlock eos = do
...
             genPayload slot
             -- We need to know leader's secret key to create a block.
-            leader <-
+            precomputedLeader <-
                 lift $ maybeThrow
                     (BGInternal "no leader")
                     (leaders ^? ix (fromIntegral $ getSlotIndex siSlot))
             secrets <-
                 unInvSecretsMap . view asSecretKeys <$> view blockGenParams
+            cert <- lift $ getPskByIssuer (Right precomputedLeader)
```
volhovM: Is it alright that transitive delegation is not considered here?

pva701:
@volhovM @gromakovsky I've fixed and force pushed (not tested yet)

lib/src/Pos/Generator/Block/Logic.hs
```diff
@@ -80,12 +82,14 @@ genBlock eos = do
         EpochOrSlot (Right slot@SlotId {..}) -> withCurrentSlot slot $ do
             genPayload slot
             -- We need to know leader's secret key to create a block.
```
gromakovsky: This comment is obsolete.

lib/src/Pos/Generator/Block/Logic.hs
```diff
@@ -80,12 +82,14 @@ genBlock eos = do
...
-            leader <-
+            precomputedLeader <-
```
gromakovsky: Well, I don't know why you renamed it (leader is leader even if she delegated), but it's not essential.

lib/src/Pos/Generator/Block/Logic.hs
```diff
@@ -80,12 +82,14 @@ genBlock eos = do
...
                 lift $ maybeThrow
                     (BGInternal "no leader")
                     (leaders ^? ix (fromIntegral $ getSlotIndex siSlot))
             secrets <-
                 unInvSecretsMap . view asSecretKeys <$> view blockGenParams
+            transCert <- lift $ getDlgTransPsk precomputedLeader
+            let leader = maybe precomputedLeader (addressHash . pskDelegatePk . snd) transCert
```
gromakovsky: It's not leader. It should be called _creator_ or something.

gromakovsky:
And also should be tested.

pva701:
tested, it works

PR #1713, opened 2017-10-04
[CSM-483] Fix stream chunk handling in 'makeRequest'

int-index:
same as #1712


int-index:
(I didn't know I could change the base branch, so I closed this PR and opened another one)

PR #1724, opened 2017-10-05
[CSL-1744] Remove mpc threshold checks from 'genTestnetDistribution'

gromakovsky:
These checks aren't useful at all nowadays, because mpc threshold considers
stakes, while this function generates balances.
And they are a bit annoying when one needs to generate proper inititializer.
This changed allowed me to simplify 'TestnetBalanceOptions' generation and
fix rarely failing tests.


gromakovsky:
One test failed with the error from CSL-1741. I will restart.

PR #1730, opened 2017-10-06
[CSL-1748] Add devnet config

gromakovsky:


lib/configuration.yaml
```diff
@@ -14869,6 +14869,51 @@ mainnet_launch_base:
...
+      <<: *mainnet_base_genesis
+      spec:
+        <<: *mainnet_base_spec
+        blockVersionData:
+          <<: *mainnet_base_blockVersionData
+          slotDuration:      20000 # set to desired value
+        protocolConstants:
+          <<: *mainnet_base_protocolConstants
+          protocolMagic: 170987650
+          k: 6 # set to desired value
```
volhovM: what about `recoveryHeadersMessage`? Will it be 2200 as in mainnet?

gromakovsky: Will something bad happen if it's 2200? It should be greater than `k` and that's all we want, right?

volhovM: Well, technically yes, but in order to test block retrieval it would be better to stay sane. I suggest taking `recoveryHeadersMessage` so that retrieving blockchain is several batches, not the single on. E.G. let it be 200 here.

gromakovsky: Ok.

gromakovsky: @volhovM fixed.

PR #1738, opened 2017-10-09
[CSL-1708] Use latest version of log-warper

adinapoli-iohk:
This PR upgrades the dependency stack to depend on the latest pre-3.x revision of `log-warper`, which includes @neongreen 's work on directories plus the memory improvements, which should result in a more stable and less heavy memory footprint.

The migration was almost painless, apart from a couple of places where I had to deal with the fact I have now removed the spurious constraints on a `PureLogger` inside `log-warper` (for which we were using `UndecidableInstances` before). This means I had to either reshuffle some monadic stack around (in a trivial way) or had to create trivial `newtype` wrappers (like in the explorer case).

Please let me know if any of what I have done looks fishy!

### DIFF
--- a/godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
+++ b/godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
@@ -18,7 +18,7 @@ import           Control.Lens                   (at, (%=), (.=))
 import qualified Crypto.Random                  as Rand
 import           System.Wlog                    (CanLog, HasLoggerName (..), LogEvent,
                                                  NamedPureLogger (..), WithLogger,
-                                                 runNamedPureLog)
+                                                 launchNamedPureLog, runNamedPureLog)
 
 import           Pos.Core                       (BlockVersionData, EpochIndex,
                                                  HasGenesisData, HasProtocolConstants,
@@ -102,8 +102,10 @@ runPureTossWithLogger
     => GtGlobalState
     -> PureToss a
     -> m (a, GtGlobalState)
-runPureTossWithLogger gs pt = do
-    (res, newGS, _) <- runPureToss gs pt
+runPureTossWithLogger gs (PureToss act) = do
+    seed <- Rand.drgNew
+    let unwrapLower inner = pure $ fst $ Rand.withDRG seed inner
+    (res, newGS) <- launchNamedPureLog unwrapLower (runStateT act gs)
     return (res, newGS)
 
 evalPureTossWithLogger
--- a/godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
+++ b/godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
@@ -92,9 +92,9 @@ runPureToss
 runPureToss gs (PureToss act) = do
     seed <- Rand.drgNew
     let ((res, newGS), events) =
-            fst . Rand.withDRG seed $    -- run MonadRandom
-            runNamedPureLog  $           -- run NamedPureLogger
-            (flip runStateT gs) act      -- run State
+            fst . Rand.withDRG seed $ -- run MonadRandom
+            runNamedPureLog  $        -- run NamedPureLogger
+            (usingStateT gs) act      -- run State
     pure (res, newGS, events)
 
 runPureTossWithLogger
--- a/godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
+++ b/godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
@@ -14,11 +14,11 @@ module Pos.Ssc.GodTossing.Toss.Pure
 
 import           Universum
 
-import           Control.Lens                   (at, sequenceAOf, (%=), (.=))
+import           Control.Lens                   (at, (%=), (.=))
 import qualified Crypto.Random                  as Rand
 import           System.Wlog                    (CanLog, HasLoggerName (..), LogEvent,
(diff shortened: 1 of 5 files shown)


godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
```diff
@@ -39,11 +39,8 @@ type MultiRichmenSet   = HashMap EpochIndex RichmenSet
 -- require randomness even though they are deterministic. Note that running
 -- them with the same seed every time is insecure and must not be done.
 newtype PureToss a = PureToss
-    { getPureToss :: NamedPureLogger (
-                     StateT GtGlobalState (
-                     Rand.MonadPseudoRandom Rand.ChaChaDRG)) a
-    } deriving (Functor, Applicative, Monad,
-                CanLog, HasLoggerName, Rand.MonadRandom)
+    { getPureToss :: StateT GtGlobalState (NamedPureLogger (Rand.MonadPseudoRandom Rand.ChaChaDRG)) a
+    } deriving (Functor, Applicative, Monad, MonadState GtGlobalState, CanLog, HasLoggerName, Rand.MonadRandom)
```
gromakovsky: We use 80 characters as line length limit, in some cases it can be 100 if necessary, but this line has 111 characters, so I don't get why you changed `deriving` clause formatting here.
Cosmetic, but still.

godtossing/Pos/Ssc/GodTossing/Toss/Pure.hs
```diff
@@ -94,23 +91,19 @@ runPureToss
...
     -> m (a, GtGlobalState)
-runPureTossWithLogger gs (PureToss act) = do
-    seed <- Rand.drgNew
-    let unwrapLower a = pure $ sequenceAOf _1 $     -- (f a, b) -> f (a, b)
-                        fst $ Rand.withDRG seed $   -- run MonadRandom
-                        runStateT a gs              -- run State
-    (res, newGS) <- launchNamedPureLog unwrapLower act
+runPureTossWithLogger gs pt = do
+    (res, newGS, _) <- runPureToss gs pt
     return (res, newGS)
```
gromakovsky: It's not the same behaviour as it was before. `launchNamedPureLog` also uses `dispatchEvents` internally.

adinapoli-iohk: Argh, I suspected this reshuffling was fishy, dang!

Thanks, will amend accordingly 😉 

> Also CI failed.

That's because I'm stuck on butchering my `pkgs/default.nix` every time as I'm on Darwin and for a long-standing issue regenerating the packages will put in a dependency on `hfsevent` instead of correctly picking up `hinotify`. Please bear with me! 😛

domenkozar: Sorry :(

adinapoli-iohk:
@gromakovsky We should be in business now, sorry for the snafu! 😅

update/Pos/Update/Poll/Pure.hs
```diff
@@ -29,12 +29,13 @@ import           Pos.Update.Poll.Types     (BlockVersionState (..),
...
-    { getPurePoll :: NamedPureLogger (State Poll.PollState) a
-    } deriving (Functor, Applicative, Monad, CanLog, HasLoggerName)
+    { getPurePoll :: StateT Poll.PollState (NamedPureLogger Identity) a
+    } deriving (Functor, Applicative, Monad, CanLog, HasLoggerName, MonadState Poll.PollState)
 
 runPurePollWithLogger :: Poll.PollState -> PurePoll a -> (a, Poll.PollState, [LogEvent])
-runPurePollWithLogger ps =
-    (\((a, c), b) -> (a, b, c)) . flip runState ps . runNamedPureLog . getPurePoll
+runPurePollWithLogger ps pp =
+    let innerMonad = flip runStateT ps . getPurePoll $ pp
```
ChShersh: In `universum` there's [`usingStateT`](https://github.com/serokell/universum/blob/3a4e3ac2f2f49cd88db878eef5e140df33536994/src/Monad/Trans.hs#L50) which is just `flip runStateT`. Very minor improvement but pattern of `flip runStateT` or `flip runReaderT` or `flip RunWriterT` occurs too often so there exist helper functions for that in `universum`. Maybe you don't know about such functions. But if you find it convenient to use them you can get rid of `flip` (if you don't like `flip` of course).

adinapoli-iohk: I don't mind getting rid of `flip` here, and thanks for reminding me about those functions -- I saw that a while ago but eventually forgot about 'em and was too lazy to look them up 😁  

Allright, let me quickly amend the PR!

int-index:
> for which we were using `UndecidableInstances` before

There's nothing bad about `UndecidableInstances` and there's no reason to avoid it.

adinapoli-iohk:
@gromakovsky Can we pull the trigger on this or are we waiting for some extra review to land?

gromakovsky:
I would merge it, but we are not allowed to merge anything to cardano-sl-1.0 without approval from @dcoutts and @fersel.

adinapoli-iohk:
Gotcha, makes sense! Just to give us a little more reassurance that this PR won't break terribly things, here is a profiling report from @arybczak running a node for 8hrs with the ekg fix on top of this PR (if I have understood correctly):

alright

arybczak:
That is correct, it was a run with this PR included.

PR #1742, opened 2017-10-09
[CSL-1740] Ensure that fake change address is always long enough

gromakovsky:
Previous implementation was using one of output addresses as a fake
change address. But it can happen that 'getNewAddress' generates an address
which is longer than the fake change address, which in turn will lead to
tx fee being too small and invalid tx will be created.
Note that this problem is quite important already, even though we can only
create bootstrap era addresses with HD payload. For more details about it
please see my comment in CSL-1740.

The solution is a bit hacky, but it seems there is no better way to resolve
problem. 'getFakeChangeAddress' was added to 'MonadAddresses' which must
create an address with size greater than or equal to maximal possible size of
address generated by 'getNewAddress'. An address generated by
'getFakeChangeAddress' is used to create a fake transaction to compute fee.
In rare cases fee can be slightly greater than limit (by about 1-3 bytes
multiplied by ratio), but it's quite negligible.
Unfortunately we can't generate a real change address before computing fee,
because it requires a passphrase, but we don't want to require passphrase to
compute fee.

### DIFF
--- a/auxx/src/Mode.hs
+++ b/auxx/src/Mode.hs
@@ -45,6 +45,8 @@ import           Pos.Core                         (Address, HasConfiguration,
                                                    HasPrimaryKey (..),
                                                    IsBootstrapEraAddr (..), IsHeader,
                                                    deriveFirstHDAddress,
+                                                   largestPubKeyAddressBoot,
+                                                   largestPubKeyAddressSingleKey,
                                                    makePubKeyAddress, siEpoch)
 import           Pos.Crypto                       (EncryptedSecretKey, PublicKey,
                                                    emptyPassphrase)
@@ -239,9 +241,15 @@ instance MonadKnownPeers AuxxMode where
 instance MonadFormatPeers AuxxMode where
     formatKnownPeers = OQ.Reader.formatKnownPeersReader (rmcOutboundQ . acRealModeContext)
 
-instance (HasConfiguration, HasInfraConfiguration) => MonadAddresses AuxxMode where
+instance (HasConfiguration, HasInfraConfiguration) =>
+         MonadAddresses AuxxMode where
     type AddrData AuxxMode = PublicKey
-    getNewAddress = withCompileInfo def $ makePubKeyAddressAuxx
+    getNewAddress = makePubKeyAddressAuxx
+    getFakeChangeAddress = do
+        epochIndex <- siEpoch <$> getCurrentSlotInaccurate
+        gsIsBootstrapEra epochIndex <&> \case
+            False -> largestPubKeyAddressBoot
+            True -> largestPubKeyAddressSingleKey
 
 type instance MempoolExt AuxxMode = EmptyMempoolExt
 
@@ -258,7 +266,7 @@ instance (HasConfigurations) =>
 -- choose suitable stake distribution. We want to pick it based on
 -- whether we are currently in bootstrap era.
 makePubKeyAddressAuxx ::
-       (HasConfiguration, HasInfraConfiguration, HasCompileInfo)
+       (HasConfiguration, HasInfraConfiguration)
     => PublicKey
     -> AuxxMode Address
 makePubKeyAddressAuxx pk = do
@@ -268,7 +276,7 @@ makePubKeyAddressAuxx pk = do
 
 -- | Similar to @makePubKeyAddressAuxx@ but create HD address.
 deriveHDAddressAuxx ::
-       (HasConfiguration, HasInfraConfiguration, HasCompileInfo)
+       (HasConfiguration, HasInfraConfiguration)
     => EncryptedSecretKey
     -> AuxxMode Address
 deriveHDAddressAuxx hdwSk = do
--- a/core/Pos/Core/Address.hs
+++ b/core/Pos/Core/Address.hs
@@ -41,6 +41,14 @@ module Pos.Core.Address
        , createHDAddressNH
        , createHDAddressH
 
+       -- * Maximal sizes (needed for tx creation)
+       , largestPubKeyAddressBoot
+       , maxPubKeyAddressSizeBoot
+       , largestPubKeyAddressSingleKey
+       , maxPubKeyAddressSizeSingleKey
+       , largestHDAddressBoot
+       , maxHDAddressSizeBoot
+
          -- * Internals
        , AddressHash
        , addressHash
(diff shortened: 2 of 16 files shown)


gromakovsky:
I don't know if it should go to cardano-sl-1.0, but maybe it should.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -128,6 +131,20 @@ instance
     getNewAddress (accId, passphrase) = do
         clientAddress <- L.newAddress RandomSeed passphrase accId
         decodeCTypeOrFail (cadId clientAddress)
+    -- We are creating a sample HD address here. It has the same
+    -- length (76) as the biggest one created by Daedalus.
+    getFakeChangeAddress = do
+        let passphrase = DBA.replicate 10 255
+        SecretKey sk <- snd <$> keyGen
+        encSK <- mkEncSecret passphrase sk
+        case deriveLvl2KeyPair
```
pva701: I don't think it's acceptable, `deriveLvl2KeyPair` adds redundant overhead for creating tx. also it can generate different size for different addresses because it depends on secret key

gromakovsky: > I don't think it's acceptable, deriveLvl2KeyPair adds redundant overhead for creating tx.

Do you mean performance overhead or what? I don't think it will be big overhead, tx creation has a lot of logic. And `deriveLvl2KeyPair` is almost always used once anyway.
Do you propose to just use hardcoded constant address?

> also it can generate different size of addresses because it depends on secret key

Yes, I realized it today already, will fix.

pva701: I realized that we do a lot of `deriveLvl2KeyPair` during building of a tx and one more won't add performance overhead

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -137,6 +142,23 @@ instance
...
+                 (IsBootstrapEraAddr True)
+                 (ShouldCheckPassphrase False)
+                 passphrase
+                 encSK
+                 maxBound
+                 maxBound of
+            Nothing -> throwM $ InternalError "getFakeChangeAddress failed"
+            Just (addr, _)
+                | biSize addr == maxHDAddressSizeBoot -> return addr
```
pva701: dunno why you don't want to use a hardcoded address with the max length, it will look safer. wouldn't you like to check that `biSize addr >= maxHDAddressSizeBoot`?

core/Pos/Core/Address.hs
```diff
@@ -372,3 +378,25 @@ isBootstrapEraDistrAddress (addrAttributesUnwrapped -> AddrAttributes {..}) =
...
+
+----------------------------------------------------------------------------
+-- Maximal size
+----------------------------------------------------------------------------
+
+-- | Maximal size of PubKey address with BootstrapEra
+-- distribution. Actual size depends on CRC32 value which is
+-- serialized using var-length encoding.
+maxPubKeyAddressSizeBoot :: Byte
+maxPubKeyAddressSizeBoot = 43
```
volhovM: How can one ensure these values make sense? Can you maybe add a short notice on how they were calculated, so anybody can validate numbers without applying too much effort?

lib/src/Pos/Generator/Block/Mode.hs
```diff
@@ -375,6 +379,19 @@ instance MonadBlockGenBase m => MonadBListener (BlockGenMode ext m) where
...
     getNewAddress = pure
+    getFakeChangeAddress = pure (go 0)
+      where
+        -- We want it to be deterministic. It must be consistent with
+        -- the way we construct address in block-gen. If it's changed,
+        -- tests will fail, so we will notice it.
+        go :: Word8 -> Address
+        go i =
+            let addr =
+                    makePubKeyAddressBoot . fst $
```
volhovM: Why do you consider only boot addresses here?

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
volhovM: AFAIU it can't be `>` since `maxHDAddressSizeBoot` is exactly what it says -- it's a _maximum_ size.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -137,6 +142,23 @@ instance
...
+        SecretKey sk <- snd <$> keyGen
+        encSK <- mkEncSecret passphrase sk
+        case deriveLvl2KeyPair
+                 (IsBootstrapEraAddr True)
+                 (ShouldCheckPassphrase False)
+                 passphrase
+                 encSK
+                 maxBound
+                 maxBound of
+            Nothing -> throwM $ InternalError "getFakeChangeAddress failed"
```
volhovM: It'd be good to specify which `getFakeChangeAddress` instance has failed exactly.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -137,6 +142,23 @@ instance
     getNewAddress (accId, passphrase) = do
         clientAddress <- L.newAddress RandomSeed passphrase accId
         decodeCTypeOrFail (cadId clientAddress)
+    -- We are creating a sample HD address here. It has the same
+    -- length (76) as the biggest one created by Daedalus.
+    getFakeChangeAddress = do
+        let passphrase = DBA.replicate 10 255
+        SecretKey sk <- snd <$> keyGen
+        encSK <- mkEncSecret passphrase sk
```
volhovM: Why didn't `emptyPassphrase` do it for you?

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
gromakovsky: If it's `>`, it means there is a bug. There is an issue to write tests for this function. In particular we want to test that it terminates. If at some point `maxHDAddressSizeBoot` stops being maximum, we will notice it by failing tests. But if it's `>=`, maybe we won't.

core/Pos/Core/Address.hs
gromakovsky: Hm, I think I should add tests for these values where addresses are generated from fixed values and their lengths are asserted.

lib/src/Pos/Generator/Block/Mode.hs
gromakovsky: Because currently block-gen only uses `makePubKeyAddressBoot` to generate addresses.
It's a bit fragile to rely on it here, but, as I wrote in the comment above, if it stops being true, tests will fail and we will notice it immediately.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
gromakovsky: Hm, I was trying hard to create an address with length 76 and was trying different passphrases in particular. I guess `emptyPassphrase` would work too, I will check.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -137,6 +142,23 @@ instance
     getNewAddress (accId, passphrase) = do
         clientAddress <- L.newAddress RandomSeed passphrase accId
         decodeCTypeOrFail (cadId clientAddress)
+    -- We are creating a sample HD address here. It has the same
+    -- length (76) as the biggest one created by Daedalus.
+    getFakeChangeAddress = do
```
pva701: I don't like this function is tail recursive. You suggest believing it will work and find an address quickly instead of using a predefined constant address with deterministic guarantees. There are already a lot of kostils in the wallet which we just believe.

I insist on rewriting it to deterministic manner.

lib/src/Pos/Generator/Block/Mode.hs
volhovM: Alright, this makes sense then.

auxx/src/Mode.hs
```diff
@@ -222,9 +224,15 @@ instance MonadKnownPeers AuxxMode where
...
     formatKnownPeers = OQ.Reader.formatKnownPeersReader (rmcOutboundQ . acRealModeContext)
 
-instance (HasConfiguration, HasInfraConfiguration) => MonadAddresses AuxxMode where
+instance (HasConfiguration, HasInfraConfiguration) =>
+         MonadAddresses AuxxMode where
     type AddrData AuxxMode = PublicKey
     getNewAddress = makePubKeyAddressAuxx
+    getFakeChangeAddress = do
+        epochIndex <- siEpoch <$> getCurrentSlotInaccurate
+        gsIsBootstrapEra epochIndex <&> \case
```
volhovM: 👍

lib/test/Test/Pos/Core/AddressSpec.hs
```diff
@@ -6,22 +6,91 @@ module Test.Pos.Core.AddressSpec
...
+            genPubKeyAddrBoot largestPubKeyAddressBoot 43
+
+        let genPubKeyAddrSingleKey = pure . makePubKeyAddress
+                (IsBootstrapEraAddr False) . toPublic
+        largestAddressProp "PubKey address with SingleKey distribution"
+            genPubKeyAddrSingleKey largestPubKeyAddressSingleKey 78
+
+        let genHDAddrBoot :: SecretKey -> Gen Address
+            genHDAddrBoot sk = frequency
+                [ (1, genHDAddrBootEmptyPass sk)
```
volhovM: Wow, TIL you can do forward-bindings in a single `let` block. It's non obvious though. Maybe it makes sense to move `getHDAddrBoot` after `genHDAddress*`?

lib/test/Test/Pos/Core/AddressSpec.hs
```diff
@@ -6,22 +6,91 @@ module Test.Pos.Core.AddressSpec
...
+
+largestAddressProp :: Text -> (SecretKey -> Gen Address) -> Address -> Byte -> Spec
+largestAddressProp addressDescription genAddress largestAddress expectedLargestSize = do
+    describe (toString addressDescription) $ do
+
+        it (formatToString ("has the size we expect: "%memory) expectedLargestSize) $ do
+            biSize largestAddress `shouldBe` expectedLargestSize
+
+        prop "is indeed the largest address" $ do
+            forAll (deterministicKeyGen . BS.pack <$> vectorOf 32 arbitrary) $ \(_, sk) ->
```
volhovM: Isn't there an `Arbitrary` instance for `pk/sk`?

gromakovsky: There is, but it takes a key from a predefined pool of keys. That pool consists of only 128 keys. Here I want to consider more keys.

PR #1746, opened 2017-10-10
Update application version and versions in .cabal files

gromakovsky:


gromakovsky:
Bumping cabal versions is generally good because we are delivering new version of software, but it also has more important impact besides being generally good, specifically, reports sent to reporting server contain version taken from .cabal files, so we can know which version produced particular report.

PR #1748, opened 2017-10-10
[CSL-1636] Remove Pos.Auxx.Hacks

int-index:
Use 'getCurrentSlotInaccurate' in 'makePubKeyAddressAuxx' to determine whether it's the bootstrap era at the moment.


int-index:
> I know that `listaddr` was deadlocking, did you figure out what was the reason?

I didn't investigate this, just changed `getCurrentSlotBlocking` to `getCurrentSlotInaccurate`.

> Also you reported that `getCurrentSlotInaccurate` was way too inaccurate, wasn't it a problem here?

Yes, back when I was implementing a Brick-based interface for `auxx`, I've seen `getCurrentSlotInaccurate` reporting highly inaccurate values (off by a few million slots). But it shouldn't be a problem with neither `unlockStakeEpoch = maxBound` (because all slots are in bootstrap era) nor `unlockStakeEpoch = 0` (because all slots are in reward era). I will test anyway, just to make sure.

int-index:
I've run `listaddr` with unknown current slot (so `getCurrentSlotInaccurate` had to guess) with different configurations (`unlockStakeEpoch = maxBound` and `unlockStakeEpoch = 0`). The addresses are different, as expected:

1. Bootstrap era addr: `Ae2tdPwUPEZ3Fd8HkQabvTJo3Ues7o2kNXXcK6LgGBfYwTM3pxpn5pijrBu
`
2. Reward era addr: `5FCjkr138i9vbi2TiprdLL1TSbQDD342QYf8NnZt6yZEJeADMgc9RAH48KpfegNa1cYcXFCVh4ezCFcfC56rembu8eCi5R3JKtf6yqGJLAY`

int-index:
Testing how `getCurrentSlotInaccurate` behaves with other `unlockStakeEpoch` values, and how accurate its guesses are, is better left for a separate issue, I think.

PR #1754, opened 2017-10-11
[CSL-1747] Various launcher improvements

neongreen:
This PR is supposed to go into cardano-sl-1.0. It contains the following changes:

* The launcher now has pretty logs with timestamps.
* The launcher can be configured to put logs into a folder. This is nice because now if the node doesn't start or can't be killed or something, we'll find out.
* On Linux and macOS, the launcher sends SIGKILL if the node doesn't respond to SIGTERM.

If this PR is merged, we will have to change the launcher command-line arguments in Daedalus:

* Set `--node-timeout` to 30 seconds instead of 5 seconds (sigkilling the node after 5 seconds is cruel)
* Set `--launcher-logs-prefix` appropriately (e.g. `%appdata%\Daedalus\logs\pub`)

### DIFF
--- a/tools/src/launcher/Main.hs
+++ b/tools/src/launcher/Main.hs
@@ -128,7 +128,8 @@ optionsParser = do
     -- Other args
     loNodeTimeoutSec <- option auto $
         long    "node-timeout" <>
-        help    "How much to wait for the node to exit before killing it." <>
+        help    "How much to wait for the node to exit before killing it \
+                \(and then how much to wait after that)."<>
         metavar "SEC"
     loReportServer <- optional $ strOption $
         long    "report-server" <>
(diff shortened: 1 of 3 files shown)


neongreen:
> It should go to cardano-sl-1.0 to be included into the next update.

I have changed the base of the PR to cardano-sl-1.0.

> AFAIK, node-timeout is set to something like 5 seconds, but sending SIGKILL after only 5 seconds is probably too tough. 

Okay, makes sense. Let's increase node-timeout to 30 seconds, I don't think anything bad will happen.

> Since this PR goes to 1.0, you should provide some description

Will do.

tools/src/launcher/Main.hs
```diff
@@ -240,91 +266,101 @@ main = do
...
              -- TODO: does the wallet have some kind of log?
-             putText "Killing the node"
-             liftIO $ do
-                 Process.terminateProcess nodeHandle
-                 cancel nodeAsync
+             killNode nodeHandle nodeAsync
+  where
+    killNode nodeHandle nodeAsync = do
+        logInfo "Killing the node"
+        liftIO $ Process.terminateProcess nodeHandle
```
gromakovsky: I want to have `tryAny` here and logging an exception. One hypothesis is that `terminateProcess` sometimes fails.

neongreen: Okay, I'll add it (I think that `terminateProcess` doesn't throw exceptions but, well, I'm not sure)

neongreen: @gromakovsky done, please rerereview.

gromakovsky: > I think that terminateProcess doesn't throw exceptions but, well, I'm not sure

I am not sure either, but I know that `TerminateProcess` from Win API can fail and given that `terminateProcess` returns `()`, I expect it to throw an exception in case of failure (`process` seems to be solid library).

gromakovsky:
FYI: CI failed.

PR #1755, opened 2017-10-11
[CSL-1554] Blockgen to auxx

volhovM:


lib/src/Pos/GState/Context.hs
```diff
@@ -29,15 +33,40 @@ data GStateContext = GStateContext
...
-    , _gscSlottingVar :: TVar SlottingData
+    , _gscSlottingVar :: SlottingVar
     -- Fields are lazy to be used with future.
     }
 
 makeClassy ''GStateContext
 
+--class HasGStateContext ctx where
+--    gStateContext :: Getter ctx GStateContext
+
```
gromakovsky: Shouldn't it be removed?

auxx/src/BlockGen.hs
```diff
@@ -0,0 +1,43 @@
...
+import           Pos.Wallet                  (getAllUserSecrets, getSecretKeys)
+
+import           Command.Types               (GenBlocksParams (..))
+import           Mode                        (AuxxMode)
+
+
+generateBlocks :: HasConfigurations => GenBlocksParams -> AuxxMode ()
+generateBlocks GenBlocksParams{..} = do
+    seed <- liftIO $ maybe randomIO pure bgoSeed
+    putText $ "Generating with seed " <> show seed
```
gromakovsky: I think using `logInfo` would be more reasonable here.

auxx/src/BlockGen.hs
```diff
@@ -0,0 +1,43 @@
...
+
+    allSecrets <- mkAllSecretsSimple . map encToSecret . getAllUserSecrets <$> getSecretKeys
```
gromakovsky: There is `getSecretKeysPlain`.

auxx/src/BlockGen.hs
```diff
@@ -0,0 +1,43 @@
...
+
+    let bgenParams =
+            BlockGenParams
+                { _bgpSecrets         = allSecrets
+                , _bgpGenStakeholders = gdBootStakeholders genesisData
+                , _bgpBlockCount      = fromIntegral bgoBlockN
+                , _bgpTxGenParams     = def
```
gromakovsky: Why is it `def`? It should be configurable.

auxx/src/BlockGen.hs
```diff
@@ -0,0 +1,43 @@
...
+                , _bgpInplaceDB       = True
+                , _bgpSkipNoKey       = True
+                }
+    void $ evalRandT (genBlocks bgenParams) (mkStdGen seed)
+    -- We print it twice because there can be a ton of logs and
+    -- you don't notice the first message.
+    putText $ "Generated with seed " <> show seed
```
gromakovsky: Here it also makes sense to use `logInfo` instead.

auxx/src/BlockGen.hs
volhovM: 1. It shouldn't be, task doesn't say anything about it
2. It's hard to make it configurable because parsing optionals is hard if you use parsec not optparse.

gromakovsky: The task says that you should move block-gen to auxx, _move_ means that you shouldn't throw-away existing features, they should be preserved.
The most important use case is to generate empty blocks, with `def` it's not achievable.

auxx/src/BlockGen.hs
```diff
@@ -0,0 +1,43 @@
+-- | Block generation.
+
+module BlockGen
```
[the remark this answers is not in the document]
volhovM: Moved.

auxx/src/Command/BlockGen.hs
```diff
@@ -0,0 +1,47 @@
...
+                , _bgpBlockCount      = fromIntegral bgoBlockN
+                -- tx generation is disalbed for now
+                , _bgpTxGenParams     = def & tgpTxCountRange .~ (0,0)
+                , _bgpInplaceDB       = True
+                , _bgpSkipNoKey       = True
+                }
+    withCompileInfo def $ void $ evalRandT (genBlocks bgenParams) (mkStdGen seed)
+    -- We print it twice because there can be a ton of logs and
+    -- you don't notice the first message.
+    putText $ "Generated with seed " <> show seed
```
gromakovsky: Is it intentional that first we print it using `logInfo` and then using `putText`?

docs/how-to/generate-blockchain.md
```diff
@@ -1,30 +1,19 @@
...
     --system-start 0                             \
     --configuration-file node/configuration.yaml \
     --configuration-key mainnet_full             \
-    --blocks 1000                                \
-    --generated-db some-db                       \
-    --secret PATH_TO_SECRET_KEY                  \
-    --tx-count "(0,0)"                           \
-    --append
+    --db-path some-db                            \
 ```
```
gromakovsky: What is it? AFAIK this command as-is won't work. You need to specify a list of command or pass `repl`.

docs/how-to/generate-blockchain.md
```diff
@@ -1,30 +1,19 @@
...
-*  `some-db` is a path to generated database.
-*  `PATH_TO_SECRET_KEY` is the path to a secret key we're using for generating blockchain.
-   This parameter can be specified **multiple** times. Please note that during Bootstrap era
-   we should use secret key Bootstrap stakeholders delegated heavily to.
-
-It is possible to use a custom seed for generating blocks. Use `--seed` parameter
-to define integer number that will be used as a seed, for example `--seed 100`.
-If the value of seed is not defined, random integer number will be used instead.
+Then ensure you have all the needed keys imported (see `add-key` and 
+`add-key-pool` commands). Next is running `generate-blocks <N> <seed>?` command, where
```
gromakovsky: This doc is mostly for DevOps, so `add-key-pool` doesn't need to be mentioned, I think.

docs/how-to/generate-blockchain.md
```diff
@@ -1,30 +1,19 @@
...
+`<N>` is number of blocks to generate and `<seed>` is optional generation seed.
```
gromakovsky: AFAIU, when tx payload generation is disabled, seed doesn't mean anything, so probably it's better not to mention it.

auxx/src/Command/BlockGen.hs
volhovM: Yes, that's exactly what comment says.

docs/how-to/generate-blockchain.md
volhovM: Let it stay, at least as a sanity check.

docs/how-to/generate-blockchain.md
volhovM: Alright.

gromakovsky:
So did you test it?

volhovM:
@gromakovsky yes, it does, I've just tested it in dev mode/1k blocks.

docs/how-to/generate-blockchain.md
```diff
@@ -1,30 +1,20 @@
...
-*  `1000` is a length of generated blockchain (number of blocks),
-*  `some-db` is a path to generated database.
-*  `PATH_TO_SECRET_KEY` is the path to a secret key we're using for generating blockchain.
-   This parameter can be specified **multiple** times. Please note that during Bootstrap era
-   we should use secret key Bootstrap stakeholders delegated heavily to.
-
-It is possible to use a custom seed for generating blocks. Use `--seed` parameter
-to define integer number that will be used as a seed, for example `--seed 100`.
-If the value of seed is not defined, random integer number will be used instead.
+Then ensure you have all the needed keys imported (see `add-key` and
```
gromakovsky: TODO: write that we need to add primary key after CSL-1765 is in master.

PR #1756, opened 2017-10-12
[CSE-218] Refactor Explorer's API

sectore:
to get rid of `catchExplorerError`. It's very similar as we have done with `wallet`'s API before (see https://github.com/input-output-hk/cardano-sl/blob/master/wallet/src/Pos/Wallet/Web/Api.hs).


explorer/src/Pos/Explorer/Web/Api.hs
```diff
@@ -18,85 +19,104 @@ module Pos.Explorer.Web.Api
...
+type ExplorerVerb verb = VerbMod ExplorerVerbTag verb
+
+-- | Shortcut for common api result types.
+type ExRes verbMethod a = ExplorerVerb (verbMethod '[JSON] a)
+
+instance ModifiesApiRes ExplorerVerbTag where
+    type ApiModifiedRes ExplorerVerbTag a = Either ExplorerError a
+    modifyApiResult _ =
+        try . catchEndpointErrors . (either throwM pure =<<)
+        where
```
ksaric: Indentation rules - https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md#indentation

explorer/src/Pos/Explorer/Web/Api.hs
```diff
@@ -18,85 +19,104 @@ module Pos.Explorer.Web.Api
...
+-- which allows to catch exceptions thrown by Explorer's endpoints.
+type ExplorerVerb verb = VerbMod ExplorerVerbTag verb
+
+-- | Shortcut for common api result types.
+type ExRes verbMethod a = ExplorerVerb (verbMethod '[JSON] a)
+
+instance ModifiesApiRes ExplorerVerbTag where
+    type ApiModifiedRes ExplorerVerbTag a = Either ExplorerError a
+    modifyApiResult _ =
+        try . catchEndpointErrors . (either throwM pure =<<)
```
ksaric: I prefer to have a parameter here than to have this chain and wonder what's _really_ going on.

explorer/src/Pos/Explorer/Web/Error.hs
```diff
@@ -2,17 +2,21 @@
...
+import           Control.Lens        (makePrisms)
 import qualified Data.Text.Buildable
 import           Formatting          (bprint, stext, (%))
 import           Universum
 
-data ExplorerError =
+newtype ExplorerError =
     -- | Some internal error.
-    Internal !Text
+    Internal Text
```
ksaric: If it's going to be used through `Lens`es, maybe put an underscore here and import it without the underscore.

explorer/src/Pos/Explorer/Web/Api.hs
```diff
@@ -105,22 +125,22 @@ type GenesisPagesTotal = API
...
     :> QueryParam "filter" CAddressesFilter
-    :> Get '[JSON] (Either ExplorerError PageNumber)
+    :> ExRes Get PageNumber
 
 type GenesisAddressInfo = API
     :> "genesis"
     :> "address"
     :> QueryParam "page" Word
     :> QueryParam "pageSize" Word
     :> QueryParam "filter" CAddressesFilter
```
Martoon-00: You can use `DQueryParam` from `Pos.Util.Servant` here. It will take `def` value if nothing is specified.

explorer/src/Pos/Explorer/Web/Api.hs
```diff
@@ -18,85 +19,104 @@ module Pos.Explorer.Web.Api
...
+-- | Shortcut for common api result types.
+type ExRes verbMethod a = ExplorerVerb (verbMethod '[JSON] a)
+
+instance ModifiesApiRes ExplorerVerbTag where
+    type ApiModifiedRes ExplorerVerbTag a = Either ExplorerError a
+    modifyApiResult _ =
+        try . catchEndpointErrors . (either throwM pure =<<)
+        where
+            catchEndpointErrors :: (MonadCatch m) => m a -> m (Either ExplorerError a)
+            catchEndpointErrors = tryJust $ \e -> (e ^? _Internal) $> e
```
Martoon-00: Here there is no need to do so since you have only one constructor of `ExplorerError`
Just `try` should suffice

explorer/src/documentation/Main.hs
```diff
@@ -75,6 +76,9 @@ main = do
 instance HasSwagger api => HasSwagger (MultipartForm a :> api) where
     toSwagger Proxy = toSwagger $ Proxy @api
 
+instance HasSwagger v => HasSwagger (VerbMod mod v) where
+    toSwagger _ = toSwagger $ Proxy @v
```
Martoon-00: Don't you mind moving it to `Pos.Util.Servant`?
We already have similar instance for wallet, but I couldn't perform this movement earlier, because at that time adding even a single dependency to subproject time-outed CI

explorer/src/Pos/Explorer/Web/Error.hs
Martoon-00: As for me, it looks like departing from the usual workflow ; -)

explorer/src/Pos/Explorer/Web/Api.hs
Martoon-00: Oh, looks like there is no such function yet. But you can add one here:
https://github.com/input-output-hk/cardano-sl/blob/master/lib/src/Pos/Util/Servant.hs#L512

explorer/src/Pos/Explorer/Web/Error.hs
sectore: I don't know the rules, but with `WalletError` (and others) we are doing the same thing (w/o underscore for data constructors) https://github.com/input-output-hk/cardano-sl/blob/master/lib/src/Pos/Wallet/Web/Error/Types.hs#L16-L27

explorer/src/Pos/Explorer/Web/Api.hs
Martoon-00: Would do you mean under parameter here?

sectore: ```modifyApiResult _ = try . catchEndpointErrors . (either throwM pure =<<)```
is pointfree style of following:
```modifyApiResult proxy action = try . try $ (either throwM pure =<<) action```.

We can't set a parameter `proxy`, because it is not used at the right hand side of `modifyApiResult`. We will get a warning instead: 
```
cardano-sl/explorer/src/Pos/Explorer/Web/Api.hs:57:21: warning: [-Wunused-matches]
    Defined but not used: ‘proxy’ 
```

But we could use 
```modifyApiResult _ action = try . try $ (either throwM pure =<<) action``` 
to avoid to have a pointfree style. 

@ksaric Would the last version (not pointree) of `modifyApiResult` ^ fix your concerns?

Comment (ksaric) [reply explorer/src/Pos/Explorer/Web/Api.hs]: @sectore Yes, I presume we don't need `proxy` here.
It's better, but I still don't get why we have 2 `try`s here. And `pure bind` (`pure =<<`).
Why not `try (either throwM action)`? Isn't that sufficient? I still haven't compiled this locally, so I'm running blind here. Nonetheless, it looks kind of boilerplate-y.

Comment (sectore) [reply explorer/src/Pos/Explorer/Web/Api.hs]: @ksaric Yeah, good to have these questions! Because it helps me to understand everything more and more, too :) 
```
try (either throwM pure action)
```
will throw
```
Couldn't match expected type ‘Either e0 (Either ExplorerError a)’
                  with actual type ‘IO (Either ServantErr a)’
    • In the third argument of ‘either’, namely ‘action’
      In the first argument of ‘try’, namely
        ‘(either throwM pure action)’
      In the expression: try (either throwM pure action)
...
``` 
Agree, the best thing is to play around to answer your questions by `ghc`.

Another questions for me was: Why the hell do we need to add `Proxy ExplorerVerbTag`? It's because type families are not injective (we are using type families of Servant) and `ghc` won't be able to know about `ExplorerVerbTag`. (see https://arow.info/blog/posts/2015-07-10-servant-intro.html#why-pass-layout-twice BTW: This article helps me a lot to understand what's really going on here).

ksaric: I will take a look at this `Either e0 (Either ExplorerError a)` thing tomorrow, thanks.

Martoon-00: Yep, two `try`s there because we want to return `Either ServantErr (Either ExplorerError a)`. First exception is servant-required, and second one is what we want to return.

Martoon-00: Btw, in wallet code we also cough _all_ other exceptions here and converted them to `ServantErr` or `WalletError`, because otherwise on unexpected error we just got `500 Internal error` without any explanations of what's exactly going on; maybe it also occurred to you to suffer from it )

PR #1757, opened 2017-10-12
[CSL-1760] Document configuration and genesis + clean-up

gromakovsky:
# Document configuration and genesis + clean-up

In this PR I:
1. Made some changes in code relevant to genesis generation. They only matter for testnet genesis, so shouldn't break anything.
2. Deleted some obsolete/useless/duplicated files and stuff.
3. Added `docs/configuration.md` which describes configuration.
4. Figured out how to setup internal staging cluster, provided configuration template and instructions.

Now again, step-by-step.

## Changes in code

1. Maximal balance for `testnetInitializer` is now computed correctly, considering `avvmBalanceFactor`.
2. `avvmBalanceFactor` no longer applies to fake avvm balances, because I find it more intuitive.
3. `genPrimaryKey` command now prints `StakeholderId`.
4. `genTestnetDistribution` now doesn't divide by zero when there are no richmen.
5. A log message that genesis generation can take a lot of time is printed when we generate genesis data.

## Clean-up

1. Deleted `secrets/` folder, it's absolutely useless after genesis rework.
2. Deleted `scripts/generate/genesis.sh`, it also was absolutely useless.
3. Deleted `node/mainnet-genesis.json`, it was duplicated.
4. Deleted `mainnet-genesis-stub.json`, it was useless, we don't need stubs now, we have actual genesis data.
5. Deleted `Configs for mainnet launch tooling` section from the configuration file, it's not needed after mainnet is deployed.
6. Deleted testnet configurations, they didn't make sense anyway.
7. Deleted useless `mainnet_example_generated` configuration.

## Configuration.md

My primary focus was on describing genesis-related parts and other things necessary to deploy internal staging and later testnet. So there are some TODOs. I believe they can be done later (in another PR).

Described things include:

* What is configuration, how it's specified (CLI and format).
* Details description of genesis:
  - What it is. Format, meaning of fields.
  - How it's specified.
  - Related tools.
  - How to generated it for testnet-like clusters.
* Partially described other things from configuration (not much, see TODOs).
* Which configurations we have in the repository.

Remaining TODOs:
* Describe parts of `core` configuration other than genesis, `infra` configuration, `update` configuration, `ssc` configuration and finish describing `node` configuration.
* Do something with `Generating genesis for mainnet` chapter and old documentation on this topic.

## Internal staging

As per @georgeee we are going to have 5 clusters running: internal staging (unstable), testnet staging, testnet, mainnet staging, mainnet. Currently we want to have internal staging and it should as close as possible to what we will later deploy for testnet. I added necessary configurations for `internal_staging` and described the workflow.

I tested the described workflow locally and it worked.


docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+
+* One way is to provide genesis data itself, stored in json
+  format. Genesis data explicitly states balances of all addresses,
+  all bootstrap stakeholders, heavyweight delegation certificates,
+  etc. It makes it clearer which genesis is used, but it doesn't
+  describe _how_ this genesis was constructed. Also it requires more
+  steps to create genesis data, than to create a spec. Current design
+  implies that this mechanism should be used only for mainnet where we
+  want to have full understanding of which addresses have which
+  stake. The format is demonstrated below. `hash` is the hash of
```
avieth: Should mention the hashing algorithm.

gromakovsky:
It will be finished after I get requirements about testnet.

domenkozar:
DevOps team will review the docs this week.

deepfire:
I'm on it.

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+  format. Genesis data explicitly states balances of all addresses,
+  all bootstrap stakeholders, heavyweight delegation certificates,
+  etc. It makes it clearer which genesis is used, but it doesn't
+  describe _how_ this genesis was constructed. Also it requires more
+  steps to create genesis data, than to create a spec. Current design
+  implies that this mechanism should be used only for mainnet where we
+  want to have full understanding of which addresses have which
+  stake. The format is demonstrated below. `hash` is the hash (Blake2b
+  256 bits) of canonically encoded `mainnet-genesis.json` file. It can
+  be computed manually using `scripts/js/genesis-hash.js` script.
```
deepfire: I wonder if there's consensus to drop the JS code and switch to a single Haskell code/tool for this hash.

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+        hash: 5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb
+```
+
+* Another way is to provide a specification how to generate genesis
+  data. It allows one to specify how many nodes should have stake,
+  which bootstrap stakeholders should be used, what should be total
+  balance, etc. When node is launched with spec as genesis, it
+  automatically constructs genesis data from it. This way to specify
+  genesis is supposed to be used for everything except
+  mainnet. Example:
```
deepfire: ..and also anything production-y, like `staging` and `testnet`, right?

Can't see how we could want to use auto-genesis for those, since we want to maximise similarity to mainnet.

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+          unlockStakeEpoch: 18446744073709551615 # last epoch (maxBound @Word64)
+        protocolConstants: &dev_core_genesis_spec_protocolConstants
+          k: 2
+          protocolMagic: 55550001
+          vssMinTTL: 2
+          vssMaxTTL: 6
+        ftsSeed: "c2tvdm9yb2RhIEdndXJkYSBib3JvZGEgcHJvdm9kYSA="
+        heavyDelegation: {}
+        avvmDistr: {}
+```
```
deepfire: This makes me wonder.

We probably want to keep a close watch on the exact list of differences between the auto-genesis and the `mainnet` one.  This list should not contain any frivolous differences, lest we endanger ourselves with self-induced blindness..

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+`testnetInitializer` was provided above. The most interesting part
+there is `distribution` field:
+
+```
+  distribution:
+   testnetRichmenStakeDistr: []
+```
+
+It means that generated richmen will also be bootstrap stakeholders
+and will be participating in SSC. Note that richmen's keys are
```
deepfire: The reference to SSC begs a reference to the background documentation, where one could self-educate.

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+        ftsSeed: "c2tvdm9yb2RhIEdndXJkYSBib3JvZGEgcHJvdm9kYSA="
+        heavyDelegation: {}
+        avvmDistr: {}
+```
+
+### Initializer
+
+There are two substantially different types of genesis spec, with the
+difference in `initializer` value. Initializer can be either
+`testnetInitializer` or `mainnetInitializer`.
```
deepfire: Again, as per a previous comment -- is the key auto-genesis convenience worth the dangers of divergence between `mainnet` and `testnet`? 

Answer to this question could have non-trivial consequences on the usefulness of `testnet` for the purpose or prototype development and testing.  cc @dcoutts

docs/configuration.md
gromakovsky: For staging – yes.
For testnet – we didn't plan to do it, because we need significantly different genesis for testnet anyway, so this similarity shouldn't really matter. However, at least for the first testnet we will be using the same way to specify genesis as for mainnet for performance reasons.
It's not yet described in the document, I am waiting for @georgeee to finalize the requirements (I was told about them only on Monday in the evening and only briefly).

docs/configuration.md
gromakovsky: I am not sure we want to care much about differences in mainnet genesis and testnet genesis.
From what I know, testnet genesis will be significantly different from the mainnet one (for example, all AVVM balances will be divided by 10). I am not aware of the goal to have as similar genesises as possible.

docs/configuration.md
gromakovsky: I am not sure if we have any (up-to-date) documentation, I will add a link if I find it.

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+### `devnet` configuration
+
+There is a configuration called `devnet` which should be used to setup
+clusters for developers or QA to test something. There are some values
+which should be set carefully before using this configuration, they
+depend on particular task:
+
+* `slotDuration`, `k` are parameters of the protocol, set to
+  reasonable defaults, but sometimes we may want to use different
+  values.
```
deepfire: One thing that I had to look up in the source code is the relationship between `k` and the number of slots in an epoch.

I'm sure it must have been documented somewhere, but I couldn't find it back then, and I think it would help if this was clarified here.

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+* `networkDiameter`, `mdNoBlocksSlotThreshold`,
+  `recoveryHeadersMessage` depend on `k`, comments are provided in
+  configuration file.
```
deepfire: I agree these probably aren't central, buf if this file _is_ documentation for the configuration, shouldn't they be explained here, primarily?

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+generate-key --path <SECRET_KEY_PATH>`. Generate as many keys as there
+should be core nodes. Last line looks like this:
+
+> [keygen:INFO] Successfully generated primary key and dumped to secrets/testnet/node4.sk, stakeholder id: 39f2bb9fd75ac348e6c92467e61eb3e1418a394f5a2105f9e014b666, PK (base64): /fbQqCqUPdPImYA2djkGYMpj9HZGkDeKhTc/mLcPG7sdhJSR6Ou0sB3J5OII2VdWTXSHRM1cPdNwZsaNba4rcg==
+
+You should use stakeholder ids as keys in `bootStakeholders`
+map. Values are stakeholders' weights, having all weights equal to 1
+is fine.
+
+The second step is VSS certificates generation. Use `cardano-keygen
```
deepfire: What are VSS certificates?  Again, a pointer to background documentation would be helpful here.

docs/configuration.md
```diff
@@ -0,0 +1,297 @@
...
+
+```
+  distribution:
+    testnetCustomStakeDistr:
+      bootStakeholders:
+        75fc2050ea497eb615461d01170679d143fb70c07c0baf3db54c6237: 1
+        eaba957b871c4d5d9f5eab1a4183a037b3b0a59e31a52185d37627f1: 1
+      vssCerts:
+        75fc2050ea497eb615461d01170679d143fb70c07c0baf3db54c6237:
+          expiryEpoch: 5
```
deepfire: What does `expiryEpoch` mean?

docs/configuration.md
jmitchell: > we need significantly different genesis for testnet anyway

I'm interested in hearing more about this when the info is available. I share Serge's intuition that testnet, for the most part, should be as similar to mainnet's configuration as possible.

However, as long as the initial genesis data resulting from the spec is sufficiently similar to mainnet's JSON in terms of constants and such, I'm not terribly concerned about automating it. We would just need a way to verify what initial genesis data are used.

docs/configuration.md
jmitchell: Personally I'm not so concerned about the initial accounts and balances being completely similar to mainnet, but any deviation in the protocol constants should be clearly called out and justified.

docs/configuration.md
jmitchell: Perhaps including "Shared Seed Computation" and linking to https://cardanodocs.com/technical/blocks/. I didn't recall what SSC referred to either.

docs/configuration.md
jmitchell: It's described at https://github.com/input-output-hk/cardano-sl/tree/cardano-sl-1.0/scripts/prepare-genesis#32-genesis-json-format-explanation as:

> index of epoch when this certificate will expire

@gromakovsky, judiciously linking to parts of the existing genesis generation doc or deprecating it and integrating any other context it provides into this doc may help.

docs/configuration.md
jmitchell: Agreed. Recent experiences tell me this is 10 on mainnet, but I'm unclear whether that's hardcoded somewhere or in any way configurable.

jmitchell: Ah, it's fixed as `10 * k`. See https://cardanodocs.com/technical/. It also links to https://github.com/input-output-hk/cardano-sl/blob/9ee12d3cc9ca0c8ad95f3031518a4a7acdcffc56/core/Pos/Core/Constants/Raw.hs#L161.

docs/configuration.md
gromakovsky: Clarification: when I said _significantly different_, I meant balances (it wasn't clear indeed).
Almost all other constants should be the same (except e. g. protocol magic and maybe 1-2 more constants).

docs/configuration.md
gromakovsky: Initializer determines balances, stakes and bootstrap stakeholders. In testnet we need differences balances. For example, we need few thousands of addresses with some balance, apart from AVVM output. In mainnet we have only AVVM output (and 3 more addresses).
This difference between mainnet and testnet is intentional.

docs/configuration.md
gromakovsky: Added.

docs/configuration.md
gromakovsky: Added `Prerequisites` chapter in the beginning with the relevant link.

docs/configuration.md
gromakovsky: Ok, I described these values (also I realized that `networkDiameter` doesn't depend on `k` and removed it from this line, but described it anyway), but in general the issue is mostly describing genesis configuration, so there will be some TODOs left for new issues (describing genesis configuration is more important now).

docs/configuration.md
deepfire: cc @domenkozar -- based on the DevOps standup review.

docs/configuration.md
deepfire: cc @domenkozar, based on DevOps review.

docs/configuration.md
deepfire: @gromakovsky, would you mind adding this reference -- and, perhaps a bit expand on its significance in this particular context?

docs/configuration.md
deepfire: @gromakovsky, bump.

docs/configuration.md
gromakovsky: > This would entail maintenance of a documented list of these differences, naturally.

I added lists of differences to `Our configurations` for long-term clusters (2 mainnet stagings and internal staging).

docs/configuration.md
gromakovsky: I don't think that link is a good source of information. Instead, I wrote a short explanation in Prerequisites:
```
SSC is Shared Seed Computation, it's performed by rich nodes (with big stake) to agree upon the seed which will be used for leaders selection for an epoch.
```

docs/configuration.md
gromakovsky: See `Genesis data format` section, I described it there (basically copied from the old doc).

As I wrote in the first message in this PR, I suggest that other activities related to the old documentation (e. g. merging it with this one) will be performed later, because it's not crucial now.

docs/configuration.md
dcoutts: We can certainly have a Haskell tool if that makes things easier, it'd only be about 5 lines long too.

George does have a point that for transparency it is useful to keep an obviously independent implementation for others to use to verify our genesis block, but that doesn't mean we need to use it in our workflow if it is not convenient.

gromakovsky:
@georgeee 
1. I rebased on latest master, resolved conflicts.
2. I removed `mainnetInitializer` altogether. Also removed mainnet_generated_example configuration key.
3. I restored `mainnet_base` configuration description, because it still exists and people may wonder what it means.

I think it's ready for another review.

core/Pos/Core/Genesis/Types.hs
```diff
@@ -129,64 +128,46 @@ instance Buildable FakeAvvmOptions where
...
-                     "  vss certificates: "%build%"\n"%
-                     "  non-avvm balances: "%build%"\n"%
-                     "}\n"
-                    )
-                    miStartTime
-                    miBootStakeholders
-                    miVssCerts
-                    miNonAvvmBalances
+    build GenesisInitializer {..} = bprint
+        ("TestnetInitializer {\n"%
```
volhovm: s/Testnet/Genesis/g

core/Pos/Core/Genesis/Types.hs
```diff
@@ -129,64 +128,46 @@ instance Buildable FakeAvvmOptions where
 
 -- | This data type contains various options presense of which depends
 -- on whether we want genesis for mainnet or testnet.
```
volhovm: Updating this comment may be a good idea too.

core/Pos/Core/Genesis/Types.hs
gromakovsky: Fixed.

core/Pos/Core/Genesis/Types.hs
gromakovsky: Fixed.

gromakovsky:
>  Demo doesn't work.

No.

docs/configuration.md
```diff
@@ -0,0 +1,696 @@
...
+  total balance and whether to use HD addresses or simple PubKey
+  addresses. Assignment of balances and stakes is described in
+  detail [below](#balances-and-stakes).
+* `fakeAvvmBalance`. It contains number of fake AVVM seeds and how
+  balance of each seed. These seeds can be used to redeem ADA.
+* `avvmBalanceFactor`. Real AVVM balances will be multiplied by this
+  factor. Fake AVVM balances are not affected.
+* `useHeavyDlg`. If it's `True`, then for each bootstrap stakeholders
+  there will be a heavyweight delegation certificate from this
+  stakeholder to a core node. Otherwise bootstrap stakeholders will
```
volhovm: "core node" here is ambiguous and it wasn't defined previously in the document. I suggest to use "rich node" instead.

docs/configuration.md
```diff
@@ -0,0 +1,696 @@
...
+
+In _genesis spec_ there are two fields which determine balances and
+stakes: `avvmDistr` and `initializer`. `avvmDistr` is supposed to
+contain the result of vending (which can be found in
+`lib/configuration.yaml`, see `mainnet_avvmDistr`), but it also can be
+empty if you don't need AVVM addresses (or it can be modified in
+a different way, however you want). Overview of `initializer` is
+described [above](#initializer).
+
+In the following we describe how
```
volhovm: In the following ... paragraph i suppose.

docs/configuration.md
```diff
@@ -0,0 +1,696 @@
...
+
+* `cardano-node-simple` and `cardano-node` also have command line option to
+  dump genesis data. The option is
+  `--dump-genesis-data-to genesis.json` (data will be dumped to
+  `genesis.json`).
+
+### Generating genesis for testnet
+
+For testnet we can't use `spec` for genesis, at least because
+converting genesis spec to genesis data is currently very slow and we
```
volhovm: It's not really obvious how these statements are related (you use "and"). I suppose that hd address generation is a bottleneck?

docs/configuration.md
```diff
@@ -0,0 +1,696 @@
...
+
+* One way is to provide genesis data itself, stored in JSON
+  format. Genesis data explicitly states balances of all addresses,
+  all bootstrap stakeholders, heavyweight delegation certificates,
+  etc. It makes it clearer which genesis is used, but it doesn't
+  describe _how_ this genesis was constructed. Also it requires more
+  steps to create genesis data, than to create a spec.  This mechanism
+  is used for mainnet (both production and staging) so that everyone
+  can open this file and see all balances (and everything else from
+  genesis data) at any time. It also will be used for testnet, at
```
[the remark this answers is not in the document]
gromakovsky: Why?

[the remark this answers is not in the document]
volhovm: Because we had like 6 testnets before the mainnet, i suppose. Which testnet are you talking about?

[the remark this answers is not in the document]
gromakovsky: You can forget about those testnets, they are history. I am talking about normal testnet which will be launched in few weeks/months.

docs/configuration.md
gromakovsky: Apparently this phrase is acceptable too, but I missed the comma https://ell.stackexchange.com/a/94858

PR #1758, opened 2017-10-12
[CSL-1767] fix block generator leak

avieth:
`catMaybes <$> mapM genBlock [startEOS .. finishEOS]` caused the entire list of blocks to be retained.

Instead of constructing the list of blocks, some value of a monoid `t` is strictly constructed according to some injector `Maybe (Blund SscGodTossing) -> t`. Using `const ()` means the blocks will never be retained. `genBlock` writes them out to disk, which is all we want for the block generator application.

### DIFF
--- a/lib/src/Pos/Generator/Block/Logic.hs
+++ b/lib/src/Pos/Generator/Block/Logic.hs
@@ -41,7 +41,6 @@ import           Pos.Lrc.Context             (lrcActionOnEpochReason)
 import qualified Pos.Lrc.DB                  as LrcDB
 import           Pos.Txp                     (MempoolExt, MonadTxpLocal,
                                               TxpGlobalSettings)
-import           Pos.Util.Chrono             (OldestFirst (..))
 import           Pos.Util.CompileInfo        (HasCompileInfo, withCompileInfo)
 import           Pos.Util.Util               (HasLens', maybeThrow, _neHead)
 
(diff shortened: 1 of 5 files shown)


lib/src/Pos/Generator/BlockEvent.hs
```diff
@@ -190,7 +190,10 @@ genBlocksInTree secrets bootStakeholders blockchainTree = do
             }
     -- Partial pattern-matching is safe because we specify
     -- blockCount = 1 in the generation parameters.
-    OldestFirst [block] <- genBlocks blockGenParams
+    -- FIXME avieth what if it's 'Nothing'? Then the list is empty.
```
gromakovsky: Look at the comment above, it explains why pattern-matching is safe.

avieth:
> changed the title from [CSL-1617] fix block generator leak to [CSL-1767] fix block generator leak

Thanks :) 1617 is apparently ingrained into my mind

lib/src/Pos/Generator/BlockEvent.hs
avieth: `genBlock` can give `Nothing` (haven't looked into why).

gromakovsky: Ah, I see, yes, it was changed to `Nothing` after this code and comment were written.

avieth: If there's no secret for the leader of a given slot then `Nothing` is given. Maybe that should be an error and the whole thing should fail?

gromakovsky:
Rebase on top of latest master to fix stack2nix build.

lib/src/Pos/Generator/BlockEvent.hs
```diff
@@ -190,8 +190,10 @@ genBlocksInTree secrets bootStakeholders blockchainTree = do
             , _bgpTxpGlobalSettings = txpSettings
             }
     -- Partial pattern-matching is safe because we specify
-    -- blockCount = 1 in the generation parameters.
-    OldestFirst [block] <- genBlocks blockGenParams
+    -- blockCount = 1 and skipNoKey = False in the generation parameters.
+    let inj Nothing   = []
```
int-index: Why redefine `maybeToList`?

lib/src/Test/Pos/Block/Logic/Util.hs
```diff
@@ -92,7 +92,10 @@ bpGenBlocks
 bpGenBlocks blkCnt enableTxPayload inplaceDB = do
     params <- genBlockGenParams blkCnt enableTxPayload inplaceDB
     g <- pick $ MkGen $ \qc _ -> qc
-    lift $ evalRandT (genBlocks params) g
+    -- Injector for genBlocks, so that it retains all of the blunds in a list.
+    let inj Nothing   = []
```
int-index: Why redefine `maybeToList`?

lib/src/Pos/Generator/Block/Logic.hs
```diff
@@ -60,22 +59,32 @@ type BlockTxpGenMode g ctx m =
...
+        let numberOfBlocks = params ^. bgpBlockCount
+        tipEOS <- getEpochOrSlot <$> lift getTipHeader
+        let startEOS = succ tipEOS
+        let finishEOS = toEnum $ fromEnum tipEOS + fromIntegral numberOfBlocks
+        foldM' genOneBlock mempty [startEOS .. finishEOS]
+
+    genOneBlock t eos = ((t <>) . inj) <$> genBlock eos
+
+    foldM' _ !base [] = return base
+    foldM' combine !base (x : xs) = do
```
int-index: What about applying a worker-wrapper transformation here, so that `combine` is not passed at each step? Could be good for both perfomance and clarity.

```
foldM' combine = go
  where
    go !base [] = return base
    go !base (x:xs) = do
      it <- combine base x
      go it xs
```

avieth:
@int-index I've implemented your suggestions.

PR #1766, opened 2017-10-13
[CSL-1769] Use log-warper 1.3.3

int-index:
Since @volhovM told me upgrading to the latest `log-warper` would fix the issue, that's what I did. Well, the good news is that I finished the upgrade, the bad news is that it didn't fix the issue.

### DIFF
--- a/explorer/src/Pos/Explorer/Socket/Holder.hs
+++ b/explorer/src/Pos/Explorer/Socket/Holder.hs
@@ -27,16 +27,16 @@ module Pos.Explorer.Socket.Holder
        , ccConnection
        ) where
 
-import           Control.Lens             (makeClassy)
-import qualified Data.Map.Strict          as M
-import qualified Data.Set                 as S
-import           Network.EngineIO         (SocketId)
-import           Network.SocketIO         (Socket)
-import           Serokell.Util.Concurrent (modifyTVarS)
-import           System.Wlog              (NamedPureLogger, WithLogger,
-                                           launchNamedPureLog)
+import           Control.Lens     (makeClassy)
+import qualified Data.Map.Strict  as M
+import qualified Data.Set         as S
+import           Network.EngineIO (SocketId)
+import           Network.SocketIO (Socket)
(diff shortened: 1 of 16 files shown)


explorer/src/Pos/Explorer/Socket/Holder.hs
```diff
@@ -83,9 +83,16 @@ mkConnectionsState =
 withConnState
     :: (MonadIO m, WithLogger m)
     => ConnectionsVar
-    -> NamedPureLogger (StateT ConnectionsState STM) a
+    -> StateT ConnectionsState (NamedPureLogger STM) a
     -> m a
-withConnState var = launchNamedPureLog $ atomically . modifyTVarS var
+withConnState var m = do
```
volhovM: mda

int-index: life is tough

int-index:
Alright, https://github.com/serokell/log-warper/pull/27 fixes it! Thank you, @volhovM

avieth:
No more stdout logging in csl master? Is this patch responsible for that? How can I get it back? @int-index @volhovM

int-index:
The patch should've fixed the flag that disables logging. I only tested that logging is indeed disabled it when lcConsoleOutput is set to False, but not that logging still works when it's set to True...

avieth:
I put `printOutput: true` in the top level of the configuration file but still only warnings and errors are printed.

avieth:
```
topLevelParams :: [Text]
topLevelParams =
    ["rotation", "showTime", "printOutput", "filePrefix" ]
```

I think I need to set `termSeverity` to `Debug`, but I can't because it's not in the `topLevelParams`.

int-index:
I created https://issues.serokell.io/issue/CSL-1778 to track this.

int-index:
> I think I need to set `termSeverity` to `Debug`

Right, @volhovM told me the same, so I suppose that's what we should do.

PR #1768, opened 2017-10-14
[CSL-1770] Integrate report server 0.4

volhovM:
I've also decided to put git revision into `rBuild` report field so it required me to pass `HasCompileInfo` in several places. Tests and block-gen use fake compile info (default instance) as they never report anything (though using methods that can report).

### DIFF
--- a/lib/src/Pos/Generator/Block/Logic.hs
+++ b/lib/src/Pos/Generator/Block/Logic.hs
@@ -18,7 +18,7 @@ import           System.Random               (RandomGen (..))
 import           System.Wlog                 (logWarning)
 
 import           Pos.AllSecrets              (HasAllSecrets (..), unInvSecretsMap)
-import           Pos.Block.Core              (mkGenesisBlock)
+import           Pos.Block.Core              (Block, mkGenesisBlock)
 import           Pos.Block.Logic             (applyBlocksUnsafe, createMainBlockInternal,
                                               normalizeMempool, verifyBlocksPrefix)
 import           Pos.Block.Slog              (ShouldCallBListener (..))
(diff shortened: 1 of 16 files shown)


lib/src/Pos/Generator/Block/Logic.hs
```diff
@@ -117,10 +118,12 @@ genBlock eos = do
                     Just <$> usingPrimaryKey leaderSK (genMainBlock slot (Right . swap <$> transCert))
   where
     genMainBlock slot proxySkInfo =
+        withCompileInfo def $
         lift $ createMainBlockInternal @SscGodTossing slot proxySkInfo >>= \case
             Left err -> throwM (BGFailedToCreate err)
             Right mainBlock -> verifyAndApply $ Right mainBlock
     verifyAndApply block =
+        withCompileInfo def $
```
gromakovsky: Looks strange to have 3 `withCompileInfo` in `genBlock`. Maybe you need to provide type signatures explicitly.

volhovM: I think this difference is mostly cosmetical, but i can do it, sure.

gromakovsky: The problem is that if `withCompileInfo` is applied to the same code multiple times with different values, something bad will happen, AFAIK (and it won't be prevented by compiler).
So having less `withCompileInfo` makes reasoning easier.
Later someone may change `def` to something else in one place and forget about another one.

volhovM: In general -- yes. This particular place -- no. But anyway, I agree.

PR #1769, opened 2017-10-15
[CSL-744] Haskeline-based REPL

int-index:
it's ready for review, but I'm going to add some documentation a bit later (perhaps tomorrow)

The general idea is that Haskeline UI is running in a separate thread and communicates with `auxxPlugin` via a couple of `MVar`s. The logic of communication is neatly encapsulated in `createAuxxRepl` and the rest of the code doesn't need to deal `MVar`s.

### DIFF
--- a/auxx/src/Main.hs
+++ b/auxx/src/Main.hs
@@ -15,8 +15,8 @@ import qualified Pos.Client.CLI        as CLI
 import           Pos.Communication     (OutSpecs, WorkerSpec)
 import           Pos.Core              (Timestamp (..), gdStartTime, genesisData)
 import           Pos.Launcher          (HasConfigurations, NodeParams (..), NodeResources,
-                                        bracketNodeResources, loggerBracket, runNode,
-                                        runRealBasedMode, withConfigurations)
+                                        bracketNodeResources, loggerBracket, lpConsoleLog,
+                                        runNode, runRealBasedMode, withConfigurations)
 import           Pos.Network.Types     (NetworkConfig (..), Topology (..),
                                         topologyDequeuePolicy, topologyEnqueuePolicy,
                                         topologyFailurePolicy)
@@ -26,10 +26,11 @@ import           Pos.Util.CompileInfo  (HasCompileInfo, retrieveCompileTimeInfo,
 import           Pos.Util.UserSecret   (usVss)
 import           Pos.WorkMode          (EmptyMempoolExt, RealMode)
 
-import           AuxxOptions           (AuxxOptions (..), getAuxxOptions)
+import           AuxxOptions           (AuxxAction (..), AuxxOptions (..), getAuxxOptions)
 import           Mode                  (AuxxContext (..), AuxxMode, AuxxSscType,
                                         CmdCtx (..), realModeToAuxx)
 import           Plugin                (auxxPlugin)
+import           Repl                  (WithCommandAction, withAuxxRepl)
 
 -- 'NodeParams' obtained using 'CLI.getNodeParams' are not perfect for
 -- Auxx, so we need to adapt them slightly.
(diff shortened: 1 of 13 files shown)


auxx/src/Plugin.hs
```diff
@@ -43,52 +44,30 @@ import           Mode                       (AuxxMode)
...
+    -> Either (WithCommandAction AuxxMode) Text
     -> (WorkerSpec AuxxMode, OutSpecs)
-auxxPlugin AuxxOptions {..} =
-    case aoAction of
-        Repl    -> worker' runCmdOuts $ runWalletRepl
-        Cmd cmd -> worker' runCmdOuts $ runWalletCmd cmd
+auxxPlugin AuxxOptions{..} = \case
+        Left WithCommandAction{..} -> worker' runCmdOuts $ \sendActions -> do
+            printAction <- getPrintAction
+            printAction "... the auxx plugin is ready"
```
int-index: This message doesn't disrupt the REPL if the user started entering a command before the `auxxPlugin` is ready to handle commands :tada:

auxx/src/Main.hs
```diff
@@ -104,5 +105,10 @@ action opts@AuxxOptions {..} = withConfigurations conf $ do
 main :: IO ()
 main = withCompileInfo $(retrieveCompileTimeInfo) $ do
     opts <- getAuxxOptions
-    let loggingParams = CLI.loggingParams "auxx" (aoCommonNodeArgs opts)
-    loggerBracket loggingParams $ runProduction $ action opts
+    let loggingParams = (CLI.loggingParams "auxx" (aoCommonNodeArgs opts))
+                        { lpConsoleLog = Just False }
```
gromakovsky: I think console log should be disabled only for repl, am I wrong?

auxx/src/Repl.hs
```diff
@@ -0,0 +1,174 @@
...
+        --
+    }
+
+-- | The action to post a command to be executed by the auxx plugin and wait for
+-- its result. This action blocks until the result is available: that is, until
+-- command execution completes or an exception is thrown.
+type PutCommandAction = Command -> AuxxRepl CommandResult
+
+-- | Create the auxx REPL (as an abstract @'IO' ()@ action) and the interface to
+-- interact with it ('WithcommandAction'). You are supposed to fork the REPL
```
gromakovsky: `WithCommandAction` (typo)

auxx/src/Repl.hs
```diff
@@ -0,0 +1,174 @@
...
+auxxReplOn printActionVar putCommand = Haskeline.runInputT settings $ do
+    printAction <- Haskeline.getExternalPrint
+    putMVar printActionVar (liftIO . printAction . toString)
+    greetings
+    loop (prompt putCommand)
+  where
+    greetings = do
+        outputTextLn "Welcome to Cardano SL Auxx REPL"
+        outputTextLn helpMessage
+    settings = Haskeline.Settings Haskeline.noCompletion Nothing True
```
gromakovsky: Why not use explicit records names?

auxx/src/Main.hs
int-index: The node tries to print to `stdout`, it disrupts the REPL.

auxx/src/Repl.hs
int-index: thanks, fixed

auxx/src/Repl.hs
int-index: good idea, fixed

auxx/src/Repl.hs
```diff
@@ -0,0 +1,178 @@
...
+    putMVar printActionVar (liftIO . printAction . toString)
+    greetings
+    loop (prompt putCommand)
+  where
+    greetings = do
+        outputTextLn "Welcome to Cardano SL Auxx REPL"
+        outputTextLn helpMessage
+    settings = Haskeline.Settings
+        { Haskeline.complete       = Haskeline.noCompletion
+        , Haskeline.historyFile    = Nothing -- TODO: pass it from CLI
```
gromakovsky: I created https://issues.serokell.io/issue/CSL-1779
We can do it later, but can you add issue id to this TODO?

auxx/src/Main.hs
gromakovsky: You probably got me wrong, I mean that if `aoAction opts` is `Cmd`, do we want to disable console output?

auxx/src/Repl.hs
int-index: done

auxx/src/Main.hs
int-index: ah, I see. Alright, fixed.

PR #1773, opened 2017-10-16
Change directory to workdir in CI so stack's cache can be retained

lehins:
A quick explanation why this fix actually works.

`STACK_WORK`, as well as `CACHED_STACK_WORK` are relative directories, since stack doesn't support absolute ones.

Actual issue appears on this line:
```
IF NOT DEFINED APPVEYOR_PULL_REQUEST_NUMBER IF EXIST %STACK_WORK% xcopy /q /s /e /r /k /i /v /h /y %STACK_WORK% %CACHED_STACK_WORK%
```
Because current directory was previously changed it simply can't find `STACK_WORK` and skips the copying.


domenkozar:
#1782

PR #1779, opened 2017-10-17
[CSL-1279] Part I: Remove 'ssc' from a bunch of places

neongreen:


neongreen:
@int-index because the `Ssc` class is defined in `cardano-sl-ssc` and `SscGodTossing` is defined on a different level. I don't like `ssc ~ SscGodTossing` either but I don't think it matters much because I'll eliminate it completely in another PR and adding `ssc ~ SscGodTossing` leads to fewer diff lines than replacing `ssc` with `SscGodTossing` everywhere.

int-index:
`SscGodTossing` is defined in `cardano-sl-godtossing`, and if you are going to remove `ssc` as a parameter, you'll have to depend everything on `cardano-sl-godtossing` anyway. I don't see a reason not to do it sooner than later.

explorer/src/Pos/Explorer/Socket/App.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell     #-}
 {-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE GADTs               #-}
```
gromakovsky: 🤔

explorer/src/Pos/Explorer/Socket/Methods.hs
```diff
@@ -338,8 +339,8 @@ addressSetByTxs tx txs =
     S.fromList $ txOutAddress <$> txs'
 
 getBlockTxs
-    :: forall ssc ctx m . ExplorerMode ctx m
-    => Block ssc -> m [TxInternal]
+    :: forall ssc ctx m . (ExplorerMode ctx m, ssc ~ SscGodTossing)
```
gromakovsky: Why is `ssc` parameter needed here at all?

lib/src/Pos/Arbitrary/Block.hs
```diff
@@ -77,26 +78,27 @@ instance Arbitrary (T.Body (T.GenesisBlockchain ssc)) where
...
 
-instance ( Arbitrary (SscPayload ssc)
-         , Arbitrary (SscProof ssc)
+instance ( Arbitrary (SscPayload SscGodTossing)
+         , Arbitrary (SscProof SscGodTossing)
          , Bi Raw
-         , SscHelpersClass ssc
+         , SscHelpersClass SscGodTossing
          , HasConfiguration
+         , ssc ~ SscGodTossing
```
gromakovsky: Why is this line needed?

lib/src/Pos/Arbitrary/Block.hs
```diff
@@ -205,8 +207,8 @@ instance Arbitrary T.MsgGetBlocks where
     arbitrary = genericArbitrary
     shrink = genericShrink
 
-instance (Arbitrary (SscPayload ssc), Arbitrary (SscProof ssc), Bi Raw, SscHelpersClass ssc, HasConfiguration) =>
-         Arbitrary (T.MsgHeaders ssc) where
+instance (Arbitrary (SscPayload ssc), Arbitrary (SscProof ssc), Bi Raw, SscHelpersClass ssc, HasConfiguration, ssc ~ SscGodTossing) =>
```
gromakovsky: Are there any benefits in using `ssc ~ SscGodTossing` here?

lib/src/Pos/Arbitrary/Block.hs
```diff
@@ -215,23 +217,26 @@ instance ( Arbitrary $ SscPayload ssc
...
             ) block key
 
-newtype BlockHeaderList ssc = BHL
-    { getHeaderList :: ([T.BlockHeader ssc], [PublicKey])
-    } deriving (Eq)
+newtype BlockHeaderList = BHL
+    { getHeaderList :: ([T.BlockHeader], [PublicKey])
+    }
+
+deriving instance Eq T.BlockHeader => Eq BlockHeaderList
```
gromakovsky: Why did you change it to standalone deriving?

lib/src/Pos/Arbitrary/Block.hs
```diff
@@ -364,8 +369,11 @@ generateBHL createInitGenesis startSlot slotCount = BHL <$> do
 -- part of the verification parameters are guaranteed to be valid, as are the
 -- slot leaders and the current slot.
 newtype HeaderAndParams ssc = HAndP
```
gromakovsky: Why does it still have `ssc` parameter?

lib/src/Pos/Arbitrary/Block.hs
```diff
@@ -364,8 +369,11 @@ generateBHL createInitGenesis startSlot slotCount = BHL <$> do
...
-    { getHAndP :: (T.VerifyHeaderParams ssc, T.BlockHeader ssc)
-    } deriving (Eq, Show)
+    { getHAndP :: (T.VerifyHeaderParams ssc, T.BlockHeader)
+    }
+
+deriving instance Eq T.BlockHeader => Eq (HeaderAndParams ssc)
+deriving instance Show T.BlockHeader => Show (HeaderAndParams ssc)
```
gromakovsky: Why standalone?

lib/src/Pos/Block/BListener.hs
```diff
@@ -19,19 +19,20 @@ import           Pos.Block.Types       (Blund)
...
+import           Pos.Ssc.GodTossing.Type (SscGodTossing)
 
 class Monad m => MonadBListener m where
     -- Callback will be called after putting blocks into BlocksDB
     -- and before changing of GStateDB.
     -- Callback action will be performed under block lock.
     onApplyBlocks
-        :: forall ssc . SscHelpersClass ssc
-        => OldestFirst NE (Blund ssc) -> m SomeBatchOp
+        :: SscHelpersClass SscGodTossing
```
gromakovsky: `Pos.Ssc.GodTossing.Type` contains this instance, I think it's better to write `HasConfiguration =>` instead.

lib/src/Pos/Block/BListener.hs
```diff
@@ -42,11 +43,11 @@ instance {-# OVERLAPPABLE #-}
     onRollbackBlocks = lift . onRollbackBlocks
 
 onApplyBlocksStub
-    :: forall ssc m . (SscHelpersClass ssc, Monad m)
-    => OldestFirst NE (Blund ssc) -> m SomeBatchOp
+    :: forall ssc m . (SscHelpersClass ssc, Monad m, ssc ~ SscGodTossing)
```
gromakovsky: Same question, I will stop asking it again and again.

lib/src/Pos/Block/Core/Genesis/Misc.hs
```diff
@@ -76,55 +77,60 @@ instance BiSsc ssc => Buildable (GenesisBlock ssc) where
...
 -- Smart constructors
 ----------------------------------------------------------------------------
 
 type SanityConstraint ssc
-     = ( HasDifficulty $ BlockHeader ssc
-       , HasHeaderHash $ BlockHeader ssc
+     = ( HasDifficulty BlockHeader
+       , HasHeaderHash BlockHeader
        , HasConfiguration
+       , ssc ~ SscGodTossing
```
gromakovsky: Is `ssc` parameter needed?

lib/src/Pos/Block/Logic/Util.hs
```diff
@@ -150,7 +151,7 @@ calcOverallChainQuality =
     getCurrentSlotFlat >>= \case
         Nothing -> pure Nothing
         Just curFlatSlot ->
-            calcOverallChainQualityDo curFlatSlot <$> DB.getTipHeader @ssc
+            calcOverallChainQualityDo curFlatSlot <$> DB.getTipHeader @SscGodTossing
```
gromakovsky: I hope this type application won't be needed after CSL-1279 is fully finished.

neongreen:
> Why does it still have ssc parameter?

image

I've got a big list of things that have the `ssc` parameter but shouldn't, this will be addressed in subsequent PRs

lib/src/Pos/Block/Network/Listeners.hs
```diff
@@ -46,33 +47,33 @@ blockListeners oq = constantListeners $ map ($ oq)
 -- headers from some checkpoints that are older than optional @to@
 -- field.
 handleGetHeaders
-    :: forall pack ssc ctx m.
-       (WorkMode ssc ctx m)
+    :: forall pack ctx m.
+       (WorkMode SscGodTossing ctx m)
```
gromakovsky: Why is `ssc` still parameter of `WorkMode`? 🤔

lib/src/Pos/Block/RetrievalQueue.hs
```diff
@@ -14,7 +14,7 @@ import           Pos.Communication.Types (NodeId)
 
 -- | Task that block retrieval queue is asked to do.
 data BlockRetrievalTask ssc = BlockRetrievalTask
```
gromakovsky: Why did you leave `ssc` parameter?

lib/src/Pos/Context/Functions.hs
```diff
@@ -45,8 +46,8 @@ genesisLeaders = followTheSatoshiUtxo (gdFtsSeed genesisData) utxo
   where
     GenesisUtxo utxo = genesisUtxo
 
-genesisBlock0 :: forall ssc . (HasConfiguration, SscHelpersClass ssc)
-    => GenesisBlock ssc
+genesisBlock0 :: forall ssc . (HasConfiguration, SscHelpersClass ssc, ssc ~ SscGodTossing)
```
gromakovsky: These constraints are likely useless.

lib/src/Pos/DB/Block.hs
```diff
@@ -196,46 +197,46 @@ loadDataByDepth getter extraPredicate depth h = do
...
-    => BlockCount -> HeaderHash -> m (NewestFirst [] (BlockHeader ssc))
+    :: (SscHelpersClass ssc, MonadDBRead m, HasConfiguration, ssc ~ SscGodTossing)
+    => BlockCount -> HeaderHash -> m (NewestFirst [] BlockHeader)
 loadHeadersByDepth = loadDataByDepth getHeaderThrow (const True)
 
 -- | Load headers which have depth less than given and match some criterion.
 loadHeadersByDepthWhile
-    :: (SscHelpersClass ssc, MonadDBRead m, HasConfiguration)
-    => (BlockHeader ssc -> Bool)
+    :: (SscHelpersClass ssc, MonadDBRead m, HasConfiguration, ssc ~ SscGodTossing)
```
gromakovsky: A lot of lines with strange constraints.

explorer/src/Pos/Explorer/Socket/App.hs
neongreen: GADTs is needed for the `~` syntax

explorer/src/Pos/Explorer/Socket/Methods.hs
neongreen: It's likely not needed but I tried to avoid removing `ssc` from more places than I wanted to (`Block` and `BlockHeader`) because it would've increased the size of this PR (e.g. if you remove `ssc` from a `forall`, you also need to change all type applications when there are type applications)

lib/src/Pos/Arbitrary/Block.hs
neongreen: Nice, thanks. I removed it.

lib/src/Pos/Arbitrary/Block.hs
neongreen: Nope, other than the fact that it already was in my clipboard.

lib/src/Pos/Arbitrary/Block.hs
neongreen: Because previously it was automatically deriving a `Eq (T.BlockHeader ssc)` constraint for the instance, but after I removed `ssc` it broke.

lib/src/Pos/Block/Logic/Util.hs
neongreen: Yep.

neongreen:
> to be honest I don't like it, because there are a lot of things to fix now

I think that they will be pretty easy to fix because pretty much everything you pointed out can be found simply by grepping for ` ssc ` in code (with an exception of standalone deriving, but I have that written down). I don't see how “I'll remove some `ssc` in one PR and the rest in another PR” is harder to verify than “I'll remove all `ssc` in one PR”.

explorer/src/Pos/Explorer/Socket/App.hs
gromakovsky: It's true, but there is no `~` syntax in this file.

lib/src/Pos/Arbitrary/Block.hs
gromakovsky: It seems very strange that this module doesn't see `Eq T.BlockHeader` instance and this constraint is needed at all.

explorer/src/Pos/Explorer/Socket/App.hs
neongreen: Hm-m-m.

lib/src/Pos/Block/BListener.hs
neongreen: Good point, thanks.

neongreen:
@gromakovsky I have fixed most of the things you were unhappy about (I think).

lib/src/Pos/Block/Core/Union/Types.hs
```diff
@@ -30,39 +28,21 @@ import           Pos.Core.Block
 ----------------------------------------------------------------------------
 
 -- | Ssc w/ buildable blockchain
-type BiSsc ssc =
-    ( Ssc ssc
-    , Bi (GenericBlockHeader (GenesisBlockchain ssc))
-    , Bi (GenericBlockHeader (MainBlockchain ssc))
+type BiSsc =
```
gromakovsky: This name looks even more weird now (as well as the comment above).

lib/src/Pos/Block/Pure.hs
```diff
@@ -63,7 +62,10 @@ data VerifyHeaderParams ssc = VerifyHeaderParams
       -- ^ Maximal allowed header size. It's applied to 'BlockHeader'.
     , vhpVerifyNoUnknown :: !Bool
       -- ^ Check that header has no unknown attributes.
-    } deriving (Show, Eq)
+    }
+
+deriving instance Eq BlockHeader => Eq VerifyHeaderParams
```
gromakovsky: I really don't get why `Eq BlockHeader =>` and such stuff may be needed. `BlockHeader` is a concrete type and there is an `Eq` instance for it (we import `Pos.Block.Core`, `Pos.Core`, `Universum` here).

lib/src/Pos/Block/Core/Union/Types.hs
neongreen: Sure, I'm going to remove it.

lib/src/Pos/Block/Pure.hs
neongreen: I haven't investigated it because I wanted the code to compile and that's all. I'll fix it in part II.

PR #1786, opened 2017-10-18
[CSL-1193] Get rid of explorer code in cardano-sl lib and with-explorer flag

pva701:


gromakovsky:
There are 6 conflicts already, btw.

lib/src/Pos/Generator/BlockEvent.hs
```diff
@@ -200,10 +204,11 @@ genBlocksInStructure ::
        , Functor t, Foldable t)
     => AllSecrets
     -> GenesisWStakeholders
+    -> TxpGlobalSettings
     -> Map Path BlockDesc
     -> t Path
     -> RandT g m (t BlundDefault)
-genBlocksInStructure secrets bootStakeholders annotations s = do
+genBlocksInStructure secrets bootStakeholders txpSettings annotations s = do
```
gromakovsky: Why did you add it as arguments everywhere in this module instead of putting into reader context?

lib/src/Pos/Generator/BlockEvent/DSL.hs
```diff
@@ -119,11 +120,12 @@ runBlockEventGenT
     :: BlockTxpGenMode g ctx m
     => AllSecrets
     -> GenesisWStakeholders
+    -> TxpGlobalSettings
     -> BlockEventGenT g m ()
     -> RandT g m BlockScenario
-runBlockEventGenT secrets genStakeholders m = do
+runBlockEventGenT secrets genStakeholders txpSettings m = do
```
gromakovsky: Why not add to context instead?

lib/src/Pos/Generator/BlockEvent.hs
pva701: makes sense

pva701:
@gromakovsky I've fixed all issues and force pushed, can you approve if everything is ok?

gromakovsky:
> The Travis CI build failed 

:loh:

PR #1791, opened 2017-10-19
[CSL-1778] Set termSeverity for other log templates

int-index:


gromakovsky:
Travis failure is not related.

PR #1792, opened 2017-10-19
[CSM-367] Add endpoint to check local time difference

sectore:
_Task:_
https://issues.serokell.io/issue/CSM-367

_How to test:_
1. Build everything by running `./scripts/build/cardano-sl.sh` and `./scripts/build/daedalus-bridge.sh`
2. Run wallet api
```
tmux
scripts/launch/demo-with-wallet-api.sh
```
3. Test endpoint w/ `Node`
```
cd daedalus
node
> var api = require('../output/Daedalus.ClientApi')
undefined
> var js = require('../src/tls_workaround.js')
undefined
> var ca = js.readCA('{absolute-path-to}/cardano-sl/scripts/tls-files/ca.crt')
undefined
> var tls = api.tlsInit(ca)
undefined
> api.localTimeDifference(tls).then(console.log).catch(console.log);
Promise { <pending> }
> 0
```

The time unit of the result is [`Microsecond`](https://en.wikipedia.org/wiki/Microsecond):
- `0` means we have no difference
- All of `> 0` means that we have got - _ooooh my_ - a difference


infra/Pos/NtpCheck.hs
```diff
@@ -63,3 +69,20 @@ timeDifferenceWarnInterval = fromIntegral (Infra.ccTimeDifferenceWarnInterval in
 
 timeDifferenceWarnThreshold :: HasInfraConfiguration => Microsecond
 timeDifferenceWarnThreshold = fromIntegral (Infra.ccTimeDifferenceWarnThreshold infraConfiguration)
+
+type NtpDiffVar = TVar Microsecond
```
int-index: I would expect this variable to contain `NtpStatus`.

sectore: @int-index Why? `NtpDiffVar` does contain the time difference in Microsecond only. But it says nothing about the status. See mkNtpDiffVar and endpoint handler `localTimeDifference`.

sectore: @int-index Ok, I see. Check last commit, pls.

infra/Pos/NtpCheck.hs
```diff
@@ -63,3 +69,14 @@ timeDifferenceWarnInterval = fromIntegral (Infra.ccTimeDifferenceWarnInterval in
...
 timeDifferenceWarnThreshold :: HasInfraConfiguration => Microsecond
 timeDifferenceWarnThreshold = fromIntegral (Infra.ccTimeDifferenceWarnThreshold infraConfiguration)
+
+type NtpStatusVar = TVar NtpStatus
+
+-- Helper to get ntp status
+mkNtpStatusVar :: ( NtpCheckMonad m , Mockables m [ CurrentTime, Delay] )
+    => m NtpStatusVar
+mkNtpStatusVar = do
+    status <- newTVarIO NtpSyncOk
```
int-index: I think you should use an MVar and create it empty. The reason is that if the front-end makes a request before the first NTP check is performed, we'd probably want to wait for the check rather than say that everything is OK (when it's possibly not).

I don't know what the desired behavior is with regards to this, so I'm going to approve, but keep this suggestion in mind. And if the current behavior is intended, perhaps add a clarifying comment.

wallet/src/Pos/Wallet/Web/Methods/Misc.hs
```diff
@@ -91,6 +93,21 @@ syncProgress =
...
+-- NTP (Network Time Protocol) based time difference
+----------------------------------------------------------------------------
+
+localTimeDifference :: MonadWalletWebMode m => m Word
+localTimeDifference =
+    mkNtpStatusVar >>= readMVar >>= pure . diff
+  where
+    diff :: NtpStatus -> Word
+    diff = \case
+        NtpSyncOk -> 0
```
akegalj: Well this doesn't seem fully accurate.

My understanding if it is `NtpSyncOk` that doesn't mean `localTimeDifference == 0` but it means `localTimeDifference < timeDifferenceWarnThreshold` which isn't the same.

akegalj: Also, I am not sure we would like to use `timeDifferenceWarnThreshold` for this purpose (which is to notify user via UI) , or we should define another constant `timeDifferenceUserWarningThreshold` and warn the user to adjust its clock before (ie, user threshold should be smaller then `timeDifferenceWarnThreshold`) it actually hits the limit.

@pva701 @flyingleafe thoughts ?

sectore: > if it is NtpSyncOk that doesn't mean localTimeDifference == 0

@akegalj ^ That's correct. However, the first idea of this endpoint is to keep the returned value as simple as possible. Which means: `Daedalus` is interested in a value of time difference if we have `NtpDesync` only. If we have `NtpSyncOk` we won't return a time difference, even if the difference (`d`) is `d >= 0 && d < timeDifferenceWarnThreshold`. In this case we just send a value of `0` to hide the computation by using `timeDifferenceWarnThreshold` and just to say that everything is ok.

If there is a need, we can change the returned value of this endpoint of course ;)

pva701: sounds reasonable to me

PR #1793, opened 2017-10-19
[CSM-500] Remember pending transaction from the very beginning

Martoon-00:
This is important, because if we remember it once transaction creation is done,
then creation may complete only after the moment when transaction gets to
blocks and then gets detected by tracker, so effectively tracker misses the tx.

Solution is to add new pending transaction status, `PtxCreating`. which
is not to display to frontend, and such transaction may be removed if
full cycle of transaction creation finishes with fatal error (unlike
`PtxApplying`).
Transaction remembered with this status still caries all pending meta info,
so tracker can successfully update status of that transaction.

Tested on:
* Normal creation
* Slot-length artificial delay between transaction submission over network and local application (original issue case)
* Fatal and non-fatal failure scenarios

### DIFF
--- a/wallet/src/Pos/Wallet/Web/ClientTypes/Instances.hs
+++ b/wallet/src/Pos/Wallet/Web/ClientTypes/Instances.hs
@@ -107,6 +107,7 @@ type instance OriginType CPtxCondition = Maybe PtxCondition
 
 instance ToCType CPtxCondition where
     encodeCType = maybe CPtxNotTracked $ \case
+        PtxCreating{}       -> CPtxInvisible
         PtxApplying{}       -> CPtxApplying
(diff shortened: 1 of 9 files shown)


akegalj:
@Martoon-00 figured out this changes DB scheme - so safe migration should be added.

by @Martoon-00 
> looks like it's not, because serialization of `PtxCondition` has changed entirely

Martoon-00:
Or I can just move new `PtxCreating` constructor to the bottom of the list
Magic :star2:

Martoon-00:
@akegalj done & tested

Martoon-00:
I messed up with branches while rebasing upon master, new PR is #1877

PR #1796, opened 2017-10-19
[CSL-1784] Force flush on block write

volhovM:
Possible bug causing CSL-1784 is the fact that `putData` wasn't flushing handle, so this scenario was possible:
1. We write block and put info about it in gstate, we update it.
2. System IO doesn't flush the data, so it's not yet on the disk.
3. Somebody turns machine off.

### DIFF
--- a/lib/src/Pos/DB/Block.hs
+++ b/lib/src/Pos/DB/Block.hs
@@ -57,43 +57,45 @@ module Pos.DB.Block
 
 import           Universum
 
-import           Control.Lens           (at, _Wrapped)
-import           Data.ByteArray         (convert)
-import qualified Data.ByteString        as BS (readFile, writeFile)
-import           Data.Default           (Default (def))
-import           Ether.Internal         (HasLens (..))
-import           Formatting             (build, formatToString, sformat, (%))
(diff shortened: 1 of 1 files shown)


lib/src/Pos/DB/Block.hs
```diff
@@ -496,7 +498,9 @@ getData fp = flip catch handle $ liftIO $
         | otherwise = throwM e
 
 putData ::  (MonadIO m, Bi v) => FilePath -> v -> m ()
-putData fp = liftIO . BS.writeFile fp . serialize'
+putData fp v = liftIO $
+    bracket (openBinaryFile fp WriteMode) hClose $ \h ->
+        BS.hPut h (serialize' v) >> hFlush h
```
gromakovsky: This whole thing is in `IO`, so you could've used `withBinaryFile` instead (it's in `base`).

volhovM: Not that important, but thanks for the hint.

volhovM: ok, i'll implement it

volhovM:
(2) I'll create another PR.

PR #1799, opened 2017-10-20
[CSL-1279] Part III: merge -ssc and -godtossing, remove 'Ssc'

neongreen:
This one is already big enough. Future parts will bring:

* Removal of the `.GodTossing` part from more places
* Renaming of everything with `gt` in it
* Removal of other `Ssc*` classes
* Removal of `SscAlgo`
* Conversion of `SscBlock` into a type


ssc/Pos/Ssc/Types.hs
```diff
@@ -0,0 +1,222 @@
...
+    , sscLocal  :: !(TVar SscLocalData)
+    }
+
+----------------------------------------------------------------------------
+-- SscBlock
+----------------------------------------------------------------------------
+
+-- [CSL-1156] Find a better way for this
+--
+-- NB. It must be a newtype instead of a type (unlike e.g. 'TxpBlock' and
```
int-index: This comment is outdated

ssc/Pos/Ssc/GodTossing/Instance.hs
```diff
@@ -0,0 +1,26 @@
...
+import           Pos.Binary.Core                    ()
+import           Pos.Binary.GodTossing              ()
+import           Pos.Core                           (HasConfiguration)
+import           Pos.Ssc.Class.Helpers              (SscHelpersClass (..))
+import           Pos.Ssc.Core                       (defaultSscPayload,
+                                                     stripSscPayload)
+import           Pos.Ssc.GodTossing.Functions       (sanityChecksSscPayload)
+import           Pos.Ssc.VerifyError                (SscVerifyError (..))
+
+instance HasConfiguration => SscHelpersClass where
```
int-index: Remove `SscHelpersClass` in the current iteration, please. It has no parameters and its instance defined in the same package, so you're not going to hit any circular dependencies, should be rather painless.

int-index:
> Conversion of SscBlock into a type

Don't do that, perhaps. `newtype` > `type`

neongreen:
> Remove SscHelpersClass in the current iteration, please.

Ok.

> This comment is outdated

I was going to do it after changing `SscBlock`. I think that it makes more sense to change `SscBlock` to a newtype than _not_ do it, because everywhere else we have blocks-as-Either (Block, TxpBlock, UpdateBlock, etc) and so in this case the newtype just makes things less consistent for no obvious gain.

neongreen:
@int-index I have addressed your comments.

CODEOWNERS
```diff
@@ -33,7 +33,7 @@ Wallet/                         @akegalj @flyingleafe
 txp/                            @gromakovsky @pva701 @neongreen
 update/                         @gromakovsky @pva701
 lrc/                            @pva701
-godtossing/                     @neongreen
+ssc/                            @neongreen
```
gromakovsky: I want to share ownership of `ssc/` :)
Can you add me in the next part?

neongreen: Okay.

PR #1802, opened 2017-10-21
[CSM-503] Optimize change addresses evaluation

Martoon-00:
Tested with outgoing/incoming and a2a transactions

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
+++ b/wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
@@ -33,6 +33,9 @@ module Pos.Wallet.Web.Tracking.Sync
 
        , buildTHEntryExtra
        , isTxEntryInteresting
+
+       -- For tests
+       , evalChange
        ) where
 
 import           Universum
--- a/wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
+++ b/wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
@@ -46,6 +46,7 @@ import           Control.Monad.Catch (handleAll)
 import qualified Data.DList as DL
 import qualified Data.HashMap.Strict as HM
 import qualified Data.HashSet as HS
+import qualified Data.List.NonEmpty as NE
 import qualified Data.Map as M
 import           Ether.Internal (HasLens (..))
 import           Formatting (build, sformat, (%))
@@ -69,7 +70,7 @@ import           Pos.GState.BlockExtra (foldlUpWhileM, resolveForwardLink)
 import           Pos.Slotting (MonadSlots (..), MonadSlotsData, getSlotStartPure, getSystemStartM)
 import           Pos.StateLock (Priority (..), StateLock, withStateLockNoMetrics)
 import           Pos.Txp (MonadTxpMem, flattenTxPayload, genesisUtxo, getLocalTxsNUndo, topsortTxs,
-                          unGenesisUtxo, utxoToModifier)
+                          unGenesisUtxo, utxoToModifier, _txOutputs)
 import           Pos.Util.Chrono (getNewestFirst)
 import           Pos.Util.LogSafe (logInfoS, logWarningS)
 import qualified Pos.Util.Modifier as MM
@@ -335,7 +336,11 @@ trackingApplyTxs (eskToWalletDecrCredentials -> wdc) dbUsed getDiff getTs getPtx
             addedHistory = maybe camAddedHistory (flip DL.cons camAddedHistory) (isTxEntryInteresting thee)
 
             usedAddrs = map (cwamId . snd) theeOutputs
-            changeAddrs = evalChange (constructAllUsed dbUsed camUsed) (map snd theeInputs) (map snd theeOutputs)
+            changeAddrs = evalChange
+                              (constructAllUsed dbUsed camUsed)
+                              (map snd theeInputs)
+                              (map snd theeOutputs)
+                              (length theeOutputs == NE.length (_txOutputs $ taTx tx))
 
             mPtxBlkInfo = getPtxBlkInfo blkHeader
             addedPtxCandidates =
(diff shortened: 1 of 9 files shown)


wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -478,13 +471,14 @@ rollbackModifierFromWallet wid newTip CAccModifier{..} = do
...
+    :: [CWAddressMeta]  -- ^ Own input addresses of tx
+    -> [CWAddressMeta]  -- ^ Own outputs addresses of tx
     -> [CId Addr]
-evalChange allAddresses inputs outputs
-    | null inputs || null outputs = []
-    | otherwise = filter (isTxLocalAddress allAddresses (NE.fromList inputs)) outputs
+evalChange inputs outputs
+    | [] <- inputs = []
+    | inp : _ <- inputs =
+        let srcAccount = addrMetaToAccount inp
```
pva701: oh, I realized that plan was great but didn't consider transactions which we send from account to the same account. now we can't differ change address from destination address.

on the other hand, AFAIU we can get all addresses of account and check each of output addresses. what do you think?

Martoon-00: Actually this problem exists for a long time, and we didn't care much. For `A->A` transactions we suggested that all its outputs are changes. And now processing is the same.
(Btw now I wonder, with such logic, don't we remove extra change addresses on rollback?)

pva701: >Actually this problem exists for a long time

why? we took our addressed and determined change address as "not presented in our addresses".
and I suggest doing a similar thing here.

yes, there will still cases when we are wrong, but it will be ok for sending from one account to the same account. not it's broken.

>Btw now I wonder, with such logic, don't we remove extra change addresses on rollback?

hm, IIRC we were discussing it and decided to do 
`type CustomAddresses = HashMap (CId Addr) HeaderHash`
so we will remove it only when block which added this address will be rolled back.

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -292,20 +293,32 @@ syncWalletWithGStateUnsafe encSK wTipHeader gstateH = setLogger $ do
...
             (maybe (error "No genesis block corresponding to header hash") pure <=< DB.getHeader)
 
+constructAllUsed
+    :: [(CId Addr, HeaderHash)]
+    -> VoidModifier (CId Addr, HeaderHash)
+    -> HashSet (CId Addr)
+constructAllUsed dbUsed modif =
+    HS.map fst $
+    HS.fromMap $
+    void $
```
gromakovsky: Btw, there is `Pos.Util.getKeys = fromMap . void`.

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -292,20 +293,32 @@ syncWalletWithGStateUnsafe encSK wTipHeader gstateH = setLogger $ do
...
+    MM.modifyHashMap modif $
+    HM.fromList $
```
gromakovsky: Hm, perhaps we need `SetModifier` 🤔

wallet/test/Test/Pos/Wallet/Web/Tracking/SyncSpec.hs
```diff
@@ -66,3 +79,61 @@ twoApplyTwoRollbacksSpec = walletPropertySpec twoApplyTwoRollbacksDesc $ do
...
+    deriving Show
+
+instance Arbitrary AddressesFromSameAccounts where
+    arbitrary = do
+        wId <- arbitrary
+        accIdx <- arbitrary
+        let genAddrs n = map (uncurry $ CWAddressMeta wId accIdx) <$> vectorOf n arbitrary
+        inpAddrs <- choose (1, 5) >>= genAddrs
+        outAddrs <- choose (1, 5) >>= genAddrs
+        usedBase <- (inpAddrs ++) <$> (choose (1, 10) >>= flip vectorOf arbitrary)
```
Martoon-00: instead
> `(choose (1, 10) >>= flip vectorOf arbitrary)`
it can be
> `resize 10 $ listOf1 arbitrary`

wallet/test/Test/Pos/Wallet/Web/Tracking/SyncSpec.hs
```diff
@@ -66,3 +79,61 @@ twoApplyTwoRollbacksSpec = walletPropertySpec twoApplyTwoRollbacksDesc $ do
...
+        (changeAddrs, extraUsed) <- elements [False, True] >>= \case
```
Martoon-00: There is `oneof` )

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -424,14 +437,41 @@ rollbackModifierFromWallet wid newTip CAccModifier{..} = do
...
+    -> [CWAddressMeta] -- ^ Own input addresses of tx
+    -> [CWAddressMeta] -- ^ Own outputs addresses of tx
     -> [CId Addr]
-evalChange allAddresses inputs outputs
-    | null inputs || null outputs = []
-    | otherwise = filter (isTxLocalAddress allAddresses (NE.fromList inputs)) outputs
+evalChange allUsed inputs outputs
+    | [] <- inputs = [] -- It means this transaction isn't our outgoing transaction.
+    | inp : _ <- inputs =
+        let srcAccount = addrMetaToAccount inp in
```
Martoon-00: Why not just
```
let nayn = 1
    kot = 2
in eto nyan kot
```

wallet/test/Test/Pos/Wallet/Web/Tracking/SyncSpec.hs
pva701: `oneof` takes monadic actions

Martoon-00:
Ok, now it seem to work. Consider it approved :green_apple:

PR #1805, opened 2017-10-22
Clean-up log templates

gromakovsky:
Log templates contained some garbage which is older than universe (like
`time-slave` handler).
Some logger names (like `comm`, `dht`) are not used nowadays.
Some stuff was redundant (e. g. `Debug` severity for a child of another
handler with `Debug severity).

I also added some comments and renamed one file.


scripts/log-templates/update-log-config.yaml
```diff
@@ -1,20 +1,9 @@
-# Same as log-template.yaml, but without files, 
+# Same as log-template-demo.yaml, but without files,
```
volhovM: Without files? What does it mean? It has files output. I guess the difference is in rotation.

gromakovsky: I have no idea, I am not competent enough to clarify this comment.

PR #1806, opened 2017-10-22
[CSL-1705] Describe transaction processing (txp)

gromakovsky:
I've described the whole algorithm of tx payload verification without delving into implementation details.
It should be reviewed carefully by people familiar with code and also these people should make sure that I indeed didn't forget anything.

I also deleted one redundant check (it was duplicated).

### DIFF
--- a/docs/txp.md
+++ b/docs/txp.md
@@ -32,54 +32,91 @@ places.
 ## Global transaction processing
 
 Global transaction processing can be described by presenting an
-algorithm to solve the following problem: given a sequence of blocks
-`B₀, B₁, B₂, …` (where `B₀` is the first genesis block) check whether
-transactions from these blocks are valid. The algorithm is similar to
-the one from Bitcoin (for example, UTXO is used to prevent
-double-spending), but is more complicated because there are more
-features.
+algorithm to solve the following problem:
 
-For verification we maintain some state (called GState) which
-corresponds to the application of a sequence of blocks. It's
-associated with the hash of the header of the last applied block
-(i. e. this hash is integral part of this state). So, given a sequence
-of blocks `B₀, B₁, B₂, …` we can compute sequence of states `S₀, S₁,
-S₂, …`. There is also genesis state `G` which takes places before any
-block is applied. This state is derived from genesis data (see ??? TODO).
+> Given a sequence of blocks `B₀, B₁, B₂, …` (where `B₀` is the first genesis block) check whether
+transactions from these blocks are valid.
 
-In theory transaction verification can be stateless (except genesis
-state `G`) and be described without mentioning any additional state
+The algorithm is similar to the one from Bitcoin (for example, UTXO is used to prevent
+double-spending), but is more complicated due to various reasons, including:
+  
+* Extendable (via update mechanism) data structures
+* Maintaining stakes in addition to balances
+
+
+We describe global txp as stateful algorithm:
+* Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))
+* `S₁, S₂, …` are maintained as sequential application of blocks  `B₀, B₁, B₂, … ` to state `S₀`  
+* State transition function. Given GState `S` and a main block `B` return either an error describing why `B` is invalid (w.r.t. tx payload) or new GState `S'`
+  ```
+  verifyAndApplyGState :: GState -> MainBlock -> Either TxPayloadVerificationError GState
+  -- ^ Note, that function definition and types are different in code and are put here for reader's convenience
+  ```
+  * Note, that state transition function is defined only for main blocks. This is done because genesis blocks (as defined in Ouroboros) don't have tx payload and thus are not considered by global txp.
+  * Recall that tx payload contains transactions (stored in Merkle tree) and their witnesses.
+
+Note, that in theory transaction verification can be stateless (except genesis
+state `S₀`) and be described without mentioning any additional state
 (apart from blocks themselves). But in practice it would be very
-inefficient and inconvenient, so we present the verification algorithm
-in a stateful way.
+inefficient and inconvenient.
 
-The problem: given GState `S` and a block `B` return either an error
-describing why `B` is invalid (w.r.t. tx payload) or new GState
-`S'`. Genesis blocks don't have tx payload and are ignored by global
-txp. Recall that tx payload contains transactions (stored in Merkle
-tree) and their witnesses.
+For sake of simplicity, we describe state transition function in two parts:
+* Verification: given GState `S` and a main block `B`, check whether `B` is valid (and can be applied to `S`)
+* Modification: given GState `S` and a main block `B` (successfully verified against `S`), produce `S'`
 
 ### GState
 
-In this section we describe parts of GState relevant to transaction
-processing.
+In this section we describe parts of GState relevant to transaction processing (GState is an actual structure in code used by few more components).
 
-* UTXO (unspent transaction outputs). It's a map from `TxIn` to
-  `TxOutAux` which contains all unspent outputs. `TxOutAux` is just an
-  alias for `TxOut`. Later it can be extended if we want to associate
+* UTXO (unspent transaction outputs)
+  ```
+  type TxId = Hash Tx
+  data TxOut = TxOut
+     { txOutAddress :: !Address
+     , txOutValue   :: !Coin
+     }
+  data TxIn
+     = TxInUtxo
+     { -- | Which transaction's output is used
+       txInHash  :: !TxId
+       -- | Index of the output in transaction's outputs
+     , txInIndex :: !Word32
+     }
+     | TxInUnknown !Word8 !ByteString
+     
+  data TxOutAux = TxOutAux
+     { toaOut   :: !TxOut
+     -- ^ Tx output
+     }
+     
+     
+  type Utxo = Map TxIn TxOutAux
+  ```
+  
+  * UTXO is a map from `TxIn` to
+  `TxOutAux` ([code](https://github.com/input-output-hk/cardano-sl/blob/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/txp/Pos/Txp/Toil/Types.hs#L58)) which contains all unspent outputs.  
+  * `TxOutAux` is just an alias for `TxOut` ([code](https://github.com/input-output-hk/cardano-sl/blob/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/txp/Pos/Txp/Core/Types.hs#L160)). Later it can be extended if we want to associate
   more data with unspent outputs (e. g. slot of the block in which
   this transaction appeared).
-  For example, if a transaction `A` has 1 output (`out1`) which hasn't been
+  * Example: if a transaction `A` has 1 output (`out1`) which hasn't been
   spent yet, UTXO will have a pair `(TxIn (hash A) 0, out1)`.
-* Stakes. They are not needed for tx verification, but they are needed
-  for other parts of the protocol (for example, leader election
-  process called FTS) and they are maintained by tx processing.
-  We store total stake and a map from `StakeholderId` to its stake (`Coin`).
-* Adopted `BlockVersionData`. This data type contains various
-  parameters of the algorithm which can be updated by Update
-  System (details are not covered in this document). Txp doesn't
-  modify this value, it only reads it. Some values are relevant for
-  transaction processing:
+* Stakes
+  ```
+  type StakesMap = HashMap StakeholderId Coin
+  type TotalStake = Coin -- this type doesn't exist in code and put here for convenience of reader
+  ```
+  * Stakes are used by other parts of the protocol (for example, leader election
+  process Follow-the-satoshi or FTS)
+  * Stakes are not needed for tx verification
+  * Stakes are maintained by tx processing: we store total stake and a map from `StakeholderId` to its stake (`Coin`)
+* Adopted `BlockVersionData`.
+  This data type ([code](https://github.com/input-output-hk/cardano-sl/blob/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/core/Pos/Core/Types.hs#L318)) contains various
+  parameters of the algorithm
+  which can be updated by Update System (details are not covered in this document).
+  
+  Txp doesn't modify this value, it only reads it.
+  
+  Some values are relevant for transaction processing:
   * `maxTxSize` – maximal size of a transaction.
   * `scriptVersion` – determines the rules which should be used to
     verify scripts.
@@ -87,7 +124,7 @@ processing.
     transaction has enough fees.
   * `unlockStakeEpoch` – epoch when bootstrap era ends.
 
-### Unknown data checks
+### Unknown data handling
 
 Many types in Cardano-SL are designed in an extensible way. They can
 be extended via softfork in future (e. g. new data can be attached or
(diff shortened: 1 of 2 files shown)


docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+* `Attributes` is basically a map from 1-byte integers to arbitrary
+  values. Some values can be parsed into known data types, other
+  values are treated as _unparsed fields_. `Attributes` are part of
+  various data types, e. g. `Tx`. In version 0 `Tx` always has empty
+  attributes. But in version 1 we may add more data to `Tx` (put it
+  into `Attributes`). In this case new software will parse this data
+  and old software will treat it as unparsed fields.
+* Unknown address type. Each valid address in Cardano-SL has a
+  type. It can be one of fixed address types or unknown address type
+  which can't be interprted by current version of software.
```
georgeee: > interprted

typo

docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+   case we can't know whether that input is valid and how much value
+   it has. So we perform only basic
+   checks: [outputs checks](#outputs-checks)
+   and [consistency check](#consistency-check). Note that we don't
+   perform
+   [transaction's attributes check](#transaction-attributes-check),
+   because `verifyAllIsKnown` is `False`.
+3. All inputs have known format (i. e. `TxInUtxo`). In this case
+   (which should be the most common one) we can check all inputs'
+   legitimacy, compare input and output sums, check fee, etc. All
```
georgeee: > inputs' legitimacy, compare input and output sums, check fee
>
> etc.

Need more detailed explanation. Some of these checks are described later, but not all. E.g. what's legitimacy. And there shouldn't be "etc.". I would propose to have here bullet-list and in subsections of this section show every check.

georgeee: > All inputs must exist in UTXO

Important to note that this is simple hashtable check w/o modification of the hashtable

docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+  and a redeem signature), we require that address is a `Redeem`
+  address and its spending data is `Redeem` spending data with the
+  same key as in the witness. Witness itself is checked in the same
+  way as `PkWitness` except that signature scheme is the one from
+  RSCoin and signing tag is `SignRedeemTx`.
+* If witness has unknown type with tag `t`, corresponding address must
+  have unknown type with tag `t`. Witness itself is valid only if
+  `verifyAllIsKnown` is `False`.
+
+Note: even though address doesn't contain its spending data, it's easy
```
georgeee: Not sure I follow, don't we miss info about that `Address'` contains attributes?

georgeee: Also, this note is cryptic and requires one to dive into code to be understood

docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+   (it probably indicates there is a mistake in currently adopted fee
+   policy). Then we compare actual fee of tx with minimal required fee
+   and if it's greater than or equal to the minimal value, transaction
+   is valid with regards to this check.
+2. Fee policy is `TxFeePolicyUnknown`. In this case the check is
+   skipped (i. e. transaction is considered valid with respect to fee
+   policy). It's questionable which behavior we want in this case, but
+   currently it's implemented this way.
+
+Note: if all unspent outputs corresponding to inputs of transaction
```
georgeee: This isn't a note, but another consensus rule which should be subitem for `1.`.

docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+If all checks above pass, we modify GState appropriately. We modify
+stakes and UTXO. UTXO modification is trivial:
+* All `TxInUtxo` inputs with are deleted from UTXO.
+* For each output `out` with index `i` we put `(TxInUtxo (hash tx) i,
+  out)` into UTXO.
+
+Stakes modification is a bit more complex, but is also very
+intuitive. We have `TxOut`s corresponding to inputs and outputs of
+transaction. Each `TxOut` can be converted to a list of
+`(StakeholderId, Coin)` pairs. It depends on `AddrStakeDistribution`
```
georgeee: >  It depends on AddrStakeDistribution of given address. For BootstrapEraDistr the value of TxOut is distributed among bootstrap stakeholders proportional to their weights. If distribution is SingleKeyDistr id, then the value of TxOut will be assigned to id. In case of MultiKeyDistr stake will be distributed among multiple stakeholders according to specified portions. 


Bullet list please

docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+   legitimacy, compare input and output sums, check fee, etc. All
+   inputs must exist in UTXO. In the following we describe the case
+   when all inputs have `TxInUtxo` type and we assume that for each
+   input we know corresponding unspent output (if we don't,
+   transaction is invalid).
+
+All inputs of a transaction must be different and each input must be
+properly certified by its witness.
+
+### Witness checks
```
georgeee: This and following sections should be `####`

docs/txp.md
gromakovsky: I wrote that these checks are described below and rephrased some stuff here.

docs/txp.md
gromakovsky: Okay.

pva701:
@gromakovsky
your comment for 97bdd9a:
>This check was redundant, because in `verifyTxUtxo` we `find` an unknown input
and if it exists, we will reach `verifyInputs` function.

looks wrong, it should be like
>This check was redundant, because in `verifyTxUtxo` we `find` an unknown input
and if it **doesn't** exist, we will reach `verifyInputs` function.

pva701:
also, I don't think this check should be removed, it's pretty small. at least if you want to remove this check I would rename `verifyInputs` to `verifyKnownInputs` to show this function doesn't expect to see unknown inputs

docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+| Our   | Adopted | Check? |
+| ----- | ------- | ------ |
+| 1.2.3 |  1.2.3  | Yes    |
+| 1.2.3 |  1.2.4  | No     |
+| 1.2.3 |  1.2.2  | No     |
+| 1.2.3 |  1.3.2  | No     |
+| 1.2.3 |  1.1.1  | Yes    |
+| 2.2.8 |  1.9.9  | Yes    |
+
+If `(major, minor)` of our version is greater than of adopted
```
pva701: don't you think it's better to describe why check isn't done if `(major, minor)` is same but alt is different?

docs/txp.md
```diff
@@ -186,16 +186,22 @@ with unknown type. There are three cases:
...
-   input we know corresponding unspent output (if we don't,
-   transaction is invalid).
+   legitimacy, compare input and output sums, check fee, etc.
+   These checks are described in more details below.
+
+In the following (until [tx size check section](#tx-size-check)) we
+consider case (3), when all inputs are `TxInUtxo`. The next step is to
+lookup all inputs in UTXO. If at least one input is not found,
+transaction is invalid. We further assume that for each input we know
+corresponding unspent output.  All inputs must exist in UTXO. In the
```
pva701: >  All inputs

two spaces before `All`

docs/txp.md
```diff
@@ -186,16 +186,22 @@ with unknown type. There are three cases:
...
+following we describe the case when all inputs have `TxInUtxo` type
```
pva701: you wrote:
>In the following (until [tx size check section](#tx-size-check)) we consider case (3), when all inputs are `TxInUtxo`.

and then the similar sentence 
>In the following we describe the case when all inputs have `TxInUtxo` type

did you really want to repeat it? though I understand why did you repeat it, it looks awkward.

docs/txp.md
```diff
@@ -0,0 +1,334 @@
...
+  with `SignTx` tag.
+* If witness is `ScriptWitness` (which contains validator and redeemer
+  scripts), we require that address is a `Script` address and its
+  spending data is `Script` spending data with validator script from
+  the witness. Witness itself is valid if few conditions are met.
+  * Both redeemer and validator script have the same version.
+  * If `verifyAllIsKnown` is `True`, script version (which is same for
+    redeemer and validator) must be known. Currently it means that
+    version must be equal to 0, it's the only version known to our software.
+  * Plutus validation script build from redeemer and validator for
```
pva701: I don't get this sentence, maybe you wanted to write
>Plutus validation script **built** from redeemer and validator

docs/txp.md
```diff
@@ -296,12 +301,25 @@ Stakes modification is a bit more complex, but is also very
...
-be assigned to `id`. In case of `MultiKeyDistr` stake will be
-distributed among multiple stakeholders according to specified
-portions. These lists are concatenated and we have two lists of
+of given address.
+* For `BootstrapEraDistr` the value of `TxOut` is distributed among
+bootstrap stakeholders proportional to their weights. The behavior
+depends on whether the value is less than the sum of weights of all
+stakeholders (this sum is called `bootDustThreshold`). If it's greater
+than or equal to `bootDustThreshold`, then each stakeholder receives
+`val / weights_sum` coins (`val` is the value of `TxOut`) and one
```
pva701: I don't get why you mentioned:
>`TxOut` is distributed among bootstrap stakeholders proportional to their weights.

but 
>`val / weights_sum` coins

it should be `val * w_i / weights_sum` coins

docs/txp.md
```diff
@@ -296,12 +301,25 @@ Stakes modification is a bit more complex, but is also very
...
+stakeholder may receive less then their weight and other stakeholders
+won't receive anything.
+* If distribution is `SingleKeyDistr id`, then the value of `TxOut` will
+be assigned to `id`.
+* In case of `MultiKeyDistr` stake will be distributed among multiple
+stakeholders according to specified portions. The first stakeholder
+may receive slightly more than others due to rounding.
+
+Then these lists are concatenated and we have two lists of
 `(StakeholderId, Coin)` pairs: the first one is how much stake each
```
pva701: you wrote
>We have TxOuts corresponding to inputs and outputs of transaction. Each TxOut can be converted to a list of (StakeholderId, Coin) pairs. It depends on 

if so then it should be:
Then these lists are concatenated and we have two lists of (StakeholderId, Coin) pairs: the first one is how much stake each stakeholder should **lose**, the second one is how much stake each stakeholder should **gain**

gromakovsky:
Well, `verifyKnownInputs` sounds fine, will rename.

docs/txp.md
gromakovsky: I think it's quite non-essential, because markdown renderer should take care of that. Anyway, I will fix it here.

docs/txp.md
gromakovsky: I think it should be described in some document about update system, it's not specific to txp. We have the same rules in other places (e. g. there are attributes in blocks and they are checked the same way).

docs/txp.md
gromakovsky: Yes.

docs/txp.md
gromakovsky: Right.

docs/txp.md
gromakovsky: I forgot to delete one of these sentences during previous refactoring of this document.

docs/txp.md
```diff
@@ -45,17 +45,19 @@ associated with the hash of the header of the last applied block
 (i. e. this hash is integral part of this state). So, given a sequence
 of blocks `B₀, B₁, B₂, …` we can compute sequence of states `S₀, S₁,
 S₂, …`. There is also genesis state `G` which takes places before any
-block is applied. This state is derived from genesis data (see ??? TODO).
+block is applied. This state is derived from genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json) for reference).
```
gromakovsky: This file (mainnet-genesis.json) as-is won't be understood by an external reader. My idea was to put a link to the document created in CSL-1688 here, but since CSL-1688 is WIP, I left a TODO.

docs/txp.md
```diff
@@ -64,22 +66,39 @@ tree) and their witnesses.
...
-  System (details are not covered in this document). Txp doesn't
-  modify this value, it only reads it. Some values are relevant for
-  transaction processing:
+* Stakes
+  ```
+  type StakesMap = HashMap StakeholderId Coin
+  type TotalStake = Coin -- this type doesn't exist in code and put here for convenience of reader
+  ```
+  * Stakes are used by other parts of the protocol (for example, leader election
+  process Follow-the-satoshi or FTS)
```
gromakovsky: Sentences should end with commas.

docs/txp.md
```diff
@@ -267,7 +282,7 @@ inputs, tx is considered invalid. Otherwise this check succeeds and
 the difference between sum of inputs and sum of outputs is considered
 to be transaction fee.
 
-#### Fee check
+##### Fee check
```
gromakovsky: When there are 5 levels of nesting, it's likely a good indicator that something should be changed :/

gromakovsky:
@georgeee once you are satisfied with this documentation, I and either @pva701 or @neongreen should make another pass.

docs/txp.md
georgeee: Should they? It's bullet-list. Default formatting I've seen is capital letters at start and no trailing comma/dot

docs/txp.md
```diff
@@ -0,0 +1,434 @@
...
+Witness `PkWitness` contains:
+
+* public key `pk`
+* signature `sig`
+
+We require:
+
+* Address is a `PubKey` address
+* Address spending data is created with public key `pk` (\*)
+* Signature `sig` is a valid signature of transaction (validity checked with `pk`)
```
georgeee: **Important**: need an accurate description of how signature check is made:
 
- Which tag used
- What data included into tag

docs/txp.md
```diff
@@ -0,0 +1,434 @@
...
+
+* redeem public key `pk`
+* redeem signature `sig`
+
+We require:
+
+
+* Address is a `Redeem` address 
+* Address spending data is created with redeem public key `pk`
+* Signature `sig` is a valid signature of transaction (validity checked with `pk`)
```
georgeee: **Important**: need an accurate description of how signature check is made:
 
- Which tag used
- What data included into tag

docs/txp.md
```diff
@@ -0,0 +1,434 @@
...
+    - If the size of transaction is greater than the limit from the adopted `BlockVersionData`, transaction is
+invalid.
+- *Bootstrap era check*: if the block was created during bootstrap era, all
+output addresses must have `BootstrapEraDistr` distribution
+  - Predicate "was created during bootstrap era" is checked with block's slot field
+    and adopted `BlockVersionData` (which contains data which slot is defined last for bootstrap era)
+
+#### Outputs checks
+
+For each output address we perform the following checks:
```
georgeee: @gromakovsky is this outputs check list complete?

I don't see an *important* check that all outputs are different.

CC @neongreen

docs/txp.md
gromakovsky: Oh, sorry, I meant dot. It doesn't matter whether it's a bullet-list or not. These are different sentences, they should be separated by dots. If it's all one sentence, then there shouldn't be capital letters in between.

At least for Russian it's true, not sure about English.

gromakovsky:
I am reading it now and can already mention some TODOs:
1. It should be clearly stated that GState is updated on per-transaction basis. Otherwise one may think that all transactions from `TxPayload` are verified considering the same GState (which is not true).
2. It misses invariants of `TxPayload` type. Maybe they should be described somewhere else, but we don't have this `somewhere else`, so it's better to have them here. And actually there's only one invariant anyway.

docs/txp.md
```diff
@@ -0,0 +1,458 @@
...
+* Currently adopted tx fee policy is unknown
+
+For this check we know:
+
+* Transaction fee `txFee` (from *Sums check*)
+* Transaction size `txSize` (from *Transaction size check*)
+* Currently adopted tx fee policy (as part of `BlockVersionData`), which is of known type
+
+###### Size-linear policy
+Only policy supported by CSL v1.0 is `TxFeePolicyTxSizeLinear`.
```
gromakovsky: I'll change CSL to Cardano SL, because, well, this is document will be read by external reader, I don't think that using our internal abbreviations is a good idea.

docs/txp.md
```diff
@@ -0,0 +1,459 @@
...
+(apart from blocks themselves). But in practice it would be very
+inefficient and inconvenient.
+
+For sake of simplicity, we describe state transition function in two parts:
+* Verification: given GState `S` and a main block `B`, check whether `B` is valid (and can be applied to `S`)
+* Modification: given GState `S` and a main block `B` (successfully verified against `S`), produce `S'`
+
+### GState
+
+In this section we describe parts of GState relevant to transaction processing (GState is an actual structure in code used by few more components).
```
gromakovsky: What do you mean by `actual structure in code`? There is no `GState` type in code (or anything equivalent or similar).

docs/txp.md
gromakovsky: It turns out English rules are simpler and not having dots is valid 😕

docs/txp.md
```diff
@@ -0,0 +1,459 @@
...
+
+We require:
+
+* Address is a `Script` address
+* Address spending data is created with validator script same as `validator`
+* Both `redeemer` and `validator` scripts have the same version.
+* If `verifyAllIsKnown` is `True`, script version (which is same for
+    redeemer and validator) must be known.
+    - For Cardano SL 1.0 it means that version must be equal to 0,
+      it's the only version known to our software.
```
gromakovsky: Btw, we should do one more check here, but we currently don't https://issues.serokell.io/issue/CSL-1818

docs/txp.md
```diff
@@ -0,0 +1,459 @@
...
+2. If `verifyAllIsKnown` is `True`, address' type must be known.
+3. Address can't be a redeem address.
+
+#### Inputs checks
+
+For each input `i` we perform:
+
+- *Inputs known check*: if `verifyAllIsKnown` is `True`, type of input `i` should be known
+- *UTXO check*: If input `i` is of type `TxInUtxo`, it should have corresponding record in UTXO
+- *Difference check*: input `i` must be different from all other inputs in transaction
```
gromakovsky: Note: this particular check isn't performed if there is at least one input with unknown type. I think it should be performed, but it's not.

docs/txp.md
```diff
@@ -0,0 +1,459 @@
...
+
+In addition we perform number of consolidated checks (checks which consider whole set of inputs):
+
+* Witness check (separately for each witness)
```
gromakovsky: Note: this particular check isn't performed if there is at least one input with unknown type. I think it should be performed, but it's not.

gromakovsky:
https://github.com/input-output-hk/cardano-sl/pull/1806#issuecomment-343185044 = FIXED

docs/txp.md
gromakovsky: Done.

docs/txp.md
gromakovsky: Done.

gromakovsky:
Two issues about probably missing checks:
• https://issues.serokell.io/issue/CSL-1818
• https://issues.serokell.io/issue/CSL-1853

docs/txp.md
gromakovsky: Fixed.

docs/txp.md
gromakovsky: Fixed.

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+
+* check that witness corresponds to the address of corresponding input
+* check witness is correct
+
+Witness checks are omitted if (and only if):
+* There is an unknown transaction input.
+
+We describe checks for each type in folowing sections.
+
+###### Public key witness
```
pva701: >`######`

omg, I am sad it's maximal possible depth of `#`
tbh I don't like so deep subsections, the indent doesn't differ from outer subsection 
also, I would write which particular witness you describe, `TxInWitness` in this case.

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+For it, we need to compute minimal required fee for this transaction.
+This value is computed using fixed-precision arithmetic (precision is 1e-9) using formula:
+
+```
+minFee = a + b * txSize
+```
+
+There are two cases check fails (and transaction rejected):
+
+* Minimal fee is negative or is greater than maximal number of coins in the system
```
pva701: Don't you think it's a little bit counter-intuitive to describe when the check is failed instead of when it's passed?

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+
+   ```minFee < 0 || minFee > maxBound @Coin```
+   - It probably indicates there is a mistake in currently adopted fee
```
pva701: Why do you use bullet-list consisting of the one item? Why not to just write it without bullet label?

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+
+- For `BootstrapEraDistr` the value `val` of `txOut` is distributed among
+bootstrap stakeholders proportional to their weights.
+
+    + If `val >= bootDustThreshold`, then each stakeholder receives
+`val * w_ᵢ / weights_sum` coins
+      - one stakeholder also receives the remainder (the choice is deterministc)
+    + Otherwise:
+      - some stakeholders will receive the same stake as their weights
+      - one stakeholder may receive less then their weight
```
pva701: >one stakeholder may receive less **than** their weight

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+Each `txOut` can be converted to a list of `(StakeholderId, Coin)` pairs.
+
+This is being done with considering `AddrStakeDistribution` attribute of address corresponding to `txOut`:
+
+- For `BootstrapEraDistr` the value `val` of `txOut` is distributed among
+bootstrap stakeholders proportional to their weights.
+
+    + If `val >= bootDustThreshold`, then each stakeholder receives
+`val * w_ᵢ / weights_sum` coins
+      - one stakeholder also receives the remainder (the choice is deterministc)
```
pva701: again, don't like one-element bullet list

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+
+Note, that `TxInUtxo` is only input type supported in CSL v1.0. For unknown inputs GState is not modified.
+
+#### Computing stakes for transaction output
+
+Each `txOut` can be converted to a list of `(StakeholderId, Coin)` pairs.
+
+This is being done with considering `AddrStakeDistribution` attribute of address corresponding to `txOut`:
+
+- For `BootstrapEraDistr` the value `val` of `txOut` is distributed among
```
pva701: IMHO this section is described very informally, what are weights? What is `bootDustThreshold` and why is it needed? What does mean `The first stakeholder may receive slightly more than others`?

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+
+#### Stakes modification
+
+Stakes modification is a bit more complex than UTXO modification, but is also very intuitive:
+
+1. Each `TxInUtxo` input corresponds to some `TxOut`, we compute stakes for it
+   * For all inputs, stake lists are concatenated into one, `inStakes`
+   * This list denotes how much stake each stakeholder should lose
+2. Each output is `TxOut`, we compute stakes for it
+   * For all inputs, stake lists are concatenated into one, `outStakes`
```
pva701: > For all **outputs**

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+Local transaction processing is needed for transaction relay. Node can
+receive a standalone transaction from the network and then it needs to
+verify it against its current state (global + mempool) and:
+
+* Apply and relay further (to neighbors)
+* Reject
+
+There are few differences between local txp and global txp:
+
+* In local txp we consider not only GState, but also mempool. We
```
pva701: >In local txp we consider not only GState, but also mempool.

I am not sure this phrase is absolutely clear for a reader because each transaction is part of mempool, I would write **we consider mempool for the moment when tx comes** or smth like that.

docs/txp.md
georgeee: >  I would write which particular witness you describe, TxInWitness in this case.
Elaborate please

docs/txp.md
georgeee: It's better than putting one sentence. Bullet lust means adding more details to broader point, which is the case here

docs/txp.md
georgeee: It's better than putting one sentence. Bullet lust means adding more details to broader point, which is the case here

docs/txp.md
georgeee: Disagree. Here we state that GState is accompanied by smth else, that's it

docs/txp.md
georgeee: Fixed

docs/txp.md
georgeee: Not principal, you can rewrite if you want

docs/txp.md
georgeee: Fixed

docs/txp.md
georgeee: Fixed

docs/txp.md
pva701: I meant we also have `TxWitness`

docs/txp.md
pva701: lol, mne len'

docs/txp.md
```diff
@@ -0,0 +1,486 @@
...
+- *Attributes known check*: if `verifyAllIsKnown` is `True`, all transaction attributes must be known.
+- *Transaction size check*: size of transaction doesn't exceed size limit
+    - Transaction size is computed as number of bytes in serialized `TxAux` (which contains transaction and its witness)
+    - Transaction size limit is part of `BlockVersionData`.
+    - If the size of transaction is greater than the limit from the adopted `BlockVersionData`, transaction is
+invalid.
+- *Bootstrap era check*: if the block was created during bootstrap era, all
+output addresses must have `BootstrapEraDistr` distribution
+  - Predicate "was created during bootstrap era" is checked with block's slot field
+    and adopted `BlockVersionData` (which contains data which slot is defined last for bootstrap era)
```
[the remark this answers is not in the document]
gromakovsky: > are you sure it's correct?

Well, it contains epoch, but from that information you can trivially figure out which slot is last.

docs/txp.md
gromakovsky: Where would you write it?

docs/txp.md
gromakovsky: > I am not sure this phrase is absolutely clear for a reader because each transaction is part of mempool, I would write we consider mempool for the moment when tx comes or smth like that.

Imo, everything is clear. What is incorrect way to interpret it? 🤔

PR #1808, opened 2017-10-22
[CSM-502] Split MonadWalletWebMode

Martoon-00:
Eventually, it's used only in `Launcher.hs` and `(servant) Handlers.hs`

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Methods/Payment.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/Payment.hs
@@ -16,7 +16,8 @@ import           Formatting                     (sformat, (%))
 import qualified Formatting                     as F
 
 import           Pos.Client.KeyStorage          (getSecretKeys)
-import           Pos.Client.Txp.Balances        (getOwnUtxos)
+import           Pos.Client.Txp.Addresses       (MonadAddresses)
+import           Pos.Client.Txp.Balances        (MonadBalances (..))
 import           Pos.Client.Txp.History         (TxHistoryEntry (..))
 import           Pos.Client.Txp.Util            (computeTxFee, runTxCreator)
 import           Pos.Communication              (prepareMTx)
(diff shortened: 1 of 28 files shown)


wallet/src/Pos/Wallet/Web/Mode.hs
```diff
@@ -209,15 +214,9 @@ type MonadWalletWebMode' ctx m =
     , AddrData m ~ (AccountId, PassPhrase)
     )
 
-type MonadWalletWebMode ctx m = MonadWalletWebMode' ctx m
```
pva701: memno :+1:

wallet/src/Pos/Wallet/Web/Mode.hs
```diff
@@ -3,13 +3,13 @@
 {-# LANGUAGE TypeOperators       #-}
 
 module Pos.Wallet.Web.Mode
-       ( WalletWebMode
+       ( WalletMempoolExt
```
pva701: tbh I would like you to remove this export because it may cause dependence of other modules on `Pos.Wallet.Web.Mode` (by mistake or if you aren't familiar with code pretty well). It's better to leave it only in `WalletMode.hs`

Martoon-00: Right

Martoon-00:
In all the history we have `addHistory` which modifies db. `Payment` and `Redeem` also have it.
`Restore` is all about adding new wallet

pva701:
but why didn't you reflect which of them read-only and which write?

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -88,15 +78,15 @@ data MoneySource
     | AddressMoneySource CWAddressMeta
     deriving (Show, Eq)
 
-getMoneySourceAddresses :: MonadWalletWebMode ctx m => MoneySource -> m [CWAddressMeta]
```
Martoon-00: However these 3 getters are worth having modified constraint

pva701:
for example `getTxFee`

pva701:
and `getCurChainDifficulty`

Martoon-00:
Oh really. I need to take a closer look

pva701:
@Martoon-00 ping me (in slack) when you address all issues, I would like to prioritize merge of this PR because of dependance modules on `Mode.hs` doesn't allow to sleep me.

Martoon-00:
:smile:

PR #1810, opened 2017-10-23
Fix build

gromakovsky:


domenkozar:
FYI this impurity will be fixed in DEVOPS-464

PR #1818, opened 2017-10-25
[CSL-1774] Auxx lang proof-of-concept

int-index:
This is an implementation of a language for `cardano-sl-auxx` REPL. Although it doesn't have many features so far, I put a lot of thought into the design of its internals and ended up with a beautiful, extensible design.

It's a dynamically-typed imperative language with strict evaluation, similarly to other scripting languages. Here's a quick overview of its features:

* auto-generated help for built-ins. Since it's a scripting language, many commands are implemented in Haskell and accessed as built-ins, so it's important to make adding new commands easy. To lower the overhead, the `help` command doesn't need to be updated: it automatically detects what arguments each command accepts (using a free applicative trick) and with 0 effort you get a short description of the command, like this:

    ```
    propose-update i: Int
                   block-version: BlockVersion
                   software-version: SoftwareVersion
                   bvm: BlockVersionModifier
                   update: ProposeUpdateSystem*
    ```

  additionally, you can also add a textual description of the command, which will be broken into words and aligned automatically, with word rewrapping (so you need not worry about carefully placing line breaks). In the future it'll be possible to be smart about this and use actual terminal width for word-wrapping (right now the desired width is a constant).

* extremely flexible invocation syntax with positional arguments, keyword arguments, optional arguments, 0+ and 1+ repeated arguments (variadic). Besides, all of these can be mixed in a single command invocation. For instance, if a command `cmd` accepts `x`, `y`, and optionally `z`, you could write `cmd x:1 y:2 z:3`, or `cmd 1 2 3`, or `cmd 1 2`, or `cmd y:2 x:1 z:3`, or `cmd z:3 1 2`, etc.

* a convenient DSL for consuming arguments. As mentioned above, there are many ways to invoke a command. But it doesn't impose any overhead on the implementor: consuming commands is done with a set of simple primitives (`getArg`, `getArgOpt`, etc) in `do` notation (with `-XApplicativeDo`).

* literal syntax for blockchain-related definitions. Besides usual numeric and string literals (with full support for scientific notation like `1.2e+3` in numeral literals and quote escaping like `"a \" b"` in string literals), there are also literals for addresses, public keys, stakeholder ids, block versions, etc. You can just copy-paste a hash and the lexer will recognize it! (also filepaths, just start them with `/` or `./`)

* convenient (but not sloppy) implicit conversions. If a function returns a public key, but a command requires a stakeholder id, it's alright — the `addressHash` function will be invoked automatically. But this behavior is up to the implementor of a command, and possible conversions show up in the help message.

* sequential command execution. Previously, in non-interactive mode there was `T.splitOn "," ` to split multiple commands (ridiculous, right? what if there were commas in command arguments?), and it wasn't even a possibility to execute multiple commands at once in REPL. Now you can separate commands with a semicolon, it's part of the language grammar, and the result will be taken from the last executed command (as in `do` notation).

* subexpressions. One of the main issues I had without a language was that if I needed to pass an address somewhere, I first had to do `listaddr` to get hold of the public key, then copy it as an argument to `addr-distr`, then copy the resulting address and paste it as argument to (let's say) `balance`. Now it's possible to use proper sub-expressions for this, no silly copy-paste required:

    ```
    auxx> balance (addr pk:0 distr:boot)
    11137499999752500
    ```

* a test-suite for the lexer, parser, and argument system. Both unit tests and property tests.


int-index:
@georgeee While you were reviewing, I was writing PR description, so check it out. Regarding examples, here's one:

```
auxx> propose-update i: 0
                     block-version: 1.0.0
                     ~software~cardano-sl:1
                     (bvm unlock-stake-epoch: 3)
                     (upd-bin "linux" /dev/null)
                     (upd-bin "windows" /dev/null)
8fc8180f32a3644ef8654334f8d953d4462601e2682ed4c67a698483944172b9
```

auxx/src/Command/Proc.hs
```diff
@@ -0,0 +1,468 @@
...
+    , cpExec = \addr' -> do
+        addr <- toLeft addr'
+        balance <- getBalance addr
+        return $ ValueNumber (fromIntegral . unsafeGetCoin $ balance)
+    , cpHelp = "check the amount of coins on the specified address"
+    },
+
+    CommandProc
+    { cpName = "bvd"
+    , cpArgumentConsumer = do pure ()
```
gromakovsky: Is `do` preset here intentionally? 🤔

auxx/src/Command/Proc.hs
```diff
@@ -0,0 +1,468 @@
...
+        bvmMaxHeaderSize <- getArgOpt tyByte "max-header-size"
+        bvmMaxTxSize <- getArgOpt tyByte "max-tx-size"
+        bvmMaxProposalSize <- getArgOpt tyByte "max-proposal-size"
+        bvmMpcThd <- getArgOpt tyCoinPortion "mpc-thd"
+        bvmHeavyDelThd <- getArgOpt tyCoinPortion "heavy-del-thd"
+        bvmUpdateVoteThd <- getArgOpt tyCoinPortion "update-vote-thd"
+        bvmUpdateProposalThd <- getArgOpt tyCoinPortion "update-proposal-thd"
+        let bvmUpdateImplicit = Nothing -- TODO
+        let bvmSoftforkRule = Nothing -- TODO
+        let bvmTxFeePolicy = Nothing -- TODO
```
gromakovsky: slojnaaa

int-index: I decided to leave these as `TODO` for now because it wasn't possible to specify these parameters in the previous version, so it's something for a separate issue.

auxx/src/Command/Proc.hs
int-index: not really, just an artifact of how I wrote code (started with `do pure ()` and extended as necessary)

auxx/src/Command/TyProjection.hs
```diff
@@ -0,0 +1,166 @@
+module Command.TyProjection
```
gromakovsky: This module has only one comment which is about some hack. Can you at least write a сomment to the whole module here?
I see that most of the definitions are trivial, but the what it all means is not clear for an unprepared reader.

int-index: I must document other modules as well, but not in this sprint probably (writing good docs takes time). In short, this module is used for matching runtime values to Haskell types. If you have `ValueNumber 12.5`, can it be converted to an `Int`? (no, because of the fractional part). But `ValueNumber 4` can. And `TyProjection Int` is what does the conversion (it also contains the type name for an error message)

core/Pos/Crypto/Signing/Types/Signing.hs
```diff
@@ -111,6 +111,12 @@ parseFullPublicKey s = do
     b <- Base64.decode s
     PublicKey <$> first fromString (CC.xpub b)
 
+-- | Parse 'PublicKey' from base16 encoded string.
+parseFullPublicKeyHex :: Text -> Either Text PublicKey
```
gromakovsky: No-no-no, please, let's stop it. I find it very bad that we don't have a single standard way of displaying a public key.
Base64 is currently used in JSON format. Base16 is currently used in scripts.
Unfortunately it seems that we can't change any of these two. And I propose to treat base64 as the standard way of displaying public keys and treat base16 as «only for scripts».
Nobody uses scripts nowadays and probably they won't be used for a long time, while canonical json contains genesis data which is quite useful in some cases.

core/Pos/Util/Justify.hs
```diff
@@ -0,0 +1,51 @@
+module Pos.Util.Justify
+       ( leftAlign
+       ) where
+
+import           Universum          hiding ((<>))
+
+import qualified Data.List          as List
+import qualified Data.List.NonEmpty as NonEmpty
+import           Data.Semigroup
```
gromakovsky: Implicit imports are still prohibited :kot:

core/Pos/Util/Justify.hs
```diff
@@ -0,0 +1,51 @@
...
+initLine w = Line (Text.length w) (w :| [])
+
+appendLine :: Line -> Line -> Line
+appendLine line1 line2 =
+  Line
+    (lineWidth line1 + 1 + lineWidth line2)
+    (lineWords line1 <>    lineWords line2)
+
+nonEmptyTails :: NonEmpty a -> NonEmpty [a]
+nonEmptyTails = NonEmpty.fromList . List.tail . List.tails . NonEmpty.toList
```
gromakovsky: What about https://www.stackage.org/haddock/lts-9.1/base-4.9.1.0/Data-List-NonEmpty.html#v:tails?

auxx/test/Test/Auxx/Lang/ArgumentSpec.hs
```diff
@@ -0,0 +1,68 @@
...
+acExample1 = do
+    x <- getArg tyString "s"
+    y <- getArg tyInt "i"
+    return (x, y)
+
+tyString :: TyProjection String
+tyString = TyProjection "String" (preview _ValueString)
+
+tyInt :: TyProjection Int
+tyInt = TyProjection "Int" (toBoundedInteger <=< preview _ValueNumber)
```
gromakovsky: Is it intentionally copy-pasted?

int-index: No, I forgot to un-copy-paste when changed module structure

core/Pos/Util/Justify.hs
int-index: Oops, missed this one.

auxx/src/Command/Proc.hs
```diff
@@ -0,0 +1,468 @@
...
+               \ can be specified manually (by default it uses the current epoch \
+               \ to determine whether we want to use bootstrap distr)"
+    },
+
+    CommandProc
+    { cpName = "addr-hd"
+    , cpArgumentConsumer = getArg tyInt "i"
+    , cpExec = \i -> do
+        sks <- getSecretKeysPlain
+        sk <- evaluateWHNF (sks !! i)
```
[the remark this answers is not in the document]
int-index: Because the secret key has no NFData instance and WHNF is enough to force errors from (!!). I would prefer NF here as well.

auxx/src/Lang/Name.hs
```diff
@@ -0,0 +1,54 @@
...
+newtype Letter = Letter { getLetter :: Char }
+    deriving (Eq, Ord, Show)
+
+unsafeMkLetter :: Char -> Letter
+unsafeMkLetter = Letter
+
+instance Arbitrary Letter where
+    arbitrary = Letter <$> arbitrary `suchThat` isAlpha
+
+newtype Name = Name (NonEmpty (NonEmpty Letter))
```
gromakovsky: Why do we need such representation? Why do we treat `patak-bardaq` as `[patak, bardaq]` and not just `patak-bardaq`?

auxx/src/Lang/Parser.hs
```diff
@@ -0,0 +1,94 @@
...
+tok p = terminal (preview p)
+
+inBrackets
+    :: Getting (First ()) Token BracketSide
+    -> Prod r e Token a
+    -> Prod r e Token a
+inBrackets p r =
+    tok (p . _BracketSideOpening) *> r <* tok (p . _BracketSideClosing)
+
+gExpr :: Grammar r (Prod r Text Token Expr)
```
gromakovsky: I think it would be good to have a comment with this grammar in some more common convention, because this one is hard to read for people not familiar with `Earley` library (or least me). I guess it won't take much time to write it. And it's ok to do it in another PR which adds documentation to the code (AFAIU, there will be one at some point).

auxx/src/Plugin.hs
```diff
@@ -60,20 +67,45 @@ auxxPlugin AuxxOptions{..} = \case
...
+    printAction " " -- for exit by SIGPIPE
     liftIO $ hFlush stdout
 #if !(defined(mingw32_HOST_OS))
     delay $ sec 3
     liftIO $ exitImmediately ExitSuccess
 #endif
+  where
+    printAction = putText
+
+withValueText :: Monad m => (Text -> m ()) -> Lang.Value -> m ()
```
gromakovsky: Why did you do it this way instead of defining `instance Buildable Value` and simply using `printAction value`? It would like more conventional to me.
Is it just to conveniently  format `ValueList`? I guess it should be a big problem to format a list of values separated by spaces :/
And btw there is `listBuilder` in `Serokell.Util.Text` :)
(dinosaur)

auxx/src/Lang/Interpreter.hs
```diff
@@ -0,0 +1,75 @@
...
+eval :: Expr -> T m Value
+eval = \case
+    ExprUnit -> return ValueUnit
+    ExprLit l -> return (literalToValue l)
+    ExprGroup exprs -> evalExprGroup exprs
+    ExprProcCall procCall ->
+        evalProcCall =<< traverse eval procCall
+
+evalExprGroup :: NonEmpty Expr -> T m Value
+evalExprGroup (x :| xs) = case nonEmpty xs of
```
gromakovsky: Is it equivalent to `NE.last . traverse eval`?

core/Pos/Crypto/Signing/Types/Signing.hs
int-index: base16 is used in logs, which is why I thought it was the default. I'll change the language to use base64, but then it would be more convenient to change the logs as well.

gromakovsky: Ah, in `Buildable`, yes, I see. It's really bardaq :(

core/Pos/Util/Justify.hs
int-index: Not the same, it'll have an extra element in front

auxx/src/Lang/Name.hs
int-index: it guarantees the correct structure by code, not by a comment, so we get generic methods (i.e. `genericShrink`)

auxx/src/Lang/Parser.hs
int-index: You can read everything to the right of `rule` as regular parser combinators (with an `Applicative` interface), but you don't need to worry about left recursion.

auxx/src/Plugin.hs
int-index: it was supposed to be a dirty and quick solution for printing :) ideally I want printed values to be valid language expressions (so one could copy REPL output and paste it back as input), and this requires a formatter for `Expr` (it's on my TODO list)

auxx/src/Lang/Interpreter.hs
int-index: yes but it's single-pass. I'll add a comment

core/Pos/Util/Justify.hs
gromakovsky: I think it would be good to write a comment about it, because otherwise other people may also have this question when they see this code.

auxx/src/Lang/Name.hs
gromakovsky: I am not sure you got me right.
My question is: suppose we had `Name = Name (NonEmpty Letter)`, would it make some existing functionality impossible? Suppose `Letter` allowed `-`, then `patak-bardaq` would be `Name "patak-bardaq"` (pseudo-code), would it be essentially different from `Name ["patak", "bardaq"]` (which we have now) somewhere?

auxx/src/Plugin.hs
gromakovsky: Okay, can you create an issue about all the things you are going to do (write comments, implement a formatter for `Expr`, etc.)?

auxx/src/Lang/Lexer.hs
```diff
@@ -216,9 +216,9 @@ pAddress = do
 
 pPublicKey :: Lexer PublicKey
 pPublicKey = do
-    str <- pSomeAlphaNum
+    str <- (<>) <$> takeP (Just "base64") 86 <*> string "=="
```
gromakovsky: I don't get this line, can you clarify?

auxx/src/Lang/Name.hs
int-index: Then we'd get names that start with a hyphen, end with a hyphen, and contain multiple hyphens sequentially

auxx/src/Lang/Lexer.hs
int-index: Read 86 symbols that end with `==`, then we parse it as base64. I tested

auxx/src/Lang/Name.hs
gromakovsky: Is it bad? :)

auxx/src/Lang/Lexer.hs
gromakovsky: Ah, I forgot that public key has fixed length, ok then.

auxx/src/Lang/Name.hs
int-index: I'd like to have more free lexical space for future additions, so broadly speaking yes. For instance `--` could be used for comments or decrement

auxx/src/Plugin.hs
int-index: Yes, I have them in orgmode at the moment. But I'd rather create separate issues, not one

gromakovsky: Okay, I will approve this PR after you create them.

int-index: Created (see YT)

PR #1824, opened 2017-10-26
[CSL-1785] Get rid of too much copy-paste in `Pos.Explorer.Txp.Local`

pva701:


explorer/src/Pos/Explorer/Txp/Local.hs
```diff
@@ -146,68 +121,17 @@ eTxProcessTransactionNoLock itw@(txId, txAux) = reportTipMismatch $ runExceptT $
...
-    case pRes of
-        Left er -> do
-            logDebug $ sformat ("Transaction processing failed: " %build) txId
-            throwError er
-        Right _ ->
-            logDebug
-                (sformat ("Transaction is processed successfully: " %build) txId)
+    pure $
+        EProcessTxContext
+        { _eptcExtraBase = eet
```
ksaric: How about just adding an additional type to `ProcessTxContext`? Something like:
```
data ProcessTxContext a = ProcessTxContext
    { _ptcAdoptedBVData :: !BlockVersionData
    , _ptcUtxoBase      :: !Utxo
    , _ptxAdditional      :: !a
    }
```
Wouldn't that remove the additional `pctx` from `txProcessTransactionAbstract` since you could pass `ext` directly?

pva701: your suggestion makes sense but I don't get how it helps to remove `ptcx`,
`ext = ExplorerExtra` but `pctx = ExplorerExtraTxp`. so if it doesn't help I don't see advantages, though, it can remove several lines of code.

ksaric: I thought `pctx` was `ProcessTxContext` or `EProcessTxContext`. Ok.

pva701: or, sorry, you are right, I meant `a=ExplorerExtraTxp` (`a` from your definition of `ProcessTxContext`)

txp/Pos/Txp/Logic/Local.hs
```diff
@@ -54,7 +60,7 @@ makeLenses ''ProcessTxContext
 instance HasLens Utxo ProcessTxContext Utxo where
     lensOf = ptcUtxoBase
 
--- Base monad for tx processing in.
+--- Base monad for tx processing in.
```
gromakovsky: wut

pva701: lool, forgot to remove

explorer/src/Pos/Explorer/Txp/Local.hs
ksaric: Right, but you can have multiple cases since the base structure is the same?
For example:
- in one case `a=ExplorerExtraTxp`
- in another case `a=()` - the _default/base_ structure

pva701: but why it's better than current approach? it won't reduce the number of type parameters.

ksaric: > Wouldn't that remove the additional `pctx` from `txProcessTransactionAbstract` since you could pass ext directly?

Not on the structure itself, but I thought that it would remove it from the function. Since `a = ext` we don't need another type parameter around. At least, that's what I thought, but I didn't try to compile it.

gromakovsky: @ksaric it seems you confused `ExplorerExtra` and `ExplorerExtraTxp`.

pva701: I used your way b/c it reduces number of redundant instances and add opportunity to add new field to `ProcessTxContext` without duplication

PR #1827, opened 2017-10-26
[DEVOPS-463] add a simpler API for hash-installer

cleverca22:


jmitchell:
@cleverca22: I recommend enabling stylish haskell as explained at https://github.com/input-output-hk/cardano-sl-style-guides/blob/master/haskell-style-guide.md

auxx/src/HashInstaller.hs
```diff
@@ -0,0 +1,17 @@
+module Main where
+
+import qualified Data.ByteString.Lazy as BSL
+import           Formatting           (sformat, string, (%))
+import           Pos.Crypto           (Hash, SignTag (SignUSVote),
+                                       emptyPassphrase, encToPublic, hash,
+                                       hashHexF, safeSign, unsafeHash,
+                                       withSafeSigner, withSafeSigners)
```
deepfire: Most of those imports are not require, I guess..
Also maintaining them is a pain, so just using qualified names might help.

cleverca22: i'm rebuilding with `-Wall` to see what is extra

auxx/src/HashInstaller.hs
```diff
@@ -0,0 +1,14 @@
+module Main where
+
+import qualified Data.ByteString.Lazy as BSL
+import           Formatting           (sformat)
+import           Pos.Crypto               (hashHexF)
```
jmitchell: Why did the spacing end up weird here even after using stylish? Maybe there's a tab.

cleverca22: i think thats a refactor i did after stylish, re-running it

jagajaga:
It should be in `auxx`.

cleverca22:
auxx is overly strict about having every parameter present and valid, even ones that have zero impact on the hash
`cardano-auxx cmd --commands "hash-installer Daedalus-installer-1.0.3253.pkg" --system-start 0`
with the requirement that `node/configuration.yaml` exists relative to the working directory, or you also specified `--configuration-file`
vs
`cardano-hash-installer Daedalus-installer-1.0.3253.pkg` with no extra flags and no expectations on other files being present

georgeee:
@cleverca22 anyway it's not good to duplicate code and create many single-purpose executables. It's not very hard to adjust `cardano-auxx` according to your request (to have less params) and this way should be preferred

cleverca22:
yeah, that will also work

PR #1828, opened 2017-10-26
[CSL-1802] Fix compatibility with mainnet

gromakovsky:
See the last commit message for details.

### DIFF
--- a/core/Pos/Core/Block.hs
+++ b/core/Pos/Core/Block.hs
@@ -34,12 +34,14 @@ module Pos.Core.Block
 
 import           Universum
 
-import           Control.Lens         (makeLenses)
-import           Control.Monad.Except (MonadError (throwError))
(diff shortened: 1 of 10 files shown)


georgeee:
>  Note: we can't use 'VssCertificatesMap', because we serialize it as
> a 'HashSet', but in the very first version of mainnet this map was
> serialized as a 'HashMap' (and 'VssCertificatesMap' was just a type
> alias for that 'HashMap').

Won't `VssCertificatesMap` be an incompatibility itself? @gromakovsky

gromakovsky:
> Won't VssCertificatesMap be an incompatibility itself?

What do you mean, I don't get it? AFAIK, we don't store it in acid-state for example, so changing types shouldn't matter as long as we don't change binary formats 🤷‍♂️

PR #1829, opened 2017-10-27
[CSL-1399] [CSL-1695] block decode bench

avieth:
Compares block deserializing with checked versus unchecked input, and
also in cases where the decoded value is not forced, because forcing it
will induce even more work.

Try running it yourself: `./bench/Serialization/Block +RTS -T -RTS --regress allocated:iters`.
It shows an improvement with the changes in volhovm/csl1399-unsafe-deserialize, but an even bigger improvement (deserialize time cut in half, allocations cut two thirds) when we don't verify nor force the deserialized block (verifying forces the block, hence no change).

This is *almost* immediately applicable: when serving a block, neither verify nor force the `Block` value that's deserialized from the database. But time-warp doesn't allow that: it demands a `MsgBlock` and will re-serialize it before sending, thereby forcing the `Block` etc. etc.
Ideally we could just throw the already-serialized representation (from the DB) down to the peer without re-serializing. How to do this while maintaining some sort of type safety could be tricky.

Slightly unrelated note: I found that using the `Arbitrary` instance was by far the easiest way to get a hold of a valid `Block`. `genBlocks` is simply unusable as it comes with so much overhead and class constraints and side-effects. I have to invent a new "mode" or perhaps steal the `AuxxMode`... Is there no simple pure way to get a `Block`?


bench/Serialization/Block.hs
```diff
@@ -0,0 +1,73 @@
...
+  generateBlock =
+    loggerBracket loggingParams $
+    runProduction $ do
+      let block :: Block
+          !block = Right arbitraryBlock
+          !serialized = serialize' block
+      return (block, serialized)
+
+  arbitraryBlock :: HasConfigurations => MainBlock
+  arbitraryBlock = unGen arbitrary (mkQCGen 0) 10000
```
avieth: By adjusting the last argument we can get blocks of different sizes. I played with it and found that choosing 10 gives a block that serializes to 1.8kb. Running that through the tests, the unchecked version is nearly 8 times faster. Forcing the value seems to have little effect, probably because the merkle tree is so small.

gromakovsky:
> Try running it yourself: ./bench/Serialization/Block +RTS -T -RTS --regress allocated:iters.

How do I build it?

gromakovsky:
>  Is there no simple pure way to get a Block?

There are `Pos.Block.Core.Main.Misc.mkMainBlock` and `Pos.Block.Core.Genesis.Misc.mkGenesisBlock`. Not sure if it's simple enough, but if you need something simpler, then use `arbitrary` as you do.

avieth:
> How do I build it?

```
stack install criterion
stack ghc bench/Serialization/Block.hs
```

gromakovsky:
> Ideally we could just throw the already-serialized representation (from the DB) down to the peer without re-serializing. How to do this while maintaining some sort of type safety could be tricky.

We have `AsBinary` newtype in `Pos.Binary.Class.Primitive`. It's just a wrapper over `ByteString` with phantom parameter. I am not sure it will be directly usable, but I think that eventually we should make it possible to send already-serialized representation without calling any decoding function.
But we still need to have unchecked deserialization, because sometimes we want to fetch some data from DB not to just send it somewhere, but to use it in a structured form.

avieth:
> We have AsBinary newtype in Pos.Binary.Class.Primitive. It's just a wrapper over ByteString with phantom parameter. I am not sure it will be directly usable, but I think that eventually we should make it possible to send already-serialized representation without calling any decoding function.

Oh nice, I'll check it out.

> But we still need to have unchecked deserialization, because sometimes we want to fetch some data from DB not to just send it somewhere, but to use it in a structured form.

I'm of the opinion that all deserialization should be unchecked. That's to say, it should only check that the bytes are of the right form; that they parse to *some* value of the given type, but should *never* check that the value conforms to some extra constraints. That's not the business of the encoder/decoder.

volhovM:
@avieth please continue development of this benchmark in `volhovm/csl1399-unsafe-deserialization`, I've moved your changes there.

PR #1832, opened 2017-10-28
Link daedalus bridge module in daedalus-bridge.sh

StephenCharles:
According to the documentation https://cardanodocs.com/for-contributors/building-from-source/ you should be able to install the daedalus-client-api module after running daedalus-bridge.sh. I found that I had to explicitly run `npm link` to get it to work. So lets add it to the script.

### DIFF
--- a/scripts/build/daedalus-bridge.sh
+++ b/scripts/build/daedalus-bridge.sh
@@ -20,3 +20,6 @@ echo "3. Building Bridge..."
 cd daedalus
 npm install
 npm run build:prod
+
+echo "4. Linking Bridge..."
+npm link


scripts/build/daedalus-bridge.sh
```diff
@@ -20,3 +20,6 @@ echo "3. Building Bridge..."
 cd daedalus
 npm install
 npm run build:prod
+
+echo "4. Linking Bridge..."
+npm link
```
Martoon-00: As for me, I'm glad that `npm link` was absent here, because it fails for me when done without `sudo`, while it doesn't for others.
However I have little authority to decide on this

StephenCharles: Hmm well if that's the case for many people, maybe adding a note to the documentation would be better.

akegalj: I also believe adding note should be better. There was a separate script that does:
```
daedalus-bridge.sh
npm link
```
which was used just for this purpose, but I don't know when and why it was removed.

@DominikGuzei @darko-mijic do you run `npm link` explicitly or do you use some script from daedalus repo that handles this?

Instead of this proposed solution I would create daedalus-bridge-link.sh that does what is described above.

denisshevchenko:
I just ran `./scripts/build/daedalus-bridge.sh` from the latest `master` - it works, without `npm link`...

StephenCharles:
Did you already have the bridge linked locally? Seems like it if you never had to run `npm link`

denisshevchenko:
> Did you already have the bridge linked locally? 

No. I did on the clean computer (Arch Linux 64bit).

StephenCharles:
Interesting, I also am using Arch 64 bit. Can you verify that it is linked? For me I ran into errors running the daedalus wallet locally, until I ran `npm link` explicitly in the dadalus bridge repo.

int-index:
@denisshevchenko Linking `daedalus-bridge` is a step for building Daedalus itself. The bridge will build successfully without `npm link`. But building Daedalus will fail, because it won't find `daedalus-bridge`.

StephenCharles:
I realize this PR is quite low priority compared to others but if anyone has a preference for what should be done, I will happily do it. Otherwise I will happily close this ahah.

ksaric:
@StephenCharles I think it's a good idea. The alternative is to send some sort of argument with `--link` and link then, but since the whole point of this bridge is to enable the (JS) communication, I don't see why we need to complicate with any arguments.
@akegalj Can you approve, I think it's good for as long as we are using the bridge - I know I myself spent some time trying to figure out what's missing.

There is a script on the Daedalus side that links from this - https://github.com/input-output-hk/daedalus/blob/master/scripts/link-bridge.sh

DominikGuzei:
@StephenCharles @ksaric you can close any PR related to the purescript bridge as the frontend is not using it anymore 😉 probably you should setup a PR to delete it from `cardano-sl`

StephenCharles:
OO interesting. So just to clarify, the frontend does not use the bridge anymore, so I can delete the `daedalus` folder located at root level? Also is this the pr that removed the dependency of the daedalus-bridge? https://github.com/input-output-hk/daedalus/pull/557/files

DominikGuzei:
Yeah @StephenCharles, this PR removed the final ("soft") dependency on the bridge. I guess you can delete the whole `daeadalus` folder but better discuss that with @akegalj to be sure 😉

DominikGuzei:
@StephenCharles the only thing to be careful about: the current official release of Daedalus + cardano-sl does still use the bridge, so im not 100% sure how safe it is already to delete it completely.

StephenCharles:
Alright I'm going to play around with this locally.

gromakovsky:
See #2016

StephenCharles:
Closing as bridge is being removed see: https://github.com/input-output-hk/cardano-sl/pull/2022

PR #1833, opened 2017-10-28
[CSL-1691] Check protocolMagic in VerInfo

pva701:


wallet/node/Main.hs
```diff
@@ -139,6 +139,7 @@ action (WalletNodeArgs (cArgs@CommonNodeArgs{..}) (wArgs@WalletArgs{..})) =
     conf :: ConfigurationOptions
     conf = CLI.configurationOptions $ CLI.commonArgs cArgs
 
+
```
gromakovsky: Are you serial? Why do you commit it every time? Also `stack clean cardano-sl-wallet` works fine.

PR #1836, opened 2017-10-28
[CSM-531] MERGE ONLY IN CASE OF EMERGENCY Revert "[CSM-491] Add timeout reports in blistener"

Martoon-00:
After addition of these logs we started to observe the issue, when after
new epoch start BListener application hang, mentioned logging emitted infinite
warnings about timeout.

Further investigation showed strange behaviour of `logWarningWaitInf`
function. It forks a thread via `withAsync` which periodically drops warnings
and this thread gets killed when primary action completes.
However, in our case, looks like `uninterruptibleCancel` hangs trying to kill
that forked thread, which in turn occurs to be on `delay` at that moment.
(And all these happens only upon a new epoch for some reason)

I hope someone explains me how could it be. Till that moment or some progress in investigation, I remove this logging-on-timeout.

### DIFF
--- a/infra/Pos/Slotting/Util.hs
+++ b/infra/Pos/Slotting/Util.hs
@@ -7,7 +7,6 @@ module Pos.Slotting.Util
        , getSlotStart
        , getSlotStartPure
        , getSlotStartEmpatically
-       , getCurrentEpochSlotDuration
        , getNextEpochSlotDuration
        , slotFromTimestamp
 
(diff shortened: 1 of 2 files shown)


gromakovsky:
I don't think we need this hotfix in master.
1. There shouldn't be such problem in master, because there we don't call BListener during LRC.
2. Even if there was such problem, I think we can afford to take some time to fix stuff in master.

Note, however, that in `cardano-sl-1.0` this problem does exist until #1772 is merged. So I am going to reopen #1835 and close this PR.

PR #1838, opened 2017-10-29
[CSL-1794] Update versions in configuration

gromakovsky:
We are going to propose and vote for an update to block version 0.1.0.
We will also need to set application version to 3 for csl-daedalus.
Previosuly cardano-sl software was launched with version 2, but it was by
mistake, because last confirmed version was 0. This mistake didn't have any
impact though. I changed it to 0 to be correct.
Wallet configs now don't reuse `mainnet_base_update`, but instead define
all versions explicitly, because the previous solution was error-prone.

### DIFF
--- a/node/configuration.yaml
+++ b/node/configuration.yaml
@@ -279,10 +279,10 @@ mainnet_base: &mainnet_base
 
   update: &mainnet_base_update
     applicationName: cardano-sl
-    applicationVersion: 2
+    applicationVersion: 0
     lastKnownBlockVersion:
       bvMajor: 0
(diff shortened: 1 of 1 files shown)


cleverca22:
```
test/Test/Pos/Configuration.hs:16:14: error:
    • Couldn't parse /tmp/nix-build-cardano-sl-1.0.2.drv-0/node/configuration.yaml: Unknown alias `mainnet_base_update`
```

gromakovsky:
Thanks for noticing, fixed (hopefully). Will keep an eye on CI.

PR #1839, opened 2017-10-29
[CSM-531] Allow timeout-logging to work under `uninterruptibleMask`

Martoon-00:


stack.yaml
```diff
@@ -55,7 +55,7 @@ packages:
   extra-dep: true
 - location:
     git: https://github.com/serokell/time-warp-nt.git
-    commit: 48cb189a412c5a606888285944b9a8f32fcd0d30 # master
+    commit: c8deb5dd65acd1f4a0fd3a106cb08fe6baecbd20 # master
```
Martoon-00: Can't merge PR until commit here appears in `time-warp-nt:master`

infra/Pos/Util/TimeLimit.hs
```diff
@@ -60,7 +61,10 @@ logWarningLongAction secure delta actionTag action =
     -- 'withAsync' is assumed to take care of this, and indeed it does for
     -- 'Production's implementation, which uses the definition from the async
     -- package: 'uninterruptibleCancel' is used to kill the thread.
-    withAsync (waitAndWarn delta) (const action)
+    --
+    -- thinking even more about it, unmasking auxilary thread is crucial if
+    -- this function is going to be called under 'mask'.
```
avieth: Shouldn't it be OK so long as it's not called in an *uninterruptible* masking state?

Where are the call sites in which `logWarningLongAction` is called in masked state?

gromakovsky: https://issues.serokell.io/issue/CSL-1837

`applyBlocks` calls `onApplyBlocks` from `BListener` which uses `logWarningLongAction` in wallet's implementation (`onApplyBlocksWebWallet`).

Martoon-00: Actually, I wanted it to work under `MaskedUninterruptibly` state only because in future there accidentally could be a similar case, when someone calls `bracket` or smth other big piece of code, and later someone another inserts innocent `logWarningWait`...

PR #1841, opened 2017-10-30
Add mainnet staging configuration with short epoch

gromakovsky:


gromakovsky:
Note: `startTime` is arbitrary, should be set to appropriate value, but I don't know when it's gonna be deployed.

PR #1844, opened 2017-10-30
CSL-1806 Add new features to genesis creation necessary for testnet

gromakovsky:
* [X] Remove `TestnetCustomStakeDistr`
* [X] Add `AvvmBalanceFactor`
* [X] Add heavyDlg boolean parameter to mark that heavyweight delegation certificates should also be generated for richmen
* [X] Allow to optionally specify --configuration-seed parameter for genesis generation (which will override one from genesis-spec )

During my work on part 3 (genesis heavy delegation) I also performed a refactoring because I didn't like some code. Specifically:
1. I introduced `RichSecrets` type instead of `(SecretKey, EncryptedSecretKey, VssKeyPair)` because this tuple was used in many place and I think it was a good sign to wrap it into a data type. Also I got rid of a function which was returning 5-elements tuple, because I think 5-elements tuples should be avoided.
2. Now only necessary data is generated for different types of nodes. For rich nodes we generate primary key and VSS key. For poor nodes we generate only `EncryptedSecretKey`s. Previously we had a mess, all this data was generated for all nodes despite being unnecessary.
3. `userSecretWithGenesisKey` and `updateUserSecretVSS` functions were used together (only in one place) and had similar logic, so I merged them into single function.
4. `mkGenesisDelegation` function was a bit ridiculous. For example, it was converting a map to list and then list to the same map. Also it was basically checking that keys in `HashMap` are different. I removed those checks and added another function. Now we have `mkGenesisDelegation` and `recreateGenesisDelegation`.
5. Some code in `keygen` has been refactored accordingly (e. g. we dump only appropriate data for rich and poor nodes).

So the fourth commit should be reviewed carefully.


gromakovsky:
It's finished. @pva701 please review.

core/Pos/Core/Configuration.hs
```diff
@@ -139,6 +142,12 @@ withGenesisSpec theSystemStart conf@CoreConfiguration{..} val = case ccGenesis o
                 GeneratedGenesisData {..} =
                     generateGenesisData (gsInitializer spec) maxTnBalance
 
+                applyAvvmBalanceFactor :: HashMap k Coin -> HashMap k Coin
+                applyAvvmBalanceFactor = map (applyCoinPortionDown ggdAvvmBalanceFactor)
+                finalAvvmDistr :: GenesisAvvmBalances
+                finalAvvmDistr =
+                    ggdAvvm <> gsAvvmDistr spec & over coerced applyAvvmBalanceFactor
```
pva701: `over coerced` soo clever

lib/test/Test/Pos/Lrc/WorkerSpec.hs
```diff
@@ -77,6 +77,8 @@ genGenesisInitializer = do
     tiTestBalance <- genTestnetBalanceOptions
     tiFakeAvvmBalance <- arbitrary
     tiAvvmBalanceFactor <- arbitrary
+    -- Currently these tests don't work well with genesis delegation.
```
pva701: :loh:

PR #1847, opened 2017-11-01
[CSL-1279] Part IV: Final SSC cleanup

neongreen:
In this PR:

* `GodTossing` was removed from all modules.
* All types and functions with `gt` or `godtossing` were renamed.
* Four useless SSC-related classes were removed. There are none left.
* `SscBlock` was made a type instead of a newtype.
* The module hierarchy was changed a bit (look at cardano-sl-ssc.cabal).
* Unused `Pos.Ssc.GodTossing.Genesis.*` modules were removed.
* @gromakovsky was added to the owners of `ssc/`.
* The `--ssc-algo` parameter was removed.


ssc/Pos/Ssc/GState/BlockLogic.hs
```diff
@@ -0,0 +1,255 @@
+{-# LANGUAGE RankNTypes #-}
+
+module Pos.Ssc.GState.BlockLogic
```
gromakovsky: We usually call such modules `VAR` (verification + application + rollback).

ssc/Pos/Ssc/Message.hs
```diff
@@ -0,0 +1,111 @@
...
+    toSscTag _ = OpeningMsg
+
+instance HasSscTag MCShares where
+    toSscTag _ = SharesMsg
+
+instance HasSscTag MCVssCertificate where
+    toSscTag _ = VssCertificateMsg
+
+-- TODO: someone who knows networking should take a look because this really
+-- doesn't look like something that anyone should ever have to write
```
gromakovsky: :)

ssc/Pos/Ssc/GState/BlockLogic.hs
neongreen: thanks, I will rename the module

lib/configuration.yaml
```diff
@@ -104,7 +104,7 @@ dev: &dev
       bvAlt: 0
     systemTag: system
 
-  # GodTossing
+  # SSC (formerly known as GodTossing)
   gt: &dev_gt
```
gromakovsky: I am reviewing commit with message `Rename everything with “gt” or “godtossing”` and I see `gt` :suspect:

gromakovsky: Why not rename to `ssc`?

neongreen: I didn't rename it because I thought that it would break things. (For the same reason I didn't rename the DB key used for SSC data.) Now that you have mentioned it, I think that it won't break things so I'll rename it.

lib/src/Pos/Web/Api.hs
```diff
@@ -65,28 +65,27 @@ type HealthCheckApi =
 baseNodeApi :: Proxy BaseNodeApi
 baseNodeApi = Proxy
 
--- | GodTossing specific API.
-type GodTossingApi =
+-- | SSC specific API.
```
gromakovsky: These APIs should be merged. Previously we had «base» API and GodTossing-specific API. They were separated, because SSC wasn't equivalent to GodTossing. Now it is, so it doesn't make sense to keep them separated.

lib/configuration.yaml
gromakovsky: It's better not to rename DB key indeed, but configuration format can be changed without breaking anything, imo.

lib/src/Pos/Web/Api.hs
```diff
@@ -65,28 +65,27 @@ type HealthCheckApi =
...
+sscApi = Proxy
 
--- | Servant API which provides access to full node internals with
--- GodTossing SSC.
-type GtNodeApi =
+-- | Servant API which provides access to full node internals with SSC.
+type SscNodeApi =
     BaseNodeApi
     :<|>
     "god_tossing"
```
gromakovsky: Also there is no need to prefix `ssc` api with `god_tossing`.
Don't worry about breaking things, this server is just a swalka which is not used for anything significant.

lib/configuration.yaml
gromakovsky: Btw, DB key was and is `ssc`, so there is no need to change it anyway.

gromakovsky:
```
lib/src/Pos/Behavior.hs:    , bcSscBehavior    :: !SscBehavior       -- ^ SSC (GodTossing)
lib/src/Pos/Behavior.hs:        bcSscBehavior    <- o A..: "godTossing"
ssc/Pos/Ssc/Behavior.hs:-- godTossing:
```

Rename it as well maybe?

gromakovsky:
```
wallet/node/Main.hs:import           Pos.Web              (serveWebGT)
wallet/node/Main.hs:    plugins = mconcat [ convPlugins (pluginsGT wArgs)
wallet/node/Main.hs:pluginsGT ::
wallet/node/Main.hs:pluginsGT WalletArgs {..}
wallet/node/Main.hs:    | enableWeb = [serveWebGT webPort (Just walletTLSParams)]
wallet/node/Main.hs:        let sscParams = CLI.gtSscParams cArgs vssSK (npBehaviorConfig currentParams)
```

Something obsolete, there shouldn't be `GT` in the name, it's not a car.

gromakovsky:
`initGtDB` can be renamed. Also `rollbackGT`, `rollbackGTDo`.

lib/configuration.yaml
neongreen: No, some other DB key.

ssc/Pos/Ssc/GState.hs
```diff
@@ -0,0 +1,342 @@
...
+    -- calculating the seed for N+1-th epoch, we should still use data from
+    -- N-th epoch.
+    richmen <- getSscRichmenFromLrc "sscCalculateSeed" (epoch - 1)
+    sscRunGlobalQuery $ sscCalculateSeedQ epoch richmen
+
+sscCalculateSeedQ
+    :: EpochIndex
+    -> RichmenStakes
+    -> SscGlobalQuery (Either SscSeedError SharedSeed)
+sscCalculateSeedQ _epoch richmen =
```
gromakovsky: I would delete `_epoch` argument, because it was here only because of nist beacon.

txp/Pos/Txp/Toil/Failure.hs
```diff
@@ -34,8 +34,8 @@ data ToilVerFailure
     | ToilOverwhelmed !Int -- ^ Local transaction storage is full --
                             -- can't accept more txs. Current limit is attached.
     | ToilNotUnspent !TxIn -- ^ Tx input is not a known unspent input.
-    | ToilOutGTIn { tInputSum  :: !Integer
-                  , tOutputSum :: !Integer}
+    | ToilOutSscIn { tInputSum  :: !Integer
```
gromakovsky: wtf ;d
Toil Output Greater Than Input

wallet/src/Pos/Wallet/Web/Pending/Functions.hs
```diff
@@ -63,7 +63,7 @@ isReclaimableFailure = \case
     ToilSlotUnknown          -> True
     ToilOverwhelmed{}        -> True
     ToilNotUnspent{}         -> False
-    ToilOutGTIn{}            -> False
+    ToilOutSscIn{}           -> False
```
gromakovsky: palehchi pozh

wallet/node/Main.hs
```diff
@@ -106,14 +106,14 @@ walletProd WalletArgs {..} = first one $ worker walletServerOuts $ \sendActions
         walletAddress
         (Just walletTLSParams)
 
-pluginsGT ::
+pluginsSsc ::
```
gromakovsky: I'd remove `Ssc` suffix here.

txp/Pos/Txp/Toil/Failure.hs
neongreen: lol mda

PR #1848, opened 2017-11-01
[CSM-547] Document HD Wallets code in preparation for audit

pva701:


core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+    so child's public and secret keys can be _derived_ from parent
+    by this index using parent's public-secret pair.
+    On the other hand, a child can't know a parent in reasonable time
+    without some extra information.
+
+* Used structure of tree
+    In general case a structure of tree can be arbitrary but
+    we use specific structure.
+    We have 2 layers tree, nodes of the first layer are called "accounts"
+    and nodes of the second layer are called "addresses" and being leaves of a tree.
```
Martoon-00: I think `being` used incorrectly

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+    There are 2^32 accounts and each account has 2^32 addresses, so wallet has
```
Martoon-00: maybe better would be
`There could be possibly 2^32 different accounts ...`

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+    2^64 addresses in total.
+
+    The overall scheme is imagined on the picture below.
```
Martoon-00: scheme is pridumana on picture below?
I think `represented` fits here better

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+
+                                 (wallet's public key and secret key)
+                                  /             |                  \
+                (account[0] (pk, sk))    (account[1] (pk, sk)) ... (account[2^32-1] (pk, sk))
+                /                  \                                    \
+    (address[0][0] (pk, sk)) ... (address[0][2^32-1] (pk, sk)) ... (address[2^32-1][2^32-1] (pk, sk))
+
+    So each address belongs to one account.
+    Addresses are being addresses in usual sense:
```
Martoon-00: better without `being`

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+    they can contain money, money can be spent from them,
+    address can be used as change address and so on.
+
+    Accounts also reflects a intuition behind word "account":
```
Martoon-00: `reflect an` ... `behind the word`

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+    each account contains money, you can spent money from it,
```
Martoon-00: Is that try?

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+    you can have different accounts for different purposes.
+
+* Derivation process
+    The deriving public key can be performed using an only public key of a parent node.
```
Martoon-00: `deriving of`

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,127 @@
...
+    @deriveHDPublicKey@ does it.
+    More knowledge about underlying cryptography you can see here
```
Martoon-00: Znanie nilza uvidet', mozno tolko popytatsa propit

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    The overall scheme is represented on the picture below.
+
+                                 (wallet's public key and secret key)
+                                  /             |                  \
+                (account[0] (pk, sk))    (account[1] (pk, sk)) ... (account[2^32-1] (pk, sk))
+                /                  \                                    \
+    (address[0][0] (pk, sk)) ... (address[0][2^32-1] (pk, sk)) ... (address[2^32-1][2^32-1] (pk, sk))
+
+    So each address belongs to one account.
+    Addresses are addresses in common sense:
```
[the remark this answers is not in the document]
volhovM: Better throw away this sentence at all:
Address can contain money, be used as a change address, ...

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    they can contain money, money can be spent from them,
+    address can be used as change address and so on.
+
+    Accounts also reflect an intuition behind the word "account":
```
volhovM: Same, just enumerate features, throw away "reflect a intuition..." bit.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    balance of each account can be computed as sum of balances of addresses which belong to it,
+    you can have different accounts for different purposes and so on.
+
+* Derivation process
+    To derive secret key, parent secret key and passphrase are necessary.
+    @deriveHDSecretKey@ does it.
```
volhovM: Derivation is done using X.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    Description of underlying cryptography you can get here
```
volhovM: You can get more detailed cryptography specification here:...

volhovM: Also it is wrong because we don't use BIP-32, because it works on secp256k1 only, we use HD/Ed25519 derivation scheme. Here's a spec:
https://cardanolaunch.com/assets/Ed25519_BIP.pdf

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--private-child-key
+
+    The deriving of public key can be performed using just a public key of a parent node.
+    @deriveHDPublicKey@ does it, but we don't use it yet.
```
volhovM: Again, "it is done using 'x' function", remove "but we don't..." bit.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    Description of underlying cryptography you can get here
```
volhovM: Remove this link.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#public-parent-key--public-child-key
+
+    It's not allowed to derive a public key from parent public key for
+    indicies which are greater than or equal to 2^31.
+    For these indicies secret key is required.
```
volhovM: parent secret key*

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    Such way to derive public keys is called "hardened derivation":
```
volhovM: This feature is called "hardened derivation".

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    We also don't use it yet.
```
volhovM: 2bad

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    See more here
+    https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--public-child-key
```
volhovM: Remove the link.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+
+    Note: all these functions don't require number of derivated level,
```
volhovM: "derivation level number" maybe?

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    only credential information (public key or secret key with passphrase)
+    and index of a child.
+    Also we can derive public and secret keys for any arbitrary number of levels.
+
+* HD address payload
+    There were mentioned "account" and "address" entities,
```
volhovM: "Account" and "address" entities were mentioned,

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    but only addresses appear in the blockchain.
+    When we create a transaction, for each TxOut we specify address and coins.
```
volhovM: we specify X for each Y

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    So neither transaction nor blocks know anything about accounts and wallets.
```
volhovM: So neither transaction or block knows ...

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+
+    We want to be able to track our addresses in the blockchain
+    to compute balance of wallet and accounts,
```
volhovM: wallet and related accounts

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    but as said in _General scheme_ section we aren't able
```
volhovM: as it is said... we're not able

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    to determine a parent without any additional info.
```
volhovM: without _some_ particular info

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    So we can't determine account and wallet which address belongs to.
+
+    For each leaf let's store path from the root to this leaf along with an address.
+    So path is a list of indexes of derivation on each level.
```
volhovM: derivation indices*

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    To hide path from other parties we will encrypt it.
+    This encrypted derivation path is called @HDAddressPayload@
+    and stored in attributes of Address datatype.
+
+    Note: length of derivation path in our case is 2: @[account index, address index]@.
+
+* Encryption of payload
```
volhovM: Payload encryption.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    To encrypt derivation path let's use AEAD scheme using ChaCha20 and Poly1305.
+    To see more information go to https://tools.ietf.org/html/rfc7539.
```
volhovM: More information is available here: ...
Don't force users to go somewhere.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    Hash of root public key is used as a symmetric key for ChaChaPoly1305.
+    This hash called @HDPassphrase@ and function @deriveHDPassphrase@ generates
+    it by root public key.
+
+    Encryption of payload performs @packHDAddressAttr@ which takes @HDPassphrase@ and
+    derivation path and returns @HDAddressPayload@.
+    It serializes derivation path using @Bi@ instance for it and then encrypts produced bytes sequence.
+
+* Decryption of payload
```
volhovM: Payload decryption.

core/Pos/Crypto/HD.hs
```diff
@@ -1,4 +1,129 @@
...
+    To decrypt encrypted derivation path we have to derive @HDPassphrase@ again
+    from root public key and then try to decrypt @HDAddressPayload@.
+    If it's successfully decrypted then it implies address belongs to our tree
+    and, vice versa, address doesn't belong to our tree if decryption is failed.
+
+    Note: publishing of root public key gives other parties opportunity to reveal
+    all your addresses in the blockchain.
+    So it's not safe for the user to share his root public key (like he can do
+    it with a usual public key) because he loses his anonymity.
```
volhovM: "lose anonymity" doesn't seem to be a correct phrase. Use "it is unsafe to share root pk (like he can do) because it implies user deanonymization".

core/Pos/Crypto/HD.hs
```diff
@@ -29,61 +29,54 @@ This module provides basic operations under HD wallets.
...
-    We also don't use it yet.
-    See more here
-    https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#private-parent-key--public-child-key
-
-    Note: all these functions don't require number of derivated level,
+    indicies which are greater than or equal to 2^31:
+    for these indicies parent secret key is required.
+    This feature is called "hardened derivation":
+
+    Description of underlying cryptography you can get here
```
volhovM: "You can get X here: .." instead of "X you can get here".

docs/hd.md
```diff
@@ -0,0 +1,124 @@
+# Cardano SL HD Wallets
+
+## General scheme
+    
+HD wallet has tree-based structure.
```
adinapoli-iohk: "_An_ HD wallet has _a_ tree structure." sounds a bit better.

docs/hd.md
```diff
@@ -0,0 +1,124 @@
...
+The root of the tree is user's public-secret key pair.
```
adinapoli-iohk: I'm a bit on the fence whether or not we should say that each node contain such `(pk,sk)` tuple, because technically you can always derive a `pk` from a `sk`. Maybe we should clarify that:

"Although it's always possible to derive a `pk` from a `sk` due to the mathematical properties of ECC (_Elliptic Curves Cryptography_), for illustration purposes we represent each node as always be composed by a `(public key, secret key)` pair".

docs/hd.md
```diff
@@ -0,0 +1,124 @@
+# Cardano SL HD Wallets
```
adinapoli-iohk: We should include a section called `Glossary` where we clarify terms which might be ambiguous or non-obvious.

For example, I would start by writing something like:

```
## Glossary
Public Key: Explains what a Public Key is.
Secret Key: Say that it corresponds to (essentially) a Private key.
Extended Public Key:
Extended Private Key:
...
```

docs/hd.md
```diff
@@ -0,0 +1,124 @@
...
+
+## General scheme
+    
+HD wallet has tree-based structure.
+The root of the tree is user's public-secret key pair.
+Each node in the tree corresponds to public-secret key pair.
+
+Each child node is identified by an index of this child,
+so child's public and secret keys can be _derived_ from parent
+by this index using parent's public-secret key pair.
```
adinapoli-iohk: I think the main problem with this documentation is that it's not crystal clear about the relationship between a private key, a public key, an extended public key and an extended private key is.

To begin with, I think we should try to pick a naming scheme as close as possible as [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Specification_Key_derivation), so people which are familiar with "standard" HD wallets won't be too confused.

I'm even OK with the fact we pick different names, but we should be super precise about:

* When do we mean a `public key` intended as a point on an elliptic curve?
* When do we mean an `extended public key` intended as the tuple (public key, chain code) ?
* When do we mean a `private key` intended as random bits that, when multiplied by the fixed-point G on the elliptic curve, yields a `public key`?
* When do we mean an `extended private key` intended as the tuple (private key, chain code)?

Unfortunately the confusion stems from the fact that, in the code, we have a `newtype` wrapper around an `XPub` which is called `PublicKey` (if memory does not betray me).

pva701:
@adinapoli-iohk I can address your issues about public-secret keys, but tbh I don't get the necessity of it.

I would notice the following points:
  1. The HD derivation doesn't depend on the particular scheme used for public/private keys, so it doesn't 
  matter whether we use extended scheme or simple plain keys.
  I suppose this documentation aims to show overall HD scheme used by us.
  2. This documentation doesn't dive into cryptographic details. For instance, derivation processes aren't 
  described very precisely from viewpoint of cryptography.
There is the comprehensive description of our crypto here https://cardanolaunch.com/assets/Ed25519_BIP.pdf and I referred to it in the doc.

On the other hand, it's enough to a reader to realize general public-secret pair properties, for instance, a public key can be generated from a secret in a single way but not vice versa. And I assumed auditors have such knowledge.

So my main point is: there is not an appropriate place to describe our public-secret key scheme. It's better to do in a separate doc. But I agree it makes sense to add `Glossary` section with a brief description of public/private keys (not extended) to avoid misunderstandings.

adinapoli-iohk:
> So my main point is: there is not an appropriate place to describe our public-secret key scheme. It's better to do in a separate doc. But I agree it makes sense to add Glossary section with a brief description of public/private keys (not extended) to avoid misunderstandings.

I really cannot make a statement about this because my naive assumption would be that there is no better place than a document inside a GitHub project to be specific about our HD wallet structure and our key derivation mechanism (although we can, of course, treat the crypto primitives as a blackbox).

> On the other hand, it's enough to a reader to realize general public-secret pair properties, for instance, a public key can be generated from a secret in a single way but not vice versa. And I assumed auditors have such knowledge.

I see what you mean, and I guess it all boils down to which audience this document is being produced for and which is the purpose of this document. If it's a testament of how we implement, on the technical level, our HD wallet derivation scheme, I suspect we might want to be extremely precise. If this has to be intended as a "primer" I agree there is no need to be highly specific.

I'll check in with @dcoutts as I'm very curious to hear what he thinks.

adinapoli-iohk:
Thanks @dcoutts ! Filed here: https://issues.serokell.io/issue/CSM-560

PR #1855, opened 2017-11-03
[CSM-550] Group pagination parameters in one datatype

flyingleafe:
Investigation on whether or not pagination parameters can be grouped led to the following results:

1. Pagination parameters can indeed be grouped
2. Pagination parameters can have default values which are passed to API handlers automatically, which looks way more convenient than dealing with `Maybe`s in each handler.
3. As long as `extended` query parameter (which I dared to rename to `response_type`, because it seems more self-explanatory) and `Daedalus-Response-Format` header serve the same purpose, their values can be combined in single value, which is then passed to handlers.

Besides of that, I couldn't resist doing small refactorings here and there. 

These solutions are not final, but they clearly show that we can do interesting things with our API, keeping our handlers neat and tidy even if API specification isn't very convenient to implement.

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/Types.hs
+++ b/wallet-new/src/Cardano/Wallet/API/Types.hs
@@ -23,7 +23,6 @@ import qualified Servant.Server.Internal as SI
 
 import           Servant
 import           Servant.API.Sub         ((:>))
-import           Servant.Swagger
 import           Test.QuickCheck
 
 --
@@ -114,10 +113,6 @@ instance ( HasServer (apiType a :> res) ctx
         inRouteServer @(apiType a :> res) route $
         \f a -> f $ fromMaybe def a
 
-instance HasSwagger (apiType a :> res) =>
-    HasSwagger (WithDefaultApiArg apiType a :> res) where
-    toSwagger _ = toSwagger (Proxy @(apiType a :> res))
-
 -- | Type aliases for query params and headers which have default values.
 type DQueryParam s a = WithDefaultApiArg (QueryParam s) a
 type DHeader s a = WithDefaultApiArg (Header s) a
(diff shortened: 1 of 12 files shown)


wallet-new/src/Cardano/Wallet/API/Types.hs
```diff
@@ -11,13 +11,146 @@ import           Universum
...
+type family MergeArgs a where
+    MergeArgs (a -> a -> b) = a -> b
+
+-- | `AlternativeApiArg` is used when we want to provide a way
+-- to pass one argument via two ways, e. g. via request header or query parameter.
+data AlternativeApiArg (argA :: * -> *) (argB :: * -> *) a
+
+instance ( ApiHasArgClass argA a
+         , ApiHasArgClass argB a
+         , ApiArg argA a ~ Maybe b
```
Martoon-00: Extra constraint?

flyingleafe: Because we need to choose between two parameters, and we cannot meaningfully do it if both of them are always present (we can always choose the first one, but this doesn't make sense).
But thanks for the comment, actually - I realized that first constraint is unnecessary, and constraint in `HasServer` can be generalized to requiring any  `Alternative`.

wallet-new/src/Cardano/Wallet/API/Types.hs
```diff
@@ -1,23 +1,161 @@
...
+         , Default c
+         ) =>
+         HasServer (WithDefaultApiArg apiType a :> res) ctx where
+    type ServerT (WithDefaultApiArg apiType a :> res) m =
+        UnmaybeArg (ServerT (apiType a :> res) m)
+    route =
+        inRouteServer @(apiType a :> res) route $
+        \f a -> f $ fromMaybe def a
+
+instance HasSwagger (apiType a :> res) =>
```
adinapoli-iohk: Minor nitpick: If we move the `HasSwagger` definition inside `V1.Swagger` we don't need to depend on `servant-swagger`. It's something I'm trying to achieve as part of the design: there should be separation of concerns between the logic & the documentation.

What do you think, @flyingleafe ?

flyingleafe: It's sensible.

adinapoli-iohk:
Thanks for your contribution, @flyingleafe !

PR #1856, opened 2017-11-03
[CSL-1758] Track installed updates in MiscDB

int-index:


int-index:
I had to rebase it after #1840 — @gromakovsky, review again?

PR #1861, opened 2017-11-03
Merge cardano-sl-1.0

gromakovsky:


lib/configuration.yaml
```diff
@@ -14866,6 +14866,54 @@ mainnet_dryrun_wallet_macos64: &mainnet_dryrun_wallet_macos64
...
+      bvAlt: 0
+    systemTag: none
+
+mainnet_staging_short_epoch_wallet_win64:
+  <<: *mainnet_staging_short_epoch_full
+  update:
+    <<:
+    applicationName: csl-daedalus
+    applicationVersion: 0
+    lastKnownBlockVersion:
```
volhovM: I don't understand why `lastKnownBlockVersion` is overriding one in `mainnet_staging_short_epoch_full` with the same value, but it's not related to this merge pr.

PR #1863, opened 2017-11-04
[CSL-1775] Finish moving Txp into txp/

neongreen:


neongreen:
Appveyor is broken :(

> ```
> git checkout -qf FETCH_HEAD
> Restoring build cache
> Error downloading cache item: Remote server returned 500: 
>   The process cannot access the file 'D:\cache-data\build-cache-53120
>     \project-265590\da39a3ee5e6b4b0d3255bfef95601890afd80709
>     \82eb3ad6bfe5a56e0e930991a581f6e732a27fea.zip.001\data'
>   because it is being used by another process.
> ```

lib/cardano-sl.cabal
```diff
@@ -60,6 +59,8 @@ library
                         Pos.Behavior
 
                         -- Something
+                        Pos.Types
```
gromakovsky: Why did you move it here? 🤔 
They were alphabetically sorted (or very close to be sorted)

txp/Pos/Txp.hs
```diff
@@ -0,0 +1,7 @@
+-- | Txp system reexports.
+
+{-# OPTIONS_GHC -F -pgmF autoexporter #-}
+
+import           Pos.Arbitrary.Txp         ()
+import           Pos.Arbitrary.Txp.Network ()
```
gromakovsky: `Pos.Arbitrary.Txp.Network` imports `Pos.Arbitrary.Txp`, so the import of `Pos.Arbitrary.Txp` is likely redundant. On the other hand, it's kinda weird that this dependency is not the other way around 😕 
Anyway, it's not important. I'd like `Pos.Arbitrary.Txp` to import `Pos.Arbitrary.Txp.Network`, but not too much.

lib/cardano-sl.cabal
neongreen: Because `-- Txp` was a wrong place for it

txp/Pos/Txp.hs
neongreen: > I'd like Pos.Arbitrary.Txp to import Pos.Arbitrary.Txp.Network

me too, but I like the green checks so I might fix it in a different PR

PR #1865, opened 2017-11-05
Update OpenSSL

neongreen:
The previous minor version has become unavailable, causing Appveyor builds to fail.


neongreen:
(Should I change the base to cardano-sl-1.0 and then merge cardano-sl-1.0 into master so that we would have the fix in both branches?)

PR #1867, opened 2017-11-06
[CSM-552] purs to js migration

akegalj:
This PR implements https://issues.serokell.io/issue/CSM-552 - it rewrites `isValidRedemptionKey` and `isValidPaperVendRedemptionKey` into js.

`isValidPaperVendRedemptionKey` is tested with samples from https://cardanodocs.com/technical/wallet-frontend/#redeem-valid-paper-vend : 
```
$ node
> var bla = require('./src/redemption_key_validation.js')
undefined
> bla.isValidPaperVendRedemptionKey('lwIF94R9AYRwBy0BkVVpLhwtsG3CmqDvMahlQr3xKEY=')
false
> bla.isValidPaperVendRedemptionKey('CUUCxbH5pPc6RC1dvJKetJPVTG7X4WKPmBsoeZMbiQok')
true
```

`isValidPaperVendRedemptionKey` is tested:
```
> bla.isValidRedemptionKey('lwIF94R9AYRwBy0BkVVpLhwtsG3CmqDvMahlQr3xKEwm')
false
> bla.isValidRedemptionKey('lwIF94R9AYRwBy0BkVVpLhwtsG3CmqDvMahlQr3xKEw=')
true
> bla.isValidRedemptionKey('lwIF94R9AYRwBy0BkVVpLhwtsG3CmqDvMahlQr3xKE==')
true
> bla.isValidRedemptionKey('lwIF94R9AYRwBy0BkVVpLhwtsG3CmqDvMahlQr3xK===')
false
```

TODO:
 * [x] rewrite `isValidPaperVendRedemptionKey`
 * [x] rewrite `isValidRedemptionKey`
 * [ ] ~simplify `isValidRedemptionKey` with js lib that has base64 decode and base64url decode~
 * [x] write automated tests that check rewritten function against its alternative from purescript

EDIT: after careful consideration my conclusion is that current code is not that complex and it does things in a correct manner (in contrast to most of tried js libraries)

TESTED:
```
Successfull tests for redemption key: 5500
Failed tests for redemption key: 0
redemptionKeyValid: 400
redemptionKeyInvalid: 5100
Successfull tests for paper vend redemption key: 5500
Failed tests for paper vend redemption key: 0
paperVendRedemptionKeyValid: 122
paperVendRedemptionKeyInvalid: 5378
```

### DIFF
--- a/daedalus/src/redemption_key_validation.js
+++ b/daedalus/src/redemption_key_validation.js
@@ -70,48 +70,139 @@ function decodeUtf8(buf) {
     decoder.decode(buf);
 }
 
+// Implements: "Valid redemption key should be base64 and base64url decodable, it should end with '=' and it should be 44 chars long."
+function isValidRedemptionKey(code) {
+    var res = true;
+    try {
+        // Check if atob is defined (if we are in browser - it should be as we are in Electron)
+        // Note that with Electron we will probably run the first branch (so, atob is defined) but just in case we add fallback - as there are some mentiones that it was not working sometimes https://github.com/electron/electron/issues/8529
+        if(atobIsDefined) {
+            // If we are in browser then:
+            //  * convert input data from base64url into base64 - as vending was done in a messy way with both base64 and base64 url
+            //  * decode base64 with atob - note that atob usually does base64 validity check under the hood
+            //  * convert string into Uint8Array with unsafeStringToUint8ArrayOrCharCodes
+            //  * decode Uint8Array into a utf-8 string
+            decodeUtf8(unsafeStringToUint8ArrayOrCharCodes(atob(toRfc4648(code))));
+        } else {
+            // If we are not in browser then:
+            //  * convert input data from base64url into base64 - as vending was done in a messy way with both base64 and base64 url
+            //  * decode base64 using node facilities but additionally run manual base64 validity check (which Node doesn't implement)
+            //
+            //  This branch probably won't run at all in Electron.
+            decodeNode(toRfc4648(code));
+        }
+    } catch (err) {
+        res = false;
+    }
+    // Return true if:
+    //  * its base64 or base64url decodable - ada vending state mess
+    //  * ends with '=' - base64 padds string with optional '=' or two '=' signs. In ada vending phase we were using base64 configured where '=' was always added at the end.
+    //  * length of base64 string should be 44 chars long. Its 32 byte decoded data - but this check is more exact becase of optional '=' in base64. See example:
+    //      > base64url.toBuffer("qL8R4QIcQ_ZsRqOAbeRfcZhilN_MksRtDaEAAAArM-A=").length
+    //          32
+    //      > base64url.toBuffer("qL8R4QIcQ_ZsRqOAbeRfcZhilN_MksRtDaEAAAArM-A").length
+    //          32
+    //  Later example above is 32 byte long but its not valid ada redemption key because ada redemption key ends with '=' (lib used in above example is https://www.npmjs.com/package/base64url)
+    return res && code.endsWith('=') && code.length == 44;
+}
+
+// Implements: "Valid paper vend redemption key should be base58 decodable 32 byte stream."
+function isValidPaperVendRedemptionKey(code) {
+    try {
+        return bs58.decode(code).length == 32;
+    } catch (err) {
+        return false;
+    }
+}
+
+function tests() {
+    // Before running tests be sure to build client api. For more information see cardano docs
+    const api = require('../output/Daedalus.ClientApi')
+
+    // taken from https://stackoverflow.com/a/1349462/1924817
+    var rardomString = function(len, charSet) {
+        charSet = charSet || 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+        var randomString = '';
+        for (var i = 0; i < len; i++) {
+            var randomPoz = Math.floor(Math.random() * charSet.length);
+            randomString += charSet.substring(randomPoz,randomPoz+1);
+        }
+        return randomString;
+    }
+
+    // generate base64 samples
+    // Some will be valid, some won't
+    var b64Length = function(l, charSet) {
+        var b64 = rardomString(l, charSet) // valid base64 but might be invalid redemption key (depending on the size)
+        var b64Pad1 = b64 + "="   // should be valid depending on the size
+        var b64Pad2 = b64 + "=="  // should be valid depending on the size
+        var b64Pad3 = b64 + "===" // invalid
+        return [b64, b64Pad1, b64Pad2, b64Pad3]
+    }
+
+    var testList = []
+    var numberOfGroupTests = 100;
+    var b64Charset = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
+    var b64urlCharset = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_'
+    var b58Charset = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'
+    for (var i = 0; i < numberOfGroupTests; i++) {
+        // generate base64 test strings
+        Array.prototype.push.apply(testList, b64Length(40, b64Charset))
+        Array.prototype.push.apply(testList, b64Length(41, b64Charset))
+        Array.prototype.push.apply(testList, b64Length(42, b64Charset))
+        Array.prototype.push.apply(testList, b64Length(43, b64Charset))
+        Array.prototype.push.apply(testList, b64Length(44, b64Charset))
+        Array.prototype.push.apply(testList, b64Length(45, b64Charset))
+
+        // generate base64url test strings
+        Array.prototype.push.apply(testList, b64Length(40, b64urlCharset))
+        Array.prototype.push.apply(testList, b64Length(41, b64urlCharset))
+        Array.prototype.push.apply(testList, b64Length(42, b64urlCharset))
+        Array.prototype.push.apply(testList, b64Length(43, b64urlCharset))
+        Array.prototype.push.apply(testList, b64Length(44, b64urlCharset))
+        Array.prototype.push.apply(testList, b64Length(45, b64urlCharset))
+
+        // generate base58 test strings
+        Array.prototype.push.apply(testList, [rardomString(40, b58Charset)])
+        Array.prototype.push.apply(testList, [rardomString(41, b58Charset)])
+        Array.prototype.push.apply(testList, [rardomString(42, b58Charset)])
+        Array.prototype.push.apply(testList, [rardomString(43, b58Charset)])
+        Array.prototype.push.apply(testList, [rardomString(44, b58Charset)]) // most of the time this will be correct base58
+        Array.prototype.push.apply(testList, [rardomString(45, b58Charset)])
+        Array.prototype.push.apply(testList, [rardomString(46, b58Charset)])
+    }
+
+    var redemptionKeyTrue = 0 
+    var redemptionKeyFalse = 0 
+    var paperVendRedemptionKeyTrue = 0 
+    var paperVendRedemptionKeyFalse = 0 
+    for (var i = 0, len = testList.length; i < len ; i++) {
+        if(isValidRedemptionKey(testList[i]) == api.isValidRedemptionKey(testList[i])) {
+            isValidRedemptionKey(testList[i])? redemptionKeyTrue++: redemptionKeyFalse++;
+        } else {
+            console.log("Test failed for redemption key: " + testList[i])
+        }
+        if(isValidPaperVendRedemptionKey(testList[i]) == api.isValidPaperVendRedemptionKey(testList[i])) {
+            isValidPaperVendRedemptionKey(testList[i])? paperVendRedemptionKeyTrue++: paperVendRedemptionKeyFalse++;
+        } else {
+            console.log("Test failed for paper vend redemption key: " + testList[i])
+        }
+    }
+
+    console.log("Successfull tests for redemption key: " + (redemptionKeyTrue + redemptionKeyFalse))
+    console.log("Failed tests for redemption key: " + (testList.length - redemptionKeyTrue - redemptionKeyFalse))
+    console.log("redemptionKeyTrue: " + redemptionKeyTrue)
+    console.log("redemptionKeyFalse: " + redemptionKeyFalse)
+
+    console.log("Successfull tests for paper vend redemption key: " + (paperVendRedemptionKeyTrue + paperVendRedemptionKeyFalse))
+    console.log("Failed tests for paper vend redemption key: " + (testList.length - paperVendRedemptionKeyTrue - paperVendRedemptionKeyFalse))
+    console.log("redemptionKeyTrue: " + paperVendRedemptionKeyTrue)
+    console.log("redemptionKeyFalse: " + paperVendRedemptionKeyFalse)
+
+}
+
 module.exports = {
-    // Implements: "Valid redemption key should be base64 and base64url decodable, it should end with '=' and it should be 44 chars long."
-    isValidRedemptionKey: (code) => {
-        var res = true;
-        try {
-            // Check if atob is defined (if we are in browser - it should be as we are in Electron)
-            // Note that with Electron we will probably run the first branch (so, atob is defined) but just in case we add fallback - as there are some mentiones that it was not working sometimes https://github.com/electron/electron/issues/8529
-            if(atobIsDefined) {
-                // If we are in browser then:
-                //  * convert input data from base64url into base64 - as vending was done in a messy way with both base64 and base64 url
-                //  * decode base64 with atob - note that atob usually does base64 validity check under the hood
-                //  * convert string into Uint8Array with unsafeStringToUint8ArrayOrCharCodes
-                //  * decode Uint8Array into a utf-8 string
-                decodeUtf8(unsafeStringToUint8ArrayOrCharCodes(atob(toRfc4648(code))));
-            } else {
-                // If we are not in browser then:
-                //  * convert input data from base64url into base64 - as vending was done in a messy way with both base64 and base64 url
-                //  * decode base64 using node facilities but additionally run manual base64 validity check (which Node doesn't implement)
-                //
-                //  This branch probably won't run at all in Electron.
-                decodeNode(toRfc4648(code));
-            }
-        } catch (err) {
-            res = false;
-        }
-        // Return true if:
-        //  * its base64 or base64url decodable - ada vending state mess
-        //  * ends with '=' - base64 padds string with optional '=' or two '=' signs. In ada vending phase we were using base64 configured where '=' was always added at the end.
-        //  * length of base64 string should be 44 chars long. Its 32 byte decoded data - but this check is more exact becase of optional '=' in base64. See example:
-        //      > base64url.toBuffer("qL8R4QIcQ_ZsRqOAbeRfcZhilN_MksRtDaEAAAArM-A=").length
-        //          32
-        //      > base64url.toBuffer("qL8R4QIcQ_ZsRqOAbeRfcZhilN_MksRtDaEAAAArM-A").length
-        //          32
-        //  Later example above is 32 byte long but its not valid ada redemption key because ada redemption key ends with '=' (lib used in above example is https://www.npmjs.com/package/base64url)
-        return res && code.endsWith('=') && code.length == 44;
-    },
-    // Implements: "Valid paper vend redemption key should be base58 decodable 32 byte stream."
-    isValidPaperVendRedemptionKey: (code) => {
-        try {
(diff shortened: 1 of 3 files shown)


adinapoli-iohk:
Looking good so far, @akegalj ! As you have tested your code already, do you think we can write tests to do this programmatically?

akegalj:
> As you have tested your code already, do you think we can write tests to do this programmatically?
In general, yes. I would need to generate arbitrary inputs and feed middleware vs rewritten functions for equality. I can write such tests tomorrow

akegalj:
@adinapoli-iohk note my direct translation from PS into JS for `isValidRedemptionKey` is a bit more complex. It can be simplified as: "Valid redemption key should be base64 and base64url decodable, it should end with '=' and it should be 44 chars long."

Direct translation looks more messy then it should be as I was trying to be exact and use same dependencies as before.

base64 && base64url decodable probably exists in js world and we might use different dependency to achieve it (and thus have simpler validation function in js). I will check this out tomorrow.

daedalus/src/redemption_key_validation.js
```diff
@@ -0,0 +1,74 @@
...
+    decoder.decode(str);
+}
+
+module.exports = {
+    isValidRedemptionKey: (code) => {
+        var res = false;
+        try {
+            if(atobIsDefined) {
+                decodeUtf8(unsafeStringToUint8ArrayOrCharCodes(atob(toRfc4648(code))));
+                res = true;
```
adinapoli-iohk: If we change the default value of `res` to `true` on line 53, then you can remove the redundant `res = true` on line 57 and line 60. Generally speaking the pessimistic approach of starting with `false` is a good one, but here is not necessary as we are wrapping all the work in a `try/catch` block anyway. What do you think?

adinapoli-iohk:
> base64 && base64url decodable probably exists in js world and we might use different dependency to achieve it (and thus have simpler validation function in js). There are more elegant solutions then this in js. I will check this out tomorrow.

Was going to suggest this anyway, as we should try to piggyback on well-tested solutions as much as possible rather than trying to write this code ourselves as it can have lots of subtleties. Looks like you have a plan so I will leave you to it and won't review up until you consider this ready to go 😉  Good progress!

daedalus/src/redemption_key_validation.js
akegalj: `true` :)

adinapoli-iohk: Excellent pun! 😀

daedalus/src/redemption_key_validation.js
```diff
@@ -0,0 +1,117 @@
...
+                //  * convert string into Uint8Array with unsafeStringToUint8ArrayOrCharCodes
+                //  * decode Uint8Array into a utf-8 string
+                decodeUtf8(unsafeStringToUint8ArrayOrCharCodes(atob(toRfc4648(code))));
+            } else {
+                // If we are not in browser then:
+                //  * convert input data from base64url into base64 - as vending was done in a messy way with both base64 and base64 url
+                //  * decode base64 using node facilities but additionally run manual base64 validity check (which Node doesn't implement)
+                //
+                //  This branch probably won't run at all in Electron.
+                decodeNode(toRfc4648(code));
```
akegalj: This probably won't be run in Electron as its a browser box with `atob` defined.

daedalus/src/redemption_key_validation.js
```diff
@@ -0,0 +1,117 @@
...
+//  Where node's implementation doesn't:
+//  > Buffer.from("====", "base64").toString("utf-8")
+//  ''
+//  > Buffer.from("aa==", "base64").toString("utf-8")
+//  'i'
+//  > Buffer.from("aa=", "base64").toString("utf-8")
+//  'i'
+//
+// And that's the reason this funciton exists: to extend Node's base64 decode with validity check.
+function decodeNode (str) {
```
akegalj: This probably won't be run in Electron as its a browser box with `atob` defined.

akegalj:
WIP put back until automated tests are written

akegalj:
Running tests shows some failed tests:

```
$ node
> var bla = require('./src/redemption_key_validation.js')
undefined
> bla.tests()
Test failed for redemption key: f7Xc-bvEC3nOWXUOW2HmgvFZ9jthgvpL-u476uec6mF=
Test failed for redemption key: HoTQtPT-8qJC4zc1GHK3sQusl63fhjbhOOIt-AkIn7==
Test failed for redemption key: 9Oof_62TkBhp3dwwaFg4_UtShgufWfovCMtuxyyHjj==
Test failed for redemption key: 6oujcOZus5t-Vo-u4T2T6lxezj5DwJ4A-0KyI2Ma05I=
Test failed for redemption key: WXJyCbMAD_VUN4Y-12YHxhMjz0N1jSQz2-TnrSUdFc==
Test failed for redemption key: V-sRQoWZMpKRZ8A1IdYTUigcP-SD4INd7OfhAypmMX==
Test failed for redemption key: PaVynKdirVOEzzZOxwZzsUbk4wEf83Smw_2ko-1grX-=
Test failed for redemption key: cOdVjWtVVAW4165-TyKvAR-7G1JG7Np6sm30NHAq8i==
Test failed for redemption key: BJClXs10FMNesHm5oDekw-KbQ-njl4Ikzqs71TBJ65J=
Test failed for redemption key: qE-aInqrwdC8zKTo4z5-7-U2nhrW9dvrBX7RI-5Bm3n=
Test failed for redemption key: VoQzRTlPkeBEKOiXnXTv_LPQkQ6QNZ_PyIwU-dWTIF==
Test failed for redemption key: _3eux05024Z_GbpCGUo6SVDFw0LZV8QUWdTiLZGiboT=
Test failed for redemption key: leund6Ol6MgEpz_pc6yNZUGeUiitvR_x9mAn9f9oiVG=
Test failed for redemption key: _6dvC8c_uhIAPDU5SD4SVyrT0A0bYvcsgYVfjUh0tu==
Test failed for redemption key: 15esg9boO-dL1O180HsX-y7jFkslmZncQ4qmdDOr41==
Test failed for redemption key: lUGJicaZPq36jSe6PJz_gxJu-HSc_sac51jJSB14TGG=
Test failed for redemption key: zdF72W5IgX3ULlitXJYeNUHY3_K82DNMrO9_pomxx6==
Test failed for redemption key: dHtxhOzqGJiRhYbFu0qvb2yn_eZQnDyqvy_sTMPFZHT=
Test failed for redemption key: z1_6yF-cgHiorX_gz_J0qEYtRDtYdFbF7F7flWvaRVn=
Test failed for redemption key: pZjIHXlVK30Ojl-9iUEl4Mc-9EEJQOzNp0ypKXAwu3==
Test failed for redemption key: YFq-0Kz1ik73pP4XY-M3QE9BQExENm3f5AbsMBoxVZS=
Test failed for redemption key: UAPkvmmrdCusLtIQRYh63rtrl--Ng_zcH5Uk0bQENH==
Test failed for redemption key: PuW9f0mfV48A_Og3gl01XlMb5_yObTeLESbcdbXOo_==
Test failed for redemption key: _xAYXE0kIKqhY6OcWSfoihXDC_VFY8eNvd4sZgoUvoK=
Test failed for redemption key: r9ekF-kcJ-yYyxvt4mmx0EyM0Wm3MqZfDbn_uyWYXa==
Test failed for redemption key: OFBWveQGhpD13HNGM_31B_YKK-6c8JVEuIZyV511tK==
Test failed for redemption key: 0_gAuZh_3z_kho95FXK6DktGC8rOa-Qgr5QC91wi4Y==
Test failed for redemption key: KVwsEp-OfRHGAOgJIprV_RoNdrGyM-Bd6KDTGldO8W8=
Test failed for redemption key: l8YOmWDLxjzdKfpWCyBQt_-HAl-TCNHxM0vCm64_gl==
Test failed for redemption key: jdshVodJKxKCw-__UWHhlWzY3oxkSJhECRAEBCgHfh5=
Test failed for redemption key: CA9S-sExfAD7lX6Y5CK3-62ljo2TimZxBikQxlOxJcq=
Test failed for redemption key: rAZ_iGcLZnBu9-ZMmgp5vwJWNHxpleNctsIx0zBw-J==
Test failed for redemption key: s09CBKNt_CMCPzIOymPMQ9mR2SZvkme_9yCn4JyHvk==
Test failed for redemption key: KshDnYIy6J2k6Kf9Ploq-U0pnx-nIJvmekJO6-_HlYQ=
Test failed for redemption key: dv-hN_7bVjNF4g0_bK0uYm7H0eXNjaESriMsEXCqbUO=
Test failed for redemption key: tnQVEL2GmXPyMhwcUfT_asJXBMnMJ_4t3NqygNOA88Y=
Test failed for redemption key: wZDRp7mYMVOAh-RqlWcccDuwjwo6rcyAw4w__8IXvF==
Test failed for redemption key: 1WzRwk9197oJjLJdQeFInJNlZ3-5PXa2C987cFe-zV==
Test failed for redemption key: sJfmGO-uW-nlMdT2y5Fofsz8qOEUB52PaOfPFR5CUp==
Test failed for redemption key: 8wQtG_9CBMeBcpjOo_jJzcSMvSq4WQ_9ylhZw4nd6o==
Test failed for redemption key: TsmI_tnVnUaST6_ZzS5Miyyvcy4InRTyBih0KsoJdm==
Test failed for redemption key: zb4KYLrr-Wa8-IR5l4c-Q2tublevAJQxF9ij9375Ei==
Test failed for redemption key: nehkWxBMqVBAj_1SiK4GOqiChUeFkoSZZobxBXFfaE_=
Test failed for redemption key: 5QUm6LZyAV_R-cOFvN1J6-jcMhjvGFzK9thzDmlG5i==
Test failed for redemption key: BWX7_x5sYnnai5CZL_oXcxKNXCttDVMw51rL92_-8db=
Test failed for redemption key: G2UF2vk91gERgwH9-BOgm59jSIyRU8lcHC8DthYd5i-=
Test failed for redemption key: Co5H5IdePY-w3sUW1bhLfiDxmy9UkS-qK6tdItgOsnm=
Test failed for redemption key: Xg_fJzyul_AdNKCCPT9c-V8gyyuX2bjG2LWtHAK8xF==
Test failed for redemption key: iL_FDy9fC7GCzeSAnUTS7STGamKWGmV7_yoT_YpYGm==
Test failed for redemption key: GKKTC_diA-OB4g2gVI6OfSmkh-lpaEBgJuzCRjyAwX==
Test failed for redemption key: _2tmnZw0NaEiH7MKXgZ9QmrQVz8axj_uY9h8C-bej3p=
Test failed for redemption key: ltYVh9_YOQxlSBlTP1XphB3xSYwEV2IlG5rfY__O0n==
Test failed for redemption key: yM_iIpaZMPCs_5g2MsKz2fUKN5qXjtvImhkjV420GL==
Test failed for redemption key: 1ZOKX_O7ztvPyPYkmGx1lp4qi33mvj91NN_mIvKQql==
Test failed for redemption key: 2-epEG6lJ1g9YLkoox__IaS-bRpGih7Ji_fQd-zeAdD=
Test failed for redemption key: P9dRa-dEXDgnLKY93mv-BXEsVceRJw499Zb-vecSrj==
Test failed for redemption key: v_-Ih9ndH3YPElGeQLO1Rz7Asxmeq_N35nSvDze48tB=
Test failed for redemption key: WRbTb2CMwubxrqQ-iHo_bxm3aMxHFc6n-5DDMPgIyb==
Test failed for redemption key: D7wU9lb-tdgGGrTiO4RArmUxmq-WkJqdTLolToVkTYO=
Successfull tests for redemption key: 5441
Failed tests for redemption key: 59
redemptionKeyTrue: 341
redemptionKeyFalse: 5100
Successfull tests for paper vend redemption key: 5500
Failed tests for paper vend redemption key: 0
redemptionKeyTrue: 129
redemptionKeyFalse: 5371
```

Will investigate.

akegalj:
@adinapoli-iohk 
After latest commit tests pass:
```
$ node
> var bla = require('./src/redemption_key_validation.js')
undefined
> bla.tests()
Successfull tests for redemption key: 5500
Failed tests for redemption key: 0
redemptionKeyValid: 400
redemptionKeyInvalid: 5100
Successfull tests for paper vend redemption key: 5500
Failed tests for paper vend redemption key: 0
paperVendRedemptionKeyValid: 122
paperVendRedemptionKeyInvalid: 5378
```

The problem was because one liner:
```
return str.replace('_', '/').replace('-', '+');
```

replaces only first occurrences of '_' and '-' and correct way to replace all occurrences should be done with `RegExp` as its in PS implementation - but I missed it as I was careless:
```
return str.replace(new RegExp('_', 'g'), '/').replace(new RegExp('-', 'g'), '+');
```

Example:
```
> '---'.replace('-', '+')
'+--'
> '---'.replace(new RegExp('-', 'g'), '+')
'+++'
```

More about rfc4648: https://en.wikipedia.org/wiki/Base64#RFC_4648

PS implementation that was translated is here https://github.com/menelaos/purescript-b64/blob/master/src/Data/String/Base64/Internal.purs#L49 : 
```
toRfc4648 :: String -> String
toRfc4648 = replaceAll "-" "+" <<< replaceAll "_" "/"
```

And the missing bit is from `replaceAll` https://github.com/menelaos/purescript-stringutils/blob/v0.0.6/src/Data/String/Utils.purs#L264 : 
```
replaceAll :: String -> String -> String -> String
replaceAll old new str = replace (mkRegex old) new str
  where
    -- Helper function to construct a `Regex` from an input string
    mkRegex :: String -> Regex
mkRegex str = unsafePartial (fromRight (regex (escapeRegex str) global))
```

where I have skipped `mkRegex` part as I thought its not essential.

daedalus/src/redemption_key_validation.js
```diff
@@ -0,0 +1,208 @@
...
+    var leadingQuanta = "^([A-Za-z0-9+/]{4})*";
+    var finalQuantum =
+        "([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}(?:=)?|[A-Za-z0-9+/]{2}(?:=){0,2})$";
+    var reValidBase64 =
+        new RegExp([reEmptyString, "|", leadingQuanta, finalQuantum].join(""));
+
+    if (!reValidBase64.test(str)) { throw new Error("Invalid input string");}
+    result = Buffer.from(str, "base64").toString("utf-8");
+
+    return result;
```
adinapoli-iohk: My only minor nitpick here is the use of regexes which looks like an hairy approach (although your excellent code coverage is making up for that). Let me see if I can propose something slightly different: if it's not feasible, we simply go back to your regex-based approach and I will be happy with that :)

OK, so here is my thought process: You say we cannot use the `Buffer` API because it is too lenient with malformed input (which is, it won't throw an error, but produce a "chopped" string), so here is an idea about an invariant: "An input is a valid base64-encoded string if and only if, after converting it to base64 and then to `ascii`, it preserves its original length".

To put it differently, we could write something similar (pardon my broken Node.js / JS skills):

```javascript
function decodeNode(original) {
  /* Note how we do not need to convert to UTF-8. RFC 4648 describes the Base64 Alphabet as:
  "A 65-character subset of US-ASCII is used, enabling 6 bits to be
   represented per printable character.  (The extra 65th character, "=",
   is used to signify a special processing function.)"
   */
  var converted    = Buffer.from(original, "base64").toString("ascii");

  // Note how `Buffer.from` performs a copy, so we can still access the old string size.
  vas isValidBase64 = converted.size() === original.size();
  if (!isValidBase64) throw new Error("Invalid input string");
  
  return converted;
}
```

Do you think I am being naive or this could actually work?

daedalus/src/redemption_key_validation.js
```diff
@@ -0,0 +1,208 @@
...
+    try {
+        // Check if atob is defined (if we are in browser - it should be as we are in Electron)
+        // Note that with Electron we will probably run the first branch (so, atob is defined) but just in case we add fallback - as there are some mentiones that it was not working sometimes https://github.com/electron/electron/issues/8529
+        if(atobIsDefined) {
+            // If we are in browser then:
+            //  * convert input data from base64url into base64 - as vending was done in a messy way with both base64 and base64 url
+            //  * decode base64 with atob - note that atob usually does base64 validity check under the hood
+            //  * convert string into Uint8Array with unsafeStringToUint8ArrayOrCharCodes
+            //  * decode Uint8Array into a utf-8 string
+            decodeUtf8(unsafeStringToUint8ArrayOrCharCodes(atob(toRfc4648(code))));
```
adinapoli-iohk: In general, do we need conversion to `utf-8`? Why `ascii` is not enough?

daedalus/src/redemption_key_validation.js
akegalj: > we simply go back to your regex-based approach

Note that its not *my* regex based approach - but its a copy paste from https://github.com/menelaos/purescript-b64/blob/master/src/Data/String/Base64.js#L11 . My modification was just to remove `Either` constructors.

I was trying to have exact implementation as before - not to introduce any regression.

My line of thought is - should we  modify implementation (as with your proposal we are doing that)?

> converting it to base64 and then to ascii, it preserves its original length

this sounds plausible, and I will try it out. Note what I have commented above though - this might and probably will change the implementation (we might just not care about that)

daedalus/src/redemption_key_validation.js
akegalj: In our case it is, true. I was just doing the exact translation in the first pass.

akegalj:
> Sorry for being overly fussy

please, don't apologize for this. I like being exact

daedalus/src/redemption_key_validation.js
adinapoli-iohk: Ah, interesting! (And to be 100% clear I'm not pointing fingers here, just trying to reason together if there is a better way to handle this 😉 ).

You make a valid point, this might change the implementation but hopefully not the end result (and I hope your tests will catch any regression). Let's see how it goes, and worst come to worst we will make a note in the code saying this is a direct porting on the PS code so the implementation is not any more fragile than the original 😉

daedalus/src/redemption_key_validation.js
```diff
@@ -0,0 +1,53 @@
...
+
+const bs58 = require("bs58");
+
+// Convert base64url into base64
+function toRfc4648(str) {
+    return str.replace(new RegExp('_', 'g'), '/').replace(new RegExp('-', 'g'), '+');
+}
+
+// Checks is input string valid base64 or base64url
+function isValidBase64Url(code) {
```
akegalj: Validity function has been factored out into a separate function

daedalus/src/redemption_key_validation.js
```diff
@@ -5,95 +5,27 @@
...
-            decodeUtf8(unsafeStringToUint8ArrayOrCharCodes(atob(toRfc4648(code))));
-        } else {
-            // If we are not in browser then:
-            //  * convert input data from base64url into base64 - as vending was done in a messy way with both base64 and base64 url
-            //  * decode base64 using node facilities but additionally run manual base64 validity check (which Node doesn't implement)
-            //
-            //  This branch probably won't run at all in Electron.
-            decodeNode(toRfc4648(code));
-        }
+            atob(toRfc4648(code));
```
akegalj: validity has been changed from 
```
decodeUtf8(unsafeStringToUint8ArrayOrCharCodes(atob(toRfc4648(code))));
```
into `atob(toRfc4648(code));` as we don't care about the end result - only about validity.

daedalus/src/redemption_key_validation.js
```diff
@@ -115,94 +47,7 @@ function isValidPaperVendRedemptionKey(code) {
     }
 }
 
-function tests() {
```
akegalj: Tests have been removed as they can't be run within the Node now (as `atob` is not defined in the Node). 

I have checked with Nikola Glumac and he confirmed to me that `isValidRedemptionKey` and `isValidPaperVendRedemptionKey` will be used only in Electron render process where we have `atob` defined.

Previously we have shown that `isValidPaperVendRedemptionKey` works as expected.

We can show that `isValidRedemptionKey` works too by analysing previous implementation that was run https://github.com/menelaos/purescript-b64/blob/master/src/Data/String/Base64.purs#L120 : 
```
decode :: String -> Either Error String
decode str =
  if atobIsDefined
    then
      unsafeStringToUint8ArrayOfCharCodes <$> atob (toRfc4648 str)
        >>= decodeUtf8
    else
runFn3 _decodeNode Left Right (toRfc4648 str)
```

where in Electron render process the first branch of `if` statement was run:
```
unsafeStringToUint8ArrayOfCharCodes <$> atob (toRfc4648 str)
        >>= decodeUtf8
```

and if we remove `decodeUtf8` and `unsafeStringToUint8ArrayOfCharCodes` as we don't need them in our use case - we are left with only `atob (toRfc4648 str)` which translates to our js implementation. (we have previously verified and fixed `toRfc4648` and `atob` is the same as it has been FFI from js.

adinapoli-iohk:
@akegalj Looks like you will need to regenerate `pkgs/default.nix`:

https://gist.github.com/601326c637f8311e200e04a1014d4642

Once CI is green, I'm happy to merge!

adinapoli-iohk:
Thanks @akegalj ! I ended up squashing this one, hoping this won't rub @domenkozar the wrong way with regards the new `git flow` we are trying to adopt 😅 

(ps. I'm very happy to take recommendation on how I should have behaved in this situation, but I thought that considering we had 13 comments for what it seemed a minor feature, squashing was cleaner).

PR #1868, opened 2017-11-06
[WIP][CSM-544] Hook V0 API inside `wallet-new`

adinapoli-iohk:
This PR paves the way to start integrating the old version of the API into `wallet-new`.

I have rewrote parts of what was `wallet/node/NodeOptions` into `wallet-new` counterparts, seizing the chance to break down things differently in order to simultaneously understand a bit better what we are doing here. **To be 100% clear, this is not touching the old wallet code, this is 100% wallet-new bound and that's why we are merging the PR into `alfredo/wallet-new`**.

Further PRs will expand the work done on this one to start implementing V1 functionalities in terms of the V0 building blocks.


adinapoli-iohk:
I had to temporarily disable TLS for now as for some reason even if passing the correct path at `--tlskey=scripts/tls-files/server.key` would result in:

```
HandshakeFailed (Error_Misc "no keys found\nCallStack (from HasCallStack):\n  error, called at ./Network/Wai/Handler/WarpTLS.hs:247:20 in warp-tls-3.2.4-IKRexZDxUaN2V73k7nnd5h:Network.Wai.Handler.WarpTLS")
```

For now the solution was to disable tls when running with `--wallet-debug`. I will create a ticket to restore the functionality so we don't forget.

Merging by the virtue of the fact this is not impacting anything but the `alfredo/wallet-new` branch.

PR #1870, opened 2017-11-06
[CSM-555] Switch from () to NoContent for endpoints with no result

adinapoli-iohk:
This slightly-invasive PR switch from `()` to `NoContent` for endpoints which do not return a result. This was necessary to avoid `servant-swagger` generating invalid specs.

Prior to this PR, it was not possible to validate the spec generated:

screen shot 2017-11-06 at 09 45 17

After this PR, we can:

screen shot 2017-11-06 at 16 25 27

`NoContent` is a type defined in `Servant.API.ContentTypes` which is then erased by response bodies but at the same time ensures our Swagger spec validates correctly.

If you guys can think of a less-invasive way to support this, I'm all ears (one possibility would be to `fmap NoContent` each unit-returning endpoint inside `servantHandlers`, but if feels conceptually cleaner to give each handler a proper return type directly in its definition).

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Methods/History.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/History.hs
@@ -41,6 +41,7 @@ import           Pos.Wallet.Web.State         (AddressLookupMode (Ever), MonadWa
 import           Pos.Wallet.Web.Util          (decodeCTypeOrFail, getAccountAddrsOrThrow,
                                                getWalletAccountIds, getWalletAddrs,
                                                getWalletAddrsSet)
+import           Servant.API.ContentTypes     (NoContent (..))
 
 
 type MonadWalletHistory ctx m =
@@ -67,8 +68,8 @@ getFullWalletHistory cWalId = do
 
     let localHistory = unfilteredLocalHistory `Map.difference` blockHistory
 
-    logTxHistory "Block" blockHistory
-    logTxHistory "Mempool" localHistory
+    _ <- logTxHistory "Block" blockHistory
+    _ <- logTxHistory "Mempool" localHistory
 
     fullHistory <- addRecentPtxHistory cWalId $ localHistory `Map.union` blockHistory
     walAddrs    <- getWalletAddrsSet Ever cWalId
@@ -152,8 +153,10 @@ addHistoryTxMeta
     :: MonadWalletDB ctx m
     => CId Wal
     -> TxHistoryEntry
-    -> m ()
-addHistoryTxMeta cWalId = addHistoryTxsMeta cWalId . txHistoryListToMap . one
+    -> m NoContent
+addHistoryTxMeta cWalId txhe = do
+    _ <- addHistoryTxsMeta cWalId . txHistoryListToMap . one $ txhe
+    return NoContent
 
 -- This functions is helper to do @addHistoryTx@ for
 -- all txs from mempool as one Acidic transaction.
@@ -161,10 +164,11 @@ addHistoryTxsMeta
     :: MonadWalletDB ctx m
     => CId Wal
     -> Map TxId TxHistoryEntry
-    -> m ()
+    -> m NoContent
 addHistoryTxsMeta cWalId historyEntries = do
     metas <- mapM toMeta historyEntries
     addOnlyNewTxMetas cWalId metas
+    return NoContent
   where
     toMeta THEntry {..} = CTxMeta <$> case _thTimestamp of
         Nothing -> liftIO getPOSIXTime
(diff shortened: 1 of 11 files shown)


wallet/src/Pos/Wallet/Web/Methods/Logic.hs
```diff
@@ -247,17 +249,18 @@ markWalletReady cid isReady = do
...
     removeWallet wid
     removeTxMetas wid
     removeHistoryCache wid
     deleteSecretKey . fromIntegral =<< getAddrIdx wid
+    return NoContent
 
-deleteAccount :: MonadWalletLogic ctx m => AccountId -> m ()
-deleteAccount = removeAccount
+deleteAccount :: MonadWalletLogic ctx m => AccountId -> m NoContent
+deleteAccount aid = removeAccount aid >> return NoContent
```
Martoon-00: Since you go with one-liner, why not `removeAccount aid $> NoContent`

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -208,8 +213,7 @@ addRecentPtxHistory wid currentHistory = do
...
     :: (Container t, Element t ~ TxHistoryEntry, WithLogger m, MonadIO m)
-    => Text -> t -> m ()
-logTxHistory desc =
-    logInfoS .
-    sformat (stext%" transactions history: "%listJson) desc .
-    map _thTxId . toList
+    => Text -> t -> m NoContent
+logTxHistory desc t = do
+    logInfoS . sformat (stext%" transactions history: "%listJson) desc . map _thTxId . toList $ t
+    return NoContent
```
Martoon-00: Seems strange that logging function returns some `NoContent`

Martoon-00:
JG! 
I also don't see a nice way to do it shorter and without modifying all the handlers around.

Although, maybe overlapping `instance ToSchema ()` would help to perform the same job? Actually we already have similar hack for documentation, so it will look alright if it works.

wallet/src/Pos/Wallet/Web/Methods/Logic.hs
adinapoli-iohk: No particular reason and I don't have a strong opinion on whether or not we should use one or the other, so I'm happy to go with the applicative style here.

wallet/src/Pos/Wallet/Web/Methods/History.hs
adinapoli-iohk: That's possibly an oversight during my refactoring, let me see if that was actually required and if not I agree, that's a mistake, it should really be `m ()` 😉

adinapoli-iohk:
> Although, maybe overlapping instance ToSchema () would help to perform the same job?

That doesn't seem to be a great way to go. To begin with, I don't like overlapping instances and secondly `NoContent` looks like what we want here as it's more explicit in its intention, so I might be tempted to stick to this approach 😉

Martoon-00:
Another possible way would be to modify only module which declares endpoints
https://github.com/input-output-hk/cardano-sl/blob/master/wallet/src/Pos/Wallet/Web/Server/Handlers.hs#L29
modifying required handlers as `handler` -> `fmap (\() -> NoContents) ... handler`, it looks less invasive.
What do you think?

adinapoli-iohk:
> another possible way would be to modify only module which declares endpoints

Yes, that was the other solution I was advocating in the PR description. It's surely doable, but I have refrained from it because I have eventually realised this design it's clearer in its intentions: after all we are writing http handlers, and as such it's totally expected for a http handler to return either some form of payload, or none (i.e. `NoContent`). In a nutshell, such return type for those functions seems to be more _domain specific_.

let's see what @flyingleafe thinks, I'm happy either way if you have a strong preference on one or the other.

adinapoli-iohk:
> which are reused in other parts of code, and which now require these _ <- prefixes there;

Yes, I totally see your point @flyingleafe , and I'm not happy with the current design either. Technically you could avoid the wild card doing something (apparently) even worse (which is why I haven't pursue that route):

```
NoContent <- gonnaExplode
```

Technically that's fine here as the `NoContent` type is inhabited by only 1 type constructor, but in other situations I would steer away from this approach as you might easily end up with a runtime exception in case of a failed pattern match.

Whilst we could remove the wildcard here, this got me thinking that probably the cringing stems from the fact we are reusing _whole_ http handlers in other parts of the logic. In theory, one might advocate that handlers needs to map an HTTP request to an HTTP response. All the rest should be abstracted away into domain specific items. In pseudocode:

```
myHandler input = do
  finalResult <- myDomainSpecificLogic
  transform finalResult
```

That way `myDomainSpecificLogic` can return any type (including `()`) and we can reuse it all across the board. We'll get there, I'm sure 😀

PR #1871, opened 2017-11-06
[DEVOPS-493] csl-daedalus `applicationVersion`:  bump to 5

deepfire:


gromakovsky:
Why 5? Do I recall correctly that mainnet is running 2 and staging is running 3?

deepfire:
@gromakovsky, `rc-staging` runs `4`: https://github.com/input-output-hk/internal-documentation/wiki/Daedalus-installer-history

node/configuration.yaml
```diff
@@ -14817,7 +14817,7 @@ mainnet_wallet_macos64: &mainnet_wallet_macos64
   update:
     <<:
     applicationName: csl-daedalus
-    applicationVersion: 3
+    applicationVersion: 5
```
gromakovsky: Why did you change it here and on line 14808? It should be changed only for staging (i. e. `mainnet_dryrun…` configurations), no?

deepfire: @gromakovsky, good catch!  Thank you!

deepfire:
@domenkozar, this would dismiss review. It might be wise not to delay this any further.

domenkozar:
If all PRs can have issue attached, this one should as well.

PR #1858, opened 2017-11-03
[CSM-525] More public logs in wallet

Martoon-00:
Log errors and other interesting stuff in wallet.
This PR contains bunch of small improvements in logging, please see commit history for details

**This PR enables very important logging, diagnosis of wallet-related issues is hindered without it**
**This PR blocks #1874**

PR #1874, opened 2017-11-07
[CSM-551] Split request and response in servant logs

Martoon-00:
Like in daedalus logs, requests are now logged immediatelly, and responses are logged later once endpoint completed. It allows to make sure whether endpoint was or wasn't called, even if it may, potentially, take enormous amount of time in complete.

Each request-response pair also have its unique id, so that we can find one by another in logs easily.

default

**NOTE:**
**Do not merge until #1858 gets merged**. Once it merged, conflict will arise and I'll rebase this PR upon new `cardano-sl-1.0`, this should resolve CI failure as well

### DIFF
--- a/node/src/Pos/Util/Servant.hs
+++ b/node/src/Pos/Util/Servant.hs
@@ -48,10 +48,12 @@ import           Control.Monad.Catch     (handleAll)
 import           Control.Monad.Except    (ExceptT (..), MonadError (..))
 import           Data.Default            (Default (..))
 import           Data.Reflection         (Reifies (..), reflect)
+import qualified Data.Text               as T
 import qualified Data.Text.Buildable
 import           Data.Time.Clock.POSIX   (getPOSIXTime)
 import           Formatting              (bprint, build, builder, formatToString, sformat,
                                           shown, stext, string, (%))
+import           GHC.IO.Unsafe           (unsafePerformIO)
 import           Serokell.Util           (listJsonIndent)
 import           Serokell.Util.ANSI      (Color (..))
 import           Servant.API             ((:<|>) (..), (:>), Capture, QueryParam,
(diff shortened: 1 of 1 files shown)


node/src/Pos/Util/Servant.hs
```diff
@@ -399,6 +401,22 @@ instance ( HasServer (apiType a :> LoggingApiRec config res) ctx
...
+
+-- | We really want all servant servers to have non-overlapping ids,
+-- so singleton counter here.
+requestsCounter :: TVar Integer
+requestsCounter = unsafePerformIO $ newTVarIO 0
+{-# NOINLINE requestsCounter #-}
+
+allocRequestId :: MonadIO m => m RequestId
+allocRequestId =
+    atomically $ RequestId <$> modifyTVarS requestsCounter (identity <<+= 1)
```
adinapoli-iohk: I guess this is just an observation (read bikeshedding) but the "unrolled" version is yes one line
longer but is more straightforward to understand for the uninitiated as it uses just the standard `stm` primitives:

```haskell
allocRequestId = atomically $ do
  modifyTVar' requestsCounter (+ 1)
  RequestId <$> readTVar
```

Always on the bikeshedding front, I would call this function simply `nextRequestId`.

Martoon-00:
Thanks for your review!
Changed that `allocRequestId`, now it should look better.

adinapoli-iohk:
Thanks @Martoon-00 ! I guess we just need one of the gate keepers to go and merge it.

adinapoli-iohk:
@Martoon-00 would you please apply this patch to `pkgs/default.nix` so we can get the CI green?

https://gist.github.com/5c8b21f8154abfae4176435bda063208

Martoon-00:
Actually, I would like to merge this after #1858, because the former is more important and complex, and merging one of them will produce conflicts in another. And I hope that on next cardano-sl-1.0 merge here stack2nix issue gets resolved by itself

Martoon-00:
Closed in favour of ^

PR #1878, opened 2017-11-07
[CSM-559] Uniform naming for response types

adinapoli-iohk:
🚴 🚴   **Beware, bikeshedding in its purest form lies ahead!** 🚴 🚴 

This PR performs a bunch of renames and remove code which was originally committed as part of CSM-550 but wasn't really necessary (after all less code is better, right? 😉 ).

In particular, as part of this PR I propose we switch from `response_type` to simply `response_format`. The reason is that at the moment we have an HTTP header called `Daedalus-Response-Format` which has the same semantic, but we are naming the two differently (type vs format), which is confusing.

To bikeshed even further, the Oxford dictionary refers to a `type` as:

> A category of people or things having common characteristics.

Whereas `format` is defined as

> The way in which something is arranged or set out.

I would vote to go with `format`, as it describe better our intent (we are changing the shape of the response by passing a special parameter).

Let me know what you guys think!


wallet-new/spec/swagger.json
```diff
@@ -1036,10 +1036,6 @@
                         "description": "Invalid `new` or `old`"
                     },
                     "200": {
-                        "schema": {
-                            "items": [],
-                            "type": "array"
-                        },
```
Martoon-00: Btw am I right that swagger doc will be wiped out from git eventually?

adinapoli-iohk: No, they will stay 😉  The main point of having them in git is that we can always have a persistent remote link where we can point people at.

for example, we can already say:

"Hey, go to `editor.swagger.io` , click `File -> Import URL` and then paste the following:

https://raw.githubusercontent.com/input-output-hk/cardano-sl/5aeca683cd0bf32cdd9d92aaad6427acca42aaa0/wallet-new/spec/swagger.json

"

What would be quite a boon as they would be able to fetch the exact snapshot of the API at each stage of the development, without relying on things like https://cardanodocs.com/technical/wallet/api/ which can very easily get obsolete/outdated. "

adinapoli-iohk:
@Martoon-00 @flyingleafe Ok, I'm merging this one in the virtue of the fact I'm starting some integration work on the `/api/v1/wallets` API and this is useful to have to pick up the pace (that for a prototype should be quick fast).

I'm happy to reconsider when we are all online (either via Slack or on the issue tracker).

PR #1879, opened 2017-11-07
[CSM-523] Get rid of nonatomic deletions in key storage

Martoon-00:
I actually wanted to modify our wallet creating endpoints so that, instead of
`makeWallet emptyPassphrase >> changePassphrase emptyPassphrase userPassphrase`
they do the same in one hope (for security purposes).
But then I saw that only `importWallet` endpoint suffers from that, so just an other little change is here


Martoon-00:
I indeed didn't try to beat the problem if nonatomic operations in keystorage in general, just a small change which moves us towards that point )

Martoon-00:
I indeed didn't try to beat the problem if nonatomic operations in keystorage in general, just a small change which moves us towards that point )

PR #1882, opened 2017-11-08
[CSL-1798] Move definition of Block into cardano-sl-core

pva701:


core/Pos/Core/Txp.hs
```diff
@@ -42,26 +43,27 @@ module Pos.Txp.Core.Types
        -- * Block message
```
gromakovsky: `-- * Block message`? 🤔 

I bet some changes will be requested eventually, so consider removing this line.

explorer/src/Pos/Binary/Explorer.hs
```diff
@@ -6,9 +6,8 @@ import           Universum
 
 import           Pos.Binary              ()
 import           Pos.Binary.Class        (Cons (..), Field (..), deriveSimpleBi)
-import           Pos.Core                (HeaderHash, Timestamp)
+import           Pos.Core                (HeaderHash, Timestamp, TxUndo)
```
gromakovsky: I suppose `TxUndo` shouldn't be in core, but it's rather small, so I won't be too sad seeing it there.

core/Pos/Core/Delegation.hs
```diff
@@ -0,0 +1,56 @@
+-- | Delegation payload types.
+
+module Pos.Core.Delegation
```
gromakovsky: I believe this module should also contains some stuff from `Pos.Core.Types`, namely:
```
       , ProxySigLight
       , ProxySKLight
       , ProxySigHeavy
       , ProxySKHeavy
```

core/Pos/Core/Ssc.hs
```diff
@@ -112,7 +110,7 @@ mkCommitmentsMapUnsafe = CommitmentsMap
 
 -- | Opening reveals secret.
 newtype Opening = Opening
-    { getOpening :: (AsBinary Secret)
+    { getOpening :: AsBinary Secret
```
gromakovsky: palekhchi

lib/src/Pos/Binary/Slog.hs
```diff
@@ -1,6 +1,6 @@
 -- | Binary serialization of Slog types.
 
-module Pos.Binary.Block.Slog
+module Pos.Binary.Slog
```
gromakovsky: Why?

gromakovsky: Well, maybe we should separate `Block` and `Slog`, but currently `Slog` is part of `Block`, so I don't see a reason why this module isn't called `Pos.Binary.Block.Slog` 🤷‍♂️

lib/src/Pos/Block/Base.hs
```diff
@@ -0,0 +1,169 @@
...
+
+----------------------------------------------------------------------------
+-- Main smart constructors
+----------------------------------------------------------------------------
+
+type SanityConstraintM
+     = ( HasDifficulty BlockHeader
+       , HasHeaderHash BlockHeader
+       , HasConfiguration
+       )
```
gromakovsky: I think `HasDifficulty BlockHeader` and `HasHeaderHash BlockHeader` are in scope or can be added into scope by importing something from `Core`. So I guess `SanityConstraintM` can be deleted and `HasConfiguration` should be enough.

lib/src/Pos/Block/Base.hs
```diff
@@ -0,0 +1,169 @@
...
+
+----------------------------------------------------------------------------
+-- Genesis smart constructors
+----------------------------------------------------------------------------
+
+type SanityConstraintG
+     = ( HasDifficulty BlockHeader
+       , HasHeaderHash BlockHeader
+       , HasConfiguration
+       )
```
gromakovsky: This one is the same as `SanityConstraintM`, I don't think it's a good idea to separate them.

ssc/Pos/Ssc/Class/LocalData.hs
```diff
@@ -0,0 +1,46 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE RankNTypes          #-}
+
+-- | This module defines type class for local data storage.
+
+module Pos.Ssc.Class.LocalData
```
neongreen: Bad merge.

core/Pos/Core/Update.hs
```diff
@@ -413,59 +408,3 @@ mkUpdateProof = hash
 -- TODO: I don't like that 'Some' is used here
 -- —@neongreen
 type UpdateBlock = Either (Some IsGenesisHeader) (Some IsMainHeader, UpdatePayload)
```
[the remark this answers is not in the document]
pva701: how?

ssc/Pos/Ssc/Base.hs
```diff
@@ -2,7 +2,7 @@
 
 -- | Core functions from SSC.
 
-module Pos.Ssc.Core.Functions
+module Pos.Ssc.Base
```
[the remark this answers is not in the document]
pva701: `Slavka`

lib/src/Pos/Util/JsonLog.hs
```diff
@@ -36,17 +36,18 @@ import           Mockable                      (Catch, Mockable, realTime)
...
 import           Pos.Communication.Relay.Logic (InvReqDataFlowLog)
 import           Pos.Core                      (HasConfiguration, SlotId (..), gbHeader,
                                                 gbhPrevBlock, getSlotIndex, headerHash,
                                                 mkLocalSlotIndex)
+import           Pos.Core.Block                (Block, mainBlockTxPayload)
+import           Pos.Core.Block.Genesis.Lens   (genBlockEpoch)
+import           Pos.Core.Block.Main.Lens      (mainBlockSlot)
+import           Pos.Core.Txp                  (txpTxs)
 import           Pos.Crypto                    (hash, hashHexF)
-import           Pos.Txp                       (JLTxR (..), MemPoolModifyReason, txpTxs)
```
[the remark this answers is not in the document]
pva701: `import           Pos.Core.Txp                  (txpTxs)`

ssc/Pos/Ssc/Base.hs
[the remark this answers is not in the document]
pva701: Your feeling is right

core/Pos/Core/Delegation.hs
pva701: Somnevaus'

```
    Module imports form a cycle:
             module ‘Pos.Crypto.Signing.Tag’ (./Pos/Crypto/Signing/Tag.hs)
            imports ‘Pos.Core.Configuration.Protocol’ (./Pos/Core/Configuration/Protocol.hs)
      which imports ‘Pos.Core.Genesis.Types’ (./Pos/Core/Genesis/Types.hs)
      which imports ‘Pos.Core.Delegation’ (./Pos/Core/Delegation.hs)
      which imports ‘Pos.Core.Configuration’ (./Pos/Core/Configuration.hs)
      which imports ‘Pos.Core.Genesis.Canonical’ (./Pos/Core/Genesis/Canonical.hs)
      which imports ‘Pos.Core.Address’ (./Pos/Core/Address.hs)
      which imports ‘Pos.Crypto.Signing’ (./Pos/Crypto/Signing.hs)
      which imports ‘Pos.Crypto.Signing.Tag’ (./Pos/Crypto/Signing/Tag.hs)
```

core/Pos/Core/Update.hs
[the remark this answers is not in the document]
neongreen: By not using them

All functions that were taking a XxxBlock will now take a Block

core/Pos/Core/Delegation.hs
gromakovsky: But it seems you don't need full `HasConfiguration` here, I believe you need only a small part (so you can import only `Pos.Core.Configuration.SmallPart`) which might resolve the problem.

core/Pos/Core/Update.hs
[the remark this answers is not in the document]
gromakovsky: But then update system functions will have access to TxpPayload, for example, and we won't be sure that it doesn't use TxpPayload.

gromakovsky:
I will approve after @neongreen approves.

PR #1885, opened 2017-11-08
[CSM-557] Fix `instance FromHttpApiData (CId w)`

Martoon-00:
We don't want to throw decoding error, just return `False` if address is invalid


adinapoli-iohk:
@Martoon-00 Mind addressing the CI issue? It looks like a transient Appveyor problem, but it seems I don't have enough permissions to restart the job 😞

Martoon-00:
That job is not mandatory, we can merge now

adinapoli-iohk:
Nikola reports this is now fixed, good job @Martoon-00 !

PR #1890, opened 2017-11-09
[CSE-246] Un-/Subscription to last page of epochs

sectore:
_Task_: https://issues.serokell.io/issue/CSE-246

_How to test:_
1. [Build and run](https://github.com/input-output-hk/cardano-sl/tree/master/explorer/frontend#short-version-of-installation) `Explorer` locally
2. Run `nodes` with `tmux`: `scripts/launch/explorer-with-nodes.sh`
3. Open http://localhost:3100/ in your browser and click to any epoch at `latest blocks` table to open an epoch page. Check adding of new blocks there.


ksaric:
Nice work, @sectore :sunglasses: 
Will take a detailed look tomorrow, but looks very nice.

ksaric:
I thought about it. We don't need to remember the `EpochIndex` for the client since the changes on an `Epoch` are always occurring for the last one - https://github.com/input-output-hk/cardano-sl/pull/1890/files#diff-47bd247d999ad481c628a180a89567aaR47
We simply check if the last `Epoch` has some new `Slot` and if it has, we notify.
That should simplify the model a bit.

sectore:
@ksaric The client does not have any information about the last `epoch`. Imagine the current last `Epoch` is `10`. When an user is entering a slot page (`/epoch/3`) the client does not know if epoch `3` is the latest `epoch` or not. So the client subscribes to `SocketBlocksPageUpdated` once an user does enter a `/epoch/x` page. In this case the server has to decide if the client will be notified by `SocketBlocksPageUpdated` or not. That's why we do need an `EpochIndex` to know, what the "current" `epoch` of `/epoch/x` page at client-side is to compare it with latest `epoch` on backend-side.

ksaric:
@sectore That can be solved without tracking what `Epoch` the client want's to listen - it does not require state on the server side. You could simply check if the `Epoch` the client want's to listen is the latest and if yes, add the subscription. If not, don't add the subscription. 
I mean, adding `EpochIndex` to `ClientContext` may not seem like a little bit data, but if we have a lot of users it won't be a little bit data. And it would be simpler without tracking the state.
Right?

sectore:
@ksaric In this case the client have no information that its subscription has not been accepted. Currently  the client stores all its subscriptions at an own state (https://github.com/input-output-hk/cardano-sl/blob/master/explorer/frontend/src/Explorer/Types/State.purs#L59), so that they can be unsubscribed if needed. For example if an user leaves `epoch/x` page the client will unsubscribe `SocketBlocksPageUpdated`. This will fire an unsubscription into an empty space, because there was no subscription before. But the backend will try to handle it. 

The other thing is, when a client is subscribing to latest `epoch` (e.g. epoch `10`), then it will be notified every time if we have new blocks in `epoch 10`. But if we don't store `EpochIndex` all notifications will be sent again and again, even if the latest `epoch` is `11` and not `10` anymore. That means the client will receive new blocks from `epoch 11`, but it was interested on `epoch 10` only. How would you solve this without storing  an `EpochIndex`?

ksaric:
> This will fire an unsubscription into an empty space, because there was no subscription before.

Well, it's not like this isn't happening right now. I'm not saying it's often, but it does occur.

> The other thing is, when a client is subscribing to latest epoch (e.g. epoch 10), then it will be notified every time if we have new blocks in epoch 10. But if we don't store EpochIndex all notifications will be sent again and again, even if the latest epoch is 11 and not 10 anymore. That means the client will receive new blocks from epoch 11, but it was interested on epoch 10 only. How would you solve this without storing an EpochIndex?

The client has subscribed to epoch 10. The epoch switches to 11. 
1. The transitions from one epoch to another are quite a rare event in production and wouldn't be a problem if the client got the subscription from epoch 11 and simply ignore it.
2. If 1. is problematic, we could simply notice that the epoch that we received is not the one we are after (we want epoch 10, we receive epoch 11) and send an unsubscribe (as in - "oh, I'm not interested in the epoch 11 subscriptions, I just wanted epoch 10 subscriptions".

Seems like I'm making a fuss over some simple state, but the issues we had with socket.io before was exactly because of some simple state and the explosion of possible transitions between them.
If it's not a performance problem, and I don't think it is, let's keep it simple (KISS principle).

sectore:
> Well, it's not like this isn't happening right now. I'm not saying it's often, but it does occur.

If you've seen something like this, please open an YT issue for it.

explorer/src/Pos/Explorer/Socket/App.hs
```diff
@@ -174,9 +181,13 @@ periodicPollChanges connVar closed =
                             Just blocks -> return blocks
             let newBlunds = fromMaybe [] mNewBlunds
 
-            -- notify about blocks and blocks with offset
+            -- notify changes depending on new blocks
             unless (null newBlunds) $ do
+                -- 1. last page of blocks
                 notifyBlocksLastPageSubscribers
+                -- 2. last page of epochs
+                mSlotId :: Maybe SlotId <- lift $ getCurrentSlot @ctx
```
ksaric: Remove type signature ` :: Maybe SlotId`? It would also remove the extra import for the `SlotId`.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -640,6 +641,15 @@ epochSlotSearch epochIndex slotIndex = do
         errMsg :: Text
         errMsg = sformat ("No blocks on epoch "%build%" page "%build%" found!") epoch page
 
+-- | Get the @Page@s number from an @Epoch@ or throw an exception.
+getEpochPagesOrThrow
```
ksaric: If you want to move the function outside the function scope, move it into the `Helpers` section.

explorer/src/Pos/Explorer/Socket/Methods.hs
```diff
@@ -302,6 +333,18 @@ notifyTxsSubscribers
...
     view csTxsSubscribers >>= broadcast @ctx TxsUpdated cTxEntries
 
+notifyEpochsLastPageSubscribers
+    :: forall ctx m . ExplorerMode ctx m
+    => EpochIndex -> ExplorerSockets m ()
+notifyEpochsLastPageSubscribers currentEpoch = do
+    recipients <- view $ csEpochsLastPageSubscribers
+    -- ^ subscriber
+    maxPage <- getEpochPagesOrThrow currentEpoch
+    -- ^ max no. of epoch pages
```
ksaric: `-- ^ max no. of epoch pages` to `-- ^ last epoch page` and change variable name to `lastPage`.
Otherwise, we can just call `epochPageSearch` with `Nothing` and get the first page.

PR #1892, opened 2017-11-10
More precise test output for binary tests

NorfairKing:
The output now looks like so:

```
MainExtraBodyData
    binary encode and decode are inverses
    cbor flat term valid
    cbor canonical representation
```

It makes figuring out where the problem is a bit easier, and it also makes it easier to extend `binaryTest` with more, general tests.

### DIFF
--- a/lib/test/Test/Pos/Helpers.hs
+++ b/lib/test/Test/Pos/Helpers.hs
@@ -104,6 +104,15 @@ type IdTestingRequiredClassesAlmost a = (Eq a, Show a, Arbitrary a, Typeable a)
 
 type IdTestingRequiredClasses f a = (Eq a, Show a, Arbitrary a, Typeable a, f a)
 
+identityTestSpec
+  :: forall a. (IdTestingRequiredClassesAlmost a)
+  => (forall b. IdTestingRequiredClassesAlmost b => Spec)
+  -> Spec
+identityTestSpec s = describe (typeName @a) (s @a)
+  where
+    typeName :: forall x. Typeable x => String
+    typeName = show $ typeRep (Proxy @a)
+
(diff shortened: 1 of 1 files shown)


adinapoli-iohk:
Hey @NorfairKing , thanks for your contribution!

Unfortunately that part of the code underwent a bit of refactoring so there are conflicts now. Would you mind addressing those? Can I also ask Mr. @gromakovsky to chime in and report whether the approach taken by this PR is a good starting point?

Cheers folks!

lib/test/Test/Pos/Helpers.hs
```diff
@@ -104,6 +104,15 @@ type IdTestingRequiredClassesAlmost a = (Eq a, Show a, Arbitrary a, Typeable a)
 
 type IdTestingRequiredClasses f a = (Eq a, Show a, Arbitrary a, Typeable a, f a)
 
+identityTestSpec
+  :: forall a. (IdTestingRequiredClassesAlmost a)
+  => (forall b. IdTestingRequiredClassesAlmost b => Spec)
+  -> Spec
+identityTestSpec s = describe (typeName @a) (s @a)
+  where
+    typeName :: forall x. Typeable x => String
```
gromakovsky: Please make it top-level definition, because now it's copy-pasted from `identityTest`. I know it's small, but still.
Also, I think the definition is wrong, it should be `typeName = show $ typeRep (Proxy @x)` rather than `typeName = show $ typeRep (Proxy @a)` (it works because it's called as `typeName @a`, but it's still wrong conceptually). Can you fix it?

gromakovsky:
This PR became stale. I am closing it, but feel free to reopen after fixing one thing I mentioned above and resolving conflicts.

PR #1895, opened 2017-11-10
[CSM-563] Bump servant to 0.12

Martoon-00:
Feel free to point out extra code and places which lack of explanation

### DIFF
--- a/lib/src/Pos/Util/Servant.hs
+++ b/lib/src/Pos/Util/Servant.hs
@@ -47,6 +47,8 @@ import           Universum
 import           Control.Lens (Iso, iso, makePrisms)
 import           Control.Monad.Catch (handleAll)
 import           Control.Monad.Except (ExceptT (..), MonadError (..))
+import           Data.Constraint ((\\))
+import           Data.Constraint.Forall (Forall, inst)
 import           Data.Default (Default (..))
 import           Data.Reflection (Reifies (..), reflect)
 import qualified Data.Text.Buildable
@@ -102,12 +104,15 @@ class ApiHasArgClass apiType a where
         => Proxy (someApiType n a) -> String
     apiArgName _ = formatToString ("'"%string%"' field") $ symbolVal (Proxy @n)
 
-type ApiHasArgInvariant apiType a res =
-    Server (apiType a :> res) ~ (ApiArg apiType a -> Server res)
+class ServerT (apiType a :> res) m ~ (ApiArg apiType a -> ServerT res m)
+   => ApiHasArgInvariant apiType a res m
+instance ServerT (apiType a :> res) m ~ (ApiArg apiType a -> ServerT res m)
+      => ApiHasArgInvariant apiType a res m
 
 type ApiHasArg apiType a res =
     ( ApiHasArgClass apiType a
-    , ApiHasArgInvariant apiType a res
+    , ApiHasArgInvariant apiType a res Handler  -- this one for common cases
+    , Forall (ApiHasArgInvariant apiType a res)  -- and this is generalized one
     )
 
 instance KnownSymbol s => ApiHasArgClass (Capture s) a
@@ -133,14 +138,19 @@ class ModifiesApiRes t where
 -- 'ModifiesApiRes'.
 data VerbMod (modType :: *) api
 
-instance (HasServer (Verb mt st ct $ ApiModifiedRes mod a) ctx,
-          ModifiesApiRes mod) =>
+instance ( HasServer (Verb mt st ct $ ApiModifiedRes mod a) ctx
+         , HasServer (Verb mt st ct a) ctx
+         , ModifiesApiRes mod
+         ) =>
          HasServer (VerbMod mod $ Verb (mt :: k1) (st :: Nat) (ct :: [*]) a) ctx where
     type ServerT (VerbMod mod $ Verb mt st ct a) m =
          ServerT (Verb mt st ct a) m
 
     route = inRouteServer @(Verb mt st ct $ ApiModifiedRes mod a) route $
             serverHandlerL' %~ modifyApiResult (Proxy @mod)
+
+    hoistServerWithContext _ pc nt s =
+        hoistServerWithContext (Proxy :: Proxy (Verb mt st ct api)) pc nt s
 
 instance HasSwagger v => HasSwagger (VerbMod mod v) where
     toSwagger _ = toSwagger $ Proxy @v
@@ -191,6 +201,7 @@ instance ApiHasArgClass apiType a =>
     apiArgName _ = apiArgName (Proxy @(apiType a))
 
 instance ( HasServer (apiType a :> res) ctx
+         , HasServer res ctx
          , ApiHasArg apiType a res
          , FromCType (ApiArg apiType a)
          , ReportDecodeError res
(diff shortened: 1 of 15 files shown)


lib/src/Pos/Util/Servant.hs
```diff
@@ -133,15 +139,22 @@ class ModifiesApiRes t where
...
 data VerbMod (modType :: *) api
 
-instance (HasServer (Verb mt st ct $ ApiModifiedRes mod a) ctx,
-          ModifiesApiRes mod) =>
+instance ( HasServer (Verb mt st ct $ ApiModifiedRes mod a) ctx
+         , ReflectMethod mt
+         , KnownNat st
+         , AllCTRender ct a
+         , ModifiesApiRes mod
+         ) =>
```
adinapoli-iohk: This is a bit intimidating. Any chance we could document what's going on here and why we need those extra constraints ?

lib/src/Pos/Util/Servant.hs
```diff
@@ -229,16 +247,39 @@ instance (ApiHasArgClass apiType a, ApiArg apiType a ~ Maybe b) =>
...
+    -- @forall m. ServerT (WithDefaultApiArg apiType :> res) m ~ (e -> d)@
+    -- (for some @e@ and @d@).
+    -- However GHC doesn't allow @forall@ in constraints.
+    --
+    -- But we can use 'Forall' thing instead.
+    -- @Forall (ApiHasArgInvariant (WithDefaultApiArg apiType) a res)@
+    -- constraint is already provided here (see 'ApiHasArg' definition),
+    -- so we can instantiate it's quantification parameter @m@ to concrete
+    -- @m1@ and @m2@ required by this function, getting desired constraints.
+    hoistServerWithContext _ pc (nt :: forall x. m1 x -> m2 x) s =
```
adinapoli-iohk: Excellent explanation, although I have to admit it's quite mind-blowing and it makes my head hurt 😉

lib/src/Pos/Web/Server.hs
```diff
@@ -106,23 +106,28 @@ convertHandler nc nodeDBs txpData handler =
...
     catchServant = throwError
 
-nat
-    :: forall ext ctx m .
-    ( MyWorkMode ctx m, MempoolExt m ~ ext)
-    => m (WebMode ext :~> Handler)
-nat = do
+withNat
+    :: forall ext ctx api m .
+    (MyWorkMode ctx m, MempoolExt m ~ ext, HasServer api '[])
```
adinapoli-iohk: Oh! Why do we need that `HasServer api '[]` constraint, with the `'[]`?

stack.yaml
```diff
@@ -113,9 +113,12 @@ extra-deps:
 - concurrent-extra-0.7.0.10       # not yet on Stackage
 # - purescript-bridge-0.8.0.1
 - directory-1.3.1.0               # https://github.com/malcolmwallace/cpphs/issues/8
-- servant-0.10                    # servant-multipart supports version servant-10 only
-- servant-server-0.10             # so it triggers another dependencies to be v10
-- servant-multipart-0.10
+- servant-0.12
+- servant-server-0.12             # so it triggers another dependencies to be v10
```
adinapoli-iohk: `# so it triggers another dependencies to be v10` is now redundant and can be removed 😉

adinapoli-iohk:
@Martoon-00 I'm keen on getting this on `master` as soon as we can so that we can benefit from the upgraded stack in the `wallet-new` branch as well. I would suggest to address the comments nitpicks, fix CI and go directly for the merge 😉

lib/src/Pos/Util/Servant.hs
flyingleafe: @adinapoli-iohk 
This is because `hoistServerWithContext` is defined as the same as `hoistServerWithContext` for `Verb mt st ct a`, and we have `HasServer` constraint only for `Verb mt st ct (ApiModifierRes mod a)`. These constraints imply `HasServer (Verb mt st ct a)`.

Though I still have a couple of questions:
- Isn't it simpler and more clear to directly require `HasServer (Verb mt st ct a)` constraint (as you did everywhere else)?
- Is there actually any solid reason to not use `hoistServerWithContext` implementation for `Verb mt st ct (ApiModifier res a)`?

cc @Martoon-00

lib/src/Pos/Web/Server.hs
flyingleafe: @adinapoli-iohk `'[]`' denotes empty context, and we don't use any context for our handlers indeed -- in this function we use simple `hoistServer`, not `hoistServerWithContext`.

wallet/src/Pos/Wallet/Web/Api.hs
```diff
@@ -111,6 +111,9 @@ instance ReportDecodeError (WalletVerb (Verb (mt :: k1) (st :: Nat) (ct :: [*])
     reportDecodeError _ err = throwM (DecodeError err) ^. from serverHandlerL'
 
 instance ( HasServer (Verb mt st ct $ ApiModifiedRes WalletVerbTag a) ctx
+         , AllCTRender ct a
+         , KnownNat st
+         , ReflectMethod mt
          , Reifies config ApiLoggingConfig
          , ReflectMethod mt
```
flyingleafe: Duplicated constraint.
Btw, why not just require `HasServer (WalletVerb (Verb mt st ct)) ctx`? AFAIU, these constraints (excluding `Reifies` and `Buildable` ones) are here only to imply this constraint. Am I missing something?

lib/src/Pos/Util/Servant.hs
Martoon-00: > Isn't it simpler and more clear to directly require HasServer (Verb mt st ct a) constraint (as you did everywhere else)?

You're right, I'm switching to it..

> Is there actually any solid reason to not use hoistServerWithContext implementation for Verb mt st ct (ApiModifier res a)?
If I delegate to it, then I also need a function of signature `m a -> m (ApiModifiedRes mod a)`, and I have only roughly `Handler a -> Handler (ApiModifiedRes mod a)` which is important for errors catching.

lib/src/Pos/Web/Server.hs
adinapoli-iohk: Ah, perfect -- Incidentally I have seen this cropping up also in the `servant-quickcheck` package, which made me realise that's safe to use (was originally concerned we were restricting our types too much by doing that 😉  )

wallet/src/Pos/Wallet/Web/Api.hs
Martoon-00: Actually, `HasServer (WalletVerb (Verb mt st ct)) ctx` doesn't use logging. Those constraints would be implied by `HasServer (LoggingApi ...) ctx` or `HasLoggingServer ...`.
For some reason they are not enough as well though.

As justification I can generalize this instance and move to `Util.Servant`

Martoon-00:
Thanks for all your valuable comments!
I hope that once CI is pleased, we can merge

adinapoli-iohk:
@Martoon-00 ouch, more trouble with the CI! Would you please apply this patch?

https://gist.github.com/96ca8a7d01ebb8c35a40d141ad8e6de4

Almost there!

adinapoli-iohk:
🎉

PR #1896, opened 2017-11-11
[CSL-1854] Get rid of MonadBlock abstraction

pva701:


db/Pos/DB/GState/Common.hs
```diff
@@ -69,29 +75,25 @@ writeBatchGState = dbWriteBatch' GStateDB
...
-getTipBlockGeneric
-    :: forall block header undo m.
-       MonadBlockDBGeneric header block undo m
-    => m block
-getTipBlockGeneric = getTipSomething "block" (dbGetBlock @_ @block)
+getTipBlock :: MonadDBRead m => m Block
+getTipBlock = getTipSomething "block" dbGetBlock
+
+-- | Returns header of block that was requested from Block DB.
+getHeader
```
gromakovsky: Why is it in `GState` module? It's not stored in GState DB, it's stored in Block DB.

update/Pos/Update/Poll/Modifier.hs
```diff
@@ -1,14 +1,64 @@
--- | Functions working with PollModifier.
+-- | Types related to Poll monad.
```
gromakovsky: Strange change.

db/Pos/DB/Class.hs
```diff
@@ -91,6 +92,14 @@ class DBIteratorClass i where
 
 type IterType i = (IterKey i, IterValue i)
 
+newtype RawBlock = RawBlock
+    { unRawBlock :: ByteString
+    }
+
+newtype RawUndo = RawUndo
+    { unRawUndo :: ByteString
```
gromakovsky: Imo, it'd be better to have `newtype Serialized a = Serialized ByteString`, but for CSL-1854 this solution is enough.

db/Pos/DB/Class.hs
```diff
@@ -104,13 +113,29 @@ class (HasConfiguration, MonadBaseControl IO m, MonadThrow m) => MonadDBRead m w
...
+    dbGetRawBlock = lift . dbGetRawBlock
+    dbGetRawUndo = lift . dbGetRawUndo
+
+
+type MonadBlockDBRead m = (MonadDBRead m, BlockchainHelpers MainBlockchain)
+
+getBlock :: MonadBlockDBRead m => HeaderHash -> m (Maybe Block)
+getBlock x = do
+    mBS <- fmap unRawBlock <$> dbGetRawBlock x
+    pure $ rightToMaybe . decodeFull =<< mBS
```
gromakovsky: Why do you discard decode error? `Nothing` means that there is no block. Decode error should trigger `DBMalfored` exception.

db/Pos/DB/Class.hs
```diff
@@ -138,6 +163,8 @@ class MonadDBRead m => MonadDB m where
     -- with given key from DB corresponding to given tag.
     dbDelete :: DBTag -> ByteString -> m ()
 
+    -- | Put given blund into the Block DB.
+    dbPutRawBlund :: (Block, ByteString) -> m ()
```
gromakovsky: Why not `RawUndo`?

db/Pos/DB/Class.hs
```diff
@@ -39,6 +39,10 @@ module Pos.DB.Class
        , DBIteratorClass (..)
```
gromakovsky: You forget to update the description of this module.

lib/src/Pos/Block/Types.hs
```diff
@@ -33,6 +34,8 @@ instance NFData Undo
 -- | Block and its Undo.
 type Blund = (Block, Undo)
 
+type RawBlund = (Block, ByteString)
```
gromakovsky: Why not `RawUndo`?

lib/src/Pos/DB/Block.hs
```diff
@@ -72,27 +47,49 @@ import           System.IO.Error (IOError, isDoesNotExistError)
...
+
+
+----------------------------------------------------------------------------
+-- BlockDB related methods
+----------------------------------------------------------------------------
+
+getUndo :: MonadDBRead m => HeaderHash -> m (Maybe Undo)
+getUndo x = do
+    mBS <- fmap unRawUndo <$> dbGetRawUndo x
+    pure $ rightToMaybe . decodeFull =<< mBS
```
gromakovsky: Same question (see `getBlock`).

lib/src/Pos/DB/Block.hs
```diff
@@ -496,11 +273,19 @@ getData fp = handle handler $ liftIO $
...
 
+getRawData ::  forall m . MonadIO m => FilePath -> m ByteString
+getRawData = liftIO . BS.readFile
+
 putData ::  (MonadIO m, Bi v) => FilePath -> v -> m ()
 putData fp v = liftIO $
     bracket (openBinaryFile fp WriteMode) hClose $ \h ->
         BS.hPut h (serialize' v) >> hFlush h
 
+putRawData ::  MonadIO m => FilePath -> ByteString -> m ()
```
gromakovsky: Get rid of this copy-paste please.

lib/src/Pos/DB/Block.hs
```diff
@@ -101,151 +98,44 @@ import           Pos.Util.Chrono (NewestFirst (..))
...
--- | Returns header of block that was requested from Block DB.
-blkGetHeader
-    :: (HasConfiguration, MonadDBRead m)
-    => HeaderHash -> m (Maybe BlockHeader)
-blkGetHeader = dbGetBi BlockIndexDB . blockIndexKey
+    => HeaderHash -> m (Maybe ByteString)
+getRawBlock = blockDataPath >=> getData
 
 -- Get undo data for block with given hash from Block DB. This
 -- function has too strict constraint, consider using 'blkGetUndo'.
```
gromakovsky: The comment is outdated.

lib/src/Pos/DB/Block.hs
```diff
@@ -101,151 +98,44 @@ import           Pos.Util.Chrono (NewestFirst (..))
...
-getUndo :: (HasConfiguration, MonadRealDB ctx m) => HeaderHash -> m (Maybe Undo)
-getUndo = undoDataPath >=> getData
+getRawUndo :: (HasConfiguration, MonadRealDB ctx m) => HeaderHash -> m (Maybe ByteString)
+getRawUndo = undoDataPath >=> getData
```
gromakovsky: The implementation is outdated.

lib/src/Pos/DB/Block.hs
```diff
@@ -289,192 +155,103 @@ decodeOrFailPureDB
...
-    => Blund
+    => RawBlund
     -> m ()
-dbPutBlundPureDefault (blk,undo) = do
+dbPutRawBlundPureDefault (blk, rawUndo) = do
+    let blund :: Blund
+        blund =
+            either (error "Couldn't deserialize undo")
+                   (blk,)
+                   (decodeFull rawUndo)
```
gromakovsky: It's absolutely weird to decode and then immediately encode. I suggest that we store blocks and undos separately in pure db. Can you create an issue at least?

update/Pos/Update/Poll/Modifier.hs
```diff
@@ -1,14 +1,55 @@
...
+                                 NumSoftwareVersion, SoftwareVersion, StakeholderId)
+import           Pos.Core.Update (UpId)
+import           Pos.Slotting.Types (SlottingData)
+import           Pos.Update.Poll.Types (BlockVersionState, ConfirmedProposalState, ProposalState)
+import           Pos.Util.Modifier (MapModifier)
+
+-- | PollModifier is used in verification. It represents operation which
+-- one should apply to global state to obtain result of application of
+-- MemPool or blocks which are verified.
+data PollModifier = PollModifier
```
gromakovsky: Why did you move it here?

auxx/src/Mode.hs
```diff
@@ -183,29 +180,14 @@ instance {-# OVERLAPPING #-} CanJsonLog AuxxMode where
 instance HasConfiguration => MonadDBRead AuxxMode where
     dbGet = realModeToAuxx ... dbGet
     dbIterSource tag p = hoist (hoist realModeToAuxx) (dbIterSource tag p)
+    dbGetRawBlock = realModeToAuxx ... DB.dbGetRawBlockRealDefault
```
int-index: This should be `dbGetRawBlock = realModeToAuxx ... dbGetRawBlock` instead. Note that we don't want to choose the `Default` implementation here — if another implementation is used in `Real` mode, we won't have to remember to update `auxx` accordingly.

auxx/src/Mode.hs
```diff
@@ -183,29 +180,14 @@ instance {-# OVERLAPPING #-} CanJsonLog AuxxMode where
...
+    dbGetRawUndo = realModeToAuxx ... DB.dbGetRawUndoRealDefault
```
int-index: Same as above, don't commit to a particular impl here.

auxx/src/Mode.hs
```diff
@@ -183,29 +180,14 @@ instance {-# OVERLAPPING #-} CanJsonLog AuxxMode where
...
-    dbGetUndo   = realModeToAuxx ... dbGetUndo @BlockHeader @Block @Undo
-    dbGetHeader = realModeToAuxx ... dbGetHeader @BlockHeader @Block @Undo
-
-instance (HasConfiguration, HasSscConfiguration) =>
-         MonadBlockDBGeneric (Some IsHeader) SscBlock () AuxxMode
-  where
-    dbGetBlock  = realModeToAuxx ... dbGetBlock
-    dbGetUndo   = realModeToAuxx ... dbGetUndo @(Some IsHeader) @SscBlock @()
-    dbGetHeader = realModeToAuxx ... dbGetHeader @(Some IsHeader) @SscBlock @()
+    dbPutRawBlund = realModeToAuxx ... DB.dbPutRawBlundRealDefault
```
int-index: Same as above, don't commit to a particular impl here.

db/Pos/DB/Class.hs
pva701: I don't think so, I introduced these wrappers to avoid mismatching `Block` and `Undo` ByteStrings

lib/src/Pos/DB/Block.hs
pva701: >I suggest that we store blocks and undos separately in pure db.

we don't, why do you pay attention to it at all? :thinking: 
I don't think it does matter how we store blund in pure db

auxx/src/Mode.hs
pva701: it's already fixed

pva701: can you take away your diss?

update/Pos/Update/Poll/Modifier.hs
pva701: Because when I was moving it to `core` this datatype shouldn't be moved and I decided it should be logical split with other datatypes. Also, I find it more logical to put it here.

auxx/src/Mode.hs
int-index: It's not fixed as far as I can see.

lib/src/Pos/DB/Block.hs
```diff
@@ -72,180 +47,92 @@ import           System.IO.Error (IOError, isDoesNotExistError)
...
--- Implementations for 'MonadRealDB'
+-- BlockDB related methods
 ----------------------------------------------------------------------------
 
--- Get block with given hash from Block DB.  This function has too
--- strict constraint, consider using 'blkGetBlock'.
+getUndo :: MonadDBRead m => HeaderHash -> m (Maybe Undo)
+getUndo x = dbGetRawUndo x >>= \case
+    Nothing      -> pure Nothing
+    Just rawUndo -> eitherToThrow $ first DBMalformed $ decodeFull $ unRawUndo rawUndo
```
gromakovsky: Copy-paste 😒

lib/src/Pos/DB/Block.hs
```diff
@@ -72,180 +47,92 @@ import           System.IO.Error (IOError, isDoesNotExistError)
...
+    => HeaderHash -> m (Maybe ByteString)
+getRawBlock = blockDataPath >=> getData
 
--- Get undo data for block with given hash from Block DB. This
--- function has too strict constraint, consider using 'blkGetUndo'.
-getUndo :: (HasConfiguration, MonadRealDB ctx m) => HeaderHash -> m (Maybe Undo)
-getUndo = undoDataPath >=> getData
+-- Get serialization of an undo data for block with given hash from Block DB.
+getRawUndo :: (HasConfiguration, MonadRealDB ctx m) => HeaderHash -> m (Maybe ByteString)
+getRawUndo = undoDataPath >=> getData
```
gromakovsky: Once again, how is it supposed to work? `getData` performs `decodeFull`, but for `getRaw*` you don't need to do any decoding.

update/Pos/Update/Poll/Modifier.hs
gromakovsky: No, we put types into `Types` modules and other stuff into neighbour modules.

pva701:
>No, we put types into Types modules and other stuff into neighbour modules.

I remember, but I don't find at least one reason to do it for this case

lib/src/Pos/DB/Block.hs
gromakovsky: > we don't

I know that we don't, but I suggest that we do.

> why do you pay attention to it at all?

Because doing unnecessary roundtrip of encoding and decoding is inefficient and it will make tests slower. So I would like to have an issue about it.

gromakovsky:
> I remember, but I don't find at least one reason to do it for this case

Isn't consistency with other code a good reason?

pva701:
no zachem

db/Pos/DB/Class.hs
```diff
@@ -8,6 +8,7 @@
 -- 'MonadDBRead' contains reading and iterating capabilities. The
 -- advantage of it is that you don't need to do any 'IO' to use it
 -- which makes it suitable for pure testing.
+-- 'MonadDBRead' also give you access to the Block DB.
```
volhovM: "gives you", but actually "provides"

db/Pos/DB/Class.hs
volhovM: Pleeease can you _at least_ rename them? We're discussing this CSL-1859 issue and we operate with term `RawBlock` meaning that is an "unverified block", not a "serialized" one. 
I actually find Ivan's idea to be a pretty good thing, because CSL-1869 will anyway introduce a notion of `Serialized a` as Ivan has proposed -- so if you do it now, you minimize the diff.

db/Pos/DB/Class.hs
```diff
@@ -104,13 +106,29 @@ class (HasConfiguration, MonadBaseControl IO m, MonadThrow m) => MonadDBRead m w
...
         MonadDBRead (t m)
   where
     dbGet tag = lift . dbGet tag
     dbIterSource tag (p :: Proxy i) =
         hoist (hoist lift) (dbIterSource tag p)
+    dbGetRawBlock = lift . dbGetRawBlock
+    dbGetRawUndo = lift . dbGetRawUndo
+
+
+type MonadBlockDBRead m = (MonadDBRead m, BlockchainHelpers MainBlockchain)
```
volhovM: Can you please point out where do we use `BlockchainHelpers` instance? I don't really see it.

db/Pos/DB/Class.hs
volhovM: What, do you think, should be added/removed? I will add needed documentation about getting raw data in CSL-1869, so documenting this part is not a strict requirement.

db/Pos/DB/GState/Common.hs
```diff
@@ -6,10 +6,9 @@ module Pos.DB.GState.Common
        (
          -- * Getters
          getTip
+       , getTipSomething
```
volhovM: 1. You only use it in a single place.
2. I would prefer `getTipAbstract` (nitpicker-mode-on).

db/Pos/DB/BlockIndex.hs
```diff
@@ -0,0 +1,36 @@
+-- | Operations with block index db.
+
+module Pos.DB.BlockIndex
```
volhovM: I don't get it. Why would you create a module that provides only read capabilities over block index?

AFAIK every other db module follows the following hierarchy:
1. Reader methods
2. Writer batch functions
3. Iterators
4. Prefix keys

It seems I am missing something important. What exactly?

Update: I now seem to have more understanding on this question (after i've read through all the PR), but I would like you to explain anyway. Like, yes -- we always write blunds only, but design is controversial.

lib/src/Pos/Block/Types.hs
```diff
@@ -33,6 +35,8 @@ instance NFData Undo
 -- | Block and its Undo.
 type Blund = (Block, Undo)
 
+type RawBlund = (Block, RawUndo)
```
volhovM: Same concern applies (as i addressed for `RawBlock`).

lib/src/Pos/Block/Types.hs
```diff
@@ -47,3 +51,4 @@ instance HasDifficulty Blund where
 
 instance HasHeaderHash Block => HasHeaderHash Blund where
     headerHash = headerHash . fst
+
```
volhovM: It's a shame.

lib/src/Pos/DB/Block.hs
```diff
@@ -289,192 +152,103 @@ decodeOrFailPureDB
...
+dbPutRawBlundPureDefault ::
        forall ctx m. (HasConfiguration, MonadPureDB ctx m)
-    => Blund
+    => RawBlund
     -> m ()
-dbPutBlundPureDefault (blk,undo) = do
+dbPutRawBlundPureDefault (blk, rawUndo) = do
+    let blund :: Blund
+        blund =
+            either (error "Couldn't deserialize undo")
```
volhovM: It should be `DBMalformed`.

lib/src/Pos/Util/BackupPhrase.hs
```diff
@@ -23,6 +23,7 @@ import           Pos.Crypto (AbstractHash, EncryptedSecretKey, PassPhrase, Secre
                              unsafeAbstractHash)
 import           Pos.Util.Mnemonics (fromMnemonic, toMnemonic)
 import           Test.QuickCheck (Arbitrary (..), genericShrink)
+import           Test.QuickCheck.Instances ()
```
volhovM: I see no other changes in this file. Why did you decide to import this?

pva701: because it didn't find `Arbitrary Text` instance when I had merged w/ master. dunno why, gromak give advice to me to add this import and it helped

lib/src/Pos/Block/Types.hs
pva701: #antihype

db/Pos/DB/GState/Common.hs
pva701: I use it twice, in this file and in the `lib/DB/Block`

db/Pos/DB/Class.hs
pva701: when deserialize `Block` to validate validness

db/Pos/DB/BlockIndex.hs
pva701: 1. There are only readers methods.
2. We don't write a batch to block index db, just write some information along with putting a block to block db.
3. We don't iterate over headers, we use BlockExtra for it which is stored into GState-db
4. it's done

:paket:

db/Pos/DB/Class.hs
pva701: I propose `newtype Serialized a = Serialized ByteString`

db/Pos/DB/BlockIndex.hs
volhovM: Sorry, but it doesn't really answer my question.
I suggest to export a function instead ofa  prefix key. I find current solution unsafe (everybody from outside can write to `blockIndexKey` something which is different from `HeaderHash`.

db/Pos/DB/Class.hs
pva701: lol, @gromakovsky it seems you edited your first comment

volhovM: The first comment by gromak was there in the same form yesterday evening.
`Serialized a` works for me.

db/Pos/DB/BlockIndex.hs
pva701: breten, vso horosho?
`blockIndexKey :: HeaderHash -> ByteString`

pva701:
@volhovM @gromakovsky have been all your major issues resolved? if so, plz approve and write down minor issues you really would like to fix but it can be postponed to next refactoring PR

db/Pos/DB/Class.hs
gromakovsky: > What, do you think, should be added/removed?

This: https://i.imgur.com/HYITtDr.png

Perhaps those changes were done before you wrote your question, but the thing is they were done after I wrote my comment.

pva701:
@gromakovsky I can, but I really don't think it's important to care about insane behaviour in `PureDB`

gromakovsky:
Tests are slow and I don't think that making them even more slower is a good idea.

pva701:
ah, ok. sound reasonable

pva701:
@volhovM suuda idi

db/Pos/DB/BlockIndex.hs
volhovM: It's not an interface we're providing -- that's what I'm trying to say. We don't usually export index keys.
Yeah, whatever, I guess you won't ever understand :pes: :peka:

db/Pos/DB/Class.hs
gromakovsky: > it seems you've edited your first comment

No, if I edited it, github would show it.

PR #1897, opened 2017-11-11
[CSL-1782] Restart wallet when node exits first

int-index:
The following was happening:

* Daedalus sends a shutdown signal to the node
* The node exits faster than Daedalus
* Launcher treats it as node crash and doesn't restart Daedalus


neongreen:
In which cases does Daedalus attempt to shut down the node?

int-index:
> In which cases does Daedalus attempt to shut down the node?

When you press "Update and restart".

PR #1898, opened 2017-11-12
[CSM-553] Create default address for each account during JSON wallet import

flyingleafe:
Before default address was created only together with default account, if no accounts were specified in JSON backup. This PR introduces logic of creating default addresses for accounts which were specified in backup, but don't have any addresses in Utxo or transaction history.

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -17,8 +17,8 @@ self: {
           pname = "Cabal";
           version = "1.24.2.0";
           sha256 = "b7d0eb8e3503fbca460c0a6ca5c88352cecfe1b69e0bbc79827872134ed86340";
-          revision = "1";
-          editedCabalFile = "0jw809psa2ms9sy1mnirmbj9h7rs76wbmf24zgjqvhp4wq919z3m";
+          revision = "2";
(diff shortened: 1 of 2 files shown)


adinapoli-iohk:
Thanks @flyingleafe, you rock! Will review tomorrow morning and get back to you asap ;)

flyingleafe:
@adinapoli-iohk good point, we actually can do it as long as we have mocks now. I'll try to sort it out today.

adinapoli-iohk:
@flyingleafe Gentle reminder that if you provide some tests, we can get this PR out of the way and make the frontend folks happy 😉

flyingleafe:
@adinapoli-iohk 
I believe that I won't be able to devote time for tests here in next couple of days (I'm afraid I need at least 2-3 hours for this, and it seems like until weekend I'm tightly packed with tasks on API v1).
As soon as this is a kind of blocker for frontend, I think that it's better to merge it now as is, and add tests in separate PR. I created another issue (https://iohk.myjetbrains.com/youtrack/issue/CSM-567) for that, so it will not be forgotten.

PR #1899, opened 2017-11-12
[CSL-1821] Make auxx convenient to use with single command case

hithroc:
`AuxxMode` was replaced with `MonadAuxxMode` constraint to allow more generalized `createCommandProcs`.

`createCommandProcs` is changed to have the configuration constraints and `MonadAuxxMode` passed as arguments in `Maybe (Dict Constraint)` ([more on Dict](https://hackage.haskell.org/package/constraints-0.9.1/docs/Data-Constraint.html#t:Dict)), and returns `[Either Text CommandProc]`. That allows filtering and running the commands that do not require those constraints without having AuxxMode or configuration, like `help` or `hash-installer`.


auxx/Main.hs
```diff
@@ -69,30 +80,34 @@ runNodeWithSinglePlugin ::
...
-            lift $ runReaderT auxxAction auxxContext
-    let vssSK = unsafeFromJust $ npUserSecret nodeParams ^. usVss
-    let sscParams = CLI.gtSscParams cArgs vssSK (npBehaviorConfig nodeParams)
-    bracketNodeResources nodeParams sscParams txpGlobalSettings initNodeDBs $ \nr ->
-        runRealBasedMode toRealMode realModeToAuxx nr $
-            (if aoNodeEnabled then runNodeWithSinglePlugin nr else identity)
-            (auxxPlugin opts command)
+        runWithoutNode = do
+          rawExec Nothing Nothing opts Nothing command
+    (flip catch) (\(_ :: SomeException) -> runWithoutNode) $ withConfigurations conf $ do
```
int-index: 1. `flip catch` is called `handle`
2. don't catch `SomeException`, it could be (1) an async exception, (2) a fatal error

and it looks like `auxxPlugin` itself is within this `catch`: does this mean that if I run with `node` enabled, then fail with an exception during command execution, your handler will catch it and run without node?

I suggest that you don't rely on exceptions when determining whether to run with node or without.

auxx/src/Command/Help.hs
```diff
@@ -54,7 +54,7 @@ withTypeName (TypeNameEither tn1 tn2) needWrap =
     t' = withTypeName tn1 DontNeedWrap <> " | " <>
          withTypeName tn2 DontNeedWrap
 
-mkHelpMessage :: [CommandProc m] -> Text
+mkHelpMessage :: [Either Text (CommandProc m)] -> Text
```
int-index: Since it's a WIP PR I hope you already planned to replace `Text` here with something more informative, at least include the name:

```
data UnavailableCommand = UnavailableCommand
    { ucName :: Name
    , ucReason :: Text
    }
```

auxx/src/Command/Proc.hs
```diff
@@ -51,68 +52,80 @@ import           Lang.Name (Name)
...
-createCommandProcs printAction sendActions = fix $ \commands -> [
-
-    CommandProc
+       forall m. (HasCompileInfo, MonadIO m, CanLog m, HasLoggerName m)
+    => Maybe (Dict HasConfigurations)
+    -> Maybe (Dict (MonadAuxxMode m))
+    -> PrintAction m
+    -> Maybe (SendActions m)
+    -> [Either Text (CommandProc m)]
+createCommandProcs hasConfigurations hasAuxxMode printAction mSendActions = fix $ \commands -> ([
```
int-index: I'd prefer to see `rights . fix` here, rather than `rights` at use site.

auxx/src/Command/Proc.hs
```diff
@@ -51,68 +52,80 @@ import           Lang.Name (Name)
...
 createCommandProcs ::
-       (HasConfigurations, HasCompileInfo)
-    => PrintAction AuxxMode
-    -> SendActions AuxxMode
-    -> [CommandProc AuxxMode]
-createCommandProcs printAction sendActions = fix $ \commands -> [
-
-    CommandProc
+       forall m. (HasCompileInfo, MonadIO m, CanLog m, HasLoggerName m)
+    => Maybe (Dict HasConfigurations)
```
int-index: I don't see a single command that uses `needsConfigurations` but doesn't need `MonadAuxxMode`. Since `MonadAuxxMode` already includes `HasConfigurations`, a separate `Dict` for `HasConfigurations` seems redundant to me.

auxx/src/Command/Proc.hs
```diff
@@ -447,7 +521,7 @@ procConst name value =
     }
 
 class ToLeft a b where
-    toLeft :: Either a b -> AuxxMode a
+    toLeft :: MonadAuxxMode m => Either a b -> m a
```
int-index: `ToLeft` is used for conversions, it's very unfortunate to have this huge `MonadAuxxMode` constraint here, now that your change allows for simpler monads. Instead, add `m` as a class parameter and move the constraint to specific instances:

```
class ToLeft m a b where
     toLeft :: Either a b -> m a

 instance MonadAuxxMode m => ToLeft Address PublicKey where
```

auxx/src/Mode.hs
```diff
@@ -79,6 +81,43 @@ data CmdCtx = CmdCtx
     }
 
 type AuxxMode = ReaderT AuxxContext Production
+type MonadAuxxMode m =
```
int-index: Nice constraint list. Initially I thought to suggest replacing it with `WorkMode`, but it looks like it's actually a bit smaller (doesn't include `Slog`, for instance), so let's leave it as is.

auxx/src/Plugin.hs
```diff
@@ -46,30 +48,61 @@ import           Repl (WithCommandAction (..))
...
 
 auxxPlugin ::
-       (HasConfigurations, HasCompileInfo)
-    => AuxxOptions
-    -> Either (WithCommandAction AuxxMode) Text
-    -> (WorkerSpec AuxxMode, OutSpecs)
-auxxPlugin AuxxOptions{..} = \case
-    Left WithCommandAction{..} -> worker' runCmdOuts $ \sendActions -> do
+       (HasCompileInfo, MonadAuxxMode m, Mockable Delay m)
+    => Dict HasConfigurations
```
int-index: Doesn't make any sense, since `MonadAuxxMode m` includes `HasConfigurations`

auxx/src/Command/Proc.hs
```diff
@@ -51,68 +52,80 @@ import           Lang.Name (Name)
...
+    do
+        Dict <- needsAuxxMode
+        return CommandProc
+            { cpName = "s"
+            , cpArgumentConsumer = getArg (tyPublicKey `tyEither` tyInt) "pk"
+            , cpExec = fmap ValueStakeholderId . toLeft . Right
+            , cpHelp = "stakeholder id (hash) of the specified public key"
+            },
+
+    do
```
int-index: to avoid indentation changes I suggest doing it like this:

```
needsAuxxMode >>= \Dict ->
return CommandProc
    { ...
    }
```

auxx/src/Command/Help.hs
hithroc: Yes, this is a planned change.

auxx/Main.hs
gromakovsky: > (1) an async exception

It's wrong, `catch` from `Universum` doesn't catch async exceptions. Other than that, I agree with
> I suggest that you don't rely on exceptions when determining whether to run with node or without.

int-index:
@gromakovsky The `MonadAuxxMode` is strictly more general. Two observations:

* we'll need `Dict` either way, because there's also `HasConfigurations`
* some commands (which must work in both `AuxxMode` and `IO`) will need to be in abstract monads anyway

Therefore, for uniformity of the approach, passing `Maybe (Dict (MonadAuxxMode m))` looks good to me.

gromakovsky:
Okay, I think for now both options will do and it's not essential which one is chosen, because CSL-1823 is coming.

auxx/src/Command/Proc.hs
```diff
@@ -189,47 +208,56 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
         Tx.sendToAllGenesis sendActions stagp
         return ValueUnit
     , cpHelp = "create and send transactions from all genesis addresses \
-               \ for <duration> seconds, <delay> in ms. <conc> is the \
-               \ number of threads that send transactions concurrently. \
-               \ <mode> is either 'neighbours', 'round-robin', or 'send-random'"
```
int-index: accidental whitespace change

auxx/src/Command/Proc.hs
```diff
@@ -189,47 +208,56 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
...
+    { cpName = name
     , cpArgumentConsumer =
         (,) <$> getArg tyInt "i"
             <*> getArgSome tyTxOut "out"
     , cpExec = \(i, outputs) -> do
         Tx.send sendActions i outputs
         return ValueUnit
     , cpHelp = "send from #i to specified transaction outputs \
-               \ (use 'tx-out' to build them)"
+              \ (use 'tx-out' to build them)"
```
int-index: accidental whitespace change

auxx/src/Command/Proc.hs
```diff
@@ -189,47 +208,56 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
...
+    let name = "vote" in
+    needsSendActions name >>= \sendActions ->
+    needsAuxxMode name >>= \Dict ->
+    return CommandProc
+    { cpName = name
     , cpArgumentConsumer =
         (,,) <$> getArg tyInt "i"
-             <*> getArg tyBool "agree"
-             <*> getArg tyHash "up-id"
+            <*> getArg tyBool "agree"
```
int-index: accidental whitespace change

auxx/src/Command/Proc.hs
```diff
@@ -189,47 +208,56 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
...
+            <*> getArg tyHash "up-id"
     , cpExec = \(i, decision, upId) -> do
         Update.vote sendActions i decision upId
         return ValueUnit
     , cpHelp = "send vote for update proposal <up-id> and \
-               \ decision <agree> ('true' or 'false'), \
-               \ using secret key #i"
```
int-index: accidental whitespace change

auxx/src/Command/Proc.hs
```diff
@@ -278,20 +309,23 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
         -- one with existential (relevant via singleton-style GADT) quantification.
         ValueHash . unsafeCheatingHashCoerce <$> Update.propose sendActions params
     , cpHelp = "propose an update with one positive vote for it \
-               \ using secret key #i"
+              \ using secret key #i"
```
int-index: whitespace

auxx/src/Command/Proc.hs
```diff
@@ -413,29 +460,37 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
             addrHD <- deriveHDAddressAuxx sk
             printAction $
                 sformat ("    #"%int%":   addr:      "%build%"\n"%
-                         "          pk:        "%fullPublicKeyF%"\n"%
-                         "          pk hash:   "%hashHexF%"\n"%
-                         "          HD addr:   "%build)
+                        "          pk:        "%fullPublicKeyF%"\n"%
```
int-index: whitespace

auxx/src/Mode.hs
```diff
@@ -79,6 +81,43 @@ data CmdCtx = CmdCtx
...
+    , Eq (Promise m (Maybe ())) -- what?
+    , MonadRandom m
+    , Default (MempoolExt m)
+    , MonadTxpLocal (BlockGenMode (MempoolExt m) m)
+    , MonadFail m
+    )
+
+-- For MonadAuxxMode we need to make sure that AddrData m in MonadAdresses
+-- is indeed PublicKey, as it is for AuxxMode.
+class (AddrData m ~ PublicKey) => AddrDataIsPublicKey m
```
int-index: I don't understand why this class is defined. Can't you just use `AddrData m ~ PublicKey` above?

Besides, seeing that you have only a single instance here, `instance AddrDataIsPublicKey`, it made me realize that you could do this:

```
type MonadAuxxMode m = m ~ AuxxMode
```

hithroc:
The `--node-enabled` CLI option was replaced with `--mode` which accepts one of the 4 possible arguments:

* `light`, start auxx in light mode, no configuration required, minimal subset of commands
* `with-config`, start auxx in `AuxxMode`, full set of commands, full configuration required
* `with-node`, same as `with-config`, but starts auxx as a node plugin.
* `auto`, tries to load the configuration on startup and start in `with-config` mode, if fails starts in `light` mode.

auxx/src/Mode.hs
gromakovsky: For `Slog` there are only `Has` classes, so it's not required because this list contains `MonadReader AuxxContext m` :)
Also it looks a bit weird, because if you have `MonadReader AuxxContext` and some basic stuff which `Production` has, you have all other classes from this list.

auxx/Main.hs
```diff
@@ -69,30 +81,47 @@ runNodeWithSinglePlugin ::
...
-            (auxxPlugin opts command)
+        runWithoutNode = do
+          rawExec Nothing opts Nothing command
+
+    let configToDict :: HasConfigurations => Production (Maybe (Dict HasConfigurations))
+        configToDict = return (Just Dict)
+
+    hasConfigurations <- case aoStartMode of
+        Automatic -> (flip catch) (\(_ :: ConfigurationException) -> return Nothing)
+                    . (flip catch) (\(_ :: ConfigurationError)     -> return Nothing)
```
gromakovsky: Also I think it'd be good to print a message like:
> Using light mode because \<reason\>

because sometimes user may be sure they have valid configuration, but something may be unexpectedly wrong.

auxx/Main.hs
```diff
@@ -22,17 +23,28 @@ import           Pos.Network.Types (NetworkConfig (..), Topology (..), topologyD
...
-import           AuxxOptions (AuxxAction (..), AuxxOptions (..), getAuxxOptions)
+import           AuxxOptions (AuxxAction (..), AuxxOptions (..), AuxxStartMode (..), getAuxxOptions)
 import           Mode (AuxxContext (..), AuxxMode, CmdCtx (..), realModeToAuxx)
-import           Plugin (auxxPlugin)
+import           Plugin (auxxPlugin, rawExec)
 import           Repl (WithCommandAction, withAuxxRepl)
 
 -- 'NodeParams' obtained using 'CLI.getNodeParams' are not perfect for
 -- Auxx, so we need to adapt them slightly.
-correctNodeParams :: AuxxOptions -> NodeParams -> Production (NodeParams, Bool)
```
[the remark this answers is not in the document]
hithroc: Okay, that is a leftover of when I was trying to meddle with everything without generalizing `WithCommandAction`. There's no actual need to generalize it, yes.

auxx/Main.hs
hithroc: I thought about that, but that makes it more complicated to use in scripts, where sometimes we want to check the output of a specific command. Not sure if that's even an issue or not, so I'll add the message.

auxx/Main.hs
```diff
@@ -69,30 +71,52 @@ runNodeWithSinglePlugin ::
...
+        configToDict = return (Just Dict)
+
+    hasConfigurations <- case aoStartMode of
+        Automatic -> do
+            mode <- (flip catch) (\(_ :: ConfigurationException) -> return Nothing)
+                  . (flip catch) (\(_ :: ConfigurationError)     -> return Nothing)
+                  $ withConfigurations conf configToDict
+            case mode of
+                Nothing -> putStrLn ("Starting in light mode" :: Text) >> return Nothing
+                x -> putStrLn ("Starting in standalone mode" :: Text) >> return x
```
int-index: First, I think we called this mode "with-config", not "standalone" (but I'm okay with a rename if you do it everywhere in the codebase)

Second, I'd like to see "mode" as the first word in the message, like this:

```
Mode: with-config
```

so it would be easier to find visually in the logs

auxx/Main.hs
```diff
@@ -69,30 +71,52 @@ runNodeWithSinglePlugin ::
...
-            (auxxPlugin opts command)
+        runWithoutNode = do
+          rawExec Nothing opts Nothing command
+
+    let configToDict :: HasConfigurations => Production (Maybe (Dict HasConfigurations))
+        configToDict = return (Just Dict)
+
+    hasConfigurations <- case aoStartMode of
+        Automatic -> do
+            mode <- (flip catch) (\(_ :: ConfigurationException) -> return Nothing)
```
int-index: `handle @ConfigurationException (const Nothing)` would be more concise

auxx/src/Command/Proc.hs
```diff
@@ -278,20 +309,23 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
         -- one with existential (relevant via singleton-style GADT) quantification.
         ValueHash . unsafeCheatingHashCoerce <$> Update.propose sendActions params
     , cpHelp = "propose an update with one positive vote for it \
-               \ using secret key #i"
+             \ using secret key #i"
```
int-index: :unamused: Can you check for whitespace issues yourself?

auxx/src/Mode.hs
```diff
@@ -80,6 +81,9 @@ data CmdCtx = CmdCtx
 
 type AuxxMode = ReaderT AuxxContext Production
 
+class (m ~ AuxxMode, HasConfigurations, HasCompileInfo) => MonadAuxxMode m
+instance (m ~ AuxxMode, HasConfigurations, HasCompileInfo) => MonadAuxxMode AuxxMode
```
int-index: You already have `m ~ AuxxMode` to the left, just leave it as `m` on the RHS

auxx/src/Command/Proc.hs
```diff
@@ -134,33 +144,37 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
...
     , cpExec = \parts -> do
         distr <- case parts of
             AddrDistrPart sId coinPortion :| []
               | coinPortion == maxBound ->
                 return $ SingleKeyDistr sId
             _ -> eitherToFail $
-                 mkMultiKeyDistr . Map.fromList $
-                 map (\(AddrDistrPart s cp) -> (s, cp)) $
-                 toList parts
+                mkMultiKeyDistr . Map.fromList $
```
int-index: whitespace

auxx/src/Command/Proc.hs
```diff
@@ -428,14 +475,22 @@ createCommandProcs printAction sendActions = fix $ \commands -> [
...
-    CommandProc
+    return CommandProc
     { cpName = "help"
     , cpArgumentConsumer = do pure ()
     , cpExec = \() -> do
         printAction (mkHelpMessage commands)
         return ValueUnit
     , cpHelp = "display this message"
-    }]
+    }] :: [Either UnavailableCommand (CommandProc m)]) -- The explicit type signature is required
```
int-index: it shouldn't be required, figure it out please

auxx/Main.hs
gromakovsky: So now you print the mode, but don't print the reason. I think printing the reason is useful, because you may be sure that everything is ok for `with-config` mode, but for some reason `Light` one is used.
But I also realized that then you can pass `--mode with-config` and you will see the error message, so ok.

PR #1903, opened 2017-11-14
[CSL-1859] Serialization/verification granularity proposal

volhovm:
So this is a "proposal" PR. It will only be merged after we agree on proposal document. Proposal document is what is going to determine specification document/implementation later.

Discussions to be held here in comments only. Whenever we come up with a need to change doc, change it.

No force-pushes, please! This is a public branch!

### DIFF
--- a/docs/proposals/serialization.md
+++ b/docs/proposals/serialization.md
@@ -101,20 +101,6 @@ need to agree on is Q2: U1 is irrelevant, U4 is solved by
 deserializing to unsafe `T` or `T0` (and maybe skipping D2 D3 as
 suggested by Q1).
 
-<!--
-If we aim to do U2.1, i think we can implement it using
-newtypes. E.g. if we have `S` as subfield of `T` and we only want to
-deserialize `S`, we can create `newtype SAsT = SAsT S` which will be
-encoded as `S`, but decoded as `T`, though avoiding other `T`'s
-fields.
-
-Another suggestion is to use `data SAsT = ...` effectively duplicating
-T, but containing only one related subfield `S`.
-
-And yes, maybe default haskell laziness can help us here too (though
-I'm not sure). What if `S` subfield of `T` is not strict?
--->
-
 ### Solution 1: Approach with "RawX and X".
 
 Described in https://github.com/input-output-hk/cardano-sl/pull/1823#issuecomment-342475355
@@ -206,3 +192,11 @@ lenses applicable to `T`). To construct `Verified Fully T` you'll need
 to do full verification and/or serialization with all checks. It is
 much more verbose than using phantom types, though it is not obvious
 whether using this approach will produce bigger diff.
+
+## Decision
+
+Decision was made in the following PR:
+https://github.com/input-output-hk/cardano-sl/pull/1903
+
+We will adopt solution 3 (phantom types) and extend it with type families (as 
+suggested in the PR) if needed.
--- a/docs/proposals/serialization.md
+++ b/docs/proposals/serialization.md
@@ -185,3 +185,27 @@ CSL code/core uses valid block already and unsafe blocks will be used
 in the diffusion layer only, so these module sets have minimum number
 of intersections thus minimizing ability to make a mistake too.
 
+### Solution 5: `Verified a` newtype
+
+Introduce newtype:
+```
+data Fully
+
+newtype Verified level s = Verified s
+```
+
+Then rewrite definitions in the following way:
+
+```
+-- old:
+applyBlock :: GState -> Block -> GState
+
+-- new:
+applyBlock :: GState -> Verified Fully Block -> GState
+```
+
+For type `T` you'll need to define lenses for `Verified a T` (all
+lenses applicable to `T`). To construct `Verified Fully T` you'll need
+to do full verification and/or serialization with all checks. It is
+much more verbose than using phantom types, though it is not obvious
+whether using this approach will produce bigger diff.
--- a/docs/proposals/serialization.md
+++ b/docs/proposals/serialization.md
@@ -14,8 +14,6 @@ consists of the following steps:
  * __(D2)__ Check CBOR structure (bytestring is a correct CBOR encoded data)
  * __(D3)__ Check datatype canonicity
  * __(D4)__ Construct/deserialize datatype `T`
-   * __(D4.1)__ An option is to deserialize only some part of data (header
-     of block)
  * __(D5)__ Validate `T`.
    * __(D5.1)__ Validate parts of `T`.
 
@@ -33,19 +31,15 @@ them in terms of use cases:
      * Only D1 is needed (just get those bytes and send them).
  * __(U2)__ Diffusion (as designed for Shelley) of blocks: receive block
    and distribute it on fly.
-     * __(U2.1)__ D1, D2, D3 and D4.1 are needed. Regarding blocks, we don't want
-       to deserialize payload at all, checking header is enough.
-     * __(U2.2)__ One can argue that may want to have D4 + D5.1 instead of
-       D4.1, if it's not significantly slower.
+     * We need D4 + D5.1 (complete validation is not required). We don't
+       want to have partial deserialization (though it's possible).
  * __(U3)__ Apply received block(s) to GState.
      * D1-D5 (complete procedure) are needed.
  * __(U4)__ Reading data (sometimes lots of it) from DB for different
    node local actions (LRC, GState checks, API for Middleware -- all
    of them require db traversing or LCA computations).
-     * D1,D4 needed (we trust local database, so we want to turn off
-       as much checks as it's possible).
-     * AFAIU @avieth assumes D2-D3 is also needed and we shouldn't
-       optimize here because it's too dangerous anyway.
+     * D1,D2,D4 needed. We can omit D3: we trust local database, 
+       so let's turn off as much checks as it's possible.
 
 ## Solutions and comparison.
 
@@ -102,22 +96,22 @@ like? Well, you should consider the fact that now you're obliged to
 perform all the subfields checks by yourself. Hopefully, this will be
 easy to maintain. Just a notice.
 
-I claim that if we're aiming to solve U2.2 then the only thing we
+I claim that if we're aiming to solve U2 then the only thing we
 need to agree on is Q2: U1 is irrelevant, U4 is solved by
 deserializing to unsafe `T` or `T0` (and maybe skipping D2 D3 as
 suggested by Q1).
 
-If we aim to do U2.1, i think we can implement it using
-newtypes. E.g. if we have `S` as subfield of `T` and we only want to
-deserialize `S`, we can create `newtype SAsT = SAsT S` which will be
-encoded as `S`, but decoded as `T`, though avoiding other `T`'s
-fields.
-
-Another suggestion is to use `data SAsT = ...` effectively duplicating
-T, but containing only one related subfield `S`.
-
-And yes, maybe default haskell laziness can help us here too (though
-I'm not sure). What if `S` subfield of `T` is not strict?
+# If we aim to do U2.1, i think we can implement it using
+# newtypes. E.g. if we have `S` as subfield of `T` and we only want to
+# deserialize `S`, we can create `newtype SAsT = SAsT S` which will be
+# encoded as `S`, but decoded as `T`, though avoiding other `T`'s
+# fields.
+# 
+# Another suggestion is to use `data SAsT = ...` effectively duplicating
+# T, but containing only one related subfield `S`.
+# 
+# And yes, maybe default haskell laziness can help us here too (though
+# I'm not sure). What if `S` subfield of `T` is not strict?
 
 ### Solution 1: Approach with "RawX and X".
 
--- a/docs/proposals/serialization.md
+++ b/docs/proposals/serialization.md
@@ -101,17 +101,19 @@ need to agree on is Q2: U1 is irrelevant, U4 is solved by
 deserializing to unsafe `T` or `T0` (and maybe skipping D2 D3 as
 suggested by Q1).
 
-# If we aim to do U2.1, i think we can implement it using
-# newtypes. E.g. if we have `S` as subfield of `T` and we only want to
-# deserialize `S`, we can create `newtype SAsT = SAsT S` which will be
-# encoded as `S`, but decoded as `T`, though avoiding other `T`'s
-# fields.
-# 
-# Another suggestion is to use `data SAsT = ...` effectively duplicating
-# T, but containing only one related subfield `S`.
-# 
-# And yes, maybe default haskell laziness can help us here too (though
-# I'm not sure). What if `S` subfield of `T` is not strict?
+<!--
+If we aim to do U2.1, i think we can implement it using
+newtypes. E.g. if we have `S` as subfield of `T` and we only want to
+deserialize `S`, we can create `newtype SAsT = SAsT S` which will be
+encoded as `S`, but decoded as `T`, though avoiding other `T`'s
+fields.
+
+Another suggestion is to use `data SAsT = ...` effectively duplicating
+T, but containing only one related subfield `S`.
+
+And yes, maybe default haskell laziness can help us here too (though
+I'm not sure). What if `S` subfield of `T` is not strict?
+-->
 
 ### Solution 1: Approach with "RawX and X".
 
--- a/docs/proposals/serialization.md
+++ b/docs/proposals/serialization.md
@@ -1,0 +1,187 @@
+# Block retrieval optimization
+
+This document aims to provide a detailed overview of a certain
+development problem and suggest/compare several possible solutions. We
+will mostly talk about _speeding up_ data
+deserialization/verification and diffusion interface.
+
+## Terms and definitions
+
+Arbitrary datatype `T` construction/deserialization routine usually
+consists of the following steps:
+
+ * __(D1)__ Read bytes (from network or any type of storage)
+ * __(D2)__ Check CBOR structure (bytestring is a correct CBOR encoded data)
+ * __(D3)__ Check datatype canonicity
+ * __(D4)__ Construct/deserialize datatype `T`
+   * __(D4.1)__ An option is to deserialize only some part of data (header
+     of block)
+ * __(D5)__ Validate `T`.
+   * __(D5.1)__ Validate parts of `T`.
+
+Current codebase only provides interface to do all of it at
+once. User runs D1 and feeds bytestring to decoder. Decoder performs
+D2-D3, parses subfields and then constructs datatype (D4) with a
+special constructor (D5).
+
+It is currently impossible to perform (D4.1) and (D5.1), though it can
+be implemented.
+
+Now let's describe features that we'd like to have. We'll formulate
+them in terms of use cases:
+ * __(U1)__ Respond block requests quickly.
+     * Only D1 is needed (just get those bytes and send them).
+ * __(U2)__ Diffusion (as designed for Shelley) of blocks: receive block
+   and distribute it on fly.
+     * __(U2.1)__ D1, D2, D3 and D4.1 are needed. Regarding blocks, we don't want
+       to deserialize payload at all, checking header is enough.
+     * __(U2.2)__ One can argue that may want to have D4 + D5.1 instead of
+       D4.1, if it's not significantly slower.
+ * __(U3)__ Apply received block(s) to GState.
+     * D1-D5 (complete procedure) are needed.
+ * __(U4)__ Reading data (sometimes lots of it) from DB for different
+   node local actions (LRC, GState checks, API for Middleware -- all
+   of them require db traversing or LCA computations).
+     * D1,D4 needed (we trust local database, so we want to turn off
+       as much checks as it's possible).
+     * AFAIU @avieth assumes D2-D3 is also needed and we shouldn't
+       optimize here because it's too dangerous anyway.
+
+## Solutions and comparison.
+
+First of all it is needed to say that solving U1 is completely
+orthogonal to solving U2-U4. We will just modify database typeclass so
+it is possible to read raw data instead of already deserialized
+value. So let's move to U2-U4.
+
+Current state of CSL-1399 PR aims to cover U4 only, as it was
+described in the YT task. It does it in a wery compact and concise
+way, though preventing us from implementing U2. Decoupling
+(de)serialization and datatype verification would give us more
+flexibility to solve U2. At least it seems to be. Before I move to
+suggesting other solutions, I find it important to say that CSL-1399
+PR idea can be applied to get rid of D2-D3 as it's required by
+U4. Still, if we go with decoupling approach, it might be a good idea
+to leave ReaderT (or patch cborg) to skip heavy D3. Let's call this
+suggestion __Q1__ to refer to it later.
+
+Let's have a closer look at the validation process itself. Currently
+we have a bunch of types `T` that conform the following structure:
+ 1. Defined as `data T = UnsafeT { .. }`
+ 2. There is `mkT` function that works in the same constructor
+    does. It defines several checks that define internal validity of
+    `T`.
+ 3. We support invariant "every instance of type `T` is valid" by not
+    using `UnsafeT` where it is possible to construct incorrect `T`.
+
+Pros of this approach:
+ * It's easy to implement
+ * It integrates into decoder nicely so we shouldn't ever think that
+   `T`'s subfield `S` can be invalid. If it's `decode`d, it's
+   instantiated. Then it's valid. We can simply feed `S` to `makeT`.
+
+Cons of this approach is that it, well, combines verification and
+deserialization, which we want to avoid.
+
+The first dichotomy i would like to highlight is:
+ * __(Q2)__ is it alright to break invariant (3)? Rephrasing: may we
+   assume that type `T` can be valid to different degree? Or we should
+   create `Ti` for each validation step? (like `T0` -- no checks
+   enforced, `T1` -- some checks enforced, ... `Tn` -- valid). If yes,
+   which `n` should we take?
+
+We need to figure out whether it's alright to accept Q2, it's a tough
+tradeoff: if we reject Q2 it can lead to massive code duplication and
+unnecessary overabstraction.
+
+There's another issue with `T` verification. Imagine it's now
+completely separate and we already have `T` deserialized, but not
+verified (`T` or `T0` according to `Q1`). How would `verifyT` look
+like? Well, you should consider the fact that now you're obliged to
+perform all the subfields checks by yourself. Hopefully, this will be
+easy to maintain. Just a notice.
+
+I claim that if we're aiming to solve U2.2 then the only thing we
+need to agree on is Q2: U1 is irrelevant, U4 is solved by
+deserializing to unsafe `T` or `T0` (and maybe skipping D2 D3 as
+suggested by Q1).
+
+If we aim to do U2.1, i think we can implement it using
+newtypes. E.g. if we have `S` as subfield of `T` and we only want to
+deserialize `S`, we can create `newtype SAsT = SAsT S` which will be
+encoded as `S`, but decoded as `T`, though avoiding other `T`'s
+fields.
+
+Another suggestion is to use `data SAsT = ...` effectively duplicating
+T, but containing only one related subfield `S`.
+
+And yes, maybe default haskell laziness can help us here too (though
+I'm not sure). What if `S` subfield of `T` is not strict?
+
+### Solution 1: Approach with "RawX and X".
+
+Described in https://github.com/input-output-hk/cardano-sl/pull/1823#issuecomment-342475355
+
+It solves the problem by rejecting Q2. Though, there is a serious
+drawback with it: You must propagate "Raw" to all record
+fields. Imagine you have the following datatype set
+
+```
+data C = C { ... }
+data B = B { bC :: C, ... }
+data A = A { aB :: B, ... }
+```
+
+Now imagine you want to turn off validation of C and A. It is
+impossible to do it without propagating "Raw" to B:
+
+```
+data C = C { ... }
+data RawC = RawC { ... }
+
+data B = B { bC :: C, ... }
+data RawB = RawB { bC :: RawC, ... }
+
+data A = A { aB :: B, ... }
+data RawA = A { aB :: RawB, ... }
+```
+
+You now should see where it is going. Imagine you want to have
+`RawB1` (which doesn't check everything) and `RawB2` (which checks
+only `bC`, but not, say, `bD`): how would you implement it without
+duplicating datatypes?
+
+### Solution 2: using GADT/DataKinds
+
+Another no-Q2 solution suggested in
+https://github.com/input-output-hk/cardano-sl/pull/1823#issuecomment-342478781
+
+Has the very same drawback as the first solution. It is actually a
+cosmetic improvement that gives another tradeoff:
+1. Now you can't mistakenly define `RawA` and `A` in different
+   modules (which feels like chaos).
+2. But now you should use this parameter `v` everywhere you use block.
+
+### Solution 3: phantom types
+
+Same as 3, but instead of having two constructors you now have:
+
+```
+data Block (v :: Verification) = Block {..}
+```
+
+This doesn't have the (bad) possibility to have different
+implementations of block for verified and unverified parameter.
+
+### Solution 4: watching your code.
+
+Q2-accepting solution that works essentially as:
+1. Use `T`, but use implicit function invariants (e.g. "this function
+   works on valid `T` only").
+2. Use `verifyT` or partial verifications.
+
+The main idea behind the confidence it will work is that most of the
+CSL code/core uses valid block already and unsafe blocks will be used
+in the diffusion layer only, so these module sets have minimum number
+of intersections thus minimizing ability to make a mistake too.
+
--- a/docs/proposals/serialization.md
+++ b/docs/proposals/serialization.md
@@ -55,7 +55,7 @@ it is possible to read raw data instead of already deserialized
 value. So let's move to U2-U4.
 
 Current state of CSL-1399 PR aims to cover U4 only, as it was
-described in the YT task. It does it in a wery compact and concise
+described in the YT task. It does it in a very compact and concise
 way, though preventing us from implementing U2. Decoupling
 (de)serialization and datatype verification would give us more
 flexibility to solve U2. At least it seems to be. Before I move to
@@ -75,20 +75,21 @@ we have a bunch of types `T` that conform the following structure:
     using `UnsafeT` where it is possible to construct incorrect `T`.
 
 Pros of this approach:
- * It's easy to implement
+ * It was easy to implement (and it is concise).
  * It integrates into decoder nicely so we shouldn't ever think that
    `T`'s subfield `S` can be invalid. If it's `decode`d, it's
    instantiated. Then it's valid. We can simply feed `S` to `makeT`.
+ * Aligns with Q1 pretty well.
 
 Cons of this approach is that it, well, combines verification and
 deserialization, which we want to avoid.
 
 The first dichotomy i would like to highlight is:
- * __(Q2)__ is it alright to break invariant (3)? Rephrasing: may we
-   assume that type `T` can be valid to different degree? Or we should
-   create `Ti` for each validation step? (like `T0` -- no checks
-   enforced, `T1` -- some checks enforced, ... `Tn` -- valid). If yes,
-   which `n` should we take?
+ * __(Q2)__ is it alright to break invariant (3) (see 2 paragraphs
+   before)? Rephrasing: may we assume that type `T` can be valid to
+   different degree? Or we should create `Ti` for each validation
+   step? (like `T0` -- no checks enforced, `T1` -- some checks
+   enforced, ... `Tn` -- valid). If yes, which `n` should we take?
 
 We need to figure out whether it's alright to accept Q2, it's a tough
 tradeoff: if we reject Q2 it can lead to massive code duplication and


volhovM:
Alright, so i cleaned up conversation history a bit. I will post comments that are to be discussed here.

volhovM:
by @gromakovsky :

>There's another issue with T verification. Imagine it's now completely separate and we already have T deserialized, but not verified (T or T0 according to Q1). How would verifyT look like? Well, you should consider the fact that now you're obliged to perform all the subfields checks by yourself. Hopefully, this will be easy to maintain. Just a notice.

Yes, that's another advantage of the current approach. For instance, we have:
```
mkTx
    :: MonadFail m
    => NonEmpty TxIn -> NonEmpty TxOut -> TxAttributes -> m Tx
```
Inside `mkTx` we can assume that `TxIn`s and `TxOut`s are valid, so we only need to check invariants of `Tx` itself.
If we have `verifyTx :: MonadFail m => Tx -> m ()`, then we should also call `verifyTxIn` for all `TxIn`s and `verifyTxOut` for all `TxOut`s.
One can argue that `TxIn` and `TxOut` are simple types without any invariants. However:
1. `TxOut` contains `Coin` which has an internal invariant, so with `verify` approach we have to also have `verifyTxOut`.
2. `TxIn` doesn't have nested invariants, so we indeed don't need to use `verifyTxIn`, but can we be sure that we will never add any checks to `TxIn`? And if we do it, we may forget to use `verifyTxIn` in `verifyTx` and the code will compile (unless we do something sophisticated, but I don't know what).

Even if we call `verifyX` for all `X`s from some type `Y`, we may later add another field to `Y` and forget to add `verifyY` usage to `verifyX`.

P. S. Perhaps it can be solved using TH.

**Upd**: after I wrote the comment above, I read the article fully and remembered about verified/unverified types. I see that it's not that bad.
Also I remembered that the key problem with what we have now is that we can't have unverified value, but we need it for diffusion layer.

> data Block (v :: Verification) = Block {..}

Does `v` apply to all fields also? Will we have
```
data Tx (v :: Verification) = Tx
    { _txInputs     :: !(NonEmpty TxIn)  -- ^ Inputs of transaction.
    , _txOutputs    :: !(NonEmpty (TxOut v)) -- ^ Outputs of transaction.
    , _txAttributes :: !TxAttributes     -- ^ Attributes of transaction
    } deriving (Eq, Ord, Generic, Show, Typeable)
```
?

volhovM:
Dialog from slack with @avieth:


# Alexander Vieth:

> We need to figure out whether it's alright to accept Q2, it's a tough tradeoff: if we reject Q2 it can lead to massive code duplication and unnecessary overabstraction.

Yes, it _can_ lead to this, but I don't think it necessarily will.

> [Solution 1] solves the problem by rejecting Q2. Though, there is a serious drawback with it: You must propagate "Raw" to all record fields.

I don't think you have to propagate it to all record fields. The way I imagine it is to define, as this "raw" type, something which matches the serialized format almost exactly, and to think about it as a fundamentally different thing from its "non-raw" counterpart. This is somewhat relevant to...

> define RawA and A in different modules (which feels like chaos).

I think defining them in different modules could actually be a great thing. Having a datatype which corresponds exactly to the serialized format of a block is good, because the serialized format of a block really is an essential part of what we're making here. At the moment it's defined only implicitly: as the `ByteString` image of some function, rather than as a datatype proper.

Stated another way: it's important to be able to speak about (in Haskell) the block as it appears out there in the network (possibly invalid) versus the block as we would like it to appear while we use it to do blockchain logic (always valid, with certain transformations applied like a merkle tree in the tx payload).

> Can we choose `U2.2` over `U2.1`

For diffusion layer block deserialization I think all we will need is to decode the bytes into `(header, body)` where `body` can just be a `ByteString` (undecoded, unverified) and `header` is a decoded but unverified block header. That's because the diffusion layer never needs to know anything about the contents of the body, but it will need to know some header details: who created the block, what's the header hash (to use as a key in the cache), etc.

# Mikhail Volkhov

> I don't think it necessarily will.
> I don't think you have to propagate it to all record fields

We yet didn't come up with a proper solution that can prove your point, so I believe otherwise (and I'm honestly trying :(). I just can't imagine any solution that will:
1. Will provide us with `T0` that is essentially `T` but with other verification level
2. Will not propagate the notion of "verified"ness into record fields.

It seems impossible (and i've already pointed it out in my writeup).

>corresponds exactly to the serialized format of a block is good

You're missing the point here, i was highlighting Q2 tradeoff, rather than "having a special type for serialized data". See the part before "solution 1". 

> it's important to be able to speak about (in Haskell) the block as it appears out there in the network (possibly invalid) versus the block as we would like it to appear while we use it to do blockchain logic

It's an obvious thing that you would like to have, but I thought we (you and George) have already agreed on requirements (U1-U4) that are more flexible.

>I think all we will need is to decode the bytes into `(header, body)`...

Sorry, but it doesn't answer my question. What you state in this paragraph is that "it would be good to use U2.1" (this is already a requirement so there's no need to highlight that it's needed) and what i'm asking is to (try to) estimate whether U2.2 may be the same as U2.1. U2.2 may be just an easier requirement to implement yielding almost the same results.

https://gist.github.com/volhovM/5d356d68f77f56080a4ec28a57d728e7
I've updated the doc (added solution (3) with phantom types)

[16:48] 
It also has problem with "verifiability" propagation.

# Alexander Vieth

> > I don't think it necessarily will.
> > I don't think you have to propagate it to all record fields
> We yet didn't come up with a proper solution that can prove your point

Ok I suppose the mismatch here is what each of us considers to be (un)necessary. I've come to the conclusion that a verified block, and a (possibly invalid) serialized block, are two different things which merit two different types. Each is of interest to us, but must not be conflated with the other (hence the distinct types). So when you warn of "code duplication and unnecessary overabstraction" I don't sympathize. Having similar functions defined for verified block as well as for this new possibly unverified block type is not code duplication, because they are fundamentally different functions over different domains. I should add however that I don't believe every function on our existing block type will have a counterpart for the serialized block type. There will be a few innocuous things like `getHeader`, `getBody`, etc. but surely all of the nontrivial work on a block should only be done on a verified block, so we wouldn't "duplicate" it anyway.

> I just can't imagine any solution that will:
> 1. Will provide us with `T0` that is essentially `T` but with other verification level
> 2. Will not propagate the notion of "verified"ness into record fields.

We don't necessarily want `T0` that is `T` but with other verification level. The "raw" type doesn't have to mimic the verified type in form and it almost certainly won't.

> It's an obvious thing that you would like to have, but I thought we (you and George) have already agreed on requirements (U1-U4) that are more flexible.

These requirements are consistent with what I described. The idea is to factor deserialization of a verified block through deserialization of a possibly invalid block. For U2 we deserialize the possibly invalid block, making no application-level checks for validity, and doing no possibly expensive transformations such as building a merkle tree. For U3 we do what's done for U2, but then run it through `verifyBlock` to get a verified block. For U4 we would read the unverified block type from the database and then compute whatever we need from it: a complete verification may not be necessary, maybe we only need a part of the verified block structure.

> Sorry, but it doesn't answer my question. What you state in this paragraph is that "it would be good to use U2.1" (this is already a requirement so there's no need to highlight that it's needed) and what i'm asking is to (try to) estimate whether U2.2 may be the same as U2.1. U2.2 may be just an easier requirement to implement yielding almost the same results.

If I understand correctly, U2.2 would include application-level verification, so I'm inclined to say no, it's not an acceptable replacement for U2.1.

RE: new solution 3
> This doesn't have the (bad) possibility to have different implementations of block for verified and unverified parameter.

There's the mismatch: having different implementations for verified versus unverified blocks does not strike me as a bad thing. Quite the opposite.

volhovM:
Now answering questions.

To @gromakovsky: 

> Points out it's easy to make a mistake if verification is manual (if we add extra field and forget about it)

Yes, it's the case, it's written in the document. Bad thing to have, sadly. On the bright side, once it's there, it's easy to validate what exactly is verified.

> Asks about verification propagation problem

The problem is open and it is also addressed in the document. I argue that if we separate types (reject Q2), then we will have to deal with verification propagation somehow. 

To @avieth:

>I don't believe every function on our existing block type will have a counterpart.

Yes, this makes sense in fact. I still think that using phantom types (solution 4) is the best way to granulate verification (instead of solution 1 you're supporting), so consider it as an option.

>We don't necessarily want T0 that is T but with other verification level. The "raw" type doesn't have to mimic the verified type in form and it almost certainly won't.

Well, I think it doesn't make sense. Let's consider our favorite `Block` example. AFAIU `RawBlock` should be "unverified" block and it should be convertible to `Block` easily -- so it should have all the `Block` fields except for those we can build on-fly while doing verification. This covers most of the types `T` verification granularity proposal will affect, like transactions, coins etc.

> The idea is to factor deserialization of a verified block through deserialization of a possibly invalid block. For U2 we deserialize the possibly invalid block, making no application-level checks for validity, and doing no possibly expensive transformations such as building a merkle tree. For U3 we do what's done for U2, but then run it through verifyBlock to get a verified block

> These requirements are consistent with what I described. <describes his vision>

Ok, good, it took me a while to understand this paragraph but i think that first sentence captures the whole point.

>If I understand correctly, U2.2 would include application-level verification, so I'm inclined to say no, it's not an acceptable replacement for U2.1.

No, that's not what i suggest. Let me elaborate on the difference.
 * __U2.1__ I have `ByteString` (representing `Block`, for example), but I will deserialize only part of it (like, first N bytes representing header) and check this part. This saves time as we don't deserialize most of the block (header is quite small). Profit.
 * __U2.2__ Whenever i have `ByteString` representing `T` we deserialize `T` completely to avoid missing possibly bad encoded data (which is possible if we choose U2.1), though we perform no checks by default (application level consistency is not enforced).

As far as I see, you vote for U2.2 which is, in my opinion, a better solution (both in terms of type safety and ease of implementation), though, maybe, is a little bit slower than U2.1 (which is hopefully alright).

>There's the mismatch: having different implementations for verified versus unverified blocks does not strike me as a bad thing. Quite the opposite.

I got your point, though it's pretty subjective. Performance gain may be significant, but maintaining this code duplication (again, I don't understand how can you implement `RawBlock` and `Block` without, say, fields duplication) seems to be somewhere on `[hard..impossible]` scale.

volhovM:
Honestly, I don't know what to do in this case. My solution priority list is `[3,4]` for now, though both seem terrible to me. Everything else is even more terrible.

volhovM:
@avieth any comments?

docs/proposals/serialization.md
```diff
@@ -0,0 +1,212 @@
...
+       D4.1, if it's not significantly slower.
+ * __(U3)__ Apply received block(s) to GState.
+     * D1-D5 (complete procedure) are needed.
+ * __(U4)__ Reading data (sometimes lots of it) from DB for different
+   node local actions (LRC, GState checks, API for Middleware -- all
+   of them require db traversing or LCA computations).
+     * D1,D4 needed (we trust local database, so we want to turn off
+       as much checks as it's possible).
+     * AFAIU @avieth assumes D2-D3 is also needed and we shouldn't
+       optimize here because it's too dangerous anyway.
```
avieth: Yes D2 is certainly needed. How can we get the datatype structure if we don't use its CBOR encoding? As for D3, I suppose we could drop it, but honestly I don't think it's worth it.

volhovm: Agree. I will change it to `D1, D2, D4 needed`, ok? I personally think it will be a noticeable optimization since, for example, `decodeMapSkel` does `reverse` and also O(n) order checks. It's a pretty expensive computation for large `n`s -- of course it doesn't change the asymptotic (still linear), but reduces the constant (~3 times faster?).

gromakovsky:
I think solution (3) is the best one here. It doesn't solve all the problems, but it allows us to do D5 without D4, which partially solves U2 and U4. It would be good to also be able to deserialize only part of data (D4.1) and avoid D3 (needed for U4). These things can be achieved separately. For the former I like `SAsT` approach. For the latter I think we can use `ReaderT` approach as in #1823 (not exactly the same code, but the same idea).
I think we should start implementing solution (3).

volhovm:
Thank you for the comment @gromakovsky. I agree with everything and, indeed, find (3) to be the most plausible and effective solution here. 

Though you say that "It doesn't solve all the problems" -- can you please point out which requirement does it not satisfy then?

gromakovsky:
As I said,
1. It doesn't allow us to deserialize only part of data (e. g. only header from a block). I know your point that this requirement is now so important if we disable verification (i. e. you assume that deserialization without verification is cheap), but it's still useful (because deserialization is not as cheap as absence of deserialization) and it should be proved by some benchmarks.
2. It doesn't allow us to skip canonicity checks (which I believe should be possible).

avieth:
> I think solution (3) is the best one here.

My objection to solution 3 remains. Quoting the doc: "This doesn't have the (bad) possibility to have different implementations of block for verified and unverified parameter." That's a deal-breaker for me. I *want* to have different implementations for the existing (verified) block and the unverified block, because they *really are* different. Sure it's nice to have one unified representation but it doesn't work here. Square peg, two elliptical holes.

Reiterating: it's useful to have the actual serialized forms of things first-class in our programs. That's to say, datatypes which mirror exactly the encodings of block bodies, headers, transactions, etc.. Typical practice in many Haskell applications is to leave the encoding implicit, a kind of under-the-hood implementation detail that you don't even have to observe (derive your JSON instances and get automagic servant http communcation for example) but that's no good in this setting because these encoded things are a core part of the system. The high-level Haskell representations need not be shared among different clients, but the encoded forms do, so we should have a way to deal with them in our Haskell libraries. It's these encoded forms which will have trivial and fast encode/decode.

gromakovsky:
> My objection to solution 3 remains. Quoting the doc: "This doesn't have the (bad) possibility to have different implementations of block for verified and unverified parameter." That's a deal-breaker for me. I want to have different implementations for the existing (verified) block and the unverified block, because they really are different. Sure it's nice to have one unified representation but it doesn't work here. Square peg, two elliptical holes.

AFAIU, you are talking about D4.1, i. e. ability to deserialize only part of the data. If not, please clarify with an example which exactly difference you want between verified and unverified block.
As I said, solution 3 doesn't make D4.1 possible and it's not the only thing we should do.
However, for D4.1 another solution was proposed (which combines with 3). If you want to deserialize only header, you can add `data HeaderAsBlock = HeaderAsBlock (Header verified) (RestOfBlock unverified)`. `Bi` instance will be a bit complex and will use the same encoding as for `Block`.

> Reiterating: it's useful to have the actual serialized forms of things first-class in our programs. That's to say, datatypes which mirror exactly the encodings of block bodies, headers, transactions, etc.. 

Is it about `ByteString` wrappers? If yes, then it's all about U1 and it's orthogonal to the rest of the problem. I think the most reasonable approach for it is to have `newtype Serialized a = Serialized ByteString`. And actually we already have such newtype in master. I don't think there is anything else to discuss about U1 (but maybe you have some objections). The main concern is U2 and U4.

avieth:
> AFAIU, you are talking about D4.1, i. e. ability to deserialize only part of the data. If not, please clarify with an example which exactly difference you want between verified and unverified block.

It's somewhat related to D4.1 but that item doesn't capture it all. Technically the entire datum would be deserialized: it would be produced from a `ByteString` but would not be a `ByteString`.  It would have more structure than a `ByteString`, but it would not be a verified `Block`, nor even of the same form as the existing `Block` type. Think of it like tokenization.

Forget about partial deserialization, I think it's a misleading goal. Instead, think only of full deserialization (i.e. deserialization, no qualifier), which is simple and straightforward, along with, shall we call it, *partial restructuring*: pure functions from less-structured types to more-structured types. The "raw block" type is less structured than the fully-verified `Block` type, so define a pure function to go from the former to the latter (`RawBlock -> Either InvalidBlock Block`). There's absolutely no need to mix that complexity into deserialization.

volhovm:
I've just updated the proposal, two things were changed:
1. I've removed U2.1/U2.2 separation, now U.2 is former U2.2. As a result, D4.1 was removed too. And also the part about `SAsT`.
2. U4 now contains "D1,D2,D4".

volhovm:
>I want to have different implementations for the existing (verified) block and the unverified block, because they really are different.

Sorry, but let's officially forbid want-reasoning. We're talking about serious tradeoffs (again: it's highly possible that implementing, eg, solution (1), will be hard/impossible to maintain). What requirement are you referring to? Solution (3) satisfies all of them, at least it seems so. Please prove me wrong. I don't see any real factor that would make me choose solution (1) over (3).

>It's somewhat related to D4.1 but that item doesn't capture it all. Technically the entire datum would be deserialized: it would be produced from a ByteString but would not be a ByteString.

This is confusing. Sounds a lot like partial deserialization (whatever it is called, if you leave some fields unserialized, it's partial deserialization). Can you formulate it as a requirement? Also, please, provide the related usecase example.

> less-structured types

So, like, `A` contains field `merkleA :: Merkle B`, while `RawA` contains field `listA :: [B]`, and `rawAtoA :: RawA -> A` builds merkle tree out of the list, right? Is it a usecase?

avieth:
> Sorry, but let's officially forbid want-reasoning.

What do you mean? Surely it's OK to state what I want. That's an important part of designing things.

> We're talking about serious tradeoffs (again: it's highly possible that implementing, eg, solution (1), will be hard/impossible to maintain). What requirement are you referring to? Solution (3) satisfies all of them, at least it seems so. Please prove me wrong. I don't see any real factor that would make me choose solution (1) over (3).

Well, I suppose I'm saying that solution 3 missing the mark, for reasons that do not necessarily appear in the document. And that's OK. Why should the document be completely exhaustive? I didn't write it, so naturally it doesn't contain everything that I want. This is normal; this is why we collaborate.

> This is confusing. Sounds a lot like partial deserialization (whatever it is called, if you leave some fields unserialized, it's partial deserialization). Can you formulate it as a requirement? Also, please, provide the related usecase example.
> ...
> So, like, A contains field `merkleA :: Merkle B`, while `RawA` contains field `listA :: [B]`, and `rawAtoA :: RawA -> A` builds merkle tree out of the list, right? Is it a usecase?

Exactly. The use case is bringing in a deserialized but unverified block from the wire or database and refining/verifying it only if necessary. It begins life as a less-structured thing (but more structured than a `ByteString`) and can be brought up to a more-structured and verified thing if need be.

volhovm:
>What do you mean?

I didn't want to sound rude or anything! Just saying that we need to have detailed argumentation of each opinion and requirement, because it's complex to agree on the solution.

>This is normal

Ok. The purpose of the document was to figure out the numerous requirements and details which felt contradictory to me unless formulated. So it wasn't obvious for me that something left aside. 

Can you please provide at least several examples of datatypes that should have altered raw versions?  `TxPayload` is one of them, is there anything else? Maybe we can live without this requirement for a while?

By the way, here's a list of the types that currently force internal invariants by not letting user to construct incorrect version of it. Just to roughly esmitate how many duplication we will have (also everything that is in-between them -- e.g. if `T` contains `RawCoin`, it will become `RawT`).
* `Coin`
* `LocalSlotIndex`
* `GenericBlockHeader`
* `GenericBlock`
* `VssCertificatesMap`
* `SystemTag`
* `UpdateProposal`
* `SlottingData`
* `Tx`
* `TxPayload`
* `DlgPayload`

georgeee:
> I suppose I'm saying that solution 3 missing the mark, for reasons that do not necessarily appear in the document. And that's OK. Why should the document be completely exhaustive? I didn't write it, so naturally it doesn't contain everything that I want.

@avieth, purpose of document was to handle discussion on proposals. So if it's not exhaustive, you're welcome to contribute.

@volhovm @avieth for me solution 3 with data families seems isomorphic to solution 1. Do we have strict reasons not to go that way (that said, where we need `MerkleTree A`/`[A]` we'll put data family and otherwise we won't have any more boilerplate as with raw types.

gromakovsky:
## Partial deserialization

> Technically the entire datum would be deserialized: it would be produced from a ByteString but would not be a ByteString.
> Forget about partial deserialization, I think it's a misleading goal.

I think that partial deserialization is still useful in some cases. E. g. when you receive a `VssCertificate`, it's a good idea to decode only `vcSigningKey` first, then check whether it's the one you expect and only then decode the rest.

But it's not a major use case, so I agree that we can forget about it for a while.

## Different representations

> Exactly. The use case is bringing in a deserialized but unverified block from the wire or database and refining/verifying it only if necessary. It begins life as a less-structured thing (but more structured than a ByteString) and can be brought up to a more-structured and verified thing if need be.

Ok, it's good that it was clarified. It wasn't obvious before. I think we are still looking at the problem from different perspectives.

## (3) vs (1) comparison

> for me solution 3 with data families seems isomorphic to solution 1

Let's consider `Tx` type to see the difference.

### Existing approach

Currently we have

```
data Tx = UnsafeTx
    { _txInputs     :: !(NonEmpty TxIn)  -- ^ Inputs of transaction.
    , _txOutputs    :: !(NonEmpty TxOut) -- ^ Outputs of transaction.
    , _txAttributes :: !TxAttributes     -- ^ Attributes of transaction
    } deriving (Eq, Ord, Generic, Show, Typeable)

type TxAttributes = Attributes ()

data TxIn
    = TxInUtxo
    { -- | Which transaction's output is used
      txInHash  :: !TxId
      -- | Index of the output in transaction's outputs
    , txInIndex :: !Word32
    }
    | TxInUnknown !Word8 !ByteString

data TxOut = TxOut
    { txOutAddress :: !Address
    , txOutValue   :: !Coin
    } deriving (Eq, Ord, Generic, Show, Typeable)

newtype Coin = Coin -- constructor isn't exported
    { getCoin :: Word64
    } deriving (Show, Ord, Eq, Generic, Hashable, Data, NFData)

```

There two invariants ensures by constructors:
1. All `Coin`s are in range `[0, maxCoinVal]`. Ensured by `mkCoin` constructor (which probably shouldn't use `error`, but it's another matter).
2. All `Coin`s are positive. Ensured by `mkTx` constructor.

Also there are invariants that `txInputs` and `txOutputs` are not empty, but they are enforced by `NonEmpty` type structure.

### Solution (1)

`TxAttributes` and `TxIn` don't have any invariants (even deep ones), they remain intact.
For `Tx`, `TxOut` and `Coin` we need `Raw` versions.

```
-- This type is not changed. Ideally we don't want to export the constructor, but unfortunately then we won't be able to use `RecordWildCards` which is probably too painful.
data Tx = UnsafeTx
    { _txInputs     :: !(NonEmpty TxIn)  -- ^ Inputs of transaction.
    , _txOutputs    :: !(NonEmpty TxOut) -- ^ Outputs of transaction.
    , _txAttributes :: !TxAttributes     -- ^ Attributes of transaction
    } deriving (Eq, Ord, Generic, Show, Typeable)

-- No instance Bi Tx

instance Buildable Tx where…

data RawTx = RawTx
    { _rtxInputs     :: !(NonEmpty TxIn)  -- ^ Inputs of transaction.
    , _rtxOutputs    :: !(NonEmpty RawTxOut) -- ^ Outputs of transaction.
    , _rtxAttributes :: !TxAttributes     -- ^ Attributes of transaction
    } deriving (Eq, Ord, Generic, Show, Typeable)

instance Buildable RawTx where…

verifyTx :: RawTx -> Either InvalidTx Tx
verifyTx RawTx {..} = do
    when hasOutputEqualToZero $ Left InvalidTx
    verifiedOutputs <- mapM (first InvalidOutput . verifyTxOut) txOutputs
    return UnsafeTx { _txInputs = _rtxInputs, _txOutputs = verifiedOutputs, _txAttributes = _rtxAttributes }

-- we can also have `Tx -> RawTx` and unsafe `RawTx -> Tx` (e. g. to read from DB) and other functions

data TxOut = TxOut
    { txOutAddress :: !Address
    , txOutValue   :: !Coin
    } deriving (Eq, Ord, Generic, Show, Typeable)

data RawTxOut = RawTxOut
    { txOutAddress :: !Address
    , txOutValue   :: !RawCoin
    } deriving (Eq, Ord, Generic, Show, Typeable)

-- verifyTxOut and other helpers should be provided

newtype Coin = Coin -- constructor isn't exported
    { getCoin :: Word64
    } deriving (Show, Ord, Eq, Generic, Hashable, Data, NFData)

newtype RawCoin = RawCoin -- constructor is exported
    { getCoin :: Word64
    } deriving (Show, Ord, Eq, Generic, Hashable, Data, NFData)

-- verifyCoin and other helpers should be provided
```

### Solution (3)

`TxAttributes` and `TxIn` don't have any invariants (even deep ones), they remain intact.
`Tx`, `TxOut` and `Coin` types are changed.

```
data Tx (v :: Verification) = Tx
    { _txInputs     :: !(NonEmpty TxIn)  -- ^ Inputs of transaction.
    , _txOutputs    :: !(NonEmpty (TxOut v)) -- ^ Outputs of transaction.
    , _txAttributes :: !TxAttributes     -- ^ Attributes of transaction
    }

-- TODO: is it ok to export `Tx` constructor? Likely no. We want it to be exported for unverified `Tx` and don't want for verified `Tx`. I don't see how to solve it with this approach

instance Buildable (Tx v) where…

instance Bi (Tx Unverified) where… -- only unverified

verifyTx :: Tx Unverified -> Either InvalidTx (Tx Verified)
verifyTx Tx {..} = do
    when hasOutputEqualToZero $ Left InvalidTx
    verifiedOutputs <- mapM (first InvalidOutput . verifyTxOut) txOutputs
    return Tx { _txInputs = _rtxInputs, _txOutputs = verifiedOutputs, _txAttributes = _rtxAttributes }

-- we can also have `Tx Verified -> Tx Unverified` and unsafe `Tx Unverified -> Tx Verified` (e. g. to read from DB) and other functions

data TxOut (v :: Verification) = TxOut
    { txOutAddress :: !Address
    , txOutValue   :: !(Coin v)
    } deriving (Eq, Ord, Generic, Show, Typeable)

-- It's ok to export `TxOut` constructor

newtype Coin (v :: Verification) = Coin
    { getCoin :: Word64
    } deriving (Show, Ord, Eq, Generic, Hashable, Data, NFData)

-- Same problem about exporting constructor as for `Tx`

instance Buildable (Coin v) where…

instance Bi (Coin Unverified) where… -- only unverified

```

### Differences

1. First of all, the main difference between these solutions and our current solution is that we can represent unverified value in code. That's what we are doing it for.
2. Solution (1) requires writing two type definitions instead of one. It's bad if we want verified and unverified types to have the same structure. It's especially bad for types with many records (e. g. `UpdateProposal`).
3. Solution (1) allows us to have different structure for verified and unverified types (for the cost described above). According to @avieth, it's a useful feature, but I guess it's useful not for all types.
4. For solution (3) it's not clear how to export only 1 constructor (which would be good).
5. There are functions/instances which make sense for both types (verified and unverified). Solution (1) requires us to write them twice. E. g. it's true for `Buildable` instances (I mentioned them in my example).
6. Some functions make sense only for verified or only unverified types. For such functions difference5 is not essential. Note that with solution (3) we can define a function once for both types or only for one type or 2 different implementations for two types.
7. Solution (3) depends on complex features, which is probably bad.


## Conclusion

I have a conclusion in mind, but I want to eat, will write it a bit later.

## P. S.

Sorry if I wrote obvious things, I wrote them mostly for myself, but you can also read them if you want.

volhovm:
> solution 3 with data families 

I undestood what you are talking about, but it's not mentioned everywhere. Just to clarify:
```
data Verified = V | UT
data family VerifiedExtra t

data instance VerifiedExtra (Block V) = MerkleTree
data instance VerifiedExtra (Block UT) = ListWrapper

data Block (v :: Verified) = Block
  { fieldCommon    :: Int
  , fieldDependent :: VerifiedExtra (Block v)
  }
```
This may work indeed.

volhovm:
@gromakovsky  thank you for your comment, it was helpful in fact. Regarding your example of applying (3):

> We want it to be exported for unverified `Tx` and don't want for verified `Tx`

1. Use constructor methods (`mkRawTx` and `verifyTx :: Tx UV -> Tx V`, also inverse of the latter).
2. Export both -- maybe rename to `UnsafeTx` as it's done now: constructing `Tx V` with `UnsafeTx` is indeed unsafe, constructing `Tx UV` with `UnsafeTx` is actually safe so it doesn't make lot of sense, but i think it's a minor semantic problem.

Now, regarding differences section: I agree with everything except for 3 (phantom types + data kinds), 4 (i have just explained it) and 7 (what is complex feature?).

>The only concern is: how do these approaches combine? If T has S inside and we want to use different approaches for T and S, is it possible? 

It's a known problem and it's not solved by (1) (unless you want to have exponential number of type declarations, e.g. `RawXwithRawYRawZ` and `RawXwithYRawZ` etc). I think that (3) can solve it using type-level verification negation (type families?). E.g. `data T (v :: Ver) = T { tSubtype :: S (Neg Ver) }`.

gromakovsky:
> (what is complex feature?)

```
+{-# LANGUAGE DataKinds      #-}
+{-# LANGUAGE GADTs          #-}
+{-# LANGUAGE KindSignatures #-}
```

I think all three extensions are advanced features. It doesn't mean that we shouldn't use this approach, but in general I'd prefer simpler solution if there are no other differences.

> It's a known problem and it's not solved by (1) (unless you want to have exponential number of type declarations, e.g. RawXwithRawYRawZ and RawXwithYRawZ etc).

Wait, where does exponential number of type declarations come from? AFAIU, we either want everything to be verified or want everything to be unverified. Maybe there are few cases when we want something different (having some part verified), but it shouldn't be exponential.
With solution (1) there will be `2n` types instead of `n`, with solution (3) there will be `n` types.
My concern is: suppose we pick approach (1) for `TxPayload` and approach (3) for `Tx`. Will we combine them nicely? Or we pick (1) for `TxPayload` (again), but (3) for `GenericBlock`.

Also, about your list with types with internal invariants:
1. You forgot at least `VssCertificate` (and btw, previously its constructor wasn't exported, then it was exported but wasn't renamed to `Unsafe`).
2. Note that almost all types have internal invariants if we consider the whole structure (i. e. including all types stored inside that type).

> I agree with everything except for 3 (phantom types + data kinds)

Yes, I see how that extension of (3) allows us to have different representations. It looks good when we want have different but similar representation and it looks weird if we want representations to be completely different (e. g. have verified block with all stuff we have now and unverified block with `Header` and `SomeOtherStuff`).

georgeee:
We had a conversion with @gromakovsky @volhovm . This is our conclusion:

We should use solution 3 (phantom types).

For current codebase it's more than sufficient:
  - list of transactions is perhaps the only place where we want to distinguish on types, but yet `MerkleTree Tx` can be replaced by simple list, b/c MerkleTree is used only on validation step and then may be thrown away
  - some checks like checks being done for `Coin` are very cheap, so it worth leaving them at their current place
  - expensive checks can be easily opted out by providing this validate function which will transform `T a` to `T b`, where `a, b` are phantom types

If in future we would decide to have different structure for unverified type, we will still be able to implement this with data family made on phantom type, see example at the end of [Mikhail's comment](https://github.com/input-output-hk/cardano-sl/pull/1903#issuecomment-346824975).

georgeee:
@avieth waiting for your feedback!

avieth:
> @avieth waiting for your feedback!

Solution 3 remains less-than-ideal for me because of the proposed advantage given in the document:

> This doesn't have the (bad) possibility to have different implementations of block for verified and unverified parameter.

I want to have structurally different datatypes for verified and unverified. The idea is that the unverified datatype is sometimes bigger than (never smaller than) the verified one, so we don't get an isomorphism of the two, only an injection from verified to unverified (analogous to the injection from unverified to `ByteString` given by the canonical encoder). 

Frankly I believe a phantom type data kinds type family approach is *at least* as complex and costly as introducing some new types for the unverified things. I also don't see the use for having a composite with some things verified and some things unverified. Just construct the particular verified component from the unverified composite and use that directly. E.g. if you want a verified transaction payload from an unverified block, you don't have to make a new block type where only the transaction payload is verified, you can just use `verifyTransactionPayload :: Pos.Binary.Txp.TxPayload -> Either Invalid Pos.Txp.TxPayload` to get the thing you need.

The principal point I want to highlight is that we need a first-class representation of types as they appear on the wire. Invalid blocks / transactions / whatever are a normal thing that we have to deal with so we must not pretend as though they don't exist or that they don't make any sense. They do! Somebody can send us an invalid yet correctly-encoded block and we must be able to recognize exactly that: decoding passed, verification did not, but here's the decoded thing anyway.

Defining structurally different types where needed is the simplest solution and it seems to me the only complaints against it are worries about boilerplate due to partial validation (having to define a new type for every intermediate partial verification) but I don't think this is a legitimate issue.

georgeee:
@avieth But thing is that we have no structural difference between verified and unverified types for all data types we've got at the moment.

So having Raw data type hierarchy wil effectively copy existing hierarchy. Just for sake of marking data type "unverified".

> Frankly I believe a phantom type data kinds type family approach is at least as complex and costly as introducing some new types for the unverified things.

It would be if nearly all types in raw hierarchy would have to be redefined. But we see an opposite case: that most part of structure would be preserved and only some types in these hierarchy changed.

E.g. it's useful to keep crypto data types as `ByteString` and use constructors from` cryptonite` only on verification phase (because these constructors are comparatively expensive). But then we have to amend types of several leafs of the hierarchy, yet for all non-leaf nodes (and also most leaf nodes) type/structure won't change. It would be insane to replicate whole structure with `Raw` hierarchy.

Data families address it perfectly.

gromakovsky:
> I want to have structurally different datatypes for verified and unverified. The idea is that the unverified datatype is sometimes bigger than (never smaller than) the verified one, so we don't get an isomorphism of the two, only an injection from verified to unverified (analogous to the injection from unverified to ByteString given by the canonical encoder).

The only example where it's desirable to have different representations of verified and unverified value was `TxPayload` where you want unverified value to have `[Tx]` and verified one to have `MerkleTree Tx`. It's indeed a good point and for this purpose we can't use solution (3), only solution (1). However:
1. We serialize transactions as a list anyway and the only place where we use `MerkleTree` not as just a list is the place where we check that block header corresponds to block. But it should happen only once for a block (when convert unverified value to a verified one), so we can just change `MerkleTree Tx` to `[Tx]`. There are currently no plans to use `MerkleTree` in other places, so we can simplify this code.
2. Even if at some point we want to use `MerkleTree` for other purposes and decide to have `TxPayload` which we have now (with `MerkleTree Tx`) or if you give another example of a type where different representations are needed, we can use approach (1) for `TxPayload` and approach (3) for other types. It's not necessary to use the same approach for all types. Different approaches may be more suitable for different types.

> The principal point I want to highlight is that we need a first-class representation of types as they appear on the wire. Invalid blocks / transactions / whatever are a normal thing that we have to deal with so we must not pretend as though they don't exist or that they don't make any sense. They do! Somebody can send us an invalid yet correctly-encoded block and we must be able to recognize exactly that: decoding passed, verification did not, but here's the decoded thing anyway.

Ok! `TxPayload Unverified` is the type which represents (possibly) invalid transaction. And `TxPayload Verified` is a type which represents (to some extent) valid transaction.

volhovm:
George and Ivan already pointed out everything, so only two small comments:

> if you want a verified transaction payload from an unverified block, you don't have to make a new block type where only the transaction payload is verified, you can just use verifyTransactionPayload :: Pos.Binary.Txp.TxPayload -> Either Invalid Pos.Txp.TxPayload to get the thing you need.

In approach (3) you just take the `TxPayload Unver` field from `Block Unver` and verify it. This is as simple.

>we need a first-class representation of types as they appear on the wire

In approach (3) `X Unver` is exactly the representation you're looking for.

avieth:
> But thing is that we have no structural difference between verified and unverified types for all data types we've got at the moment.

Ok, but I expect there will be structural differences. In the diffusion layer I want a structurally different block: not `Block Unverified`, but something like a header, parent hash, and blob for the body.

> Ok! TxPayload Unverified is the type which represents (possibly) invalid transaction. And TxPayload Verified is a type which represents (to some extent) valid transaction.

But with the phantom type solution they must have the same representation. 

> In approach (3) you just take the TxPayload Unver field from Block Unver and verify it. This is as simple.
> ...
> In approach (3) X Unver is exactly the representation you're looking for.

Right but `X Unver` and `X Ver` must have the same representation, and that's what I'm complaining about. So we go with this data family approach, but then these are no longer phantom types (the representation depends upon them), and we just have a roundabout way of defining 2 different types, so why not go direct and define two different types *without* this extra machinery?

gromakovsky:
> Ok, but I expect there will be structural differences. In the diffusion layer I want a structurally different block: not Block Unverified, but something like a header, parent hash, and blob for the body.

Ok, if you want to have different representations of `Block Unverified` and `Block Verified`, then we can use approach (1) for blocks.

The thing is: you mentioned only `TxPayload` as the type for which you want verified and unverified versions to have different representations. And the reason was that constructing `MerkleTree` is expensive, so you don't want to have `MerkleTree` in `TxPayload`. But we are going to drop `MerkleTree`, which means we don't need to have different representations.

If there are other types for which we need different representations, then no problem we can use approach (1) for them (as soon as you convince us that different representations are necessary), but for other types I find it better to use approach (3). For example, `Tx` type.

volhovm:
>I suppose it's a confusing term, partial deserialization.

It's intuitively clear that "binary blob" is a "serialized" value. We either have ADT-like datatype or we have some `ByteString` fields (or just one, newtype-like, field) that represent CBOR-encoded value inside. 

> and blob for the body.

What does "blob" mean here then? 🤔 

> fully deserialize a less structured type.

If you had a string `bs` and you turned it into some datatype `B` which has field `leftover :: ByteString`, and `leftover` is basically substring of `bs`, then `B` is "partially deserialized" IMO. I think that it's valid to assume that what i say is more or less what everybody would expect from this term to express. We wanted to separate verification (which can be done several times, converting less structured things to more structured)  and deserialization (which is done only once (and by saying that, i mean we run cbor decoder once only), and by default produces least-structured datatype).

> we'll need to use some different representations

gromakovsky:
> Actually, why? We've already agreed that we don't want to have partial deserialization. That's what you've said.

I've reread my big post, that's what I wrote there:
> But it's not a major use case, so I agree that we can forget about it for a while.

When I wrote it I was thinking about partial deserialization for stuff like `Commitment`, which is processed rarely. Having ability to partially deserialize can speed-up `Commitment` processing, but it's quite negligible, considering how often we need to do it.
I also was thinking about partial deserialization for blocks, where we deserialize only headers. But for some reason I decided that it wasn't useful and didn't even write about it :/
Now I am trying to recall my reasoning, but I can't do it :(
And now I think that representing main block as `(header, blob)` makes sense and is useful for diffusion layer, because there we want to verify header, relay block and only then process the rest of the block.
I wish I could recall my thoughts from Nov 24 :/
Probably it was also discussed during after call about all this stuff, but I don't remember what we discussed about it then.

> It's reflected in the doc :\

To be more precise, it's only reflected in the history of the doc. It's removed from the doc currently.

P. S. Also side note: perhaps we should represent blobs as CBOR items, not `ByteString`, because our `Bi` class doesn't allow putting arbitrary `ByteString` «as-is».

gromakovsky:
@volhovm anyway, I don't see a big problem with representing unverified block as `(header, blob)`. And the advantage is that we can do one phase from diffusion layer faster (because we don't need to perform any decoding for body).
We still can convert this representation to `Block Verified` without unnecessary overhead (including DB case, where we are sure that unverified block is valid) and then use it in all the code we have now.

volhovm:
@gromakovsky 
>it's only reflected in the history of the doc

Yes: there was a U2.1-U2.2 tradeoff and we already decided that U2.2 is what we'll go with. https://github.com/input-output-hk/cardano-sl/pull/1903/commits/7a777063339435adf787646f6fed67d5bbb04d63#diff-450f574a311c67771c3d6d74099056b4L36

>perhaps we should represent blobs as CBOR items, not ByteString, because our Bi class doesn't allow putting arbitrary ByteString «as-is».

We can't do it w/o deserializing the value. See YT comments on CSL-1869.

gromakovsky:
> We can't do it w/o deserializing the value. See YT comments on CSL-1869.

It's a pity, probably we should change interface of `Pos.Binary.Class` or even `cborg` then :(

georgeee:
So to summarize:

1. We're happy with phantom type + data family for data structures

2. There's an open question how to implement a case for partial validation:

We want to read and validate header, but omit most checks (except for hash) for body, don't deserialize body, but read it a sequence of bytes/tokens

How we reflect it with type? Special type to store some CBOR structure?
How we bind it with cborg/time-warp?

avieth:
> So to summarize:
> 
> 1. We're happy with phantom type + data family for data structures

In my view it's *vastly* over-complicated but I suppose if you keep it localized and we don't have to deal with it at the diffusion layer, then it's fine.

> 2. There's an open question how to implement a case for partial validation:
> 
> We want to read and validate header, but omit most checks (except for hash) for body, don't deserialize body, but read it a sequence of bytes/tokens
> 
> How we reflect it with type? Special type to store some CBOR structure?
> How we bind it with cborg/time-warp?

I'll tell you how I plan to do it in the diffusion layer: define a *new, structurally different* type for the block which shows just the things we need, and includes a blob (byte string) for the body. I'll define the encoder to just read in the things we need, and leave all else undecoded (as byte strings). I'll give this type a new time-warp `MessageCode` and make a new conversation handler for it, and let the inspec/outspec system do its thing (assuming it works, I don't believe it's been put to the test).

No change to time-warp required, no change to CBOR required.

volhovm:
> I'll give this type a new time-warp MessageCode and make a new conversation handler for it

I'm fine with it as long as you keep compatibility with older clients. If diffusion layer wants to have another representation of `Block`, it's totally fine, unless we have to deal a lot with it in user logic. 

>no change to CBOR required.

I doubt it. We were thinking about how to implement decoding something into `(Header,ByteStringBinaryBlob)` and it's actually more complex than we thought.

But that's fine in general. We should be moving forward even if it costs us some mistakes. Maybe our solution is not optimal as well. We'll see. :)

PR #1912, opened 2017-11-15
CSL-1800 Move not csl-specific utility modules into a separate package

gromakovsky:
Now we have `cardano-sl-util` package, which contains various utility functions which are not cardano-sl-specific. There are two reasons why it has been done:
1. It allows us to move `Pos.Crypto` into a separate package (because `Pos.Crypto` depends on some utility modules).
2. It makes `cardano-sl-core` smaller, this package has more than a hundred modules, so we want to move something from it.

Previously we had `Pos.Util.Util` module which was very big and contained groups of utilities. I put some big groups into separate modules (only those which logically formed a separate group and looked good to be separated). For instance, orphan instances for 3rdparty types/classes are now in `Pos.Util.Orphans`.


gromakovsky:
It's ready for review.

PR #1913, opened 2017-11-15
[CSL-1856] Remove light delegation

volhovM:
I've removed light delegation logic code, though leaving light block signatures (in order to maintain backwards compatibility).

I also ran weeder on our project and thus removed lots of unused dependencies. I recommend reviewing this PR per-commit (there are only two of them).


auxx/src/Command/Proc.hs
```diff
@@ -321,32 +322,6 @@ createCommandProcs hasAuxxMode printAction mSendActions = rights . fix $ \comman
...
-                dataFlow
-                    "pskLight"
-                    (immediateConcurrentConversations sendActions ccPeers)
-                    (MsgTransaction OriginSender) psk
-                logInfo "Sent lightweight cert"
-        return ValueUnit
-    , cpHelp = ""
-    },
-
     let name = "delegate-heavy" in
```
int-index: Perhaps rename `delegate-heavy` to simply `delegate` now?

volhovM: I decided to leave it because doing otherwise would break compatibility with some FAQs. And, anyway, light delegation will be _reworked_ afaiu so no need to call heavy delegation "just delegation".

core/cardano-sl-core.cabal
```diff
@@ -148,10 +148,9 @@ library
 
   build-depends:       QuickCheck
                      , aeson
-                     , ansi-terminal
+                     , ansi-terminal
```
gromakovsky: 😒

volhovM: My bad, I was using vim to edit those configs, I only have trailing space hook in emacs.

volhovM:
Got it.

PR #1914, opened 2017-11-16
[CSL-1881] Add tmux version check in script.

denisshevchenko:


scripts/launch/demo.sh
```diff
@@ -7,6 +7,24 @@ if ! [ -n "$TMUX" ]; then
   exit 1
 fi
 
+# Make sure we're using proper version of tmux.
+# IMPORTANT: this check works on Linux only!
```
gromakovsky: It should also work on MacOS, we have MacOS users of this script.

denisshevchenko:
@gromakovsky Fixed.

PR #1917, opened 2017-11-16
Allow to disable TLS with --no-tls

Martoon-00:


wallet-new/server/Cardano/Wallet/Server/CLI.hs
```diff
@@ -1,22 +1,23 @@
+{-# LANGUAGE ApplicativeDo #-}
```
adinapoli-iohk: Sorry to be a pain in the ass, but any chance we can *not* use `ApplicativeDo`? It feels a little weird that we are using a package like `optparse-applicative` abusing the `do` notation. I know we are doing this virtually everywhere in cardano, but I thought that for new code we could try something different, if possible 😉

wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -79,11 +76,11 @@ walletBackend :: (HasConfigurations, HasCompileInfo)
               -> Plugin WalletWebMode
 walletBackend WalletBackendParams {..} =
     first one $ worker walletServerOuts $ \sendActions -> do
-      walletServeImpl
+      trace @Text ("Wallet tls: " <> show walletTLSParams) $ walletServeImpl
```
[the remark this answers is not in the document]
Martoon-00: Oh right, forgot about it

wallet-new/server/Cardano/Wallet/Server/CLI.hs
Martoon-00: When I've read this ^, I realized what was wrong :))
Arguments were passed in wrong order to `TlsParams` constructor

Martoon-00:
Fixed

PR #1918, opened 2017-11-16
Merge "alfredo/wallet-new" feature branch into master

adinapoli-iohk:
This slightly beefy PR brings into `master` a new project called `wallet-new`, which is the home for the V1 redesign.

It also moves `Pos.Util.BackupPhrase` from `cardano-sl` into `cardano-sl-wallet`, which is a useful step towards having `wallet-new` depend from less dependencies (which is something we still cannot enforce now, but we will try in the future).

cc @gromakovsky @georgeee


lib/src/Pos/Client/CLI/NodeOptions.hs
```diff
@@ -13,33 +13,31 @@ module Pos.Client.CLI.NodeOptions
...
-import           Pos.Util.BackupPhrase (BackupPhrase, backupPhraseWordsNum)
-import           Pos.Util.CompileInfo (CompileTimeInfo (..), HasCompileInfo, compileInfo)
-import           Pos.Util.TimeWarp (NetworkAddress)
+import           Pos.Client.CLI.Options       (CommonArgs (..), commonArgsParser, optionalJSONPath)
+import           Pos.HealthCheck.Route53      (route53HealthCheckOption)
+import           Pos.Network.CLI              (NetworkConfigOpts, networkConfigOption)
+import           Pos.Statistics               (EkgParams, StatsdParams, ekgParamsOption,
+                                               statsdParamsOption)
+import           Pos.Util.CompileInfo         (CompileTimeInfo (..), HasCompileInfo, compileInfo)
+import           Pos.Util.TimeWarp            (NetworkAddress)
```
gromakovsky: Can you reformat it back with new stylish-haskell config?

wallet/src/Pos/Wallet/Web/Server/Runner.hs
```diff
@@ -9,35 +9,37 @@
...
+import           Pos.Wallet.Web.Methods         (addInitialRichAccount)
+import           Pos.Wallet.Web.Mode            (WalletWebMode, WalletWebModeContext (..),
+                                                 WalletWebModeContextTag)
 import           Pos.Wallet.Web.Server.Launcher (walletApplication, walletServeImpl, walletServer)
-import           Pos.Wallet.Web.Sockets (ConnectionsVar)
-import           Pos.Wallet.Web.State (WalletState)
-import           Pos.Web (TlsParams)
+import           Pos.Wallet.Web.Sockets         (ConnectionsVar)
+import           Pos.Wallet.Web.State           (WalletState)
+import           Pos.Web                        (TlsParams)
```
gromakovsky: Wrong formatting here as well.

wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -16,13 +13,13 @@ import           Universum
...
-                             unsafeAbstractHash)
-import           Pos.Util.Mnemonics (fromMnemonic, toMnemonic)
-import           Test.QuickCheck (Arbitrary (..), genericShrink)
+import           Crypto.Hash         (Blake2b_256)
+import           Pos.Binary          (Bi (..), serialize')
+import           Pos.Crypto          (AbstractHash, EncryptedSecretKey, PassPhrase, SecretKey,
+                                      VssKeyPair, deterministicKeyGen, deterministicVssKeyGen,
+                                      safeDeterministicKeyGen, unsafeAbstractHash)
+import           Pos.Util.Mnemonics  (fromMnemonic, toMnemonic)
+import           Test.QuickCheck     (Arbitrary (..), elements, genericShrink, vectorOf)
```
gromakovsky: Wrong formatting here as well.

wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -0,0 +1,113 @@
...
+                                 wwmcSendActions)
+import           Pos.Wallet.Web.Account (findKey, myRootAddresses)
+import           Pos.Wallet.Web.Methods.Restore (addInitialRichAccount)
+import           Pos.Wallet.Web.Pending.Worker (startPendingTxsResubmitter)
+import qualified Pos.Wallet.Web.Server.Runner as V0
+import           Pos.Wallet.Web.Sockets (getWalletWebSockets, launchNotifier, upgradeApplicationWS)
+import           Servant (Handler, Server, serve)
+import           System.Wlog (logInfo, modifyLoggerName)
+
+import qualified Data.ByteString.Char8 as BS8
```
volhovM: Wrong grouping.

wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -0,0 +1,113 @@
...
+      walletServeImpl
+        (getApplication sendActions)
+        walletAddress
+        -- Disable TLS if in debug mode.
+        (if (isDebugMode walletRunMode) then Nothing else walletTLSParams)
+  where
+    -- Gets the Wai `Application` to run.
+    getApplication :: SendActions WalletWebMode -> WalletWebMode Application
+    getApplication sendActions = do
+      logInfo "DAEDALUS has STARTED!"
```
volhovM: Wrong indentation.

wallet-new/server/Main.hs
```diff
@@ -0,0 +1,137 @@
...
+module Main where
+
+import           Universum
+
+import           Data.Aeson.Encode.Pretty      (encodePretty)
+import qualified Data.ByteString.Lazy.Char8    as BL8
+import           Data.Maybe                    (fromJust)
+import           Formatting                    (sformat, shown, (%))
+import           Mockable                      (Production (..), currentTime, runProduction)
+import           Pos.Communication             (ActionSpec (..))
```
volhovM: Wrong grouping.

wallet-new/src/Cardano/Wallet/API/Types.hs
```diff
@@ -0,0 +1,134 @@
+{-- Types shared between different API versions. --}
+{-# LANGUAGE ConstraintKinds       #-}
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE DefaultSignatures     #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE FlexibleInstances     #-}
```
volhovM: We usually use default extensions in cabal file for these things.

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
```diff
@@ -0,0 +1,115 @@
...
+--
+instance Migrate V0.CWalletAssurance V1.AssuranceLevel where
+    eitherMigrate V0.CWAStrict = pure V1.StrictAssurance
+    eitherMigrate V0.CWANormal = pure V1.NormalAssurance
+
+instance Migrate V1.AssuranceLevel V0.CWalletAssurance where
+    eitherMigrate V1.StrictAssurance = pure V0.CWAStrict
+    eitherMigrate V1.NormalAssurance = pure V0.CWANormal
+
+--
```
volhovM: Is it a ... separator? Or you wanted to write a comment? :thinking:

wallet-new/stack.yaml
```diff
@@ -0,0 +1,24 @@
+resolver: lts-9.1
```
volhovM: Woah, why is it a separate project instead of a subproject? I guess it's totally wrong. 
Nobody wants to deal with two `defaut.nix`. Also did you check that `pkgs/generate.sh` works correctly?

wallet-new/stack.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - acid-state-0.14.3
+  - log-warper-1.3.3
+  - time-units-1.0.0
+  - fmt-0.5.0.0
+  - servant-0.12
+  - servant-server-0.12
+  - servant-client-0.12
+
+nix:
+  shell-file: ../shell.nix
```
gromakovsky: Hm, I think it's better to have only one `stack.yaml` per repository.

adinapoli-iohk:
@volhovM Please let's be pragmatic with these changes. I'm not going to amend 6k lines of code for a prototype, but it *will* get polished when the time will come. I just don't want a feature branch to span several months.

wallet-new/stack.yaml
adinapoli-iohk: Right, I agree. This is a relic from the prototype.

adinapoli-iohk:
> There are few scripts which should be updated when a new package is added:
> • scripts/build/cardano-sl.sh
> • scripts/haskell/lint.sh
> • scripts/haskell/update-cabal-versions.sh 

@gromakovsky I have addresses all your feedback apart from amending the last 2 scripts. The reason being this is a prototype, and we don't want our CI to fail due to it, but chances are we still want the CI to at least build it, which I hope my amendment to the first script will ensure.

Things like making the project compliant to the style guides & address all the linting suff (as @volhovM rightly pointed out) can be addressed at a later stage but our priority here is to prototype how the API would look like, and I'm afraid we cannot afford putting too much time in adding the bells & whistles _just yet_.

I will create tickets in our backlog to not forget though, I promise! 😉

wallet/test/Test/Pos/Types/Identity/BinarySpec.hs
```diff
@@ -0,0 +1,19 @@
...
+       ( spec
+       ) where
+
+import           Test.Hspec            (Spec, describe)
+import           Universum
+
+import           Pos.Util.BackupPhrase (BackupPhrase)
+
+import           Test.Pos.Helpers      (binaryTest)
+import           Test.Pos.Util         (withDefConfiguration, withDefInfraConfiguration)
```
gromakovsky: And (apparently last one) another file with wrong import formatting (I am permissive regarding everything in wallet-new, but not in other places).

scripts/build/cardano-sl.sh
```diff
@@ -192,6 +192,9 @@ if [[ $clean == true ]]; then
   echo "Cleaning cardano-sl-wallet"
   stack clean cardano-sl-wallet
 
+  echo "Cleaning cardano-sl-wallet-new"
+  stack clean cardano-sl-wallet-new
```
gromakovsky: One more change is needed in this script. There are lines like these:
```
  elif [[ $var == "wallet" ]]; then
    spec_prj="wallet"
  elif [[ $var == "explorer" ]]; then
    spec_prj="explorer"
```

You should add `wallet_new` case there.
P. S. Yes, these scripts are stupid and a bit tedious to maintain, because it's bash and we don't have bash experts unfortunately :(

adinapoli-iohk: Ok, makes sense -- maybe we want to build cardano with just `wallet-new` or similar. Leave it with me.

adinapoli-iohk:
Ok lieutenant @gromakovsky , your wish, my command 😉 

It should be all done now. I recommend for once merging this PR with a squashed commit 😀

PR #1923, opened 2017-11-17
[CSL-1884] Fix benchmarks so they work for in-memory benchmarking.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-1884

The benchmarking now works with everything in-memory, so it's not "real" since we don't have any disk access slowdowns. We have everything available in memory.

When running the memory/space benchmark I got:
```
Case                               Allocated      GCs
getBlocksTotal 100 blocks        622,766,056    1,190
getBlocksTotal 1000 blocks       659,519,408    1,254
getBlocksTotal 10000 blocks    1,020,147,528    1,877
getBlocksPage 100 blocks         798,029,096    1,496
getBlocksPage 1000 blocks      3,671,720,216    6,720
getBlocksPage 10000 blocks   160,749,876,408  303,805
```

When running time benchmark I got:
```
benchmarking getBlocksTotal 100 blocks    ~ 40 ms
benchmarking getBlocksTotal 1000 blocks   ~ 40 ms
benchmarking getBlocksTotal 10000 blocks  ~ 40 ms
benchmarking getBlocksPage 100 blocks     ~ 40 ms
benchmarking getBlocksPage 1000 blocks    ~ 40 ms
benchmarking getBlocksPage 10000 blocks   ~ 40 ms
```
An average of 40 ms for the execution of the two functions `getBlocksTotal` and `getBlocksPage`.
It kind of makes sense, since we have everything in memory and we are using paging (last 10 blocks).
The memory benchmark seems to make sense since we have to fill out the `CBlockEntries` data and it takes quite a bit of data in it.


explorer/bench/Bench/Pos/Explorer/ServerBench.hs
```diff
@@ -3,222 +3,124 @@ module Bench.Pos.Explorer.ServerBench
...
-        mode = def {
-            emmGetTipBlock         = pure $ genTestArgs ^. gtaTipBlock,
-            emmGetPageBlocks       = \_ -> pure $ Just $ genTestArgs ^. gtaBlockHeaderHashes,
-            emmGetBlundFromHH      = \_ -> pure $ Just $ genTestArgs ^. gtaBlundsFromHeaderHashes,
-            emmGetSlotStart        = \_ -> pure $ Just $ genTestArgs ^. gtaSlotStart,
-            emmGetLeadersFromEpoch = \_ -> pure $ Just $ genTestArgs ^. gtaSlotLeaders
-        }
+getBlocksPageBench (testParams, extraContext) =
+    withDefConfigurations $
+        runExplorerTestMode testParams extraContext $ getBlocksPage Nothing (Just 10)
```
sectore: Could we use `Just defaultPageSize` here?

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
```diff
@@ -3,222 +3,124 @@ module Bench.Pos.Explorer.ServerBench
...
 
-    runExplorerTestMode testParams extraContext $ getBlocksPage pageNumber (Just 10)
+    pure (testParams, extraContext)
   where
-    -- The page number we send to the function
-    pageNumber :: Maybe Word
-    pageNumber = genTestArgs ^. gtaPageNumber
+    -- | Generated test parameters.
+    testParamsGen :: IO ExplorerTestParams
+    testParamsGen = generate arbitrary
```
sectore: Why not using `generate arbitrary` directly in `line 56` without creating an extra `testParamsGen` function? This `where` clause could be removed then.

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
```diff
@@ -3,222 +3,124 @@ module Bench.Pos.Explorer.ServerBench
...
+        runExplorerTestMode testParams extraContext $ getBlocksPage Nothing (Just 10)
+
+-- | This is used to generate the test environment. We don't do this while benchmarking
+-- the functions since that would include the time/memory required for the generation of the
+-- mock blockchain (test environment), and we don't want to include that in our benchmarks.
+generateTestEnv
+    :: BlockNumber
+    -> SlotsPerEpoch
+    -> IO BenchmarkTestParams
+generateTestEnv totalBlocksNumber slotsPerEpoch = do
```
sectore: I would rename this function to `generateTestParams`

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
ksaric: It's on purpose. When reading, you can easily see what instance is being generated.

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
ksaric: Fixed.

explorer/bench/Bench/Pos/Explorer/ServerBench.hs
sectore: That's what my editor can show me and it done it well :) Anyway, it's up to you to change it or not ;)

sectore: P.S. Screenshot - it's Atom in this case :arrow_down: .
type

ksaric: Lucky you. 
I can't get Atom to work :sob:

PR #1925, opened 2017-11-17
CSL-1895 Fix warnings in wallet-new, make it build, integrate with CI

gromakovsky:


gromakovsky:
In this particular PR I did everything manually. I know that @volhovM used `weeder` recently in another PR. Regarding `importify`: I tried to use it with `cardano-sl`, but unfortunately it's not ready for such big projects yet.

PR #1926, opened 2017-11-17
[CSL-1812] Error reporting in auxx lang 

hithroc:
Type erros and argument errors are now pretty printed like this:
```
auxx> propose-update x:0 false
Invalid arguments for 'propose-update':
  Missing keys: block-version,bvm,software-version,vote-all
  Irrelevant keys: x
  Following type errors occured:
    Couldn't match expected type Int with actual value ValueBool False!
```

Parse errors are now also pretty printed and they look like this
```
auxx> this is going to be a @#$%^&* parse error
Parse error at 1:23-1:30
Unexpected TokenUnknown (UnknownChar '@'), expected procedure call w/o arguments, or literal, or parenthesized expression, or atom, or argument
this is going to be a @#$%^&* parse error
                      ^^^^^^^
```

Possible improvements: pretty print tokens in parse errors, and pretty print values in type errors.


auxx/src/Lang/DisplayError.hs
```diff
@@ -0,0 +1,128 @@
...
+        "Command" <+> (squotes . highlight . nameToDoc) name
+    <+> "is unavailable. Try" <+> (squotes . highlight) "help"
+ppEvalError (InvalidArguments name procError) =
+        "Invalid arguments for" <+> (squotes . highlight . nameToDoc) name <> ":"
+    <$> indent 2 (ppProcError procError)
+
+
+renderLine :: Int -> Int -> Text -> Doc
+renderLine start end str = (text $ unpack str)
+  <$> highlight (indent start . hcat . replicate (end - start) $ char '^')
```
int-index: Since it's an error, maybe use red rather than yellow here?

auxx/src/Lang/DisplayError.hs
```diff
@@ -0,0 +1,128 @@
...
+
+renderFullLine :: Text -> Doc
+renderFullLine str = renderLine 0 (length str) str
+
+ppParseError :: ParseError -> Doc
+ppParseError (ParseError str (Report {..})) =
+      "Parse error at" <+> text (show span)
+  <$> "Unexpected" <+> text unconsumedDesc <> ", expected"
+  <+> hcat (punctuate (text ", or ") $ map (text . unpack) expected)
```
int-index: This is the third time I see `text . unpack` here. Try doing something like this:

```
import qualified Text.PrettyPrint.ANSI.Leijen

text = Text.PrettyPrint.ANSI.Leijen.text . Universum.toText
```

auxx/src/Lang/DisplayError.hs
```diff
@@ -0,0 +1,128 @@
...
+  <$> renderLines
+  where
+    unconsumedDesc = maybe "end of input" show . head . fmap snd $ unconsumed
+    strLines = NE.nonEmpty $ take spanLines . drop (spanLineStart - 1) $ lines str
+    renderLines = case strLines of
+        Nothing -> error "Wtf"
```
int-index: please include the position, `error "ppParseError/renderLines: Wtf"`

auxx/src/Lang/DisplayError.hs
```diff
@@ -0,0 +1,128 @@
...
+renderFullLine str = renderLine 0 (length str) str
+
+ppParseError :: ParseError -> Doc
+ppParseError (ParseError str (Report {..})) =
+      "Parse error at" <+> text (show span)
+  <$> "Unexpected" <+> text unconsumedDesc <> ", expected"
+  <+> hcat (punctuate (text ", or ") $ map (text . unpack) expected)
+  <$> renderLines
+  where
+    unconsumedDesc = maybe "end of input" show . head . fmap snd $ unconsumed
```
int-index: Is using `head` safe here? If yes, explain why (in a comment). If not, don't use it.

auxx/src/Lang/DisplayError.hs
int-index: Also add a comment that explains why this case is impossible

auxx/src/Lang/DisplayError.hs
```diff
@@ -0,0 +1,128 @@
...
+    spanColumnStart = fromIntegral . toNat . locColumn . spanStart $ span
+    spanColumnEnd   = fromIntegral . toNat . locColumn . spanEnd   $ span
+    spanLineStart   = fromIntegral . toNat . locLine   . spanStart $ span
+    spanLineEnd     = fromIntegral . toNat . locLine   . spanEnd   $ span
+    spanLines = spanLineEnd - spanLineStart + 1
+
+    strEndLoc = loc 1 (fromInteger . toInteger . length $ str)
+
+    addColumn col l = loc (locLine l) (locColumn l + col)
+    isTokenUnknown (TokenUnknown _) = True
```
int-index: `isTokenUnknown = isRight . matching _TokenUnknown`

auxx/src/Lang/Lexer.hs
```diff
@@ -141,17 +143,27 @@ detokenize = unwords . List.map tokenRender
...
 tokenize' = parseMaybe (between pSkip eof (many pToken))
 
-pToken :: Lexer Token
-pToken = (try pToken' <|> pUnknown) <* pSkip
+pToken :: Lexer (Span, Token)
+pToken = withPosition (try pToken' <|> pUnknown) <* pSkip
+  where
+    posToNum :: (Num a, Num b) => SourcePos -> (a, b)
+    posToNum SourcePos{..} =
+        ( fromInteger . toInteger . unPos $ sourceLine
```
int-index: `fromInteger . toInteger = fromIntegral`

auxx/src/Lang/Lexer.hs
```diff
@@ -141,17 +143,27 @@ detokenize = unwords . List.map tokenRender
...
+        , fromInteger . toInteger . unPos $ sourceColumn)
+    withPosition p = do
+        pos1 <- (uncurry loc) . posToNum <$> getPosition
```
int-index: Why parenthesize `uncurry loc`?

auxx/src/Lang/Parser.hs
```diff
@@ -66,16 +67,16 @@ gExpr = mdo
         ] <?> "atom"
     return ntExpr
 
-pExpr :: Parser Text [Token] Expr
+pExpr :: Parser Text [(Span, Token)] Expr
```
int-index: Here you can use abstract `s` instead of `Span`, too

auxx/src/Plugin.hs
```diff
@@ -79,10 +79,12 @@ rawExec mHasAuxxMode AuxxOptions{..} mSendActions = \case
         printAction <- getPrintAction
         let commandProcs = createCommandProcs mHasAuxxMode printAction mSendActions
         printAction "... the auxx plugin is ready"
-        forever $ withCommand $ \line -> do
-            expr <- eitherToThrow $ Lang.parse line
-            value <- eitherToThrow =<< Lang.evaluate commandProcs expr
-            withValueText printAction value
+        forever $ withCommand $ \line -> case Lang.parse line of
+            Left parseError -> printAction (show . Lang.ppParseError $ parseError)
```
int-index: What about updating `runWalletCmd` accordingly? Move this logic to a helper, and remove `Exception` instances for `ParseError` and `EvalError`

auxx/src/Plugin.hs
```diff
@@ -79,10 +79,12 @@ rawExec mHasAuxxMode AuxxOptions{..} mSendActions = \case
...
+            Right expr -> Lang.evaluate commandProcs expr >>= \eitherValue ->
+                case eitherValue of
+                    Left evalError -> printAction (show . Lang.ppEvalError $ evalError)
```
int-index: When I see `show` I automatically assume that the output will be ugly. Provide `renderAuxxDoc :: Doc -> Text` instead (you can implement it as `show` with a comment that it's actually beautiful)

auxx/src/Plugin.hs
```diff
@@ -77,14 +76,8 @@ rawExec ::
...
         printAction <- getPrintAction
-        let commandProcs = createCommandProcs mHasAuxxMode printAction mSendActions
         printAction "... the auxx plugin is ready"
-        forever $ withCommand $ \line -> case Lang.parse line of
-            Left parseError -> printAction (show . Lang.ppParseError $ parseError)
-            Right expr -> Lang.evaluate commandProcs expr >>= \eitherValue ->
-                case eitherValue of
-                    Left evalError -> printAction (show . Lang.ppEvalError $ evalError)
-                    Right value -> withValueText printAction value
+        forever $ withCommand $ \line -> runCmd mHasAuxxMode line mSendActions printAction
```
int-index: You could eta-reduce this if you make `line` the last parameter of `runCmd` (I find it weird that you pass it before configurations)

auxx/src/Plugin.hs
```diff
@@ -115,6 +105,29 @@ runWalletCmd mHasAuxxMode line mSendActions = do
...
+    => Maybe (Dict (MonadAuxxMode m))
+    -> Text
+    -> Maybe (SendActions m)
+    -> PrintAction m
+    -> m ()
+runCmd mHasAuxxMode line mSendActions printAction = do
+    let commandProcs = createCommandProcs mHasAuxxMode printAction mSendActions
+    case Lang.parse line of
+        Left parseError -> printAction (Lang.renderAuxxDoc . Lang.ppParseError $ parseError)
+        Right expr -> Lang.evaluate commandProcs expr >>= \eitherValue ->
```
int-index: Why not `\case`?

auxx/src/Lang/DisplayError.hs
```diff
@@ -0,0 +1,141 @@
...
+
+ppArgumentError :: ArgumentError -> Doc
+ppArgumentError ae@ArgumentError{..} =
+    if isEmptyArgumentError ae
+    then empty
+    else vcat $ missingKeysDoc <> irrelevantKeysDoc <> irrelevantPosDoc
+  where
+    setToDoc = hcat . punctuate comma . map (highlight . nameToDoc) . S.toList
+    missingKeysDoc =
+        if S.null aeMissingKeys
```
gromakovsky: Btw, `null` and `toList` from `Universum` are polymorphic, but I don't mind if you use `S.{toList, null}`.

auxx/src/Lang/Lexer.hs
```diff
@@ -141,17 +143,27 @@ detokenize = unwords . List.map tokenRender
...
-tokenize' :: Text -> Maybe [Token]
+tokenize' :: Text -> Maybe [(Span, Token)]
 tokenize' = parseMaybe (between pSkip eof (many pToken))
 
-pToken :: Lexer Token
-pToken = (try pToken' <|> pUnknown) <* pSkip
+pToken :: Lexer (Span, Token)
+pToken = withPosition (try pToken' <|> pUnknown) <* pSkip
+  where
+    posToNum :: (Num a, Num b) => SourcePos -> (a, b)
```
gromakovsky: Why not change it to `posToLoc`? AFAIU, it's only used in combination with `loc` (`uncurry loc . posToNum`).

auxx/src/Lang/Parser.hs
```diff
@@ -66,16 +67,14 @@ gExpr = mdo
         ] <?> "atom"
     return ntExpr
 
-pExpr :: Parser Text [Token] Expr
+pExpr :: Parser Text [(s, Token)] Expr
 pExpr = parser gExpr
 
-newtype ParseError = ParseError (Report Text [Token])
+data ParseError = ParseError Text (Report Text [(Span, Token)])
```
gromakovsky: Can you use records here please? I have no idea what `Text` denotes by looking at this type definition.

(it becomes clear if I look below, but I believe it should be clear by just looking at the definition)

gromakovsky:
@int-index please approve if all your comments have been addressed.

PR #1929, opened 2017-11-18
[CSL-1781, CSL-1874] Extract client and generator packages

Pastafarianist:
Summary of changes:

Two new modules are introduced: `cardano-sl-client` and `cardano-sl-generator`.
* `cardano-sl-client` contains the following modules which had previously been in `cardano-sl`:
    * `Pos.Client.KeyStorage`
    * `Pos.Client.Txp.*`
    * `Pos.Communication.Methods` which is split between `Pos.Client.Txp.Network` and `Pos.Client.Update.Network`
    * `Pos.Communication.Tx` which is renamed to `Pos.Client.Txp.Network`
    * `Pos.Communication.Update` which is renamed to `Pos.Client.Update.Network`
    * `Test.Pos.Client.Txp.*`
* `cardano-sl-generator` contains the following modules:
    * `Pos.Generator`
    * `Pos.Generator.*`
    * `Test.Pos.Block.Logic.*`

`Test.Pos.Helpers` is split between `cardano-sl` and `cardano-sl-generator`.

Also, some minor cleanup is done.
* `build-depends` is sorted alphabetically where it wasn't.
* Miscellaneous changes: see `generator/src/Pos/Generator/Block/Mode.hs` for example.

Notes of interest:

* There is a lot of boilerplate. In particular, I had to copy `Test.hs` and `Spec.hs` everywhere.
* `cardano-sl-client` exports lots of modules from `Pos.Client.Txp.*`, but there is no wrapper module `Pos.Client.Txp` with reexports. Also, there is `Pos.Client.Update.Network`, but there is no `Pos.Client.Update`. I'm not sure whether this is good style.


client/src/Pos/Client/Update/Network.hs
```diff
@@ -26,6 +30,18 @@ submitVote
 submitVote enqueue voteUpd = do
     sendVote enqueue voteUpd
```
gromakovsky: `submitVote = sendVote`, I propose to leave only one function.

PR #1930, opened 2017-11-19
[CSL-1869] MsgBlock extension/improvements

volhovm:
This PR introduces two changes:
1. I've implemented new type of `MsgBlock` that sends raw bytestring over the network. Listener logic is now replying with new message type only, though in recovery code old reply (`MsgBlockDirect`) is still supported. 
2. I improved performance of `getHeadersFromToIncl` (which is now called `getHeadersRange`) -- it now uses forward links and doesn't query `O(n)` headers, operating only on header hashes level. It should speed up responding to `MsgBlock`.

### DIFF
--- a/db/Pos/DB/Class.hs
+++ b/db/Pos/DB/Class.hs
@@ -130,7 +130,7 @@ instance {-# OVERLAPPABLE #-}
 type MonadBlockDBRead m = (MonadDBRead m, BlockchainHelpers MainBlockchain)
 
 getDeserialized
-    :: (MonadBlockDBRead m, Bi v)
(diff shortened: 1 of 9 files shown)


gromakovsky:
I agree, I also don't want to change network communication layer, because it's a quite complicated procedure (involves `OutSpec` stuff and such).
Also I don't know whether it's ok to do it in master. Still no answer from @georgeee on this questions :/

volhovm:
I'm closing this PR since the solution is wrong and should be discussed again.
Part of it was cherry-picked as a separate issue: https://github.com/input-output-hk/cardano-sl/pull/1971

PR #1932, opened 2017-11-20
[CSLD-146] Move BUILDING CARDANO SL FROM SOURCE chapter to internal docs.

denisshevchenko:
Since we have https://github.com/input-output-hk/cardano-sl/tree/master/docs for "internal technical docs", it would be better if this chapter was here.


denisshevchenko:
cc @domenkozar

domenkozar:
1. We should add a link at existing docs to point here
2. We should remove `how-to/build-cardano-sl-core-from-scratch.md` as a duplicate

denisshevchenko:
@domenkozar Thanks, fixed.

PR #1935, opened 2017-11-20
[CS-24] Design payments endpoints for V1

adinapoli-iohk:
This PR ships the tentative subset of the V1 API to perform payments in Cardano. In boils down to three simple endpoints, as `/api/txs/payments/{address}/{transaction}` has been reported as not being in use anymore by @DominikGuzei :

screen shot 2017-11-20 at 17 53 55

As you can see we now have only the `transactions` resource with the notable exception of the `fees` nesting. The logic is unimplemented, but what I want to have you guys review here is the new structure and the new semantic of the endpoints. In particular:

* `POST /transactions` now accepts a list of output "money distributions" by default, with single input, single output being a special case of the former;

* `GET /transactions` is what used to be `histories`.

Share you feelings!


wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -361,53 +355,95 @@ instance Arbitrary PasswordUpdate where
...
                       <*> arbitrary
-                      <*> fmap getPositive arbitrary
+                      <*> arbitrary
+                      <*> arbitrary
 
 type TxId = Text
 
--- | A Wallet Transaction.
+-- | A 'Wallet''s 'Transaction'.
+-- TODO(adinapoli) This still doesn't include the full payload frontend might
```
flyingleafe: Indeed. Probably we should think about adding such data as `isOutgoing` flag right now?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -348,11 +342,11 @@ instance Arbitrary AccountUpdate where
...
 
 -- | A type incapsulating a password update request.
-data PasswordUpdate = PasswordUpdate
-  { -- | The old password.
+data PasswordUpdate = PasswordUpdate {
     pwdOld :: !Text
-    -- | The new password.
+    -- ^ The old password.
   , pwdNew :: !Text
+    -- ^ The new password.
```
flyingleafe: Don't know how I missed that in previous PRs: why `pwdOld` and `pwdNew` are not `SpendingPassword`s here?

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
```diff
@@ -8,26 +8,28 @@ import qualified Cardano.Wallet.API.V1.Addresses as Addresses
...
 
 handlers :: Server Addresses.API
 handlers =  listAddresses
        :<|> newAddress
 
 listAddresses :: PaginationParams
               -> Handler (OneOf [Address] (ExtendedResponse [Address]))
-listAddresses PaginationParams {..} =
+listAddresses PaginationParams {..} = do
+  addresses <- liftIO $ generate (vectorOf 2 arbitrary)
```
flyingleafe: I remember how you asked me to send you my `.dir-locals.el` which sets up 4-spaces indentation. Apparently you forgot to put it to use :wink:

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -258,14 +257,26 @@ instance ToDocs EstimatedFees where
   descriptionFor _ = "Estimated fees for a `Payment`."
 
 instance ToDocs Payment where
-  descriptionFor _ = "A request for exchange of `Coins` from one entity to another."
+  descriptionFor _ = "An exchange of `Coin`(s) from one source to one or more destinations."
```
flyingleafe: Minor: word "exchange" seemingly doesn't fit here. Why not use word "transaction"?

wallet-new/src/Cardano/Wallet/API/V1/Transactions.hs
```diff
@@ -0,0 +1,19 @@
...
+    =    "transactions" :> Summary "Generates a new transaction from the source to one or multiple target addresses."
+                        :> ReqBody '[JSON] Payment
+                        :> Post '[JSON] Transaction
+    :<|> "transactions" :> Summary "Returns the transaction history, i.e the list of all the past transactions."
+                        :> WalletRequestParams
+                        :> Get '[JSON] (OneOf [Transaction] (ExtendedResponse [Transaction]))
+    :<|> "transactions" :> "fees"
+                        :> Summary "Estimate the fees which would originate from the payment."
+                        :> ReqBody '[JSON] Payment
+                        :> Post '[JSON] EstimatedFees
```
flyingleafe: I remember how we discussed a good idea of having hierarchical URI structure for such api endpoints (i. e. `/api/v1/transactions/<wallet_id>/<account_id>/<address_id>`). I don't remember that we ever dismissed that idea. What is your plan on this? Are you going to implement such design later? I personally don't think that we should implement something that we know will have to be re-implemented later.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -361,53 +355,95 @@ instance Arbitrary PasswordUpdate where
...
 data Payment = Payment
-  { -- | The source Account.
-    pmtSourceAccount      :: !Account
-    -- | The destination Address.
-  , pmtDestinationAddress :: !Address
-    -- | The amount for this payment.
-  , pmtAmount             :: !Coins
+  { pmtSourceWallet   :: !WalletId
+    -- ^ The source Wallet.
+  , pmtSourceAccount  :: !AccountId
```
flyingleafe: See my comment about hierarchical URIs. If we'll have them, this info about source will be included not into this datatype, but into URL captions.

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
adinapoli-iohk: @flyingleafe Proof! It's here! 😀 

screen shot 2017-11-21 at 12 12 29

I don't know what happened here! Maybe the editor is required my to press <tab> to indent correctly and is not forcing the indentation from first principles? I will have to investigate.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
adinapoli-iohk: I have refrained from using "transaction" to not confuse the reader as the input type is `Payment` but we eventually yield a `Transaction`. What about "transfer"? Sounds like a nice middle-ground to me 😉

wallet-new/src/Cardano/Wallet/API/V1/Transactions.hs
adinapoli-iohk: Let me explain the reasoning behind this.

As written [in this RESTful best-practices guide](http://www.restapitutorial.com/media/RESTful_Best_Practices-v1_1.pdf), API are for consumers and the API should make sense to them before everything else, or we have lost in our attempt to make their lives easier.

In the section labeled "Resource URI Examples", there are a bunch of examples for hierarchical resources, and you see  how deeply hierarchical resources would need to be properly interspersed with piece of URLs. For example:

`GET http://www.example.com/customers/33245/orders/8769/lineitems`

Now, this is a properly designed piece of API. You can immediately establish that we are getting all the `lineitems` for order `8769` of customer `33245`. this is because who wrote this enforced the well-known convention of putting free-variables in the URL which are qualified entirely by the URL path which precedes them.

In other terms, here `33245` is immediately after `customers` therefore it must be a customer id. Same goes for `8769`, which follows `orders`.

Now, look at this:

`/api/v1/transactions/J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg/10/5oP9ib6ym3XdFXvmLXfUVZWw537vicy11ppa5bo8Ku9weByZAUaXWknjiA4dzK84Si`

First of all, this is really hard for a human to parse and visualise, which is not great for ergonomics. Secondly, this betrays the principle above. What's `J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg`? Is this a transaction ID, a wallet ID, something different? There is no way to tell and you cannot use conventions because this way we have just violated those.

In short, I think hierarchical resources makes lot of sense, if applicable. Here the resource we are dealing with is a `Transaction`. _Transaction and wallet does not form a hierarchy because transaction does not logically nests wallets_. If we had some hierarchical nesting, then it would have made lot of sense to have something like `/api/v1/transactions/<wallet_id>/accounts/<account_id>/addresses<address_id>`, but I think that for this one we should go with my proposal.

I hope this clarifies. What do you think?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
adinapoli-iohk: Excellent point, will amend.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
adinapoli-iohk: Not sure about this last one as I would like to see maybe if we can coalesce `isOutgoing` and `isLocal` into a single datatype. Do you think it's a good idea or we should just stick with the 2 separated booleans? I was thinking something along the line of (I suck at names, feel free to adjust/propose):

```
data TransactionType =
    LocalTransaction
  | OutgoingTransaction
```

and then have something like (in the final json):

```
{
  "type": "local"
}
```

Terrible idea?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
flyingleafe: Yes, "transfer" seems good indeed.

wallet-new/src/Cardano/Wallet/API/V1/Transactions.hs
flyingleafe: >Transaction and wallet does not form a hierarchy because transaction does not logically nests wallets.
I'm not sure if I understood your point here. I personally think that if we want to give a user an opportunity to get transaction history not only for a whole wallet, but also for a _particular account_ or a _particular address_, then hierarchical URI make a lot of sense. And we probably do want to give such opportunity (we actually have it in current API, though its implementation is far from perfect). I agree though that we should include `accounts` and `addresses` parts in URI for clarity.

Even if you think that we don't want to provide such opportunity, we should have wallet ID included in the URI anyway (now it looks like the endpoint returns transaction history for _all_ the wallets mixed together, that is totally not what we want :wink:)

Regarding POST requests - this is a bit less intuitive. My vision is that using hierarchical URIs we can provide an option to more finely restrict allowed money source for a new transaction, i. e. a user can use money from the whole wallet, only from single account or only from single address. Though I admit that such URI structure for _creating_ a transaction looks less intuitive and that we probably don't really need to provide such options for user. But if we eventually go with hierarchical URIs for GET requests (getting transaction history), same URI structure for POST requests will look nice and symmetric.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
flyingleafe: I like this idea, though then there should be three types of transaction:
```
data TransactionType =
    LocalTransaction
  | IncomingTransaction
  | OutgoingTransaction
```
I believe though that you had that in mind, just accidentally forgot one line :).

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
flyingleafe: @adinapoli-iohk 🤔 
Check out the value of `enable-dir-local-variables` variable in your emacs, it should be set to `t` for dir-locals to work. If it is set to `t`, then I don't have other immediate ideas on why this doesn't work for you.
Anyway, you got my general point - let's indent with 4 spaces, like in the rest of the project :)

adinapoli-iohk: Aye, it's definitely picking the file up as `stylish-haskell` correctly format on save. No clue! Will investigate. But for the sake of indentation absolutely, leave it with me.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
adinapoli-iohk: I like it, let's do it 😉

wallet-new/src/Cardano/Wallet/API/V1/Transactions.hs
adinapoli-iohk: I have reflected on what you said, and I see where you are coming from, trust me, but let me explain why I think might be cleaner to go with `/transactions` only 😉 

What you describe are all _filtering_ problems, for which we have an epic [here](https://iohk.myjetbrains.com/youtrack/issue/CSL-1905) . Given a collection, you want to zoom into a subset of elements satisfying some criteria. 

Obviously here you have two choices; you can either nest everything under `/wallets` (like, for example `/wallets/{wallet_id}/accounts/{account_id}/addresses/{address}/transactions`) or to keep everything separated into a separate resource and filter accordingly (e.g. `/transactions?wallet_id=xxx&account_id=xxx&address=xxx`). The first, by experience, comes with some drawbacks:

- It doesn't fully reflect an hierarchy. Typically RESTful resources, when nested, indicates _"a collection of"_. The very same definition of `hierarchy`, in the dictionary, says: _"an arrangement or classification of things according to relative importance or inclusiveness._". Here you don't have the concept of inclusiveness, because a wallet is conceptually composed of accounts & addresses, but _not_ by transactions. Transactions represents separate resources in the system and as such should be separated at the RESTful level.

- It's not composable. In this particular case does not matter, because you cannot further "slice & dice" the data, but you cannot add more filters without breaking the API, which is something you can do if you filter via query params.

Said that, I think for the trifecta `<wallets, accounts, addresses>` it *does* make sense to have the nested representation, I just think transaction shouldn't belong there. If we would follow an arbitrary nesting strategy, you wouldn't know where to stop with nesting. By the same principle, you might in future want to go with:

```
/wallets/{wallet_id}/accounts/{account_id}/addresses/{address_id}/transactions/{transaction_id}/...
```

...and you wouldn't know where to draw a line. The line I'm proposing here is very simple: if you can think about a URL as representing a _view_ or a _zoom_ on a collection, then you should nest the resources, otherwise represent the resources separately.

Said that, considering what's written [here](http://www.restapitutorial.com/lessons/restquicktips.html):

> Producing a great API is 80% art and 20% science. 

and provided frontend devs would also need to use and be comfortable with this API, I suggest maybe @darko-mijic gives us his unbiased opinion on what makes most sense, by experience 😀

DominikGuzei: I fully agree with @adinapoli-iohk here, for us it's only easy to construct URL paths for nested resources (because the path itself never changes, only the provided IDs -> e.g: `/api/v1/transactions/<wallet_id>/accounts/<account_id>/addresses<address_id>`).

For (optional/mixed) filtering params, it's easier to move them into query params, since we can hand them over as a simple object with any combination of properties (which are then turned into query params by the request).

Also for POST requests i fully agree with @adinapoli-iohk that we absolutely have to send the wallet addreses etc. as body params and not include them in the URL (we even had an URL encoding bug because of this in Daedalus / cardano … which was unnecessary).

Standard REST: "you want to create a transaction? Then it's `POST /api/transactions`

darko-mijic: I fully agree with @DominikGuzei and @adinapoli-iohk.

adinapoli-iohk: Thanks for chiming in, guys! Ok, looks like we were able to take an informative decision. I'm very happy therefore to stick with the proposed design and review this later in case we find out it's not suitable for our endeavours, but at least we get the prototype rolling and we don't get stuck on design decisions we can review before freezing V1 in place.

PR #1936, opened 2017-11-20
[CSL-2031] logic/diffusion split

avieth:
Here's the work done so far on defining and separating logic from diffusion.

## WorkMode

The main difficulty here is that much of the logic layer interface is currently expressed via class constraints (see `WorkMode`). We have to decide in which monad the diffusion layer will work. Should it be something that must satisfy `WorkMode` (i.e. `RealMode`)? Absolutely not, because in such a monad you can do way too much. We'd prefer to have something weaker like `Production`/`IO`. The diffusion layer assumes a much weaker `DiffusionWorkMode m`, and takes the logic layer assumptions as a function parameter of type `Logic m`.

## MessageLimited

I'm forced to change the message limits infrastructure. This piece is the business of the diffusion layer, but which is in informed by the logic layer. Currently there's 2 classes, one for "pure" limits and one for impure limits, which enforces a `MonadGState` constraint. I've changed it to be one class, `MessageLimited t m`, so that the "pure" case is simply `MessageLimited t Identity`. In practice, things with pure limits have `Applicative m => MessageLimited t m` instances. Types which have impure limits (formerly used the `MonadGState` constraints) now have `Data.Reflection.Given` constraints which are discharged when setting up the diffusion layer, using the logic layer. They're also found in the existing `Pos.Launcher` area, just to keep things unbroken.

## Workers and OutSpecs

It doesn't make any sense to put `OutSpecs` on workers, as we do now, because workers are logic-layer things. We'll just throw all of these `OutSpecs` into the diffusion layer, since they're derived from the diffusion layer interface (it determines the set of all possible network requests).


lib/src/Pos/Logic/Types.hs
```diff
@@ -0,0 +1,171 @@
...
+import           Pos.Core.Types            (HeaderHash, StakeholderId,
+                                            ProxySKHeavy)
+import           Pos.Core.Txp              (TxId, TxAux)
+import           Pos.Core.Update           (UpId, UpdateVote, UpdateProposal, BlockVersionData)
+import           Pos.Ssc.Message           (MCOpening, MCShares, MCCommitment,
+                                            MCVssCertificate)
+
+-- | The interface to a diffusion layer, i.e. some component which takes care
+-- of getting data in from and pushing data out to a network.
+data Logic m = Logic
```
avieth: Here's the interface so far (may discover other things are needed as I move forward).

Some things to note:
- See comments surrounding the `KeyVal` type for relayed things.
- Adopted `BlockVersionData` must be here, because it's used to implement type-directed message limits.
- Ultimately, the `m` parameter will *not* be `RealMode`. It must be something weaker like `Production`/`IO`. That's because the diffusion layer must use the logic layer features, and diffusion layer must not have `RealMode`.

avieth: @georgeee

avieth:
@dcoutts @georgeee ready for review. I've left as much as possible in-tact and copied everything I needed into diffusion layer modules under `Pos.Diffusion`.

We'll need a full implementation of a `Pos.Logic.Types.Logic`. Perhaps someone from the core team can do this, but I don't mind doing it myself either.

avieth:
I've been trying to do `pkgs/generate.sh` so that CI may have a hope of passing but it just hangs. `stack2nix` doesn't seem to be doing anything, but I can't ctrl+c or ctrl+z and must SIGKILL to make it stop.

avieth:
@georgeee @dcoutts Can I get a review please? Changes hit master very fast in this repository and it's tedious to keep up with a daily nontrivial rebase.

lib/src/Pos/Logic/Types.hs
```diff
@@ -0,0 +1,207 @@
...
+import           Pos.Core.Txp              (TxId)
+import           Pos.Core.Update           (UpId, UpdateVote, UpdateProposal, BlockVersionData)
+import           Pos.Security              (SecurityParams (..))
+import           Pos.Ssc.Message           (MCOpening, MCShares, MCCommitment,
+                                            MCVssCertificate)
+import           Pos.Util.Chrono           (NewestFirst, OldestFirst, NE)
+
+-- | The interface to a diffusion layer, i.e. some component which takes care
+-- of getting data in from and pushing data out to a network.
+data Logic m = Logic
```
volhovm: Why is it called `Logic` if it's used as csl interface for diffusion layer? Why didn't you decide to go with set of constraints instead? 
I do not understand design decisions of this PR tbh, I don't find PR description deep enough, it's more stating things rather than explaining them.

lib/src/Pos/Logic/Types.hs
```diff
@@ -0,0 +1,207 @@
...
+    { runLogicLayer = identity
+    , logic         = dummyLogic
+    }
+
+  where
+
+    dummyLogic :: Applicative m => Logic m
+    dummyLogic = Logic
+        { ourStakeholderId   = stkhldId
+        , getBlock           = \_ -> pure (error "dummy: can't get block")
```
volhovm: Why using `error` here is safe?

lib/src/Pos/Diffusion/Full.hs
```diff
@@ -0,0 +1,445 @@
...
+                , maxHeaderSize = bvdMaxHeaderSize <$> getAdoptedBVData logic
+                }
+
+            diffusion :: Diffusion d
+            diffusion = Diffusion {..}
+
+        return DiffusionLayer {..}
+
+  where
+    -- TBD will we need any resources here?
```
volhovm: I don't see any rational usecases of resources acquisition that may be related to diffusion layer. It's one-per-runtime thing and it doesn't interact with anything except some in-memory storage.

lib/src/Pos/Diffusion/Full.hs
```diff
@@ -0,0 +1,445 @@
...
+        networkConfig <- intNetworkConfigOpts networkConfigOpts
+
+        -- Make the outbound queue using network policies.
+        oq :: OQ.OutboundQ (EnqueuedConversation d) NodeId Bucket <-
+            initQueue networkConfig mEkgStore 
+
+        let -- VerInfo is a diffusion-layer-specific thing. It's only used for
+            -- negotiating with peers.
+            ourVerInfo :: VerInfo
+            ourVerInfo = VerInfo (getProtocolMagic protocolMagic) lastKnownBlockVersion ins (outs <> workerOuts)
```
volhovm: Block version can and will change while diffusion layer is running. How do we deal with that fact?

lib/src/Pos/Diffusion/Full/Types.hs
```diff
@@ -0,0 +1,43 @@
...
+type DiffusionWorkMode m
+    = ( WithLogger m
+      , CanJsonLog m
+      , MonadMockable m
+      , MonadIO m
+      , HasConfiguration
+      , HasInfraConfiguration
+      , HasUpdateConfiguration
+      , HasSscConfiguration
+      , HasNodeConfiguration
```
volhovm: `HasConfigurations`.

infra/Pos/Communication/Limits/Types.hs
```diff
@@ -61,45 +58,48 @@ class MessageLimited a where
...
          )
-         => MessageLimitedPure (Either a b) where
-    msgLenLimit = 1 + max (Left <$> msgLenLimit) (Right <$> msgLenLimit)
-
-instance ( MessageLimitedPure a
-         , MessageLimitedPure b
+         => MessageLimited (Either a b) m where
+    getMsgLenLimit _ = f <$> getMsgLenLimit Proxy <*> getMsgLenLimit Proxy
+      where
+      f limA limB = 1 + max (Left <$> limA) (Right <$> limB)
```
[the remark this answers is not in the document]
avieth: Unless I made a mistake, this should be exactly as it was before: 1 byte plus the maximum of the limits on either of the types.

Ok as I wrote this out I see what you're thinking. It's definitely confusing: the type `Limit a` has a `Num` instance. What we have here is two `Limit (Either a b)` terms which are, under the hood, just numbers, and `max` takes their maximum as numbers.

lib/src/Pos/Diffusion/Full.hs
avieth: Oh interesting point. I'm not sure if we have ever dealt with that! `VerInfo` is and always has been constant for the life time of the process.

lib/src/Pos/Diffusion/Full/Delegation.hs
```diff
@@ -0,0 +1,45 @@
+{-# LANGUAGE RankNTypes #-}
+
+-- | Server listeners for delegation logic
+--
+-- TBD where is the other half of the story: delegation requests?
```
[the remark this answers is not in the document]
avieth: Yes. I don't know where to find the delegation request making pieces. Are `sendTx` and similar used for this?

lib/src/Pos/Logic/Types.hs
avieth: Oops, the comment is totally wrong. I must have copy/pasted the diffusion file and then re-wrote it to have the logic stuff.

avieth: > Why didn't you decide to go with set of constraints instead?

Using typeclass constraints could get us to a similar solution, but it wouldn't save us any work nor simplify anything. Ultimately we would have to discharge these constraints, and that amounts to giving exactly what we have here: values of type `Logic` and `Diffusion` or some union of the two.

lib/src/Pos/Communication/Limits.hs
```diff
@@ -206,94 +224,120 @@ instance MessageLimited MCShares where
...
+instance (MessageLimited a m, Functor m) => MessageLimited (DataMsg a) m where
     --type LimitType (DataMsg a) = LimitType a
     getMsgLenLimit _ = (fmap . fmap) DataMsg (getMsgLenLimit (Proxy @a))
 
 ----------------------------------------------------------------------------
 ---- Txp
 ----------------------------------------------------------------------------
 
-instance MessageLimited TxAux where
+data TxpLimits m = TxpLimits
```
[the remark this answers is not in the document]
avieth: Right I see now that they're all derived from the `BlockVersionData`. I'll make the change.

lib/src/Pos/Diffusion/Types.hs
```diff
@@ -0,0 +1,88 @@
...
+import           Pos.Core.Block                   (Block, BlockHeader, MainBlockHeader)
+import           Pos.Core.Types                   (HeaderHash)
+import           Pos.Core.Txp                     (TxAux)
+import           Pos.Core.Update                  (UpId, UpdateVote, UpdateProposal)
+import           Pos.Ssc.Message                  (MCOpening, MCShares, MCCommitment,
+                                                   MCVssCertificate)
+
+-- | The interface to a diffusion layer, i.e. some component which takes care
+-- of getting data in from and pushing data out to a network.
+data Diffusion m = Diffusion
```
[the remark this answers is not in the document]
avieth: The `Logic` interface has `postPskHeavy :: ProxySKHeavy -> m Bool` which is used by the delegation relay (set up in the full diffusion layer `Pos.Diffusion.Full`).

lib/src/Pos/Diffusion/Types.hs
```diff
@@ -0,0 +1,88 @@
...
+    { -- Get all blocks from a set of checkpoints to a given tip.
+      getBlocks          :: NodeId
```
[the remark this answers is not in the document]
avieth: This is more or less the type that block retrieval code uses so I thought I'd keep it the way it was for the initial patch. Eventually we'll be able to get rid of the `NodeId` parameter (diffusion layer will figure out whom to ask).

lib/src/Pos/Diffusion/Types.hs
```diff
@@ -0,0 +1,88 @@
...
+      -- SSC: send our shares (diffusion layer takes care of relaying shares
+      -- for other stakeholders).
+    , sendSscShares      :: MCShares -> m ()
+      -- SSC: send our commitment (diffusion layer takes care of relaying
+      -- commitments for other stakeholders).
+    , sendSscCommitment  :: MCCommitment -> m ()
+    }
+
+-- | Failure description for getting blocks from the diffusion layer.
+data GetBlocksError = GetBlocksError Text
```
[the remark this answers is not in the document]
avieth: Could make an exception hierarchy for these and call it `DiffusionLogicError`.

avieth:
Could we keep the review rolling? @georgeee @volhovm I'd like some feedback on the logic interface in particular `lib/src/Pos/Logic/Types.hs`. Also: I can start porting existing stuff to make the full logic layer, or someone from the core team could do it. Can we get that decided and scheduled?

CC @dcoutts

lib/src/Pos/Diffusion/Full/Delegation.hs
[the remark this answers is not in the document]
volhovm: https://github.com/input-output-hk/cardano-sl/blob/master/auxx/src/Command/Proc.hs#L390

lib/src/Pos/Logic/Types.hs
```diff
@@ -0,0 +1,204 @@
...
+      -- went wrong.
+      --
+      -- NB light delegation is apparently disabled in master.
+    , postPskHeavy      :: ProxySKHeavy -> m Bool
+
+      -- Recovery mode related stuff.
+      -- TODO get rid of this eventually.
+    , recoveryInProgress :: m Bool
+
+      -- TBD does this need to be in the monad? It's not constant?
```
volhovm: Looks like it is constant.

lib/src/Pos/Diffusion/Types.hs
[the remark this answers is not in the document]
volhovm: I do not understand the separation. Why there's no `sendPskHeavy` from `postPskHeavy` like there is `sendTx` from `postTx`?

volhovm:
@avieth You didn't address all my comments and didn't make any changes (am i missing something?), so I thought there's no need for another review yet. There's a very important question about the design choice of monad configuration (record) instead of typeclass approach that i'd like to see explanation about. You basically duplicate all the existing logic instead of using `HasX` typeclasses we have.

lib/src/Pos/Logic/Types.hs
avieth: Right I guess it's monadic in existing code just because it's produced from a reader/lens combo.

lib/src/Pos/Diffusion/Types.hs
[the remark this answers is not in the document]
avieth: Because I couldn't find `sendPskHeavy` :)

lib/src/Pos/Diffusion/Full/Delegation.hs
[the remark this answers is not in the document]
avieth: Thanks

lib/src/Pos/Diffusion/Types.hs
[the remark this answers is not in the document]
avieth: You pointed me to the auxx implementation. I'll copy that out and put it here.

avieth:
> There's a very important question about the design choice of monad configuration (record) instead of typeclass approach that i'd like to see explanation about. You basically duplicate all the existing logic instead of using HasX typeclasses we have.

My justification is that the `HasX` typeclass appraoch doesn't buy us anything; there's no advantage to it over this record-based approach. Suppose we did the `HasX` approach. Then we have a bunch of `HasX` constraints to discharge in order to run the program. How do we do that? We define all the necessary terms, i.e. we create something that is at least as big as the `Logic l` and `Diffusion d` records combined. So in the end we've done all the work we would have done in the record approach, but with the added pain of defining a whack of classes and instances that we have to continue to maintain.

avieth:
## MessageLimited is a weird class

I had some trouble with those `MessageLimited` instances which depend upon the mutable latest adopted block version data. I made it work by defining `class HasBlockVersionData m` (suggested by @volhovm ) and giving instances which use `MonadGState m`, as well as instances which use `Given (m BlockVersionData)`; the former for existing cardano-sl, the latter for use by the diffusion layer, which should not introduce a `MonadGState` constraint on its monad parameter (the g state comes from and is encapsulated by the logic layer).

Having a `MessageLimited` instance as it stands now implies that the type of the limited thing and the monad in which the limit is computed jointly determine how to compute the limit, but I think this is dishonest and too onerous. It means the monad itself has to contain all of the necessary information--in this case, the mutable latest adopted `BlockVersionData`--and so this becomes a necessary part of any monad we want to use.

Alternatively, the particular choice of limit (`:: m (Limit t))` determines those types which it limits, along with those monads which are capable of carrying the computation. Such a limit definition can use the capabilities of `m`, *and* definitions which are not a part of `m`, like for instance a `Logic m` record. The honest truth is that the `BlockVersionData`-backed limits we have now are legit in any monad which can do the IO necessary to read the mutable latest adopted data, but I believe they're expressed by a type class only for the perceived convenience of not having to explicitly give them at the call site of `recvLimited`.

But is it such a convenience? It's probably less awkward if we do away with `MessageLimited` altogether, and use terms rather than types in order to come up with limits.

Instead of a `MessageLimited t m`, just define a term of type `m (Limit t)` (or `Limit t` for things which don't need any monadic stuff). Just as `MessageLimited` instances for composite types make use of other such instances, limit terms for composite types can use other limit terms. For all of the types for which we `recvLimited`, we just have to import the relevant limit.

```Haskell
-- Give the limit, don't expect it to magically fall out of a class instance
recvWithLimit :: Monad m => ConversationActions snd rcv m -> m (Limit rcv) -> m (Maybe rcv)
recvWithLimit cactions mkLimit = mkLimit >>= recvLimited cactions . toWord32
```

For the `BlockVersionData`-dependent limited things like block headers, their limits can be derived from `Logic m`:

```Haskell
blockHeaderLimit :: Functor m => Logic m -> m (Limit BlockHeader)
blockHeaderLimit logic = bvdMaxHeaderSize <$> adoptedBVData logic
```

Since the diffusion layer has a `Logic m` in scope, it can come up with thiese limits wherever necessary.

With a solution like this, we don't have to bend and twist the monad in order to accommodate limits, nor resort to `Data.Reflection`. In particular, we don't have to bake the `m BlockVersionData` into the monad itself, we can just grab it from the logic layer. It looks overall simpler to me. Thoughts?

## Injecting dependencies into the monad itself versus defining dependencies over the monad

Staying with the theme: does the monad determine the implementation, or does the implementation determine the monad?

When working with classes like those found in `mtl` (`MonadReader r m | m -> r`), the monad determines the implementation. They go to say: this monad `m` is *intrinsically* a reader / state / writer / whatever, there's no choice left to be made.

When working with free monads, all of the terms are symbolic, and their actual implementation is picked later on, targeting some other monad. Their implementations determine the target monad. Note that this is just a slick way of doing dependency injection via higher-order functions: you could lambda-abstract over the implementations and do your work in the target monad determined by those implementations.

The goals of `mtl`-style classes and of free monads / dependency injection have some overlap: to do something symbolically/abstractly, leaving open a choice of concrete implementation. For the former, the implementation is chosen by picking a type constructor; for the latter, by picking a term. Anyway, it's a desirable thing, so it's no surprise that it comes up in cardano-sl. I've noticed that we typically (always?) have the monad determine the implementation.

Both of these approaches are equally expressive. The disadvantage of the `mtl`-style monad-determines-the-implementation variant is that we have to define a new monad and a whack of class instances for each implementation. Also, we have to bundle up all of the term-level dependencies (those things which are used by the implementations) in a reader context, just so that the *type* can determine the implementation.

For the logic/diffusion record-based design I've chosen implementation-determines-the-monad. The logic layer interface comes as a first-class term rather than out of a hidden typeclass dictionary. This just saves one step of indirection. If we did a class-based monad-determines-the-implementation approach, the logic and diffusion interface records would still be there, they'd just be a mishmash of typeclass dictionaries.

Just some food for though regarding the monad transformer stack discussion CC @georgeee @dcoutts

georgeee:
CC @int-index

lib/src/Pos/Diffusion/Full.hs
```diff
@@ -0,0 +1,449 @@
...
+import           Pos.Diffusion.Subscription.Common (subscriptionListeners)
+import           Pos.Diffusion.Subscription.Dns (dnsSubscriptionWorker)
+import           Pos.Diffusion.Subscription.Dht (dhtSubscriptionWorker)
+import           Pos.Update.Configuration (lastKnownBlockVersion)
+import           Pos.Util.OutboundQueue (EnqueuedConversation (..))
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+-- Orphan instance to get the adopted block version data through reflection.
+instance (Given (m BlockVersionData)) => HasAdoptedBlockVersionData m where
```
int-index: This orphan instance doesn't look good to me, it conflicts with other existing instances. With what `m` instantiations is it used? Would it be possible to specialize `m`? I.e. `instance Given (Production BlockVersionData) => HasAdoptedBlockVersionData Production`

lib/src/Pos/Logic/Types.hs
```diff
@@ -0,0 +1,203 @@
...
+
+-- | Failure description for getting the tip of chain from the logic layer.
+data GetTipError = GetTipError Text
+
+deriving instance Show GetTipError
+instance Exception GetTipError
+
+-- | A diffusion layer: its interface, and a way to run it.
+data LogicLayer m = LogicLayer
+    { runLogicLayer :: forall x . m x -> m x
```
int-index: `m` in negative position again :disappointed:

int-index: it's really hard to reason about manual dictionaries with `m` in negative positions when it comes to unlifting (I go into more details in CSL-1823)

auxx/src/Command/BlockGen.hs
```diff
@@ -24,7 +27,7 @@ import           Lang.Value (GenBlocksParams (..))
 import           Mode (MonadAuxxMode)
 
 
-generateBlocks :: MonadAuxxMode m => GenBlocksParams -> m ()
+generateBlocks :: forall m . MonadAuxxMode m => GenBlocksParams -> m ()
```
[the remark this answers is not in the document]
avieth: Oops, leftovers. I made a change and then thought better and undid it. I'll fix this.

lib/src/Pos/Communication/Limits.hs
```diff
@@ -206,48 +218,57 @@ instance MessageLimited MCShares where
...
-instance MessageLimited UpdateProposal where
+instance (HasAdoptedBlockVersionData m, Functor m) => MessageLimited UpdateProposal m where
     -- FIXME Integer -> Word32
-    getMsgLenLimit _ = Limit . fromIntegral <$> DB.gsMaxProposalSize
+    getMsgLenLimit _ = Limit . fromIntegral <$> maxProposalSize
 
-instance MessageLimited (UpdateProposal, [UpdateVote]) where
+-- Overlapping because there's an instance for (a, b) but that one doesn't
+-- express what we want: a limit on the size of the list of votes.
+instance {-# OVERLAPPING #-} (HasAdoptedBlockVersionData m, Monad m) => MessageLimited (UpdateProposal, [UpdateVote]) m where
```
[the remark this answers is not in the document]
avieth: I'm all for it! I think there should have been one to begin with. However it's out of the scope of this change.

lib/src/Pos/Diffusion/Full.hs
avieth: The only instance it conflicts with is the `MonadGState`-backed one, used to keep the existing cardano-sl implementation up and running. That one isn't imported here so it's ok (no overlap).

If we specialized it then the full diffusion layer implementation given here would also have to specialize, but we don't want to specialize it. If we did, we'd have to lift / natural transform the monad parameter at use sites and that's cumbersome as you know (negative position `m`).

lib/src/Pos/Logic/Types.hs
avieth: So long as we never pick a concrete `m`, we don't have to do lifts or natural transformations.

Notice that the full diffusion layer implementation leaves `m` free subject only to some constraints `DiffusionWorkMode m, MonadFix m`. We can pick the concrete `m` and let GHC solve the constraints; no need to apply natural transformations.

You don't change the implementation (logic or diffusion layer record) in order to accommodate your choice of monad; you change your choice of monad to accommodate the implementation.

georgeee:
@avieth We had a call with @int-index , @gromakovsky , @volhovm .

We came to agreement that currently we have all-in mix in CSL monad stack approaches and till this discussion finalizes (which will hopefully happen soon), anything is legit (and everything would have to be rewritten finally, that's also the story).

So discard now most of feedback regarding record-passing style. This PR was much useful for @int-index in his work on CSL-1823

avieth:
> What are we going to do with VerInfo is Full?

Are you referring to the problem with mutable `BlockVersionData`? Currently in master the `lastKnownBlockVersion :: HasUpdateConfiguration => BlockVersion` is used at launch time. I do the same thing in the full diffusion layer. If this is actually not OK then we should create a new ticket. CC @georgeee

volhovm:
@avieth alright, but please create a ticket and make sure it's major and planned to be done soon (idk how sprints are usually planned).

avieth:
`scripts/launch/demo.sh 4` works. That's to be expected: it uses the old code, not the diffusion/logic split, for we haven't implemented a real logic layer yet.

avieth:
Travis mac fails :(

> Dec 05 19:00:58 tail: 'dist/test/test.log' has become inaccessible: No such file or directory

akegalj:
@avieth 
> Travis mac fails :(

Merge latest master to your branch. @gromakovsky addressed this problem already

avieth:
> Merge latest master to your branch. @gromakovsky addressed this problem already

This branch includes tip of master. 86ba0d19ba3e96bd46d7b978faf518e339a0c1c6 is in there

gromakovsky:
Last workaround was to decrease `k` in tests https://github.com/input-output-hk/cardano-sl/commit/596dc4ed6925899f511a4baf6dc5ffd69ca1c39a#diff-61feea0f7a7bc722b74eedc3907be256R195
It helped to make some builds pass, but apparently it doesn't prevent all timeouts :(
I think it's better to just restart the build in this PR until it's properly investigated in CSL-1966.

avieth:
> @avieth alright, but please create a ticket and make sure it's major and planned to be done soon (idk how sprints are usually planned).

https://iohk.myjetbrains.com/youtrack/issue/CSL-2022

avieth:
Mac build still failing D: it timed out

volhovm:
@avieth did you rebase on master? It should work (after several restarts?..).

avieth:
> @avieth did you rebase on master? It should work (after several restarts?..).

Latest master is in this branch but I squashed a commit and force pushed to get it restarted. Fingers crossed.

volhovm:
@avieth you'll need to rebase after https://github.com/input-output-hk/cardano-sl/pull/2071 is merged.

volhovm:
@avieth #2071 is merged.

avieth:
Mac builds are the only failing travis builds, and they both fail because they ran for too long.

avieth:
As for appveyor:

```
[00:00:00] Build started
[00:00:00] SET BRANCH_CACHE_DIR=%CACHE_DIR%\%APPVEYOR_REPO_BRANCH:/=_%
[00:00:00] SET CACHED_STACK_ROOT=%BRANCH_CACHE_DIR%\sr
[00:00:00] SET CACHED_STACK_WORK=%BRANCH_CACHE_DIR%\sw
[00:00:00] $env:VERSION = ($env:APPVEYOR_BUILD_VERSION).split(".")[0] + "." + ($env:APPVEYOR_BUILD_VERSION).split(".")[1]
[00:00:00] echo VERSION=%VERSION%
[00:00:00] VERSION=1.0
[00:00:00] git clone -q https://github.com/input-output-hk/cardano-sl.git C:\projects\cardano-sl
[00:00:18] git fetch -q origin +refs/pull/1936/merge:
[00:00:19] git checkout -qf FETCH_HEAD
[00:00:21] Restoring build cache
[00:00:24] 
[00:00:26] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...1%
[00:00:29] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...10%
[00:00:31] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...20%
[00:00:33] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...30%
[00:00:39] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...40%
[00:00:44] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...50%
[00:00:52] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...60%
[00:00:59] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...70%
[00:01:05] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...80%
[00:01:09] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...90%
[00:01:09] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...100%
[00:02:08] Cache 'c:\c\master' - Unzipping...
[00:02:08] Cache 'c:\c\master' - Restored
[00:02:08] 
[00:02:10] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...1%
[00:02:13] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...10%
[00:02:16] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...20%
[00:02:18] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...30%
[00:02:21] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...40%
[00:02:24] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...50%
[00:02:27] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...60%
[00:02:29] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...70%
[00:02:32] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...80%
[00:02:35] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...90%
[00:02:35] Cache 'c:\c\master' - Downloading (1,033,943,078 bytes)...100%
[00:03:34] Cache 'c:\c\master' - Unzipping...
[00:03:34] Cache 'c:\c\master' - Restored
[00:03:34] 
[00:03:35] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...1%
[00:03:36] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...10%
[00:03:38] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...20%
[00:03:39] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...30%
[00:03:40] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...40%
[00:03:41] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...50%
[00:03:42] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...60%
[00:03:44] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...70%
[00:03:45] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...80%
[00:03:46] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...90%
[00:03:46] Cache 'c:\c\cardano-sl-1.0' - Downloading (517,765,678 bytes)...100%
[00:04:19] Cache 'c:\c\cardano-sl-1.0' - Unzipping...
[00:04:19] Cache 'c:\c\cardano-sl-1.0' - Restored
[00:04:24] xcopy /q /s /e /r /k /i /v /h /y C:\projects\cardano-sl "%WORK_DIR%"
[00:04:35] 1291 File(s) copied
[00:04:35] cd "%WORK_DIR%"
[00:04:36] Echo %APPVEYOR_BUILD_VERSION% > build-id
[00:04:36] Echo BRANCH_CACHE_DIR = %BRANCH_CACHE_DIR%
[00:04:36] BRANCH_CACHE_DIR = c:\c\master
[00:04:36] IF EXIST %CACHE_DIR% dir %CACHE_DIR%
[00:04:36]  Volume in drive C is Windows
[00:04:36]  Volume Serial Number is D4AB-4044
[00:04:36] 
[00:04:36]  Directory of c:\c
[00:04:36] 
[00:04:36] 12/12/2017  03:30 PM    <DIR>          .
[00:04:36] 12/12/2017  03:30 PM    <DIR>          ..
[00:04:36] 12/02/2017  06:04 PM    <DIR>          cardano-sl-1.0
[00:04:36] 11/01/2017  08:39 PM    <DIR>          master
[00:04:36]                0 File(s)              0 bytes
[00:04:36]                4 Dir(s)  33,667,502,080 bytes free
[00:04:36] IF EXIST %CACHED_STACK_ROOT% xcopy /q /s /e /r /k /i /v /h /y %CACHED_STACK_ROOT% %STACK_ROOT%
[00:05:12] 6879 File(s) copied
[00:05:12] IF EXIST %CACHED_STACK_WORK% xcopy /q /s /e /r /k /i /v /h /y %CACHED_STACK_WORK% %STACK_WORK%
[00:05:57] 12690 File(s) copied
[00:05:57] (New-Object Net.WebClient).DownloadFile('https://slproweb.com/download/Win64OpenSSL-1_0_2m.exe', "$($env:USERPROFILE)\Win64OpenSSL.exe")
[00:05:58] Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
[00:05:58] At line:1 char:101
[00:05:58] + ... eb.com/download/Win64OpenSSL-1_0_2m.exe', "$($env:USERPROFILE)\Win64O ...
[00:05:58] +                                                  ~~~~~~~~~~~~~~~~
[00:05:58]     + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
[00:05:58]     + FullyQualifiedErrorId : WebException
[00:05:58]  
[00:05:58] Command executed with exception: Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
```

gromakovsky:
Appveyor is fixed in master: https://github.com/input-output-hk/cardano-sl/pull/2097

avieth:
Appveyor passes woo! ᕕ(⌐■_■)ᕗ

Mac travis still times out. Perhaps we could increase the time limit?

volhovm:
I think we can't -- 1.5h is a maximum value for travis, that has been a problem for a long time. You just restart and hope for better afaiu.

PR #1919, opened 2017-11-16
[CSL-1883] Support tx with multiple recipients

akegalj:
Issue: https://iohk.myjetbrains.com/youtrack/issue/CSL-1883

## Tested manually:

Cleaning db:
```
[akegalj@notebook cardano-sl]$ scripts/clean/db.sh
```
Running a local cluster:
```
[akegalj@notebook cardano-sl]$ scripts/launch/demo-with-wallet-api.sh
```
Generating genesis:
```
[akegalj@notebook cardano-sl]$ stack exec -- cardano-keygen --system-start 0  generate-keys-by-spec --genesis-out-dir tmp-secrets
[keygen:INFO] using configurations: ConfigurationOptions {cfoFilePath = "node/configuration.yaml", cfoKey = "default", cfoSystemStart = Just 0}
[keygen:INFO] withConfiguration using custom system start time 0
[keygen:INFO] Processing command
[keygen:INFO] Generating testnet data into tmp-secrets/keys-testnet
[keygen:INFO] 16 keyfiles are generated
[keygen:INFO] Generating fake avvm data into tmp-secrets/keys-fakeavvm
[keygen:INFO] 10 fake avvm seeds are generated
[keygen:INFO] tmp-secrets generated successfully
```

Creating an empty wallet:
```
$ curl -i -H "Accept: application/json" -H "Content-Type: application/json" -X POST -d "{   \"cwInitMeta\": {     \"cwName\": \"My wallet\",      \"cwAssurance\": \"CWANormal\",     \"cwUnit\": 0   },   \"cwBackupPhrase\": {       \"bpToList\": [\"squirrel\", \"material\", \"silly\", \"twice\", \"direct\", \"slush\", \"pistol\", \"razor\", \"become\", \"junk\", \"kingdom\", \"flee\"]     } }" -k https://localhost:8090/api/wallets/new
HTTP/2 200 
date: Mon, 20 Nov 2017 11:20:40 GMT
server: Warp/3.2.13
content-type: application/json;charset=utf-8

{"Right":{"cwId":"Ae2tdPwUPEZMeiDfNHZ45V7RoaSqd4oSMuG4jo7asvmNHS193EEad1tUkeT","cwMeta":{"cwName":"My wallet","cwAssurance":"CWANormal","cwUnit":0},"cwAccountsNumber":1,"cwAmount":{"getCCoin":"0"},"cwHasPassphrase":false,"cwPassphraseLU":1.511176839764960291e9}}
```

Creating another empty wallet:
```
[...]

PR #1937, opened 2017-11-20
csl1883 + tests for master

akegalj:
Original issue here: https://github.com/input-output-hk/cardano-sl/pull/1919

The purpose of this PR is to create tests which are available in the `master` branch and which are missing in #1919 branch.

For more info see #1919 

## TODO
* [x] write automated tests for `/api/txs/payments/batch`
       - this will be variant of https://github.com/input-output-hk/cardano-sl/blob/master/wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs#L45
* [x] remove `"purescript": "^0.11.5",` from `package.json`


daedalus/package.json
```diff
@@ -51,6 +51,7 @@
     "blakejs": "^1.1.0",
     "bs58": "^4.0.1",
     "json-loader": "^0.5.7",
+    "purescript": "^0.11.5",
```
adinapoli-iohk: Are you sure this is needed?

akegalj: @adinapoli-iohk good catch! Thanks, I will remove it.

That thing is added by `npm install` on my machine automatically and usually I do check it - but it was late. Thanks

added to TODO list

akegalj:
@adinapoli-iohk got it.

Lets proceed as you said:
> feel free to crack on with the tests, but my preference would be to merge just #1919 on cardano-sl-1.0, then cherry-pick the merge into master and then open a separate PR with just the tests against master

we will close this PR when tests are written.

adinapoli-iohk:
Thanks @akegalj , appreciated as always!

akegalj:
@pva701 note that current batch payment spec is super set of newPayment spec. So it will cover both.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -57,6 +59,21 @@ newPayment passphrase srcAccount dstAddress coin policy =
...
 
+newPaymentBatch
+    :: MonadWalletTxFull ctx m
+    => PassPhrase
+    -> NewBatchPayment
+    -> m CTx
+newPaymentBatch passphrase NewBatchPayment {..} = do
+    src <- decodeCTypeOrFail npbFrom
+    dst <- nonEmpty npbTo `whenNothing`
+        throwM (RequestError "At least one recipient required.")
```
pva701: you can use `NonEmpty` for `npbTo`

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
```diff
@@ -58,10 +71,19 @@ oneNewPaymentSpec = walletPropertySpec oneNewPaymentDesc $ do
     srcAddr <- getAddress srcAccId
     -- Dunno how to get account's balances without CAccModifier
     initBalance <- getBalance srcAddr
+    let topBalance = unsafeGetCoin initBalance `div` 2
     -- `div` 2 to leave money for tx fee
-    coins <- pick $ mkCoin <$> choose (1, unsafeGetCoin initBalance `div` 2)
-    void $ lift $ newPayment pswd srcAccId dstCAddr coins def
-    dstAddr <- lift $ decodeCTypeOrFail dstCAddr
+    -- FIXME: this might be inafficient
```
pva701: we should write a function which split `K` into `k` parts and put it to `Pos.Util`. it's used in several places

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
akegalj: yes, but then the same logic 
```
nonEmpty npbTo `whenNothing`
       throwM (RequestError "At least one recipient required.")
```

would have to be pushed inside `FromJSON` instance again (because there is no `FromJSON NonEmpty` and I would not use it as fail message is much more clear this way - mentioning recipients). 

On a second though I can add `FromJSON NonEmpty` and use generic fail message "Expected at least one element".

Do you think this would be better ?

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
```diff
@@ -70,22 +92,29 @@ oneNewPaymentSpec = walletPropertySpec oneNewPaymentDesc $ do
...
-        txMetasDst <- maybeStopProperty "Dst wallet doesn't exist" =<< lift (WS.getWalletTxHistory dstWalId)
-        void $ expectedOne "TxMeta for dst wallet" txMetasDst
+    -- (!!) is safe here as group will guarantee that
+    let countedWalIds = map (length &&& (!! 0)) . group . sort $ walId:dstWalIds
+    forM_ countedWalIds $ \(c, wId) -> do
+        txMetas <- maybeStopProperty "Wallet doesn't exist" =<< lift (WS.getWalletTxHistory wId)
+        -- FIXME: I am sure that if there was one output, there will
+        -- be exactly one txMeta. If there are multiple outputs tx meta
+        -- count is a bit complicated to calculate (not sure yet)
+        when (c == 1) $
```
pva701: It should be one `TxMeta` for each mentioned wallet. It's very strange if there is greater than one. I would investigate why this statement is needed to pass tests. A valid number of `TxMeta`s guarantees correct tx history.

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
```diff
@@ -39,14 +43,23 @@ spec :: Spec
 spec = withCompileInfo def $
        withDefConfigurations $
        describe "Wallet.Web.Methods.Payment" $ modifyMaxSuccess (const 10) $ do
-    describe "newPayment" $ do
-        describe "One payment" oneNewPaymentSpec
+    -- TODO: add seperate test for newPayment
+    -- this ^ is not really needed as newPayment endpoint is a real subset
```
pva701: I would remove this TODO if so.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
pva701: I think it would be better to check inside `FromJSON NewPaymentBatch`. it would just guarantee that you have at least one element and you don't need it check somewhere else. 
however, IMHO, it's more or less minor, I don't mind your approach.

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
akegalj: Can you be more specific?

I could write this as 
```
coins <- pick $ vectorOf destLen $ mkCoin <$> choose (1, unsafeGetCoin initBalance `div` (2*destLen))
```

where `destLen` is `n` and `initBalance` is `K` in your case. This would be a bit conservative as each output would contain max `K/(2*n)` coins. 

If we go with:
```
coins <- pick $ vectorOf destLen $ mkCoin <$> choose (1, unsafeGetCoin initBalance `div` destLen)
```

Then each output would contain max `K/n` coins and in this case we might trigger `Insufficient founds` error because we won't have enough money to count in for fees.

With:
```
let topBalance = unsafeGetCoin initBalance `div` 2
 coins <- pick $ suchThat (vectorOf destLen $ mkCoin <$> choose (1, topBalance)) ((< topBalance) . fromIntegral . sumCoins)
```

each output would contain max `K/2` coins so it increases search space (at least if some bug is triggered by bigger amounts being sent). Not sure how important it is but search space is a bit bigger.

> we should write a function which split K into n parts and put it to Pos.Util. it's used in several places

So my conclusion is that you thing we should have something like:
```
coinPart coin n = mkCoin $ unsafeGetCoin coin `div` n
```

am I right?

(I am not aware of other places - so maybe pointing this out would understand better what you meant)

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
akegalj: > It should be one TxMeta for each mentioned wallet

right, I got it confused with `CTxMeta` . I will fix that

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
pva701: If we have a function which split `K` into `n` parts, then we can choose `K <- choose (n, topBalance)` and then call this function

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
pva701: Ok I am wrong, in a real system, it should be one for all mentioned destination addresses.
It should be one only for source addresses in the test and zero for others addresses

pva701: I suppose only for source address `c == 1` and it works

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
akegalj: ok, I think I got it. Let me try and ping you back

pva701:
Also, I don't see an update of purescript api. dunno whether it should be done :man_shrugging:

wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
akegalj: yes, I agree.

akegalj:
@pva701 
> Also, I don't see an update of purescript api. dunno whether it should be done

it won't be needed. This is needed only for Binance exchange and they are integrating via direct http endpoints. Purescript api is not used by any exchange so far and Daedalus UI is also moving away from it. We will deprecate it really soon

akegalj:
@pva701 @adinapoli-iohk it should be ready for re-review.

I am happy with squashing these commits when merging them.

I have simplified json serilization from:
```
{
    from: "the_input_moneysource"
  , to: 
      [ {address: "addr1", amount: 30}
      , {address: "addr2", amount: 40}
      , {address: "addr3", amount: 30}
      ]
  , inputSelectionPolicy: null 
}
```
to:
```
{
    from: "the_input_moneysource"
  , to: 
      { ["addr1", 30]
      ,  ["addr2", 40]
      ,  ["addr3", 30]
      ]
  , inputSelectionPolicy: null 
}
```

After it builds on my machine I will test json serilization one more time and edit this comment to confirm (as I never really serialized json tuples - ie used `FromJSON a, FromJSON b => FromJSON (a,b)` instance)

adinapoli-iohk:
Sorry to be a pain in the neck @akegalj , but I'm not very comfortable with this serialisation scheme:

```
{
    from: "the_input_moneysource"
  , to: 
      { ["addr1", 30]
      ,  ["addr2", 40]
      ,  ["addr3", 30]
      ]
  , groupingPolicy: null 
}
```

Why are we giving up lovely and self descriptive JSON objects in favour of nasty heterogeneous lists? 😉  What am I missing?

akegalj:
> Why are we giving up lovely and self descriptive JSON objects in favour of nasty heterogeneous lists? 😉 What am I missing?
It looks simpler (in my mind TM) I am ok with list of objects though. Will revert back

adinapoli-iohk:
We are almost there, @akegalj ! Would you address the conflicts so we can this merged? :)

akegalj:
@adinapoli-iohk This should be it - waiting for CI

adinapoli-iohk:
@akegalj I have decided to squash these into one clean commit which is easy to revert, in case. Thanks Ante!

PR #1938, opened 2017-11-21
Corrections in commands (from old chapter)

denisshevchenko:


denisshevchenko:
cc @domenkozar. Sorry, I forgot to add these corrections from old chapter. :-(

PR #1939, opened 2017-11-21
[CSL-1867] Describe update system's consensus rules

pva701:
Description of US consensus rules.

### DIFF
--- a/docs/block-processing/overall.md
+++ b/docs/block-processing/overall.md
@@ -5,14 +5,13 @@
 Block processing can be described by presenting an algorithm to solve the following problem:
 
 > Given a sequence of blocks `B₀, B₁, B₂, …` 
-(where `B₀` is the first genesis block) check whether block payload from these blocks are valid.
+(where `B₀` is the first genesis block) check whether these blocks are valid.
 
 We describe block processing as stateful algorithm:
 * Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))
 * `S₁, S₂, …` are maintained as sequential application of blocks  `B₀, B₁, B₂, … ` to state `S₀`
-* State transition function. 
-We maintain some state called GState which corresponds to the application of a sequence of blocks
-Given GState `S` and a block `B` 
+* We maintain some state called GState which corresponds to the application of a sequence of blocks.
+* State transition function. Given GState `S` and a block `B` 
 return either an error describing why `B` is invalid or new GState `S'`
   ```
   verifyAndApplyGState :: GState -> Block -> Either BlockVerificationError GState
@@ -67,8 +66,7 @@ possible behaviours:
 1. Consider such block invalid.
 2. Do as many checks as we can and ignore checks which can't be done
    because data is unknown. This behaviour depends on which type of
-   data we are processing, each particular case is described in more
-   details below.
+   data we are processing.
 
 The behaviour depends on two protocol versions: version used by this
 software and last adopted version. We verify that data in blocks is
--- a/docs/block-processing/overall.md
+++ b/docs/block-processing/overall.md
@@ -1,4 +1,42 @@
-# Unknown data handling
+# Block processing
+
+## Task definition
+
+Block processing can be described by presenting an algorithm to solve the following problem:
+
+> Given a sequence of blocks `B₀, B₁, B₂, …` (where `B₀` is the first
+genesis block) check whether block payload from these blocks are valid.
+
+We describe block processing as stateful algorithm:
+* Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))
+* `S₁, S₂, …` are maintained as sequential application of blocks  `B₀, B₁, B₂, … ` to state `S₀`
+* State transition function. Given GState `S` and a main block `B` return either an error describing why `B` is invalid (w.r.t. tx payload) or new GState `S'`
+  ```
+  verifyAndApplyGState :: GState -> MainBlock -> Either BlockPayloadVerificationError GState
+  -- ^ Note, that function definition and types are different in code and are put here for reader's convenience
+  ```
+  * Note, that state transition function is defined only for main blocks. This is done because genesis blocks 
+  (as defined in Ouroboros) don't have payload.
+
+Note, that in theory payload verification can be stateless (except genesis
+state `S₀`) and be described without mentioning any additional state
+(apart from blocks themselves). But in practice it would be very
+inefficient and inconvenient.
+
+For sake of simplicity, we describe state transition function in two parts:
+* Verification: given GState `S` and a main block `B`, check whether `B` is valid (and can be applied to `S`)
+* Modification: given GState `S` and a main block `B` (successfully verified against `S`), produce `S'`
+
+### Block payload
+
+Block payload consist of:
+* `UpdatePayload`. 
+  It's checked by update system component and is described in the [Update system consensus rules](us.md).
+* `TxPayload`. 
+  It's checked by transaction processing component and is described in the [Transaction processing](txp.md).
+* TODO: others 
+
+## Unknown data handling
 
 Many types in Cardano-SL are designed in an extensible way. They can
 be extended via softfork in future (e. g. new data can be attached or
@@ -59,7 +97,7 @@ adopted version) check is not done.
 Note: when we say that attributes must be known, it means that
 unparsed fields must be empty.
 
-## verifyAllIsKnown flag
+### verifyAllIsKnown flag
 
 Let's assume that version comparison as described above.
 Let's denote flag `verifyAllIsKnown` and assign it value:
--- a/docs/block-processing/overall.md
+++ b/docs/block-processing/overall.md
@@ -4,32 +4,33 @@
 
 Block processing can be described by presenting an algorithm to solve the following problem:
 
-> Given a sequence of blocks `B₀, B₁, B₂, …` (where `B₀` is the first
-genesis block) check whether block payload from these blocks are valid.
+> Given a sequence of blocks `B₀, B₁, B₂, …` 
+(where `B₀` is the first genesis block) check whether block payload from these blocks are valid.
 
 We describe block processing as stateful algorithm:
 * Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))
 * `S₁, S₂, …` are maintained as sequential application of blocks  `B₀, B₁, B₂, … ` to state `S₀`
-* State transition function. Given GState `S` and a main block `B` return either an error describing why `B` is invalid (w.r.t. tx payload) or new GState `S'`
+* State transition function. 
+We maintain some state called GState which corresponds to the application of a sequence of blocks
+Given GState `S` and a block `B` 
+return either an error describing why `B` is invalid or new GState `S'`
   ```
-  verifyAndApplyGState :: GState -> MainBlock -> Either BlockPayloadVerificationError GState
+  verifyAndApplyGState :: GState -> Block -> Either BlockVerificationError GState
   -- ^ Note, that function definition and types are different in code and are put here for reader's convenience
   ```
-  * Note, that state transition function is defined only for main blocks. This is done because genesis blocks 
-  (as defined in Ouroboros) don't have payload.
 
-Note, that in theory payload verification can be stateless (except genesis
+Note, that in theory verification can be stateless (except genesis
 state `S₀`) and be described without mentioning any additional state
 (apart from blocks themselves). But in practice it would be very
 inefficient and inconvenient.
 
 For sake of simplicity, we describe state transition function in two parts:
-* Verification: given GState `S` and a main block `B`, check whether `B` is valid (and can be applied to `S`)
-* Modification: given GState `S` and a main block `B` (successfully verified against `S`), produce `S'`
+* Verification: given GState `S` and a block `B`, check whether `B` is valid (and can be applied to `S`)
+* Modification: given GState `S` and a block `B` (successfully verified against `S`), produce `S'`
 
 ### Block payload
 
-Block payload consist of:
+Block payload consists of:
 * `UpdatePayload`. 
   It's checked by update system component and is described in the [Update system consensus rules](us.md).
 * `TxPayload`. 
@@ -77,7 +78,8 @@ equal to the adopted version. That's because in this case:
 1. Authors of this software are aware of the adopted version.
 2. Each issued block must be formed with respect to adopted version.
 
-Comparison is quite tricky here. Table below demonstrates it.
+Comparison of software protocol version and last adopted one is quite tricky here. 
+Table below demonstrates it. The last column stands for whether we check data in block is known.
 
 | Our   | Adopted | Check? |
 | ----- | ------- | ------ |
@@ -99,7 +101,7 @@ unparsed fields must be empty.
 
 ### verifyAllIsKnown flag
 
-Let's assume that version comparison as described above.
-Let's denote flag `verifyAllIsKnown` and assign it value:
+Let's assume that version comparison is done as described above.
+Let's define flag `verifyAllIsKnown` and assign it value:
 * `True` if check is to be done, i.e. only known data should be considered valid (unknown data is prohibitied)
 * `False` otherwise
--- a/docs/block-processing/us.md
+++ b/docs/block-processing/us.md
@@ -10,6 +10,7 @@
     + [Update proposal states](#update-proposal-states)
     + [Software and block versions](#software-and-block-versions)
     + [Adoption of block version](#adoption-of-block-version)
+      - [Competing block version](#competing-block-version)
       - [Block validation according to adopted version](#block-validation-according-to-adopted-version)
       - [Example](#example)
  * [Verification](#verification)
@@ -45,7 +46,7 @@ which checks validity of update payload of blocks and update corresponding state
 
 ### Update payload
 
-Update payload consists of at most one `UpdatePayload` and list of votes, it reflects the datatype:
+Update payload consists of at most one `UpdatePayload` and list of votes, it's reflected by the datatype:
 ```
 data UpdatePayload = UpdatePayload
     { upProposal :: !(Maybe UpdateProposal)
@@ -160,20 +161,29 @@ if portion of block issuers' stakes, which issued blocks of this `BlockVersion` 
 Intuitively, the threshold which needed to adopt `BlockVersion` is decreasing in each epoch but cannot become
 less than some reasonable minimal value (`spMinThd`).
 
-There is datatype which can also be updated by `BlockVersionModifier`:
-```
-data SoftforkRule = SoftforkRule
-    { srInitThd      :: !CoinPortion
-    , srMinThd       :: !CoinPortion
-    , srThdDecrement :: !CoinPortion
-    }
-```
-
 So we check this rule at the beginning of each epoch for each _confirmed_ proposal 
 and adopt one of block version from _confirmed_ proposals if this version satisfies the rule.
 
-By the way, we call block version **competing** if it might become adopted. 
-It implies there is at least one confirmed proposal with such block version.
+
+#### Competing block version
+Block version is called **competing** if it may become adopted and 
+there is a confirmed proposal with this block version.
+There are rules defining whether block version may become adopted after current last adopted version.  
+The following rules regarding major and minor versions take place:
+  * The proposed major version must be equal to or greater by `1` last adopted one.
+  * If the proposed major version is greater than last adopted one by `1`, then minor version must be `0`.
+  * If the proposed major version is equal to the last adopted one, then minor version 
+    can be either same as the last adopted one or greater by `1`.
+
+Also the following rules regarding alternative versions take place (assuming checks above passed):
+  * If `(Major, Minor)` of proposed version is equal to `(Major, Minor)` of
+    last adopted version, then alternative version must be equal to alternative version of last adopted version.
+  * Otherwise `(Major, Minor)` of proposed version is lexicographically greater 
+    than `(Major, Minor)` of last adopted version. In this case
+    other proposed block versions with same `(Major, Minor)` are considered. 
+    Let's call set of these block version `X`:
+    * If `X` is empty, given alternative version must be `0`.
+    * Otherwise it must be in `X` or greater than maximum from `X` by `1`.
 
 #### Block validation according to adopted version
 
@@ -187,7 +197,7 @@ others stakeholders will validate blocks according to the current _adopted_ bloc
 
 #### Example
 
-Assume there are four stakeholders with stakes `[0.25, 0.26, 0.1, 0.39]`,
+Assume there are four stakeholders with stakes `[0.25, 0.26, 0.39, 0.1]`,
 `lastKnownBlockVersion` is `0.1.0` for all stakeholders' softwares and 
 last adopted block version is also `0.1.0`.
 
@@ -211,22 +221,22 @@ So the block is issued without new attribute but tagged by the new version.
 Stakeholders with old software expects that all attributes are known because
 last adopted block version equals `lastKnownBlockVersion` for its softwares.
 
-Then the 3rd epoch starts and the threshold (which is regulated by `SoftforkRule`) 
-for adopting new block version is dumped to `0.6`.
+Then the 3rd epoch starts and the threshold (which is regulated by genesis `SoftforkRule`) 
+for adopting new block version is `0.85`.
 
 Then the second stakeholder downloads this update and when it's turn to issue block, 
 the stakeholder issues block like it has been done by the first stakeholder.
 Also the second one issues one more block as leader, so it produced 2 blocks.
 
-Then 4th epoch starts and sum of issuers stakes is `0.25 + 0.26 = 0.51 < 0.6` 
-(we take into account the second stakeholder only once). Adopting threshold is dumped to the minimal value `0.6`.
+Then 4th epoch starts, adopting threshold is `0.80` and sum of issuers stakes is `0.25 + 0.26 = 0.51 < 0.80` 
+(we take into account the second stakeholder only once). 
 So last adopted block version is still `0.1.0`, `lastKnownBlockVersion` is `0.2.0` for 
 the first and the second stakeholders and `0.1.0` for the third and the fourth.
 
 Then the third stakeholder downloads this update and issues block in the same manner like previous ones.
 
-When 5th epoch starts the sum of issuers stakes is `0.25 + 0.26 + 0.1 = 0.61 > 0.6` and 
-it's time to bump last adopted block version.
+When 5th epoch starts, adopting threshold is `0.75` and
+sum of issuers stakes is `0.25 + 0.26 + 0.39 = 0.9 > 0.75` and it's time to bump last adopted block version.
 So last adoped version is `0.2.0` now and `lastKnownBlockVersion` is `0.1.0` 
 for the latter stakeholder and `0.2.0` for others.
 
@@ -250,7 +260,7 @@ First of all, we check that block version of a verifiable block is _adopted_ or 
 * _No duplicated proposer check_: that no one stakeholder sent two update proposals within current epoch. 
 We prohibit it because of opportunity spam update proposals from one stakeholder.
 
-* _Update proposal size check_: proposal is not exceed maximal proposal size.
+* _Update proposal size check_: proposal doesn't exceed maximal proposal size.
   * Update proposal size is computed as number of bytes in serialized `UpdateProposal`.
   * Update proposal size limit is part of `BlockVersionData`.
   * If the size of a proposal is greater than the limit from the adopted `BlockVersionData`, the proposal is invalid.
@@ -270,28 +280,17 @@ So everyone can send update proposal to network, but it can get into block only 
 
 #### Protocol version checks
 
-* _Protocol version following check_: protocol version from proposal can follow last adopted protocol version.  
-The following rules regarding major and minor versions take place:
-  * The proposed major version must be equal to or greater by `1` last adopted one.
-  * If the proposed major version is greater than last adopted one by `1`, then minor version must be `0`.
-  * If the proposed major version is equal to the last adopted one, then minor version 
-can be either same as the last adopted one or greater by `1`.
-
-  Also the following rules regarding alternative versions take place (assuming checks above passed):
-  * If `(Major, Minor)` of proposed version is equal to `(Major, Minor)` of
-last adopted version, then alternative version must be equal to alternative version of last adopted version.
-  * Otherwise `(Major, Minor)` of proposed version is lexicographically greater 
-than `(Major, Minor)` of last adopted version and in this case
-other proposed block versions with same `(Major, Minor)` are considered (let's call this set `X`):
-    * If `X` is empty, given alternative version must be `0`.
-    * Otherwise it must be in `X` or greater than maximum from `X` by `1`.
+* _Protocol version following check_: 
+proposed block version must adhere to the rules described [above](#competing-block-version).
 
 * _BlockVersionModifier consistenty check_: `BlockVersionModifier` from proposal is consistent 
-with `BlockVersionData` for adopted protocol version or with `BlockVersionModifier` from _competing_ proposals.  
+with `BlockVersionData` for adopted protocol version or 
+with `BlockVersionModifier` from _competing_ proposals with the same protocol version.  
 There are three cases to check:
   * Block version from proposal is equal to adopted version. 
     In this case each field of `BlockVersionModifier` from proposal must be either `Nothing` or 
-    `Just` with the same value from `BlockVersionData` of last adopted version.
+    `Just` with the same value from `BlockVersionData` of last adopted version, 
+    i.e. we forbid changing any protocol constants. 
   * Block version from proposal is equal to one from another proposal. 
     `BlockVersionModifier`s from them must be the same. 
      If several proposals with the same block version exist, then all `BlockVersionModifier` must be equal.
@@ -338,7 +337,7 @@ On each main block the following algorithm is performed
     So at the beginning of sorted list,  _approved_ proposals will follow, if there is more than one,
     approved explicitly and so on.
 4. All proposals in each group except a head become _discarded_.
-5. Head proposal of each group become _confirmed_ if it's _approved_.
+5. Head proposal of each group become _confirmed_ if it's _approved_ and _discarded_ if it's _rejected_.
 
 ### Block version adoption algorithm
 
@@ -348,9 +347,9 @@ On each genesis block the following algorithm is performed:
 1. Compute threshold for current epoch for adoption of block version, let's call it `T`.
 2. For each competing block version do the following steps:
     1. Compute summary stake of stakeholder which issued a block of this block version and 
-        corresponding to slots no greater than `8k`. 
-        Such blocks can't be rolled back, hence, summary stake can't be changed.  
-        Notice, if a stakeholder issued two or more blocks with this block version, it's counted only once.
+       corresponding to slots no greater than `8k` of the previous epoch. 
+       Such blocks can't be rolled back, hence, summary stake can't be changed.  
+       Notice, if a stakeholder issued two or more blocks with this block version, it's counted only once.
     2. If summary stake is greater than total stake multiple by `T` then this block version can be adopted.
 3. Take all block versions from the previous step which can be adopted, let's call them `S`.  
    If `S` is empty then there is nothing to adopt, otherwise take maximal block version from `S`, let's call it `V`.
--- a/docs/block-processing/us.md
+++ b/docs/block-processing/us.md
@@ -241,11 +241,7 @@ and updates should be handled carefully.
 
 ### Block header verification
 
-First of all, we check that one of statements for a verifiable block is true:
-* block version equals to last adopted block version. 
-It means the block is created by software where `lastKnownBlockVersion` equals last adopted block version.
-* block version is greater than last adopted block version and this one is confirmed. 
-It means the block is created by software updated by _confirmed_ update proposal.
+First of all, we check that block version of a verifiable block is _adopted_ or _competing_.
 
 ### Update proposal verification
 
--- a/docs/block-processing/us.md
+++ b/docs/block-processing/us.md
@@ -40,7 +40,8 @@ ability to users to vote for updates to decide which one will be accepted.
 
 This document describes a global part of update system consensus rules.
 Global consensus rules is a part of block processing, 
-which checks validity of update payload of blocks and update corresponding state when blocks are applied/rolled back. 
+which checks validity of update payload of block, verify main block header and 
+update corresponding state when blocks are applied/rolled back. 
 
 ## Update system model
 
@@ -193,7 +194,7 @@ haven't updated its software yet, hence,
 they can't validate blocks of block version from the proposal.
 
 To avoid a situation when adversary stakeholder issues invalid block of _competing_ version
-others stakeholders will validate blocks according to the current _adopted_ block version.
+other stakeholders will validate blocks according to the current _adopted_ block version.
 
 #### Example
 
@@ -251,14 +252,14 @@ and updates should be handled carefully.
 
 ### Block header verification
 
-First of all, we check that block version of a verifiable block is _adopted_ or _competing_.
+First of all, we check that block version of a verifiable block is _adopted_ or [competing](#competing-block-version).
 
 ### Update proposal verification
 
 #### General checks
 
 * _No duplicated proposer check_: that no one stakeholder sent two update proposals within current epoch. 
-We prohibit it because of opportunity spam update proposals from one stakeholder.
+We prohibit it because of opportunity to spam update proposals from one stakeholder.
 
 * _Update proposal size check_: proposal doesn't exceed maximal proposal size.
   * Update proposal size is computed as number of bytes in serialized `UpdateProposal`.
@@ -271,7 +272,8 @@ is `True`, all update proposal attributes must be known.
 * _Uniqueness of proposal check_: that there is no active proposal with the same id.
 
 * _Software version check_: numeric software version of application is 1 more 
-than of last confirmed proposal for this application.
+than of last confirmed proposal for this application. 
+Numeric software version should be zero for a new application name.
 
 * _Stake to include into block check_: sum of positive votes for this proposal is enough 
 to be included into block (at least `updateProposalThd`).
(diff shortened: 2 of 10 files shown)


docs/us.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Poll and decision agreement rules
+
+Each stakeholder can send update proposal in the blockchain,
+and others stakeholders can send their votes "for" or "against" it.
+Making decision whether approve or reject an update proposal depends on sum of stake "for" and "against" it.
+
+There are two rules to make decision about update proposal:
+* **Explicit agreement**: if an update proposal has greater than 50% stake "for" or "against" 
+then the proposal becomes approved or rejected corresponding.
```
gromakovsky: s/corresponding/respectively

docs/us.md
```diff
@@ -0,0 +1,225 @@
...
+We first describe global part and then describe how local part differs.
+
+## Problem definition
+TBA
+
+## Update system model
+
+### Poll and decision agreement rules
+
+Each stakeholder can send update proposal in the blockchain,
```
gromakovsky: I'm not sure if it's written below, but I'd already say that update proposal can appear in blockchain only if it has some votes for it from stakeholders (there is a threshold, you know).

docs/us.md
```diff
@@ -0,0 +1,225 @@
...
+
+There are two rules to make decision about update proposal:
+* **Explicit agreement**: if an update proposal has greater than 50% stake "for" or "against" 
+then the proposal becomes approved or rejected corresponding.
+* **Implicit agreement**: if an update proposal has been proposed in the blockchain `U` slots ago,
+and it has more stakes "for" then "against", then the proposal becomes approved and rejected otherwise.
+
+### Update proposal states
+
+Update proposal may be in three states:
```
gromakovsky: Well, initially we had 5 states (active, approved, rejected, confirmed, discarded), so I'd stick to this notion.
Decided is not a separate state, it's just a union of two states.

docs/us.md
```diff
@@ -0,0 +1,225 @@
...
+it wasn't still decided by stakeholders whether to approve or to reject this proposal.
+So, a poll is active and stakeholders' votes which get into the blockchain affect the decision.
+
+* **Decided**  
+An update proposal continues being _active_ until it becomes _decided_, it may be _decided_ to approve
+or to reject.
+If a proposal is rejected then it doesn't affect any consensus rules and we don't consider it anymore.
+Note: rejected proposal may become active again if rollback occurs.
+
+* **Confirmed**
```
gromakovsky: You forgot **Discarded** state. One may argue that proposal can't be in this state, because when it's discarded, it's, well, literally discarded (i. e. thrown away from the consideration), but I think it's still useful to have this notion in some cases.

docs/us.md
```diff
@@ -0,0 +1,225 @@
...
+
+Each stakeholder can send update proposal in the blockchain,
+and others stakeholders can send their votes "for" or "against" it.
+Making decision whether approve or reject an update proposal depends on sum of stake "for" and "against" it.
+
+There are two rules to make decision about update proposal:
+* **Explicit agreement**: if an update proposal has greater than 50% stake "for" or "against" 
+then the proposal becomes approved or rejected corresponding.
+* **Implicit agreement**: if an update proposal has been proposed in the blockchain `U` slots ago,
+and it has more stakes "for" then "against", then the proposal becomes approved and rejected otherwise.
```
gromakovsky: than

gromakovsky: approved **or** rejected, not **and**?

docs/us.md
```diff
@@ -0,0 +1,225 @@
...
+```
+So if a value is going to be updated then a field is `Just` and `Nothing` otherwise.
+
+### Adoption of block version
+
+Assume a proposal which is bumping `BlockVersion` became _confirmed_.
+Though the proposal is already _confirmed_ along with block version but stakeholders
+haven't updated its software yet, hence, they can't validate block of _confirmed_ block version.
+
+To avoid a situation when adversary stakeholder issues invalid block of just _confirmed_ block version,
```
gromakovsky: For block versions we use term **competing** rather than **confirmed** to avoid ambiguity of term **confirmed**.

docs/us.md
```diff
@@ -0,0 +1,225 @@
...
+others stakeholders will validate blocks according to previous **adopted** block version, 
+until _confirmed_ becomes **adopted**.
+
+Informally `BlockVersion` is **adopted** if sum of block issuers' stakes, 
+which issued blocks of this `BlockVersion` at least once, is a majority.
```
gromakovsky: Why majority? Not just majority, it depends on `SoftforkRule`, initially we want it to be 90% and after a month we want 60% (and never less than 60%).

gromakovsky: Well, maybe that's ok, because it's _informally_ and below you describe it more formally… 🤷‍♂️

pva701: mostly copied from 
https://cardanodocs.com/cardano/update-mechanism/

docs/us.md
pva701: Frankly, I am not sure it should be mentioned here, this rule is kind of just in case, to avoid an influence of bad men. I wouldn't mention it here...

docs/unknown-data.md
```diff
@@ -0,0 +1,67 @@
...
+done only if `alt` component is the same as adopted one. In
+other cases (i. e. when our `(major, minor)` is less than from
+adopted version) check is not done.
+
+Note: when we say that attributes must be known, it means that
+unparsed fields must be empty.
+
+## verifyAllIsKnown flag
+
+Let's assume that version comparison as described above
```
gromakovsky: It should be:
>  Let's assume that version comparison as described at the end of previos section (Unknown data handling) is done.

update/Pos/Update/Poll/Logic/Apply.hs
```diff
@@ -300,27 +304,32 @@ applyImplicitAgreement (flattenSlotId -> slotId) cd hh = do
...
                 NE.groupWith groupCriterion deepProposals
         mapM_ applyDepthCheckDo winners
   where
     upsAppName = svAppName . upSoftwareVersion . upsProposal
-    resetAllDecisions (a:|xs) = a :| map (\x->x {dpsDecision = False}) xs
-    groupCriterion a =
-        ( upsAppName $ dpsUndecided a
-        , dpsDifficulty a)
+    discardAllExceptHead (a:|xs) = a :| map (\x->x {dpsDecision = False}) xs
+    groupCriterion = upsAppName . dpsUndecided
```
gromakovsky: I suggest that we don't change the criterion unless you can prove that it effectively doesn't change anything (is equivalent to the old one in our case).
Also now it contradicts to the comment you wrote:
> 1. Group psoposals by (application name, difficulty when proposal became decided)

docs/us.md
```diff
@@ -0,0 +1,352 @@
...
+
+#### Block validation according to adopted version
+
+Assume a proposal which is bumping `BlockVersion` became _confirmed_.
+Though the proposal is already _confirmed_ but stakeholders
+haven't updated its software yet, hence, 
+they can't validate blocks of block version from the proposal.
+
+To avoid a situation when adversary stakeholder issues invalid block of _competing_ version
+others stakeholders will validate blocks according to the previous _adopted_ block version.
```
gromakovsky: Why previous? Current, no?

docs/us.md
```diff
@@ -0,0 +1,352 @@
...
+
+Then the third stakeholder downloads this update and issues block in the same manner like previous ones.
+
+When 5th epoch starts the sum of issuers stakes is `0.25 + 0.26 + 0.1 = 0.61 > 0.6` and 
+it's time to bump last adopted block version.
+So last adoped version is `0.2.0` now and `lastKnownBlockVersion` is `0.1.0` 
+for the latter stakeholder and `0.2.0` for others.
+
+When it's turn to issue block of the fourth stakeholder, 
+it issues block with `0.1.0` block version and with block format for `0.1.0`.
```
gromakovsky: No, in our implementation this stakeholder won't even try to issue a block.

docs/us.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Verification
+
+### Block header verification
+
+First of all, we check that one of statements for a verifiable block is true:
+* block version equals to last adopted block version. 
+It means the block is created by software where `lastKnownBlockVersion` equals last adopted block version.
+* block version is greater than last adopted block version and this one is confirmed. 
+It means the block is created by software updated by _confirmed_ update proposal.
```
gromakovsky: I guess it can be simplified to say that block version should be either confirmed or competing. And the code can be simplified as well.

docs/us.md
```diff
@@ -0,0 +1,352 @@
...
+ 
+* _Uniqueness of proposal check_: that there is no active proposal with the same id.
+
+* _Software version check_: numeric software version of application is 1 more 
+than of last confirmed proposal for this application.
+
+* _Stake to include into block check_: sum of positive votes for this proposal is enough 
+to be included into block (at least `updateProposalThd`).
+We wouldn't like adversary stakeholders to have opportunity to spam network by `UpdateProposal`, 
+so only stakeholders with some significant amount of stake can send proposals.
```
gromakovsky: Everyone can send them, but they can be included only along with enough votes.

docs/us.md
```diff
@@ -0,0 +1,352 @@
...
+  * Block version from proposal is new one. 
+    In this case we must check `BlockVersionModifier` from the proposal is consistent with adopted `BlockVersionData`.  
+    The following checks take place:
+    * script version from proposal must be equal to or greater by `1` adopted one.
+    * max block size from proposal is at most two times greater than adopted one.
+    * unlock stake epoch must not be changed if boostrap era is over
+
+### Votes verification
+
+For each vote the following checks must be performed:
```
gromakovsky: Where is the stake check?

docs/unknown-data.md
pva701: there is no previous section

docs/us.md
pva701: how to check it? and why shouldn't we check that it is last adopted?

update/Pos/Update/Poll/Logic/Apply.hs
pva701: as you said in slack: there is one difficulty because otherwise, we processed it on the previous block.

pva701:
>I want to notice that you should also describe checks which are part of internal invariants of the involved types (e. g. see mkUpdateProposal).

:cho: how is it related to consensus rules? you can notice I didn't write the `UpdateProposal` type here and related, internal invariants forces to do it. and again, this docs doesn't describe everything about US just consensus rules

docs/us.md
pva701: done

docs/unknown-data.md
gromakovsky: Ok, but where is `is done` part? The second part of the sentence doesn't have a verb.

docs/us.md
gromakovsky: Sorry, I wrote garbage. Either competing or adopted.

gromakovsky: We have this:
```
-- | Check whether BlockVersion is confirmed.
isConfirmedBV :: MonadPollRead m => BlockVersion -> m Bool
isConfirmedBV = fmap (maybe False bvsIsConfirmed) . getBVState
```

We should rename all usages of `confirmed` to `competing` when we talk about block versions.
Then we have this:
```
-- | Check whether block with given 'BlockVersion' can be created
-- according to current Poll.
--
-- Specifically, one of the following conditions must be true.
-- • Given block version is equal to last adopted block version.
-- • '(major, minor)' of given block version is greater than
-- '(major, minor)' of adopted block version and this block version
-- is confirmed.
canCreateBlockBV :: MonadPollRead m => BlockVersion -> m Bool
canCreateBlockBV bv = do
    lastAdopted <- getAdoptedBV
    isConfirmed <- isConfirmedBV bv
    let toMajMin BlockVersion {..} = (bvMajor, bvMinor)
    return
        (bv == lastAdopted ||
         (toMajMin bv > toMajMin lastAdopted && isConfirmed))
```
The check `toMajMin bv > toMajMin lastAdopted` is likely useless, because if the version is competing, its `(Maj, Min)` must be greater than the last adopted one. Do you agree?

update/Pos/Update/Poll/Logic/Apply.hs
gromakovsky: I wasn't 100% sure about that, but if you agree (thoughtfully, not just trust me), then ok, let's simplify it (and it looks more correct anyway).

gromakovsky:
> :cho: how is it related to consensus rules? you can notice I didn't define the UpdateProposal type here and related, internal invariants force to do it. and again, this docs doesn't describe everything about US just consensus rules

Because if a block contains a proposal with invalid signature, this block is invalid. This rule is not described anywhere. If you know better place for it, let me know, but it must be described somewhere (otherwise people will think we are silly and forgot this check).

pva701:
Why doesn't 'txp.md' describe similar invariants? These invariants is more related to block-processing rather than consensus rules. I would describe them into `block-processing/overall.md`

gromakovsky:
There is one https://github.com/input-output-hk/cardano-sl/blob/a43e18d58f345e302729f0b0c5491ab97946b723/docs/txp.md#general-checks

```
Also there is a check which applies to whole TxPayload:

    Number of transactions must be the same as number of TxWitnesses.
```

But I've forgotten about invariants of `Tx` (also we should mention invariant of `Coin` somewhere).

pva701:
Isn't better to introduce new doc with invariants description?

update/Pos/Update/Poll/Logic/Apply.hs
pva701: I had understood when was writing this code, but wrote in just in case

docs/us.md
pva701: I agree, will fix now.

pva701: done

gromakovsky:
> Isn't better to introduce new doc with invariants description?

I think it would be good to have a document where all types are described along with their invariants.
Probably it's ok to do it in another issue.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+
+Block processing can be described by presenting an algorithm to solve the following problem:
+
+> Given a sequence of blocks `B₀, B₁, B₂, …` (where `B₀` is the first
+genesis block) check whether block payload from these blocks are valid.
+
+We describe block processing as stateful algorithm:
+* Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))
+* `S₁, S₂, …` are maintained as sequential application of blocks  `B₀, B₁, B₂, … ` to state `S₀`
+* State transition function. Given GState `S` and a main block `B` return either an error describing why `B` is invalid (w.r.t. tx payload) or new GState `S'`
```
gromakovsky: > a main block `B`

Not necessarily main block. We are talking about overall block processing, not txp.

> (w.r.t. tx payload)

Not only tx payload.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+  ```
+  verifyAndApplyGState :: GState -> MainBlock -> Either BlockPayloadVerificationError GState
```
gromakovsky: > BlockPayloadVerificationError

We want to check not only payload, but full block. Can you remove `Payload` (`BlockVerificationError`)?

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+  -- ^ Note, that function definition and types are different in code and are put here for reader's convenience
+  ```
+  * Note, that state transition function is defined only for main blocks. This is done because genesis blocks 
+  (as defined in Ouroboros) don't have payload.
+
+Note, that in theory payload verification can be stateless (except genesis
```
gromakovsky: block verification

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+state `S₀`) and be described without mentioning any additional state
+(apart from blocks themselves). But in practice it would be very
+inefficient and inconvenient.
+
+For sake of simplicity, we describe state transition function in two parts:
+* Verification: given GState `S` and a main block `B`, check whether `B` is valid (and can be applied to `S`)
```
gromakovsky: Not only main block. Even now (when we cover only txp and US) genesis blocks are important, because US has a lot of logic when processing genesis block.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+* Modification: given GState `S` and a main block `B` (successfully verified against `S`), produce `S'`
+
+### Block payload
+
+Block payload consist of:
```
gromakovsky: Consists.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+done only if `alt` component is the same as adopted one. In
+other cases (i. e. when our `(major, minor)` is less than from
+adopted version) check is not done.
+
+Note: when we say that attributes must be known, it means that
+unparsed fields must be empty.
+
+### verifyAllIsKnown flag
+
+Let's assume that version comparison as described above.
```
gromakovsky: … is done (where is verb)?

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+which issued blocks of this `BlockVersion` at least once, takes a significant part of the stake.
+
+Formally, let's say a proposal became _confirmed_ in `s` epoch and current epoch is `t`:  
+if portion of block issuers' stakes, which issued blocks of this `BlockVersion` at least once, is greater than
+`max spMinThd (spInitThd - (t - s) * spThdDecrement)`, then proposal's `BlockVersion` becomes **adopted**.
+
+Intuitively, the threshold which needed to adopt `BlockVersion` is decreasing in each epoch but cannot become
+less than some reasonable minimal value (`spMinThd`).
+
+There is datatype which can also be updated by `BlockVersionModifier`:
```
gromakovsky: It may sound a bit confusing, because one may think that one can update the **data type**, not the value of this type. Maybe it's just me 🤷‍♂️

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+The last adopted version is `0.1.0`, hence, the first stakeholder issues a block:
+* `BlockVersion` of this block is set to `0.2.0`
+* block created by rules for last adopted block version, i.e. `0.1.0`. 
+
+So the block is issued without new attribute but tagged by the new version.
+Stakeholders with old software expects that all attributes are known because
+last adopted block version equals `lastKnownBlockVersion` for its softwares.
+
+Then the 3rd epoch starts and the threshold (which is regulated by `SoftforkRule`) 
+for adopting new block version is dumped to `0.6`.
```
gromakovsky: I suggest that in this example we use values from mainnet, where we have initial threshold 0.9, minimal threshold 0.6 and decrement 0.05.

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+than of last confirmed proposal for this application.
+
+* _Stake to include into block check_: sum of positive votes for this proposal is enough 
+to be included into block (at least `updateProposalThd`).
+We wouldn't like adversary stakeholders to have opportunity to spam blockchain by `UpdateProposal`.
+So everyone can send update proposal to network, but it can get into block only along with enough votes.
+
+#### Protocol version checks
+
+* _Protocol version following check_: protocol version from proposal can follow last adopted protocol version.
```
gromakovsky: I think it's better to move these rules above, to the place where you define **competing** block version. The word **competing** is used few times above, but the reader will know its actual meaning only after these lines.

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+1. Take all proposals which were decided at least `k` block ago.  
+   It may be determined by chain difficulty.
+2. Group proposals by application name.
+3. Sort each group using the following rule:
+  * Construct tuple  
+    (decision, whether decision is implicit, positive stake, slot when it has been proposed)  
+    by each proposal. So a tuple has type `(Bool, Bool, Coin, SlotId)`
+  * Compare in decreasing order.  
+    So at the beginning of sorted list,  _approved_ proposals will follow, if there is more than one,
+    approved explicitly and so on.
```
gromakovsky: I am wondering if it's theoretically possible that all 4 values in tuple will be the same 🤔

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+4. All proposals in each group except a head become _discarded_.
+5. Head proposal of each group become _confirmed_ if it's _approved_.
```
gromakovsky: And it also becomes discarded if it was rejected, right?

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Block version adoption algorithm
+
+In this section we describe block version adoption algorithm more precisely.
+
+On each genesis block the following algorithm is performed:
+1. Compute threshold for current epoch for adoption of block version, let's call it `T`.
+2. For each competing block version do the following steps:
+    1. Compute summary stake of stakeholder which issued a block of this block version and 
+        corresponding to slots no greater than `8k`.
```
gromakovsky: Slot is determined by two numbers (epoch and local slot). Please make it clear which epoch we're talking about.

update/Pos/Update/Poll/Logic/Base.hs
```diff
@@ -65,33 +65,28 @@ import           Pos.Util.Util (leftToPanic)
 ----------------------------------------------------------------------------
 
 -- | Check whether BlockVersion is confirmed.
```
volhovm: Update comment also?

update/Pos/Update/Poll/Logic/Apply.hs
```diff
@@ -300,27 +304,32 @@ applyImplicitAgreement (flattenSlotId -> slotId) cd hh = do
...
+    :: forall m . ApplyMode m
     => EpochIndex -> HeaderHash -> ChainDifficulty -> m ()
 applyDepthCheck epoch hh (ChainDifficulty cd)
     | cd <= blkSecurityParam = pass
     | otherwise = do
         deepProposals <- getDeepProposals (ChainDifficulty (cd - blkSecurityParam))
+        -- 1. Group psoposals by application name
+        -- 2. Sort proposals in each group by tuple
+        --     (decision, whether decision is implicit, positive stake, slot when it has been proposed)
+        -- 3. All proposals in each group except the head we make discarded
```
volhovm: make discarded == discard.

docs/block-processing/overall.md
volhovm: "GState" is undefined here. Either define it or use "state".

docs/block-processing/overall.md
volhovm: Also `Block`, not `MainBlock`.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+
+We describe block processing as stateful algorithm:
+* Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))
+* `S₁, S₂, …` are maintained as sequential application of blocks  `B₀, B₁, B₂, … ` to state `S₀`
+* State transition function. Given GState `S` and a main block `B` return either an error describing why `B` is invalid (w.r.t. tx payload) or new GState `S'`
+  ```
+  verifyAndApplyGState :: GState -> MainBlock -> Either BlockPayloadVerificationError GState
+  -- ^ Note, that function definition and types are different in code and are put here for reader's convenience
+  ```
+  * Note, that state transition function is defined only for main blocks. This is done because genesis blocks
```
volhovm: This doesn't seem to be true -- genesis blocks indeed do not have a "payload', but they have leaders list and we verify/apply them in the same way we do it with main blocks.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+
+The behaviour depends on two protocol versions: version used by this
+software and last adopted version. We verify that data in blocks is
+known if protocol version used by this software is greater than or
+equal to the adopted version. That's because in this case:
+
+1. Authors of this software are aware of the adopted version.
+2. Each issued block must be formed with respect to adopted version.
+
+Comparison is quite tricky here. Table below demonstrates it.
```
volhovm: Comparison of what? Can you put more verbose description here? It's not very clear what is "check" and what is being demonstrated.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,105 @@
...
+other cases (i. e. when our `(major, minor)` is less than from
+adopted version) check is not done.
+
+Note: when we say that attributes must be known, it means that
+unparsed fields must be empty.
+
+### verifyAllIsKnown flag
+
+Let's assume that version comparison as described above.
+Let's denote flag `verifyAllIsKnown` and assign it value:
```
volhovm: Misused "denote": http://sentence.yourdictionary.com/denote

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+   It may be determined by chain difficulty.
+2. Group proposals by application name.
+3. Sort each group using the following rule:
+  * Construct tuple  
+    (decision, whether decision is implicit, positive stake, slot when it has been proposed)  
+    by each proposal. So a tuple has type `(Bool, Bool, Coin, SlotId)`
+  * Compare in decreasing order.  
+    So at the beginning of sorted list,  _approved_ proposals will follow, if there is more than one,
+    approved explicitly and so on.
+4. All proposals in each group except a head become _discarded_.
```
volhovm: I don't understand this by the way. So if you have some approved proposals in the beginning of the list, but you take the proposals:
1. Explicit first
2. Bigger stake first
3. Newest first.

What is the motivation? Why not confirm >1 proposals at once?

docs/block-processing/overall.md
pva701: This paragraph may answer your questions:
```
The behaviour depends on two protocol versions: version used by this
software and last adopted version. We verify that data in blocks is
known if protocol version used by this software is greater than or
equal to the adopted version. That's because in this case:
```

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+### Adoption of block version
+
+Informally, `BlockVersion` is **adopted** if sum of block issuers' stakes, 
+which issued blocks of this `BlockVersion` at least once, takes a significant part of the stake.
+
+Formally, let's say a proposal became _confirmed_ in `s` epoch and current epoch is `t`:  
+if portion of block issuers' stakes, which issued blocks of this `BlockVersion` at least once, is greater than
+`max spMinThd (spInitThd - (t - s) * spThdDecrement)`, then proposal's `BlockVersion` becomes **adopted**.
+
+Intuitively, the threshold which needed to adopt `BlockVersion` is decreasing in each epoch but cannot become
```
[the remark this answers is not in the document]
pva701: @gromakovsky can you explain why threshold isn't fixed?

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+
+#### Block validation according to adopted version
+
+Assume a proposal which is bumping `BlockVersion` became _confirmed_.
+Though the proposal is already _confirmed_ but stakeholders
+haven't updated its software yet, hence, 
+they can't validate blocks of block version from the proposal.
+
+To avoid a situation when adversary stakeholder issues invalid block of _competing_ version
+others stakeholders will validate blocks according to the current _adopted_ block version.
```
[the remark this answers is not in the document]
pva701: It forces nodes with new software to construct blocks using rules for the current adopted version, not for `lastKnownBlockVersion` of software, despite the fact, `lastKnowBlockVersion` is greater than last adopted.

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+
+#### General checks
+
+* _No duplicated proposer check_: that no one stakeholder sent two update proposals within current epoch. 
+We prohibit it because of opportunity spam update proposals from one stakeholder.
+
+* _Update proposal size check_: proposal is not exceed maximal proposal size.
+  * Update proposal size is computed as number of bytes in serialized `UpdateProposal`.
+  * Update proposal size limit is part of `BlockVersionData`.
+  * If the size of a proposal is greater than the limit from the adopted `BlockVersionData`, the proposal is invalid.
```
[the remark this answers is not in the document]
pva701: what? didn't get you.

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+* _Protocol version following check_: protocol version from proposal can follow last adopted protocol version.  
+The following rules regarding major and minor versions take place:
+  * The proposed major version must be equal to or greater by `1` last adopted one.
+  * If the proposed major version is greater than last adopted one by `1`, then minor version must be `0`.
+  * If the proposed major version is equal to the last adopted one, then minor version 
+can be either same as the last adopted one or greater by `1`.
+
+  Also the following rules regarding alternative versions take place (assuming checks above passed):
+  * If `(Major, Minor)` of proposed version is equal to `(Major, Minor)` of
+last adopted version, then alternative version must be equal to alternative version of last adopted version.
```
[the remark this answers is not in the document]
pva701: >I can't propose alternative version bump without changing (major,minor)?
yep, `alt` version reflects the fact there is several competing update proposal with the same `(major, minor)`. 
so if stakeholder decided to publish an update, they should take max `alt` version among all _competing_ proposals with the same `(major, minor)` version and increment it for own proposal.
and when one of these proposals with the same `(major, minor)` version is adopted, others proposals will be discarded.

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+  * Otherwise `(Major, Minor)` of proposed version is lexicographically greater 
+than `(Major, Minor)` of last adopted version and in this case
+other proposed block versions with same `(Major, Minor)` are considered (let's call this set `X`):
```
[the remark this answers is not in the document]
pva701: tbh I don't understand what don't you understand, and dunno how to clarify it. I've slightly rewritten this sentence, dunno whether it helps...

docs/block-processing/us.md
```diff
@@ -0,0 +1,359 @@
...
+
+* _BlockVersionModifier consistenty check_: `BlockVersionModifier` from proposal is consistent 
+with `BlockVersionData` for adopted protocol version or with `BlockVersionModifier` from _competing_ proposals.  
+There are three cases to check:
+  * Block version from proposal is equal to adopted version. 
+    In this case each field of `BlockVersionModifier` from proposal must be either `Nothing` or 
+    `Just` with the same value from `BlockVersionData` of last adopted version.
+  * Block version from proposal is equal to one from another proposal. 
+    `BlockVersionModifier`s from them must be the same. 
+     If several proposals with the same block version exist, then all `BlockVersionModifier` must be equal.
```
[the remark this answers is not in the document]
pva701: You should issue new proposal with new version (either bump alt version, if another protocol behaviour isn't changed in your proposal or bump `(maj, min)` version).

So this check makes sense because it's difficult to satisfy that update proposals with the same version have the same behaviour (yep, we expect it), so we try to check at least this minimal check of equality `BlockVersionModifiers`

docs/block-processing/us.md
pva701: >Why not confirm >1 proposals at once?

AFAIU we expect the linear sequence of updates, we force it by checking that software version of proposal is greater by one than current confirmed software version.
it implies that an issuer of the update is aware of which version he updates.

so we can't confirm >1 proposals at once because then next issuer of update won't be aware which version is used by each stakeholder and won't know which software he updates.

docs/block-processing/us.md
pva701: why `must`? we check that proposed protocol version can follow last adopted protocol version

docs/block-processing/overall.md
pva701: @gromakovsky It wasn't defined in initial `txp.md` :(

docs/block-processing/us.md
pva701: dunno why I mentioned `SoftforkRule` datatype here...

docs/block-processing/overall.md
gromakovsky: It was defined in my initial version https://github.com/input-output-hk/cardano-sl/blob/25e8136aa58acf4f2a53b9aa6853bde20fe75604/docs/txp.md#global-transaction-processing

docs/block-processing/us.md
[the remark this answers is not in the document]
gromakovsky: Explained here: https://serokell.slack.com/archives/G3B0JT7AL/p1511707364000009?thread_ts=1511637343.000052&cid=G3B0JT7AL
tl;dr: see CSL-750.

update/Pos/Update/Poll/Logic/Base.hs
```diff
@@ -65,33 +65,28 @@ import           Pos.Util.Util (leftToPanic)
...
-isConfirmedBV :: MonadPollRead m => BlockVersion -> m Bool
-isConfirmedBV = fmap (maybe False bvsIsConfirmed) . getBVState
+isCompetingBV :: MonadPollRead m => BlockVersion -> m Bool
+isCompetingBV = fmap (maybe False bvsIsConfirmed) . getBVState
```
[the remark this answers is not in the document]
volhovm: `Active` may become adopted as well, but `active` is not competing.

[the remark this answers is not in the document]
pva701: `Active` is the status of update proposal, but competing is block version state.

are you just talking about renaming this function?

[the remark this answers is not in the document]
gromakovsky: _Perhaps_ we should slightly modify this definition
> By the way, we call block version competing if it might become adopted.

For instance, version 100.500.228 can be adopted, but only if many different things happen. We probably shouldn't say that it implies that the corresponding proposal is confirmed, but instead make a requirement that there must be a confirmed proposal with this block version.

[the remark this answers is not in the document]
pva701: done

pva701:
I've rebased on master

docs/block-processing/overall.md
```diff
@@ -0,0 +1,107 @@
...
+Block processing can be described by presenting an algorithm to solve the following problem:
+
+> Given a sequence of blocks `B₀, B₁, B₂, …` 
+(where `B₀` is the first genesis block) check whether block payload from these blocks are valid.
+
+We describe block processing as stateful algorithm:
+* Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))
+* `S₁, S₂, …` are maintained as sequential application of blocks  `B₀, B₁, B₂, … ` to state `S₀`
+* State transition function. 
+We maintain some state called GState which corresponds to the application of a sequence of blocks
```
gromakovsky: I suggest that you move this sentence above and make it a separate bullet point (the first one). So that it appears before 
> Initial state `S₀` is derived from blockchain genesis data…

Also you forgot a dot.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,107 @@
...
+
+There are few more examples, but these three should demonstrate the
+general idea. If we encounter unknown data (unparsed fields, unknown
+tx input type, unknown address type, etc.) in a block, there are two
+possible behaviours:
+1. Consider such block invalid.
+2. Do as many checks as we can and ignore checks which can't be done
+   because data is unknown. This behaviour depends on which type of
+   data we are processing, each particular case is described in more
+   details below.
```
gromakovsky: No, it's not described in details below anymore. Probably it's better to just drop the part of this sentence after the comma.

update/Pos/Update/Poll/Logic/Apply.hs
```diff
@@ -300,27 +304,32 @@ applyImplicitAgreement (flattenSlotId -> slotId) cd hh = do
...
 -- discarded).
 applyDepthCheck
-    :: ApplyMode m
+    :: forall m . ApplyMode m
     => EpochIndex -> HeaderHash -> ChainDifficulty -> m ()
 applyDepthCheck epoch hh (ChainDifficulty cd)
     | cd <= blkSecurityParam = pass
     | otherwise = do
         deepProposals <- getDeepProposals (ChainDifficulty (cd - blkSecurityParam))
+        -- 1. Group psoposals by application name
```
gromakovsky: psoposals

update/Pos/Update/Poll/Logic/Apply.hs
```diff
@@ -300,27 +304,32 @@ applyImplicitAgreement (flattenSlotId -> slotId) cd hh = do
...
+        -- 2. Sort proposals in each group by tuple
+        --     (decision, whether decision is implicit, positive stake, slot when it has been proposed)
+        -- 3. All proposals in each group except the head we discard
+        -- 4. Concat all groups and process all proposals
```
gromakovsky: I'd write `Concatenate` instead of `concat`, you're writing comment English.

update/Pos/Update/Poll/Logic/Apply.hs
```diff
@@ -300,27 +304,32 @@ applyImplicitAgreement (flattenSlotId -> slotId) cd hh = do
...
+    :: forall m . ApplyMode m
     => EpochIndex -> HeaderHash -> ChainDifficulty -> m ()
 applyDepthCheck epoch hh (ChainDifficulty cd)
     | cd <= blkSecurityParam = pass
     | otherwise = do
         deepProposals <- getDeepProposals (ChainDifficulty (cd - blkSecurityParam))
+        -- 1. Group psoposals by application name
+        -- 2. Sort proposals in each group by tuple
+        --     (decision, whether decision is implicit, positive stake, slot when it has been proposed)
+        -- 3. All proposals in each group except the head we discard
```
gromakovsky: It doesn't look like a grammatically correct sentence. Why not
> We discard all proposals in each group except the head

?

docs/block-processing/us.md
```diff
@@ -0,0 +1,358 @@
...
+  don't confuse them.
+* [Block processing](block-processing.md).
+
+## Overview
+Update system gives ability to developers to propose updates of software applications and
+ability to users to vote for updates to decide which one will be accepted.
+
+This document describes a global part of update system consensus rules.
+Global consensus rules is a part of block processing, 
+which checks validity of update payload of blocks and update corresponding state when blocks are applied/rolled back.
```
gromakovsky: We also check header of main block (block version check).

docs/block-processing/us.md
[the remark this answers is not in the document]
gromakovsky: s/others stakeholders/other stakeholders

docs/block-processing/us.md
```diff
@@ -0,0 +1,358 @@
...
+and the software gives up issuing a block. So this slot left without a block.
+
+You can see despite the provided update system is powerful, it's also very complicated 
+and updates should be handled carefully.
+
+## Verification
+
+### Block header verification
+
+First of all, we check that block version of a verifiable block is _adopted_ or _competing_.
```
gromakovsky: I think it makes sense to refer to `Adoption of block version`.

docs/block-processing/us.md
```diff
@@ -0,0 +1,358 @@
...
+
+### Update proposal verification
+
+#### General checks
+
+* _No duplicated proposer check_: that no one stakeholder sent two update proposals within current epoch. 
+We prohibit it because of opportunity spam update proposals from one stakeholder.
```
gromakovsky: **to** spam maybe?

docs/block-processing/us.md
```diff
@@ -0,0 +1,358 @@
...
+    In this case each field of `BlockVersionModifier` from proposal must be either `Nothing` or 
+    `Just` with the same value from `BlockVersionData` of last adopted version, 
+    i.e. we forbid changing any protocol constants. 
+  * Block version from proposal is equal to one from another proposal. 
+    `BlockVersionModifier`s from them must be the same. 
+     If several proposals with the same block version exist, then all `BlockVersionModifier` must be equal.
+  * Block version from proposal is new one. 
+    In this case we must check `BlockVersionModifier` from the proposal is consistent with adopted `BlockVersionData`.  
+    The following checks take place:
+    * script version from proposal must be equal to or greater by `1` adopted one.
```
gromakovsky: than adopted one

docs/block-processing/us.md
```diff
@@ -0,0 +1,358 @@
...
+    * max block size from proposal is at most two times greater than adopted one.
+    * unlock stake epoch must not be changed if boostrap era is over
```
gromakovsky: I don't like your punctuation. Two sentences have a dot in the end, but this one doesn't. For some reason there are no capital letters after dots. It looks strange to me, though I don't know English punctuation well.

docs/block-processing/us.md
```diff
@@ -0,0 +1,358 @@
...
+2. For each competing block version do the following steps:
+    1. Compute summary stake of stakeholder which issued a block of this block version and 
+       corresponding to slots no greater than `8k` of the previous epoch. 
+       Such blocks can't be rolled back, hence, summary stake can't be changed.  
+       Notice, if a stakeholder issued two or more blocks with this block version, it's counted only once.
+    2. If summary stake is greater than total stake multiple by `T` then this block version can be adopted.
+3. Take all block versions from the previous step which can be adopted, let's call them `S`.  
+   If `S` is empty then there is nothing to adopt, otherwise take maximal block version from `S`, let's call it `V`.
+4. `V` becomes new adopted version. Other competing versions can't become adopted because they are less than `V`, 
+   so they aren't competing anymore and thrown away from the consideration.
```
gromakovsky: It's not fully correct. If 1.0.0 is competing and 0.5.5 becomes adopted, 1.0.0 is still competing.

docs/block-processing/overall.md
```diff
@@ -0,0 +1,107 @@
+# Block processing
+
+## Task definition
+
+Block processing can be described by presenting an algorithm to solve the following problem:
+
+> Given a sequence of blocks `B₀, B₁, B₂, …` 
+(where `B₀` is the first genesis block) check whether block payload from these blocks are valid.
```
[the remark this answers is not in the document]
pva701: sorry I was nevivonoy during the prev review  :(

update/Pos/Update/Poll/Logic/Apply.hs
pva701: English motherf***er, do you speak it? (c)

docs/block-processing/overall.md
gromakovsky: I meant moving it before 
> Initial state `S₀` is derived from blockchain genesis data (see [mainnet genesis JSON](https://raw.githubusercontent.com/input-output-hk/cardano-sl/e7cfb1724024e0db2f25ddd2eb8f8f17c0bc497f/node/mainnet-genesis.json))

Then it will be clear that `S₀` is actually GState.

docs/block-processing/us.md
```diff
@@ -0,0 +1,360 @@
...
+  * Update proposal size limit is part of `BlockVersionData`.
+  * If the size of a proposal is greater than the limit from the adopted `BlockVersionData`, the proposal is invalid.
+
+* _Attributes known check_: if `verifyAllIsKnown` (defined [here](overall.md#verifyallisknown-flag))
+is `True`, all update proposal attributes must be known.
+ 
+* _Uniqueness of proposal check_: that there is no active proposal with the same id.
+
+* _Software version check_: numeric software version of application is 1 more 
+than of last confirmed proposal for this application.
```
gromakovsky: Please write that it should be zero for a new application name.

pva701: done

PR #1944, opened 2017-11-21
[CSL-1819] Update `push-update-on-mainnet.md`

hithroc:
`push-update-on-mainnet.md` is updated according to the new auxx interface, as well as the rest of the documentation.

`bvm` argument in `propose-update` command is now optional for convinience.

### DIFF
--- a/auxx/src/Command/Proc.hs
+++ b/auxx/src/Command/Proc.hs
@@ -8,6 +8,7 @@ module Command.Proc
 import           Universum
 
 import           Data.Constraint (Dict (..))
+import           Data.Default (def)
 import           Data.List ((!!))
 import qualified Data.Map as Map
 import           Formatting (build, int, sformat, stext, (%))
@@ -322,7 +323,7 @@ createCommandProcs hasAuxxMode printAction mSendActions = rights . fix $ \comman
         puVoteAll <- getArg tyBool "vote-all"
         puBlockVersion <- getArg tyBlockVersion "block-version"
         puSoftwareVersion <- getArg tySoftwareVersion "software-version"
-        puBlockVersionModifier <- getArg tyBlockVersionModifier "bvm"
+        puBlockVersionModifier <- fromMaybe def <$> getArgOpt tyBlockVersionModifier "bvm"
         puUpdates <- getArgMany tyProposeUpdateSystem "update"
         pure ProposeUpdateParams{..}
     , cpExec = \params ->
--- a/docs/how-to/push-update-on-mainnet.md
+++ b/docs/how-to/push-update-on-mainnet.md
@@ -249,7 +249,7 @@ These hashes are Blake2b_256 hashes of CBOR-encoded contents of the files.
 There is a simple command to calculate the hash of an installer:
 
 ```
-stack exec -- cardano-auxx cmd --commands "hash-installer <FILEPATH>"
+./scripts/hash-installer.sh <FILEPATH>
 ```
 
 Note:
--- a/docs/how-to/push-update-on-mainnet.md
+++ b/docs/how-to/push-update-on-mainnet.md
@@ -137,7 +137,7 @@ The following variables will be used in all following commands:
 CONFIG_KEY=mainnet_dryrun_full  # Should be the same as on the running nodes!
 RELAY_PEER=<IP>:3000            # A host of any relay node.
 
-COMMONOPTS="--system-start 0 --configuration-file ../cardano-sl/node/configuration.yaml --configuration-key ${CONFIG_KEY}"
+COMMONOPTS="--system-start 0 --configuration-file ../cardano-sl/lib/configuration.yaml --configuration-key ${CONFIG_KEY} --mode=with-config"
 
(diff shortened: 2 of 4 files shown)


docs/how-to/push-update-on-mainnet.md
```diff
@@ -137,7 +137,7 @@ The following variables will be used in all following commands:
 CONFIG_KEY=mainnet_dryrun_full  # Should be the same as on the running nodes!
 RELAY_PEER=<IP>:3000            # A host of any relay node.
 
-COMMONOPTS="--system-start 0 --configuration-file ../cardano-sl/node/configuration.yaml --configuration-key ${CONFIG_KEY}"
+COMMONOPTS="--system-start 0 --configuration-file ../cardano-sl/lib/configuration.yaml --configuration-key ${CONFIG_KEY} --mode=with-config"
```
int-index: Is with-config sufficient? Don't you need with-node to keep up with the current epoch?

docs/how-to/push-update-on-mainnet.md
```diff
@@ -183,44 +183,42 @@ depends on whether you want to update `cardano-sl` or `csl-daedalus`.
...
-of _triples_ (two in `csl-daedalus` case, one for Windows and one for
-macOS). Each triple is `<system tag> <installer filename> none`. Once
-again, those are filenames in the current dir, not arbitrary
-paths. `none` stands for binary diff (this feature is not used for
-now). System tags should corresponds to `systemTag` values in
-configuration file.
+case and omitted in `cardano-sl` case. Those arguments are update proposals
+for a system. You can specify an arbitary amount of them
+(two in `csl-daedalus` case, one for Windows and one for macOS).
+They're constructed with `upd-bin` function, which takes two arguments:
```
int-index: `upd-bin` takes three arguments, two of them optional

docs/how-to/test-update-system.md
```diff
@@ -53,10 +53,10 @@ Then you will appear in `repl` mode and would need to perform few actions:
 #### Import secret keys
 
 ```
-add-key nodes2/testnet1.key primary
-add-key nodes2/testnet2.key primary
-add-key nodes2/testnet3.key primary
-add-key nodes2/testnet4.key primary
+add-key ./nodes2/testnet1.key true
```
int-index: I think using `primary:true` would be more clear here

docs/how-to/test-update-system.md
```diff
@@ -139,19 +149,19 @@ Wait for 30-60 seconds.
 
 Execute
 ```
-vote 1 y <upId>
-vote 2 y <upId>
-vote 3 y <upId>
+vote 1 true <upId>
```
int-index: `agree:true` would be less cryptic

docs/how-to/push-update-on-mainnet.md
hithroc: I know, I thought it is better to not overcomplicate things and not mention argument that is not actually used at the moment (according to the previous version of the documentation).

docs/how-to/push-update-on-mainnet.md
hithroc: All previous examples do not have `--node-enabled` flag, so yes, `with-config` should be sufficient.

int-index: At the time of writing the old version of the document I am not sure if `--node-enabled` existed. I'd like @gromakovsky's opinion on whether `with-node` would be better for testing updates (I think yes, because then you have up-to-date blockchain)

auxx/src/Command/Proc.hs
```diff
@@ -322,7 +323,7 @@ createCommandProcs hasAuxxMode printAction mSendActions = rights . fix $ \comman
         puVoteAll <- getArg tyBool "vote-all"
         puBlockVersion <- getArg tyBlockVersion "block-version"
         puSoftwareVersion <- getArg tySoftwareVersion "software-version"
-        puBlockVersionModifier <- getArg tyBlockVersionModifier "bvm"
+        puBlockVersionModifier <- fromMaybe def <$> getArgOpt tyBlockVersionModifier "bvm"
```
int-index: :+1:

docs/how-to/push-update-on-mainnet.md
gromakovsky: The goal of this issue is to fix outdated parts of some documents. It should be discussed with DevOps whether we want to use `with-node` mode. Currently there are no benefits in doing it, because versions should be provided manually anyway. We can implement better interface which will set proper versions automatically (in `with-node` mode), but it will require keeping auxx running which might be not convenient.

docs/how-to/push-update-on-mainnet.md
```diff
@@ -251,7 +249,7 @@ These hashes are Blake2b_256 hashes of CBOR-encoded contents of the files.
 There is a simple command to calculate the hash of an installer:
 
 ```
-stack exec -- cardano-auxx $COMMONOPTS $AUXXOPTS cmd --commands "hash-installer <FILEPATH>"
+stack exec -- cardano-auxx cmd --commands "hash-installer <FILEPATH>"
```
gromakovsky: Why not `scripts/hash-installer.sh`?

PR #1945, opened 2017-11-21
CSL-1799 Move Pos.Crypto into separate package 

gromakovsky:
`Pos.Crypto.*` modules are now in `cardano-sl-crypto` package. The only exception is `Pos.Crypto.HDDiscovery`, which is still in `cardano-sl` lib. I don't know anything about this module, I see it uses utxo, so perhaps it shouldn't be in `Pos.Crypto` namespace (which is supposed to be sorta low-level), but I decided not to touch it.

The only difficulty was that some crypto modules depended on `Pos.Core.Configuration.ProtocolConstants`. I resolved it by moving `ProtocolMagic` newtype into `Pos.Crypto.Configuration` and introducing `HasCryptoConfiguration` constraint.


gromakovsky:
Note: I want it to be merged after #1929.

PR #1946, opened 2017-11-21
[CSL-1192] Extract delegation package

Pastafarianist:
This PR introduces a new package called `cardano-sl-delegation`.

Summary of changes:

* All modules from `Pos.Delegation[.*]` were moved from `lib` into `delegation`
* A new module called `Pos.Delegation.RichmenComponent` was added to `delegation`. This module mirrors similar modules from `ssc` and `update`.
* `Pos.Delegation.Logic.Common` was refactored. `invalidateProxyCaches` was moved to `f` and `mkDelegationVar` was kept in `lib` in `Pos.Delegation.HelpersLib`.
* `DlgBlock` and `DlgBlund` were introduced (and will probably be removed in CSL-1156).
* `DlgConfiguration` was introduced.

Other notes:

* Tests remain in `lib`.


cleverca22:
`ERROR: you need to (run ./pkgs/generate.sh or apply the patch in https://gist.github.com/64bac5ecb23affd698ee85a47d06d401) and commit the changes`
(from the travis stack2nix job)

lib/cardano-sl.cabal
```diff
@@ -97,20 +92,18 @@ library
                         Pos.Web.Server
                         Pos.Web.Types
 
+                        Pos.Delegation.HelpersLib
```
gromakovsky: Why is it here and not near `Pos.Delegation.Listeners` and `Pos.Delegation.Worker`?

PR #1950, opened 2017-11-22
Merge cardano-sl-1.0 to master

gromakovsky:


lib/configuration.yaml
```diff
@@ -14843,7 +14843,7 @@ mainnet_dryrun_wallet_win64: &mainnet_dryrun_wallet_win64
   update:
     <<:
     applicationName: csl-daedalus
-    applicationVersion: 5
+    applicationVersion: 4
```
volhovm: Is it legal? What is the motivation of this change?

volhovm: This looks suspicious, though i see it's `mainnet_dryrun_*` cluster.

PR #1953, opened 2017-11-23
[CS-99] Added 3-level / n-level HD address and key derivation.

et4te:
Summary:

* Adds deriveLvl3KeyPair function for deriving an address from 3 path elements.
* Adds deriveHDAddressKeyPair function for deriving addresses with arbitrary path lengths.

This pull request was previously closed and reset due to author / signature issues.

### DIFF
--- a/core/Pos/Core/Address.hs
+++ b/core/Pos/Core/Address.hs
@@ -21,6 +21,7 @@ module Pos.Core.Address
        , addrSpendingDataToType
        , addrAttributesUnwrapped
        , deriveLvl2KeyPair
+       , deriveLvl3KeyPair
        , deriveFirstHDAddress
 
        -- * Pattern-matching helpers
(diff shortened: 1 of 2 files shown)


core/Pos/Core/Address.hs
```diff
@@ -346,6 +347,46 @@ deriveLvl2KeyPair ibea scp passphrase wsKey accountIndex addressIndex = do
...
 
+-- | Derives a HD wallet address and private key from a root key and a path
+-- which is 3 levels deep (i, j, k).
+deriveLvl3KeyPair
+    :: Bi Address'
+    => IsBootstrapEraAddr
+    -> ShouldCheckPassphrase
+    -> PassPhrase
+    -> EncryptedSecretKey
+    -> (Word32, Word32, Word32)
```
adinapoli-iohk: Just a thought: was about adding a comment specifying what those three `Word32` are about? Ideally eventually we will type-synonym those to something like `DerivationIndex`, but for now what about:

```
-> (Word32, Word32, Word32)
-- ^ (account index, chain index, address index)
```

The name `chain index` was stolen from [here](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#master-key-generation). Maybe that's not a good idea, but I thought speaking as much as possible the BIP-32 vocabulary could be a good thing 😉

core/Pos/Core/Address.hs
```diff
@@ -346,6 +347,46 @@ deriveLvl2KeyPair ibea scp passphrase wsKey accountIndex addressIndex = do
...
+    -> Maybe (Address, EncryptedSecretKey)
+deriveLvl3KeyPair isBootstrap checkPass passphrase rootKey (i, j, k) =
```
adinapoli-iohk: What about renaming those variables into something more domain-specific, like `(accountIndex, chainIndex, addressIndex)`?

core/Pos/Core/Address.hs
```diff
@@ -346,6 +347,46 @@ deriveLvl2KeyPair ibea scp passphrase wsKey accountIndex addressIndex = do
     -- We don't need to check passphrase twice
     createHDAddressH ibea (ShouldCheckPassphrase False) passphrase hdPass wKey [accountIndex] addressIndex
```
adinapoli-iohk: This is just a thought, but if I squint my eyes I can see similarities between `createHDAddressH` and your `deriveHDAddressKeyPair`. With a bit of refactoring, in theory, we should be able to generalise `createHDAddressH` to simply create addresses of arbitrary depth, and then our `deriveLvl2KeyPair` and `deriveLvl3KeyPair` might become trivial functions which offload everything to `deriveHDAddressH`.

Do you think I'm being mental here? 😀

et4te: Totally agree here. I was a little worried to make too many changes to others code but certainly deriveHDAddressKeyPair is meant to be general. I will make the changes such that the functions use the same code.

adinapoli-iohk: Thanks man! 😉

core/Pos/Core/Address.hs
et4te: Updated in latest commit.

core/Pos/Core/Address.hs
et4te: Updated in latest commit.

adinapoli-iohk:
> I don't think this is good way to solve problem. Why we're implementing some solution without preparing specification for this quite non-straightforward question of how we handle addresses (sequential/random/etc.)?

Ah, I'm not sure if you already dropped off the call on Tuesday, but this is what was agreed. We will go with 3-level addresses for the time being and research in parallel the feasibility to move everything to sequential. At that point there will be specifications etc. 😉 

> Also, why this PR only introduces library function w/o using it? I would say we shouldn't introduce unused code, any function which is unused should be a candidate for immediate removal

I hear you, but at the same time this was the only way for Edward to deliver the feature incrementally (as a matter of fact he opened another PR which consolidate his work). Do you think you can live with that if we add a comment that this is a function which will be used soon to migrate to 3-level addresses?

et4te:
> I have checked and @et4te opened #1959 which seems to be a superset of this one, so I'm also
> happy to close this one in favour of just #1959, if Edward confirms it includes everything from this one.

Hey guys, yes indeed #1959 is a superset of these changes. CSL-1916 is rebased on CS-99.

adinapoli-iohk:
In which case, @georgeee , I propose a middle-ground:

- We close this PR;
- We leave Edward work on #1959 
- Once that's ready and the design documented (I'm sure we can ask Edward to write an accurate PR description or even a markdown document with the design spec) we review and hopefully merge #1959 (for which we will of course need a more thorough review involving @dcoutts and the rest of the chaps).

Would that make you happy? 😉

adinapoli-iohk:
This is quite stale by now. Let's reopen if relevant.

PR #1957, opened 2017-11-23
[CSL-1890] Add updatePassword endpoint to V1 API

Martoon-00:
This adds implementation of `V1.updatePassword` endpoint via corresponding endpoint in `V0`.
Also some tests on password.

### DIFF
--- a/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
@@ -62,9 +62,7 @@ updatePassword
     => WalletId -> PasswordUpdate -> m Wallet
 updatePassword wid PasswordUpdate{..} = do
     wid' <- migrate wid
-    _ <- join $ V0.changeWalletPassphrase <$> migrate wid
-                                          <*> pure pwdOld
-                                          <*> pure pwdNew
+    _ <- V0.changeWalletPassphrase wid' pwdOld pwdNew
     V0.getWallet wid' >>= migrate
 
 deleteWallet :: WalletId
--- a/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
@@ -3,7 +3,7 @@ module Cardano.Wallet.API.V1.Handlers.Wallets where
 import           Universum
 
 import qualified Pos.Wallet.Web.ClientTypes.Types as V0
-import qualified Pos.Wallet.Web.Methods.Restore as V0
+import qualified Pos.Wallet.Web.Methods as V0
 
 import           Cardano.Wallet.API.V1.Migration
 import           Cardano.Wallet.API.V1.Types as V1
@@ -57,10 +57,15 @@ listWallets PaginationParams {..} = do
       }
     _ -> return $ OneOf $ Left example
 
-updatePassword :: WalletId
-               -> PasswordUpdate
-               -> MonadV1 Wallet
-updatePassword _ _ = liftIO $ generate arbitrary
+updatePassword
+    :: (MonadWalletLogic ctx m)
+    => WalletId -> PasswordUpdate -> m Wallet
+updatePassword wid PasswordUpdate{..} = do
+    wid' <- migrate wid
+    _ <- join $ V0.changeWalletPassphrase <$> migrate wid
+                                          <*> pure pwdOld
+                                          <*> pure pwdNew
+    V0.getWallet wid' >>= migrate
 
 deleteWallet :: WalletId
              -> MonadV1 NoContent
--- a/wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs
+++ b/wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs
@@ -49,7 +49,7 @@ instance ToJSON Core.PassPhrase where
 
 instance FromJSON Core.PassPhrase where
     parseJSON Null        = mempty
-    parseJSON x@(String pp) = case mkPassPhrase pp of
-        Left e    -> typeMismatch e x
+    parseJSON (String pp) = case mkPassPhrase pp of
+        Left e    -> fail e
         Right pp' -> pure pp'
     parseJSON x           = typeMismatch "parseJSON failed for PassPhrase" x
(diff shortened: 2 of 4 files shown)


wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs
```diff
@@ -27,13 +27,16 @@ instance FromJSON BackupPhrase where
 instance ToJSON CFilePath where
   toJSON (CFilePath c) = toJSON c
 
+-- FIXME: This couple of instances is not complementary,
+-- can't write roundtrip test for them.
+-- It doesn't matter though, because we will hardly need to return password
```
adinapoli-iohk: Yes, this is something I want to fix for V1. I my opinion Aeson instances should always roundtrip 😉

wallet-new/test/MarshallingSpec.hs
```diff
@@ -5,28 +5,62 @@ import           Universum
...
+        aesonRoundtripProp @AssuranceLevel Proxy
+        aesonRoundtripProp @Payment Proxy
+        aesonRoundtripProp @PaymentDistribution Proxy
+        aesonRoundtripProp @NewWallet Proxy
+        aesonRoundtripProp @Core.Coin Proxy
+        aesonRoundtripProp @TransactionGroupingPolicy Proxy
+        aesonRoundtripProp @TransactionType Proxy
+        aesonRoundtripProp @Transaction Proxy
+
+    describe "Special" $ do
```
adinapoli-iohk: "Special" is a bit vague. What about "Invariants"?

wallet-new/test/MarshallingSpec.hs
```diff
@@ -5,28 +5,62 @@ import           Universum
...
+        describe "password" $ do
+            prop "empty" $
```
adinapoli-iohk: "empty" is a bit generic. What about `prop "empty string decodes to empty passphrase"`? Same applies for the rest of the two functions.

adinapoli-iohk: Also, there doesn't seem to be properties in the QuickCheck sense, but more unit tests, so I would suggest change `prop` to simply `it`. What do you think?

wallet-new/test/MarshallingSpec.hs
```diff
@@ -5,28 +5,62 @@ import           Universum
...
 aesonRoundtrip (_ :: proxy a) = forAll arbitrary $ \(s :: a) -> do
     decode (encode (toJSON s)) === Just s
+
+aesonRoundtripProp
+    :: (Arbitrary a, ToJSON a, FromJSON a, Eq a, Show a, Typeable a)
+    => proxy a -> Spec
+aesonRoundtripProp proxy =
+    prop ("Aeson " <> show (typeRep proxy) <> " roundtrips") (aesonRoundtrip proxy)
+
+decodesTo :: FromJSON a => LByteString -> (a -> Bool) -> Property
```
adinapoli-iohk: I think this is better written as a normal `hspec` `Expectation` rather than a `Property`.

wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs
Martoon-00: Then we can rewrite `ToJSON` so that for empty passphrase it produces `null`.
But whether do we need `instance ToJSON PassPhrase` at all?

wallet-new/test/MarshallingSpec.hs
Martoon-00: Would be nice, I get used more to it

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
```diff
@@ -57,10 +57,15 @@ listWallets PaginationParams {..} = do
...
-updatePassword :: WalletId
-               -> PasswordUpdate
-               -> MonadV1 Wallet
-updatePassword _ _ = liftIO $ generate arbitrary
+updatePassword
+    :: (MonadWalletLogic ctx m)
+    => WalletId -> PasswordUpdate -> m Wallet
+updatePassword wid PasswordUpdate{..} = do
+    wid' <- migrate wid
+    _ <- join $ V0.changeWalletPassphrase <$> migrate wid
```
flyingleafe: Why do `migrate wid` again and not use `wid'`?

wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs
flyingleafe: I don't understand why don't these instances roundtrip, honestly. `Arbitrary` instance for `PassPhrase` doesn't allow for passphrases which are not 0 or 32 bytes long. It seems like that if these instances don't roundtrip, then one of them is buggy =\

adinapoli-iohk:
>  It seems like that if these instances don't roundtrip, then one of them is buggy =\

@flyingleafe @Martoon-00 I want to review those and others (like the one for `ApplicationName`) in planning on Monday. The fact we are not removing `newtype` wrappers from our JSON instances (via `defaultOptions {  unwrapUnaryRecords = True }`). I know this sounds terribly vague but for now I just want to plant the seed on your brains, we will discuss when the time comes  😉 

@Martoon-00  --  There are conflicts now, would you also mind addressing those? Cheers! 😉

Martoon-00:
Okay, ready for rereview

wallet-new/test/MarshallingSpec.hs
```diff
@@ -6,28 +6,61 @@ import           Cardano.Wallet.API.V1.Errors (WalletError)
...
+        aesonRoundtripProp @TransactionGroupingPolicy Proxy
+        aesonRoundtripProp @TransactionType Proxy
+        aesonRoundtripProp @Transaction Proxy
+        aesonRoundtripProp @WalletError Proxy
+
+    describe "Invariants" $ do
+        describe "password" $ do
+            it "empty string decodes to empty password" $
+                jsonString "" `decodesTo` (== emptyPassphrase)
+            it "base-64 string of length 32 decodes to nonempty password" $
```
akegalj: isn't it base16 string?

Martoon-00: Thanks, fixed

akegalj:
@adinapoli-iohk I believe this could be merged (looks good to me)

Some stuff from this PR would be useful to have for further tests.

adinapoli-iohk:
Yep, happy to do so as soon as @Martoon-00 fixes the conflicts 😁

Martoon-00:
@adinapoli-iohk Some jobs in travis passed, can we merge?

adinapoli-iohk:
@Martoon-00 Let me wait for https://github.com/input-output-hk/cardano-sl/pull/1986 to be merged so that we don't "pollute" with more commit something that Ivan already reviewed, than will pull the trigger on this one. Thanks!

Martoon-00:
Closed in favor of #2023

PR #1958, opened 2017-11-23
[CSL-1839] Catch and logs exceptions right after logger initialization 

hithroc:
Ported #1753 loggerBrackets, and added exception handling and logging.

### DIFF
--- a/auxx/src/Main.hs
+++ b/auxx/src/Main.hs
@@ -17,6 +17,7 @@ import           Pos.Network.Types   (NetworkConfig (..), Topology (..),
                                       topologyDequeuePolicy, topologyEnqueuePolicy,
                                       topologyFailurePolicy)
 import           Pos.Ssc.SscAlgo     (SscAlgo (GodTossingAlgo))
+import           Pos.Util            (logException)
 import           Pos.Util.UserSecret (usVss)
 import           Pos.WorkMode        (RealMode)
 
@@ -70,6 +71,7 @@ action opts@AuxxOptions {..} = withConfigurations conf $ do
 
 main :: IO ()
 main = do
-  opts <- getAuxxOptions
-  let loggingParams = CLI.loggingParams "auxx" (aoCommonNodeArgs opts)
-  loggerBracket loggingParams . runProduction $ action opts
+    opts <- getAuxxOptions
+    let loggingParams = CLI.loggingParams "auxx" (aoCommonNodeArgs opts)
+    loggerBracket loggingParams . logException "auxx" $ do
+        runProduction $ action opts
--- a/auxx/src/Main.hs
+++ b/auxx/src/Main.hs
@@ -12,7 +12,7 @@ import           System.Wlog         (logInfo)
 import qualified Pos.Client.CLI      as CLI
 import           Pos.Core            (Timestamp (..), gdStartTime, genesisData)
 import           Pos.Launcher        (NodeParams (..), bracketNodeResources,
-                                      runRealBasedMode, withConfigurations)
+                                      loggerBracket, runRealBasedMode, withConfigurations)
 import           Pos.Network.Types   (NetworkConfig (..), Topology (..),
                                       topologyDequeuePolicy, topologyEnqueuePolicy,
                                       topologyFailurePolicy)
@@ -69,4 +69,7 @@ action opts@AuxxOptions {..} = withConfigurations conf $ do
         lift $ runReaderT auxxAction auxxContext
 
 main :: IO ()
-main = getAuxxOptions >>= runProduction . action
+main = do
+  opts <- getAuxxOptions
+  let loggingParams = CLI.loggingParams "auxx" (aoCommonNodeArgs opts)
+  loggerBracket loggingParams . runProduction $ action opts
--- a/core/Pos/Util/Util.hs
+++ b/core/Pos/Util/Util.hs
@@ -67,6 +67,9 @@ module Pos.Util.Util
        -- * Aeson
        , parseJSONWithRead
 
+       -- * Exceptions
+       , logException
+
        -- * Instances
        -- ** Lift Byte
        -- ** Lift HashMap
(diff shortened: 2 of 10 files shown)


hithroc:
1. I asked @gromakovsky about async exceptions and he told me to not catch them and use `safe-exceptions`.

int-index:
@gromakovsky I don't understand, wouldn't we want to log async exceptions as well?

wallet/node/Main.hs
```diff
@@ -115,6 +116,8 @@ action (WalletNodeArgs (cArgs@CommonNodeArgs{..}) (wArgs@WalletArgs{..})) =
 main :: IO ()
 main = do
     args <- getWalletNodeOptions
-    CLI.printFlags
-    putText "[Attention] Software is built with wallet part"
-    runProduction $ action args
+    let loggingParams = CLI.loggingParams "node" (wnaCommonNodeArgs args)
+    loggerBracket loggingParams . handleAny (usingLoggerName "node" . logError . show) $ do
```
ksaric: `usingLoggerName "wallet"`?

explorer/src/explorer/Main.hs
```diff
@@ -48,8 +50,10 @@ printFlags = do
 main :: IO ()
 main = do
     args <- getExplorerOptions
-    printFlags
-    runProduction $ action args
+    let loggingParams = CLI.loggingParams "node" (enaCommonNodeArgs args)
+    loggerBracket loggingParams . handleAny (usingLoggerName "auxx" . logError . show) $ do
```
ksaric: `usingLoggerName "explorer"`?

hithroc: Well, I took the names straightly from #1753, but I think that is a good point.

gromakovsky:
> I don't understand, wouldn't we want to log async exceptions as well?

Hm, I am not sure, maybe we want. Especially if we use `onException` (I didn't think about this function). Probably catching async exceptions in `main` is ok if they are rethrown after being caught. What I don't like:
1. If we use `onException` which catches an async exception, it will rethrow it synchronously. I don't know whether it's bad to throw an async exception synchronously, but it sounds like a bad idea.
2. If we catch `StackOverflow`, is it ok to call `logError` (which is quite complex function actually)? Probably it won't succeed. But we can try, we won't die if we try.

I can't imagine any other bad cases for main thread, so ok, I don't mind catching async exceptions as well (using `onException`).

hithroc:
I will probably use `withException` then, because `onException` doesn't actually give you the exception.

auxx/src/Main.hs
```diff
@@ -69,4 +70,8 @@ action opts@AuxxOptions {..} = withConfigurations conf $ do
         lift $ runReaderT auxxAction auxxContext
 
 main :: IO ()
-main = getAuxxOptions >>= runProduction . action
+main = do
+  opts <- getAuxxOptions
+  let loggingParams = CLI.loggingParams "auxx" (aoCommonNodeArgs opts)
+  loggerBracket loggingParams . handleAny (usingLoggerName "auxx" . logError . show) $ do
```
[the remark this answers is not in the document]
georgeee: 2-space indentation?

wallet/node/Main.hs
adinapoli-iohk: Yes, was going to say the same, good idea.

core/Pos/Util/Util.hs
```diff
@@ -619,3 +624,14 @@ parseJSONWithRead :: Read a => A.Value -> A.Parser a
...
+----------------------------------------------------------------------------
+-- Exceptions
+----------------------------------------------------------------------------
+
+-- | Catch and log an exception, then rethrow it
+logException :: LoggerName -> IO a -> IO a
+logException name = E.handle (\e -> handler e >> E.throw e)
+  where
+    handler :: E.SomeException -> IO ()
+    handler = usingLoggerName name . logError . show
```
gromakovsky: It's better to use `displayException` instead of `show`. Also there in `instance Buildable SomeException` in `serokell-util`. You can import `Serokell.Util.Exceptions` and use `pretty` here.

gromakovsky:
@int-index I think everything has been fixed, please re-review.

pkgs/default.nix
```diff
@@ -1263,6 +1263,7 @@ self: {
             node-sketch
             optparse-applicative
             parsec
+            safe-exceptions
```
int-index: Why?

adinapoli-iohk:
> I'd prefer squashing the review commits into the main one here before we merge.

In theory that can be done via a squashed merge here directly from the command line :) (unless people have other stronger opinions).

wallet/node/Main.hs
```diff
@@ -115,6 +116,8 @@ action (WalletNodeArgs (cArgs@CommonNodeArgs{..}) (wArgs@WalletArgs{..})) =
 main :: IO ()
 main = do
     args <- getWalletNodeOptions
-    CLI.printFlags
-    putText "[Attention] Software is built with wallet part"
-    runProduction $ action args
+    let loggingParams = CLI.loggingParams "wallet" (wnaCommonNodeArgs args)
+    loggerBracket loggingParams . logException "wallet" $ do
```
gromakovsky: I don't think it's a good idea to use `wallet` logger name here. End-users run `cardano-node` with `log-config-prod.yaml` (from the root of the repository). Only logs sent to `node` logger (and subloggers) will be sent to files. So basically there will be no logs.
I suggest that we leave `node` here (this line and one above) in cardano-sl-1.0 branch to minimize changes (otherwise we'd have to modify log config which would be another moving part).

gromakovsky:
I think it makes sense to have two commits: one which moves `loggerBracket` to `Main` and another one which adds exceptions logging. They are self-contained and dedicated to different problems and having them separate will make further merge to master easier (because the first part is in master already).

core/Pos/Util/Util.hs
```diff
@@ -619,3 +625,14 @@ parseJSONWithRead :: Read a => A.Value -> A.Parser a
...
     either (fail . toString) pure . readEither @String <=<
     parseJSON
+
+----------------------------------------------------------------------------
+-- Exceptions
+----------------------------------------------------------------------------
+
+-- | Catch and log an exception, then rethrow it
+logException :: LoggerName -> IO a -> IO a
+logException name = E.handle (\e -> handler e >> E.throw e)
```
int-index: According to the new guidelines use `safe-exceptions` (`catchAsync`)

PR #1959, opened 2017-11-23
[CSL-1916] Change wallet backend code to use 3 level derivation

et4te:
Summary:
* chainGenesisIndex added to constants for firstHDAddress.
* max address size upped from 76 to 81 to account for the new derivation path.
* minor re-factorings / chainIndex added where code failed.


adinapoli-iohk:
Quite stale by now, let's reopen if necessary.

PR #1960, opened 2017-11-23
[CSL-1903] Check for coin overflow in payment endpoints

Martoon-00:


scripts/launch/demo.sh
```diff
@@ -10,7 +10,7 @@ fi
 # Make sure we're using proper version of tmux.
 tmux_actual_version=$(tmux -V | awk '{print $2}')
 # All tmux versions contain two numbers only.
-tmux_proper_versions=("2.3" "2.4" "2.5" "2.6")
+tmux_proper_versions=("2.3" "2.4" "2.5" "2.6" "master")
```
adinapoli-iohk: Sounds like this is outside the scope of this PR, let's remove it 😉

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -392,7 +392,11 @@ prepareTxRawWithPicker inputPicker utxo outputs (TxFee fee) = do
     when (totalMoney == mkCoin 0) $
         throwError $ GeneralTxError "Attempted to send 0 money"
 
-    let moneyToSpent = totalMoney `unsafeAddCoin` fee
+    moneyToSpent <- case integerToCoin (sumCoins [totalMoney, fee]) of
+        -- we don't care about exact number if user desires all money in the world
+        Left _  -> throwError $ NotEnoughMoney maxBound
```
adinapoli-iohk: I'm unsure about this error. Are we sure that's descriptive enough for frontend to use? @DominikGuzei , does this make any difference for you folks? I don't think it does, but checking just in case 😉

Martoon-00: That's the same error they receive when user simply has no enough money on his account to spend. Here's indeed the exact same case

adinapoli-iohk: Ah, perfect! Then simply remove the tmux change and I'm happy to merge! :)

Martoon-00:
@adinapoli-iohk Since tmux change is already here and CI passed, why can't we just rebase this PR upon master? It would be the same if I move that commit with tmux change to another PR right now

adinapoli-iohk:
> Since tmux change is already here and CI passed, why can't we just rebase this PR upon master?

I hear you, it's just that I'm uncomfortable with fixing something else on a PR which is supposed to tackle a completely different issue. Ok, here's my middle-ground: @gromakovsky , @georgeee , this PR also adds a new version of `tmux` to the `demo.sh` script, and it's necessary to @Martoon-00 to be able to launch the cluster.

Are you OK with merging the change as part of this PR?

gromakovsky:
Ideally I would prefer having that change in a separate PR (for various reasons), but if the rest of this PR is ready to be merged and since CI is green, I can live with merging that change as part of this PR.
I agree that it's a bad practice and should be done rarely (ideally never).

adinapoli-iohk:
Thanks @gromakovsky , happy to go with this.

PR #1961, opened 2017-11-24
[CSL-1899] Document `CFilePath` better in swagger

Martoon-00:
Something like this?

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Swagger/Instances/Schema.hs
+++ b/wallet/src/Pos/Wallet/Web/Swagger/Instances/Schema.hs
@@ -1,7 +1,4 @@
-{-# LANGUAGE OverloadedLists     #-}
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeOperators       #-}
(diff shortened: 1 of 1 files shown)


wallet/src/Pos/Wallet/Web/Swagger/Instances/Schema.hs
```diff
@@ -102,6 +99,21 @@ instance ToSchema (FileData tag) where
...
+                & example ?~ "keys/1.key"
+        return $ named "FilePath" schema'
+      where
+        desc = "Path to file.\n \
+               \Note that it is represented as JSON-string, \
+               \one may need to properly escape content. For instance:\n \
+               \curl ... -X \"\\\\\"1.key\"\\\\\".\n\
+               \Also, when on Windows, don't forget to double-escape \
+               \ backslashes, e.g. \
+               \ \"C:\\\\\\\\\\\\\\\\keys\\\\\\\\1.key\""
```
Martoon-00: Don't ask why...

Martoon-00:
We with @adinapoli-iohk agreed that it's not required to be mentioned in description, because proper swagger UI already marks request body is mandatory in another way

PR #1964, opened 2017-11-24
Update stack.yaml & scripts

Pastafarianist:


scripts/haskell/update-cabal-versions.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
volhovm: Nope.

gromakovsky: 🤷‍♂️

Pastafarianist: BTW it doesn't even work with `sh`.

```
$ sh scripts/haskell/update-cabal-versions.sh 1.0.3
scripts/haskell/update-cabal-versions.sh: 10: scripts/haskell/update-cabal-versions.sh: Syntax error: "(" unexpected
```

PR #1966, opened 2017-11-24
[CSL-1910] New settings endpoints for V1

adinapoli-iohk:
~~This is almost there but it's still a WIP as I would like to solve some JSON-schema issues with our current `SoftwareVersion`, which I cannot do directly without breaking V0-compatibility (will do after lunch), but I thought about opening the PR now to gather early feedback~~. After several yak shaved and numerous paint layers to the shed, I have rationalised the old `settings` and `info` endpoints into two:

- `/api/v1/node-settings` : Retrieves the static settings for a node, for example the project version, the protocol slot duration, etc.

- `/api/v1/node-info`: Retrieves the _dynamic_ settings for a node, for example the time difference or the sync progress.

Furthermore, I have added this concept of unit of measure for quantities which are non obvious at first glance. This means now we would represent things this way in our JSON:

screen shot 2017-11-24 at 16 58 55

screen shot 2017-11-24 at 13 45 19

Please let me know if this is over-engineering. I thought it would have been quite a boon for 3rd parties and exchanges as it's basically self-documenting.


DominikGuzei:
@adinapoli-iohk this looks great! I don't think that clarity without bigger drawbacks can be called over-engineering 😆

adinapoli-iohk:
Happy to hear that, @DominikGuzei ! Will merge as soon as I solve the pesky conflicts 😁

PR #1970, opened 2017-11-25
[Cosmetics] Fix more lint/style problems

volhovm:
Again, cosmetic changes only (except for one subtle logic change in cbor reference implementation, but it doesn't matter). I wanted to push them as CSL-1399, but it wasn't merged, so here we go. 
cc @pva701


volhovm:
Ok, passing CI is the only thing I was waiting for (i don't think it should be reviewed since it's super trivial).

adinapoli-iohk:
Thank you @volhovm !

PR #1971, opened 2017-11-25

volhovm:
Server side block processing now should be faster. What this PR does is basically traversing header hashes (only) oldest to newest (using forward links) instead of traversing block headers newest to oldest (which required deserializing headers).

So I added forward links, limited number of blocks to give back to `recoveryHeadersMessage`. Tested the whole thing.

### DIFF
--- a/lib/src/Pos/Block/Logic/Header.hs
+++ b/lib/src/Pos/Block/Logic/Header.hs
@@ -9,14 +9,15 @@ module Pos.Block.Logic.Header
        , classifyHeaders
        , getHeadersFromManyTo
        , getHeadersOlderExp
-       , getHeadersFromToIncl
+       , getHeadersRange
        ) where
 
 import           Universum
+import           Unsafe (unsafeLast)
 
+import           Control.Lens (to)
 import           Control.Monad.Except (MonadError (throwError))
 import           Control.Monad.Trans.Maybe (MaybeT (MaybeT), runMaybeT)
-import           Data.List.NonEmpty ((<|))
 import qualified Data.Text as T
 import           Formatting (build, int, sformat, (%))
 import           Serokell.Util.Text (listJson)
@@ -31,7 +32,6 @@ import           Pos.Core (BlockCount, EpochOrSlot (..), HasConfiguration, Heade
                            epochOrSlotG, getChainDifficulty, getEpochOrSlot, headerHash,
                            headerHashG, headerSlotL, prevBlockL)
 import           Pos.Core.Block (BlockHeader)
-import           Pos.Core.Configuration (genesisHash)
 import           Pos.Crypto (hash)
 import           Pos.DB (MonadDBRead)
 import qualified Pos.DB.Block.Load as DB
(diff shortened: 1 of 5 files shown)


lib/src/Pos/Block/Logic/Header.hs
```diff
@@ -353,34 +353,67 @@ getHeadersOlderExp upto = do
...
+    olderHd <- fromMaybeM "can't retrieve older header" $ DB.getHeader older
+    let olderD = olderHd ^. difficultyL
+    let newerD = newerHd ^. difficultyL
+
+    -- Proving newerD >= olderD
+    when (newerD == olderD) $
+        throwError $ "getHeadersRange: newer and older headers have " <>
+                     "the same difficulty, but are not equal"
+    when (newerD < olderD) $
+        throwError "getHeadersRange: newer header is less dificult that older one"
```
gromakovsky: than

lib/src/Pos/Block/Logic/Header.hs
```diff
@@ -353,34 +353,67 @@ getHeadersOlderExp upto = do
...
+
+    -- How many epochs does this range cross.
+    let genDiff :: Int
+        genDiff = fromIntegral $ newerHd ^. epochIndexL - olderHd ^. epochIndexL
+    -- Number of blocks is difficulty difference + number of genesis blocks.
+    let depthDiff :: Int
+        depthDiff = genDiff + fromIntegral (newerD - olderD)
+
+    -- We load these depthDiff blocks.
+    let cond _curHash depth = depth < depthDiff
```
gromakovsky: Omg, what's the purpose of doing all this stuff with difficulty (which looks rather complicated), when you can simply check whether `_curHash == newer`?

lib/src/Pos/Block/Logic/Header.hs
```diff
@@ -353,34 +353,67 @@ getHeadersOlderExp upto = do
...
+
+    -- This is [oldest..newest) headers.
+    allExceptNewest <- GS.loadHashesUpWhile older cond
+
+    -- Sometimes we will get an empty list, if we've just switched the
+    -- branch (after first checks are performed here) and olderHd is
+    -- no longer in the main chain.
+    when (null $ allExceptNewest ^. _OldestFirst) $ throwError $
```
gromakovsky: We can use snapshots to avoid it, btw.

lib/src/Pos/Block/Logic/Header.hs
volhovm: Because i don't want to traverse more than `depthDiff` blocks ever. Replacing the check with the one you suggest would open an DOS attack vector.

lib/src/Pos/Block/Logic/Header.hs
volhovm: Indeed. It's not really important here, but if you would like to see it, please tell me, I'll figure it out.

lib/src/Pos/Block/Logic/Header.hs
volhovm: Fixed

lib/src/Pos/Block/Logic/Header.hs
gromakovsky: Can you create an issue and let's postpone it (~for eternity~)?

lib/src/Pos/Block/Logic/Header.hs
gromakovsky: I don't get it. Currently this function can load all blocks, there is no upper bound other than that. If you pass `const (const True)` as `cond`, there will be the same bound.

Can you elaborate on DOS attack vector which your solution makes impossible?

volhovm: Oh, indeed, sorry -- I assumed that nobody can query more than `recoveryHeadersMessage` blocks at once. I guess i should put this limitation there.

Attack is basically asking to send all the chain, from the very first block, to the last one. When i put limitation on the `depthDiff` it will be better.

Indeed, `cond` can be expressed both in terms of depth (how it is done now) and it rems of `curHash == newer`. There's no difference.

lib/src/Pos/Block/Logic/Header.hs
volhovm: https://iohk.myjetbrains.com/youtrack/issue/CSL-1950

lib/src/Pos/Block/Logic/Header.hs
gromakovsky: I still think that it would be simpler, if `cond` was checking `curHash == newer && maybe True (depth <=) depthLimitM` rather than manipulating difficulties. But probably it's subjective, so I won't continue this discussion.

volhovm:
This PR is closed. I'll reopen one pointing on cardano-sl-1.0.

PR #1975, opened 2017-11-25
[CSL-1780] Extract block package

Pastafarianist:
This PR introduces a new package called `cardano-sl-block`, based mostly on `Pos.Block.*` modules. Summary of accompanying changes:

* Moved modules:
    * `Pos.Arbitrary.Block`
    * `Pos.Binary.Block.Types`
    * `Pos.Block.*`
    * `Pos.DB.Block[.*]`
    * `Pos.GState.BlockExtra`
    * `Pos.Lrc[.*]`
    * `Pos.Util.JsonLog`
* New module: `Pos.Block.Configuration`. I moved `ccNetworkDiameter` and `ccRecoveryHeadersMessage` from `NodeConfiguration` there because many functions in `Pos.Block` depend on them. I also patched `configuration.yaml`.
* New module: `Pos.Block.WorkMode`. It contains `BlockWorkMode`, which is a subset of constraints from `Pos.WorkMode.Class`. Furthermore, it contains a number of constraints for instances which are unavailable in `cardano-sl-block`, specifically, the ones from `Pos.Binary.Communication`, `Pos.Communication.Message` and `Pos.Communication.Limits`.
* New module: `Pos.GState.SanityCheck`. It exports `sanityCheckDB` which was moved from `Pos.DB.DB`.
* New module: `Pos.Lrc.Genesis`. It has `genesisLeaders` which had previously been in `Pos.Context.Functions`.
* `[Monad]LastKnownHeader[Tag]`, `[Monad]ProgressHeader[Tag]` and `[Monad]RecoveryHeader[Tag]` were moved from `Pos.Context.Context` to `Pos.Block.Types`. I'm not sure that this is the best place to put them; please suggest otherwise if you have any better ideas.
* `Pos.Ssc.Util` is deleted and its contents (`toSscBlock`) are moved into `Pos.Block.Logic.Internal`.

There is an issue in `wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs`: it defines `instance {From,To}JSON Coin`, which is already defined in `Pos.Aeson.Core`. That wasn't a problem before as those instances had not been available to `Cardano.Wallet.Orphans.Aeson`, but now they have leaked through. I've commented out the lines defining the instances in `Cardano.Wallet.Orphans.Aeson`, but I believe we shouldn't merge commented-out code, and we'll have to find some solution instead.


volhovm:
Before everything else:
1. Fix build scripts (`./b`).
2. Run `weeder` on the project (please) and ensure you didn't introduce needless dependencies.

block/src/Pos/Block/Types.hs
```diff
@@ -51,3 +66,18 @@ instance HasDifficulty Blund where
 
 instance HasHeaderHash Block => HasHeaderHash Blund where
     headerHash = headerHash . fst
+
+data LastKnownHeaderTag
```
volhovm: I think it'd be a good idea to move the documentation from `Context.hs` here.

block/src/Pos/Block/WorkMode.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE DataKinds #-}
+
+-- | Various constraints needed for block processing.
+
+module Pos.Block.WorkMode
```
volhovm: I'd actually rename it into `BlockWorkMode`.

lib/src/Pos/Context/Context.hs
```diff
@@ -61,21 +55,6 @@ import           Pos.Util.UserSecret (HasUserSecret (..), UserSecret)
 -- NodeContext
 ----------------------------------------------------------------------------
 
-data LastKnownHeaderTag
```
Pastafarianist: There is no documentation to move, unfortunately.

Pastafarianist:
`weeder` reports that `cardano-sl-block` does not introduce any new weeds:

```
= Package cardano-sl-block =
No warnings
```

However, there are a lot of warnings regarding other projects. I've uploaded the full report [here](https://gist.github.com/Pastafarianist/2effb0ffae5180b91e4acad05c8f5bb6).

wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs
```diff
@@ -28,7 +28,7 @@ instance FromJSON BackupPhrase where
 instance ToJSON CFilePath where
   toJSON (CFilePath c) = toJSON c
 
-deriveJSON defaultOptions ''Core.Coin
+-- deriveJSON defaultOptions ''Core.Coin
```
adinapoli-iohk: Why have this been commented out?

Pastafarianist: @adinapoli-iohk please see the last paragraph in the PR description.

adinapoli-iohk: @Pastafarianist Sorry, my bad! This is what happens by glossing over the PR description 😉 . If the instances are equivalent I'm ok with removing the code from the `Wallet.Orphans.Aeson` module, as I can use the ones in `Pos.Aeson.Core`.

If they are not, we will need to come up with a different strategy. The tension in the design space is that we are trying to design `wallet-new` with backward compatibility in mind, so we ideally don't want neither to modify the `Pos.Aeson.Core` instance to fits `wallet-new` needs nor the other way around.

block/src/Pos/Block/WorkMode.hs
gromakovsky: Maybe call it simply `Mode`? We have many modules with this name.

gromakovsky:
> This PR introduces a new module called `cardano-sl-block`

It's called `package`, not module.

block/src/Pos/Block/BlockWorkMode.hs
```diff
@@ -0,0 +1,83 @@
...
+    , Bi MsgBlock
+    , Message MsgGetHeaders
+    , Message MsgHeaders
+    , Message MsgGetBlocks
+    , Message MsgBlock
+    , MessageLimited MsgGetHeaders
+    , MessageLimited MsgHeaders
+    , MessageLimited MsgGetBlocks
+    , MessageLimited MsgBlock
+    )
```
gromakovsky: Btw, you could've used `Each` like here https://github.com/input-output-hk/cardano-sl/blob/c60901742e5fab76f34f7207ee91960e2da93131/ssc/Pos/Ssc/Message.hs#L96, but it's 100% minor.

lib/src/Pos/Context/Context.hs
gromakovsky: :pepe_sad:

gromakovsky:
Btw,
```
[ 3 of 18] Cardano.Wallet.Orphans.Aeson

/home/gromak/Work/cardano-sl/wallet-new/src/Cardano/Wallet/Orphans/Aeson.hs:16:1: warning: [-Wunused-imports]
    The qualified import of ‘Pos.Core.Types’ is redundant
      except perhaps to import instances from ‘Pos.Core.Types’
    To import instances alone, use: import Pos.Core.Types()
```

Pastafarianist:
@adinapoli-iohk here is a slightly cleaned-up and reformatted instance that is generated by `deriveJSON defaultOptions ''Core.Coin`:

```haskell
instance ToJSON Coin where
    toJSON =
        \value_0 ->
            case value_0 of
                Coin arg1_1 ->
                    Data.Aeson.Types.Internal.object
                        [ Data.Aeson.TH.keyValuePairWith
                              toJSON
                              (Data.Text.pack "getCoin")
                              arg1_1
                        ]
    toEncoding =
        \value_2 ->
            case value_2 of
                Coin arg1_3 ->
                    Data.Aeson.Encoding.Internal.wrapObject
                        (Data.Aeson.Encoding.Internal.econcat
                             (Data.OldList.intersperse
                                  Data.Aeson.Encoding.Internal.comma
                                  [ Data.Aeson.Encoding.Internal.text
                                        (Data.Text.pack "getCoin") Data.Aeson.Encoding.Internal.><
                                    (Data.Aeson.Encoding.Internal.colon Data.Aeson.Encoding.Internal.><
                                     toEncoding arg1_3)
                                  ]))

instance FromJSON Coin where
    parseJSON =
        \value_4 ->
            case value_4 of
                Data.Aeson.Types.Internal.Object recObj_5 ->
                    Coin Data.Functor.<$>
                    Data.Aeson.TH.lookupField
                        parseJSON
                        "Coin"
                        "Coin"
                        recObj_5
                        (Data.Text.pack "getCoin")
                other_6 ->
                    Data.Aeson.TH.parseTypeMismatch'
                        "Coin"
                        "Coin"
                        "Object"
                        (Data.Aeson.TH.valueConName other_6)
```

Basically, this means that `Coin {getCoin = 0}` will be encoded as `{"getCoin": 0.0}`. However, the instance in `Pos.Aeson.Core` works in such way that `Coin` is encoded identically to `Word64`, that is, `Coin {getCoin = 0}` becomes `0.0`.

I believe `wallet-new` must not use instances that differ from the ones in Core. Is the removal of the current instance from `wallet-new` going to break anything?

adinapoli-iohk:
> I believe wallet-new must not use instances that differ from the ones in Core. Is the removal of the current instance from wallet-new going to break anything?

@Pastafarianist Unfortunately it's not that simple. We want to offer a nice and clean API to exchanges and users, so we cannot reuse most instances from Core _as-they-are_, but we usually need to:

1. Write our own (orphaned) instances so we can expose a nicer API;
2. Write newtypes to completely customise the API at our own will in a safe way.

1. Is easier because 2. comes with the `newtype` wrapper overhead, but it's an acceptable tradeoff as that impact just the programmers and not the end users.

In general, be advised that the current version of the wallet uses a bunch of `Core` instances directly anyway (`SoftwareUpdate` is an example) so such changes will always break backward compatibility.

In general I don't think it's a good design for the middleware to depend _directly_ from the `Core` instances, because it doesn't give us any protection against breaking changes.

As far as this PR goes, if there is no other way to have the orphans defined in `wallet-new` without clashes, I will be forced to go with option 2. 😉

Pastafarianist:
Are you sure that `deriveJSON defaultOptions ''Core.Coin` is nicer and cleaner than the one defined in Core? The former exposes the implementation details (specifically, the name of the `getCoin` constructor), while the latter produces just a number.

adinapoli-iohk:
> Are you sure that deriveJSON defaultOptions ''Core.Coin is nicer or cleaner than the one defined in Core? The former exposes the implementation details (specifically, the name of the getCoin constructor), while the latter produces just a number.

Oh, don't get me wrong: in this case the latter is *exactly* what we need, `Core.Coin` should really be just an alias for a number. I was thinking more about the general case 😉

Pastafarianist:
In the general case we'll probably have to go with `newtype` wrappers since the instances from Core are likely to leak anyway.

If this issue is resolved, can you approve the PR?

PR #1977, opened 2017-11-27
[CSM-567] integration tests for import export of wallet

v0d1ch:
@gromakovsky I need some help regarding this.

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Backup.hs
+++ b/wallet/src/Pos/Wallet/Web/Backup.hs
@@ -11,29 +11,47 @@ import           Universum
 
 import qualified Data.HashMap.Strict as HM
 import qualified Data.SemVer as V
-
 import           Pos.Crypto (EncryptedSecretKey)
+import           Pos.Crypto.Signing.Safe (emptyPassphrase, safeKeyGen)
+import           Pos.Wallet.Web.ClientTypes (AccountId (..), CAccountMeta (..),
+                                             CId, CWalletMeta (..), Wal)
 import           Pos.Util.Util (maybeThrow)
 import           Pos.Wallet.Web.Account (AccountMode, getSKById)
-import           Pos.Wallet.Web.ClientTypes (AccountId (..), CAccountMeta (..), CId,
-                                             CWalletMeta (..), Wal)
 import           Pos.Wallet.Web.Error (WalletError (..))
 import           Pos.Wallet.Web.State (getAccountMeta, getWalletMeta)
 import           Pos.Wallet.Web.Util (getWalletAccountIds)
+import           Test.QuickCheck (Arbitrary (..), elements)
 
 currentBackupFormatVersion :: V.Version
 currentBackupFormatVersion = V.initial & V.major .~ 1
 
-newtype WalletMetaBackup = WalletMetaBackup CWalletMeta
-newtype AccountMetaBackup = AccountMetaBackup CAccountMeta
+newtype WalletMetaBackup = WalletMetaBackup CWalletMeta deriving Show
+newtype AccountMetaBackup = AccountMetaBackup CAccountMeta deriving Show
 
 data WalletBackup = WalletBackup
     { wbSecretKey :: !EncryptedSecretKey
     , wbMeta      :: !WalletMetaBackup
     , wbAccounts  :: !(HashMap Int AccountMetaBackup)
-    }
+    } deriving Show
 
 data TotalBackup = TotalBackup WalletBackup
+
+instance Arbitrary WalletBackup where
+    arbitrary = do
+        cwNameT <- arbitrary
+        caNameT <- arbitrary
+        wbInt   <- arbitrary
+        cwAssurance <- elements [minBound .. maxBound]
+        (_, esk) <- safeKeyGen emptyPassphrase
+        let cwUnit = 1
+            cwName = (cwNameT :: Text )
+            caName = (caNameT :: Text )
+            wMetaBackup = WalletMetaBackup $ CWalletMeta {..}
+        return $ WalletBackup
+            { wbSecretKey = esk
+            , wbMeta = wMetaBackup
+            , wbAccounts = HM.singleton wbInt (AccountMetaBackup CAccountMeta {..})
+            }
 
 getWalletBackup :: AccountMode ctx m => CId Wal -> m WalletBackup
 getWalletBackup wId = do
--- a/wallet/src/Pos/Wallet/Web/Methods/Backup.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/Backup.hs
@@ -5,6 +5,8 @@
 module Pos.Wallet.Web.Methods.Backup
        ( importWalletJSON
        , exportWalletJSON
+       , restoreWalletFromBackup
+       -- ^ just for testing
        ) where
 
 import           Universum
--- a/wallet/test/Test/Pos/Wallet/Web/Methods/BackupDefaultAddressesSpec.hs
+++ b/wallet/test/Test/Pos/Wallet/Web/Methods/BackupDefaultAddressesSpec.hs
@@ -1,0 +1,35 @@
+module Test.Pos.Wallet.Web.Methods.BackupDefaultAddressesSpec
+       ( spec
+       ) where
+
+import           Universum
+
+import           Data.Default (def)
+import           Pos.Launcher (HasConfigurations)
+import           Pos.Util.CompileInfo (HasCompileInfo, withCompileInfo)
+import           Pos.Wallet.Web.ClientTypes (CWallet(..))
+import           Pos.Wallet.Web.Methods.Backup (restoreWalletFromBackup)
+import           Test.Hspec (Spec, describe)
+import           Test.Hspec.QuickCheck (modifyMaxSuccess)
+import           Test.Pos.Util (assertProperty, withDefConfigurations)
+import           Test.Pos.Wallet.Web.Mode (walletPropertySpec)
+import           Test.QuickCheck (Arbitrary(..))
+import           Test.QuickCheck.Monadic (pick)
+
+spec :: Spec
+spec = withCompileInfo def $
+       withDefConfigurations $
+       describe "restoreAddressFromWalletBackup" $ modifyMaxSuccess (const 10) $ do
+           restoreWalletAddressFromBackupSpec
+
+restoreWalletAddressFromBackupSpec :: (HasCompileInfo, HasConfigurations) => Spec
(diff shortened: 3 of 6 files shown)


wallet/test/Test/Pos/Wallet/Web/Methods/BackupDefaultAddressesSpec.hs
```diff
@@ -0,0 +1,87 @@
...
+
+
+spec :: Spec
+spec = withCompileInfo def $
+       withDefConfigurations $
+       describe "restoreAddressFromWalletBackup" $ restoreWalletAddressFromBackupSpec
+
+restoreWalletAddressFromBackupSpec :: (HasCompileInfo, HasConfigurations) => Spec
+restoreWalletAddressFromBackupSpec = walletPropertySpec restoreWalletAddressFromBackupDesc $ do
+    -- I NEED FIRST TO CREATE WALLETBACKUP -> HOW ?
```
v0d1ch: @gromakovsky I left some code comments so you can see what is the problem for me.

wallet/src/Pos/Wallet/Web/Methods/Backup.hs
```diff
@@ -5,6 +5,8 @@
 module Pos.Wallet.Web.Methods.Backup
        ( importWalletJSON
        , exportWalletJSON
+       -- for testing purposes
+       , restoreWalletFromBackup
```
akegalj: This should not be needed  to export in order to test is `importWalletJSON` and `exportWalletJSON` work as they should

wallet/test/Test/Pos/Wallet/Web/Methods/BackupDefaultAddressesSpec.hs
```diff
@@ -0,0 +1,87 @@
...
+    -- MAYBE LIKE THIS
+    -- WALLETBACKUP CREATION
+    -- wBackup <- lift $ getWalletBackup ((CId (CHash "test")) :: CId Wal)
+
+    -- OR CREATE NEW WALLET ?
```
akegalj: I believe you should (as you did) create an *arbitrary* wallet, export it and reimport it. Then test did you get object with same properties back. Pseudo:
```
do
  wallet <- pick $ arbitrary -- this is pseudo. Read bellow for more details
  filePath <- "some filepath that works on Win/Linux"
  exportWalletJSON wallet filePath
  walletImp <- importWalletJSON filePath
  -- test various properties expected of the two. Also test is imported thing the same as the one we exported with Eq
```

For how to create an arbitrary wallet you can refer to https://github.com/input-output-hk/cardano-sl/blob/master/wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs#L47 `importSomeWallets` and then picking an arbitrary wallet from available wallets with https://github.com/input-output-hk/cardano-sl/blob/master/wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs#L51 : 
```
    rootsWIds <- lift myRootAddresses
    idx <- pick $ choose (0, l - 1)
    let walId = rootsWIds !! idx
let pswd = passphrases !! idx
```

then you can export this wallet ^ and do the rest of tests.

v0d1ch:
Hey @akegalj  you comments were most helpful. Thank You! Now test is working but failing in the `restoreWalletFromBackup` function since there is this
```
    wExists <- isJust <$> getWalletMeta wId
    if wExists
        then do
            throwM $ RequestError "Wallet with id already exists"
```
I am not familiar with the logic here but maybe I need to delete the wallet from the system once I export it  so it can do the import ?

Martoon-00:
@v0d1ch Right, to import wallet X you have to have no X at the moment :slightly_smiling_face: 
`deleteWallet` endpoint would be helpful.

v0d1ch:
I consulted @pva701 and I will change this test to use only `restoreWalletFromBackup` function and avoid saving wallet on disk and importing it again.

v0d1ch:
@gromakovsky I think this is finally ready now. @pva701 can you confirm that all is good ?

wallet/src/Pos/Wallet/Web/Backup.hs
```diff
@@ -11,30 +11,46 @@ import           Universum
...
 data TotalBackup = TotalBackup WalletBackup
 
+instance Arbitrary WalletBackup where
+    arbitrary = do
+        cwNameT <- arbitrary
+        caNameT <- arbitrary
+        wbInt   <- arbitrary
+        let hm = HM.empty
+        (_, esk) <- safeKeyGen emptyPassphrase
+        let wMetaBackup = WalletMetaBackup $ CWalletMeta { cwName = (cwNameT :: Text ), cwAssurance = CWANormal, cwUnit = 1 }
```
akegalj: you could also do `arbitrary` for assurance. By adding arbitrary for assurance:
```haskell
data CWalletAssurance = CWAStrict | CWANormal 
  deriving (Enum, Bounded)

instance Arbitrary CWalletAssurance where
  arbitrary = element [minBound..maxBound]
```

or by simply inserting in your code:
```
  cwAssurance <- element [minBound .. maxBound]
```

Also, I believe we have a syntax sugar on where you should do something like with the above:
```
instance Arbitrary WalletBackup where
+    arbitrary = do
+        cwName <- arbitrary
+        caName <- arbitrary
+        wbInt   <- arbitrary
+        let hm = HM.empty
+        (_, esk) <- safeKeyGen emptyPassphrase
+        let wMetaBackup = WalletMetaBackup $ CWalletMeta {  cwAssurance = CWANormal, cwUnit = 1, ... }
```
and `cwName` will be picked automatically.

wallet/src/Pos/Wallet/Web/Backup.hs
```diff
@@ -11,30 +11,46 @@ import           Universum
...
+    } deriving Show
 
 data TotalBackup = TotalBackup WalletBackup
 
+instance Arbitrary WalletBackup where
+    arbitrary = do
+        cwNameT <- arbitrary
+        caNameT <- arbitrary
+        wbInt   <- arbitrary
+        let hm = HM.empty
```
akegalj: why this definition? You can use just `mempty` on L51

wallet/src/Pos/Wallet/Web/Backup.hs
```diff
@@ -11,30 +11,46 @@ import           Universum
...
+        (_, esk) <- safeKeyGen emptyPassphrase
+        let wMetaBackup = WalletMetaBackup $ CWalletMeta { cwName = (cwNameT :: Text ), cwAssurance = CWANormal, cwUnit = 1 }
+        return $ WalletBackup
+            { wbSecretKey = esk
+            , wbMeta = wMetaBackup
+            , wbAccounts = HM.insert wbInt (AccountMetaBackup $ CAccountMeta { caName = caNameT }) hm
```
gromakovsky: There is `HM.singleton` to create a map with 1 key-value pair.

wallet/src/Pos/Wallet/Web/Backup.hs
v0d1ch: Hey @akegalj thanks for review!
```
cwAssurance <- element [minBound .. maxBound]
```
Did you mean `elements` ?  -> [elements](https://www.stackage.org/haddock/lts-9.14/QuickCheck-2.9.2/src/Test.QuickCheck.Gen.html#elements)
> Also, I believe we have a syntax sugar on

I am not sure what you mean, use `RecordWildCards` ?
```
instance Arbitrary WalletBackup where
    arbitrary = do
        cwNameT <- arbitrary
        caNameT <- arbitrary
        wbInt   <- arbitrary
        cwAssurance <- elements [minBound .. maxBound]
        (_, esk) <- safeKeyGen emptyPassphrase
        let cwUnit = 1
            cwName = (cwNameT :: Text )
            caName = (caNameT :: Text )
            wMetaBackup = WalletMetaBackup $ CWalletMeta {..}
        return $ WalletBackup
            { wbSecretKey = esk
            , wbMeta = wMetaBackup
            , wbAccounts = HM.singleton wbInt (AccountMetaBackup CAccountMeta {..})
            }
```

akegalj: > Did you mean elements ? -> elements

yes

> I am not sure what you mean, use RecordWildCards ?

yes

The example bellow should work:
```
Prelude> :set -XRecordWildCards
Prelude> data Test = Test { t1 :: Int, t2 :: Int } deriving Show
Prelude> t1 = 2
Prelude> Test { t2 = 2, ..}

<interactive>:15:1: warning: [-Wmissing-fields]
    • Fields of 'Test' not initialised: t1
    • In the expression: Test {t2 = 2, ..}
      In an equation for 'it': it = Test {t2 = 2, ..}

<interactive>:15:1: warning: [-Wmissing-fields]
    • Fields of 'Test' not initialised: t1
    • In the expression: Test {t2 = 2, ..}
      In an equation for 'it': it = Test {t2 = 2, ..}
Test {t1 = *** Exception: <interactive>:15:1-18: Missing field in record construction t1
```
but for some unfortunate reason it doesn't work in the repl. Probably its a repl bug. It should return `Test {t1 = 2, t2 = 2}`

v0d1ch: I get it, thanks! So I believe all is ready now

wallet/src/Pos/Wallet/Web/Backup.hs
```diff
@@ -11,30 +11,48 @@ import           Universum
...
-newtype AccountMetaBackup = AccountMetaBackup CAccountMeta
+newtype WalletMetaBackup = WalletMetaBackup CWalletMeta deriving Show
+newtype AccountMetaBackup = AccountMetaBackup CAccountMeta deriving Show
 
 data WalletBackup = WalletBackup
     { wbSecretKey :: !EncryptedSecretKey
     , wbMeta      :: !WalletMetaBackup
     , wbAccounts  :: !(HashMap Int AccountMetaBackup)
-    }
+    } deriving Show
```
adinapoli-iohk: @gromakovsky @pva701 I'm arriving late to the party, but are we sure that using the automatically-derived `Show` instance won't leak the `EncryptedSecretKey` in the logs or somewhere else?

gromakovsky: Yes, because we have
```
instance Show EncryptedSecretKey where
    show _ = "<encrypted key>"

instance B.Buildable EncryptedSecretKey where
    build _ = "<encrypted key>"
```

adinapoli-iohk: Perfect, that was what I wanted to hear, thanks 😉

PR #1978, opened 2017-11-27
[CSL-1470] Remove redundant security worker

volhovm:
pew pew pew

This deletes security worker as it wasn't even launched -- because it was useless, nobody actually noticed that it doesn't work. 

I moved recovery worker (that was in `Pos.Security` for some reason) to `Block.Worker`.


lib/src/Pos/Block/Worker.hs
```diff
@@ -197,6 +200,33 @@ onNewSlotWhenLeader slotId pske SendActions {..} = do
...
+    repeatOnInterval (const $ sec 4) $
+         recoveryCommGuard "security worker" $
+            whenM (needRecovery @ctx) $ triggerRecovery enqueueMsg
+  where
+    -- delay is needed for the system to initialize
+    afterDelay action = delay (sec 3) >> action
+    repeatOnInterval delF action = () <$ do
+        -- REPORT:ERROR 'reportOrLogE' in recovery trigger worker
+        () <$ action `catchAny` \e -> reportOrLogE "Security worker" e
+        getNextEpochSlotDuration >>= delay . delF
```
gromakovsky: Can you simplify it pozh? `delF` is now always `const $ sec 4`.

PR #1984, opened 2017-11-28
[CSL-1925] Define new solution to address generation

et4te:
Summary:
* As requested by @georgeee there is an improved proposal at docs/CSL-1925.md.
* Includes an implementation which passes the current tests using sequential indices.
* These modifications are intended as an example of what would happen if random is taken out of the equation altogether and the amount of work this introduces. The temporary solution is still intended to be CSL-1955.
 
(Rendered [here](https://github.com/et4te/cardano-sl/blob/6c0ef26722e549d9b48da0e2bf4f28d89219ee63/docs/CSL-1925.md) )

### DIFF
--- a/docs/CSL-1925.md
+++ b/docs/CSL-1925.md
@@ -4,24 +4,28 @@
 An address in Cardano is always derived from a root key using BIP32 style derivation
 except it uses a different elliptic curve than Bitcoin (Ed25519). 
 
-The problem initially raised concerns exchanges which have only one account per root
-key and thus 2^31 possible addresses. In this scenario according to the birthday
-paradox there is a non negligible chance of a collision occuring.
+The problem initially raised concerns exchanges which typically use one, single,
+giant wallet per crypto currency and all users share the same space for addresses.
+
+In the HD wallet case this translates to one account per wallet and thus the total
+space for user addresses becomes 2^31. In this scenario according to the birthday
+problem there is a non negligible chance of a collision occuring after a sufficient
+amount of addresses have been generated.
 
 The first child (or branch) of the root key is derived according to the _account index_
-which by default is 2^31 or in binary `10000000000000000000000000000000` where the
-first bit represents the hardened state. As a BIP32 path this can be represented as 
-`m/aH`. The second child is derived according to an _address index_.
+which by default lies within the range (2^31 (`10000000...`) -> 2^30) where the first 
+bit represents the hardened state. As a BIP32 path this can be represented as `m/a'`. 
 
-Both the account and address indices fall within the range 2^31 in terms of how many
-addresses can be derived.
+The second child is derived according to an _address index_ which follows the same 
+rules as the account index for derivation.
 
-In the current scheme the address size is 76 bytes, where each index takes up 5 bytes.
+In the current scheme the serialised form of an address has a size of 76 bytes, where 
+each index takes up 5 bytes (4 bytes for Word32 and 1 byte CBOR overhead).
 
 # Generation Strategies
 
 There are currently two ways of generating a unique address, as depicted by the following
-ADT.
+ADT, which is based on the source code at [wallet/src/Pos/Wallet/Web/Account.hs](https://github.com/input-output-hk/cardano-sl/blob/d7e01933912cdd127b6ffac6bd01e0fb3f15a4f5/wallet/src/Pos/Wallet/Web/Account.hs#L109).
 
 ```haskell
 data GenerationMode 
@@ -33,18 +37,17 @@ data GenerationMode
 
 The difference between `RandomMode` and `DeterministicMode` is that in `DeterministicMode` the
 application / dependent code or user must supply the seed used to generate the index instead 
-of it being generated randomly. `DeterministicMode` is there in order to be able to recover 
-the address at a later date.
+of it being generated randomly. `DeterministicMode` is used during the wallet backup and 
+recovery process in order to recover the generated indices. 
 
 The wallet uses a function called `genUniqueAddress` to generate addresses which in turn
-uses the function shown below named `generateUnique`. In RandomMode this function tries 
+uses the function shown below named `generateUnique`. In `RandomMode` this function tries 
 to ensure that the index used is unique according to an arbitrary predicate `isDuplicate`. 
 For addresses this predicate is called `doesWAddressExist` which checks storage for a 
 pre-existing address.
 
-In terms of BIP32 paths this extra index forms the `bH` in the derivation path `m/aH/bH` 
-where `bH` is the random index and H is there to denote that the function also runs 
-`isHardened` on the random index.
+In terms of BIP32 paths this extra index forms the `b'` in the derivation path `m/a'/b'` 
+where `b'` is the hardened random address index.
 
 ## Random Address Generation
 
--- a/docs/CSL-1925.md
+++ b/docs/CSL-1925.md
@@ -14,8 +14,7 @@ first bit represents the hardened state. As a BIP32 path this can be represented
 `m/aH`. The second child is derived according to an _address index_.
 
 Both the account and address indices fall within the range 2^31 in terms of how many
-addresses can be derived however only the last is generated at random, where the first
-bit is reserved for encoding whether the index is hardened or not.
+addresses can be derived.
 
 In the current scheme the address size is 76 bytes, where each index takes up 5 bytes.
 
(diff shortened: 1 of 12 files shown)


docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+
+An address in Cardano is always derived from a root key using BIP32 style derivation
+except it uses a different elliptic curve than Bitcoin (Ed25519). 
+
+The problem initially raised concerns exchanges which have only one account per root
+key and thus 2^31 possible addresses. In this scenario according to the birthday
+paradox there is a non negligible chance of a collision occuring.
+
+The first child (or branch) of the root key is derived according to the _account index_
+which by default is 2^31 or in binary `10000000000000000000000000000000` where the
```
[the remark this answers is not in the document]
adinapoli-iohk: Are we really sure "which by default is 2^31" is really the case? As you have been from the code, currently during normal operations we generate (modulo bugs 😁 ) a new address index by randomly generating a new index chosen in the hardened address space, but when we restore a wallet, what you say is indeed true: we restores accounts sequentially from `2^31`. Do you think this is a needless implementation detail or should we capture it?

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+first bit represents the hardened state. As a BIP32 path this can be represented as 
+`m/aH`. The second child is derived according to an _address index_.
+
+Both the account and address indices fall within the range 2^31 in terms of how many
+addresses can be derived.
+
+In the current scheme the address size is 76 bytes, where each index takes up 5 bytes.
```
adinapoli-iohk: If we want to be picky, we could mention that is 4 bytes for the `Word32` storage plus 1 byte for CBOR serialisation overhead. Not essential by any means though 😀

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+
+There are currently two ways of generating a unique address, as depicted by the following
+ADT.
+
+```haskell
+data GenerationMode 
+  = RandomMode
+  -- ^ Generates an address where the second index is random and hardened.
+  | DeterministicMode Word32 
+  -- ^ Generates an address where the second index is parameterised.
```
adinapoli-iohk: What about cross-referencing these ones with a link to the source code?

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+The difference between `RandomMode` and `DeterministicMode` is that in `DeterministicMode` the
+application / dependent code or user must supply the seed used to generate the index instead 
+of it being generated randomly. `DeterministicMode` is there in order to be able to recover 
+the address at a later date.
```
adinapoli-iohk: Minor bikeshedding, why not saying "[...]`DeterministicMode` is used during the wallet recovery process at a later date[..]"? This way, we give that `DeterministicMode` a purpose without necessarily implying that this is a necessity for recover (as far as we know, this might or might not be true). Mental?

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+
+The wallet uses a function called `genUniqueAddress` to generate addresses which in turn
+uses the function shown below named `generateUnique`. In RandomMode this function tries
```
adinapoli-iohk: `RandomMode`, so that rendering this in markdown will be prettier 😁

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+of it being generated randomly. `DeterministicMode` is there in order to be able to recover 
+the address at a later date.
+
+The wallet uses a function called `genUniqueAddress` to generate addresses which in turn
+uses the function shown below named `generateUnique`. In RandomMode this function tries 
+to ensure that the index used is unique according to an arbitrary predicate `isDuplicate`. 
+For addresses this predicate is called `doesWAddressExist` which checks storage for a 
+pre-existing address.
+
+In terms of BIP32 paths this extra index forms the `bH` in the derivation path `m/aH/bH`
```
adinapoli-iohk: Here is an idea: as different sources uses different conventions for the hardened/non-hardened keys (I have seen lots of them using apostrophes, like in `m/10'/20'/30'`), should we mention we are sticking to the same way [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#test-vectors) labels them? probably not necessary, but if we omit the link, I would argue it would be good to have the explanation of what `h` stands for at the beginning of the writeup, as we introduce it fairly soon (on line 14, to be precise)

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+
+Lastly `genUniqueAddress` is broken in the above implementation due to isHardened being 
+used inapropriately and being able to simplify the implementation via randomRIO. This will be 
+fixed as a matter of priority as a short term solution to the problem.
+
+# Non-Random Generation
+
+There is less chance of a collision in random generation when generating two
+distinct accounts on the same root key since the index is picked randomly. 
+In practice however this only matters when storage is not available and
```
adinapoli-iohk: That's an excellent point, and I would like us to spend a bit of time thinking more deeply on the implications of what would happen if the following happens:

- I am a large exchange with lots of addresses and a fully synced node, which means `acid-state` is sheltering me from collisions;

- I decide to delete my wallet, nuke my local storage and migrate everything to a new machine, due to server maintenance;

- I start the recovery process on the wallet, but I am still not fully synced.

- I call `POST /api/addresses` to get a branch new address for a registered-user

How would the backend behave? Will it reject the request or allow me to do that, with potentially disastrous consequences?

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
...
+A wallet with an account index of 10 would mean that from 0 to 10 there are 
+11 accounts in this wallet. In the random case one would have to store every
+seed which is much less convenient.
+
+The following algorithm solves the unique account problem sequentially.
+
+```haskell
+nextAccountId
+    :: AccountMode ctx m
+    => CId Wal                   -- FIXME Confusing name
```
adinapoli-iohk: Agreed, is something we can largely improve in V1.

docs/CSL-1925.md
et4te: I would assume perhaps naively that the fourth step here would wait until the sync in step 3 happens atomically. 

If the whole wallet is nuked and the recovery process has not been started then I would assume what would happen is:
1. A new root key is generated for the wallet based on the passphrase of the registered user.
2. This new root key is used to derive the first branch (either at random or sequentially).
3. The key resulting from the first derivation is used to derive the second branch (either at random or sequentially).

So I think actually in the case where the recovery process is atomic there is no problem. In the case where recovery fails altogether, the data may later need to be merged with the recoverable part of the failed data set.

Neither seem disastrous to me and I think that in this scenario it doesn't make a difference if you pick the index randomly or sequentially. I am not sure if I understood correctly your cases here though. :P

docs/CSL-1925.md
```diff
@@ -0,0 +1,256 @@
+
+# Problem Definition
+
+An address in Cardano is always derived from a root key using BIP32 style derivation
+except it uses a different elliptic curve than Bitcoin (Ed25519). 
+
+The problem initially raised concerns exchanges which have only one account per root
```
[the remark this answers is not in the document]
et4te: Yep I agree this needs clarification, I have updated it accordingly.

docs/CSL-1925.md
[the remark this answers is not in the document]
et4te: Ah yes true. I was assuming here that the default case was in the lower level creation of the wallet when in fact it should really reflect how it is done generally. Perhaps I was overly specific in the description, I tried to amend to reflect your points. 

With respect to hex notation, the problem is that in hex it looks like 0x80000000, which doesn't portray very well that there is just 1 bit at the front. As a compromise I shortened the binary notation to (1000000...). Let me know if you would like this changed.

docs/CSL-1925.md
et4te: I modified to include these finer details. :)

docs/CSL-1925.md
et4te: Yes this makes sense, I will stick to BIP-32 labelling and change the document accordingly to reflect this.

docs/CSL-1925.md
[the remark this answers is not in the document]
adinapoli-iohk: > With respect to hex notation, the problem is that in hex it looks like 0x80000000, which doesn't portray very well that there is just 1 bit at the front.

Agreed, maybe going with hex is misleading. Let's stick to binary as you first proposed 😉

adinapoli-iohk:
Stale at this point, so closing it.

If we want to brainstorm on some of the ideas in this PR we can do so even if it's closed, no need to keep it lingering.

PR #1989, opened 2017-11-28
[CSE-214] Socket's subscription / unsubscription specs

sectore:
Tests of:
* `subscribeAddr`
* `unsubscribeAddr`
* `subscribeBlocksLastPage`
* `unsubscribeBlocksLastPage`
* `subscribeTxs`
* `unsubscribeTxs`
* `subscribeEpochsLastPage`
* `unsubscribeEpochsLastPage`
* `unsubscribeFully`

_How to run?_
* `stack test cardano-sl-explorer --fast --test-arguments "-m Test.Pos.Explorer.Socket"`


explorer/test/Test/Pos/Explorer/Socket/MethodsSpec.hs
```diff
@@ -68,3 +122,191 @@ addressSetByTxsProp key =
...
+unsubscribeAddrProp =
+    forAll arbitrary $ \(socketId, addr) ->
+        monadicIO $ do
+            let cAddr = toCAddress addr
+            let connState = mkSubConnectionState socketId
+            -- Add a subscription first ...
+            let subscription = runSubTestMode connState $
+                                  subscribeAddr cAddr socketId
+            (_, updatedConnState) <- run subscription
+            -- ... and remove subscription
```
ksaric: You could `assert` in the middle of the tests, since it could simply not work and the end result would be `M.size sessions == 0`?

sectore: As discussed in Slack w/ Kristijan and me: Extra `assert` would compare same things as we already have with `subscribe...` tests. So there is no need for doing it here.

PR #1990, opened 2017-11-28
[CSL-1939] Check that swagger is spec is valid and up to date

Martoon-00:
Make CI check `wallet-new/spec/swagger.json`.
* It will ensure that swagger is valid via `http://online.swagger.io/validator` online API.
(If you disagree with using online API, please let's discuss this point over further, actually I feel a bit unstable about using such things)
* Also it will regenerate its own copy of doc, and compare that `wallet-new/spec/swagger.json` matches to it.

To accomplish this task, executable to generate swagger doc was added.
To avoid need in building tons of executables (okay, currently only extra one, but still), option to build with `--flag cardano-sl-wallet-new:exclude-aux` was added

### DIFF
--- a/scripts/ci/validate-wallet-swagger.sh
+++ b/scripts/ci/validate-wallet-swagger.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+
+if [[ $TRAVIS_COMMIT == "" ]]; then
+    echo "TRAVIS_COMMIT is unspecified"
+    exit 1
+fi
+
+spec_path="https://raw.githubusercontent.com/input-output-hk/cardano-sl/$TRAVIS_COMMIT/wallet-new/spec/swagger.json"
+validate="curl http://online.swagger.io/validator/debug?url=$spec_path"
+
+echo "Checking swagger spec at $spec_path"
+errors=$(curl $validate 2> /dev/null)
+
+if [[ $errors == '{}' ]]; then
+    echo "Swagger spec is valid"
+else
+    echo "Swagger spec is invalid!: $errors"
+    exit 1
+fi
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -12,6 +12,12 @@ category:            Web
 build-type:          Simple
 cabal-version:       >=1.10
 
+Flag server-only
+  default:     False
+  manual:      True
+
+  description: Build only components required for server
+
 library
   hs-source-dirs:      src
   exposed-modules:     Cardano.Wallet.Orphans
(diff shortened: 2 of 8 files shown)


wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -12,6 +12,12 @@ category:            Web
 build-type:          Simple
 cabal-version:       >=1.10
 
+Flag exclude-aux
```
adinapoli-iohk: Can we make this flag more descriptive?

wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -29,6 +35,7 @@ library
                        Cardano.Wallet.API.V1.Accounts
                        Cardano.Wallet.API.V1.Transactions
                        Cardano.Wallet.API.V1.Settings
+                       Cardano.Wallet.API.V1.Swagger
```
adinapoli-iohk: Do you think there is any way we can avoid pulling `V1.Swagger` into the library? It really shouldn't belong there.

In theory, you should be able to simply create your `wallet-new-swagger` by depending on `Cardano.wallet.API.V1.Swagger` without moving it if you include `server` in the list of `hs-source-dirs`. have you tried that?

Martoon-00: Sure, but looks like cabal dislikes when executables depend on other executables

adinapoli-iohk: But in this case you wouldn't depend from another executable, you would simply include the `server` source code into `hs-source-dirs`, which is completely executable-agnostic, in this case. What error you get if you try that?

Martoon-00: Really, it worked, thanks!

Martoon-00:
One thing I notices now is this warning on `stack build`:
```
There were multiple candidates for the Cabal entry "Main.hs" (/home/martoon/serokell/cardano-sl/wallet-new/server/Main.hs), picking /home/martoon/serokell/cardano-sl/wallet-new/spec/Main.hs
```
AFAIU it means that the only way to launch server remained is `stack exec wallet-new-server`. How much is it bad?

wallet-new/cardano-sl-wallet-new.cabal
Martoon-00: How about `server-only`?

adinapoli-iohk:
> How much is it bad?

I think this is benign and you should be able to still run `stack install` and have two nice executables `wallet-new-server` and `wallet-new-swagger` both in your stack path. Did you verify that's the case. By the experience I'm decently confident this shouldn't cause harm, but just in case I'm happy for @domenkozar to chime in :)

wallet-new/cardano-sl-wallet-new.cabal
adinapoli-iohk: Yep, sounds better :)

Martoon-00:
Yeah, it works after `stack install` as you noted, great )

Martoon-00:
Then fixes applied, waiting for the green flag )

scripts/ci/validate-wallet-swagger.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+
+if [[ $TRAVIS_COMMIT == "" ]]; then
+    echo "TRAVIS_COMMIT is unspecified"
+    exit 1
+fi
+
+spec_path="https://raw.githubusercontent.com/input-output-hk/cardano-sl/$TRAVIS_COMMIT/wallet-new/spec/swagger.json"
```
adinapoli-iohk: Very cool, this will also allow us to reference to this version externally in `editor.swagger.io`, which would be a boon in reviewing new API endpoints :)

wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -12,6 +12,12 @@ category:            Web
 build-type:          Simple
 cabal-version:       >=1.10
 
+Flag server-only
+  default:     False
+  manual:      True
+
+  description: Build only components required for server
```
adinapoli-iohk: Much better!

adinapoli-iohk:
@cleverca22 , you have been appointed from Domen as his lieutenant in his absence 😁 

A report from the trenches from Domen, @Martoon-00 , is that this patch won't work as-it-is because it's using `stack exec`, and we never build `stack` on the CI, according to Domen. Surely Michael can expand a bit more and help us out 😀

Martoon-00:
Closed in favour of #2068

PR #1991, opened 2017-11-28
[CSL-1955]  Improve `generateUnique` function

et4te:
Summary:
* Updated generateUnique to always generate hardened addresses and removed orM.


flyingleafe:
Genuine bug indeed, no hidden meaning, confirmed by @Martoon-00. Shameful.

PR #1992, opened 2017-11-28
[CSL-1968] Implement GET /api/v1/user-profile

adinapoli-iohk:
This PR implements what was called `/api/v1/profile` in the old API.

@DominikGuzei you will see I have made a precise choice: the default in case no locale is set is `en-US`, as taken from:

https://github.com/input-output-hk/daedalus/blob/master/translations/translation-runner.js#L7

I hope that's correct! This is how it looks like in Swagger:

screen shot 2017-11-28 at 17 43 46

### DIFF
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -33,6 +33,7 @@ library
                        Cardano.Wallet.API.V1.Wallets
                        Cardano.Wallet.API.V1.Errors
                        Cardano.Wallet.API.V1.Updates
+                       Cardano.Wallet.API.V1.Users
                        Cardano.Wallet.API.V1.Migration
                        Cardano.Wallet.API.V0
                        Cardano.Wallet.API.V0.Types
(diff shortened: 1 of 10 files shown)


DominikGuzei:
@adinapoli-iohk to be honest: i don't think we are using any `profile` server endpoint in Daedalus as we are saving that locally on the user's machine 😟 

@nikolaglumac that's right isn't it?

adinapoli-iohk:
If that's the case guys, even better -- the best code is the one you don't have to write 😉 

Just let me know, and will scrap these for good!

nikolaglumac:
@DominikGuzei that is correct. The problem was that saving or fetching info such as user locale from the backend was too slow - e.g. when application is starting we need to have this information straight away in order to show loading message in user's locale so we decided to save all profile info to the local storage layer.

adinapoli-iohk:
Even better guys. Case closed 😉 

Cheers!

akegalj:
@nikolaglumac that's even better.

Can we improve this communication in future for request removal if Daedalus UI decides not to use some endpoints (or features). That would help us maintain clean codebase

nikolaglumac:
@akegalj of course we can and of course we should 👍

PR #1995, opened 2017-11-29
Attempt to fix CI

gromakovsky:


stack.yaml
```diff
@@ -154,6 +154,10 @@ extra-deps:
 - megaparsec-6.2.0
 - parser-combinators-0.2.0
 - loc-0.1.3.1
+# hscolour-1.24.1 is the version from the LTS we use (9.1)
```
domenkozar: No need for this, hscolour is now hardcoded in stack2nix. Also it doesn't work

gromakovsky:
@domenkozar CI passes now (only macos builds haven't finished yet, but they have already passed the dangerous point). Can you approve please? I left only necessary changes.

volhovm:
Should be good!

PR #1997, opened 2017-11-29
[CSL-1970] Old & new wallet executable's behaviours are not consistent

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-1970

I'm not sure if it's my luck or what, but I always end up with some bizarre bug that makes no sense.

Please test this out:
1. You build it `stack build --fast -j4`
2. You open `tmux` and run script `./scripts/launch/demo-with-new-wallet-api.sh`
3. Wait for 30 seconds.
4. On new tab/terminal run `curl -X GET --cacert /home/ksaric/projects/curry/cardano-sl/scripts/tls-files/ca.crt https://localhost:8090/api/settings/sync/progress` (**change the path, please**)

All good? Right.
Now remove the `-threaded` option from the cabal file.

Repeat the test from above :up: 

The files are from original _wallet_ implementation and I keep getting results like - `{"Right":{"_spLocalCD":{"getChainDifficulty":{"getBlockCount":1}},"_spNetworkCD":{"getChainDifficulty":{"getBlockCount":24}},"_spPeers":0}}` when using the new configuration.

Fun times.

In the end, I ended up using Daedalus integration test which @nikolaglumac provided (it's currently using the branch `feature/etc-client`, but it should be merged to `master` soon I think). This branch doesn't require the daedalus-bridge, you can run it with `yarn run test` (or `npm` if you are using `npm`).
All tests pass, the backed works like it used to, and we are all happy.


adinapoli-iohk:
@ksaric I'm very sorry, but I'm struggling to understand what's going on here.

Something I have noticed is that you added a new file called `server/NodeOptions.hs` which is using the old CLI, how come? The reason I'm asking is because `server/CLI.hs` has one subtle difference: it's writing on a different DB (aka `wallet-db-new`) to not "pollute" the existing one. I might speculate this is the bizarre behaviour you are experiencing.

My recommendation would be to stick to the _new_ CLI and not commit the old one into the new code. Unless I'm missing something crucial, why do you think it's necessary?

ksaric:
The point of this PR is to check whether the "old" wallet code works. To prove that, I copied the code from "old" *wallet* and tried to run it. I choose that route since it's hard to debug the existing code. The plan was to modify the code from the "old" to "new" format and to catch the cause of this behaviour that way.
The result is that even using old *wallet* fails, even though it works in the "old" code - in the original wallet package, and *when you change the cabal file*.
If you follow the instructions you should see the weird behaviour yourself.

ksaric:
@adinapoli-iohk The issue was that the cabal file was missing the `-threaded` option :rage: 
Updated the issue comment.

adinapoli-iohk:
> The issue was that the cabal file was missing the -threaded option

Great catch!

> Also increased the waiting time to 20 seconds, since we are now generating swagger documentation and I experienced node timeouts several times.

Uhm, but that is weird, the swagger doc generation should bring very little overhead. Have you tried temporarily disabling the generation? Did you observe any difference?

ksaric:
> Uhm, but that is weird, the swagger doc generation should bring very little overhead. Have you tried temporarily disabling the generation? Did you observe any difference?

No, but that's not necessarily the case, I simply had timing issues - not sure what the underlying cause is, but it was a necessary change. Maybe my machine? :man_shrugging:

adinapoli-iohk:
Yea, possibly, go figure!

nikolaglumac:
@ksaric thanks for the notification. The branch which you used to testing will be merged into master later today as a part of this epic pull request: https://github.com/input-output-hk/daedalus/pull/565 and then you can once again use master branch for testing ;)

ksaric:
Reporting `Test suite failure for package cardano-sl-wallet-new-0.1.0.0` on AppVeyor, while locally It runs successfully like Travis shows. Merging.

nikolaglumac:
@ksaric you can now use Daedalus `master` branch. Everything is merged 🎉

PR #2000, opened 2017-11-29

volhovm:
This should speed up `MsgBlock` listener on the server side.

I've also fixed a recovery worker that wasn't triggered when we were not in recovery, though too late (it was a bug, apparently).

### DIFF
--- a/node/src/Pos/Block/Logic/Header.hs
+++ b/node/src/Pos/Block/Logic/Header.hs
@@ -9,15 +9,15 @@ module Pos.Block.Logic.Header
        , classifyHeaders
        , getHeadersFromManyTo
        , getHeadersOlderExp
-       , getHeadersFromToIncl
+       , getHeadersRange
        ) where
 
 import           Universum
+import           Unsafe                    (unsafeLast)
 
-import           Control.Lens              (_Wrapped)
+import           Control.Lens              (to, _Wrapped)
 import           Control.Monad.Except      (MonadError (throwError))
 import           Control.Monad.Trans.Maybe (MaybeT (MaybeT), runMaybeT)
-import           Data.List.NonEmpty        ((<|))
 import qualified Data.Text                 as T
 import           Formatting                (build, int, sformat, (%))
 import           Serokell.Util.Text        (listJson)
@@ -31,11 +31,10 @@ import           Pos.Block.Pure            (VerifyHeaderParams (..), verifyHeade
 import           Pos.Configuration         (HasNodeConfiguration, recoveryHeadersMessage)
 import           Pos.Core                  (BlockCount, EpochOrSlot (..),
                                             HasConfiguration, HeaderHash, SlotId (..),
-                                            blkSecurityParam, difficultyL, epochOrSlotG,
-                                            getChainDifficulty, getEpochOrSlot,
-                                            headerHash, headerHashG, headerSlotL,
-                                            prevBlockL)
-import           Pos.Core.Configuration    (genesisHash)
+                                            blkSecurityParam, difficultyL, epochIndexL,
+                                            epochOrSlotG, getChainDifficulty,
+                                            getEpochOrSlot, headerHash, headerHashG,
+                                            headerSlotL, prevBlockL)
 import           Pos.Crypto                (hash)
 import           Pos.DB                    (MonadDBRead)
 import qualified Pos.DB.Block              as DB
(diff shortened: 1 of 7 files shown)


georgeee:
Let's merge it after `CSL-1866 Benchmark block retrieval speed after CSL-1931`

node/src/Pos/GState/BlockExtra.hs
```diff
@@ -153,6 +154,29 @@ loadUpWhile morph start condition = OldestFirst . reverse <$>
...
+           HeaderHash
+        -> Int
+        -> [HeaderHash]
+        -> m [HeaderHash]
+    loadUpWhileDo curH height !res = do
+        mbNextLink <- resolveForwardLink curH
+        if | not (condition curH height) -> pure res
+           | Just nextLink <- mbNextLink ->
+               loadUpWhileDo nextLink (succ height) (curH:res)
+           | otherwise -> pure res
```
gromakovsky: Why not
```
+loadHashesUpWhile :: (MonadBlockDBRead m, HasHeaderHash a)
+    => a
+    -> (HeaderHash -> Int -> Bool)
+    -> m (OldestFirst [] HeaderHash)
+loadHashesUpWhile start condition =
+    loadUpWhile (pure . Just) identity start condition
```
?

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -328,35 +327,84 @@ getHeadersOlderExp upto = do
...
--- range @[from..to]@ will be found.
-getHeadersFromToIncl
-    :: forall ssc m .
-       (HasConfiguration, MonadDBRead m, SscHelpersClass ssc)
-    => HeaderHash -> HeaderHash -> m (Maybe (OldestFirst NE HeaderHash))
-getHeadersFromToIncl older newer = runMaybeT . fmap OldestFirst $ do
+-- | Given optional @depthLimit@, @from@ and @to@ headers where @from@
+-- is older (not strict) than @to@, and valid chain in between can be
+-- found, headers in range @[from..to]@ will be found. If difference
+-- between headers is more than @depthLimit@, error will be thrown.
```
gromakovsky: Imo, this interface is strange. I find it more intuitive to limit number of headers.
Also there is a problem: eventually you indeed want to limit number of headers, but since you limit difference in difficulty it can happen that there are more headers, because there is a genesis header in between.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -328,35 +327,84 @@ getHeadersOlderExp upto = do
...
+
+    -- How many epochs does this range cross.
+    let genDiff :: Int
+        genDiff = fromIntegral $ newerHd ^. epochIndexL - olderHd ^. epochIndexL
+    -- Number of blocks is difficulty difference + number of genesis blocks.
+    let depthDiff :: Int
+        depthDiff = genDiff + fromIntegral (newerD - olderD)
+
+    -- We load these depthDiff blocks.
+    let cond _curHash depth = depth < depthDiff
```
gromakovsky: It's harder to make an off-by-one mistake if the condition is formulated as `_curHash /= newer`.

node/src/Pos/GState/BlockExtra.hs
volhovm: Because `loadUpWhile` uses `foldlUpWhileM` which uses `blkGetBlund` thus getting extra data we don't need. I realized it only while rebasing on cardano-sl-1.0 in fact hehe.

node/src/Pos/Block/Logic/Header.hs
volhovm: Ok, you won. Since I explicitly check difference in `guard` i can change this condition.

node/src/Pos/Block/Logic/Header.hs
volhovm: @gromakovsky nice catch. I'll fix that.

node/src/Pos/GState/BlockExtra.hs
gromakovsky: Ah ok. It might be better to add yet another argument to `foldUpWhileM` to avoid doing `blkGetBlund` unconditionally there 😜

volhovm: Yes, agreed. Let's abstract. I like overabstractions.

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -247,10 +244,14 @@ syncWalletWithGStateUnsafe encSK wTipHeader gstateH = setLogger $ do
...
-            trackingApplyTxs encSK allAddresses mDiff blkHeaderTs ptxBlkInfo $
-            zip3 (gbTxs b) (undoTx u) (repeat $ getBlockHeader b)
+        applyBlock :: [CWAddressMeta] -> HeaderHash -> m (Maybe (CAccModifier,Block ssc))
+        applyBlock allAddresses hh = do
+            blundM <- DB.blkGetBlund hh
+            pure $ blundM <&> \(b,u) ->
+                let cam = trackingApplyTxs
+                              encSK allAddresses mDiff blkHeaderTs ptxBlkInfo
+                              (zip3 (gbTxs b) (undoTx u) (repeat $ getBlockHeader b))
+                in (cam,b)
```
pva701: what the heck? why do you need block `b`, why not to return only `cam`? I was looking at the code about 10 mins and didn't get why it's needed, however, if it's needed, are you sure it's not enough to return only block header? block takes more space in RAM.

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -259,16 +260,22 @@ syncWalletWithGStateUnsafe encSK wTipHeader gstateH = setLogger $ do
                 sformat ("Wallet "%build%" header: "%build%", current tip header: "%build)
                 wAddr wHeader gstateH
             if | diff gstateH > diff wHeader -> do
+                     let loadCond (_cam,b) _ = b ^. difficultyL <= gstateH ^. difficultyL
+                         mappendR (rr,bb) (r,b) = pure (r <> rr,b : bb)
```
pva701: no, you've broken this logic. it should be `rr <> r` as before. it's not associative, you can realize it if you try to mappend two `UtxoModifier`s (which is stored into `CAccModifier`)

volhovm:
@pva701 I've addressed your changes.

node/src/Pos/Block/Logic/Header.hs
gromakovsky: `difference between headers` is really obscure :/
Do you mean difference between their `ChainDiffuclties` or just number of headers?
I can figure out by looking at the implementation, but the documentation should be enough to understand the meaning of the first argument.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -328,35 +327,86 @@ getHeadersOlderExp upto = do
...
+
+    -- We load these depthDiff blocks.
+    let cond curHash _depth = curHash /= newer
+
+    -- This is [oldest..newest) headers, oldest first
+    allExceptNewest <- GS.loadHashesUpWhile older cond
+
+    -- Sometimes we will get an empty list, if we've just switched the
+    -- branch (after first checks are performed here) and olderHd is
+    -- no longer in the main chain.
```
gromakovsky: If you are going to make changes, you can put a reference to the issue you recently created about this piece of code. It's quite minor of course.

node/src/Pos/Block/Network/Listeners.hs
```diff
@@ -64,9 +65,10 @@ handleGetBlocks oq = listenerConv oq $ \__ourVerInfo nodeId conv -> do
     whenJust mbMsg $ \mgb@MsgGetBlocks{..} -> do
         logDebug $ sformat ("handleGetBlocks: got request "%build%" from "%build)
             mgb nodeId
-        mHashes <- getHeadersFromToIncl @ssc mgbFrom mgbTo
-        case mHashes of
-            Just hashes -> do
+        -- We fail if we're requested to give more than
+        -- recoveryHeadersMessage headers at once.
+        getHeadersRange @ssc (Just $ recoveryHeadersMessage - 1) mgbFrom mgbTo >>= \case
```
gromakovsky: AFAIU, currently interface of `getHeadersRange` is that the first argument is the maximal number of headers (though I am not sure and wrote that I don't like the documentation). So I guess you don't need `- 1` here.

volhovm:
@gromakovsky I've addressed things you've asked me to.

PR #2001, opened 2017-11-29
CSL-1871 Restructure cardano-sl-core

gromakovsky:
Commits are quite descriptive (imo), so I recommend reviewing it per-commit.


core/Pos/Binary/Core/Common.hs
```diff
@@ -0,0 +1,76 @@
...
+        T.mkCoinPortion <$> (decode @Word64) >>= \case
+            Left err          -> fail err
+            Right coinPortion -> return coinPortion
+
+instance Bi T.BlockCount where
+    encode = encode . T.getBlockCount
+    decode = T.BlockCount <$> decode
+
+-- serialized as vector of TxInWitness
+--instance Bi T.TxWitness where
```
adinapoli-iohk: Do you think it's worthwhile keeping this comment, if this is genuinely-unused code?

gromakovsky: Good catch, it's indeed obsolete (was obsolete even before my PR), I will delete this comment.

PR #2002, opened 2017-11-29
[CSL-1962] Implement endpoint in order to check the validity of addresses

et4te:
Summary:
* Added verifyAddress function as /addresses/x/validity route based on prior check plus AddressSpec.
* Works the same as wallet old code except also checks the length. Length should probably come from a global constant but couldn't find one.
* Ignores errors in decodeAddress since I assume we only care whether the address is valid or not.
* A couple of comment lines were added just for visual differentiation between Address / Account, can remove if necessary. I imagine these types will be refactored to their own modules eventually.


adinapoli-iohk:
@et4te This is utterly my fault for not making this clear with everybody, but would you be up for rebasing this on `alfredo/wallet-new`? Essentially, currently the protocol is for V1 code to live in the feature-branch called `alfredo/wallet-new`, from which we all branch to do new work and PRs. In a nutshell, the workflow is:

* `git checkout alfredo/wallet-new`
* `git checkout -b et4te/cslXXX-small-description`
* Open PR using as target branch `alfredo/wallet-new`
* We merge the PR
* We periodically merge `alfredo/wallet-new` into `master`, trying to stick to the [git flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) pattern.

Sorry for the confusion!

et4te:
Hey Alfredo, I rebased onto alfredo/wallet-new however now it includes all of those changes hmm...

akegalj:
> Hey Alfredo, I rebased onto alfredo/wallet-new however now it includes all of those changes hmm...

try opening a new branch from `alfredo/wallet-new` and cherry-picking your commits there

adinapoli-iohk:
@et4te Yes, that's the crucial bit, you need to fork _from_ `alfredo/wallet-new` to not have such humongous diff. A nice way to achieve that is to do what @akegalj was suggesting :)

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
```diff
@@ -31,3 +33,18 @@ listAddresses PaginationParams {..} =
...
 newAddress :: Address -> Handler Address
 newAddress a = return a
+
+-- | Verifies that an address is < 78 and base58 decodable.
+verifyAddress :: Text -> Handler AddressValidity
+verifyAddress address
+  | length address > maxAddressSize =
+      return $ AddressValidity False
+  | otherwise =
+      case decodeTextAddress address of
```
adinapoli-iohk: As we are discarding the content of the results, maybe this is written more concisely as 

``` haskell
AddressValidity $ either (const False) (const True) (decodeTextAddress address)
```

Maybe no, it's horrid, actually 😂

wallet-new/src/Cardano/Wallet/API/V1/Addresses.hs
```diff
@@ -13,3 +14,6 @@ type API = "addresses" :> WalletRequestParams
       :<|> "addresses" :> ReqBody '[JSON] Address
                        :> Summary "Creates a new Address."
                        :> Post '[JSON] Address
+      :<|> "addresses" :> Capture "address" Text
+                       :> Summary "Checks the validity of an address."
+                       :> Get '[JSON] AddressValidity
```
adinapoli-iohk: I'm wondering if this is better expresses as a `PUT`, because technically this is idempotent and doesn't query directly the model. On the other hand, it feels artificial, because it's really not an update of the model either. @DominikGuzei any preference from you folks?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -316,6 +321,19 @@ deriveJSON Serokell.defaultOptions ''Address
 instance Arbitrary Address where
   arbitrary = Address . fromString <$> elements ["DEADBeef", "123456"]
 
+-- | Whether an address is valid or not.
+newtype AddressValidity = AddressValidity { getValid :: Bool }
```
adinapoli-iohk: What about renaming `getValid` to either `isValid` or `valid`?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -316,6 +321,19 @@ deriveJSON Serokell.defaultOptions ''Address
...
+  deriving (Eq, Show, Generic)
+
+deriveJSON Serokell.defaultOptions ''AddressValidity
```
adinapoli-iohk: I'm afraid the stock `deriveJSON` instance won't cut it, as users would receive this slightly-warty JSON:

```haskell
{ 
  "getValid": true
}
```

Whereas we could have something like:

```haskell
{ 
  "valid" : true
}
```

Which looks better (at least to my eyes, which I appreciate might not be the gold standard 😁  )

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -316,6 +321,19 @@ deriveJSON Serokell.defaultOptions ''Address
...
+
+instance Arbitrary AddressValidity where
+  arbitrary = AddressValidity <$> elements [True, False]
```
adinapoli-iohk: We don't need to specify the full spectrum of possible values, cannot we simply write `arbitrary = AddressValidity <$> arbitrary`? This will simply piggyback on the `Arbitrary` instance for `Bool`, which is basically the same as writing this rolled-out-by-hand instance 😁

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
adinapoli-iohk: You can skip the double `return $` if you wrap the `case` block into `return` this way:

``` haskell
return $ case foo of
  Left _ -> True
  Right _ -> False
```

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
et4te: This was one of the conventions in Serokell where they say you should use getX in newtype constructors. I agree with you though that isValid makes more sense.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
et4te: Are you sure this happens with the Serokell defaultOptions? I looked here: https://github.com/serokell/serokell-util/blob/9f9715763134e41c9ba92a1359be92f97f93eea8/src/Serokell/Aeson/Options.hs. It looks as if the generated JSON should end up being `{ "valid" : true }` unless I am confused?

adinapoli-iohk: I think you might be right -- although it doesn't play very well with the `getXX` convention, sigh.

wallet-new/src/Cardano/Wallet/API/V1/Addresses.hs
et4te: Hmm I think one of my commits got lost somewhere because this is actually wrong. Should be a capture and then "validity" after the capture (GET /addresses/x/validity). I am not sure about the PUT in this context though I usually associate PUTs with having some effect on the state like persisting some value.

PR #2004, opened 2017-11-30
[CSL-1927] Update V1 new revision

adinapoli-iohk:
@georgeee , @domenkozar , here is the proposed new revision for the update API for V1, which includes the new endpoints to propose & vote for an update. You might notice the API is quite different from the one Domen originally proposed and that I proposed as part of [this PR](https://github.com/input-output-hk/cardano-sl/pull/1884). This is because I have had time to think about @georgeee 's review and I am now much more happy of how the API looks & feels as it's way more RESTful in my opinion, and all the custom ad-hoc actions (except for one, hehe) are gone.

[Play with the rendered API here!](https://editor.swagger.io/?url=https://gist.githubusercontent.com/adinapoli-iohk/b8d54d04101e60e83166cb32dfa9d8ff/raw/2e8193777aea781d01c0aa5ac96f50d9ef79d640/updates_v1_rev1.yaml#)

Here is a glimpse of the endpoints:

screen shot 2017-11-30 at 17 06 15

You will see how I have conceptually separated the concept of a `WalletSoftwareUpdate` (which technically represents a confirmed `UpdateProposal`) from the concept of the `UpdateProposal` itself which is something that everybody (but for now just devops) can propose.

The vote endpoint now takes the `stakeKey` into account, as @georgeee requested.

Apart from that, I feel we are close to the final revision. I have a question for you: am I right that the schema Domen is proposing (with the addition of those `installers` etc) would need to be reflected in the code by modifying the current [UpdateProposal](https://github.com/input-output-hk/cardano-sl/blob/8507d03ba928e07daea57f9a52f6dc03a9d65779/core/Pos/Core/Update.hs#L176) datatype?

Let me know!

### DIFF
--- a/wallet-new/docs/updates.md
+++ b/wallet-new/docs/updates.md
@@ -236,12 +236,14 @@ data WalletSoftwareUpdate = WalletSoftwareUpdate
     , wsuNegativeStake   :: !CCoin
     , wsuSeverity        :: !UpdateSeverity
     -- ^ The severity of the update (i.e. optional or mandatory).
-    , wsuUpdateState     :: !UpdateState
+    , wsuState     :: !UpdateState
     -- ^ The state of this update from users point of view.
-    , wsuSha256Checksum  :: !Sha256Checksum
-    -- ^ The checksum expected for this installer. It will probably
-    -- need to be passed as part of the `UpdateProposal` (maybe in an attribute?),
-    -- or we could reuse the hash which is part of the name, if possible.
+    , wsuDownloadProgress  :: !Word8
+    -- ^ A progress in percentage (from 0 to 100).
+    -- N.B. In the real implementation this can be refined to a better type.
+    , wsuId  :: !Hash
+    -- ^ The update id, which is also the checksum expected for this installer,
+    -- and correspond to the @Blake2b256@ hash of the installer.
     } deriving (Eq, Show, Generic, Typeable)
 ```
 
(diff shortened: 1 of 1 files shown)


cleverca22:
appveyor test failure
```
Test suite failure for package cardano-sl-wallet-new-0.1.0.0
    wallet-new-specs:  exited with: ExitFailure (-1073741515)
```

adinapoli-iohk:
> appveyor test failure

@cleverca22 Yea, me and @deepfire are scratching our heads on this one. The tests regularly passes on Darwin & Linux but they seems to stall on Win, which does make no sense as those are simple Aeson roundtrips -- I have asked @rockbmb to give `alfredo/wallet-new` branch a spin hoping to capture something more in the logs.

wallet-new/docs/updates.md
```diff
@@ -252,37 +254,40 @@ data UpdateSeverity = Mandatory
                     | Optional
 ```
 
-For hard forks updates (or security-critical issues) we might want to set the severity
-to `Mandatory`, for soft forks and other non-critical updates we can label an update as `Optional`.
+For security-critical issues we might want to set the severity
```
georgeee: This should spawn a task in https://iohk.myjetbrains.com/youtrack/issue/CS-113

wallet-new/docs/updates.md
```diff
@@ -236,12 +236,14 @@ data WalletSoftwareUpdate = WalletSoftwareUpdate
...
+    , wsuState     :: !UpdateState
     -- ^ The state of this update from users point of view.
-    , wsuSha256Checksum  :: !Sha256Checksum
-    -- ^ The checksum expected for this installer. It will probably
-    -- need to be passed as part of the `UpdateProposal` (maybe in an attribute?),
-    -- or we could reuse the hash which is part of the name, if possible.
+    , wsuDownloadProgress  :: !Word8
+    -- ^ A progress in percentage (from 0 to 100).
+    -- N.B. In the real implementation this can be refined to a better type.
+    , wsuId  :: !Hash
```
georgeee: Fine!

wallet-new/docs/updates.md
```diff
@@ -497,179 +520,310 @@ The way updates will transition from the *next* to the *current* can be the foll
...
+      - proposalId
+  UpdateProposalVote:
+    type: object
+    properties:
+      vote:
+        $ref: '#/definitions/Vote'
+      stakeKey:
+        type: string
+        description: "The delegation stake key for this node."
+        example: "A8ADD4BDDDFD93E4877D2746E62817B116364A1FA7BC148D95090BC7333B3673"
```
georgeee: It should not be a stake key, but rather a reference. Also, this would be useful only if user has multiple stake keys on his node available it it's the case.

I would say it's questionable whether we want it done this way (explicitly pass stake key). My initial complaint was more about you have to define which stake key would be used. I.e. show user that he send vote with this stake key (b/c this one is used by his node for operations).

georgeee: BTW, Update proposal also requires stake key to be created

wallet-new/docs/updates.md
adinapoli-iohk: Done! https://iohk.myjetbrains.com/youtrack/issue/CSL-2012

adinapoli-iohk:
Re-opening against the correct feature branch.

PR #2008, opened 2017-11-30
[CSL-1961] Implement `new address` endpoint

Martoon-00:
In the process I dared to change API of this endpoint, please point out whether it's correct or I went too far with changes :slightly_smiling_face: 
In particular, I made endpoint accepting `AccountId` and returning summary about newly created address, as old API did.

Also, marshaling test is failing for `NewAddress` due to incorrect JSON instances for `PassPhrase`, fix is in #1957

### DIFF
--- a/wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
@@ -226,11 +226,15 @@ instance ToDocs Account where
   readOnlyFields   = readOnlyFieldsFromJSON
   descriptionFor _ = "An Account."
 
+instance ToDocs WalletAddress where
+  readOnlyFields   = readOnlyFieldsFromJSON
+  descriptionFor _ = "An Address."
+
 instance ToDocs AccountUpdate where
   descriptionFor _ = updateDescr (Proxy @Account)
 
 instance ToDocs Address where
-  descriptionFor _ = "An Address."
+  descriptionFor _ = "An Address ID."
 
 instance ToDocs WalletId where
   descriptionFor _ = "A Wallet ID."
@@ -240,6 +244,9 @@ instance ToDocs Wallet where
 
 instance ToDocs NewWallet where
   descriptionFor _ = newDescr (Proxy @Wallet)
+
+instance ToDocs NewAddress where
+  descriptionFor _ = newDescr (Proxy @WalletAddress)
 
 instance ToDocs WalletUpdate where
   descriptionFor _ = updateDescr (Proxy @Wallet)
@@ -294,6 +301,9 @@ possibleValuesOf (Proxy :: Proxy a) = T.intercalate "," . map show $ ([minBound.
 instance ToSchema Account where
   declareNamedSchema = annotate fromArbitraryJSON
 
+instance ToSchema WalletAddress where
+  declareNamedSchema = annotate fromArbitraryJSON
+
 instance ToSchema AccountUpdate where
   declareNamedSchema = annotate fromArbitraryJSON
 
@@ -310,6 +320,9 @@ instance ToSchema Wallet where
   declareNamedSchema = annotate fromArbitraryJSON
 
 instance ToSchema NewWallet where
+  declareNamedSchema = annotate fromArbitraryJSON
+
+instance ToSchema NewAddress where
   declareNamedSchema = annotate fromArbitraryJSON
 
 instance ToSchema WalletUpdate where
--- a/wallet-new/src/Cardano/Wallet/API/V1/Types.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Types.hs
@@ -31,6 +31,9 @@ module Cardano.Wallet.API.V1.Types (
   -- * Accounts
   , Account (..)
   , AccountId
+  -- * Addresses
+  , WalletAddress (..)
+  , NewAddress (..)
   -- * Payments
   , Payment (..)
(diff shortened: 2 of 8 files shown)


wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -79,8 +79,16 @@ newDescr :: Typeable a => proxy a -> T.Text
 newDescr (p :: proxy a) =
     "A type represending an request for creating a(n) " <> renderType p <> "."
 
+-- | Derives all fields.
+allFieldsFromJSON :: forall a proxy. (Arbitrary a, ToJSON a) => proxy a -> Set T.Text
+allFieldsFromJSON _ =
```
adinapoli-iohk: How come did you end up adding this new function? What was the problem in writing:

``` haskell
instance ToDocs AddressInfo where
  readOnlyFields   = readOnlyFieldsFromJSON
 descriptionFor _ = "An Address."
```

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -336,6 +339,35 @@ deriveJSON Serokell.defaultOptions ''AccountUpdate
 instance Arbitrary AccountUpdate where
   arbitrary = AccountUpdate . fromString <$> pure "myAccount"
 
+-- | Summary about single address.
+data AddressInfo = AddressInfo
+  { addrId       :: !Core.Address
+  , addrBalance  :: !Core.Coin
+  , addrIsUsed   :: !Bool
+  , addrIsChange :: !Bool
+  } deriving (Show, Generic)
```
adinapoli-iohk: Naming is hard, and I'm fairly happy for now with the notion of an `AddressInfo`, but just bikeshedding on how to call this type: `WalletAddress` maybe? It might convey a little bit better the fact this is conceptually the wallet representation of an `Address` not a `Core.Address` per se.

I would also dropped the `Is` from `IsUsed`. Sending as JSON something like `{"used": true}` is perfectly self-explanatory. I would also rename `addrIsChange` to `addrIsChangeAddress`, is a bit more self-explanatory. In a nutshell, I might dare to suggest to go with:

```haskell
data WalletAddress = WalletAddress
  { addrId       :: !Core.Address
  , addrBalance  :: !Core.Coin
  , addrUsed   :: !Bool
  , addrIsChangeAddress :: !Bool
 } deriving (Show, Generic)
```

This would generate something like:

```
{ "id": "AABBCC"
, "balance": 1000
, "used": true
, "isChangeAddress": false
}
```

What do you think?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
Martoon-00: I couldn't reasonably define `Update AddressInfo`, because it's fully read-only.
So couldn't use `readOnlyFieldsFromJSON`

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
Martoon-00: Yeah, this way it looks much bette, thanks for the note!

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
Martoon-00: @adinapoli-iohk I can declare `type Update AddressInfo = ()`, it fits to current usages of `Update` type family in neat way.
But I don't know which other usages for `Update` do you have in mind, so can't say that such declaration would be correct in all usage cases. 
What do you think?

adinapoli-iohk: Hey @Martoon-00 -- that's a valid suggestion. The `Update` (and `New`) type families were created for just 2 reasons:

- As an indirection to allow us to create better, more descriptive APIs in servant so that, for example, you can express at the type level you are creating a new `Address` by typing `New Address` rather than the `NewAddress` ad-hoc type. This also allows us to change the underlying type name without breaking the API;

- In the case of `Update` is also used to computed the `diff` of the two JSON object (the input, LHS one and the RHS one) to show in Swagger which fields are readonly (Swagger will add a nice `readOnly: true` description in the specs. What you propose sound sensible, I cannot think on top of my head to any problems it might create.

Your patch has something which concerns me more, though, hehe. I tried to inspect the generated swagger specs and for some reason the `spendingPassword` is marked as required whereas in our types is marked as a `Maybe` (quite rightly):

screen shot 2017-12-01 at 13 05 12

Do you know why is that? Am I staring at the wrong API?

Martoon-00: @adinapoli-iohk Note that the same applies to spec for `NewWallet`, while this PR didn't touch it.
Perhaps this behaviour derives from `sketchSchema`:
https://github.com/GetShopTV/swagger2/pull/48/files#diff-aa239390df4e45623cfb02ef393bdab4R309
AFAIS it just takes all fields it sees and marks them as required

adinapoli-iohk: > Note that the same thing you mention applies to spec for NewWallet, while this PR didn't touch it.

Yea, you're right!

> AFAIS it just takes all fields it sees in JSON and marks them as required

Meh, that's a bit lame, but I guess this is what it is 😁

Martoon-00:
Closed in favour of #2063

PR #2009, opened 2017-12-01
[DO NOT MERGE] Cardano sl 1.0 gruping policy

akegalj:
Issue here: https://iohk.myjetbrains.com/youtrack/issue/CSL-1996

tldr; The branch is actually:
 * v1.0.2
 * cherry-pick (https://github.com/input-output-hk/cardano-sl/pull/1826 )
 * remove `Buildable (SafeLog InputSelectionPolicy)` instance as `SafeLog` has been introduced in v1.0.3 (which still isn't released)
 * change default grouping policy to `OptimizeForSize`

### DIFF
--- a/node/src/Pos/Client/Txp/Util.hs
+++ b/node/src/Pos/Client/Txp/Util.hs
@@ -234,7 +234,62 @@ makeRedemptionTx rsk txInputs = makeAbstractTx mkWit (map ((), ) txInputs)
             , twRedeemSig = redeemSign SignRedeemTx rsk sigData
             }
 
+-- | Helper for summing values of `TxOutAux`s
+sumTxOutCoins :: NonEmpty TxOutAux -> Integer
+sumTxOutCoins = sumCoins . map (txOutValue . toaOut)
+
+integerToFee :: MonadError TxError m => Integer -> m TxFee
+integerToFee =
+    either (throwError . invalidFee) (pure . TxFee) . integerToCoin
+  where
+    invalidFee reason = GeneralTxError ("Invalid fee: " <> reason)
+
+fixedToFee :: (MonadError TxError m, HasResolution a) => Fixed a -> m TxFee
+fixedToFee = integerToFee . ceiling
+
 type FlatUtxo = [(TxIn, TxOutAux)]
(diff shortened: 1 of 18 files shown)


node/src/Pos/Client/Txp/Util.hs
```diff
@@ -141,6 +141,20 @@ isCheckedTxError = \case
...
+    | OptimizeForSize      -- ^ No grouping
+    deriving (Show, Eq, Generic)
+
+instance Buildable InputSelectionPolicy where
+    build = \case
+        OptimizeForSecurity -> "securely"
+        OptimizeForSize -> "simple"
+
+instance Default InputSelectionPolicy where
+    def = OptimizeForSize
```
adinapoli-iohk: Yes, that's the crucial bit, we are effectively defaulting to `OptimiseForSize`, so that passing `null` to the payment endpoint will ensure we will call the `plainInputPicker`. Well done, @akegalj .

adinapoli-iohk:
@akegalj Oh dear, more conflicts 😢  Mind tackling those?

akegalj:
@adinapoli-iohk no, we won't merge this. We opened this just so it is easier to review. This is a separate branch that will be used to create installers so that @fersel can test these installers and hand over to a single user which has this problem (form CSL-1996)

Is my understanding right?

adinapoli-iohk:
> Is my understanding right?

You are dead right, this is what happens when tying & listening at the same time :)

akegalj:
@fersel have we test these installers?

gromakovsky:
@akegalj AFAIU, this is something old and we don't need it anymore. Am I right? Please check and close this PR if it's no longer needed.

akegalj:
@gromakovsky I believe this can be closed. Let's wait for @fersel to give a green light.

The goal of this PR is to fix wallet of few Daedalus users who are trading with a lot of ADA. I am not sure has this ever been done. 

I see our release still contains problematic selection policy (if my intuition is right?). So:
 * we did deliver fixed installer to users who had the problem
 * we didn't and we won't make special installers for the user

@fersel is the later true?

gromakovsky:
@fersel confirmed this PR can be closed: https://serokell.slack.com/archives/G2VLVLE0K/p1516959430000095

PR #1986, opened 2017-11-28
[CSL-1957] Merge feature branch `alfredo/wallet-new` into `master`

adinapoli-iohk:
@gromakovsky Apologies for another humongous PR, I think I have learnt the lesson that squashed merges and feature branches do not go along well. I have tried to cherry pick the commits which was produced after https://github.com/input-output-hk/cardano-sl/pull/1918, but I've got into painful conflict-resolution due to the refactoring which happened in core lately, so the path of least resistance seems to simply re-do the merging, this time the classic way, so that next merges are much, much smaller. Apologies 😞 

EDIT: ah, looks like that luckly despite all the commits are still shown (due to the wrong merging stategy) at least the diff looks all-right, "just" 1646 SLOC 😉 

It's still backward-compatible code, so reviewing it can be skipped or done very lightly.

PR #2010, opened 2017-12-01
[DO NOT MREGE][CSL-1952] Rename 'wallet-new-server' into 'cardano-node'.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-1952

As discussed with @adinapoli-iohk - we need to eventually completely replace the calls, renaming the old to simply `cardano-node-old` and the new as `cardano-node` would do the trick and I think we should go for it, but then we *shouldn’t* merge the PR until this one is merged - https://github.com/input-output-hk/cardano-sl/pull/1986 
This is because we want **master** to incorporate the current `wallet-new` whilst it's still not invoked. Then once we port completely the API we merge this PR and on the next merge of `alfredo/wallet-new` into `master` we should be in the position of start using `wallet-new` as if it was “the normal wallet”.

### DIFF
--- a/scripts/launch/demo-with-new-wallet-api.sh
+++ b/scripts/launch/demo-with-new-wallet-api.sh
@@ -2,4 +2,4 @@
 
 base=$(dirname "$0")
 
-WALLET_EXE_NAME='cardano-node-new' WALLET_TEST=1 "$base"/demo.sh $@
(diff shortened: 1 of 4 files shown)


ksaric:
**No merging until #1986 is merged.**

ksaric:
Rebased to branch `feature/new-wallet` - https://github.com/input-output-hk/cardano-sl/pull/2021

PR #2011, opened 2017-12-01
[CSL-1889] Implement `DELETE /api/v1/wallets/{walletId}` with V0 logic

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-1889

### DIFF
--- a/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
@@ -9,9 +9,7 @@ import qualified Pos.Wallet.Web.Methods.Restore as V0
 import           Cardano.Wallet.API.V1.Migration
 import           Cardano.Wallet.API.V1.Types as V1
 import qualified Cardano.Wallet.API.V1.Wallets as Wallets
-import           Pos.Update.Configuration (curSoftwareVersion)
-import           Pos.Util.CompileInfo (compileInfo, ctiGitRevision)
-import           Pos.Wallet.WalletMode (MonadBlockchainInfo, blockchainSlotDuration)
+import           Pos.Update.Configuration ()
 
 import           Pos.Wallet.Web.Methods.Logic (MonadWalletLogic)
 import           Servant
--- a/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
@@ -69,14 +69,15 @@ updatePassword _ _ = liftIO $ generate arbitrary
 
 -- | Deletes an exisiting wallet.
 deleteWallet
-    :: (HasConfigurations)
+    :: (HasConfigurations, MonadWalletLogic ctx m)
     => WalletId
-    -> MonadV1 NoContent
+    -> m NoContent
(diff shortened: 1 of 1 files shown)


wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
```diff
@@ -57,20 +59,27 @@ listWallets PaginationParams {..} = do
...
+    -> PasswordUpdate
+    -> MonadV1 Wallet
 updatePassword _ _ = liftIO $ generate arbitrary
 
-deleteWallet :: WalletId
-             -> MonadV1 NoContent
-deleteWallet _ = return NoContent
+-- | Deletes an exisiting wallet.
+deleteWallet
+    :: (HasConfigurations, MonadWalletLogic ctx m)
```
adinapoli-iohk: As you are calling `deleteWallet` from the main `handlers`, you don't need the extra `HasConfigurations`. Have you tried removing that?

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
```diff
@@ -57,20 +59,27 @@ listWallets PaginationParams {..} = do
...
+    => WalletId
+    -> m NoContent
+deleteWallet (WalletId walletId) =
+    V0.deleteWallet . V0.CId . V0.CHash $ walletId
```
adinapoli-iohk: Just a mental note: We are currently not doing any error handling yet, but obviously in a production-quality setting we would need to check the deletion succeeded etc. But this is for post-phase 1, no need to worry about this right now.

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
ksaric: Correct, I should have removed that once I switched from `MonadV1` because `MonadWalletLogic` contains it:
```
type MonadWalletLogic ctx m =
    ( MonadWalletLogicRead ctx m
    , MonadWalletDB ctx m
    , MonadKeys m
    )
```
```
type MonadWalletLogicRead ctx m =
    ( MonadIO m
    , MonadThrow m
    , WithLogger m
    , MonadRandom m
    , MonadSlots ctx m
    , MonadDBRead m
    , MonadBalances m
    , MonadWalletDBRead ctx m
    , MonadKeysRead m
    , MonadTxpMem WalletMempoolExt ctx m  -- TODO: remove these two once 'fixingCachedAccModifier' becomes useless
    , BlockLockMode ctx m
    , HasConfiguration
    )
```

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Wallets.hs
ksaric: Yes, it occurred to me, but I realized that writing something like:
```
case V0.deleteWallet . V0.CId . V0.CHash $ walletId of
    NoContent -> NoContent
    otherwise -> throwM $ WalletError $ "Error with wallet deletion."
```
Is the same it's now, since the type system enforces that `NoContent` is returned and in the other cases `MonadThrow` is used.

ksaric:
Rebased to `feature/new-wallet` - https://github.com/input-output-hk/cardano-sl/pull/2024

PR #2015, opened 2017-12-02
[WIP][CSL-1957][CSL-2002] Merge wallet-new into master disabling its specs on Appveyor

adinapoli-iohk:
This PR tries to sort out the mess me, and only myself are responsible for 🤢  .

Unfortunately I didn't find a better solution that squashing/rebasing only the final part (`[CSL-2002]`) and for the rest keeping the original commits. We have two choices: if we are interested in the history we should merge this normally, if we want instead to not taint the `master` history with 120 new commit we should merge this squashed **and** (this is crucial) delete `alfredo/wallet-new` for good, so that we can create a proper feature branch where we follow a much better protocol. I take full responsibility for this mess, which I'm now trying to untangle. After this I must go back and fully master `git flow` before creating more spaghetti commits 🙈 .

@gromakovsky all yours to review :)


core/cardano-sl-core.cabal
```diff
@@ -59,6 +59,10 @@ library
                        Pos.Util.Chrono
                        Pos.Util.Limits
 
+                       -- Update
+                       -- (Needed for wallet-new)
+                       Pos.Core.Update.Types
```
gromakovsky: It's re-exported from `Pos.Core.Update` module and I would prefer not to expose this module, but import `Pos.Core.Update` instead.
The reason why I find it better is that it allows us to change layout of modules in `Pos.Core.Update` without breaking the code outside `cardano-sl-core` as long as we re-export the same stuff. Now have 3 modules there, but later one of them may become too big and we may want to split it, for example. Or it can become too small and we will want to merge some modules. Or just rename for some reason.
There is a trade-off, because when you import `Pos.Core.Update` you depend on more stuff (because it contains more than just `Pos.Core.Update.Types`), that's why `cardano-sl-core` exposes not only `Pos.Core` (which re-exports virtually everything). I think current granularity (when `Pos.Core.Update` is exposed and `Pos.Core.Update.Types` is not) is a good balance between these trade-offs. `Pos.Core.Update` contains not too much stuff, but gives us some flexibility.

crypto/cardano-sl-crypto.cabal
```diff
@@ -27,6 +27,7 @@ library
                        Pos.Crypto.Signing
                        Pos.Crypto.Signing.Redeem
                        Pos.Crypto.Signing.Safe
+                       Pos.Crypto.Signing.Types.Safe
```
gromakovsky: Same thing as about exposing `Pos.Core.Update`. This one is re-exported by `Pos.Crypto.Signing.Safe`.

scripts/launch/demo.sh
```diff
@@ -78,9 +78,17 @@ fi
 # and start processing the first slot.
 if [ -z "$system_start" ]
   then
-    system_start=$((`date +%s` + 15))
+    system_start=$((`date +%s` + 20))
```
gromakovsky: Why is it necessary?

core/cardano-sl-core.cabal
adinapoli-iohk: Thank you very much for the review, @gromakovsky ! I genuinely not appreciated that module was exported via `Pos.Core.Update`, my bad! I agree with the general sentiment here, this way of exporting allows us to change things internally without breaking the external API. Leave it with me!

scripts/launch/demo.sh
adinapoli-iohk: @ksaric Reported whilst working on CSL-1951/CSL-1970 that in order for the `wallet-new` executable to work within the old launch script such bump was needed (at least on his machine). This can be plausible considering either a slower machine than average or the fact that `wallet-new` also writes on disk the updated swagger spec, and maybe that's the culprit.

However, if you think this is a roadblock in merging this PR I'm happy to revert, Mr @gromakovsky 😉

adinapoli-iohk:
Thanks again for taking the time reviewing this one @gromakovsky , I definitely owe you a beer 😉  I have addressed all the concerns you had -- let me know if you want me to decrease the timeout in `demo.sh`, otherwise I will be very keen on merging this bad boy as soon as CI passes -- cheers!

adinapoli-iohk:
@gromakovsky Sigh, Darwin builds are timing out -- is it possible to bypass them? 😞

scripts/launch/demo.sh
gromakovsky: I don't mind merging this PR with this change (because I feel your pain that you can't merge it for so long 😅), but I'd like to know what takes this time.
Previously there was 40 sec delay and it was really annoying to wait for 40 sec when you launch demo to test that things work. Then we investigated it and came to conclusion that 15 sec should be enough for everybody. Now it's going into wrong direction again and it's better not to let it be 40 sec :)

gromakovsky:
Ok, let's bypass. I know those timeouts are transient and don't indicate a real problem.

gromakovsky:
We decided to squash and merge, which means that other PRs targeting `alfredo/wallet-new` will have to be rebased onto master using `git rebase --onto master alfredo/wallet-new featureX` or by cherry-picking commits manually.

ksaric:
:tada: 
Regarding the timeout increase - https://github.com/input-output-hk/cardano-sl/pull/1997#issuecomment-348171226

adinapoli-iohk:
@ksaric To please @gromakovsky I had to revert the timeout, we will probably need to get at the bottom of that in a separate ticket (sorry!) or at last differentiate between the different wallets. Ivan was proposing to extend `demo.sh` with something like (or similar):

```
if [[ $WALLET_EXE_NAME == "wallet-new-server" ]] then system_start=$((`date +%s` + 20)) else system_start=$((`date +%s` + 15)) fi
```

PR #2018, opened 2017-12-03
Increase version of `rocksdb-haskell`

rockbmb:


stack.yaml
```diff
@@ -116,6 +116,12 @@ packages:
     commit: 6e6595f68c947107beba185392b350abd2009ba0
   extra-dep: true
 
+# Temporary, only until '1.0.1' appears in the resolver
+- location:
+    git: https://github.com/serokell/rocksdb-haskell.git
+    commit: d7e3bcc479807f976199cd52f56d233d136a7779
+  extra-dep: true
```
gromakovsky: You don't need it in LTS to be able to put this version into `extra-deps`. Just increase the number here: https://github.com/input-output-hk/cardano-sl/blob/ad55aff083039f7eff70b3b134d9ef295d30a661/stack.yaml#L144

rockbmb: Missed that part, thanks.

PR #2019, opened 2017-12-03
[CSL-1803] Increase version of 'rocksdb-haskell'

rockbmb:


ssc/cardano-sl-ssc.cabal
```diff
@@ -55,6 +55,7 @@ library
                      , mtl
                      , node-sketch
                      , parsec
+                     , rocksdb-haskell >= 1.0.1
```
gromakovsky: Why did it become necessary? You didn't change anything in `ssc/`, so it looks strange.

rockbmb: Hmm, wouldn't it cause problems to depend on a version of a package in which a problem exists, and for which a solution has been found and made available?

If we bump in one place, wouldn't it make sense to bump elsewhere as well?

gromakovsky: But the thing is: `cardano-sl-ssc` didn't depend on `rocksdb-haskell` before your changes.

rockbmb: Oh, wait - it does, but in `master`. My bad. I `cherry-picked` this command from `master` to this branch after I closed the previous PR. Let me fix this.

neongreen:
This PR was accidentally merged (and since reverted). The saga continues in https://github.com/input-output-hk/cardano-sl/pull/2047.

Issue #1993, opened 2017-11-29
Running node with wallet


VascoVisser:
I am attempting to run the daedalus wallet on my linux box.

I can build the frontend and see it attempts to connect to the backend:
```
 [01:41:57:0943] [error] CardanoClientApi::syncProgress error: {
   "stack": "Error: connect ECONNREFUSED 127.0.0.1:8090\n    at Object.exports._errnoException (util.js:1050:11)\n    at exports._exceptionWithHostPort (util.js:1073:20)\n    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1097:14)",
   "message": "connect ECONNREFUSED 127.0.0.1:8090",
   "code": "ECONNREFUSED",
   "errno": "ECONNREFUSED",
   "syscall": "connect",
   "address": "127.0.0.1",
   "port": 8090
 }
```

I have builde the backend as per instructions [1] and can launch the node, which reports the wallet part is build and enabled. 
```
$ stack exec -- cardano-node --system-start 1511917071
[Attention] We are in DEV mode
Asserts are ON
[Attention] Software is built with wallet part
System start time is 1511917071000000
Current time is 1511917078029341
Wallet is enabled!
...
```
However, it doesn't open the wallet socket. The node output shows repeatedly:
```
[node.worker.security:WARNING:ThreadId 24] [2017-11-29 01:59:13 CET] self: could not choose suitable alternative from []
[node.worker.security:ERROR:ThreadId 24] [2017-11-29 01:59:13 CET] self: enqueue instruction EnqueueAll {enqNodeType = NodeRelay, enqMaxAhead = MaxAhead 1, enqPrecedence = PHigh} failed to enqueue message MsgRequestBlockHeaders Nothing to forwarding sets [[]]
[node.worker.security:ERROR:ThreadId 24] [2017-11-29 01:59:13 CET] self: enqueue instruction EnqueueAll {enqNodeType = NodeRelay, enqMaxAhead = MaxAhead 1, enqPrecedence = PHigh} failed to enqueue message MsgRequestBlockHeaders Nothing to forwarding sets [[]]
[node.worker.subscription:ERROR:ThreadId 40] [2017-11-29 01:59:13 CET] dnsSubscriptionWorker: no relays found for index 1
[node.worker.subscription:ERROR:ThreadId 40] [2017-11-29 01:59:13 CET] dnsSubscriptionWorker: no relays found for index 1
[...]

PR #2022, opened 2017-12-04
[CSL-2009] Remove wallet purescript bridge.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2009

TBH, I removed this because I found that not being able to run stack ghci (globally) is annoying.
But it's something that needs to be done anyway.

This should make the CI faster as well.

Rebased onto `feature/new-wallet`.


adinapoli-iohk:
Resuming the conversation with @ksaric and @akegalj : In general, I'm ok with removing the bridge, because exchanges are using the RESTful API directly and Daedalus has been ported over to use the JS API directly.

Said that, we should give @darko-mijic , @DominikGuzei and Nikola visibility on this PR as it might impact them and their workflow. If they give us green lights, and other peers are well informed (aka Devops etc) I don't see great problems in removing this one. This PR also comes in the form of a nice, self-contained commit which is dead-easy to revert anyway.

akegalj:
@adinapoli-iohk @ksaric before removing the bridge we should also remove all instructions that includes the bridge from all the repos: cardano-sl, daedalus, docs, install instructions such as https://github.com/input-output-hk/cardano-sl/issues/1993 

Also we should update scripts in our repos to exclude the bridge (daedalus repo + maybe some devops scripts) as well

adinapoli-iohk:
> before removing the bridge we should also remove all instructions that includes the bridge from all the repos: cardano-sl, daedalus, docs, install instructions such as #1993

> Also we should update scripts in our repos to exclude the bridge (daedalus repo + maybe some devops scripts) as well

Good call, looks like removing the code without updating the docs or the script would cause more harm than good.

denisshevchenko:
> before removing the bridge we should also remove all instructions that includes the bridge from all the repos: cardano-sl, daedalus, docs, install instructions such as #1993

@akegalj Should I remove notes about bridge from documentation right now?

akegalj:
> Should I remove notes about bridge from documentation right now?

Yes, I believe we can

denisshevchenko:
Ok, should I remove it as well? https://cardanodocs.com/technical/wallet-frontend/#wallet-frontend-api-usage-scenario. I read:

> This is an example session that shows how the user/client can work with the Daedalus-bridge library.

So, it must be removed, right?

akegalj:
@denisshevchenko yes that whole section can be removed

denisshevchenko:
@akegalj Please check it out: https://github.com/input-output-hk/cardanodocs.com/pull/65

ksaric:
@deepfire Fixed the AppVeyor failures and I think we are good from DevOps side?
@nikolaglumac, @darko-mijic , @DominikGuzei ? Is this good? Can we remove this? I know you already said you don't use the bridge, but a confirmation would be good so I don't mess something up when merging.

nikolaglumac:
@ksaric from the standpoint of Daedalus `master` branch the bridge is no longer used. I am not sure about the older `release/0.8.x` branches nor am I sure when you plan to release this updated Backend. If this is to be released along with latest Daedalus master then I think it is fine to remove the Bridge.

cc @darko-mijic @DominikGuzei

adinapoli-iohk:
@ksaric I would wait for the last word from @darko-mijic on this, but I'm optimistic that we never make releases from `master` and that as it stand newer versions of Daedalus do not depend on it.

What's really critical though is that we merge this as a *single, squashed commit*, so that it's easily revert-able if s**t hits the fan. 😉

ksaric:
> I can see there are commits here which does not belong to this removal process, so we should either rebase or come up with a clean git history, as this PR effectively does not just remove stuff, but also adds code we might not want to.

@adinapoli-iohk It's all good. The small change that's added is the change for _explorer_ so we can run `stack ghci` globally.

darko-mijic:
After the release we are doing right now PureScript bridge will no longer be needed. So it can be removed.

adinapoli-iohk:
@ksaric Are we really sure? I can see this is also incorporating the Sev1 fixes for Binance (the endpoint to reset the pending Tx state). Not sure is what we want to incorporate as part of this PR? 😉

asbachb:
So latest stable release is not buildable anymore? https://github.com/input-output-hk/daedalus/releases/tag/0.8.0

PR #2025, opened 2017-12-04
[CSL-1927] Update V1 new revision with George's feedback

adinapoli-iohk:
@georgeee Third time lucky?

(Follows old summary)

Here is the proposed new revision for the update API for V1, which includes the new endpoints to propose & vote for an update. You might notice the API is quite different from the one @domenkozar  originally proposed and that I proposed as part of [this PR](https://github.com/input-output-hk/cardano-sl/pull/1884). This is because I have had time to think about @georgeee 's review and I am now much more happy of how the API looks & feels as it's way more RESTful in my opinion, and all the custom ad-hoc actions (except for one, hehe) are gone.

[Play with the rendered API here!](https://editor.swagger.io/?url=https://gist.githubusercontent.com/adinapoli-iohk/b8d54d04101e60e83166cb32dfa9d8ff/raw/2e8193777aea781d01c0aa5ac96f50d9ef79d640/updates_v1_rev1.yaml#)

Here is a glimpse of the endpoints:

screen shot 2017-11-30 at 17 06 15

You will see how I have conceptually separated the concept of a `WalletSoftwareUpdate` (which technically represents a confirmed `UpdateProposal`) from the concept of the `UpdateProposal` itself which is something that everybody (but for now just devops) can propose.

I have a question for you: am I right that the schema Domen is proposing (with the addition of those `installers` etc) would need to be reflected in the code by modifying the current [UpdateProposal](https://github.com/input-output-hk/cardano-sl/blob/8507d03ba928e07daea57f9a52f6dc03a9d65779/core/Pos/Core/Update.hs#L176) datatype?

Let me know!

### DIFF
--- a/wallet-new/docs/updates.md
+++ b/wallet-new/docs/updates.md
@@ -236,12 +236,14 @@ data WalletSoftwareUpdate = WalletSoftwareUpdate
     , wsuNegativeStake   :: !CCoin
     , wsuSeverity        :: !UpdateSeverity
     -- ^ The severity of the update (i.e. optional or mandatory).
-    , wsuUpdateState     :: !UpdateState
+    , wsuState     :: !UpdateState
     -- ^ The state of this update from users point of view.
-    , wsuSha256Checksum  :: !Sha256Checksum
-    -- ^ The checksum expected for this installer. It will probably
-    -- need to be passed as part of the `UpdateProposal` (maybe in an attribute?),
-    -- or we could reuse the hash which is part of the name, if possible.
+    , wsuDownloadProgress  :: !Word8
+    -- ^ A progress in percentage (from 0 to 100).
+    -- N.B. In the real implementation this can be refined to a better type.
+    , wsuId  :: !Hash
+    -- ^ The update id, which is also the checksum expected for this installer,
+    -- and correspond to the @Blake2b256@ hash of the installer.
     } deriving (Eq, Show, Generic, Typeable)
 ```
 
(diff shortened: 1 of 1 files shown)


georgeee:
```
{
        "platform": "linux",
        "url": "https://s3.iohk.io/installers/daedalus.zip",
        "checksum": "A8ADD4BDDDFD93E4877D2746E62817B116364A1FA7BC148D95090BC7333B3673"
      }
```

Instead of `url` now there should be local file.
It may change after https://iohk.myjetbrains.com/youtrack/issue/CSL-1934 implemented, but  now supplying full url in proposal makes no sense

adinapoli-iohk:
> Instead of url now there should be local file.

I hear you @georgeee , and we discussed this via direct message, but let me summarise here what we said. Essentially, there is a bit of tension in the design space: I would like to go with an API which design is "definitive" in the sense I don't want to roll out an API to public which is _versioned_ (as in `/api/v1/..`) but that can change between releases. By definition once you release a version of an API, you don't break it.

At the same time, @domenkozar wanted this API to see the light very soon so we can start using it internally to propose updates. The only "solution" I can see here is an "hack": we leave this field labeled `url` as in the "future" API, but for now we use it to store local filepath to installers. As long as the semantic of this field is clear, it shouldn't create problems. We can still model this field internally in different ways, from a simple `Text` to something more fancy like:

```haskell
data InstallerPath =
    URL Http.URL
  | Path FilePath
```

I personally think something like a plain `Text` that we parse to understand if this is an HTTP url or a path to the filesystem is enough, imho. What do you think?

adinapoli-iohk:
@georgeee Merging this as-it-is, we can discuss the `url/path` thing once @domenkozar is back.

georgeee:
I would rather go with way of supplying separately a hash of installer (as a handle) and set of mirrors separately in update proposal (also additional mirrors to be default/customized in Daedalus).

Regarding API for proposal, that should be one URL whereas in proposal we will have several URL for different mirrors (in either way). Having option to select file or url may be confusing after all. So we must highlight it in docs that the url is not anyhow reflected in proposal and serves only for executing API.

After all, feature seems useful. E.g. this will allow executing update proposal command on remote host.

PR #2005, opened 2017-11-30
[CSL-1947] API v1 endpoint `getAccount`

sectore:
_Task_ https://iohk.myjetbrains.com/youtrack/issue/CSL-1949

_Note:_ Field `accWalletId :: WalletId` of `V1.Account` has a been deactivated temporally in this PR. As mentioned in https://input-output-rnd.slack.com/archives/C819S481Y/p1511970392000449 we have to discuss how to get a `WalletId` while migrating `V0.CAccount` to  `V1.Account`.

PR #2030, opened 2017-12-04
[CSL-1947] API v1 endpoint `getAccount`

sectore:
Based on PR #2005 , but now w/ fixes and ready to go into `feature/new-wallet`.


wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
```diff
@@ -108,3 +111,60 @@ instance Migrate V0.SyncProgress V1.SyncProgress where
...
+
+instance Migrate V1.WalletId (V0.CId V0.Wal) where
+    eitherMigrate (V1.WalletId h) = pure $ V0.CId (V0.CHash h)
+
+--
+-- Following instances are friendly borrowed from @martoon's PR https://github.com/input-output-hk/cardano-sl/pull/2008
+-- TODO (jk): Use instances of his PR when it has been merged
+
+-- in old API 'V0.AccountId' supposed to carry both wallet id and derivation index
+instance Migrate (V1.WalletId, V1.AccountId) V0.AccountId where
```
adinapoli-iohk: Interesting approach, probably we can't do better than that without complicating the types further.

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
```diff
@@ -108,3 +111,60 @@ instance Migrate V0.SyncProgress V1.SyncProgress where
...
+    eitherMigrate (walId, accId) =
+        V0.AccountId <$> eitherMigrate walId <*> pure accId
+
+instance Migrate V0.AccountId (V1.WalletId, V1.AccountId) where
+    eitherMigrate accId =
+        (,) <$> eitherMigrate (V0.aiWId accId) <*> pure (V0.aiIndex accId)
```
adinapoli-iohk: If you want, this can be written using `TupleSections` also as:

```haskell
(,V0.aiIndex accId) <$> eitherMigrate (V0.aiWid accId)
```

But I think the "normal" applicative style is more elegant.

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
```diff
@@ -108,3 +111,60 @@ instance Migrate V0.SyncProgress V1.SyncProgress where
...
+--
+-- #end of TODO (jk) ^
+--
+
+instance Migrate V0.CAccountId V1.AccountId where
+    eitherMigrate cAccId = do
+        -- V0.CAccountId -> V0.AccountId -> V1.AccountId
+        a :: V0.AccountId <- eitherMigrate cAccId
+        b :: (V1.WalletId, V1.AccountId) <- eitherMigrate a
+        pure $ snd b
```
adinapoli-iohk: I find this function a bit dense. If we need to add type signatures under the form of comments we can probably do better. What about something along the lines of:

``` haskell
eitherMigrate clientAccountId = do
  oldAccountId <- eitherMigrate clientAccountId
  (walletId, newAccountId) <- eitherMigrate oldAccountId
  pure newAccountId
```

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
```diff
@@ -108,3 +111,60 @@ instance Migrate V0.SyncProgress V1.SyncProgress where
...
+
+instance Migrate V0.CAccountId V1.WalletId where
+    eitherMigrate cAccId = do
+        -- V0.CAccountId -> V0.AccountId -> V1.WalletId
+        a :: V0.AccountId <- eitherMigrate cAccId
+        b :: (V1.WalletId, V1.AccountId) <- eitherMigrate a
```
adinapoli-iohk: Ditto as above.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -321,11 +321,11 @@ data Account = Account
 deriveJSON Serokell.defaultOptions ''Account
 
 instance Arbitrary Account where
-  arbitrary = Account . fromString <$> elements ["DEADBeef", "123456"]
-                                   <*> listOf1 arbitrary
-                                   <*> arbitrary
-                                   <*> pure "My account"
-                                   <*> arbitrary
+  arbitrary = Account <$> arbitrary
+                      <*> listOf1 arbitrary
```
adinapoli-iohk: Probably it would be better to relax that `listOf1` from the `Arbitrary` instance. I know I was the guy which wrote this code, and that I probably did it in order for our Swagger spec to give back something nice rather than an empty list, but when we start getting serious about testing we should ask yourselves if your QC `Arbitrary` instances are modelling the domain correctly, and definitely an `Account` can have 0 or more addresses associated to it 😉

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
sectore: As mentioned in comment above those instances are `friendly borrowed` from @Martoon-00 's PR #2008 So kudos to Kostja :)

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
sectore: Based on same PR #2008 So I think it's ok to leave it as it is for now. If needed we could improve/change it with next review of Kostja's PR.

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
sectore: Oh yessss, naming variables like `a` or `b` make no sense here. I will change it. What we still need here are type annotations to guide `ghc` and to not get errors such as `Ambiguous type variables ...`. As always I'm open for better solutions, of course :)

adinapoli-iohk: > As always I'm open for better solutions, of course :)

@sectore Do you happen to like `TypeApplications`? Usually you need just one in the right place to allow GHC to infer the whole thing -- did you give it a whirl? 😉

adinapoli-iohk:
Thanks @sectore !

PR #2038, opened 2017-12-05
[CSL-1926] First class pagination for V1

adinapoli-iohk:
This PR ties the knot and provides a minimal but fully functional pagination support for V1 resources (**Bogus test data used below for demonstration purposes**):

pagination

It might seem too simple, but I as you can see we are navigating through 100.000 records fairly easy, so I don't the pagination overhead won't be significant: what will be costly, in practice, is how we slice & dice the data.

Feel free to propose amendment & changes to the current API.


ksaric:
@adinapoli-iohk Maybe not related to this PR, but wasn't there a plan to include `status` like you posted - https://labs.omniti.com/labs/jsend ?

wallet-new/src/Cardano/Wallet/API/Response.hs
```diff
@@ -0,0 +1,106 @@
...
+  , extMeta :: Metadata -- ^ Extra metadata to be returned.
+  } deriving (Show, Eq, Generic)
+
+deriveJSON Serokell.defaultOptions ''ExtendedResponse
+
+instance Arbitrary a => Arbitrary (ExtendedResponse a) where
+  arbitrary = ExtendedResponse <$> arbitrary <*> arbitrary
+
+-- | `PaginationParams` is datatype which combines request params related
+-- to pagination together
```
Martoon-00: Lost baby?

ksaric:
@adinapoli-iohk Also, I see that it's not formatted according to the style guide, not going to point out every single part.

adinapoli-iohk:
> Maybe not related to this PR, but wasn't there a plan to include status like you 

@ksaric That's a good idea -- I don't see any drawbacks in adding it, so we might just as well going for it 😉

wallet-new/src/Cardano/Wallet/API/Response.hs
adinapoli-iohk: Ops, possibly 😬

adinapoli-iohk:
> Also, I see that it's not formatted according to the style guide, not going to point out every single part.

@ksaric Right, for some reason I never got my editor to intend automatically 4 spaces without me hitting explicitly <tab>, despite amending my `.dir-locals.el`. Leave it with me.

wallet-new/src/Cardano/Wallet/API/Request/Pagination.hs
```diff
@@ -0,0 +1,102 @@
...
+
+import           Data.Aeson.TH
+import           Data.Default
+import qualified Serokell.Aeson.Options as Serokell
+import           Test.QuickCheck (Arbitrary (..), choose, getPositive)
+import           Web.HttpApiData
+
+-- | A `Page` is used in paginated endpoints to request access to a particular
+-- subset of a collection.
+newtype Page = Page Int
```
akegalj: Maybe we could consider `Word64` as it is unsigned (and we will have twice as much pages).

If we will need more pages then `Integer` seems the way to go

wallet-new/src/Cardano/Wallet/API/Request/Pagination.hs
```diff
@@ -0,0 +1,102 @@
...
+instance ToHttpApiData Page where
+    toQueryParam (Page p) = fromString (show p)
+
+-- | If not specified otherwise, return first page.
+instance Default Page where
+    def = Page 1
+
+-- | A `PerPage` is used to specify the number of entries which should be returned
+-- as part of a paginated response.
+newtype PerPage = PerPage Int
```
akegalj: Also maybe `Word64` ?

wallet-new/src/Cardano/Wallet/API/Request/Pagination.hs
```diff
@@ -0,0 +1,102 @@
...
+
+instance ToHttpApiData PerPage where
+    toQueryParam (PerPage p) = fromString (show p)
+
+instance Default PerPage where
+    def = PerPage defaultPerPageEntries
+
+-- | Extra information associated with pagination
+data PaginationMetadata = PaginationMetadata
+  { metaTotalPages   :: Int     -- ^ The total pages returned by this query.
```
akegalj: I would use `Word` everywhere where it is expected to have unsigned integral as it caries more info. Its a minor though

wallet-new/src/Cardano/Wallet/API/Response.hs
```diff
@@ -0,0 +1,119 @@
...
+        perPage@(PerPage pp)   = ppPerPage
+        currentPage@(Page cp)  = ppPage
+        totalPages             = ceiling (fromIntegral totalEntries / (fromIntegral pp :: Double))
+        metadata               = PaginationMetadata {
+                                 metaTotalPages = totalPages
+                               , metaPage = currentPage
+                               , metaPerPage = perPage
+                               , metaTotalEntries = totalEntries
+                               }
+        slice                  = take pp . drop ((cp - 1) * pp) . toList
```
akegalj: super minor, but you might save some parentheses 
```
slice                  = take pp . drop (pred cp * pp) . toList
```

some people don't like it though as they find `(x - 1)` more clear then `pred x`.

akegalj:
ps, how did you record gif from PR description ?

wallet-new/src/Cardano/Wallet/API/Request/Pagination.hs
adinapoli-iohk: @akegalj If we consider a very generous `10` entries per pages, we can represent up to `214748364` pages with an `Int`. Seems quite a lot for the foreseeable future. What do you think? 😉

wallet-new/src/Cardano/Wallet/API/Request/Pagination.hs
adinapoli-iohk: @akegalj Do you really think we might want to send more than `2147483648` entries in a _single_ page? 😉

wallet-new/src/Cardano/Wallet/API/Request/Pagination.hs
adinapoli-iohk: @akegalj I do agree that this way we convey the info this stuff cannot be negative, but as you said it's a minor amendment we can deal with later. I personally find though that `Int` is a bit easier to parse for the "average joe" programmer 😬

wallet-new/src/Cardano/Wallet/API/Response.hs
adinapoli-iohk: @akegalj I am an old school guy and I like my plain arithmetic 😁  Again, much easier for the average programmer to understand, and key to lower the contribution barrier 😉

adinapoli-iohk:
> ps, how did you record gif from PR description ?

@akegalj With a program called [liceCap](https://github.com/justinfrankel/licecap) 😛

PR #2039, opened 2017-12-05
[CSL-2027] Use servant-generic

neongreen:


neongreen:
Don't merge it yet, I want to convert the wallet as well.

stack.yaml
```diff
@@ -111,6 +111,11 @@ packages:
     commit: 2d261bb971bada1893753b503452d9e6e217bc4a
   extra-dep: true
 
+- location:
+    git: https://github.com/neongreen/servant-generic.git
```
gromakovsky: Maybe move it to `serokell/` to make our lives potentially easier if you are hit by a bus?

adinapoli-iohk:
> Don't merge it yet, I want to convert the wallet as well.

@neongreen Can I kindly ask you to do that as a separate PR? 😉  That might make a bit easier for us poor devs working on the moving target `feature/new-wallet` is handle the breaking changes, if any :)

neongreen:
> Maybe move it to serokell/ to make our lives potentially easier if you are hit by a bus?

fine, since two people asked for it I'll do it

neongreen:
> Can I kindly ask you to do that as a separate PR?

okay, then in an hour I'll fix default.nix and remove the WIP flag from this PR

stack.yaml
neongreen: done

ch1bo:
Hey,

sorry to chime in on your "internal" development here. But I was wondering what the benefit of record syntax data type APIs are? Also, how is route precedence treated?

Again, I'm just curious what your rationale is and maybe I should use servant-generic for my projects

neongreen:
@ch1bo the rationale is:

* You can use named fields for handlers & handler types and not be afraid of messing up the order
* When the handler doesn't match the needed type, you get a much smaller error message (`field blah ... doesn't match blah`)

We don't have overlapping routes AFAIK, but even if we did, the translation from records is very straightforward and `servant-generic` will handle them in the same way as `servant` does.

```haskell
data Foo route = Foo                type Foo 
  { a :: route :- A                     = A    
  , b :: route :- B                  :<|> B
  , c :: route :- C                  :<|> C
  }
```

PR #2045, opened 2017-12-05
[CSL-1962] Add address verification route

et4te:
See previously closed PR #2002 for previous review.


wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
```diff
@@ -33,3 +35,18 @@ listAddresses PaginationParams {..} = do
 
 newAddress :: Address -> Handler Address
 newAddress a = return a
+
+-- | Verifies that an address is < 78 and base58 decodable.
```
adinapoli-iohk: I'm wondering if this is really true -- in Cardano addresses are basically or var-len due to attributes, so are we sure this won't backfire in unexpected way?

@gromakovsky What are the ripple effects of embedding static size checking into this endpoint?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -229,6 +229,9 @@ instance ToDocs Account where
 instance ToDocs AccountUpdate where
   descriptionFor _ = updateDescr (Proxy @Account)
 
+instance ToDocs AddressValidity where
+  descriptionFor _ = "Whether an address is valid or not."
```
adinapoli-iohk: What about documenting the criteria needed for an address to be considered valid? The next exchange integrator will thank you 😀

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -305,9 +307,26 @@ instance Arbitrary Wallet where
...
 
+--------------------------------------------------------------------------------
+-- Addresses
+--------------------------------------------------------------------------------
+
+-- | Whether an address is valid or not.
+newtype AddressValidity = AddressValidity { getValid :: Bool }
+  deriving (Eq, Show, Generic)
+
+deriveJSON Serokell.defaultOptions ''AddressValidity
```
adinapoli-iohk: I know we discussed this previously, but I still think renaming this `isValid` makes much more sense than adhering blindly to code standard. I will take the bullet for the rest of the team 😅

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
et4te: @adinapoli-iohk There is a constant in `Pos.Core` called `maxPubKeyAddressSizeSingleKey` which seems to be used in the tests to check the address isn't larger than this size. Currently it is 78 bytes but basically the idea was to check against this constant so that e.g if 3-leveled derivation and the size is increased the code will still pass. The comment should be Verifies that an address is less than the max address size and is base58 decodeable.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
et4te: Haha, okay I will change it in this case.

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Addresses.hs
adinapoli-iohk: Ah, awesome. Saw the constant, but didn't appreciate it was used in the `core` tests as well. Great, looks exactly what we should be using here 😉

et4te: I fixed the comment to better reflect that the size isn't constant.

gromakovsky: > I'm wondering if this is really true -- in Cardano addresses are basically or var-len due to attributes

Yes, they are. Address can be arbitrarily large.

> There is a constant in `Pos.Core` called `maxPubKeyAddressSizeSingleKey`

Yes, but I am afraid it's not the constant you need. See its documentation:
```
-- | Maximal size of PubKey address with SingleKey
-- distribution (78).
maxPubKeyAddressSizeSingleKey :: (Bi Address, Bi Address') => Byte
maxPubKeyAddressSizeSingleKey = biSize largestPubKeyAddressSingleKey
```

Currently middleware creates HD addresses with bootstrap era distribution. There is the difference constant for it:

```
-- | Maximal size of HD address with BootstrapEra
-- distribution (76).
maxHDAddressSizeBoot :: (Bi Address, Bi Address') => Byte
maxHDAddressSizeBoot = biSize largestHDAddressBoot
```

Also now I see that the documentation is not ideal, because technically HD address in also a PubKey address. So to clarify: `maxPubKeyAddressSizeSingleKey` is maximal size of non-HD address with `SingleKey` distribution.
Secondly, these constants assume absence of unknown attributes. These constants are used only for transaction creation to properly estimate size of change address (which is important for fee calculation), so this assumption is fine. However, if you want to check validity of an arbitrary address, you probably should avoid this assumption.

To summarize:
1. If you want to check whether an address is valid assuming it was created by wallet middleware, it should be fine to use `maxHDAddressSizeBoot` (but not `maxPubKeyAddressSizeSingleKey`).
2. If you want to check an arbitrary address, you probably should avoid length check at all, because it can have arbitrary length. For instance, its `AddrStakeDistribution` can have `MultiKeyDistr` type which contains `(Map StakeholderId CoinPortion)` and can be as large as one wants. Also it can have arbitrary attributes.
3. It's not clear how to define _validity_ of an address. There are several tricky points. For instance, in bootstrap era one can send transactions only to addresses with `BootstrapEraDistr`. While we are in bootstrap era, should we consider an address with `SingleKeyDistr` valid? On the one hand, it's well-formed and can be parsed into `Address` data type. On the other hand, using such address is impossible. Another point is about unknown attributes. If our software is up-to-date, we will reject transactions to addresses with unknown attributes (because up-to-date software knows how to parse all attributes). But if our software becomes obsolete (i. e. adopted block version from blockchain becomes greater than ours), then addresses with unknown attributes become valid from our point of view. So it's not obvious whether addresses with unknown attributes should be considered valid.

et4te: I have removed the size check considering @gromakovsky comments. I agree that assessing an address based on its max size is not an ideal way of gauging whether it is valid and I think really your comment raises the question if it makes sense at all to check the validity of addresses in the wallet server. If this is a necessary feature then in my opinion we should check for validity against the presently allowed address format only (e.g BootstrapEraDistr) and this should ideally be specified in Pos.Core or wherever the addresses are specified.

adinapoli-iohk: Sorry for dragging you into a little rat nest with this PR @et4te, but I think that's all useful to intimately understand an important part of the system 😉  

>  I think really your comment raises the question if it makes sense at all to check the validity of addresses in the wallet server.

I hear you, and potentially this is something we might want to discuss with @DominikGuzei and the other frontend guys. The way I have understood it is that Daedalus uses this endpoint to validate an address the user input in the textfield when is about to send ADA to somebody, so the idea is to prevent incorrect addresses to be sent. Said that, I think it makes little sense to check the validity based on the _max_ length, and with _short addresses_ in the pipeline I'm wondering if it makes sense to check even the _min_ length.

Which brings us back to the question(s):

1. Should we be doing the validation server side, at all?
2. Which type of validation Daedalus would like?

If @DominikGuzei / @darko-mijic could reply those two questions, I guess we will be able to take an informative decision together. Thank you guys!

DominikGuzei: If there is no solid way to check the validity of an address based on the length then at least it should just do a very basic validation like does this string contain spaces or invalid chars (btw. what are the valid chars for an address?).

From the UX point of view this would be the dream (and goal): if i (accidentally) paste a BTC address into an ADA send form, the UI should alert and stop me! But im not sure if that's possible?

adinapoli-iohk: > should just do a very basic validation like does this string contain spaces or invalid chars (btw. what are the valid chars for an address?).

Sounds very sensible 😉 

> From the UX point of view this would be the dream (and goal): if i (accidentally) paste a BTC address into an ADA send form, the UI should alert and stop me! But im not sure if that's possible?

I guess that would be tricky, but if you squint hard is what this PR should provide: we call `decodeTextAddress` on the input data, so if we fail to recognise this as a valid Cardano address, we will return an error, although without telling you the exact reason why it failed. Would that be good enough for now @DominikGuzei ? 😀

DominikGuzei: Yeah, as long as we get back an error response for invalid addresses it should be fine (like it is now). If the "i pasted an address from wrong currency" case is too tricky then we have to live without it 😉

adinapoli-iohk:
@et4te Any chance you can address the conflicts? 😀

et4te:
I just used the github web conflict resolver, I think conflicts are resolved. @adinapoli-iohk

adinapoli-iohk:
Thank you @et4te , we finally made it! 😀

PR #2047, opened 2017-12-05
[CSL-1803] Increase version of 'rocksdb-haskell'

neongreen:


gromakovsky:
It doesn't work and most likely we'll use `rocksdb-haskell-ng`, so I am closing it.

PR #2049, opened 2017-12-06
[CSL-2018] Add migration roundtrip specs

akegalj:
Issue: https://iohk.myjetbrains.com/youtrack/issue/CSL-2018

This issue contains only roundtrip tests for available migration instances. If there is no A -> B, and B -> A migration written this issue doesn't add those.

For example `instance Migrate V1.Wallet V0.CWallet` is not possible to write as `V1.Wallet` doesn't contain all the information from `V0.CWallet`.


wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
```diff
@@ -1,14 +1,10 @@
 {- | This is a temporary module to help migration @V0@ datatypes into @V1@ datatypes.
 -}
-{-# LANGUAGE RankNTypes           #-}
-{-# LANGUAGE TypeSynonymInstances #-}
+{-# OPTIONS_GHC -Wno-unused-imports   #-}
```
akegalj: @adinapoli-iohk I didn't use autoexporter for this re-export module. Is this re-exporting module in a format you expect it to be?

Martoon-00:
@akegalj Could it be that you didn't import module with `type instance OriginType CCoin = Coin` ?
It lies separately from module where `CCoin` itself defined

wallet-new/test/MarshallingSpec.hs
```diff
@@ -3,11 +3,13 @@ module MarshallingSpec where
 import           Universum
 
 import           Cardano.Wallet.API.V1.Errors (WalletError)
+import           Cardano.Wallet.API.V1.Migration.Types (Migrate (..))
```
akegalj: The modules was restructured in order to be able to import it here as otherwise I was getting overlapping instances:
```
/home/akegalj/projects/serokell/cardano-sl/wallet-new/test/MarshallingSpec.hs:25:9: error:
    • Overlapping instances for ToJSON BackupPhrase
        arising from a use of 'aesonRoundtripProp'
      Matching instances:
        instance ToJSON BackupPhrase
          -- Defined in 'Cardano.Wallet.Orphans.Aeson'
        instance ToJSON BackupPhrase
```

akegalj:
@Martoon-00 
> Could it be that you didn't import module with type instance OriginType CCoin = Coin ?

ah, that's it! Feeling stupid. ty

wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -34,6 +34,8 @@ library
                        Cardano.Wallet.API.V1.Errors
                        Cardano.Wallet.API.V1.Updates
                        Cardano.Wallet.API.V1.Migration
+                       Cardano.Wallet.API.V1.Migration.Helpers
```
adinapoli-iohk: @akegalj I don't like "helper" modules, because they feel largely artificial . Let's call it simply "Cardano.Wallet.API.V1.Migration.Monads", as it would convey the idea that we are migrating between different monadic stacks 😉

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
adinapoli-iohk: What about:

- Removing `-Wno-unused-imports`
- Removing `-Wno-dodgy-exports`
- Change 

```
import           Cardano.Wallet.API.V1.Migration.Helpers
import           Cardano.Wallet.API.V1.Migration.Types
```

into

```
import  qualified Cardano.Wallet.API.V1.Migration.Helpers as Exports
import qualified Cardano.Wallet.API.V1.Migration.Types as Exports
```

- Rename

```
 module Cardano.Wallet.API.V1.Migration.Helpers
, module Cardano.Wallet.API.V1.Migration.Types
 -- * Configuration re-exports
```

into 

```
module Exports
```

Would that still gives you the need for those 2 pragmas?

wallet-new/test/MarshallingSpec.hs
```diff
@@ -51,10 +59,21 @@ spec = describe "Marshalling & Unmarshalling" $ do
...
+migrateRoundtripProp
+    :: (Arbitrary a, Migrate a b, Migrate b a, Eq a, Show a, Typeable a, Typeable b)
+    => proxy a -> proxy b -> Spec
+migrateRoundtripProp proxyA proxyB =
+    prop ("Migrate " <> show (typeRep proxyA) <> " <-> " <> show (typeRep proxyB) <> " roundtrips") (migrateRoundtrip proxyA proxyB)
 
 aesonRoundtrip :: (Arbitrary a, ToJSON a, FromJSON a, Eq a, Show a) => proxy a -> Property
 aesonRoundtrip (_ :: proxy a) = forAll arbitrary $ \(s :: a) -> do
-    eitherDecode (encode (toJSON s)) === Right s
+    eitherDecode (encode s) === Right s
```
adinapoli-iohk: Woops, good catch.

wallet-new/test/MarshallingSpec.hs
```diff
@@ -51,10 +59,21 @@ spec = describe "Marshalling & Unmarshalling" $ do
                 -- currently passphrase should be either empty or of length 32
                 decodingFails @PassPhrase "aabbcc" Proxy
 
+migrateRoundtrip :: (Arbitrary a, Migrate a b, Migrate b a, Eq a, Show a) => proxy a -> proxy b -> Property
+migrateRoundtrip (_ :: proxy a) (_ :: proxy b) = forAll arbitrary $ \(s :: a) -> do
```
adinapoli-iohk: What about renaming `s` into something a bit more descriptive and `a` and `b` into `from` and `to`? I would be curious to see if this function reads better that way :)

ksaric:
If we are able to do the roundtrip, we can maybe have (or something more elaborate, but you get the point - if we can we need to have `Migrate` to be explicit):
```
class Migrate from to where
    eitherMigrateFromTo :: from -> Either Errors.WalletError to
    eitherMigrateToFrom :: to   -> Either Errors.WalletError from
```

akegalj:
```
class Migrate from to where
    eitherMigrateFromTo :: from -> Either Errors.WalletError to
    eitherMigrateToFrom :: to   -> Either Errors.WalletError from
```
@ksaric this won't work for every migration. For example `V0.CWallet` contains more info then `V1.Wallet` thus migrating back from `V1.Wallet` into `V0.CWallet` isn't possible (as `V1.Wallet` doesn't carry enough info)

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
akegalj: @adinapoli-iohk I was able to remove those two pragmas without any further modifications (guess it is an artefact) 

```
module X (module Exports) where

import  qualified Cardano.Wallet.API.V1.Migration.Helpers as Exports
import qualified Cardano.Wallet.API.V1.Migration.Types as Exports
```
doesn't work as I am getting:
```
/home/akegalj/projects/serokell/cardano-sl/wallet-new/src/Cardano/Wallet/API/V1/Migration.hs:4:7: warning: [-Wdodgy-exports]
    The export item `module Exports' exports nothing
             
/home/akegalj/projects/serokell/cardano-sl/wallet-new/src/Cardano/Wallet/API/V1/Migration.hs:14:1: warning: [-Wunused-imports]
    The qualified import of ‘Cardano.Wallet.API.V1.Migration.Helpers’ is redundant
      except perhaps to import instances from ‘Cardano.Wallet.API.V1.Migration.Helpers’
    To import instances alone, use: import Cardano.Wallet.API.V1.Migration.Helpers()
             
/home/akegalj/projects/serokell/cardano-sl/wallet-new/src/Cardano/Wallet/API/V1/Migration.hs:15:1: warning: [-Wunused-imports]
    The qualified import of ‘Cardano.Wallet.API.V1.Migration.Types’ is redundant
      except perhaps to import instances from ‘Cardano.Wallet.API.V1.Migration.Types’
    To import instances alone, use: import Cardano.Wallet.API.V1.Migration.Types()
```
and later it breaks as it `doesn't export anything` so its missing stuff from those modules.

I believe this was possible to do somehow before.

@gromakovsky do you know how to re-export modules with a qualified names?

gromakovsky: AFAIK, for re-exports you should use unqualified imports.

>  I was able to remove those two pragmas without any further modifications (guess it is an artefact)

If there are no warnings without these pragmas, I think you there is nothing bad in using unqualified imports here for re-exports.

adinapoli-iohk: > AFAIK, for re-exports you should use unqualified imports.

Dang, this is what happens when trying to type stuff on top of your head 😁

akegalj: Yes, that works . ty @adinapoli-iohk @gromakovsky

akegalj:
@adinapoli-iohk
CI failures:

 * The job exceeded the maximum time limit for jobs, and has been terminated.
 * Insufficient memory

I believe we can ignore those ?

adinapoli-iohk:
Yes, let's ignore those.

PR #2052, opened 2017-12-06
[CSL-1948] V1 API handler `newAccount`

sectore:
_Task_ https://iohk.myjetbrains.com/youtrack/issue/CSL-1948


adinapoli-iohk:
@sectore This has conflicts now 😅 

(_Move fast and break things_ :trollface: )

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Accounts.hs
```diff
@@ -57,23 +57,17 @@ listAccounts RequestParams {..} = do
...
-        , accAddresses = mempty
-        , accName = uaccName
-        , accWalletId = wId
-        }
+newAccount
+    :: (MonadThrow m, V0.MonadWalletLogic ctx m)
+    => WalletId -> Maybe PassPhrase -> AccountUpdate -> m Account
+newAccount wId mPassPhrase accUpdate =
+    case mPassPhrase of
+        Nothing -> throwM $ Errors.toError Errors.NoPassPhrase
```
adinapoli-iohk: Are we sure this is correct? AFAIK a `PassPhrase` is completely optional and we shouldn't be failing if the wallet was created initially with no `PassPhrase`.

Also, I know this might sound confusing, but we have renamed `PassPhrase` to `SpendingPassword` as it's one of the most confusing points of the old API. There is a type synonym in `Cardano.Wallet.API.V1.Types`. Would you mind using it for external-facing code? Thanks a ton! 😉

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
```diff
@@ -19,7 +19,7 @@ type API
     :<|> "accounts" :> WalletRequestParams
                     :> Summary "Retrieves the full list of Accounts."
                     :> Get '[JSON] (OneOf [Account] (ExtendedResponse [Account]))
-    :<|> "accounts" :> Header  "Daedalus-Passphrase" Text
+    :<|> "accounts" :> Header  "Daedalus-Passphrase" PassPhrase
```
adinapoli-iohk: Let's use `SpendingPassword`, and rename `Daedalus-Passphrase` into `Daedalus-Spending-Password` or similar.

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Accounts.hs
```diff
@@ -57,23 +57,17 @@ listAccounts RequestParams {..} = do
...
-    return Account
-        { accId = newId
-        , accAmount = Core.mkCoin 0
-        , accAddresses = mempty
-        , accName = uaccName
-        , accWalletId = wId
-        }
+newAccount
+    :: (MonadThrow m, V0.MonadWalletLogic ctx m)
+    => WalletId -> Maybe PassPhrase -> AccountUpdate -> m Account
```
adinapoli-iohk: What about creating a type synonym for `AccountUpdate` and call it `NewAccount` (in `V1.Types` ?) I guess would allow programmers to better understand what's going on.

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -50,6 +50,7 @@ data WalletError =
     | SomeOtherError { weFoo :: !Text, weBar :: !Int }
     | MigrationFailed { weDescription :: !Text }
     | WalletNotFound
+    | NoPassPhrase
```
adinapoli-iohk: I think this shouldn't be an error. A user can surely not pass a spending password, as it's optional. Maybe let's rename it to `InvalidSpendingPassword`, to be thrown when the passwords do not match?

wallet-new/src/Cardano/Wallet/API/V1/Migration.hs
```diff
@@ -159,11 +170,11 @@ instance Migrate V0.CAccountId V1.WalletId where
         (walletId, _) :: (V1.WalletId, V1.AccountId) <- eitherMigrate oldAccountId
         pure walletId
 
+instance Migrate V1.PassPhrase Crypto.PassPhrase where
```
adinapoli-iohk: Are we sure we need this? AFAIK a `V1.PassPhrase` should be exactly equivalent to the Core/Crypto `PassPhrase`. Isn't this the case?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -83,6 +84,8 @@ import qualified Pos.Crypto.Signing as Core
 -- base16-encoded string.
 type SpendingPassword = Core.PassPhrase
 
+type PassPhrase = Text
```
adinapoli-iohk: Ah, this shouldn't be a text, but rather we should be using `SpendingPassword`.

wallet-new/server/Cardano/Wallet/API/V1/Handlers/Accounts.hs
sectore: @adinapoli-iohk `V0.newAccount` (https://github.com/input-output-hk/cardano-sl/blob/master/wallet/src/Pos/Wallet/Web/Methods/Logic.hs#L210-L212) does need a `PassPhrase`, so it seems that `PassPhrase` is not optional. Could you give me more details, pls?

sectore: Ah, I see... We have similar "issue" solved in `V1.newWallet`. Pls ignore my previous question ^, @adinapoli-iohk

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -4,11 +4,13 @@
 {-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE TemplateHaskell            #-}
+{-# LANGUAGE StandaloneDeriving         #-}
```
adinapoli-iohk: Why do we need `StandaloneDeriving` here? If that's just for `Generic` deriving, cannot we use `DeriveGeneric`?

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
```diff
@@ -19,9 +17,8 @@ type API
     :<|> "accounts" :> WalletRequestParams
                     :> Summary "Retrieves the full list of Accounts."
                     :> Get '[JSON] (WalletResponse [Account])
-    :<|> "accounts" :> Header  "Daedalus-Passphrase" Text
-                    :> Summary "Creates a new Account for the given Wallet."
-                    :> ReqBody '[JSON] (New Account)
+    :<|> "accounts" :> Summary "Creates a new Account for the given Wallet."
+                    :> ReqBody '[JSON] NewAccount
```
adinapoli-iohk: I think there is still value in leaving `(New Account)` there, because it means we can change the underlying type (if we ever want) without breaking the API type signature.

PR #2053, opened 2017-12-06
[CSL-2027] Use servant-generic in wallet

neongreen:


neongreen:
> We should instead have in return NoContent and amend the handler as needed.

Should I also include this change into my PR?

akegalj:
> Should I also include this change into my PR?

@neongreen I believe yes, as swagger produces old and new api docs.

adinapoli-iohk:
> Should I also include this change into my PR?

If you don't want to cross-boundaries, I'm even happy if you:

* Fix the issue locally (without committing the changes);
* Run `cardano-node-swagger` and check the the current `wallet` API now produces a valid Swagger spec;
* We open a YT ticket to address the `NoContent` thingie in a separate PR.

Whatever works for you, really 😉

neongreen:
image

neongreen:
Seems to work.

neongreen:
> If you don't want to cross-boundaries

Nah, I don't like extra bureaucracy :)

adinapoli-iohk:
Awesome @neongreen ! Excited to see how `servant-generic` works out for us. AppVeyor seems lately to be plagued with issues like these:

```
The same package name is used in multiple local packages
servant-generic used in:
- PLIndex "servant-generic-0.1.0.1"
- PLOther (PLRepo (Repo {repoUrl = "https://github.com/chpatrick/servant-generic", repoCommit = "3d2292c50fd72026a00a7f7fd44c00936261dbc5", repoType = RepoGit, repoSubdirs = "."}))
```

Any clue what's been causing them?

neongreen:
Yeah. The elusive cause is “Me forgetting how stack.yaml works” :|

PR #2054, opened 2017-12-06
[CSL-2033] Simplify new wallet response type

adinapoli-iohk:
This PR ports every endpoint in `wallet-new` to use `WalletResponse`, which was called `ExpandedResponse` before. `OneOf` is gone.

all_expanded

cc @DominikGuzei


wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -444,10 +390,9 @@ $errorExample
 |]
 
 data DescriptionEnvironment = DescriptionEnvironment {
```
et4te: Record fields do not follow conventions here which make them prone to clash on import (..).

```haskell
data DescriptionEnvironment = DescriptionEnvironment {
    deErrorExample :: !Text
  , deDefaultPerPage :: !Text
  ...
}
```

wallet-new/src/Cardano/Wallet/API/Response.hs
```diff
@@ -43,36 +45,22 @@ deriveJSON defaultOptions { constructorTagModifier = map Char.toLower . reverse
...
 
--- | An `ExtendedResponse` allows the consumer of the API to ask for
--- more than simply the result of the RESTful endpoint, but also for
--- extra informations like pagination parameters etc.
-data ExtendedResponse a = ExtendedResponse
-  { extData   :: a
+-- | An `WalletResponse` models, unsurprisingly, a response (successful or not)
+-- produced by the wallet backend.
+-- Includes extra informations like pagination parameters etc.
+data WalletResponse a = WalletResponse
```
et4te: Record fields do not follow conventions here and thus prone to clash on import (..).

wallet-new/src/Cardano/Wallet/API/Response.hs
```diff
@@ -91,16 +79,16 @@ instance (Arbitrary a, Arbitrary b) => Arbitrary (OneOf a b) where
 respondWith :: (Foldable f, Monad m)
             => RequestParams
             -> (RequestParams -> m (f a))
-            -> m (OneOf [a] (ExtendedResponse [a]))
+            -- ^ A callback-style function which, given the full set of `RequestParams`
+            -- produces some form of results in some 'Monad' @m@.
+            -> m (WalletResponse [a])
 respondWith params@RequestParams{..} generator = do
```
ksaric: Since we are already creating abstractions before they are actually used, maybe use something like:
```
class WalletResp a where
    size        :: Int
    totalPages  :: Int
```
Where we can abstract that over both a regular list `[a]` and something like a `Map k v` since we have a KV storage.
We can then have:
```
instance WalletResp [a] where
...
```
And something like:
```
instance WalletResp (Map k v) where
...
```
Bear in mind, these are initial ideas, can be improved.

wallet-new/src/Cardano/Wallet/API/Response.hs
```diff
@@ -117,3 +105,12 @@ paginate PaginationParams{..} rawResultSet =
                                }
         slice                  = take pp . drop ((cp - 1) * pp) . toList
     in (slice rawResultSet, metadata)
+
+
+-- | Creates a 'WalletResponse' with just a single record into it.
+single :: a -> WalletResponse a
```
ksaric: Also, we could probably abstract over whether the response is `single` or a list with a typeclass and a function `walletResponse` - that `single` is visually annoying to me.

wallet-new/src/Cardano/Wallet/API/Response.hs
adinapoli-iohk: @ksaric I don't want to rub you the wrong way, but I have a bit of an aversion for typeclasses used to encapsulate some form of interface 😬 😬 . In general, I resort to them when I want to stick to SPJ's definition of a way to achieve "bounded polymorphism" or when the program structure dictates it (for example `Migrate` is an example where it might make sense to have it).

For things like `WalletResponse` I think a plain record is better: It makes trivial for a newcomer to understand what's going on rather than having all these different instances scattered throughout the program which tell how to build a response. In general, I think [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python) and I'm willing to pay some price for it (in terms of boilerplate) 😉

wallet-new/src/Cardano/Wallet/API/Response.hs
adinapoli-iohk: >  that single is visually annoying to me.

It's just a plain function: what's annoying about it? 😉  Maybe you would like it to be named differently?

ksaric: It just seems to be boilerplate since most of our "regular" responses are `single`, nothing else.

adinapoli-iohk: > It just seems to be boilerplate since most of our "regular" responses are single, nothing else.

Valid point, although one may argue it's only 6 chars more to write every time, so not too bad 😉  It's not obvious to me how to do it in a simple yet transparent way (well, there are plenty of options, but you pay a price one way or the other). Ok, here is my proposal: will crack on with merging the PR as-it-is (after I fix the style guide nitpicks) and open a ticket to discuss with the team how we can get rid of the boilerplate in a concide manner (aka having our cake and eating it too :) )

PR #2055, opened 2017-12-06
[CSL-1822] Refine logging

gromakovsky:
There are various changes related to logging: what we log, which severity messages have, how logging is configured, etc.
Commit history is descriptive, imo. Most commits have few sentences of explanation.

I've almost finished, the only missing thing is to review middleware logs, will do it soon.

### DIFF
--- a/auxx/Main.hs
+++ b/auxx/Main.hs
@@ -11,7 +11,7 @@ import           Formatting (sformat, shown, (%))
 import           Mockable (Production, currentTime, runProduction)
 import qualified Network.Transport.TCP as TCP (TCPAddr (..))
 import qualified System.IO.Temp as Temp
-import           System.Wlog (logInfo)
+import           System.Wlog (LoggerName, logInfo)
 
 import qualified Pos.Client.CLI as CLI
 import           Pos.Communication (OutSpecs, WorkerSpec)
@@ -31,7 +31,10 @@ import           Pos.WorkMode (EmptyMempoolExt, RealMode)
 import           AuxxOptions (AuxxAction (..), AuxxOptions (..), AuxxStartMode (..), getAuxxOptions)
 import           Mode (AuxxContext (..), AuxxMode, CmdCtx (..), realModeToAuxx)
 import           Plugin (auxxPlugin, rawExec)
-import           Repl (WithCommandAction(..), withAuxxRepl)
+import           Repl (WithCommandAction (..), withAuxxRepl)
+
+loggerName :: LoggerName
+loggerName = "auxx"
 
 -- 'NodeParams' obtained using 'CLI.getNodeParams' are not perfect for
 -- Auxx, so we need to adapt them slightly.
@@ -88,7 +91,7 @@ action opts@AuxxOptions {..} command = do
                   $ withConfigurations conf configToDict
             mode <$ case mode of
                 Nothing -> printAction "Mode: light"
-                _ -> printAction "Mode: with-config"
+                _       -> printAction "Mode: with-config"
         Light -> return Nothing
         _ -> withConfigurations conf configToDict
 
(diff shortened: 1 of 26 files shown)


auxx/Main.hs
```diff
@@ -31,7 +31,10 @@ import           Pos.WorkMode (EmptyMempoolExt, RealMode)
 import           AuxxOptions (AuxxAction (..), AuxxOptions (..), AuxxStartMode (..), getAuxxOptions)
 import           Mode (AuxxContext (..), AuxxMode, CmdCtx (..), realModeToAuxx)
 import           Plugin (auxxPlugin, rawExec)
-import           Repl (WithCommandAction(..), withAuxxRepl)
+import           Repl (WithCommandAction (..), withAuxxRepl)
+
+loggerName :: LoggerName
```
volhovm: I believe that defining a top level function that is used once only doesn't make any sense.

lib/src/Pos/Launcher/Resource.hs
```diff
@@ -319,7 +319,7 @@ createKademliaInstance ::
     -> Word16 -- ^ Default port to bind to.
     -> m KademliaDHTInstance
 createKademliaInstance BaseParams {..} kp defaultPort =
-    usingLoggerName (lpRunnerTag bpLoggingParams) (startDHTInstance instConfig defaultBindAddress)
+    usingLoggerName (lpDefaultName bpLoggingParams) (startDHTInstance instConfig defaultBindAddress)
```
volhovm: I think that it'd better use a dedicated `kademlia` logging name instead.

lib/src/Pos/Launcher/Scenario.hs
```diff
@@ -98,13 +97,6 @@ runNode' NodeResources {..} workers' plugins' = ActionSpec $ \vI sendActions ->
         firstGenesisHash
         (gdFtsSeed genesisData)
 
-    lastKnownEpoch <- LrcDB.getEpoch
```
volhovm: Why don't we print leaders on start now?

log-config-prod.yaml
```diff
@@ -1,18 +1,9 @@
 rotation:
     logLimit: 5242880 # 5MB
     keepFiles: 20
-severity: Info
+severity: Debug
```
volhovm: Are you sure? 🤔

scripts/log-templates/log-config-qa.yaml
```diff
@@ -1,13 +1,11 @@
 # This template is used in scripts to connect to a running cluster.
 
 rotation:
-    logLimit: 2097152 # 2MB
-    keepFiles: 20
+    logLimit: 104857600 # 100MB
+    keepFiles: 2
```
volhovm: If it's for qa only, i would suggest something in between. Say, 50mb and 10 files.

log-config-prod.yaml
gromakovsky: Read vatnik.

lib/src/Pos/Launcher/Scenario.hs
gromakovsky: Read vatnik.

auxx/Main.hs
gromakovsky: Oh, I forgot to use it in another place where we still have `"auxx"` hardcoded. Will fix.

lib/src/Pos/Launcher/Resource.hs
gromakovsky: I preserved existing logic, but ok, I think it makes sense indeed.

log-config-prod.yaml
gromakovsky: Previously we had `Debug` for `node` (where almost all messages go) and `Info` for the rest (where ε messages go). Now we have `Debug` for all. It's quite minor.

scripts/log-templates/log-config-qa.yaml
gromakovsky: QA is confusing name. It's for developers who use `mainnet.sh` and similar stuff.

gromakovsky:
> BTW, can you remind what is the motivation behind default logger name?

Do you mean `lpDefaultName`? Well, maybe we can just use `mempty` by default. I think it's quite minor.

volhovm:
>Do you mean lpDefaultName? Well, maybe we can just use mempty by default. I think it's quite minor.

If we can delete a useless field from node context, i think we should. Anyway, I'm looking at this again.

scripts/log-templates/log-config-qa.yaml
volhovm: Yep, I know (I'm using it), the point was "we don't care about space since it's for devs/testing only".

PR #2056, opened 2017-12-06
docs: run `stack setup` to install `ghc` first.

5chdn:
Else you will run into the following error:

> No compiler found, expected minor version match with ghc-8.0.2 (x86_64) (based on resolver setting in /home/user/.opt/cardano-sl/stack.yaml).
>
> To install the correct GHC into /home/user/.stack/programs/x86_64-linux-nix/, try running "stack setup" or use the "--install-ghc" flag. To use your system GHC installation, run "stack config set system-ghc --global true", or use the "--system-ghc" flag.


adinapoli-iohk:
@5chdn Sorry for the uber-delay, and thanks for your contribution!

PR #2059, opened 2017-12-07
[CSL-1842] Write the guidelines for exception handling

int-index:
Rendered: https://github.com/input-output-hk/cardano-sl/blob/int-index/csl-1842/docs/exceptions.md

### DIFF
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -1,0 +1,267 @@
+# Guidelines for Exception Handling in Cardano SL
+
+To determine the preferred way to handle exceptions in a piece of code, first
+answer these questions:
+
+1. Is this a programmer mistake or a regular error?
+2. Is the code 100% pure or potentially impure?
+
+Depending on the answers, an appropriate approach can be selected.
+
+## Is this a programmer mistake or a regular error?
+
+### Programmer mistake
+
+When a particular codepath is not meant to be ever executed, but happens to be
+executed anyway, this is a programmer mistake.
+
+For instance, consider this piece of code:
+
+```
+-- Precondition (unchecked): input lists are the same length.
+zipSameLen :: [a] -> [b] -> [(a, b)]
+zipSameLen [] [] = []
+zipSameLen (a:as) (b:bs) = (a,b) : zipSameLen as bs
+zipSameLen _ _ = error "Lists of different length, precondition violated!"
+```
+
+The `zipSameLen` function is meant to work on two lists of the same length, and
+calling it with lists of different length is a mistake on the programmer's part.
+Code like `zipSameLen [] [1]` must be rejected at code review stage. However,
+the function still must have the `zipSameLen _ _` case to have exhaustive
+pattern matching.
+
+A function that can be used erroneously (has a potential for programmer mistakes)
+MUST have a comment that describes its preconditions. When it is called, the
+call MUST have an explanation that assures why these preconditions hold.
+
+For instance, when a programmer calls `Prelude.last xs`, he must add a comment
+at this call site that explains why `xs` is guaranteed to be non-empty, even if
+it seems obvious. A better option is to use functions and types that don't allow
+for errors: instead of `Prelude.last` one can use `NonEmpty.last`.
+
+In case of `zipSameLen` we could use length-indexed vectors:
+
+```
+zipSameLen :: Vec n a -> Vec n b -> Vec n (a, b)
+```
+
+It is always a tradeoff whether to allow programmer mistakes, or do type-level
+trickery to avoid them. The decision process for this is out of scope of this
+document.
+
+### Regular error
+
+Consider a function like `readFile`. It is entirely possible that the path
+passed to it may point to a file that does not exist, and the programmer can't
+do anything to prevent this. Existence of a file is not a property of our code,
+it's the property of the outside world, and we have to deal with all possible
+scenarios.
+
+Another example is parsing user input. We might expect the user to enter a
+number, but we have to consider the case that the user enters something else.
+
+In cases when the erroneous scenarios are out of our control, we consider
+these to be regular errors.
+
+### Identifying error class
+
+Can the execution of the error codepath be excluded by code review and static
+verification (more precise types?)
+
+* Yes: it is a programmer mistake
+* No: it is a regular error
+
+## Is the code 100% pure or potentially impure?
+
+### Impure code
+
+The code is considered impure when it's an `IO` action, a function that returns
+an `IO` action, or similar. The code can be _potentially_ impure when it is
+written in an abstract monad that can be instantiated to `IO`.
+
+Definitely impure:
+
+```
+readFile :: FilePath -> IO String
+```
+
+Potentially impure:
+
+```
+lookupThrow :: MonadThrow m => Map k v -> k -> m v
+```
+
+### Pure code
+
+Pure code is one that does not use `IO` operations. For example:
+
+```
+lookupMaybe :: Map k v -> k -> Maybe v
+```
+
+Notice that while it is conceivable that we have a `Map` of `IO` actions, and
+instantiate `lookupMaybe` to `Map k (IO ()) -> k -> Maybe (IO ())`, the
+instantiation does not inspect/use these `IO` actions, so it's not considered
+potentially impure.
+
+### Identifying code purity
+
+Is there an instantiation of type parameters that would mean that there are
+`IO` actions inspected/used in the code?
+
+* Yes: it is potentially impure code
+* No: it is 100% pure code
+
+## Error handling practices
+
+### Pure code, programmer mistakes
+
+Do *not*:
+
+* do this often, as if it's okay
+* use `undefined`
+* use non-exhaustive pattern matching
+
+Do:
+
+* try to use types to avoid the need in the first place
+* comment extensively (invariants and precondition, reasoning)
+* use the `error` function (or `impureThrow` with custom exception)
+* use the `HasCallStack` feature
+
+Before writing code that allows programmer mistakes, consult with colleagues how
+it would be possible to prevent them statically. There might be tricks that
+you're not aware of. Always make this trade-off conciously and responsibly.
+
+(On the other hand, if static guarantees require GADTs or type families, perhaps
+it's better to not overcomplicate code. Use your judgement.)
+
+DISCUSSION: Should we create a synonym `bug = impureThrow` in Universum? This
+would make the intention more clear.
+
+### Pure code, regular errors
+
+Do *not*:
+
+* use `error` or `impureThrow`
+* use `MonadFail`
+* return `Either Text`
+
+Do:
+
+* return `Either ErrorADT`, `Maybe`
+* wrap the underlying (pure!) monad in `ExceptT` or `CatchT`
+* use `MonadError` or `MonadThrow`
+
+Consider parsing: it is pure, but we cannot make assumptions about the input. In
+this case we might want to use `ExceptT ParseError`. Or consider a lookup in a
+`Map`, where we don't know whether the key is present -- in this case we'd like
+to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
+
+* `Maybe`, `Either e`
+* `MaybeT`, `ExceptT e`
+* `CatchT`
+
+Avoid using `Text` with the error message in place of `e` -- create a proper
+ADT. In case creating a proper ADT feels too cumbersome, use `CatchT`, which
+is equivalent to `ExceptT SomeException`.
+
+Be careful not to use `MaybeT`, `ExceptT`, and `CatchT` in potentially impure
+code. When in doubt whether the code is potentially impure, use `MonadThrow`.  
+(The reason we don't want `ExceptT` and co. in potentially impure code is that
+they add additional exception mechanisms to the one that `IO` has, and
+`catch`/`bracket` don't account for this).
+
+### Impure code, regular errors
+
+Do *not*:
+
+* use `error` or `impureThrow`
+* use `ExceptT`, `MaybeT`, or `CatchT`
+* use `Maybe` or `Either`
+* use `MonadError`
+* use `throwIO`
+
+Do:
+
+* create a custom exception type
+* use `throwM` (`MonadThrow`)
+
+We disallow the use of `throwIO` only because it is redundant in the presence of
+`throwM` and requires a stronger constraint (`MonadIO` rather than
+`MonadThrow`).
+
+Derive prisms for exception types with multiple constructors, so it's convenient
+to use them with `catchJust`.
+
+### Impure code, programmer mistakes
+
+Use the same techniques as in pure code -- `error` or `impureThrow`. There are
+two reasons for this:
+
+* we would rather catch the error sooner than later, and `impureThrow` explodes
+  when forced to WHNF, while `throwIO` or `throwM` explode when executed
+
+* when in an abstract (but potentially impure) monad, using `throwM` might
+  add an additional constraint
+
+## Packages and modules
+
+Do not import `Control.Exception` or `Control.Monad.Catch`! We use the
+`safe-exceptions` to deal with asynchronous exceptions gracefully, so import
+`Control.Exception.Safe`.
+
+## Resource handling
+
+Use `bracket` or to guarantee the release of resources. In case of concurrent
+code, avoid `forkIO` or `forkProcess` in favor of the `async` package, as it
+rethrows exceptions from the child threads. (Do not use the function `async`
+itself when you can use `withAsync`, `race`, or `concurrently`).
+
+When resource usage is non-linear, it's okay to use `ResourceT`, but prefer
+`bracket` whenever possible.
+
+## Migration
+
+We should get rid of `Mockable Throw`, `Mockable Catch`, and `Mockable Bracket`,
+as they buy us nothing compared to `MonadThrow`, `MonadCatch`, and `MonadMask`,
+but have less ecosystem support -- for instance, the `safe-exceptions` package
+doesn't use them.
+
+We should identify the parts of the code that use `ExceptT` in impure or
+potentially impure code and replace them with exceptions.
+
+We should make sure that no code imports `Control.Exception` or
+`Control.Monad.Catch`, and use `Control.Exception.Safe` instead.
+
+We should locate all usages of `forkIO` and replace with appropriate functions
+from `async`.
+
+We should find where errors which are not programmer mistakes are thrown with
+`error`, `undefined`, or `impureThrow`, and rewrite them to use correct error
+handling method. This includes usages of partial functions, such as `read`.
+
+We should find where errors are represented by `Text` and create dedicated data
+types to represent them.
+
+We should find places where preconditions/invariants are not reflected in the
+comments, and add comments.
+
+### Code references
+
+This list is not exhaustive:
+
+* `TxpGlobalVerifyMode`: https://github.com/input-output-hk/cardano-sl/blob/8507d03ba928e07daea57f9a52f6dc03a9d65779/txp/Pos/Txp/Settings/Global.hs#L37
+  -- `MonadError` in potentially impure code
+* `MonadRecoveryInfo`:
+  https://github.com/input-output-hk/cardano-sl/blob/d598003ccde5d9848a11c54eec7542b299eb7c44/lib/src/Pos/Recovery/Instance.hs#L30
+  -- `ExceptT` in impure code
+* TBD
+
+## Literature
+
+The following documents were used to create these guidelines:
+
+* https://www.schoolofhaskell.com/user/commercial/content/exceptions-best-practices
+* https://github.com/fpco/safe-exceptions#readme
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -152,7 +152,8 @@ Do:
 
 * return `Either ErrorADT`, `Maybe`
 * wrap the underlying (pure!) monad in `ExceptT` or `CatchT`
-* use `MonadError` or `MonadThrow`
+* use `MonadError` or `MonadThrow` (methods of these classes). Note
+  that if you define `f :: MonadError m => m ()`, it won't be pure
 
 Consider parsing: it is pure, but we cannot make assumptions about the input. In
 this case we might want to use `ExceptT ParseError`. Or consider a lookup in a
@@ -163,30 +164,40 @@ to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
 * `MaybeT`, `ExceptT e`
 * `CatchT`
 
-Avoid using `Text` with the error message in place of `e` -- create a proper
-ADT. In case creating a proper ADT feels too cumbersome, use `CatchT`, which
-is equivalent to `ExceptT SomeException`.
+Avoid using `Text` with the error message in place of `e` -- create a
+proper ADT. In case creating a proper ADT feels too cumbersome, use
+`CatchT`, which is equivalent to `ExceptT SomeException`. Note,
+however, that using `SomeException` in pure code is not the best
+practice, because the set of all possible exceptions is statically
+known. Use it only if you are lazy to define yet another ADT.
 
 Be careful not to use `MaybeT`, `ExceptT`, and `CatchT` in potentially impure
-code. When in doubt whether the code is potentially impure, use `MonadThrow`.  
+code. When in doubt whether the code is potentially impure, use `MonadThrow`.
 (The reason we don't want `ExceptT` and co. in potentially impure code is that
 they add additional exception mechanisms to the one that `IO` has, and
 `catch`/`bracket` don't account for this).
 
-### Impure code, regular errors
+### [Potentially] Impure code, regular errors
 
 Do *not*:
 
 * use `error` or `impureThrow`
 * use `ExceptT`, `MaybeT`, or `CatchT`
-* use `Maybe` or `Either`
 * use `MonadError`
 * use `throwIO`
+* return `m (Either e a)` if `e` has `Exception` instance
 
 Do:
 
 * create a custom exception type
 * use `throwM` (`MonadThrow`)
+
+If you want to return `m (Either e a)` from a function, it's
+recommended to define `instance TypeError "NOT AN EXC" => Exception e`
+for your type `e`. If the meaning of `e` type is not related to
+exceptional situations at all, it's not needed to define such
+instance. But for example if `e` denotes a `ParseError`, please do
+define it.
 
 We disallow the use of `throwIO` only because it is redundant in the presence of
 `throwM` and requires a stronger constraint (`MonadIO` rather than
@@ -195,7 +206,7 @@ We disallow the use of `throwIO` only because it is redundant in the presence of
 Derive prisms for exception types with multiple constructors, so it's convenient
 to use them with `catchJust`.
 
-### Impure code, programmer mistakes
+### [Potentially] Impure code, programmer mistakes
 
 Use the same techniques as in pure code -- `error` or `impureThrow`. There are
 two reasons for this:
@@ -219,8 +230,11 @@ code, avoid `forkIO` or `forkProcess` in favor of the `async` package, as it
 rethrows exceptions from the child threads. (Do not use the function `async`
 itself when you can use `withAsync`, `race`, or `concurrently`).
 
-When resource usage is non-linear, it's okay to use `ResourceT`, but prefer
-`bracket` whenever possible.
+When resource usage is non-linear, it's okay to use `ResourceT`, but
+prefer `bracket` whenever possible. Non-linear resource usage is
+anything that doesn't fit into the “allocate, use, deallocate”
+pattern.
+
 
 ## Migration
 
@@ -229,8 +243,21 @@ as they buy us nothing compared to `MonadThrow`, `MonadCatch`, and `MonadMask`,
 but have less ecosystem support -- for instance, the `safe-exceptions` package
 doesn't use them.
 
-We should identify the parts of the code that use `ExceptT` in impure or
-potentially impure code and replace them with exceptions.
+We should identify the parts of the code that use `ExceptT` or
+`MonadError` in impure or potentially impure code and replace them
+with exceptions. If code can be made pure by replacing `MonadError`
+with simple `Either`, we should do this replacement. For instance,
+`mkMultiKeyDistr :: MonadError Text m => Map StakeholderId CoinPortion
+-> m AddrStakeDistribution` becomes `mkMultiKeyDistr :: Map
+StakeholderId CoinPortion -> Either Text AddrStakeDistribution`
+
+We should get rid of `MonadFail` usages everywhere except places where
+it's required by external API. For instance, `mkTx :: MonadFail m =>
+NonEmpty TxIn -> NonEmpty TxOut -> TxAttributes -> m Tx` should be
+changed to `mkTx :: NonEmpty TxIn -> NonEmpty TxOut -> TxAttributes ->
+Either Text Tx`. We can use `eitherToFail` helper function if we want
+to use this function inside a monad from external library which uses
+`MonadFail` to fail.
 
 We should make sure that no code imports `Control.Exception` or
 `Control.Monad.Catch`, and use `Control.Exception.Safe` instead.


docs/exceptions.md
```diff
@@ -0,0 +1,246 @@
...
+We should get rid of `Mockable Throw` and `Mockable Catch`, as they buy us
+nothing compared to `MonadThrow` and `MonadCatch` but have less ecosystem
+support -- for instance, the `safe-exceptions` package doesn't use them.
+
+We should identify the parts of the code that use `ExceptT` in impure or
+potentially impure code and replace them with exceptions.
+
+We should make sure that no code imports `Control.Exception` or
+`Control.Monad.Catch`, and use `Control.Exception.Safe` instead.
+
```
gromakovsky: I wrote few more things which I want to be done. But I admit they might take quite a lot of time and I don't know how much time we can spend on this Epic.
If I had to choose only the most important points from those, I would choose these three 👍 
But I think it'd be good to also add other improvements which I suggested in comments to CSL-1842 (avoiding `Text` for errors, avoiding exceptions leakages due to `fork`s, etc.). Maybe we don't have time to do them within this Epic, but at least it will provide broader picture of the problems we currently have.

docs/exceptions.md
```diff
@@ -0,0 +1,246 @@
...
+(The reason we don't want `ExceptT` and co. in potentially impure code is that
+they add additional exception mechanisms to the one that `IO` has, and
+`catch`/`bracket` don't account for this).
+
+### Impure code, regular errors
+
+Do *not*:
+
+* use `error` or `impureThrow`
+* use `ExceptT`, `MaybeT`, or `CatchT`
```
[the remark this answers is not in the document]
int-index: I am still not convinced that using `ExceptT` here is a good idea. I believe that it can be safe when used carefully, but:

* now you don't have `MonadMask` instance, meaning that you might need to rewrite code if it needs to allocate resources in the future

* now you have to lift through an additional transformer, which might contradict the practices we're establishing in CSL-1823

* the rules for exception usage become more nuanced: is it ok to do this if it's in a helper function? what if I move this helper function to top-level? what if I put it into a separate module for reuse?

I've seen your comment that you want to allow usage like this, but I don't think that it pulls its weight. Are you sure that doing `throwM` and `try` wouldn't give you the same thing?

The transformation is mechanical:

```
        fmap convertRes . runExceptT $ do
-- becomes
        fmap convertRes . try @SyncStatus $ do
```

```
                True -> throwError SSDoingRecovery
-- becomes
                True -> throwM SSDoingRecovery
```

Adding an `Exception` instance does not sound bad to me, it's derived by GHC so you don't have to write any code.

It looks like we get the same benefits without any disadvantages if we `throwM` here.

docs/exceptions.md
int-index: > avoiding Text for errors, avoiding exceptions leakages due to forks, etc

I included these in the guidelines, but not in the migration plan. I'll make it more explicit and copy these to the migration plan.

docs/exceptions.md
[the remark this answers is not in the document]
gromakovsky: > now you don't have `MonadMask` instance, meaning that you might need to rewrite code if it needs to allocate resources in the future

I still have `MonadMask` instance, I can just use `lift` where I need it. The only place where I can't use `lift` is `throwError`, but I don't need `MonadMask` there.

> now you have to lift through an additional transformer, which might contradict the practices we're establishing in CSL-1823

It's hard to talk about it until we come to a conclusion about CSL-1823, but I am already a bit concerned about that. If we are going to prohibit all lifting, what will we do with our tests which live in `PropertyM`? There we use `lift` in many places. AFAIU, `ExceptT` case is the same. Maybe it should be discussed in CSL-1823, not here. Maybe I am just missing something.

> the rules for exception usage become more nuanced: is it ok to do this if it's in a helper function? what if I move this helper function to top-level? what if I put it into a separate module for reuse?

It's not hard to come up with concrete rules. For example, we can prohibit `ExceptT` in types of all top-level impure or potentially impure functions. If you have a nested function which has `ExceptT` in its type and you want to make it top-level, you can change its type to `m (Either Patak Sepulka)` and convert it to `ExceptT` where it's used (which is straightforward).

> It looks like we get the same benefits without any disadvantages if we `throwM` here.

Yes, I agree that there no obvious disadvantages in using `throwM`. I never had good arguments why `ExceptT` is preferable in this case. It's mostly based on my personal feeling/intuition, rather than some fact that make `ExceptT` definite winner.
1. For me `SyncStatus` conceptually is not an exception, that's why I find it a bit counterintuitive to throw it as runtime exception. And defining its `Exception` instance. It's not about writing extra code, it's just about intuition behind this. `ExceptT` name also suggests something exceptional, but I treat it simply as a transformer which provides multiple exit points.
2. `ExceptT` version is basically just sugar for the version with multiple `if`s/`case`s. And the version with multiple `case`s is exactly how we want the code to behave. It's really clear how it works if you desugar it. On the other hand, version with `throwM` involves real runtime exceptions which are implemented _somehow_.

Anyway, as I said, I don't have and never had solid arguments in favour of `ExceptT` in this case, it's mostly my personal preference. I can easily discard this preference if others prefer `throwM` version in this case.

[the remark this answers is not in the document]
gromakovsky: Btw, with `throwM` approach someone can make joke and throw `SSInFuture` from e. g. `DB.getTipHeader` function and we'll catch it and return `SSInFuture`, though perhaps we shouldn't catch it.
It's something exotic and nobody will do it of course, so you can ignore it, just something came into my mind :)

docs/exceptions.md
```diff
@@ -0,0 +1,266 @@
...
+DISCUSSION: Should we create a synonym `bug = impureThrow` in Universum? This
+would make the intention more clear.
+
+### Pure code, regular errors
+
+Do *not*:
+
+* use `error` or `impureThrow`
+* use `MonadFail`
+* return `Either Text`
```
volhovm: Why not? Sometimes we have `someFoo = runExceptT someFooDo` because every usecase would anyway run `runExceptT`  on the `someFooDo`. Or do you mean that returning `Text` error is bad? It's a reasonable tradeoff, i guess.

docs/exceptions.md
```diff
@@ -0,0 +1,266 @@
...
+Consider parsing: it is pure, but we cannot make assumptions about the input. In
+this case we might want to use `ExceptT ParseError`. Or consider a lookup in a
+`Map`, where we don't know whether the key is present -- in this case we'd like
+to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
+
+* `Maybe`, `Either e`
+* `MaybeT`, `ExceptT e`
+* `CatchT`
+
+Avoid using `Text` with the error message in place of `e` -- create a proper
```
volhovm: Again, I disagree. Consider that we're writing a message handler. What `impureFunc` returns is either data or error message to user.  Why would one use adt here? We never `catch` this error.
```
listener msg = 
  either respondError respondSuccess $ impureFunc msg
```
    
Rephrasing, sometimes `Either A B` is not a "way of exception handling", but is a handy way to return a result. One more example: let `checkData` be a impure function that has possible outcomes:
* Exit with success ~ "data is alright"
* Exit with failure ~ "data is not alright because T"
* Exit with normal error ~ "database became inaccessible (if we communicate w/ it over web)" or "we tried to write a log to file, but it doesn't exist".
* Exit with programmer error ~ "something went really wrong and it's not expected"

What you seem to miss is that sometimes it's easier to represent 1/2 difference with `Either Text a` instead of a custom ADT like `DataVerSuccess | DataVerErrorX | DataVerErrorY | ...`.

docs/exceptions.md
int-index: This is about `Text`. Define proper ADTs for your errors.

volhovm: I can't see why is it always reasonable.

docs/exceptions.md
int-index: Defining an ADT in Haskell is cheap. You never know what the caller of your function might want to do with the `Left` case, perhaps it has a way to handle specific cases. You assume that the only thing done to the error is displaying it, which is a strong assumption, and may change with time.

To be clear, this rule applies to top-level functions, where you cannot know what the caller intends to do with the return value. Maybe we want to log the error as JSON rather than text, maybe we want to return it from a HTTP request (and assign an appropriate error code), maybe we want to display it in a different language (Japanese), etc...

Just be nice to the consumers of your API and don't return `Text`. Instead, return an `ADT` and a function (or a `Buildable` instance) to display the error.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: > I still have `MonadMask` instance, I can just use lift where I need it. 

Not quite, you can't do something like this:

```
runExceptT $ do
  doStuff
  bracket ... $ do
    if b
      then throwError e -- won't work
      else return x
```

> If we are going to prohibit all lifting, what will we do with our tests which live in PropertyM?

I believe we can use it as our base monad, because the only transformer on top that we want is `ReaderT` and I am since adding `ReaderT` is equivalent to adding a parameter to a function, I believe it should be always possible.

>  If you have a nested function which has ExceptT in its type and you want to make it top-level, you can change its type to m (Either Patak Sepulka)

But returning `m (Either Patak Sepulka)` isn't recommended either, you should return `m Sepulka` and throw `Patak`. The reason is that you wouldn't want to deal with the possibility that `Patak` might be *both* thrown and returned in `Either`. Perhaps we can invent a rule that it's OK to return something as `Left` if and only if it doesn't have an `Exception` instance (preferably, it should have an instance like `instance TypeError "NOT AN EXC" => Exception Patak` to avoid accidents).

> It's mostly based on my personal feeling/intuition

I share your feelings, I see value in using `ExceptT` in this case too. But I'm having a hard time to come up with good, simple and precise rules, to detect these cases, so let's settle for a more uniform (but perhaps not as nice) solution.

docs/exceptions.md
int-index: I answered in the neighboring thread where you made the same comment.

docs/exceptions.md
volhovm: @int-index Ok, I support your point in fact, but I guess it should be explicitly stated in the document. It's not that straightforward and clear why do we ban using `Either` for returning non-exceptional failures.

int-index: I tried to keep the guidelines to the point, without justifications, so people can use them as a quick reference to make a decision. Good thing that this GH discussion will remain and can be linked to in case there are questions.

docs/exceptions.md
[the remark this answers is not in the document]
gromakovsky: It sounds quite sad. So AFAIU in some cases we have to use exceptions for control flow, which is generally considered anti-pattern, at least in some languages. Maybe it's not a big problem for Haskell, though. Or we can write deeply nested code, but it's also sad.
I am not saying that we shouldn't do it, I just don't like this world :/

Regarding your points:
> Not quite, you can't do something like this:

Well, there is a workaround to return `Either` from inside `bracket` and then have `ExceptT $ bracket …`. If it's too unpleasant, you can use exceptions instead. Permitting `ExceptT` in such cases doesn't mean that one must always use `ExceptT`. It's just another possible approach (which will sometimes be useful).

> I believe we can use it as our base monad, because the only transformer on top that we want is `ReaderT` and since adding `ReaderT` is equivalent to adding a parameter to a function, I believe it should be always possible.

As I wrote somewhere, having `PropertyM` in base monad will leave us without `MonadMask` instance, which is probably too bad. Anyway, perhaps we shouldn't continue discussing it here.

> But returning m (Either Patak Sepulka) isn't recommended either, you should return m Sepulka and throw Patak…

1. I think we can prohibit returning `Either` from potentially impure functions. We can always define an isomorphic ADT. E. g. we have `verifyBlocksPrefix :: OldestFirst NE Block -> m (Either VerifyBlocksException (OldestFirst NE Undo, PollModifier))` and we can change its return type to `m VerifyBlocksRes` where `data VerifyBlocksRes = VerifyBlocksSuccess (OldestFirst NE Undo, PollModifier) | VerifyBlocksInvalid VerifyBlocksException`. But maybe it would be too strict restriction. `Either` doesn't always carry something exceptional as its `Left`. For instance, we have `type Block = Either GenesisBlock  MainBlock`. This rule (the version without «it's OK to return something as Left if and only if it doesn't have» part) would mean we can't return `m Block`. Yes, we can and should change it do another ADT instead of using `Either`. But do we want to create an ADT for everything isomorphic to `Either`?
Also, let's consider `VerifyBlocksRes` from above. `VerifyBlocksException` might have an `Exception` instance. Should we care about possibility to `throwM` it? It's not `Either`, but isomorphic. What if `VerifyBlocksRes` has 10 constructors only one of which has an `Exception` instance?
Alternatively we can change `verifyBlocksPrefix`'s type to `OldestFirst NE Block -> m (OldestFirst NE Undo, PollModifier)`. My intuition suggests that we shouldn't do it, but maybe it's another case where I should ignore intuition.
2. I like this part:
> Perhaps we can invent a rule that it's OK to return something as Left if and only if it doesn't have an Exception instance (preferably, it should have an instance like instance TypeError "NOT AN EXC" => Exception Patak to avoid accidents).

3. Exceptions Best Practices suggests that returning `Either` is ok. Maybe not ideal but not something to prohibit. At least that's how I treat this phrase:

> Generally the solution to the ExceptT IO anti-pattern is to return an Either from more functions and throw an exception for uncommon errors.

docs/exceptions.md
```diff
@@ -0,0 +1,266 @@
...
+it would be possible to prevent them statically. There might be tricks that
+you're not aware of. Always make this trade-off conciously and responsibly.
+
+(On the other hand, if static guarantees require GADTs or type families, perhaps
+it's better to not overcomplicate code. Use your judgement.)
+
+DISCUSSION: Should we create a synonym `bug = impureThrow` in Universum? This
+would make the intention more clear.
+
+### Pure code, regular errors
```
avieth: Regular error means

> In cases when the erroneous scenarios are out of our control, we consider these to be regular errors.

but in a pure program, nothing is out of our control, so does "Pure code, regular errors" make any sense?

docs/exceptions.md
```diff
@@ -0,0 +1,266 @@
...
+instantiation does not inspect/use these `IO` actions, so it's not considered
+potentially impure.
+
+### Identifying code purity
+
+Is there an instantiation of type parameters that would mean that there are
+`IO` actions inspected/used in the code?
+
+* Yes: it is potentially impure code
+* No: it is 100% pure code
```
avieth: Not so sure about this one. `f :: forall m . Monad m => m ()` is pure, even though it can be specialized to `IO ()`. The specialization can never throw an exception / do other side-effects, because `forall m . Monad m => m ()` doesn't permit this.

docs/exceptions.md
```diff
@@ -0,0 +1,266 @@
...
+Do:
+
+* return `Either ErrorADT`, `Maybe`
+* wrap the underlying (pure!) monad in `ExceptT` or `CatchT`
+* use `MonadError` or `MonadThrow`
+
+Consider parsing: it is pure, but we cannot make assumptions about the input. In
+this case we might want to use `ExceptT ParseError`. Or consider a lookup in a
+`Map`, where we don't know whether the key is present -- in this case we'd like
+to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
```
avieth: A `Left ParseError` from parsing or a `Nothing` from a map lookup is not an error / exception, it's completely normal. If I get a parse error, nothing has gone wrong, the parser simply decided that the input was not in the language. When I get `Nothing` from a map lookup, nothing has gone wrong, the thing just wasn't in the map... completely normal.

docs/exceptions.md
```diff
@@ -0,0 +1,266 @@
...
+
+* use `error` or `impureThrow`
+* use `MonadFail`
+* return `Either Text`
+
+Do:
+
+* return `Either ErrorADT`, `Maybe`
+* wrap the underlying (pure!) monad in `ExceptT` or `CatchT`
+* use `MonadError` or `MonadThrow`
```
gromakovsky: If one uses `MonadError` or `MonadThrow`, code stops being pure (because `m` can be e. g. `ExceptT () IO` or just `IO` for `MonadThrow`).

docs/exceptions.md
```diff
@@ -0,0 +1,266 @@
...
+
+Do not import `Control.Exception` or `Control.Monad.Catch`! We use the
+`safe-exceptions` to deal with asynchronous exceptions gracefully, so import
+`Control.Exception.Safe`.
+
+## Resource handling
+
+Use `bracket` or to guarantee the release of resources. In case of
+concurrent code, avoid `forkIO` or `forkProcess` in favor of the `async`
+package, as it rethrows exceptions from the child threads.
```
avieth: Yes and prefer `withAsync` / `concurrently` / `race` wherever possible over `async` itself, because that one will *not* automatically kill the child if the parent dies. `withAsync` basically does a `bracket` so don't use it in a tail-recursive loop like `loop = waitForData >>= \it -> withAsync (process it) >> loop`, and be aware of `link` in case you don't plan to `wait` for the `Async`.

docs/exceptions.md
int-index: Well, it is classified as "potentially impure", because it can be instantiated to `IO`. This means we don't want to use `ExceptT` there, because it could do `IO`.

For instance, assume that we have `f :: forall m. Monad m => m a -> m a`. Suppose we're passing an `IO a`, and it throws. Inside `f` we could have used `ExceptT` that assumed that it was the only way to throw, but now it isn't.

docs/exceptions.md
int-index: That's why this is classified as "regular errors", everything is normal, we just have to handle these cases, which are not the main (desired) flow of events.

docs/exceptions.md
int-index: You can write code in a concrete pure monad but use methods from `MonadError` and `MonadThrow`, that's what is meant here. For instance, when you write code for `CatchT`, you use methods from `MonadCatch`.

docs/exceptions.md
avieth: Seems fine to use `ExceptT` in the definition of `f`. The resulting `forall m . Monad m => m a` cannot introduce any of its own throws, so the use of `ExceptT` is not really adding a new avenue of exception throwing to that term. If `f act :: IO a` throws an exception we are 100% sure it was thrown by `act`, and if `act` has only way exception throwing mechanism then so does `f act`.

docs/exceptions.md
avieth: Right my issue is with the term "error". They're not errors, they are part of the main desired flow of events.

docs/exceptions.md
int-index: Well, technically yes, we do control what our function take as input. But consider these three ways to give a type signature to `head`:

1. `[a] -> a`
2. `NonEmpty a -> a`
3. `[a] -> Maybe a

In case (1) we assume that the caller has full control of the passed argument, and passing `[]` is a programmer error. If the provenance of the input list external (i.e. we've read the list from a file), we (as a caller) need to check whether the list is empty. In case (2) we ensure this invariant on types. But in case (3) we give move the check into the function itself, rather than impose it on the caller. You could say that this is not an "error path", but informally you would normally think about this as a failure (we tried to extract the `head`, but there isn't one).

If we tried to formalize this, I would say in a pure function we can talk about an expected set of codepaths, and we also have checks that ensure that the input satisfies the necessary conditions for this codepath. When the checks fail, we want early from the function — this is what I mean by "regular errors in pure code".

I agree that it's not a strict definition, it depends on what you consider to be an error, but that's how we usually reason about code informally.

docs/exceptions.md
int-index: OK, perhaps this definition is overly restrictive. Could we come up with a more precise one?

docs/exceptions.md
int-index: What you consider an error depends on your expectations. When you `readFile` and the file does not exist, is it an error? No, we should've expected this. When you allocate a chunk of memory and the heap is full, is this an error? No, we are aware that the events can take this turn. If you think this way, nothing is an error.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: Does anyone else have an opinion on this? Should we add a rule about allowing local use of `ExceptT` in impure code? I do not have a strong opinion, but I'm hesitant to add more complications to the rules.

docs/exceptions.md
avieth: The lesson here is that `head :: [a] -> a` is a lie :) 

I revise my statement then: in a *total* pure program, error throwing / handling doesn't make sense. There can be programmer mistakes, sure, but the function is total so in these cases the programmer didn't realize there was a mistake.

int-index: Even in total functions, can't we talk about having a "major" codepath and various "error" codepaths? For instance, when we consider parsing, "parse errors" are called "errors" for a reason, no?

docs/exceptions.md
int-index: Yes, that's the way to use `async` that I had in mind. I added a clarification.

docs/exceptions.md
avieth: You're right it's confusing terminology. It's a parse error, but a *parse error* is not an *error in parsing*. Parsing went just fine: we successfully determined that the input string is not in the language. A better terminology would perhaps be `parse :: Parser a -> Either Unrecognized a`.

docs/exceptions.md
avieth: > When you readFile and the file does not exist, is it an error? ... When you allocate a chunk of memory and the heap is full, is this an error?

Yes and yes. Strictly speaking I'd call them exceptions but let's got hung up on semantics of error vs. exception.

> What you consider an error depends on your expectations

You're absolutely right. When I parse something I expect that the input string may not be recognized. When I lookup a key in a map I expect that it may not be in the map. Nothing went wrong in these cases. But with `readFile`, and with `IO` in general, things may go wrong for reasons outside of the programmer's control (reasons external to the program itself), and there's no obvious one-size-fits-all response to it, so we resort to exception handling.

docs/exceptions.md
avieth: I tend to use "error" to mean a programmer mistake i.e. a bug. Something that could be avoided but was not, because we're human. So regarding `head :: [a] -> a`, yes, the error would be the caller's error, because he was just supposed to know that the input must not be `[]`.

I tend to use "exception" to mean something that the programmer could not possibly have defended against / ruled out and is only capable of handling / responding to. There's no exception in `head []`, it's a plain error. But reading from a TCP socket that has been reset is an exception. The programmer did nothing wrong, the connection just broke. Failing to handle an exception could be considered an error.

I guess you could also speak of "user error" but that's very different. It doesn't mean any errors in our program. A parse error could be a user error (user gave a malformed config file or something). It can be considered an error to fail to deal with a user error.

docs/exceptions.md
```diff
@@ -0,0 +1,272 @@
...
+Do *not*:
+
+* do this often, as if it's okay
+* use `undefined`
+* use non-exhaustive pattern matching
+
+Do:
+
+* try to use types to avoid the need in the first place
+* comment extensively (invariants and precondition, reasoning)
```
ChShersh: Well, it might be even better to use `LiquidHaskell` for this. I think this is exactly the use case for `LiquidHaskell` — verify some invariants on compilation state.

docs/exceptions.md
```diff
@@ -0,0 +1,272 @@
...
+
+Before writing code that allows programmer mistakes, consult with colleagues how
+it would be possible to prevent them statically. There might be tricks that
+you're not aware of. Always make this trade-off conciously and responsibly.
+
+(On the other hand, if static guarantees require GADTs or type families, perhaps
+it's better to not overcomplicate code. Use your judgement.)
+
+DISCUSSION: Should we create a synonym `bug = impureThrow` in Universum? This
+would make the intention more clear.
```
ChShersh: I don't mind adding `bug = impureThrow` to `universum`. With `HasCallStack =>` of course.
You're welcome to open issue here: https://github.com/serokell/universum/issues

docs/exceptions.md
```diff
@@ -0,0 +1,272 @@
...
+
+* create a custom exception type
+* use `throwM` (`MonadThrow`)
+
+We disallow the use of `throwIO` only because it is redundant in the presence of
+`throwM` and requires a stronger constraint (`MonadIO` rather than
+`MonadThrow`).
+
+Derive prisms for exception types with multiple constructors, so it's convenient
+to use them with `catchJust`.
```
ChShersh: Is there any particular reason to use _prisms_ exactly instead of _affine traversals_? I understand that word _prism_ is much more clear for people. But usage of `Prism` type class enforces to use `lens` library instead of some cheaper alternative, like `microlens-platform`. As for me. even for such big project like `cardano-sl` package `microlens-platform` should be enough. Though, I guess there're a lot of dependencies which use `lens` package by themselves so we can't optimize dependencies here much...

docs/exceptions.md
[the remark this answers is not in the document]
ChShersh: I like the rule to use `ExceptT` only with `TypeError` exception instances for types. I'm not aware of better solution for _early exit_ from function (I'm aware only of `Cont`-monad solution but I can't call this solution a better one ;) ).

docs/exceptions.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Resource handling
+
+Use `bracket` or to guarantee the release of resources. In case of concurrent
+code, avoid `forkIO` or `forkProcess` in favor of the `async` package, as it
+rethrows exceptions from the child threads. (Do not use the function `async`
+itself when you can use `withAsync`, `race`, or `concurrently`).
+
+When resource usage is non-linear, it's okay to use `ResourceT`, but prefer
+`bracket` whenever possible.
```
ChShersh: Could you please clarify what is non-linear resource? It's not a common terminology... I'm aware of Linear Types and I can imagine which resources can be called _linear_ but still to make sure that terminology is the same for everybody.

gromakovsky: @int-index please clarify.

docs/exceptions.md
```diff
@@ -0,0 +1,281 @@
...
+they add additional exception mechanisms to the one that `IO` has, and
+`catch`/`bracket` don't account for this).
+
+### Impure code, regular errors
+
+Do *not*:
+
+* use `error` or `impureThrow`
+* use `ExceptT`, `MaybeT`, or `CatchT`
+* use `Maybe` or `Either`
```
gromakovsky: 1. What's wrong with `Maybe`?
2. AFAIU, you mean that `Either` shouldn't be used to represent regular errors, but it's fine to return `Either SomethingThatIsNotAnError SomethingElseThatIsNotAnError `, is it correct?

gromakovsky: Example: is it ok to return `Block`? Currently `Block` is `Either GenesisBlock MainBlock`. Of course it can be changed to a new type which is not `Either`, but suppose we don't want to do it for some reason. For example, because it's not trivial and will take some time, but one may want to return `Block` from a function right now.

gromakovsky:
So, if my understanding is correct, there are the following concerns which should be addressed before adopting the guidelines from this PR:
1. My new [comment](https://github.com/input-output-hk/cardano-sl/pull/2059/files#r157735356). Specifically, I'd like to know what's wrong with `Maybe` in impure code and some clarifications about `Either` in impure code.
2. There is a [discussion](https://github.com/input-output-hk/cardano-sl/pull/2059/files#r155497759) about using `ExceptT` in impure code in some circumstances. Specifically, we should consider permitting `ExceptT` in impure code if it's an implementation detail of a function and the first type parameter of `ExceptT` doesn't have an `Exception` instance (ideally has instance `TypeError "NOT AN EXC" => Exception Patak`).
3. @avieth has raised two concerns in two threads: https://github.com/input-output-hk/cardano-sl/pull/2059/files#r156155221 and https://github.com/input-output-hk/cardano-sl/pull/2059/files#r156127346. AFAIU, both of them are only about terminology. Please let me know if they are not only about terminology.
4. Another concern was raised by @avieth [here](https://github.com/input-output-hk/cardano-sl/pull/2059/files#r156154705). It's proposed to permit using `ExceptT` inside `f :: Monad m => m () -> m ()`.
5. @int-index should clarify the meaning of [non-linear resource usage](https://github.com/input-output-hk/cardano-sl/pull/2059/files#r157647513).
6. I pushed two small commits recently, someone should review them.

gromakovsky:
Here is what I propose to do:
0. First of all, I want to notice that we are not writing a static analyzer or something like that, so we don't need to define a strict formal set of rules which everyone must always follow. It's ok to say that some code is not recommended but is sometimes acceptable if there is a good reason to do so. Even if we prohibit some construction, during code review we may notice that it makes perfect sense in some particular case and may want to use it despite being prohibited.
Of course we should try to decrease ambiguity as much as possible, but there is no goal to describe everything formally and completely unambiguously. These are guidelines, not strict rules.
1. Permit returning `Maybe` from a potentially impure or impure function. Permit returning `Either A B`, but only if `A` doesn't have an `Exception` instance. I like the `TypeError` trick, I think we should make it recommended, but not strictly mandatory, because writing `Exception` instance for types like `GenesisBlock` is insane (at least to me) and hopefully nobody will do it anyway. But if we want to return `Either ToilVerFailure a`, we should prohibit `Exception ToilVerFailure`.
But let's wait for @int-index's reply first.
2. Me, @int-index and @shersh commented on this concern.
@shersh only says that he likes such usages of `ExceptT` when `instance TypeError => Exception e` exists. There no arguments why we should or shouldn't do it.
@int-index has a feeling that such usages of `ExceptT` are good, but has reasons not to use it and suggests that we prohibit it.
I agree with reasons not to use `ExceptT` and I don't have convincing reasons to use it (and never had, it's only my intuition). So by default I propose to leave it as is (i. e. prohibit `ExceptT` in potentially impure code completely).
3. I will read terminology concerns again more carefully and will try to improve some words. I don't have anything concrete yet.
4. Both me and @int-index think that `ExceptT` shouldn't be used in `f :: Monad m => m a -> m a`. @int-index asked @avieth if he has a better idea (unfortunately we don't).
By default we'll just leave it as is and prohibit `ExceptT` there.
5, 6. Just waiting for @int-index basically.

shersh:
@gromakovsky I think I'm not who you want to see here =)

docs/exceptions.md
int-index: Sure, we could experiment with `LiquidHaskell` in the future. Right now it does not seem mature enough for a project on the scale of Cardano, but I'd like to be proven wrong.

docs/exceptions.md
int-index: Ok, created https://github.com/serokell/universum/issues/100

docs/exceptions.md
int-index: Yes, we want `Prism` to have both construction and matching.

docs/exceptions.md
int-index: Non-linear is anything that doesn't fit into the "allocate, use, deallocate" pattern.

docs/exceptions.md
int-index: The idea was to have uniform throwing principle for potentially impure code (i.e. always throw an exception with `throwM`). Now that we plan to relax this rule to allow `Either NonExc NonExc`, then allowing `Maybe` makes sense as well.

int-index:
@gromakovsky I agree with your plan, will you incorporate the changes into the document? I answered the questions about non-linear resources and `Maybe` in impure code in the corresponding threads.

gromakovsky:
> I agree with your plan, will you incorporate the changes into the document? I answered the questions about non-linear resources and Maybe in impure code in the corresponding threads.

Ok, will do it after I get quite confident that everyone is fine with that.
Do you agree with changes from my commits?

int-index:
> Do you agree with changes from my commits?

Yes, I reviewed them back when you pushed.

gromakovsky:
TODO: write that `SomeException` should be used in pure code only if you are lazy to define a proper ADT with all possible exceptions (because in pure code you know which exceptions can be thrown).

gromakovsky:
@avieth do I understand correctly that your concerns in https://github.com/input-output-hk/cardano-sl/pull/2059/files#r156155221 and https://github.com/input-output-hk/cardano-sl/pull/2059/files#r156127346 are only about terminology? Do you agree that using `Maybe`, `Either`, ExceptT`, `Maybe` and `CatchT` to work with e. g. parse errors is reasonable?

If it's only about terminology, can you propose better terminology?

PR #2060, opened 2017-12-07
[CSL-1972] Do a single DB and mempool read when computing wallet balance

Martoon-00:
Rebased #1994

### DIFF
--- a/node/src/Pos/Client/Txp/Balances.hs
+++ b/node/src/Pos/Client/Txp/Balances.hs
@@ -1,0 +1,97 @@
+{-# LANGUAGE TypeFamilies #-}
+
+module Pos.Client.Txp.Balances
+       ( MonadBalances(..)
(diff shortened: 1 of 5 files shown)


node/src/Pos/Client/Txp/Balances.hs
```diff
@@ -0,0 +1,97 @@
+{-# LANGUAGE TypeFamilies #-}
+
+module Pos.Client.Txp.Balances
```
gromakovsky: What? Why did you add this module here? Probably by mistake.

georgeee:
@Martoon-00 @gromakovsky @adinapoli-iohk shall we close it?

adinapoli-iohk:
@georgeee I think so.

@nc6 's [current work](https://github.com/input-output-hk/cardano-sl/pull/2394) will address this and many other concerns.

PR #2067, opened 2017-12-07
[CSL-1289] Merge timewarp into csl repo

volhovm:
`time-warp-nt` is here! Welcome!
* `./networking/` contains `cardano-sl-networking` which contains all changes up to `0f58f145425650f5e68b838668f36c8342e4a0a8`
* Stylish-haskell, weeder were applied. I couldn't apply hlint because tw didn't use universum, so lots of hints are useless. It's better to do it, but maybe not in this particular PR.
* Versions were unified (`cardano-sl-networking` now has version 1.0.3).


avieth:
> time-warp-nt is here! Welcome!

Hooray!

So you applied all of the commits from time-warp-nt? I expected a simple copy/paste and commit.

volhovm:
@avieth Didn't want to spoil the statistics. We'll maybe need to delete the old time-warp repo.

networking/.gitignore
```diff
@@ -0,0 +1,28 @@
+dist
```
gromakovsky: Perhaps having a `.gitignore` not in the root is a bad idea?

networking/.travis.yml
```diff
@@ -0,0 +1,26 @@
+sudo: false
```
gromakovsky: This file is probably 100% useless.

networking/stack.yaml
```diff
@@ -0,0 +1,42 @@
+resolver: lts-9.1
+
```
gromakovsky: I'd like to avoid having a separate `stack.yaml` for `networking/`.

PR #2068, opened 2017-12-08
[CSL-1939] Check that swagger is spec is valid and up to date

Martoon-00:
Rebased #1990


Martoon-00:
This seems to work. I had to regenerate `swagger.json` in order for CI to pass, and still, as we know, generated documentation is invalid so CI fails.
Let's wait until fix appears in `feature/new-wallet` and then I rebase.
(^ is regarding travis build, appveyor fails due to doubled `servant-generics` dependency, fix is in master already)

scripts/ci/validate-wallet-swagger.sh
```diff
@@ -0,0 +1,21 @@
...
+set -e
+set -o pipefail
+
+if [[ $TRAVIS_COMMIT == "" ]]; then
+    echo "TRAVIS_COMMIT is unspecified"
+    exit 1
+fi
+
+spec_path="https://raw.githubusercontent.com/input-output-hk/cardano-sl/$TRAVIS_COMMIT/wallet-new/spec/swagger.json"
+validate="curl http://online.swagger.io/validator/debug?url=$spec_path"
```
adinapoli-iohk: @Martoon-00 Idea: As this is an external service, and as such it might not work 100% of the time, should be capture the possibility curl fails either with exit code `1` or with a non-200 error but still allow CI to be completed in green? I suspect other folks would be extremely angry with us if we break the entire CI 😁 

Another way to see this, is that we should either:

- Add another job like `hlint` in the build matrix that can still fail separately;
- Make this check step optional;

What do you think? (I might even prefer 1., but I would be curious to hear what you think)!

Martoon-00: @adinapoli-iohk Well, you can see the first approach implemented here
I'm afraid that in case of that swagger job failing CI fill be just green, and no one will notice that swagger is invalid unless he opens job matrix. If so, this validation looks pretty useless.
Do you know something about it? I've spent whole evening trying to get only swagger job failing to see how it would look like, but failed.

In this sense approach #0 (remain validate mandatory, fail only on 200 OK responses with error report) may be better, however I'm afraid of breaing CI if that swagger validator one day goes mad

adinapoli-iohk: > I'm afraid that in case of that swagger job failing CI fill be just green, and no one will notice that swagger is invalid unless he opens job matrix. If so, this validation looks pretty useless.

I see what you mean, but I think this is for now the safest bet for not hurting other developers productivity. Let me see if I can kindly ask devops to review the work and merge it in a timely fashion 😉

adinapoli-iohk:
Excellent job @Martoon-00 , I'm very happy with the end result!

I just need now to summon a bit of help from our _friendly neighborhood devops_ @cleverca22 and @deepfire in order to add the icing on the cake. Serge, Micheal, let me give you the TL;DR of this patch:

> Konstantin extended Travis with an optional job (for which failure is allowed) which will check the validity of the `wallet-new` Swagger spec.

We are trying to make this process as non-invasive as possible for other developers, but we were wondering if there is a way to alert the developers _without blocking the PR from being merged_ that the `swagger-check` job failed. To put it differently, in the current shape, failures of `swagger-check` won't result in CI to fail with a red cross. This is good as we don't want to create trouble as `wallet-new` is a fast-paced prototype, but the icing on the cake would be to _have Travis display a red cross in the status report without blocking merges_.

I suspect this is not possible if not by creating a _separate_ check, on top of the 4 we have now? In that case, we might just as well go with what Konstantin coded so far, which is already useful per se, and in the future (once `wallet-new` stabilises) make `swagger-check` mandatory. What do you think, guys?

Martoon-00:
@adinapoli-iohk How do you think, does anything yet to be done remains here?

adinapoli-iohk:
@Martoon-00 Please be patient 😉  I can see how CI is now failing -- in particular the `normal` linux job is failing with:

```
Dec 12 15:10:47 Swagger spec wallet-new/spec/swagger.json is outdated! Current md5 is 656b1b72fc34113e177c397ff2f7032c, actual md5 is 0529776ce5c8aa4e7b0dd340f619fc58
```

Are we sure we made that optional, after all? 😀

Martoon-00:
Hm, but it's because you look at `pr`, in `push` travis `normal` job is ok but build still fails )

cleverca22:
https://gist.github.com/anonymous/fa1357a8c66ccb0dde985a4d47a9fe38 needs to be applied to add cpphs to `wallet-net`

adinapoli-iohk:
@Martoon-00 looks like we finally have Hydra's webhooks, yay!

I would suggest to finally remove the swagger check from Travis and simply piggyback on the work @cleverca22 did for Hydra (if you scroll down a bit you will see how we are already checking that, it's the hook called `swagger-check`) 😀

Martoon-00:
Great! I removed those travis job and validation script.
Although we now check not spec validity as Swagger, but its validity just as JSON, I agree that this should be enough for now

adinapoli-iohk:
> Although we now check not spec validity as Swagger, but its validity just as JSON, I agree that this should be enough for now

@Martoon-00 good point! I will see if I find the time today to extend @cleverca22 's scripts to also run `md5` validation, like your online script did. I feel I'm dragging this PR forever and I don't want you to feel too frustrated by it 😁

domenkozar:
FYI: now that we're using buildkite we can just upload swagger spec for each build as an artifact and have no need to really keep it as part of the github repo, creating maintenance overhead

Martoon-00:
According to @adinapoli-iohk we need this back

adinapoli-iohk:
@Martoon-00 Yes, we need this back but it's a bit more complicated than simply reopening the PR. In order:

- The base branch should be `develop` and not `feature/new-wallet`;
- We want to do the checks on BuildKite or Hydra, not Travis;
- We want to do the checks by generating the API on the fly (you might want to coordinate with the work @KtorZ is doing on the documentation to pinpoint which files we do want  to generate & check)

My gut feeling is that we will end up with at least two hydra/Buildkite jobs: one for V0 & one for V1, and possibly a 3rd one which checks the full API (the one inclusive of development endpoints). I will leave you and Matthias to coordinate 😉

KtorZ:
Ideally, I'd like to reach a point where the swagger specs are served by the wallet backend (`/api/swagger.json` for V0 and `/api/v1/swagger.json` for V1).

I am not sure exactly then how ops want to have a hand on those files. @adinapoli-iohk, you mentioned that there used to be a dedicated executable to generate the specs  files (instead of doing it on each start-up of the backend). This sounds like a good solution to have a simple way to generate swagger files when needed and after the build has succeeded. Otherwise, GET requests on the related endpoints will do, provided the wallet backend is up-and-running somewhere...

adinapoli-iohk:
> Ideally, I'd like to reach a point where the swagger specs are served by the wallet backend (/api/swagger.json for V0 and /api/v1/swagger.json for V1).

If we want (and have time) we surely could, I think this was something that the old wallet also exposed.

> I am not sure exactly then how ops want to have a hand on those files. @adinapoli-iohk, you mentioned that there used to be a dedicated executable to generate the specs files (instead of doing it on each start-up of the backend). This sounds like a good solution to have a simple way to generate swagger files when needed and after the build has succeeded.

Yep, my idea was to have a dedicated executable (which I think also ship as part of this PR but it's a bit of a "catch-all" hammer which generates the full API) we could run to generate the spec on the fly and validate it. In terms of where we will store the artifacts eventually, I think @domenkozar was quite keen to have something like the following:

- We generate the `.json` files as part of BuildKite artifacts;
- These artifacts are private to the CI server but they are hosted on S3, and thus exposed;
- We generate machinery on `cardanodocs.com` to serve these `.json` files via a deterministic URL scheme, which could, for example, take in account the commit. Example:

```
https://cardanodocs.com/technical/api/wallet/e029c945b290336e76e447dc67e582041df04d10/v0/swagger.json
```

(Or similar). Feel free to propose alternative solutions, ofc 😉

adinapoli-iohk:
@KtorZ Just wanted also to put this one on your radar. @Martoon-00 did a great job back in the days to create some CI scripts to validate the `swagger.json` automatically, but I think we should now incorporate this PR into the grand-schema of things: even if we plan to stop committing the spec into the repo, I think there is lots of value of adding an Hydra check to generate on the fly the spec and check for its validity, and prevent the PR from being merged in case the task fails. what do you think?

KtorZ:
Yes. A failure in the Swagger specs should definitely prevent a PR from being merged. I'll have a look at this PR ASAP, it seems that there's a lovely number of conflicts already :) ..... :'(

adinapoli-iohk:
@KtorZ my 2 cents: it might be easier to simply cherry-pick backport the interesting bits and close this one off, as this PR accumulated lots of dust over the course of the months 😁

KtorZ:
I definitely agree. The backport is therefore still relevant?

adinapoli-iohk:
> I definitely agree. The backport is therefore still relevant?

I think the relevant part is the work @cleverca22 did to add the Hydra job which actually check the validity 😉

PR #2070, opened 2017-12-08
[CSL-2034] Introduce the "Example" typeclass to improve Wallet's Swagger specs

akegalj:
Issue: https://iohk.myjetbrains.com/youtrack/issue/CSL-2034
* removed empty examples in documentation
* removed some null examples in documentation
* fixed () -> NoContent for single endpoint
* response example shows success type (instead of error before)
* `WalletResponse [a]` will always generate documentation (before it could return empty list response) 
* removed `fromString` in arbitrary instances, and few more Arbitrary instances modifications


wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs
```diff
@@ -0,0 +1,58 @@
...
+import           Cardano.Wallet.API.Response
+import           Cardano.Wallet.API.V1.Types
+
+import           Test.QuickCheck (Arbitrary (..), Gen, listOf1)
+
+class Arbitrary a => Example a where
+    example :: Gen a
+    example = arbitrary
+
+instance {-# OVERLAPS #-} Arbitrary a => Example a
```
adinapoli-iohk: This looks a bit strange, why we need such empty, overlappable instance?

wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs
```diff
@@ -0,0 +1,58 @@
...
+                       <*> example
+                       <*> example
+
+instance Example Payment where
+    example = Payment <$> example
+                      <*> example
+                      <*> example
+                      <*> example -- TODO: will produce `Just groupingPolicy`
+
+-- IMPORTANT: if executing `grep "[]\|null" wallet-new/spec/swagger.json` returns any ellement - then we have to add Example instances for those objects because we don't want to see [] or null examples in our docs.
```
adinapoli-iohk: Typo `ellement` 😉

wallet/src/Pos/Wallet/Web/Api.hs
```diff
@@ -319,7 +319,7 @@ type ResetFailedPtxs =
         "For all transactions in CPtxWontApply condition, \
         \reset them to CPtxApplying condition so that they will \
         \be passed to resubmition"
-    :> WRes Get ()
+    :> WRes Get NoContent
```
adinapoli-iohk: In theory this should be already in `master`, but as long as it doesn't create conflict, I don't mind which variant we need 😉

wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs
akegalj: With it we say "Everything that has Arbitrary instance also has Example instance". So it says that every Arbitrary is also Example. If we remove it we have to explicitly define all instances... for example:
```
/home/akegalj/projects/serokell/cardano-sl/wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs:30:27: error:
    • No instance for (Example
                         cardano-sl-core-1.0.3:Pos.Core.Common.Types.Coin)
```

and dozens more (<20). So we would have to define:
```
instance Example Coin where
    example = arbitrary
```
explicitly and with `{-# OVERLAPS #-}` this is implicit. 

So I believe it is a trade off between the above boilerplate (I don't mind writing it though).

Do you think I should remove it and write instances manually ?

wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs
akegalj: Will fix, ty.

wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs
akegalj: In the meantime I have changed `OVERLAPS` with `OVERLAPPABLE` as it seems safer https://ghc.haskell.org/trac/ghc/ticket/9242#comment:30 . 

I will implement it without pragmas by explicitly writing instances (which is not hard :) ) if we decide to go this way (ie, it is 5 mins of work).

akegalj: Let me write Pros and Cons of both approaches on Monday and then we will make a final decision.

ksaric: I found this interesting approach - https://kseo.github.io/posts/2017-02-05-avoid-overlapping-instances-with-closed-type-families.html

akegalj: @ksaric Interesting. Let me read it and try playing with it on Monday. Ty

ksaric: I got interested in this issue, in the end, I ended up writing up the code, so I added a commit. You can remove it, I'm pretty sure `OVERLAPPABLE` is going to be clearer.

adinapoli-iohk: I'm aware of that blog post and I'm sure we can go via that route, although I'm just skeptical we need to bring cannons on the table (e.g. type families) for such a simple scenario, which can be solved by literally 30 seconds of vim macros to write a bunch of one-lines like this:

```
instance Example Foo
instance Example Bar
```

This will work without defining the body of the instance due to the default implementation. Boilerplat-y? Yes. Haskell-98? Yes 😁 

I still think using closed type families for avoid overlappable instances is a good trick to remember and keep in our arsenal, but I don't think the problem at hand requires them 😉

akegalj: Fixed, I believe we can merge after CI

akegalj:
@adinapoli-iohk the same issue as in https://github.com/input-output-hk/cardano-sl/pull/2049#issuecomment-351065853

Travis/push didn't succeed as it was building wrong branch (don't know why) so it has been restarted for the CI.

This is blocking https://iohk.myjetbrains.com/youtrack/issue/CSL-2051 and https://iohk.myjetbrains.com/youtrack/issue/CSL-2048

akegalj:
yeey, this can be merged @adinapoli-iohk

adinapoli-iohk:
@akegalj Very happy to merge, but this seems like a genuine failure from AppVeyor:

```
    C:\w\wallet-new\server\Cardano\Wallet\API\V1\Swagger.hs:433:7: error:
        * No instance for (Example AddressValidity)
            arising from a use of `toSwagger'
        * In the first argument of `(&)', namely `toSwagger walletAPI'
          In the first argument of `(&)', namely
            `toSwagger walletAPI & info . title .~ "Cardano Wallet API"'
          In the first argument of `(&)', namely
            `toSwagger walletAPI & info . title .~ "Cardano Wallet API"
             & info . version .~ "2.0"'
Sorry, we tried running command for 5 times and all attempts were unsuccessful!
```

akegalj:
@adinapoli-iohk 
hm, odd again. I am going to try to reproduce it once again :(

akegalj:
@adinapoli-iohk 

Sorry, seems like when CI tried to merge this branch into base feature branch it got build conflicts. 

This is the first time I got branch without git conflicts but when merged build fails. This is what confused me. Sorry for annoyance. Fixing it

wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs
```diff
@@ -11,45 +18,64 @@ class Arbitrary a => Example a where
     example :: Gen a
     example = arbitrary
 
-instance {-# OVERLAPPABLE #-} Arbitrary a => Example a
+-- | Any @Example@ instance that we want to ovveride must be matched on
+-- the type level in this definition here.
+type family (F a) :: Bool where
```
akegalj: Really good trick https://kseo.github.io/posts/2017-02-05-avoid-overlapping-instances-with-closed-type-families.html in the article. Ty for pointing it out

adinapoli-iohk:
CI is now passing, with the exception of one job which is erroring out due to insufficient memory. This can merged now.

PR #2072, opened 2017-12-08
[Cosmetic] Fix haddock compilation errors.

jcmincke:
No specific issue. 
git haddock failed due to incorrect haddock syntax

### DIFF
--- a/binary/Pos/Binary/Class/TH.hs
+++ b/binary/Pos/Binary/Class/TH.hs
@@ -85,10 +85,10 @@ data Cons = Cons
 
 data Field
     = Field {
-    -- ^ The constructor means that you want
-    -- a field to participate in serialisation/deserialization
+    -- | The constructor means that you want
+    --  a field to participate in serialisation/deserialization
       fFieldAndType :: ExpQ
-    -- ^ You're expected to write something like @[|foo :: Bar|]@ here
+    -- You're expected to write something like @[|foo :: Bar|]@ here
     }
 
 -- | Turn something like @[|foo :: Bar|]@ into @(foo, Bar)@.
--- a/explorer/src/Pos/Explorer/ExplorerMode.hs
+++ b/explorer/src/Pos/Explorer/ExplorerMode.hs
@@ -66,21 +66,21 @@ import           Test.Pos.Block.Logic.Mode (TestParams (..))
 -- testing (and running).
 type ExplorerMode ctx m =
     ( MonadDBRead m
-    -- ^ Database operations
+    --  Database operations
     , MonadSlots ctx m
-    -- ^ Slotting
+    --  Slotting
     , MonadThrow m
     , MonadCatch m
     , MonadMask m
-    -- ^ General utility operations
+    --  General utility operations
     , HasExplorerCSLInterface m
-    -- ^ For mocking external functions
+    --  For mocking external functions
     , HasGenesisRedeemAddressInfo m
-    -- ^ Genesis operations
+    --  Genesis operations
     , MonadTxpMem (MempoolExt m) ctx m
-    -- ^ Txp, could be @TxpLocalWorkMode@
+    --  Txp, could be @TxpLocalWorkMode@
     , MinWorkMode m
-    -- ^ The rest of the constraints - logger, mockable, configurations
+    --  The rest of the constraints - logger, mockable, configurations
     )
 
 ----------------------------------------------------------------------------
--- a/explorer/src/Pos/Explorer/Socket/Methods.hs
+++ b/explorer/src/Pos/Explorer/Socket/Methods.hs
@@ -344,11 +344,11 @@ notifyEpochsLastPageSubscribers
     => EpochIndex -> ExplorerSockets m ()
 notifyEpochsLastPageSubscribers currentEpoch = do
     recipients <- view $ csEpochsLastPageSubscribers
-    -- ^ subscriber
+    --  ubscriber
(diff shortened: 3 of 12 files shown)


binary/Pos/Binary/Class/TH.hs
```diff
@@ -85,10 +85,10 @@ data Cons = Cons
 
 data Field
     = Field {
-    -- ^ The constructor means that you want
-    -- a field to participate in serialisation/deserialization
+    -- | The constructor means that you want
```
akegalj: Do you know will this produce documentation describing the thing above (`Field` constructor) or field bellow (`fFieldAndType`) ?

I see here https://hackage.haskell.org/package/QuickCheck-2.10.1/docs/src/Test-QuickCheck-Test.html#Args `quickcheck` is also using `-- ^` syntax. So I am not sure is this modification correct?

adinapoli-iohk: @akegalj The two should be equivalent, according to the documentation, at least: http://haskell-haddock.readthedocs.io/en/latest/markup.html#constructors-and-record-fields

@jcmincke so basically the `-- |` syntax worked but `-- ^` resulted in an error? How weird!

explorer/src/Pos/Explorer/ExplorerMode.hs
```diff
@@ -66,21 +66,21 @@ import           Test.Pos.Block.Logic.Mode (TestParams (..))
 -- testing (and running).
 type ExplorerMode ctx m =
     ( MonadDBRead m
-    -- ^ Database operations
+    --  Database operations
     , MonadSlots ctx m
-    -- ^ Slotting
+    --  Slotting
```
adinapoli-iohk: @jcmincke What about using `*` to expose these comments in the documentation? http://haskell-haddock.readthedocs.io/en/latest/markup.html#documentation-structure-examples

explorer/src/Pos/Explorer/Socket/Methods.hs
```diff
@@ -344,11 +344,11 @@ notifyEpochsLastPageSubscribers
     => EpochIndex -> ExplorerSockets m ()
 notifyEpochsLastPageSubscribers currentEpoch = do
     recipients <- view $ csEpochsLastPageSubscribers
-    -- ^ subscriber
+    --  ubscriber
```
adinapoli-iohk: little typo 😉

explorer/src/Pos/Explorer/Txp/Global.hs
```diff
@@ -72,7 +72,7 @@ applyBlund txpBlund = do
             Left genesisBlock -> SlotId
                                   { siEpoch = genesisBlock ^. epochIndexL
                                   , siSlot  = minBound
-                                  -- ^ Genesis block doesn't have a slot, set to minBound
+                                  -- Genesis block doesn't have a slot, set to minBound
```
adinapoli-iohk: This is the correct way to go, bravo. This is a comment the programmer left to describe a subtleties of the field, we don't want it to be exposed in the docs.

explorer/src/Pos/Explorer/Web/Server.hs
```diff
@@ -827,8 +827,8 @@ cAddrToAddr cAddr@(CAddress rawAddrText) =
     in case mDecodedBase64 of
         Just addr -> do
             -- the decoded address can be both the RSCoin address and the Cardano address.
-            -- * RSCoin address == 32 bytes
-            -- * Cardano address >= 34 bytes
+            -- 1. RSCoin address == 32 bytes
```
adinapoli-iohk: According to [this section](http://haskell-haddock.readthedocs.io/en/latest/markup.html#itemized-and-enumerated-lists) the removed syntax was compatible -- do you see any immediate advantage into switching to the proposed one?

binary/Pos/Binary/Class/TH.hs
jcmincke: As I understand it:

-- | : is used above the field it documents.
-- ^ : directly follows (on the same line? , to check) the field it documents (and is followed by a comma, if needed)

adinapoli-iohk: Ah, you are probably right. For some reason I prefer the latter, some guys the former. But I appreciate this are pub topics rather than real issues 🍺 😅  (i.e. I'm completely neutral to the use of `--|` here 😉  )

explorer/src/Pos/Explorer/ExplorerMode.hs
jcmincke: Inside tuple syntax -- | , -- ^ or -- * are not legal. -- *  is ok for "top level" documentation of types/functions/...

explorer/src/Pos/Explorer/Web/Server.hs
jcmincke: -- * : raises a haddock error.  -- * is fine for top level documentation, not for comment inside the code.

adinapoli-iohk: Oh, TIL I guess, thanks!

explorer/src/Pos/Explorer/ExplorerMode.hs
adinapoli-iohk: Bad me! I genuinely thought I was staring at a top level module 😅

gromakovsky:
#2410 has been merged.

PR #2076, opened 2017-12-09
[CSL-1918] Pass launcher args via config file

Pastafarianist:
This is WIP.

- [x] Discuss whether any other modifications are necessary now that the CLI of `cardano-launcher` has changed.
- [x] Make sure that we still want the implicit propagation of parameters in `main` (L189-193 and 233-261).
- [x] Discuss whether the naming of parameters in the YAML file with `defaultOptions` is OK.


tools/src/launcher/launcher-config.yaml
```diff
@@ -0,0 +1,65 @@
...
+nodeLogConfig: scripts/log-templates/log-config.yaml
+
+# File where node stdout/err will be redirected (def: temp file)
+# nodeLogPath: node.log
+
+# Path to the wallet frontend executable (e. g. Daedalus)
+# walletPath: binaries_v000/daedalus
+
+# Arguments to be passed to the wallet frontend executable.
+walletArgs: []
```
gromakovsky: I think it'd be good to make this list non empty (I mean, add at least one entry to this example), it would be more illustrative (the purpose of example is to demonstrate as much as possible).

tools/src/launcher/launcher-config.yaml
```diff
@@ -0,0 +1,65 @@
...
+# reportServer:
+
+# Where to put launcher logs (def: console only).
+# launcherLogsPrefix:
+
+configuration:
+  filePath: lib/configuration.yaml
+  key: default
+  systemStart: 1512847931
+  seed:
```
gromakovsky: Is it valid syntax at all? 🤔

tools/src/launcher/Main.hs
```diff
@@ -91,90 +94,44 @@ data LauncherOptions = LO
...
 getLauncherOptions :: IO LauncherOptions
-getLauncherOptions = execParser programInfo
+getLauncherOptions = do
+    configPath <- unConfigPath <$> execParser programInfo
+    decoded <- Y.decodeFileEither configPath
+    case decoded of
+        Left err -> do
+            writeFile "config-error.log" $
+                sformat ("Failed to parse "%build%": "%shown) configPath err
+            throwIO $ LauncherConfigParseError configPath err
```
gromakovsky: According to the new, not yet approved, [guidelines](https://github.com/input-output-hk/cardano-sl/blob/edcbb8d3bc2659692e8145710109a1dd43f99164/docs/exceptions.md#impure-code-regular-errors) (if the link becomes obsolete, see #2059), you should use `throwM` rather than `throwIO`.

For now it doesn't matter because those guidelines are not approved, but if we force everybody to follow them before your PR is merged, this line can lead to rejection of this PR :)

tools/src/launcher/Main.hs
```diff
@@ -91,90 +94,44 @@ data LauncherOptions = LO
...
 
 getLauncherOptions :: IO LauncherOptions
-getLauncherOptions = execParser programInfo
+getLauncherOptions = do
+    configPath <- unConfigPath <$> execParser programInfo
+    decoded <- Y.decodeFileEither configPath
+    case decoded of
+        Left err -> do
+            writeFile "config-error.log" $
+                sformat ("Failed to parse "%build%": "%shown) configPath err
```
gromakovsky: Minor: IIRC, there is `string` formatter for `FilePath` (which is `String`).

tools/src/launcher/launcher-config.yaml
Pastafarianist: Why not? This is a nested dict, and an empty value for `seed` acts as if the key is not in the file at all (and it gets parsed into `cfoSeed = Nothing`).

gromakovsky: Ok, np. I don't know yaml syntax fully and was a bit lazy to google it :)

Pastafarianist:
Here is what I have done for testing:

* Took the list of parameters `scripts/launch/connect-to-cluster/mainnet-staging.sh` passes to the node.
* Swapped them into `launcher-config.yaml`.
* Made sure `stack exec -- cardano-launcher -c tools/src/launcher/launcher-config.yaml` launches the node.

A few other concerns.

- [ ] I haven't found a way to make the YAML parser interpret a missing key that should be a list as an empty list. This is needed for `walletArgs`, which I had to leave in the example config, even though `walletPath` was commented out. @neongreen, any ideas here?
- [ ] CI is broken and I don't know how to interpret Appveyor's error "Insufficient memory". It is thrown right after `IF NOT DEFINED APPVEYOR_PULL_REQUEST_NUMBER IF EXIST %STACK_ROOT% xcopy /q /s /e /r /k /i /v /h /y %STACK_ROOT% %CACHED_STACK_ROOT%`, but I don't know what this line is supposed to do or why it could fail in such way.
- [ ] We still need to find out what changes should be made to Daedalus, and in which branch, and whether the parameters should still be propagated as they are right now.

gromakovsky:
> CI is broken and I don't know how to interpret Appveyor's error "Insufficient memory". It is thrown right after IF NOT DEFINED APPVEYOR_PULL_REQUEST_NUMBER IF EXIST %STACK_ROOT% xcopy /q /s /e /r /k /i /v /h /y %STACK_ROOT% %CACHED_STACK_ROOT%, but I don't know what this line is supposed to do or why it could fail in such way.

It's fixed in cardano-sl-1.0. Will appear in master at some point. Anyway, it doesn't matter, because only optional appveyor build fails.

https://i.imgur.com/LgxBSke.png

gromakovsky:
> We still need to find out what changes should be made to Daedalus, and in which branch, and whether the parameters should still be propagated as they are right now.

https://serokell.slack.com/archives/G3K2W68AK/p1513093952000798

scripts/launch/demo.sh
```diff
@@ -141,14 +148,40 @@ while [[ $i -lt $panesCnt ]]; do
     if [[ $WALLET_TEST != "" ]] && [[ $i == $((n-1)) ]]; then
         updater_file="$config_dir/updater$i.sh"
         launcher_=$(find_binary cardano-launcher)
-        launcher_args="--node $node_ "
+
+        # TODO: make sure that this does exactly what is intended
```
Pastafarianist: @georgeee can you take a look at this and the next TODO and make sure that exactly what we want happens here?

tools/src/launcher/launcher-config.yaml
```diff
@@ -0,0 +1,76 @@
+# Path to the node executable
+nodePath: .stack-work/install/x86_64-linux-nopie/lts-9.1/8.0.2/bin/cardano-node
```
domenkozar: This is a bad idea, since it doesn't work with Nix.

Pastafarianist: This is just an example. It should be clear how to modify this under different circumstances.

PR #2081, opened 2017-12-11
[CSL-1979] Describe all types in blocks

neongreen:
I still have to describe generic/genesis blocks and addresses, and there are some todos, but otherwise the content can already be reviewed.


core/Pos/Core/Common/Types.hs
```diff
@@ -100,7 +100,7 @@ data AddrSpendingData
     -- ^ Funds can be spent by revealing a 'Script' and providing a
     -- redeemer 'Script'.
     | RedeemASD !RedeemPublicKey
-    -- ^ Funds can be spent by revealing a 'RedeemScript' and providing a
+    -- ^ Funds can be spent by revealing a 'RedeemPublicKey' and providing a
```
gromakovsky: хи-хи

docs/block-processing/types.md
```diff
@@ -0,0 +1,951 @@
...
+    = TxInUtxo
+    { txInHash  :: TxId
+    , txInIndex :: Word32
+    }
+    | TxInUnknown Word8 ByteString
+```
+
+The `TxInUnknown` constructor is needed to be able to extend `TxIn` in the
+future without breaking backwards compatibility. TODO: link wherever I
+should link
```
gromakovsky: Perhaps it's better to describe it here and then to this place from another place.

docs/block-processing/types.md
```diff
@@ -0,0 +1,951 @@
...
+```
+
+A `ProxyCert` is just a signature of `(pskDelegatePk, pskOmega)`:
+
+```
+newtype ProxyCert w = ProxyCert { unProxyCert :: CC.XSignature }
+```
+
+An invariant of `ProxySecretKey` is that its `pskCert` must be valid. TODO:
+is this right? If yes, why not `UnsafeProxySecretKey`?
```
gromakovsky: AFAIK, it's wrong, this invariant is not enforced (see e. g. `Bi` instance, it's trivial). Instead we have `verifyPsk` function and hopefully call it in all places where it's necessary.

neongreen: Is there any particular reason why we don't enforce it in `Bi`?

gromakovsky: I don't think so. Anyway, we are currently revising our approach to checked/unchecked deserialization, unsafe constructors, type invariants, etc., so probably it will be fixed in the corresponding refactoring. But for now let's describe existing code.

neongreen:
Okay, it's finished. The doc isn't very good because it doesn't include good explanations for why we include this or that field and what bad things would happen if we didn't include it, but I've already spent more time on this doc than I wanted to.

I also have the following questions:

1. Is there a good definition of “consensus data” and “extra data” that I could use in the doc?

2. Do we ever send genesis blocks or genesis headers over the network?

3. Why exactly is `psigPsk` included into a `ProxySignature`? There's an explanation in the source but I didn't understand it.

4. Why is the invariant of `ProxySecretKey` (about the cert inside being valid) not enforced? If this is an oversight, I'll change it.

5. What exactly are `_mehBlockVersion` and `_mehSoftwareVersion`? I feel that just saying “block version” and “software version” is terrible. At the very least I want to be able to link somewhere where the semantics of those fields is explained.

6. Why does `MainExtraHeaderData` include a `Hash MainExtraBodyData`, but `GenesisExtraHeaderData` doesn't include a `Hash GenesisExtraBodyData`? Is it another oversight? If yes, I guess it would be useful to check whether there are no other mistakes like this – for instance, somebody could take the definition of the block and, for each field `X`, say which field in the header guarantees the integrity of `X`. It could be a good addition to this doc.

7. Is it right that the constructors of `CommitmentsMap`, `SystemTag`, `UpdateVote`, and `CoinPortion` should have `Unsafe` prefixes, given that these types have invariants? If so, I'll make a separate PR with renamings.

8. Is it okay that this document completely skips over light signatures?

neongreen:
> Blocks and headers also have some invariants, btw.

Oops, forgot about that. Will fix now.

neongreen:
@gromakovsky please rereview my lovely pull request.

gromakovsky:
Some answers which I can provide immediately. Maybe will add more answers later.
> Do we ever send genesis blocks or genesis headers over the network?

Yes, we do. One can argue that it's unnecessary, because we can just compute them knowing main blocks, but:
1. It's still convenient to send both types of blocks, I think.
2. Imagine there is a sequence of blocks: `A B C D E` and `C` is genesis. If `B` is not from the last slot of an epoch (which is certainly possible), you need some information to figure out that it's time to compute `C`. Of course it can be less information than whole genesis block and certainly unnecessary to send leaders (because you ought to compute them anyway to verify). So there is redundancy, but it's not big (comparing to up to 21600 main blocks in an epoch) and it's done for convenience I think.

> Why exactly is psigPsk included into a ProxySignature? There's an explanation in the source but I didn't understand it.

It's better to ask @volhovm, but `dlgVerifyHeader` function gives some insights. For light delegation we need only `pskIssuerPk`, but for heavyweight delegation we also need actual PSK to check that it's present in GState.

> Why is the invariant of ProxySecretKey (about the cert inside being valid) not enforced? If this is an oversight, I'll change it.

Yes, I think it should be changed for consistency.

> What exactly are `_mehBlockVersion` and `_mehSoftwareVersion`? I feel that just saying “block version” and “software version” is terrible. At the very least I want to be able to link somewhere where the semantics of those fields is explained.

Block version can be associated with a set of protocol rules. Rules associated with `_mehBlockVersion` from a block are the rules used to create that block (i. e. the block must adhere to these rules).
`_mehSoftwareVersion` is the version of software which created given block.

> Why does MainExtraHeaderData include a Hash MainExtraBodyData, but GenesisExtraHeaderData doesn't include a Hash GenesisExtraBodyData? Is it another oversight?

Yes, I think so. Can you create an issue please? Currently it's possible for two different genesis blocks to have the same header (and identifier) which is bad.

> Is it right that the constructors of CommitmentsMap, SystemTag, UpdateVote, and CoinPortion should have Unsafe prefixes, given that these types have invariants? If so, I'll make a separate PR with renamings.

I think only `UpdateVote` should have it (and it should have a safe constructor). Constructors of other three types aren't exported, so it's not necessary to have `Unsafe` prefix there, in my opinion.

docs/block-processing/types.md
```diff
@@ -0,0 +1,1152 @@
...
+  * `BlockPSignatureLight` – a light delegation signature (not discussed in
+    this document); this signature is used when the right to issue blocks
+    has been transferred on a temporary basis
+
+  * `BlockPSignatureHeavy` – a heavy delegation signature, used when the
+    right to issue blocks has been transferred “until further notice”
+
+The thing signed by a `BlockSignature` isn't the whole block, but only
+certain parts of it (this is done because signing is an expensive process
+and we'd like to sign as few bytes as we can get away with). The exact parts
```
gromakovsky: The way it's written now is ok, but I just want to say that we sign basically `MainHeader \ BlockSignature`. The reason why we sign `MainHeader` (and not block) is because we'd like to sign fewer bytes (as you said) and the reason why we exclude `BlockSignature` is that it'd be impossible otherwise :)

PR #2083, opened 2017-12-11
connect-to-cluster: Fix configuration file path

ixmatus:
This change fixes the `connect-to-cluster` script generated by this `nix-build` invocation:

```
$ nix-build --attr connect.mainnetWallet -o connect-to-mainnet release.nix
```

The motivation behind this change is to fix this error:

```
parnell ~/Development/oss/cardano-sl ❱❱ /nix/store/ch2006v7g6cb532mr5plci53knrkf77a-wallet-connect-to-mainnet
Keeping state in state-wallet-mainnet
Launching a single node connected to 'mainnet' ...
[*production*:INFO:ThreadId 4] [2017-12-11 06:50:13.01 UTC] [Attention] Software is built with wallet part
[*production*:INFO:ThreadId 4] [2017-12-11 06:50:13.01 UTC] using configurations: ConfigurationOptions {cfoFilePath = "/nix/store/93fwvabqcpl42y0zrvmz4xhz4mmzdai4-cardano-sl/node/configuration.yaml", cfoKey = "mainnet_full", cfoSystemStart = Nothing, cfoSeed = Nothing}
cardano-node: ConfigurationParseFailure "/nix/store/93fwvabqcpl42y0zrvmz4xhz4mmzdai4-cardano-sl/node/configuration.yaml" (InvalidYaml (Just (YamlException "Yaml file not found: /nix/store/93fwvabqcpl42y0zrvmz4xhz4mmzdai4-cardano-sl/node/configuration.yaml")))
```

... which is a result of commit 2448450 moving the `node/configuration.yaml` to `lib/configuration.yaml` and not updating the path referenced in the script template within `scripts/launch/connect-to-cluster/default.nix`.

### DIFF
--- a/scripts/launch/connect-to-cluster/default.nix
+++ b/scripts/launch/connect-to-cluster/default.nix
@@ -51,7 +51,7 @@ in pkgs.writeScript "${executable}-connect-to-${environment}" ''
   ${executables.${executable}}                                     \
     ${ ifWallet "--web"}                                           \
     --no-ntp                                                       \
-    --configuration-file ${src}/node/configuration.yaml            \
+    --configuration-file ${src}/lib/configuration.yaml             \
     --configuration-key ${environments.${environment}.confKey}     \
     ${ ifWallet "--tlscert ${src}/scripts/tls-files/server.crt"}   \
     ${ ifWallet "--tlskey ${src}/scripts/tls-files/server.key"}    \


ixmatus:
Note that, after this fix, I can run the backend application successfully.

cleverca22:
i prefer to instead do things like `${src + "/lib/configuration.yaml"}`
its different in 2 ways, first, it doesnt depend on the entire source tree at runtime
second, it will directly check that the file exists at eval time, and fail early if it gets moved again

ixmatus:
@cleverca22 I agree with your suggestion however when I tried that refactor it broke the node (the `configuration.yaml` references the file `mainnet-genesis.json` with a relative path):

```
parnell ~/Development/oss/cardano-sl ❱❱ /nix/store/i5zzkqvk1jgzp8scs0hbwk01qzqws9by-wallet-connect-to-mainnet
Keeping state in state-wallet-mainnet
Launching a single node connected to 'mainnet' ...
[*production*:INFO:ThreadId 4] [2017-12-11 14:25:40.00 UTC] [Attention] Software is built with wallet part
[*production*:INFO:ThreadId 4] [2017-12-11 14:25:40.00 UTC] using configurations: ConfigurationOptions {cfoFilePath = "/nix/store/46l38i1gsam4qh0b1cp24gx6slsxnvxi-configuration.yaml", cfoKey = "mainnet_full", cfoSystemStart = Nothing, cfoSeed = Nothing}
cardano-node: /nix/store/mainnet-genesis.json: openBinaryFile: does not exist (No such file or directory)
```

I didn't go ahead with that because I only wanted to fix the immediate issue. I will see if substituting in an absolute path for the `mainnet-genesis.yaml` in the `configuration.yaml` will work okay without perturbing things too badly.

cleverca22:
ah yeah, i also ran into that problem in the docker images, it needs a new derivation containing both files, and now it gets messy, lets leave it your way for now

ixmatus:
I'm already 50% done writing the derivation, let's see if this works, it's definitely better if it does IMHO.

ixmatus:
@cleverca22 yes, that did the trick, the node is running and I can connect to it with the Daedalus client. Pushing a commit.

cleverca22:
@ixmatus oh
https://github.com/input-output-hk/daedalus/blob/devops-13/linux.nix#L32-L42
we now have this same basic code in 2 spots, the docker image, and the daedalus linux nix package

centralizing it within cardano might be cleaner

ixmatus:
Oh wow, that entire file is what I've been looking for this whole weekend. Haha.

I think a lot of that code should be pulled into `cardano-sl` abstracted from building a docker container. Ideally I could run the node as a systemd unit, and that looks like it's very close.

Additionally, I see now how you're distributing the frontend app. I originally built a Linux release and patched the elf headers of that result in a derivation.

cleverca22:
i havent had a chance to look into `yarn2nix` or `npm2nix` yet, and i'm just extracting the pre-built JS from a darwin installer

doing things right can come later

ixmatus:
@cleverca22 I tried to build it with `nixfromnpm` but it didn't go well. What you're doing is better, I ran `npm run package` to get a linux package and then I patched the elf headers so it would run on NixOS. I hadn't thought to just get the prebuilt JS and run it with a NixOS electron.

PR #2085, opened 2017-12-11
[DEVOPS-565]  appveyor:  disable branch cache updates

deepfire:
This attempts to work around the out-of-memory on AppVeyor, that happens during the `xcopy` updating the branch cache.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -132,12 +132,14 @@ test_script:
   - ps: if (-NOT (Test-Path "$($env:WORK_DIR)\$env:STACK_WORK")) { throw "ERROR... Stack cache absent $x" }
 
 after_test:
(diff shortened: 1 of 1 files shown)


gromakovsky:
If it helps, we'll need it in `cardano-sl-1.0` branch as well.

gromakovsky:
@deepfire appveyor build passed. Can you make a PR to cardano-sl-1.0 branch?

deepfire:
@gromakovsky, it appears to have helped AppVeyor -- does the problem manifest itself reliably, or we need a couple passes?

gromakovsky:
> does the problem manifest itself reliably

Yes, I think so. I haven't seen a single `continuous-integration/appveyor/branch` build which passed during last few days (and saw tens of failures).

gromakovsky:
Created #2092

PR #2090, opened 2017-12-11

volhovm:
It's a copy of #2000. It was merged by mistake and then merge was deleted by force-push.

### DIFF
--- a/node/src/Pos/Block/Logic/Header.hs
+++ b/node/src/Pos/Block/Logic/Header.hs
@@ -242,6 +242,7 @@ getHeadersFromManyTo checkpoints startM = do
 
     tip <- DB.getTipHeader @ssc
     let tipHash = headerHash tip
+    let startHash = maybe tipHash headerHash startM
 
     -- This filters out invalid/unknown checkpoints also.
     inMainCheckpoints <-
@@ -262,7 +263,8 @@ getHeadersFromManyTo checkpoints startM = do
             newestCheckpoint <-
                 maximumBy (comparing getEpochOrSlot) . catMaybes <$>
                 mapM (DB.blkGetHeader @ssc) (toList inMainCheckpoints)
-            let loadUpCond _ h = h < recoveryHeadersMessage
+            let loadUpCond (headerHash -> curH) h =
+                    curH /= startHash && h < recoveryHeadersMessage
             up <- GS.loadHeadersUpWhile @ssc newestCheckpoint loadUpCond
             res <-
                 note "loadHeadersUpWhile returned empty list" $
--- a/node/src/Pos/Block/Logic/Header.hs
+++ b/node/src/Pos/Block/Logic/Header.hs
@@ -9,15 +9,15 @@ module Pos.Block.Logic.Header
        , classifyHeaders
        , getHeadersFromManyTo
        , getHeadersOlderExp
-       , getHeadersFromToIncl
+       , getHeadersRange
        ) where
 
 import           Universum
+import           Unsafe                    (unsafeLast)
 
-import           Control.Lens              (_Wrapped)
+import           Control.Lens              (to, _Wrapped)
 import           Control.Monad.Except      (MonadError (throwError))
 import           Control.Monad.Trans.Maybe (MaybeT (MaybeT), runMaybeT)
-import           Data.List.NonEmpty        ((<|))
 import qualified Data.Text                 as T
 import           Formatting                (build, int, sformat, (%))
 import           Serokell.Util.Text        (listJson)
@@ -31,11 +31,10 @@ import           Pos.Block.Pure            (VerifyHeaderParams (..), verifyHeade
 import           Pos.Configuration         (HasNodeConfiguration, recoveryHeadersMessage)
 import           Pos.Core                  (BlockCount, EpochOrSlot (..),
                                             HasConfiguration, HeaderHash, SlotId (..),
-                                            blkSecurityParam, difficultyL, epochOrSlotG,
-                                            getChainDifficulty, getEpochOrSlot,
-                                            headerHash, headerHashG, headerSlotL,
-                                            prevBlockL)
-import           Pos.Core.Configuration    (genesisHash)
+                                            blkSecurityParam, difficultyL, epochIndexL,
+                                            epochOrSlotG, getChainDifficulty,
+                                            getEpochOrSlot, headerHash, headerHashG,
+                                            headerSlotL, prevBlockL)
 import           Pos.Crypto                (hash)
 import           Pos.DB                    (MonadDBRead)
 import qualified Pos.DB.Block              as DB
@@ -328,35 +327,86 @@ getHeadersOlderExp upto = do
                 | otherwise = selGo es ii $ succ skipped
         in selGo elems ixs 0
 
--- CSL-396 don't load all the blocks into memory at once
--- | Given @from@ and @to@ headers where @from@ is older (not strict)
--- than @to@, and valid chain in between can be found, headers in
--- range @[from..to]@ will be found.
-getHeadersFromToIncl
-    :: forall ssc m .
-       (HasConfiguration, MonadDBRead m, SscHelpersClass ssc)
-    => HeaderHash -> HeaderHash -> m (Maybe (OldestFirst NE HeaderHash))
-getHeadersFromToIncl older newer = runMaybeT . fmap OldestFirst $ do
+-- | Given optional @depthLimit@, @from@ and @to@ headers where @from@
+-- is older (not strict) than @to@, and valid chain in between can be
+-- found, headers in range @[from..to]@ will be found. If difference
+-- between headers is more than @depthLimit@, error will be thrown.
+getHeadersRange ::
+       forall ssc m. (HasConfiguration, MonadDBRead m, SscHelpersClass ssc)
+    => Maybe Word
+    -> HeaderHash
+    -> HeaderHash
+    -> m (Either Text (OldestFirst NE HeaderHash))
+getHeadersRange _ older newer | older == newer = runExceptT $ do
+    unlessM (isJust <$> (DB.blkGetHeader @ssc newer)) $
+        throwError "getHeadersRange: can't find newer-older header"
+    pure $ OldestFirst $ one newer
+getHeadersRange depthLimitM older newer = runExceptT $ do
     -- oldest and newest blocks do exist
-    start <- MaybeT $ DB.blkGetHeader @ssc newer
-    end   <- MaybeT $ DB.blkGetHeader @ssc older
-    guard $ getEpochOrSlot start >= getEpochOrSlot end
-    let lowerBound = getEpochOrSlot end
-    if newer == older
-    then pure $ one newer
-    else loadHeadersDo lowerBound (one newer) $ start ^. prevBlockL
+    newerHd <- fromMaybeM "can't retrieve newer header" $ DB.blkGetHeader @ssc newer
+    olderHd <- fromMaybeM "can't retrieve older header" $ DB.blkGetHeader @ssc older
+    let olderD = olderHd ^. difficultyL
+    let newerD = newerHd ^. difficultyL
+
+    -- Proving newerD >= olderD
+    let newerOlderF = "newer: "%build%", older: "%build
+    when (newerD == olderD) $
+        throwError $
+        sformat ("getHeadersRange: newer and older headers have "%
+                 "the same difficulty, but are not equal. "%newerOlderF)
+                newerHd olderHd
+    when (newerD < olderD) $
+        throwError $
+        sformat ("getHeadersRange: newer header is less dificult than older one. "%
+                 newerOlderF)
+                newerHd olderHd
+
+    -- How many epochs does this range cross.
+    let genDiff :: Int
+        genDiff = fromIntegral $ newerHd ^. epochIndexL - olderHd ^. epochIndexL
+    -- Number of blocks is difficulty difference + number of genesis blocks.
+    let depthDiff :: Word
+        depthDiff = fromIntegral $ genDiff + fromIntegral (newerD - olderD)
+
+    whenJust depthLimitM $ \depthLimit ->
+        when (depthDiff > depthLimit) $
+        throwError $
+        sformat ("getHeadersRange: requested "%int%" headers, but depthLimit is "%
+                 int%". Headers: "%newerOlderF)
+                depthDiff
+                depthLimit
+                newerHd olderHd
+
+    -- We load these depthDiff blocks.
+    let cond curHash _depth = curHash /= newer
+
+    -- This is [oldest..newest) headers, oldest first
+    allExceptNewest <- GS.loadHashesUpWhile older cond
+
+    -- Sometimes we will get an empty list, if we've just switched the
+    -- branch (after first checks are performed here) and olderHd is
+    -- no longer in the main chain.
+    when (null $ allExceptNewest ^. _Wrapped) $ throwError $
+        "getHeadersRange: loaded 0 headers though checks passed. " <>
+        "May be (very rare) concurrency problem, just retry"
+
+    -- It's safe to use 'unsafeLast' here after the last check.
+    let lastElem = allExceptNewest ^. _Wrapped . to unsafeLast
+    when (newerHd ^. prevBlockL . headerHashG /= lastElem) $
+        throwError $
+        sformat ("getHeadersRange: newest block parent is not "%
+                 "equal to the newest one iterated. It may indicate recent fork or "%
+                 "inconsistent request. Newest: "%build%
+                 ", last list hash: "%build%", already retrieved (w/o last): "%listJson)
+                newerHd
+                lastElem
+                allExceptNewest
+
+    -- We append last element and convert to nonempty.
+    let conv =
+           fromMaybe (error "getHeadersRange: can't happen") .
+           nonEmpty .
+           (++ [newer])
+    pure $ allExceptNewest & _Wrapped %~ conv
   where
-    loadHeadersDo
-        :: EpochOrSlot
-        -> NonEmpty HeaderHash
-        -> HeaderHash
-        -> MaybeT m (NonEmpty HeaderHash)
-    loadHeadersDo lowerBound hashes nextHash
-        | nextHash == genesisHash = mzero
-        | nextHash == older = pure $ nextHash <| hashes
-        | otherwise = do
-            nextHeader <- MaybeT $ (DB.blkGetHeader @ssc) nextHash
-            guard $ getEpochOrSlot nextHeader > lowerBound
-            -- hashes are being prepended so the oldest hash will be the last
-            -- one to be prepended and thus the order is OldestFirst
-            loadHeadersDo lowerBound (nextHash <| hashes) (nextHeader ^. prevBlockL)
+    fromMaybeM r m = ExceptT $ maybeToRight ("getHeadersRange: " <> r) <$> m
--- a/node/src/Pos/Block/Logic/Header.hs
+++ b/node/src/Pos/Block/Logic/Header.hs
@@ -329,8 +329,9 @@ getHeadersOlderExp upto = do
 
 -- | Given optional @depthLimit@, @from@ and @to@ headers where @from@
 -- is older (not strict) than @to@, and valid chain in between can be
--- found, headers in range @[from..to]@ will be found. If difference
--- between headers is more than @depthLimit@, error will be thrown.
+-- found, headers in range @[from..to]@ will be found. If the number
+-- of headers in the chain (which should be returned) is more than
+-- @depthLimit@, error will be thrown.
 getHeadersRange ::
        forall ssc m. (HasConfiguration, MonadDBRead m, SscHelpersClass ssc)
     => Maybe Word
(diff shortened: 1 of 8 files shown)


volhovm:
cc @dcoutts waiting for your approval.

node/src/Pos/GState/BlockExtra.hs
```diff
@@ -113,62 +113,68 @@ instance HasConfiguration => RocksBatchOp BlockExtraOp where
...
-    -> (r -> b -> m r)
-    -> r
+    => (HeaderHash -> m (Maybe b)) -- ^ For each header we get b(lund)
+    -> a                           -- ^ We start iterating from it
+    -> (b -> Int -> Bool)          -- ^ Condition on b and depth
+    -> (r -> b -> m r)             -- ^ Conversion function
+    -> r                           -- ^ Starting value
     -> m r
-foldlUpWhileM morphM start condition accM init =
+foldlUpWhileM getData start condition accM init =
```
avieth: I like the direction of this new type. Why not go even further: eliminate the mention of `HeaderHash`, `MonadDBRead`, and `HasHeaderHash`? This will make it more easily testable/benchmarkable. Do we even need `MonadDBRead` as is? Doesn't the `HeaderHash -> m (Maybe b)` express that?

avieth: Ah, I guess `resolveForwardLink` is the `MonadDBRead` part. But can't get resolve it *purely*, from the `b`?

node/src/Pos/Security/Workers.hs
```diff
@@ -85,8 +87,16 @@ checkForReceivedBlocksWorkerImpl
...
-    repeatOnInterval (const (sec' 4)) . recoveryCommGuard $
-        whenM (needRecovery @ctx @ssc) $ triggerRecovery enqueueMsg
+    repeatOnInterval (const (sec' 4)) $ do
+        -- TODO we should be in fact checking this predicate more often.
+        -- I suggest checking it, eg, every second. But if we succeeded
+        -- and triggered recovery, we make a longer delay, say slot duration/2,
+        -- so that we don't overload the retrieval worker.
+        need <- needRecovery @ctx @ssc
+        recoveryHappening <- recoveryInProgress
+        when (need && not recoveryHappening) $ triggerRecovery enqueueMsg
```
avieth: I don't understand the motivation for this change. Aren't the semantics the same as before?

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -328,35 +327,89 @@ getHeadersOlderExp upto = do
...
+-- found, headers in range @[from..to]@ will be found. If the number
+-- of headers in the chain (which should be returned) is more than
+-- @depthLimit@, error will be thrown.
+getHeadersRange ::
+       forall ssc m. (HasConfiguration, MonadDBRead m, SscHelpersClass ssc)
+    => Maybe Word
+    -> HeaderHash
+    -> HeaderHash
+    -> m (Either Text (OldestFirst NE HeaderHash))
+getHeadersRange _ older newer | older == newer = runExceptT $ do
```
avieth: I know it's a bit silly, but shouldn't `Just 0` for the depth limit be respected here? The number of headers in the chain is `1 > 0`.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -328,35 +327,89 @@ getHeadersOlderExp upto = do
...
-    end   <- MaybeT $ DB.blkGetHeader @ssc older
-    guard $ getEpochOrSlot start >= getEpochOrSlot end
-    let lowerBound = getEpochOrSlot end
-    if newer == older
-    then pure $ one newer
-    else loadHeadersDo lowerBound (one newer) $ start ^. prevBlockL
+    newerHd <- fromMaybeM "can't retrieve newer header" $ DB.blkGetHeader @ssc newer
+    olderHd <- fromMaybeM "can't retrieve older header" $ DB.blkGetHeader @ssc older
+    let olderD = olderHd ^. difficultyL
+    let newerD = newerHd ^. difficultyL
```
avieth: What's the significance of using difficulty rather than `getEpochOrSlot` here?

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -328,35 +327,89 @@ getHeadersOlderExp upto = do
...
+       forall ssc m. (HasConfiguration, MonadDBRead m, SscHelpersClass ssc)
+    => Maybe Word
+    -> HeaderHash
+    -> HeaderHash
+    -> m (Either Text (OldestFirst NE HeaderHash))
+getHeadersRange _ older newer | older == newer = runExceptT $ do
+    unlessM (isJust <$> (DB.blkGetHeader @ssc newer)) $
+        throwError "getHeadersRange: can't find newer-older header"
+    pure $ OldestFirst $ one newer
+getHeadersRange depthLimitM older newer = runExceptT $ do
```
avieth: This seems to have become significantly more complicated. What are the benefits? Is it faster?

node/src/Pos/Block/Logic/Header.hs
volhovm: It's minor, but you're right. I'll add an extra check.

node/src/Pos/Block/Logic/Header.hs
volhovm: 1. Complexity here is subjective, I actually find new implementation far more simple. What is scary here is amount of comments, but they are mostly redundant and help to understand what's going on.
2. Yes, it's faster. See YT task comments. This speeds up handler by not deserializing headers twice (we use forward links which are, essentially, map `HeaderHash -> HeaderHash`).

node/src/Pos/Block/Logic/Header.hs
volhovm: The particular function is operating in terms of difficulty. We can't give more than `depthLimit` blocks at once, so even if I used `EpochOrSlot` (which i actually did in the beginning), I would need to convert it to number of blocks somehow (for example, counting difficulty in traversal).

node/src/Pos/Security/Workers.hs
volhovm: No, it's not. Previously the worker wouldn't kick in if we were far away but with no recovery variable in place (that's how `recoveryCommGuard` works). Now we kick in in this condition (need + don't have variable). Anyway, this worker is optimized in master already: https://github.com/input-output-hk/cardano-sl/pull/2121

node/src/Pos/GState/BlockExtra.hs
volhovm: > why not go even further

1. Because `MonadDBRead` is a neat interface. When I'm using it, i know nobody will go crazy and pass random monadic function into `foldlUpWhileM`. It's safer and more clear. It also doesn't complicate testing and benchmarking much, because we already have pure db framework, it's just to be used.
2. Why would you even want to eliminate `HeaderHash` and `HasHeaderHash`? It's so nice to use this polymorphic function and do not think about what do you pass to it -- it should just have `HasHeaderHash`. In this way, `foldlUpWhileM` is as fast as it would be if not constrained by anything, but it's flexible -- noone will ever need to make its copy for different `a` that are `HasHeaderHash a`. Polymorphism is cheap and good.

>`resolveForwardLink` is the `MonadDBRead` part. But can't get resolve it purely, from the `b`?

I don't understand the question. `resolveForwardLink` has `HeaderHash -> m (Maybe HeaderHash)` type. In order to support both real and mockable DBs we need to have this `m` here: yes, for pure db it's just `HeaderHash -> Maybe HeaderHash`, but the more abstract thing is, the easier it is to work with it.

avieth: > Because `MonadDBRead` is a neat interface. When I'm using it, i know nobody will go crazy and pass random monadic function into `foldlUpWhileM`. It's safer and more clear. It also doesn't complicate testing and benchmarking much, because we already have pure db framework, it's just to be used.

Well, I can still go crazy and "pass" a random monadic function, by way of a bad `MonadDBRead` instance, so this buys us nothing, it only adds indirection. Which may or may not be good or bad, I'm only saying, but for testing/benchmarking it's certainly a pain. Yesterday I wanted to wire up a benchmark which uses `Int` rather than `HeaderHash`, `BlockHeader` and doesn't use a DB at all. Why should I have to invent a new monad and give it a `MonadDBRead` instance when I can just say `getHeader i = pure (Just i) :: Int -> Identity (Maybe Int))`?

> Why would you even want to eliminate `HeaderHash` and `HasHeaderHash`? It's so nice to use this polymorphic function and do not think about what do you pass to it -- it should just have `HasHeaderHash`. In this way, `foldlUpWhileM` is as fast as it would be if not constrained by anything, but it's flexible -- noone will ever need to make its copy for different a that are `HasHeaderHash a`. Polymorphism is cheap and good.

I agree, polymorphism is great, and what I'm asking for is *more* of it. Eliminate `HeaderHash`, replacing it with `headerHash`, and eliminate `HasHeaderHash header`, replacing it with `header -> m (Maybe headerHash)`. That's more polymorphic; cheaper and gooder. Then the existing function can be implemented as a specialization of the more polymorphic one.

node/src/Pos/Security/Workers.hs
avieth: It's my understanding that `recoveryCommGuard` runs its action only if recovery is not in progress. Yes, it's true

```Haskell
-- onIgnore just logs something
recoveryCommGuard action = ifM recoveryInProgress onIgnore action
```

In practice I believe this just checks whether some `TMVar` is full.

Previously: every 4 seconds we
- Check if recovery is on and if so do nothing (`recoveryCommGuard` just logs a message)
- Otherwise check whether we need recovery and if so, trigger it

That's to say: if we're not in recovery, we will always check whether we're in recovery and trigger it if necessary. Note that it's possible for recovery to be triggered elsewhere (block retrieval code) in the mean-time, resulting in two triggers where one would have sufficed.

Now: every 4 seconds we
- Check whether we need recovery
- Check whether recovery is in progress
- Trigger recovery if and only if we need it and are not currently doing it

But isn't that the same as before? `recoveryCommGuard` ensures that `not recoveryHappening` is true (unless you account for concurrency :) ) and the inner action checks whether `needRecovery` is true, and the trigger happens only then.

volhovm: >In practice I believe this just checks whether some TMVar is full.

No.
```
recoveryCommGuard actionName action =
    getSyncStatusK >>= \case
        SSKindaSynced -> action
        status ->
            logDebug $
            sformat ("recoveryCommGuard: we are skipping action '"%stext%
                     "', because "%build) actionName status
```

Previously: we're `inf` blocks late and not doing recovery. `recoveryCommGuard` won't let us trigger recovery. Because we're not `SSKindaSynced`.

Currently: we're `inf` blocks late and not doing recovery. `need == True`, `recoveryHappening = False`.

avieth: Oh, I see this in the cardano-sl-1.0 tree

```Haskell
class Monad m => MonadRecoveryInfo m where
    -- | Returns if 'RecoveryHeader' is 'Just' (which is equivalent to “we're
    -- doing recovery”).
    recoveryInProgress :: m Bool

-- | This is a helper function which runs given action only if we are
-- not doing recovery at this moment.  It is useful for workers which
-- shouldn't do anything while we are not synchronized.
recoveryCommGuard
    :: (MonadRecoveryInfo m, WithLogger m)
    => m () -> m ()
recoveryCommGuard action = ifM recoveryInProgress onIgnore action
  where
    onIgnore = logDebug "recoveryCommGuard: recovery in process, skipping action"
```

avieth: Oh right, and the instance:

```Haskell
instance
    ( Monad m, MonadIO m, MonadReader ctx m
    , HasLens RecoveryHeaderTag ctx (RecoveryHeader ssc)
    ) => MonadRecoveryInfo m where
    recoveryInProgress = do
        var <- view (lensOf @RecoveryHeaderTag)
        isJust <$> atomically (STM.tryReadTMVar var)
```

volhovm: Wow, nice catch. `cardano-sl-1.0` is too old, i might have confused implementations. Anyway, it's not harmful -- my changes introduce no semantic changes.

avieth: I thought maybe I had checked the wrong base branch. Now it's `release-1.0.4`, and maybe the recovery comm guard has changed to what you expected?

volhovm: No, you're perfectly right. My modification in this branch doesn't make any difference. I wouldn't remove it from this PR though: it will be simpler for me to merge it into `master`, because there will be a conflict (this particular piece of code is called recoveryTriggerWorker in master) and i'll check everything twice. So, yes, you're right, but let's leave it.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -224,52 +223,52 @@ classifyHeaders inRecovery headers = do
...
-        logDebug $ "getHeadersFromManyTo: loaded non-empty list of headers, returning"
-        pure res
+
+    tip <- DB.getTipHeader @ssc
+    let tipHash = headerHash tip
+
+    -- This filters out invalid/unknown checkpoints also.
+    inMainCheckpoints <-
+        noteM "no checkpoints are in the main chain" $
+        nonEmpty <$> filterM GS.isBlockInMainChain (toList checkpoints)
```
avieth: `isBlockInMainChain` is faster than `blkGetHeader` I take it? So we get a speedup here?

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -224,52 +223,52 @@ classifyHeaders inRecovery headers = do
...
+            newestCheckpoint <-
+                maximumBy (comparing getEpochOrSlot) . catMaybes <$>
+                mapM (DB.blkGetHeader @ssc) (toList inMainCheckpoints)
+            let loadUpCond _ h = h < recoveryHeadersMessage
+            up <- GS.loadHeadersUpWhile @ssc newestCheckpoint loadUpCond
+            res <-
+                note "loadHeadersUpWhile returned empty list" $
+                _Wrapped nonEmpty (toNewestFirst $ over _Wrapped (drop 1) up)
+            logDebug $ "getHeadersFromManyTo: loaded non-empty list of headers, returning"
+            pure res
```
avieth: This whole thing is shorter and that's good. One issue I have is that `startM` doesn't seem to be used.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -328,35 +327,96 @@ getHeadersOlderExp upto = do
...
+    -- This is [oldest..newest) headers, oldest first
+    allExceptNewest <- GS.loadHashesUpWhile older cond
+
+    -- Sometimes we will get an empty list, if we've just switched the
+    -- branch (after first checks are performed here) and olderHd is
+    -- no longer in the main chain.
+    -- CSL-1950 We should use snapshots here.
+    when (null $ allExceptNewest ^. _Wrapped) $ throwError $
+        "getHeadersRange: loaded 0 headers though checks passed. " <>
+        "May be (very rare) concurrency problem, just retry"
```
avieth: Ah, this is unfortunate. I take it this bug was always there?

I recommend putting a short explanation in the log message... maybe "branch may have forked in the midst of loading blocks" or something similar.

node/src/Pos/Security/Workers.hs
avieth: Ok

node/src/Pos/GState/BlockExtra.hs
```diff
@@ -113,62 +113,68 @@ instance HasConfiguration => RocksBatchOp BlockExtraOp where
...
+
 -- | Returns headers loaded up.
 loadHeadersUpWhile
     :: (MonadBlockDB ssc m, HasHeaderHash a)
     => a
     -> (BlockHeader ssc -> Int -> Bool)
     -> m (OldestFirst [] (BlockHeader ssc))
-loadHeadersUpWhile start condition =
-    loadUpWhile (view blockHeader . fst) start condition
+loadHeadersUpWhile = loadUpWhile $ fmap (fmap $ view blockHeader) . blkGetBlock
```
avieth: Is it possible to do this without `blkGetBlock`? Deserializing that is the heavy bit that we want to avoid.

node/src/Pos/Block/Logic/Header.hs
```diff
@@ -224,52 +223,52 @@ classifyHeaders inRecovery headers = do
...
+        -- block, i know the previous one".
+        then pure $ one tip
+        -- If optimization doesn't apply, just iterate down-up
+        -- starting with the newest header.
+        else do
+            newestCheckpoint <-
+                maximumBy (comparing getEpochOrSlot) . catMaybes <$>
+                mapM (DB.blkGetHeader @ssc) (toList inMainCheckpoints)
+            let loadUpCond _ h = h < recoveryHeadersMessage
+            up <- GS.loadHeadersUpWhile @ssc newestCheckpoint loadUpCond
```
avieth: As far as I can tell this should give a performance *regression* rather than improvement! Previously we would `loadHeadersByDepthWhile` in case one of the checkpoints is in the main chain, and that one is better because it doesn't have to deserialize any blocks. `loadHeadersUpWhile` apparently does, making it slower, although maybe we can patch it so that it doesn't have to deserialize blocks?

node/src/Pos/Block/Logic/Header.hs
volhovm: We perform one traversal versus two (loadHeadersByDepthWhile and then loadHeadersUpWhile).

node/src/Pos/Block/Logic/Header.hs
volhovm: > in case one of the checkpoints is in the main chain

This is not right. We did iterate `recoveryHeadersMessage` blocks deep or unless we meet a checkpoint, which was never a case with recovery, so we effectively throw one iteration away. Regarding `loadHeadersUpWhile` -- yes, it's stupid that the whole block is deserialized, i can easily optimize it. Thanks.

node/src/Pos/Block/Logic/Header.hs
volhovm: There's an issue for it: https://iohk.myjetbrains.com/youtrack/issue/CSL-1950

node/src/Pos/GState/BlockExtra.hs
volhovm: It is, i'll deliver a fix.

node/src/Pos/Block/Logic/Header.hs
avieth: Sure but in the normal (non recovery) case there's only one traversal and it includes no block deserialization. Now in every case there is block deserialization.

node/src/Pos/Block/Logic/Header.hs
avieth: Obligatory https://en.wikipedia.org/wiki/Murphy%27s_law

node/src/Pos/Block/Logic/Header.hs
volhovm: 1. You don't seem to understand the usecase and how this function was designed previously, general question is: it's still faster or almost the same speed, i believe.
2. It doesn't really matter because I'll fix `loadHeadersUpWhile` in a moment.

node/src/Pos/Block/Logic/Header.hs
volhovm: @avieth sure, i _obviously_ thought of it when writing the comment, sanity check passed. What do we do if error kicks in:
1. Locating the error message in codebase (`grep`, takes 15 seconds in total)
2. Finding comment about YT issue (3 seconds)
3. Deciding if the fix is needed and pushing it (1h at most)

:)

node/src/Pos/Block/Logic/Header.hs
avieth: > You don't seem to understand the usecase and how this function was designed previously

Sure, I'd believe this.

> it's still faster or almost the same speed, i believe.

But this is harder to believe.

> It doesn't really matter because I'll fix loadHeadersUpWhile in a moment

Ok, that should make it better.

node/src/Pos/Block/Logic/Header.hs
avieth: > Deciding if the fix is needed and pushing it (1h at most)

Just 1 hour to implement a fix for this and push it? I highly doubt that.

volhovm: It's a symbolic number, the main point is that we don't have any evidence that this issue may be problematic. There are other different places (one or two, i believe) in CSL that also have this assumption.

avieth: Sure, I'm not saying it has to be fixed here, since it's not a new bug.

avieth:
@jmitchell @domenkozar This patch should further alleviate relay stress along with https://github.com/input-output-hk/cardano-sl/pull/2146

domenkozar:
@avieth is it ready to be tested?

volhovm:
@domenkozar totally. I'm testing it on staging atm (csl-2043 cluster).

avieth:
> @avieth is it ready to be tested?

@domenkozar Yes. What's your take on the policy for merging to `release/1.0.4`? Should we merge this and https://github.com/input-output-hk/cardano-sl/pull/2146 ?

domenkozar:
@avieth correct

PR #2091, opened 2017-12-11
[CSL-2044] Make 'instance Bi VssCertData' compatible with mainnet's one

gromakovsky:
The issue is very similar to CSL-1802 (#1828). This incompatibility was broken
was 'VssCertificatesMap' became newtype. Previously it was a 'type' alias,
so it was encoded as a 'HashMap'. But 'VssCertificatesMap' newtype is encoded
as a list.
This commit changes serialization format of 'VssCertData' to the old one.

This particular problem issue is less significant than CSL-1802, because
'VssCertData' is stored only in DB, but not in blocks, so it's basically an
implementation detail. That's why it hasn't been noticed for a long time.

Alternative solution would be to stop using old format (which was a bit
stupid) and perform DB migration. I've chosen this one, because DB migration
is hard and inefficiency of the old format is insignificant, because we have
only 7 certificates nowadays anyway.

### DIFF
--- a/ssc/Pos/Binary/Ssc/Types.hs
+++ b/ssc/Pos/Binary/Ssc/Types.hs
@@ -4,26 +4,40 @@ module Pos.Binary.Ssc.Types () where
 
 import           Universum
 
-import           Pos.Binary.Class (Cons (..), Field (..), deriveSimpleBi, deriveSimpleBiCxt)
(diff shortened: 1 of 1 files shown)


gromakovsky:
@neongreen fixed.

PR #2094, opened 2017-12-12
[CSL-1939] add a hydra job to check the swagger file

cleverca22:


cleverca22:
https://github.com/input-output-hk/iohk-ops/pull/211 will have to be deployed before the hook appears within github

until then, you need to check on hydra: https://hydra.iohk.io/build/57726

PR #2096, opened 2017-12-12
Update instruction to connect Explorer w/ mainnet

sectore:
Just a record of what I've learned to test #2095 today. It might be helpful for anybody else :)

### DIFF
--- a/docs/how-to/connect-to-cluster.md
+++ b/docs/how-to/connect-to-cluster.md
@@ -4,9 +4,10 @@
 2. Make sure you’re on latest `cardano-sl-X.X` branch
 3. To build a script that will contain everything needed to connect to mainnet: `$ nix-build release.nix -A connect.mainnetWallet -o connect-to-mainnet`
 4. Alternatively you can connect to different environments and different executables by building the scripts:
-- Explorer node with mainnet: `$ nix-build release.nix -A connect.mainnetExplorer -o connect-explorer-to-mainnet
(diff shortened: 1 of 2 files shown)


gromakovsky:
Let's merge without waiting for CI maybe? This PR only changes documentation, those file don't affect CI. And currently there is a big backlog of MacOS builds.

deepfire:
@gromakovsky, agreed, let's do it.

gromakovsky:
@sectore by the way, I have a minor comment on this PR: if your branch contains only 1 commit which is not in base branch and you want to synchronize it, I think it's better to rebase instead of merging. Rebasing in this case should be as easy as merging, because there is only one commit (so you need to resolve conflicts only once). In case of this PR it should be trivial AFAUI (no conflicts).
If you rebase, the history will be cleaner, there will be less commits in `Commits` tab and nobody will have to wonder whether you modified something in your merge commit.

I don't mind merging base branch in feature branch, I think it's the right thing to do if your feature branch has many commits and significantly diverged from the base branch, because it will be easier to resolve conflicts. But when there is only 1 commit, I think there are no real advantages in doing merge (unless you working with someone else in the same branch and don't want to force push).

Anyway, it's minor, you may disagree and it won't be a problem :)

sectore:
@gromakovsky Agree and I'll have keep it in mind. Thanks for sharing your thoughts!

PR #2097, opened 2017-12-12
[CSL-2034]  fix | appveyor:  the OpenSSL package was updated, fix URL

deepfire:
cc @gromakovsky @akegalj


gromakovsky:
We also need it in `cardano-sl-1.0`.

gromakovsky:
I am going to merge this PR. It affects only appveyor config, so travis checks are irrelevant.
`continuous-integration/appveyor/branch`, but it's a known problem, the fix hasn't arrived at master yet. And that build is optional anyway.
The only build we should care about is `continuous-integration/appveyor/pr` and it passed.

PR #2104, opened 2017-12-13
[CSL-2078] Delete installer after successful exit more reliably

gromakovsky:
After we successfully apply an update in cardano-launcher we do two things
(apart from logging):
1. Write that update was installed to DB.
2. Remove update file.

The problem is that if `cardano-node` wasn't stopped for some reason (and
unfortunately it's often the case nowadays), we will fail to open DB and
won't remove installer file. It will lead to infinite loop. After user opens
updated Daedalus, they will be prompted to install an update again.

The proper fix of course is to make node exit more reliably when Daedalus
exits. However, various bad things can happen when we want to write to DB and
in any case we want to remove installer file to avoid infinite loop. So using
`finally` to achieve it seems quite reasonable.


gromakovsky:
I don't think it'll ever be merged to 1.0.4 branch. Will change base branch to `develop`.

PR #2107, opened 2017-12-13
[CSL-1940] V1 handler `newTransaction`

sectore:
_Task_: https://iohk.myjetbrains.com/youtrack/issue/CSL-1940

### DIFF
--- a/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Transactions.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Transactions.hs
@@ -6,14 +6,21 @@ import           Universum
 
 import           Cardano.Wallet.API.V1.Migration (HasCompileInfo, HasConfigurations, MonadV1,
                                                   migrate)
+import qualified Pos.Wallet.Web.ClientTypes.Types as V0
 import qualified Pos.Wallet.Web.Methods.History as V0
+import qualified Pos.Wallet.Web.Methods.History as V0
+import qualified Pos.Wallet.Web.Methods.Payment as V0
+import qualified Pos.Wallet.Web.Methods.Txp as V0
 
 import           Cardano.Wallet.API.Request
 import           Cardano.Wallet.API.Response
+import           Cardano.Wallet.API.V1.Migration (HasCompileInfo, HasConfigurations, MonadV1,
+                                                  migrate)
 import qualified Cardano.Wallet.API.V1.Transactions as Transactions
 import           Cardano.Wallet.API.V1.Types
 import qualified Data.IxSet.Typed as IxSet
 
+import qualified Data.List.NonEmpty as NE
 import           Servant
 import           Test.QuickCheck (arbitrary, generate)
 
@@ -26,8 +33,18 @@ handlers = newTransaction
         :<|> allTransactions
         :<|> estimateFees
 
-newTransaction :: Payment -> MonadV1 (WalletResponse Transaction)
-newTransaction _ = single <$> (liftIO $ generate arbitrary)
+newTransaction
+    :: forall ctx m . (V0.MonadWalletTxFull ctx m)
+    => NewPayment -> m (WalletResponse Transaction)
+newTransaction Payment {..} = do
+    let spendingPw = fromMaybe mempty pmtSpendingPassword
+    cAccountId <- migrate (pmtSourceWallet, pmtSourceAccount)
+    addrCoinList <- migrate $ NE.toList pmtDestinations
+  -- Set `OptimiseForSecurityPolicy` as default policy
+    policy <- migrate $ fromMaybe OptimiseForSecurityPolicy pmtGroupingPolicy
+    let batchPayment = V0.NewBatchPayment cAccountId addrCoinList policy
+    cTx <- V0.newPaymentBatch spendingPw batchPayment
+    single <$> migrate cTx
 
 -- | The conclusion is that we want just the walletId for now, the details
 -- in CSL-1917.
@@ -48,5 +65,4 @@ allTransactions walletId requestParams = do
                               (IxSet.fromList <$> transactions)
 
 estimateFees :: Payment -> MonadV1 (WalletResponse EstimatedFees)
-estimateFees _ = single <$> (liftIO $ generate arbitrary)
-
+estimateFees _ = single <$> liftIO (generate arbitrary)
--- a/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Transactions.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Handlers/Transactions.hs
@@ -8,23 +8,19 @@ import           Cardano.Wallet.API.V1.Migration (HasCompileInfo, HasConfigurati
                                                   migrate)
 import qualified Pos.Wallet.Web.ClientTypes.Types as V0
 import qualified Pos.Wallet.Web.Methods.History as V0
-import qualified Pos.Wallet.Web.Methods.History as V0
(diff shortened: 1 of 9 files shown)


wallet-new/server/Cardano/Wallet/API/V1/Handlers/Transactions.hs
```diff
@@ -24,8 +28,18 @@ handlers = newTransaction
...
-newTransaction _ = single <$> (liftIO $ generate arbitrary)
+newTransaction
+    :: forall ctx m . (V0.MonadWalletTxFull ctx m)
+    => NewPayment -> m (WalletResponse Transaction)
+newTransaction Payment {..} = do
+    let spendingPw = fromMaybe mempty pmtSpendingPassword
+    cAccountId <- migrate (pmtSourceWallet, pmtSourceAccount)
+    addrCoinList <- migrate $ NE.toList pmtDestinations
+  -- Set `OptimiseForSecurityPolicy` as default policy
+    policy <- migrate $ fromMaybe OptimiseForSecurityPolicy pmtGroupingPolicy
```
adinapoli-iohk: What about creating a `defaultGroupingPolicy :: TransactionGroupingPolicy` function inside `V1.Types` which we can then use here and in all the other places which needs a sensible default? Sounds more refactoring-efficient that scattering the same value in multiple places 😉

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
```diff
@@ -129,6 +130,10 @@ instance Migrate (V1.WalletId, V1.AccountId) V0.AccountId where
     eitherMigrate (walId, accId) =
         V0.AccountId <$> eitherMigrate walId <*> pure accId
 
+instance Migrate (V1.WalletId, V1.AccountId) V0.CAccountId where
+    eitherMigrate (walId, accId) =
+        V0.encodeCType <$> eitherMigrate (walId, accId)
```
adinapoli-iohk: If you write the complementary `Migrate` instances (i.e from `V0.CAccountId` to `(V1.WalletId, V1.AccountId)`) you can then add a roundtrip test.

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
```diff
@@ -168,6 +169,10 @@ instance Migrate (V0.CId V0.Addr, V0.CCoin) V1.PaymentDistribution where
         pdAmount  <- eitherMigrate cCoin
         pure $ V1.PaymentDistribution {..}
 
+instance Migrate V1.PaymentDistribution (V0.CId V0.Addr, Core.Coin) where
```
adinapoli-iohk: Ditto as above.

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
```diff
@@ -194,3 +199,9 @@ instance Migrate (Map Core.TxId (V0.CTx, POSIXTime), Word) [V1.Transaction] wher
     eitherMigrate txsMapAndSize = do
         let txsMapValues = elems . fst $ txsMapAndSize
         mapM (eitherMigrate . fst) txsMapValues
+
+instance Migrate V1.TransactionGroupingPolicy V0.InputSelectionPolicy where
```
adinapoli-iohk: Ditto as above.

wallet-new/src/Cardano/Wallet/API/V1/Transactions.hs
```diff
@@ -9,7 +9,7 @@ import           Servant
 
 type API
     =    "transactions" :> Summary "Generates a new transaction from the source to one or multiple target addresses."
-                        :> ReqBody '[JSON] Payment
+                        :> ReqBody '[JSON] NewPayment
```
adinapoli-iohk: `New Payment`, for an easy refactoring. You might need to add an extra entry to the type family in `V1.Types`.

EDIT: Actually, let's use `Payment` directly. There is no need for a `newtype` here.

The type families `New` and `Update` are useful when we have a somewhat standard CRUD resource we can update, but for a `Transaction` doesn't make sense. If we want to be consistent with all we have done so far, maybe we should have a `New Transaction` which maps to a `Payment`. Thoughts? cc @akegalj @et4te @Martoon-00

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -327,6 +330,9 @@ instance Arbitrary Payment where
                       <*> arbitrary
                       <*> arbitrary
                       <*> arbitrary
+                      <*> arbitrary
+
+type NewPayment = Payment
```
adinapoli-iohk: Unnecessary newtype.

wallet-new/src/Cardano/Wallet/API/V1/Transactions.hs
akegalj: Hm, Interesting. I guess it would make code more readable.

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
sectore: There is no counter part `Migrate (V0.CId V0.Addr, Core.Coin) V1.PaymentDistribution`. Note `Core.Coin` vs. `V0.CCoin`

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
sectore: If I remember you correctly we don't want to a `Migrate` counter part for testing only. Because currently there is no need to have `Migrate V0.V0.CAccountId ((V1.WalletId, V1.AccountId)`

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
akegalj: @sectore it would be nice to add counterpart so that we can write proper roundrip tests (this would test did we actually write v0 -> v1 correctly). 

Without counterpart we are only hoping v0 -> v1 is done correctly

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
akegalj: same as here https://github.com/input-output-hk/cardano-sl/pull/2107#discussion_r156997007

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
adinapoli-iohk: No strong feelings here, i'm mentioning this as @akegalj was very keen on having counterparts written for the sake of testing, but I'm fine with sticking with the status quo 😉

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
sectore: After talking to @akegalj and @adinapoli-iohk we've decided to add `Migration` counterparts later.

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
sectore: Just for the record https://github.com/input-output-hk/cardano-sl/pull/2107#discussion_r157008005

adinapoli-iohk:
Tested this on my machine. Will pick any outstanding CI problems as part of subsequent PRs, to unclog our PR backlog.

PR #2109, opened 2017-12-13
[CSL-2070] Merge cardano-sl-1.0 to master (Dec 12)

hithroc:


client/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -286,19 +331,21 @@ makeManyUtxoTo1Params numFrom amountEachFrom amountTo = do
...
 makeManyAddressesToManyParams
-    :: Int
+    :: InputSelectionPolicy
+    -> Int
     -> Integer
     -> Int
     -> Integer
     -> Gen CreateMTxParams
-makeManyAddressesToManyParams numFrom amountEachFrom numTo amountEachTo = do
+makeManyAddressesToManyParams inputSelectionPolicy numFrom amountEachFrom numTo amountEachTo = do
```
Martoon-00: Since `Gen` appeared here I think it would be nice to take arbitrary `InputSelectionPolicy` where possible
(but this PR seems inappropriate to do such things)

gromakovsky:
@hithroc hlint fails, please fix.

PR #2115, opened 2017-12-14
[CSL-2048] Add phantom type to BackupPhrase + fix documentation

akegalj:
This issue extends straightens `BackupPhrase` with a phantom type. We are using backup phrase for:
 * paper vend - 9 words
 * wallet backup - 12 words
 * paper wallets - 16 words ?
 * maybe more usecases in future

We are adding phantom type to differentiate between those.

Pros:
 * cleaner type signatures
      - it is clear now both in types and documentation that wallet creation requires 12 words but paper vend requires 9 words
 *  swagger documentation is able to generate right instance of 9 words vs 12 words based on this phantom type
 * smaller space for errors in future to mix those up

Cons:
 * extra language extensions (although not dangerous) in single module where `BackupPhrase` is defined
 * some boilerplate with extra class instances
     - confusion http://lpaste.net/360780

Accidentally, this PR also removes potentially wired/dangerous `SafeCopy` instance of `BackupPhrase` - we should never be able to save `BackupPhrase` on disk!

## TODO
 * [ ] ~fix boilerplate http://lpaste.net/360780~
        - tried with GADTs, failed - giving up
 * [x] fix `Arbitrary (BackupPhrase a)` to be bip39 compatible

### DIFF
--- a/wallet/src/Pos/Util/BackupPhrase.hs
+++ b/wallet/src/Pos/Util/BackupPhrase.hs
@@ -1,7 +1,12 @@
+{-# LANGUAGE DataKinds      #-}
+{-# LANGUAGE KindSignatures #-}
 -- | Module providing restoring from backup phrase functionality
 
 module Pos.Util.BackupPhrase
        ( BackupPhrase(..)
+       , MnemonicType(..)
+       , BackupPhrasePaperVend
+       , BackupPhraseNormal
        , backupPhraseWordsNum
        , toSeed
        , keysFromPhrase
@@ -13,7 +18,7 @@ import           Universum
 
 import           Crypto.Hash (Blake2b_256)
 import           Data.Text.Buildable (Buildable (..))
-import           Test.QuickCheck (Arbitrary (..), elements, genericShrink, vectorOf)
+import           Test.QuickCheck (Arbitrary (..), elements, vectorOf)
 import           Test.QuickCheck.Instances ()
 
 import           Pos.Binary (Bi (..), serialize')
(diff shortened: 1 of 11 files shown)


wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -22,18 +27,30 @@ import           Pos.Crypto (AbstractHash, EncryptedSecretKey, PassPhrase, Secre
...
-newtype BackupPhrase = BackupPhrase
+newtype BackupPhrase (a :: MnemonicType) = BackupPhrase
     { bpToList :: [Text]
     } deriving (Eq, Generic)
 
-instance Bi BackupPhrase where
+type BackupPhrasePaperVend = BackupPhrase 'PaperVendMnemonic
+type BackupPhraseNormal = BackupPhrase 'BackupMnemonic
+
+instance Typeable a => Bi (BackupPhrase a) where
```
akegalj: This is not dangerous as `Bi` is using `Typeable` only for error reporting

wallet/src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -624,7 +623,6 @@ deriveSafeCopySimple 0 'base ''CHash
 deriveSafeCopySimple 0 'base ''CId
 deriveSafeCopySimple 0 'base ''Wal
 deriveSafeCopySimple 0 'base ''Addr
-deriveSafeCopySimple 0 'base ''BackupPhrase
```
akegalj: This was odd!

akegalj:
I have tried to debug it for a quite some time and I give up. I don't get it.

Anyway, everything works except `BackupPhrase` in `CWalletInit`(for creating a new wallet in v0) every time produces 9 word mnemonic (except 12 word mnemonic). 

It should be correct, I have tested `Arbitrary` instances of `BackupPhrase` in the repl and they work correctly. But the result is bellow:

look

It would be great if someone could review because I can't figure what's going on here.

PR #2116, opened 2017-12-14
[CSL-2016] Sorting and Filtering API

adinapoli-iohk:
This PR delivers a type-safe API to do sorting & filtering for `wallet-new`, based on Servant & IxSet.
The idea is to bake the sorting & filtering information directly at the type level, and reify this into the value level creating also a correspondence between keys & _ixset indices_:

screen shot 2017-12-14 at 14 30 22

In a nutshell, once we can prove to the compiler that we can go from `[Symbol]` to `Indices resource` we can build a list of filter & sort operations in a type safe way, and offload the filtering & sorting to IxSet. Such logic is contained in the `Cardano.Wallet.API.Response.Filter.IxSet` and `Cardano.Wallet.API.Response.Sort.IxSet` modules. There is also a module called `Cardano.Wallet.API.Response.Filter.Legacy` which has been left for educational purposes (to show how we can abstract away from IxSet per se) but I think we should take the plunge and start using the `IxSet` interface straight away.

Having all of this at the type level allows us to display these filter & sort operations _automatically_ in the Swagger spec, like so:

screen shot 2017-12-14 at 14 29 05

There is also a little query language provided, which is documented here:

screen shot 2017-12-14 at 14 44 02

**Demo time!**

filtering_and_sorting


wallet-new/src/Cardano/Wallet/API/Request/Sort.hs
```diff
@@ -0,0 +1,146 @@
...
+      let newOp = SortIdentity
+      in SortOp (newOp :: SortOperation ix a) (toSortOperations req [] (Proxy :: Proxy ixs))
+  toSortOperations req (key:xs) _ =
+      case List.lookup "sort_by" (parseQueryText $ rawQueryString req) of
+          Nothing       ->
+              let newOp = SortIdentity
+              in SortOp (newOp :: SortOperation ix a) (toSortOperations req xs (Proxy @ ixs))
+          Just Nothing  ->
+              let newOp = SortIdentity
+              in SortOp (newOp :: SortOperation ix a) (toSortOperations req xs (Proxy @ ixs))
```
akegalj: there is some redundacy here with L99

adinapoli-iohk: This part was copied and adapted from the `servant` codebase -- open a PR upstream! 😁 :trollface:

PR #2120, opened 2017-12-14
[CSL-2035] Upgrade dependencies

volhovm:
Upgraded universum and log-warper.

### DIFF
--- a/auxx/src/AuxxOptions.hs
+++ b/auxx/src/AuxxOptions.hs
@@ -15,11 +15,12 @@ import           Universum
 import           Data.Version (showVersion)
 import qualified NeatInterpolation as N
 import           Options.Applicative (CommandFields, Mod, Parser, command, execParser, footerDoc,
-                                      fullDesc, header, help, helpDoc, helper, info, infoOption, long,
-                                      maybeReader, metavar, option, progDesc, subparser, value)
+                                      fullDesc, header, help, helpDoc, helper, info, infoOption,
+                                      long, maybeReader, metavar, option, progDesc, subparser,
+                                      value)
 import           Pos.Communication (NodeId)
 import           Serokell.Util.OptParse (strOption)
-import           Text.PrettyPrint.ANSI.Leijen (Doc, text, line)
+import           Text.PrettyPrint.ANSI.Leijen (Doc)
 
 import           Paths_cardano_sl (version)
 import qualified Pos.Client.CLI as CLI
@@ -35,7 +36,7 @@ data AuxxOptions = AuxxOptions
     , aoPeers          :: ![NodeId]
     -- ^ Peers with which we want to communicate
     --   TODO: we also have topology, so it can be redundant.
-    , aoStartMode           :: !AuxxStartMode
+    , aoStartMode      :: !AuxxStartMode
     }
 
 data AuxxStartMode
@@ -71,25 +72,30 @@ cmdParser = command "cmd" $ info opts desc
 -- Parse everything
 ----------------------------------------------------------------------------
 
+helpText :: Doc
+helpText = fromString @Doc $ toString @Text $ [N.text|
+    Auxx start mode. Can be 'light', 'with-config', 'with-node', 'auto'.
+    * 'light' mode requires no configuration to start, but the available.
+      command set is limited.
+    * 'with-node' mode will start auxx as a node plugin.
+    * 'with-config' mode will start auxx as a standalone application.
+    * 'auto' will try to load the configuration and start in 'with-config' mode,
+      if fails to do so will start in 'light' mode. (default: auto)
+    |]
+
 startModeParser :: Parser AuxxStartMode
-startModeParser = option startModeReader $
-    long "mode" <>
-    value Automatic <>
-    helpDoc (Just $
-           text "Auxx start mode. Can be 'light', 'with-config', 'with-node', 'auto'." <> line
-        <> text "* 'light' mode requires no configuration to start, but the available."
-        <> text " command set is limited." <> line
-        <> text "* 'with-node' mode will start auxx as a node plugin." <> line
-        <> text "* 'with-config' mode will start auxx as a standalone application." <> line
-        <> text "* 'auto' will try to load the configuration and start in 'with-config' mode,"
-        <> text " if fails to do so will start in 'light' mode. (default: auto)")
+startModeParser =
+    option startModeReader $
+    long "mode" <> value Automatic <> helpDoc (Just helpText)
   where
-    startModeReader = maybeReader (\case
-        "auto" -> Just Automatic
-        "light" -> Just $ Light
-        "with-config" -> Just $ WithConfiguration
-        "with-node" -> Just $ WithNode
-        _ -> Nothing)
+    startModeReader =
+        maybeReader $ \case
+            "auto" -> Just Automatic
+            "light" -> Just $ Light
+            "with-config" -> Just $ WithConfiguration
+            "with-node" -> Just $ WithNode
+            _ -> Nothing
+
 
 auxxOptionsParser :: Parser AuxxOptions
 auxxOptionsParser = do
--- a/auxx/src/AuxxOptions.hs
+++ b/auxx/src/AuxxOptions.hs
@@ -15,9 +15,8 @@ import           Universum
 import           Data.Version (showVersion)
 import qualified NeatInterpolation as N
 import           Options.Applicative (CommandFields, Mod, Parser, command, execParser, footerDoc,
-                                      fullDesc, header, help, helpDoc, helper, info, infoOption,
-                                      long, maybeReader, metavar, option, progDesc, subparser,
-                                      value)
+                                      fullDesc, header, help, helper, info, infoOption, long,
+                                      maybeReader, metavar, option, progDesc, subparser, value)
 import           Pos.Communication (NodeId)
 import           Serokell.Util.OptParse (strOption)
 import           Text.PrettyPrint.ANSI.Leijen (Doc)
@@ -72,8 +71,8 @@ cmdParser = command "cmd" $ info opts desc
 -- Parse everything
 ----------------------------------------------------------------------------
 
-helpText :: Doc
-helpText = fromString @Doc $ toString @Text $ [N.text|
+helpContent :: String
+helpContent = toString @Text $ [N.text|
     Auxx start mode. Can be 'light', 'with-config', 'with-node', 'auto'.
     * 'light' mode requires no configuration to start, but the available.
       command set is limited.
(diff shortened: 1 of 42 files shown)


auxx/src/AuxxOptions.hs
```diff
@@ -75,14 +76,18 @@ startModeParser :: Parser AuxxStartMode
...
+        , line
+        , text "* 'light' mode requires no configuration to start, but the available."
+        , text " command set is limited."
+        , line
+        , text "* 'with-node' mode will start auxx as a node plugin."
+        , line
+        , text "* 'with-config' mode will start auxx as a standalone application."
+        , line
+        , text "* 'auto' will try to load the configuration and start in 'with-config' mode,"
+        , text " if fails to do so will start in 'light' mode. (default: auto)" ])
```
ChShersh: This is not related to PR goal but personally I think that multiline strings should be written with some interpolation package, like, `neat-interpolation`. So you can just write:

```haskell
   help [text|
        Auxx start mode. Can be 'light', 'with-config', 'with-node', 'auto'.
        * 'light' mode requires no configuration to start, but the available.
          command set is limited.
        * 'with-node' mode will start auxx as a node plugin.
        * 'with-config' mode will start auxx as a standalone application.
        * 'auto' will try to load the configuration and start in 'with-config' mode,
          if fails to do so will start in 'light' mode. (default: auto)
        |]
```

block/src/Pos/Block/.#Worker.hs
```diff
@@ -0,0 +1 @@
+volhovm@keshet.6938:1513016755
```
ChShersh: Accidentally commited wrong file.

generator/src/Pos/Generator/Block/Logic.hs
```diff
@@ -80,7 +80,7 @@ genBlocks params inj = do
         let finishEOS = toEnum $ fromEnum tipEOS + fromIntegral numberOfBlocks
         foldM' genOneBlock mempty [startEOS .. finishEOS]
 
-    genOneBlock t eos = ((t <>) . inj) <$> genBlock eos
+    genOneBlock t eos = ((t `mappend`) . inj) <$> genBlock eos
```
ChShersh: Why `<>` is not used here? Is `t` some type from library for which we can't implement `Semigroup` instance?

auxx/src/Lang/DisplayError.hs
```diff
@@ -48,7 +48,7 @@ ppTypeError :: TypeError -> Doc
 ppTypeError TypeError{..} =
         "Couldn't match expected type" <+> (highlight $ ppTypeName teExpectedType)
     <+> "with actual value"            <+> (highlight $ show teActualValue) -- TODO: CSL-1814
-    <>  "!"
+    <+> "!"
```
hithroc: What is this change for? Why add a space before exclamation mark?

infra/Pos/Communication/Types/Protocol.hs
```diff
@@ -291,12 +297,15 @@ data MkListeners m = MkListeners
...
-instance Monad m => Monoid (MkListeners m) where
-    mempty = MkListeners (\_ _ -> []) mempty mempty
-    a `mappend` b = MkListeners act (inSpecs a `mappend` inSpecs b) (outSpecs a `mappend` outSpecs b)
+instance Monad m => Semigroup (MkListeners m) where
+    a <> b = MkListeners act (inSpecs a <> inSpecs b) (outSpecs a <> outSpecs b)
       where
         act vI pD = (++) (mkListeners a vI pD) (mkListeners b vI pD)
 
+instance Monad m => Monoid (MkListeners m) where
+    mempty = MkListeners (\_ _ -> []) mempty mempty
```
ChShersh: `(\_ _ -> [])` is just `mempty`. But some people might find lambda function more readable.

auxx/src/Lang/DisplayError.hs
```diff
@@ -84,7 +84,7 @@ ppEvalError (CommandNotSupported name) =
         "Command" <+> (squotes . highlight . nameToDoc) name
     <+> "is unavailable. Try" <+> (squotes . highlight) "help"
 ppEvalError (InvalidArguments name procError) =
-        "Invalid arguments for" <+> (squotes . highlight . nameToDoc) name <> ":"
+        "Invalid arguments for" <+> (squotes . highlight . nameToDoc) name <+> ":"
```
hithroc: Same goes for here.

lib/src/Pos/Launcher/Resource.hs
```diff
@@ -221,20 +221,22 @@ getRealLoggerConfig :: MonadIO m => LoggingParams -> m LoggerConfig
 getRealLoggerConfig LoggingParams{..} = do
     let cfgBuilder = productionB
                   <> showTidB
-                  <> maybe mempty prefixB lpHandlerPrefix
+                  <> maybe mempty logsDirB lpHandlerPrefix
```
ChShersh: `maybe mempty logsDirB lpHandlerPrefix` can be written as `maybeLogsDirB lpHandlerPrefix`. I introduced this `maybe`-builder just because it wasn't so convenient to use `maybe` for `Maybe` fields...

auxx/src/Lang/DisplayError.hs
```diff
@@ -98,7 +98,7 @@ renderFullLine str = renderLine 0 (length str) str
 ppParseError :: ParseError -> Doc
 ppParseError (ParseError str (Report {..})) =
       "Parse error at" <+> text (show span)
-  <$> "Unexpected" <+> text unconsumedDesc <> ", expected"
+  <$> "Unexpected " <+> text unconsumedDesc <+> ", expected"
```
hithroc: And here. Why the space after `Unexpected`? `<+>` concatenates with adding a space in-between, so in the end that part of the message will have two spaces. And why add space before the comma?

networking/src/Bench/Network/Commons.hs
```diff
@@ -84,19 +84,19 @@ logMeasure miEvent miId miPayload = do
...
+        ltSeverity ?= infoPlus
         commLogger
   where
-    commLogger = zoomLogger "comm" $ ltSeverity ?= Error
+    commLogger = zoomLogger "comm" $ ltSeverity ?= errorPlus
 
 loadLogConfig :: MonadIO m => Maybe FilePath -> Maybe FilePath -> m ()
 loadLogConfig logsPrefix configFile = do
-    let cfgBuilder = productionB <> (mempty { _lcFilePrefix = logsPrefix })
+    let cfgBuilder = productionB <> (mempty { _lcLogsDirectory = logsPrefix })
```
ChShersh: `mempty { _lcLogsDirectory = logsPrefix }` is `maybeLogsDirB logsPrefix`

networking/src/Network/Broadcast/OutboundQueue/Demo.hs
```diff
@@ -72,7 +72,7 @@ runEnqueue = M.runProduction . unEnqueue
 
 relayDemo :: IO ()
 relayDemo = do
-    updateGlobalLogger "*production*" (setLevel Notice)
+    updateGlobalLogger "*production*" (setLevel $ severityPlus Notice)
```
ChShersh: This is `noticePlus`.

lib/test/Test/Pos/Core/AddressSpec.hs
```diff
@@ -73,7 +73,7 @@ spec = describe "Address" $ do
 pkAndHdwAreShownDifferently :: Bool -> PublicKey -> Bool
 pkAndHdwAreShownDifferently isBootstrap pk =
     show (makePubKeyAddress (IsBootstrapEraAddr isBootstrap) pk) /=
-    (show @_ @Text (makePubKeyHdwAddress (IsBootstrapEraAddr isBootstrap)
+    (show @Text @_ (makePubKeyHdwAddress (IsBootstrapEraAddr isBootstrap)
```
gromakovsky: `@_` is sorta useless.

auxx/src/Lang/DisplayError.hs
volhovm: What? Is it a joke? 🤔 
I've replaced `<>` from semigroup with `<+>` which is something similar.

auxx/src/Lang/DisplayError.hs
volhovm: Oh, I got you. I thought `<>` is the same as `<+>` here.

block/src/Pos/Block/.#Worker.hs
volhovm: Oh wow, how did it even happen 🤔

infra/Pos/DHT/Real/Real.hs
```diff
@@ -17,6 +17,7 @@ module Pos.DHT.Real.Real
        ) where
 
 
+import           Nub (ordNub)
```
ChShersh: Sorry for inconvenience. This will be fixed in `1.0.0`.

auxx/src/Lang/DisplayError.hs
hithroc: No, they aren't.
`<$>` and `<+>` here are special Doc combinators.
https://hackage.haskell.org/package/ansi-wl-pprint-0.6.8.1/docs/Text-PrettyPrint-ANSI-Leijen.html#v:-60--43--62-

auxx/src/AuxxOptions.hs
```diff
@@ -71,25 +72,30 @@ cmdParser = command "cmd" $ info opts desc
...
-           text "Auxx start mode. Can be 'light', 'with-config', 'with-node', 'auto'." <> line
-        <> text "* 'light' mode requires no configuration to start, but the available."
-        <> text " command set is limited." <> line
-        <> text "* 'with-node' mode will start auxx as a node plugin." <> line
-        <> text "* 'with-config' mode will start auxx as a standalone application." <> line
-        <> text "* 'auto' will try to load the configuration and start in 'with-config' mode,"
-        <> text " if fails to do so will start in 'light' mode. (default: auto)")
+startModeParser =
+    option startModeReader $
+    long "mode" <> value Automatic <> helpDoc (Just helpText)
```
ChShersh: Because of the fact that you're using `neat-interpolation` now you don't need to use `helpDoc` function and convert `Text` to `Doc`. Just use [`help` function](https://www.stackage.org/haddock/lts-10.0/optparse-applicative-0.14.0.0/Options-Applicative.html#v:help). It will be just `<> help helpText` where `helpText :: String; helpText = toString [N.text|...|]`

lib/test/Test/Pos/Core/AddressSpec.hs
ChShersh: Yes, it's redundant.

PR #2121, opened 2017-12-15
[CSL-2038] More recovery heuristics

volhovm:
I've generalized `needRecovery` and related functions. I've also added a probabilistic preventive recovery triggering in case we're stuck >3 slots before current tip, but nobody sends us blocks.

### DIFF
--- a/block/src/Pos/Block/Worker.hs
+++ b/block/src/Pos/Block/Worker.hs
@@ -16,13 +16,13 @@ import           Formatting (Format, bprint, build, fixed, int, now, sformat, sh
 import           Mockable (delay, fork)
 import           Serokell.Util (enumerate, listJson, pairF, sec)
 import qualified System.Metrics.Label as Label
-import           System.Wlog (logDebug, logInfo, logWarning)
+import           System.Wlog (logDebug, logInfo, logNotice, logWarning)
 
 import           Pos.Block.BlockWorkMode (BlockWorkMode)
 import           Pos.Block.Configuration (networkDiameter)
 import           Pos.Block.Logic (calcChainQualityFixedTime, calcChainQualityM,
                                   calcOverallChainQuality, createGenesisBlockAndApply,
-                                  createMainBlockAndApply, needRecovery)
+                                  createMainBlockAndApply)
 import           Pos.Block.Network.Announce (announceBlock, announceBlockOuts)
 import           Pos.Block.Network.Logic (requestTipOuts, triggerRecovery)
 import           Pos.Block.Network.Retrieval (retrievalWorker)
@@ -47,7 +47,7 @@ import           Pos.Delegation.DB (getPskByIssuer)
 import           Pos.Delegation.Logic (getDlgTransPsk)
 import           Pos.Delegation.Types (ProxySKBlockInfo)
 import qualified Pos.Lrc.DB as LrcDB (getLeadersForEpoch)
-import           Pos.Recovery.Info (recoveryCommGuard)
+import           Pos.Recovery.Info (needTriggerRecovery, recoveryCommGuard)
 import           Pos.Reporting (MetricMonitor (..), MetricMonitorState, noReportMonitor,
                                 recordValue, reportOrLogE)
 import           Pos.Slotting (currentTimeSlotting, getSlotStartEmpatically)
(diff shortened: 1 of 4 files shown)


infra/Pos/Recovery/Info.hs
```diff
@@ -75,25 +76,37 @@ recoveryInProgress =
...
+                     "', because "%build) actionName status
+
+-- | This function checks that last known block is more than K slots
+-- away from the current slot, or current slot isn't known. It also
+-- returns False when we're actually doing recovery. So basically it
+-- returns true if we actually need to ask for tips right now.
+needTriggerRecovery :: (MonadRecoveryInfo m, HasConfiguration) => m Bool
+needTriggerRecovery = getSyncStatusK <&> \case
+    SSKindaSynced   -> False
+    SSDoingRecovery -> False
```
gromakovsky: What about `SSInFuture`? We likely don't need recovery in this case.

block/src/Pos/Block/Worker.hs
```diff
@@ -219,21 +221,34 @@ recoveryTriggerWorkerImpl SendActions{..} = do
...
-            delay $ sec 20
+
+        -- Sometimes we want to trigger recovery just in case. Maybe
+        -- we're just 5 slots late, but nobody wants to send us
+        -- blocks. It may happen sometimes, because nobody actually
+        -- guarantees that node will get updates on time. So we
+        -- sometimes ask for tips even if we're in relatively safe
+        -- situation.
+        (d :: Double) <- liftIO $ randomRIO (0,1)
+        unless (doTrigger && d < 0.005) $ do
```
gromakovsky: The probability to do it is 99.5%, is it intentional? Why not just 1?
Also you can use pattern matching for `doTrigger` instead of `when` followed by `unless`, but it's minor.

block/src/Pos/Block/Worker.hs
```diff
@@ -219,21 +221,34 @@ recoveryTriggerWorkerImpl SendActions{..} = do
...
+            getSyncStatus 3 >>= \case
```
gromakovsky: Maybe change 3 to 5? At least it will correspond to the example :)

volhovm:
@gromakovsky thanks, fixed.

infra/Pos/Recovery/Info.hs
gromakovsky: What do you think?

gromakovsky: Ah, ok, noticed that in the second commit. Perhaps it'd be better to amend this one instead, but whateva.

block/src/Pos/Block/Worker.hs
```diff
@@ -219,21 +221,33 @@ recoveryTriggerWorkerImpl SendActions{..} = do
...
-            delay $ sec 20
+
+        -- Sometimes we want to trigger recovery just in case. Maybe
+        -- we're just 5 slots late, but nobody wants to send us
+        -- blocks. It may happen sometimes, because nobody actually
+        -- guarantees that node will get updates on time. So we
+        -- sometimes ask for tips even if we're in relatively safe
+        -- situation.
+        (d :: Double) <- liftIO $ randomRIO (0,1)
+        when (not doTrigger && d < 0.005) $ getSyncStatus 5 >>= \case
```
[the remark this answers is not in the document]
volhovm: 0.005 = 1/200, so it will trigger every ~200th time. Worker kicks in every second, so it should be every 3rd minute. I might want to actually decrease it :)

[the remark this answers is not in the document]
volhovm: @gromakovsky check it pls.

volhovm:
@gromakovsky check

block/src/Pos/Block/Worker.hs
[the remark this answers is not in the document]
gromakovsky: > Worker kicks in every second

No, it doesn't. There is `delay $ sec 20`. So it's about once per hour. Once per 3 minutes would be ok.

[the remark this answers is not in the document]
gromakovsky: I'd prefer having it less than 5 minutes, because if I am an end user and I don't receive a transaction in 5 minutes, I will be worried. 5.5 is more or less ok.

block/src/Pos/Block/Worker.hs
```diff
@@ -219,21 +221,35 @@ recoveryTriggerWorkerImpl SendActions{..} = do
...
+        -- We don't want to ask for tips too frequently.
+        -- E.g. there may be a tip processing mistake so that we
+        -- never go into recovery even though we recieve
+        -- headers. Or it may happen that we will receive only
+        -- useless broken tips for some reason (attack?). This
+        -- will minimize risks and network load.
+        when (doTrigger || triggerSafety) $ delay $ sec 20
   where
     repeatOnInterval action = void $ do
         delay $ sec 1
```
gromakovsky: Below there is such line:
>  void $ action `catchAny` \e -> reportOrLogE "recoveryTriggerWorker" e

I think that if an exception is caught, we should wait for more than a second, because probably nothing will change in a second. Can you add some delay if you agree (since you are changing this code anyway)?

volhovm: Ok, sure.

PR #2127, opened 2017-12-15
[CSL-2057] Add the current GIT SHA and software version to swagger docs

akegalj:
Issue https://iohk.myjetbrains.com/youtrack/issue/CSL-2057

look

Tried pulling the text to the right with embedded html - didn't work out.


wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -406,6 +408,10 @@ This is the specification for the Cardano Wallet API, automatically generated
 as a [Swagger](https://swagger.io/) spec from the [Servant](http://haskell-servant.readthedocs.io/en/stable/) API
 of [Cardano](https://github.com/input-output-hk/cardano-sl).
 
+Git revision: **$deGitRevision**
```
akegalj: Tried pulling the text to the right with embedded html - didn't work out.

wallet-new/server/Main.hs
```diff
@@ -118,19 +118,21 @@ startEdgeNode WalletStartupOptions{..} = do
...
-    CLI.printFlags
-    logInfo "[Attention] Software is built with the wallet backend"
-    startEdgeNode cfg
+    cfg@WalletStartupOptions{..} <- getWalletNodeOptions
+    putText "Wallet is starting..."
+    let loggingParams = CLI.loggingParams loggerName wsoNodeArgs
+        conf = CLI.configurationOptions $ CLI.commonArgs wsoNodeArgs
+    loggerBracket loggingParams . runProduction $ do
+        withConfigurations conf $ do
+            liftIO generateSwaggerDocumentation
```
akegalj: As a consequence we have to wait 2 secs longer until wallet starts for swagger doc to be written.

Alternatively I can do something like  : 
```haskell
loggerBracket loggingParams $ withConfigurationOptions conf generateSwaggerDocumentation
loggerBracket loggingParams $ runProduction $ ...
```
cons of the above is that we will parse configuration two times. 

thoughts ?

adinapoli-iohk: @akegalj What about generating the Swagger docs only if `RunMode` is `DebugMode` so that we avoid compromising performances? Then, once we have [this PR](https://github.com/input-output-hk/cardano-sl/pull/2068) fully merged, I think we can finally move away from generating the Swagger docs in `Main`. Would that work as an interim measure? 😉

akegalj: god point. Will fix

wallet-new/server/Main.hs
```diff
@@ -87,8 +89,10 @@ startEdgeNode :: HasCompileInfo
               -> Production ()
 startEdgeNode WalletStartupOptions{..} = do
   withConfigurations conf $ do
-    (sscParams, nodeParams) <- getParameters
-    actionWithWallet sscParams nodeParams wsoWalletBackendParams
+      when (isDebugMode $ walletRunMode wsoWalletBackendParams) $
+          generateSwaggerDocumentation
```
akegalj: Thanks. This is like it was before though. I wanted to move swagger doc out of Production (as running it takes second or two it seems)

But this would work

util/Pos/Util/CompileInfo.hs
```diff
@@ -52,7 +53,7 @@ withCompileInfo = give
 retrieveCompileTimeInfo :: TH.Q TH.Exp
 retrieveCompileTimeInfo = do
     cti <- TH.runIO $ do
-      ctiGitRevision <- fromString <$> retrieveGit
+      ctiGitRevision <- T.strip . fromString <$> retrieveGit
```
akegalj: Strip removes space characters, but not newline character (which is problem here)

adinapoli: @akegalj My experience is different with `strip` (I have used it no problem in other projects):

screen shot 2017-12-18 at 09 48 56

Am I missing something here?

Note that `Data.Text.strip` is implemented in terms of `isSpace` from `Data.Char`, which documentation states:

```
Returns True for any Unicode space character, and the control characters \t, \n, \r, \f, \v.
```

akegalj: @adinapoli ah, you are right. I didn't read documentation of `isSpace` and that newline is considered a space:
```haskell
Prelude> import Data.Char
Prelude Data.Char> isSpace '\n'
True
```

sorry for the confusion.

PR #2130, opened 2017-12-15
[CSL-1981] Do not encrypt poor nodes secret keys in genesis generation

hithroc:
Encrypting secret keys is a very slow operation, especially when we want to generate thousands of keys.

With these changes the poor nodes secret keys are not encrypted if HD adresses are not used.


gromakovsky:
CI failed.

tools/src/keygen/Dump.hs
```diff
@@ -99,7 +100,7 @@ dumpKeyfiles (dir, pat) dlgIssuers richs poors = do
         dumpDlgIssuerSecret (dlgIssuersDir </> patternize i) sk
     forM_ (enumerate richs) $ \(i, richSecrets) ->
         dumpRichSecrets (richDir </> patternize i) richSecrets
-    forM_ (enumerate poors) $ \(i, hdwSk) ->
+    forM_ (enumerate . map poorSecretToEncKey $ poors) $ \(i, hdwSk) ->
```
gromakovsky: I find it more natural to pass `PoorSecret` to `dumpPoorSecret` function and make this conversion there.

PR #2134, opened 2017-12-16
[CSL-2087] Improve logging of primary endpoints

Martoon-00:
This PR removes some huge logging of all addresses in `newPayment` and all transactions from blocks in `getHistory` (which is compensated with logging of last 100 transactions).

Also some logs added which report stage of endpoint execution, so that we can better understand timings of all stages.

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Methods/History.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/History.hs
@@ -17,8 +17,8 @@ import qualified Data.Map.Strict            as Map
 import qualified Data.Set                   as S
 import           Data.Time.Clock.POSIX      (POSIXTime, getPOSIXTime)
 import           Formatting                 (build, sformat, stext, (%))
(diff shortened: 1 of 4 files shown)


wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -123,7 +133,11 @@ getHistoryLimited mCWalId mAccId mAddrId mSkip mLimit = do
             pure (cWalId', accIds')
         (Nothing, Just accId)   -> pure (aiWId accId, [accId])
     (unsortedThs, n) <- getHistory cWalId accIds mAddrId
-    let sortedTxh = sortByTime (Map.elems unsortedThs)
+
+    let !sortedTxh = sortByTime (Map.elems unsortedThs)
+    logDebug "getHistoryLimited: sorted transactions"
```
Martoon-00: I realize that this is not correct since I don't force full list evaluation, fixing..

PR #2138, opened 2017-12-17
cardano-sl: Fix broken cardano-bench-criterion target

erikd:
This benchmark target wouldn't compile due to changes elsewhere in
the tree.

### DIFF
--- a/lib/bench/Bench/Pos/Criterion/FollowTheSatoshiBench.hs
+++ b/lib/bench/Bench/Pos/Criterion/FollowTheSatoshiBench.hs
@@ -1,17 +1,17 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
 module Bench.Pos.Criterion.FollowTheSatoshiBench
     ( runBenchmark
     ) where
 
 import           Criterion.Main (Benchmark, bench, defaultConfig, defaultMainWith, env, whnf)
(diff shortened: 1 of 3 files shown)


lib/bench/Bench/Pos/Criterion/FollowTheSatoshiBench.hs
```diff
@@ -20,8 +20,11 @@ type UtxoSize = Int
 
 -- [CSL-192]: make a trick and use special unsafe arbitrary instances
 -- for generation of such things
-arbitraryUtxoOfSize :: UtxoSize -> Gen Utxo
-arbitraryUtxoOfSize n = fromList . take n <$> infiniteListOf arbitraryUnsafe
+arbitraryUtxoOfSize :: UtxoSize -> Gen [(StakeholderId, Coin)]
+arbitraryUtxoOfSize n = take n <$> infiniteListOf arbitraryUnsafe
+
+-- block/src/Pos/Lrc/FtsPure.hs:
+-- followTheSatoshi :: HasConfiguration => SharedSeed -> [(StakeholderId, Coin)] -> SlotLeaders
```
gromakovsky: Okay, I see why you made these changes, looks good. But in general there shouldn't be commented out code, so can you please remove it?

erikd:
Done!

PR #2142, opened 2017-12-18
[CSL-2104] Propose plan to migrate crypto schemes without breaking backward compatibility

adinapoli-iohk:
[Rendered](https://github.com/input-output-hk/cardano-sl/blob/37b54d6a8798b367897fb4da7329e88b6ec77b1e/docs/proposals/csl-2104-crypto-scheme-changes-in-address.md) 

This is the first revision of the proposal to modify how we encode & decode the `HDAddressPayload` to take into account needed changes to the crypto scheme, as per @vincenthz 's request.

@georgeee You will notice how I have not introduced a new attribute. I did this because when we spoke about this very same issue with @dcoutts he seemed to prefer to go down this route to avoid any increase in the address size which could cause problems for the short addresses workstream. Happy to discuss with you all contingency plans or amendments to the following proposals.

Let me know what you think!

### DIFF
--- a/docs/proposals/csl-2104-crypto-scheme-changes-in-address.md
+++ b/docs/proposals/csl-2104-crypto-scheme-changes-in-address.md
@@ -1,0 +1,270 @@
+<div id="table-of-contents">
+<h2>Table of Contents</h2>
+<div id="text-table-of-contents">
+<ul>
+<li><a href="#sec-1">1. Abstract</a></li>
+<li><a href="#sec-2">2. Motivation</a></li>
+<li><a href="#sec-3">3. Specification</a>
+<ul>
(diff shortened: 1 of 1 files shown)


georgeee:
Note, when you introduce new attribute, it should be formed as softfork protocol updated.

Otherwise, all looks good, it's basically an application of design provisioned for blockchain structures.

gromakovsky:
Does anybody mind if I merge it? I can do it bypassing CI, because it only adds a `.md` document.

adinapoli-iohk:
@gromakovsky oh, please do ❤️ I have meant to do this myself by I do not have such bypassing powers :)

PR #2143, opened 2017-12-18
add libiconv for osx

karknu:


domenkozar:
@karknu what error do you get otherwise?

domenkozar:
cc @adinapoli-iohk ever seen this one before?

adinapoli-iohk:
@domenkozar I have, but for a different project. Usually it can be solved by installing `iconv` via `brew`. That's peculiar though that I was never hit by this, as I install cardano via `stack build` 🤔

adinapoli-iohk:
> macOS build doesn't fail.

Ah, perhaps @karknu is building with `stack build --nix`? Anyway, I agree with Domen, this doesn't look like harming anybody 😉

karknu:
I'm simply running './scripts/build/cardano-sl.sh' from nix-shell as per https://github.com/input-output-hk/cardano-sl/blob/master/docs/how-to/build-cardano-sl-and-daedalus-from-source-code.md

neongreen:
^ The CI should pass after my commit.

dpwolfe:
I'm glad this was added. I ran into the same problem building from the nix-shell for the first time a couple days ago. This was after following the setup instructions with `v1.0.3` checked out. I fixed it by adding this line to my `.bash_profile` so that clang could find the iconv lib when linking:
```sh
export LIBRARY_PATH=$LIBRARY_PATH:/usr/lib
```

domenkozar:
@dpwolfe that might save the day, but another day it will bite you back. I suggest you remove the workaround going forward.

dpwolfe:
@domenkozar Thanks for the tip. I agree... done!

PR #2146, opened 2017-12-19
[CSL-1869] Improve block(header) server performance

avieth:
Block deserialization is skipped when sending.
Loading headers performance improved (slightly).

It's a real pain, but I've managed to verify the improvements locally.
Block generator in cardano-sl-1.0 generated a chain of poor quality!
I couldn't figure out how to get the system start time right so that
the test nodes wouldn't immediately jump into recovery mode.

Anyway, since the test node was always in recovery mode, I
noticed that the recovery mode block headers server also suffers
from the block deserialization inefficiency: it will load headers
"up" using `GS.loadHeadersUpWhile`, which in fact loads more than
just the header, it deserializes the block with undo. We can fix that
later, as our relays are probably not going to be in recovery mode (I
hope not!) so it won't affect us.

### DIFF
--- a/node/src/Pos/Binary/Communication.hs
+++ b/node/src/Pos/Binary/Communication.hs
@@ -2,9 +2,13 @@
 
 -- | Communication-related serialization -- messages mostly.
 
-module Pos.Binary.Communication () where
+module Pos.Binary.Communication
+    ( msgBlockPrefix
+    ) where
 
 import           Universum
+import qualified Data.ByteString                  as BS
+import qualified Codec.CBOR.Write                 as CBOR (toStrictByteString)
 
 import           Pos.Binary.Block                 ()
 import           Pos.Binary.Class                 (Bi (..), Cons (..), Field (..),
@@ -62,6 +66,12 @@ instance (HasConfiguration, SscHelpersClass ssc) => Bi (MsgBlock ssc) where
             0 -> MsgBlock <$> decode
             1 -> MsgNoBlock <$> decode
             t -> fail $ "MsgBlock wrong tag: " <> show t
+
+-- | Get an encoded MsgBlock from an encoded block.
+msgBlockPrefix :: BS.ByteString
+msgBlockPrefix = CBOR.toStrictByteString prefix
+  where
+    prefix = encodeListLen 2 <> encode (0 :: Word8)
 
 -- deriveSimpleBi is not happy with constructors without arguments
 -- "fake" deriving as per `MempoolMsg`.
--- a/node/src/Pos/DB/Block.hs
+++ b/node/src/Pos/DB/Block.hs
@@ -10,6 +10,8 @@ module Pos.DB.Block
        , blkGetUndo
        , blkGetBlund
 
+       , blkGetBlockBytes
+
        , deleteBlock
 
        , prepareBlockDB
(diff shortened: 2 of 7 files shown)


avieth:
It's quite difficult to get a good isolated profile of the improvements.

I'm going off of the log timestamps (I did a profiling run but oops, I used `-fprof-auto` so the files are full of noise). With this change, the get blocks request is handled and all blocks sent (2200 empty blocks, 1.4mb total) within a second. Prior to the change it took 4. This is over loopback. The get headers response still takes 4-5 seconds, because of what I explained in the pull request summary: when in recovery mode, the headers are retrieved *by way of the blocks*, whereas the headers retrieval within the get blocks listener goes by way of headers. That extra block decoding in the former is what kills the performance.

avieth:
I should mention as well that I noticed the client took a lot of time to verify the (empty) blocks, perhaps because there was something wrong with them (generator gave me a bad chain). We should think about getting some performance profiles for verification.

I just skimmed the code for this but it looks like verification could also do a lot of block deserialization: `DB.loadBlundsFromTipWhile`.

volhovm:
> it will load headers "up" using GS.loadHeadersUpWhile, which in fact loads more than just the header

It's partially (only one handler is optimized, maybe we can do more) solved here: https://github.com/input-output-hk/cardano-sl/pull/2090

node/src/Pos/DB/Block.hs
```diff
@@ -148,17 +150,17 @@ loadDataWhile
...
     -> m (NewestFirst [] a)
-loadDataWhile getter predicate start = NewestFirst <$> doIt start
+loadDataWhile getter predicate start = NewestFirst <$> doIt [] start
   where
-    doIt :: HeaderHash -> m [a]
-    doIt h
-        | h == genesisHash = pure []
+    doIt :: [a] -> HeaderHash -> m [a]
+    doIt !acc h
+        | h == genesisHash = pure (reverse acc)
```
volhovm: What happened here? Previously we were doing the right fold, now we're doing left fold and reverse. What for?

node/src/Pos/Binary/Communication.hs
```diff
@@ -63,6 +67,12 @@ instance (HasConfiguration, SscHelpersClass ssc) => Bi (MsgBlock ssc) where
             1 -> MsgNoBlock <$> decode
             t -> fail $ "MsgBlock wrong tag: " <> show t
 
+-- | Get an encoded MsgBlock from an encoded block.
```
volhovm: The comment is incorrect, you're just constructing the prefix here. You could transform it into function `AsBinary Block -> ByteString` or change the comment.

avieth: Right, stale comment.

node/src/Pos/DB/Block.hs
avieth: It performs better.

volhovm: How? Why do you think so?

avieth: I profiled it to verify, using it in a pure computation to get 10000 "block headers" (`Int`s). This changed dropped it from ~4ms to ~3ms. I don't know exactly why. I believe the issue was with the `fmap`s through the recursion. What we end up with is `fmap (d0 :) . fmap (d1 :) . fmap (d2 :) . ... . fmap (dn :) (pure [])`. Maybe inlining would help but it depends upon the choice of monad. And anyway, using the accumulator makes this tail recursive, from which I believe GHC can produce faster code.

gromakovsky:
Also please add issue id to your commit message.

avieth:
Updated the commit messages, and set time-warp-nt revision to use cardano-sl-1.0 branch.

I know, pkgs/default.nix must be updated, but whenever I run generate locally it hangs. So my strategy is to wait for travis to fail, and apply the patch it produces for me :)

avieth:
> LGTM

@dcoutts this means nothing without an official GitHub Approval ™

jmitchell:
Initial sanity checks on a new public relay that was intentionally kept out of the RR DNS pool are promising. Michael and Neil were able to send transactions back and forth, Michael starting with no local blockchain and Neil syncing from a relatively recent block.

We have just added that relay to the RR DNS to acquire more subscribers. After the connection count evens out with existing public relay averages we can start comparing metrics. Historically this has taken ~12h, but that was before the topology grew so much; it's only a rough estimate.

PR #2152, opened 2017-12-19
[CSL-1804] various Shelley networking design documents

avieth:
Copied from https://github.com/serokell/time-warp-nt/pull/113


networking/notes/shelley/BlockRetrievalRecovery.md
```diff
@@ -0,0 +1,126 @@
...
+
+This is fine, all we're doing is changing the source of the block headers.
+Instead of coming from an explicit time-warp conversation, they'll come
+from the diffusion layer (which uses a time-warp conversation, but that's
+abstracted). We can get rid of `handleUnsolicitedHeaders` because all
+we need is `handleUnsolicitedHeader`, which can be called whenever the
+diffusion layer presents a block header. `handleUnsolicitedHeader` requires
+a `NodeId` of the provenance of the block, but that's only there so it
+knows who to ask for the body, so it should be removed because it's the
+diffusion layer's responsibility to know where to find the body.
```
avieth: @dcoutts said

> And are we confident that the block cache layer can indeed do this. Will it need to track where it knows it can fetch new blocks from?

@avieth said

> Yes I imagine that's how it would work. Included in the cache would be a set of NodeIds for peers which have announced that header.

networking/notes/shelley/BlockRetrievalRecovery.md
```diff
@@ -0,0 +1,126 @@
...
+the retrieval worker.
+
+## One improvement worth mentioning
+
+RE `handleRecovery` and `getProcessBlocks`: the former requests the headers,
+and the latter requests the blocks for those headers. It's probably better to
+just request the blocks. Should we make that improvement now, or later? If
+we defer it, we'll need to put a `getHeaders` into the synchronous diffusion
+layer interface (only a similar one for `getBlocks` is there, and a block
+contains its header).
```
avieth: @dcoutts said

> If we can do it all at the same time without causing too much chaos, that'd be preferable.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+    If in state 2, reject the offer from the publisher, remain in state 2.
+    If in state 3, reject the offer from the publisher, remain in state 3.
+  - `v` is published to us
+    If in state 1, remain in state 1. This should not happen.
+    If in state 2, remember `v` and relay `k` and `v` to subscribers, enter
+      state 3.
+    If in state 3, remain in state 3. This should not happen.
+  - TTL on the key expires.
+    If in state 1, this makes no sense, since we don't have the key.
+    If in state 2, forget `v` and enter state 1.
```
avieth: @coot said

> I think it should be "forget `k`", `v` is unknown at this state.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+      This is dodgy. It means the publisher failed to deliver before TTL.
+    If in state 3, forget `k` and `v` and enter state 1.
```
avieth: @coot said

> We could also relay `k` not yet knowing `v` and then queue the requests for `v` until we know it. But this would complicate matters when we fail to get `v`, so it's better to start with the proposed machine.
> 
> There should be a fallback when the request for `v` fails, anyway.  So maybe we should have three states for `v`: `unknown`, `failed` and `known`?

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Stretch: privacy considerations
+
+There are known deanonymization schemes effective against cardano (and other
+cryptocurrencies including bitcoin) which exploit predictable patterns of
+broadcast. It's not a requirement for Shelley, but eventually we may want to
+use a mitigating broadcast scheme. Such a solution could include bouncing
+new transactions through other edge nodes, by way of an intermediate core node.
```
avieth: @coot said

> Do you foresee something like cardano on top of tor-like network?

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+  - It already has a notion of subscription, but it's topicless (or rather,
+    there's one topic called "everything"), so we'll have to introduce topics.
+    Each topic subscription is controlled by its own conversation, just as the
+    "everything" subscription is at the moment.
+    This way, the subscription protocol can remain mostly as is. The subscriber
+    sends an empty subscribe message and periodically sends empty keepalive
+    messages if the channel has been quiet for long enough.
+    The publisher will send messages (start conversations) of a type determined
+    by the topic, to every subscriber of the topic.
```
avieth: @dcoutts said

> It's not obvious to me if we want to independently subscribe to each topic we are interested in, or to do one subscription that says in the opening message which topics we are interested in. We should consider both I think and pick the best.
> 
> Note also that we are not bound by the previous approach of having a subscription conversation that simply sits idle while publish events take place completely independently as separate individual conversations. If publish messages are small (individual, block headers, transactions, mpc etc) then its plausible to have these messages arrive as part of the subscription conversation itself. Indeed I think that would be the more traditional design. It's an approach we should consider.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+    We'll get rid of this and replace it with a simple policy in which every
+    message is delivered to every subscriber for the relevant topic.
+    The `Peers` type will be simplified, since there's no longer a notion of
+    alternatives (`EnqueueOne` versus `EnqueueAll`), and renamed to
+    `Subscribers` to avoid confusion with the peer roster.
+
+  - Subscription buckets won't be necessary. These are used to organize a
+    set of mutable `Peers` values so that they can be modified concurrently
+    by independent users. With pub/sub there will only be one `Peers`
+    (`Subscribers`) value and one modifier: the subscription listener.
```
avieth: @dcoutts said

> So if we want to have non-kademlia based links, e.g. I have a number of nodes that I want to talk to each other directly, how will I do it? I think the answer is easy actually: we tell nodes by static configuration to also subscribe to these other nodes. So then yes, on the side responding to subscribers there is still only one set of subscribers. It's just on the side deciding who to subscribe to that we can add in extra nodes we want to subscribe to.
> 
> So I think this is fine, but we should mention this, that we will need private peering arrangements, but this can work by subscribing to additional peers, and there's no need to make bucket distinctions when looking at the set of subscribers.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+  - Rate limiting, precedence, max ahead will probably still be useful, both
+    for publishing and for requesting.
+
+  - Enqueue and dequeue policies aren't useful as is, because there is no
+    `NodeType`.
```
avieth: @dcoutts said

> Hmm, doesn't it just mean there's only one policy rather than lots? And actually we'll have to think about that carefully, whether we will still need some distinctions. For example a relay on the edge of the core DIF, will it not need to treat the code node it knows about differently from other nodes?

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+  - Failure policy isn't relevant for for pub/sub. If the peer fails then the
+    subscription is gone and won't return unless the subscriber tries again.
+    We'll probably want to keep the notion of a failure policy around, because
+    it is relevant for request/reponse. However, it won't be a part of pub/sub,
+    but rather the QoS/delta-Q system and peer roster, where failures will
+    lower the estimate and eventually cause the peer to fall out of favour.
```
avieth: @dcoutts said

> Hmm, I'll have to think about that. Not obvious to me yet.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+      detecting half-open TCP connections (intermittent lower-layer failures).
+
+### Publisher sends
+
+  - Announce key : indicate that a payload for this key is available.
+  - Payload value : give the payload for a key.
+
+Should the publisher acknowledge add/remove topic?
+If a subscriber adds a topic and then removes it, there may be announces for
+that topic inbound.
```
avieth: @coot said

> We could send them back with the ack message for remove, or if we don't want them, remove the subscriber only after the queue was cleared up to remove request. If we have all the requests in a FIFO this property will hold.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+## Questions, caveats, risks
+
+### Publish over subscription channel
+
+It would make a lot of sense to send the published data over the same
+conversation which is used to establish the subscription, rather than (as we
+do now) have one conversation control the lifetime of the connection, and
+starting other conversations to publish each datum.
+
+It's not clear whether this has any big advantages or disadvantages.
```
avieth: @dcoutts said

> Yes, ok, what I commented earlier, but you've said here. Fine.
>
> So yeah, to be decided.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+Decided: we'll use one time-warp conversation for the whole subscription.
+The subscriber can add/remove topics, ACK published keys, and send keepalive
+messages. The publisher can send keys and values.
+
+### Changes to time-warp
+
+The goal should be to do minimal changes to time-warp. It already enables
+what we need to do, but not in the most efficient or straightforward way.
```
avieth: @dcoutts said

> There will be a new requirement (eventually, not necessarily top priority) to be able to limit the bandwidth that all other peers can force this node into providing. Consider running a node at home without a firewall so that other nodes can contact you. That's fine so long as it does not consume all your home DSL upload bandwidth, otherwise you'd be annoyed. So we would need to be able to cap that. Timewarp does know (in principle) the amount of data sent to peers as responses to conversations, so it could in principle impose an aggregate bandwidth cap.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Topic filtering
+
+Should this be done at publisher or subscriber?
+Doing it at the publisher will reduce network traffic but the publisher will
+have to do more work. Doing it at the subscriber offloads the filtering work
+to the subscriber. Filtering won't be expensive (a pattern match) so it's
+probably best to do it at the publisher.
```
avieth: @dcoutts said

> Agree. Lets do it at the publisher side. The most expensive resource is the network traffic.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+Decided: do it publisher side.
+
+### Dead subscriptions due to half-open connections
+
+cardano-sl pull request
+[#1852](https://github.com/input-output-hk/cardano-sl/pull/1852) was made to fix
+[CSL-1676] and is relevant to pub/sub. We'll keep this solution or something
+like it in order to detect dead subscriptions.
```
avieth: @dcoutts said

> Yes we'll need something like this. It may become simpler by using the same subscription channel for the notification messages.

networking/notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Number of subscribers
+
+Each subscription will be served by one TCP connection, so the OS imposes a
+limit on the number of subscribers. It's plausible that even far below this
+limit performance will start to suffer. But it's also plausible that the number
+of subscriptions actually needed per node is low enough (not more than a
+hundred) to avoid any such problems.
+There's already a maximum subscribers configuration in the outbound queue that
+can be reused.
```
avieth: @dcoutts said

> We may need something on the inbound side, at the network-transport-tcp level to reject incoming tcp connections, and/or at the time-warp layer to reject after doing the initial peer data handshake. It depends on what basis we need to reject.
> 
> Obviously there's a cost per tcp connection (not just in the app but also the kernel) and rejecting at this stage is cheaper. We will probably need something simple like an absolute limit at the network-transport-tcp level, so that we simply do not accept more connections once we hit a limit. But if we also want to filter on the basis of info then we would need to accept, do the peer data handshake and then close the connection if we are choosing to reject that peer. That would be implemented at the time-warp layer, but we would need a network-transport extension to drop an existing heavyweight connection to a peer.

@avieth said

> https://github.com/haskell-distributed/network-transport/issues/25
> https://github.com/haskell-distributed/network-transport/issues/17

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
+# Core DIF
+
+## Notes from meeting with Neil and Duncan November 8
+
+A picture
+https://docs.google.com/drawings/d/1IbclztwUlZs8HYbkLGVJ7vS9D_lVX_7Wu-S_M2SGsD0/edit
```
avieth: @dcoutts said

> Lets try and get an svg (or png) export of this and include it in the repo here.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+  - I am called (I am in top (say) 80% of stakeholders, or the top 100
+    stakeholders).
+    The plan seems to be to limit the actual number of members rather than
+    including top n% stakeholders, which could be arbitrarily-many.
+- Expectation: core nodes don't hold coins, but have stake delegated to them.
+  - Example: IOHK and partners would make a new address with no coins and
+    delegate all their stake to these.
+  - I guess the motivation is to protect the keys that control the coins; the
+    key that controls the delegate isn't so precious, since delegation can be
+    revoked.
```
avieth: @dcoutts said

> This turns out to be fine. All coins are always held on a different keypair than balance. But we will have a offline/onlike stake key solution. Either way it doesn't affect the network or core DIF I think. We can assume there's a single key per participant known from the blockchain and either that key can be used to sign messages, or more likely a key plus a lightweight delegation cert that shows that the presented key is a valid proxy.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+- To take delegation, you don't have to be on the public delegate list.
+- Currently there is no freedom from being delegated to you; you can't
+  deny delegation.
+- There are points at which a holder *can* join and *must* join.
+  Work this out with Lars...
```
avieth: @dcoutts said

> Yeah, we don't need to worry about the freedom from being delegated to here. That's the responsibility of the delegation workstream.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Routing; building spanning tree(s)
+
+- Publish a POA address (signed) to some others (encrypted with their
+  public keys). On the blockchain? Doesn't have to be the core node's IP
+  address; could be perhaps a bridge relay which will relay back to the core
+  node.
```
avieth: @dcoutts said

> How many you publish to needs to not be an individual choice but a rule based on the known size & membership of the core DIF. For size 100 we were thinking 10, but we need a rule that works for any size. How about a linear function, bounded below at 5 and above at 10.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+- Valency min and mix. Perhaps 3 and 10.
+- Each node should randomly choose 3 to 10 others. Perhaps XOR metric of
+  their public keys.
```
avieth: @dcoutts said

> I think XOR is the plan, it means everyone knows what the connections ought to be and it's symmetric.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+    Detectable? Yes, because you know the whole graph (just not the bearer
+    addresses).
+  - Perhaps recompute the graph by using a different salt for the hash on
+    public keys for XOR metric, until the graph is connected.
+  - Nodes can also publish extra routing that they choose themselves?
+    Addresse of relay nodes?
+  - Need an amount of time between knowing the stake distr of the next epoch
+    and before the network has to be established, in which we can do this
+    network computation.
+    TODO look into this. How long do we have?
```
avieth: @dcoutts said

> The stake distribution is chosen 2k slots before the end of the epoch, but that info is only guaranteed stable after k blocks, which is guaranteed within 2k slots, ie only worst case guaranteed stable at the end of the epoch.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+  choose a bunch of peers based on some XOR, then gather some data about their
+  delta-Q, then choose a subset according to percentiles of delta-Q/performance.
+  Difference here is that the computation of delta-Q is wholistic: we're
+  concerened with the entire path (which can be known here) rather than in the
+  non-core-DIF where we're concerned only with the first hop.
+
+- Some discussion about time constraints RE: ensuring that the slot leader's
+  block goes out earlier enough that it reaches the next slot leader before
+  it's too late.
+  Always going for worst case actually lowers best case throughput (i.e. tps).
```
avieth: @dcoutts said

> I think we concluded we should send out the block at the start of the slot, and not wait.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+  throughout the network.
+
+- Matrix of delta-Q's (100 x 100). How to fill it up? We know our own, we
+  propagate our own throughout the network (must propagate in order to
+  propagate, hm...)
+  I'm about to publish. I look at the matrix and I can compute paths which
+  with some certainty will get the data to everybody within some time. Of
+  course, it depends upon what the forwarders do.
+  Minimize maximum delay.
+  Also, due to paranoia, send to a few others with some random choice.
```
avieth: @dcoutts said

> We need to clarify what this algorithm is with Neil.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+performance. It's hard enough for a single computer to make time guarantees,
+let alone a network of them.
+
+## Questions after meeting with Neil and Duncan November 8
+
+1. Nodes in the core DIF never hold coins? Only stake? Why? Does this buy any
+   security at all? I suppose the machine itself is a target, and you wouldn't
+   want to keep the private key to the _coins_ on that machine. Keeping the
+   key to the delegate is fine, because the holder of the coins can revoke if
+   there's a compromise. Correct?
```
avieth: @dcoutts said

> Yes, only stake keys on these servers, and we'll almost certainly have an online/offline stake key solution so that nodes really use an online key and present a certificate that shows that it's a valid proxy for the known stake key.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+2. Ability to deny being delegated to? Is this a must have? I would hope so.
```
avieth: @dcoutts said

> Yes, but not for this workstream I think.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+3. What does sharing "on the blockchain" really mean? How does it work? I'm
+   very skeptical. Why not flood the (non-core-dif) network with your bearer
+   address encrypted with the target's public key?
```
avieth: @dcoutts said

> > What does sharing "on the blockchain" really mean? How does it work?
> 
> Its much like sending out an MPC message which gets incorporated into the blockchain. So it's just a special kind of blockchain entry.
> 
> > Why not flood the (non-core-dif) network with your bearer address encrypted with the target's public key?
> 
> It means we have evidence of non-compliance. We could do it by broadcast but then we have no objective evidence later that someone didn't do what they were supposed to. Perhaps for v1 that is ok since we have problems gathering evidence for any of the other bits either.

@georgeee said

> What kind of evidence are you talking about?
> 
> Anyway, also vote for broadcasting.
> When one wanna join Core DIF, he broadcasts message he wants to get addresses of DIF relay nodes and retrieves these encrypted in some way. Then through DIF relay nodes he can authenticate to finally join DIF. See no need, room or rationale to do it on chain.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+4. Is the core dif construction mechanism essentially the same as the bigger
+   network? Start with some XOR kademlia-style thing, then use delta-Q to
+   refine it. Only difference is the wholistic approach, in which the entire
+   weighted graph is known?
```
avieth: @dcoutts said

> I'm not aware of anything contradicting that. And if we can make them as similar as possible that is obviously good, we can reuse logic & code.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+5. It was taken for granted that a node will know all of the pairwise delta-Q
+   of the ~100 nodes in the core DIF. How? Surely we can't expect a consistent
+   view of this. What to do in case of missing information? I'm skeptical that
+   it will just work.
```
avieth: @dcoutts said

> This is part of the propagation algorithm that is yet to be fully defined. And it doesn't need to be fully consistent.

networking/notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+6. Suppose a core dif member uses 2 relay bridges. Is it OK to give the
+   address of one of these to some nodes, and the other to some other nodes,
+   w.r.t. the delta-Q measurements? I suspect so. Some will calculate the
+   delta-Q for one relay, some for the other, but it's all effectively the
+   delta-Q to the node behind the bridge relay.
+   I don't think 1 bearer address per public key is necessary. Indeed Neil
+   was quite excited about it *not* being necessary; you can give different
+   addresses to different peers.
```
avieth: @dcoutts said

> I think that's fine. The delta-Q is about the route from A to B.

networking/notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
+# QoS / delta-Q estimation
+
+Peer discovery gives us a pool of peers but no information about link quality.
```
avieth: @dcoutts said

> Except the first & last times we exchanged UDP packets (& we could if we want maintain more, e.g. round-trip timing).

networking/notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+To inform a choice of peers to actually communicate with, we'll need to gather
+data for an estimation of the delta-Q at a given peer.
+
+The esimate must be exposed (read-only) to the user of network-transport-tcp.
+It's not determined what exactly that data should be. The distribution itself?
+The s, g, and v components?
```
avieth: @dcoutts said

> To check with Neil. I'd guess the latter is most useful.

networking/notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+getMeasurements :: Transport -> EndPointAddress -> IO DeltaQStuff
+```
+
+## Implementation
+
+Must be implemented close to the socket itself, i.e. in network-transport-tcp.
+It will rely on precise time measurements on windows of data. New control
+codes would be defined in order to delimit these windows, i.e. indicate a
+leading and trailing edge. This allows for an estimation of the s component
+of delta-Q (latency as a function of the amount of data sent).
```
avieth: @dcoutts said

> We already preserve message boundaries on lightweight channels.

networking/notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+  - When trailing edge was sent
+  - When leading edge was received remotely (different clock, possibly lying)
+  - When trailing edge was received remotely (different clock, possibly lying)
+  - When leading edge was received locally
+  - When trailing edge was received locally
+
+To get a precise timestamp for a receive, we can use the SIOCGSTAMP ioctl.
+This is the timestamp of the last packet delivered to userspace. We would call
+it after receiving a leading or trailing edge, and before doing the next
+`recv`.
```
avieth: @dcoutts said

> It may be tricky to work out how these match up with the message boundaries. We'll have to read the docs on what this ioclt means exactly for tcp connections.

networking/notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+get the TCP round-trip-time estimate. Are we interested in that?
+
+## Relationship to input QDisc
+
+network-transport-tcp endpoints may have nontrivial queueing policies. Should
+latency induced by queueing factor into the delta-Q estimate? That's to say,
+are we trying to estimate only TCP round-trip-time, or network-transport-tcp
+round-trip-time? Neither option poses any extra challenge, it's only a matter
+of deciding when to respond to the control message: before or after the
+enqueue.
```
avieth: @dcoutts said

> Good Q. Another point to discuss with Neil.

networking/notes/shelley/Discovery.md
```diff
@@ -0,0 +1,128 @@
+# Peer discovery
+
+Existing Kademlia solution can be altered and replaced without disrupting
+cardano-sl as it stands today.
+
+Andrzej will look into the original froozen/kademlia, and serokell/kademlia
+fork, and come up with an implementation.
```
avieth: @dcoutts  said

> This is sprint planning, not design, we don't need to say who. Think about someone reading this in 6 months time to understand the design.

@dcoutts said

> This is a design doc, not a planning one, so we don't need to say who is doing what. Imagine someone reading this in 6 months time, trying to get up to speed with the design of the network layer.

networking/notes/shelley/Discovery.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Required feature: discovery without join.
+
+A peer which is not publicly addressable (behind NAT) must be able to
```
avieth: @dcoutts said

> Or firewall.

networking/notes/shelley/Discovery.md
```diff
@@ -0,0 +1,128 @@
...
+
+~~Our original 3-way handshake design gets us nowhere.
+The proposal was to try to send to the joiner's host at a different (known,
+fixed) port, and include the joiner's address in the index only if the joiner
+can show that it received the sent data on that port.
+It's feasible and not difficult to arrange for the peer's send to go through,
+even behind common NAT setups. The callback port is known, so the initiator
+can simply send some data to the peer on a UDP socket bound on that port
+before joining. Typical UDP hole punching NATs will admit the server's
+response.~~
```
avieth: @dcoutts said

> I don't think we need to say what the original proposal was. We can either present two solutions and say why we're picking the second, or we can simply present our chosen solution and say why.

@dcoutts said

> I think it's not necessary to say what our "original" design idea was here, or at least not necessary to say it was the original design. We can present two options and explain why we pick one, or just present our chosen one and justify the design.

networking/notes/shelley/Discovery.md
```diff
@@ -0,0 +1,128 @@
...
+
+Even though this is easy to fool, we decided that it's useful because it offers
+some sort of automatic discovery of public addressability. A node could become
+a supernode without any configuration.
+
+I (avieth) think this is silly. Anybody who doesn't know that their computer
+is reachable from the internet is in big trouble. Surely such a user would be
+capable of setting firewall rules, and therefore ought to have the technical
+capacity required to pass an extra command line argument to cardano-node
+indicating the public address.
```
avieth: @dcoutts said

> :-) You don't need to say this. Can simply state the assumption (the assumption you think is silly).

networking/notes/shelley/Discovery.md
```diff
@@ -0,0 +1,128 @@
...
+
+Nodes will not choose their own identifiers. Allowing this would open the
+door to easy eclipse attacks. Instead, peers derive the identifiers of their
+peers as a hash of their observed addresses.
+Relevant paper detailing some attacks:
+https://syssec.kaist.ac.kr/~yongdaek/doc/kad_attack_securecomm.pdf
```
avieth: @dcoutts said

> Given the suggested options in the paper, we should explain why we go for hash(IP) rather than hash (PK).

networking/notes/shelley/Discovery.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Details
```
avieth: @dcoutts said

> This is a mix of design and development plan, we should ideally separate these aspects.

networking/notes/shelley/Interface.hs
```diff
@@ -0,0 +1,83 @@
...
+
+data LogicLayer = LogicLayer
+  { logic :: Logic
+  , runLogicLayer :: IO ()
+  }
+
+data DiffusionLayer = DiffusionLayer
+  { diffusion :: Diffusion
+  , runDiffusionLayer :: IO ()
+  }
```
avieth: @dcoutts said

> Lets document for each of these what it is exactly. For example "the interface provided by the logic layer to the diffusion layer, for the diffusion layer to make calls into the logic layer".

networking/notes/shelley/Serialization.md
```diff
@@ -0,0 +1,141 @@
...
+first byte and if it's `0`, then the `Block` is the rest of the `ByteString`.
+GHC will not check these assumptions.
+
+## A safer but more complicated approach
+
+When decoding the network input, it should be possible to retain the
+`ByteString` from which the thing was decoded, and some sort of proof or
+evidence that this `ByteString` is a valid encoding. The typed API should
+accept either the serializable value, or the serialization and the proof that
+it's a valid encoding.
```
avieth: @dcoutts said

> I don't see that this solves the problem. Take the example above of 
> 
>     data MsgGetBlock = MsgBlock Block | MsgNoBlocks
> 
> What we receive is one of these messages, and we want to decode it to find out which constructor it is (and in other examples we may need to decide other bits of the message) but we do not want to decode the block.
> 
> So getting the bytestring of the whole thing still isn't useful, it's the bytestring of the block we want. So the simplest approach is to use the CBOR-in-CBOR trick and have the `MsgBlock` message's binary encoding contain a bytestring, which itself is a further encoded thing. So this gives us an envelope/payload style thing.
> 
> That can be typed with some wrapper like:
> 
>     -- | A thing of type @a@, but still in encoded form.
>     newtype Payload a = Payload ByteString
> 
> And this can have a binary/bi instance that read the thing as a bytestring but does not decode it. So then we'd use:
> 
>     data MsgGetBlock = MsgBlock (Payload Block) | MsgNoBlocks
> 
> This trick does change the CBOR encoding, so it is not directly backwards compatible, but assuming we use the normal protocol versioning then that's ok.

networking/notes/shelley/Serialization.md
```diff
@@ -0,0 +1,141 @@
...
+recv :: ConversationActions snd rcv m -> m (Serialized rcv)
+
+-- Constructors not exported.
+-- Can only be constructed by 'decode' or 'encode'.
+-- Relies on soundness of Serializable instance.
+data Serialized t where
+  Serialized :: Serializable t => ByteString -> Serialized t
+  Deserialized :: Serializable t => t -> Serialized t
+  Both :: Serializable t => ByteString -> t -> Serialized t
+
```
avieth: @kantp said

> I like this approach.

networking/notes/shelley/Serialization.md
```diff
@@ -0,0 +1,141 @@
...
+--
+decompose :: it -> SumOfProducts (Decomposed it)
+
+-- Reserialize, using any already-serialized components to avoid extra work.
+serialize :: SumOfProducts (Decomposed it) -> ByteString
+```
+
+This is just one approach I've been toying with. It's more complicated than
+I'd like but maybe it will inspire some sort of compromise or all around better
+solution.
```
avieth: @dcoutts said

> Yeah, lets discuss the various options.

avieth (commented):
copied comments from the old PR

georgeee:
@avieth can we please:
1. Move docs to `/docs/networking/shelley`
2. Merge this PR
?

dcoutts (approved):
Yep, lets merge and continue improving.

PR #2153, opened 2017-12-19
CSL-1156 find a solution for SscBlock UpdateBlock TxpBlock

v0d1ch:


txp/Pos/Txp/Settings/Global.hs
```diff
@@ -45,8 +47,22 @@ type TxpGlobalApplyMode ctx m =
 type TxpGlobalRollbackMode m = TxpCommonMode m
 
 -- [CSL-1156] Maybe find better approach (at least wrap into normal types).
-type TxpBlock = Either (Some IsGenesisHeader) (Some IsMainHeader, TxPayload)
-type TxpBlund = (TxpBlock, TxpUndo)
+-- type TxpBlock = Either (Some IsGenesisHeader) (Some IsMainHeader, TxPayload)
+-- type TxpBlund = (TxpBlock, TxpUndo)
+
+-- | Representation of 'Block' passed to a component.
+data ComponentBlock payload =
```
gromakovsky: This type should be in `cardano-sl-core`, because it's needed for `cardano-sl-update`, `cardano-sl-txp`, etc.
`Pos.Core.Block.Union.Types` is likely the right place.

txp/Pos/Txp/Settings/Global.hs
```diff
@@ -45,8 +47,22 @@ type TxpGlobalApplyMode ctx m =
...
+    ComponentBlockGenesis (Some IsGenesisHeader)
+    | ComponentBlockMain
+       { bcmHeader  :: !(Some IsMainHeader)
+       , bcmPayload :: !payload }
+
+-- deriving instance Show a => Show (ComponentBlock a)
+
+type TxpBlock = ComponentBlock TxPayload
+type UpdateBlock = ComponentBlock UpdatePayload
```
gromakovsky: `UpdateBlock` should be defined in `cardano-sl-update`.

txp/Pos/Txp/Settings/Global.hs
```diff
@@ -45,8 +47,22 @@ type TxpGlobalApplyMode ctx m =
...
+type UndoBlock = ComponentBlock TxpUndo
```
gromakovsky: No, it doesn't make sense. `Undo` is data which is needed to rollback a block. It's not payload of a block and not part of payload.

`Undo` is whole data necessary to rollback a block, it contains `TxpUndo`, `UpdateUndo` and some other parts. Txp component only needs `TxpUndo`. There is already a type for it, so you can just use it in `TxpBlund` definition (blund = block + undo, just our ad-hoc name).

txp/Pos/Txp/Settings/Global.hs
```diff
@@ -45,8 +47,22 @@ type TxpGlobalApplyMode ctx m =
...
+type TxpBlund = (TxpBlock, UndoBlock)
```
gromakovsky: So as I said, it should be `(TxpBlock, TxpUndo)`.

txp/Pos/Txp/Logic/Global.hs
```diff
@@ -134,5 +133,4 @@ runToilAction action = runDBToil . runToilTGlobal $ action
 
 -- Zip block's TxAuxes and corresponding TxUndos.
 blundToAuxNUndo :: TxpBlund -> [(TxAux, TxUndo)]
-blundToAuxNUndo (Left _, _)                = []
-blundToAuxNUndo (Right (_, payload), undo) = zip (flattenTxPayload payload) undo
+blundToAuxNUndo (payload, undo) = zip (flattenTxPayload (bcmPayload payload)) (bcmPayload undo)
```
gromakovsky: I think it's bad, because if payload is `ComponentBlockGenesis`, it will be runtime error, AFAIU. You need to pattern-match.

block/src/Pos/Block/Logic/Internal.hs
```diff
@@ -233,7 +233,13 @@ rollbackBlocksUnsafe bsc scb toRollback = do
 toTxpBlock
     :: HasConfiguration
     => Block -> TxpBlock
-toTxpBlock = bimap convertGenesis (convertMain mbTxPayload)
+toTxpBlock block =  case block of
```
gromakovsky: I know it's WIP, but I want to say that I expect you to get rid of some duplicated code here as well.
Specifically, you can define `toComponentBlock :: (MainBlock -> payload) -> Block -> ComponentBlock payload` and then define `toTxpBlock = toComponentBlock (view mainBlockTxPayload)`.

gromakovsky: And then `toUpdateBlock = toComponentBlock (view mainBlockUpdatePayload)`, etc.

v0d1ch: This signature kinda confuses me a bit `toComponentBlock :: (MainBlock -> payload) -> Block -> ComponentBlock payload` - why do I need `Block` in there ?

gromakovsky: Because `Block` contains the data you need 🤷‍♂️ 
Look at current version of `toTxpBlock`, it also takes `Block` as an argument.
Do you have something different in mind?

v0d1ch: I think I got you, I made everything compile up to a point but there are more places where `Either` was used instead this new `ComponentBlock` type so I am continuing my work

core/Pos/Core/Block/Union/Instances.hs
```diff
@@ -54,8 +59,73 @@ instance HasDifficulty BlockHeader where
...
+
+instance HasBlockVersion (ComponentBlock a) where
+    blockVersionL = blockVersionL
+
+----------------------------------------------------------------------------
+-- HasSoftwareVersion
+----------------------------------------------------------------------------
+
+instance HasSoftwareVersion (ComponentBlock a) where
+    softwareVersionL = softwareVersionL
```
gromakovsky: These definitions are wrong, they include infinite recursion.

core/Pos/Core/Block/Main/Chain.hs
```diff
@@ -33,6 +33,7 @@ import           Pos.Core.Txp (TxPayload, TxProof, mkTxProof)
 import           Pos.Core.Update (UpdatePayload, UpdateProof, mkUpdateProof)
 import           Pos.Crypto (Hash, PublicKey, hash)
 
+
```
gromakovsky: Please avoid whitespace changes if they are the only changes in a file and are not really important (this one is certainly not important).

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -439,10 +439,10 @@ dlgApplyBlocks dlgBlunds = do
                 SomeBatchOp $ map GS.DelPostedThisEpoch $
                 HS.toList duPrevEpochPosted
         pure $ edgeOp <> transCorrections <> postedOp
-    applyBlock ((Right block), _) = do
+    applyBlock ((ComponentBlockMain header block), _) = do
         -- for main blocks we can get psks directly from the block,
         -- though it's duplicated in the undo.
-        let proxySKs = getDlgPayload $ snd block
+        let proxySKs = getDlgPayload $ block
```
gromakovsky: `$` is redundant now.

ssc/Pos/Ssc/Toss/Logic.hs
```diff
@@ -90,7 +90,10 @@ applyGenesisBlock epoch = do
 
 -- | Rollback application of 'SscPayload's in 'Toss'. First argument is
 -- 'EpochOrSlot' of oldest block which is subject to rollback.
-rollbackSsc :: (HasConfiguration, MonadToss m) => EpochOrSlot -> NewestFirst [] SscPayload -> m ()
+rollbackSsc :: (HasConfiguration, MonadToss m) =>
+    EpochOrSlot
+    -> NewestFirst [] (ComponentBlock SscPayload)
```
gromakovsky: Why do you pass `Component SscPayload` to this function now? I think it should still be `SscPayload`.

txp/Pos/Txp/Logic/Global.hs
```diff
@@ -51,13 +52,12 @@ verifyBlocks
...
 verifyBlocks verifyAllIsKnown newChain = do
-    let epoch = NE.last (getOldestFirst newChain) ^. choosing epochIndexL (_1 . epochIndexL)
+    let epoch = NE.last (getOldestFirst newChain) ^. epochIndexL
     fst <$> runToilAction @_ @() (mapM (verifyDo epoch) newChain)
   where
     verifyDo epoch = verifyToil epoch verifyAllIsKnown . convertPayload
     convertPayload :: TxpBlock -> [TxAux]
-    convertPayload (Left _)             = []
-    convertPayload (Right (_, payload)) = flattenTxPayload payload
+    convertPayload payload = flattenTxPayload (bcmPayload payload)
```
gromakovsky: How does it work if `payload` is `ComponentBlockGenesis`? Will `bcmPayload` fail? You should pattern match here I suppose.

ssc/Pos/Ssc/Toss/Logic.hs
```diff
@@ -99,7 +102,7 @@ rollbackSsc oldestEOS (NewestFirst payloads)
         resetShares
     | otherwise = do
         setEpochOrSlot (pred oldestEOS)
-        mapM_ rollbackSscDo payloads
+        mapM_ (rollbackSscDo . bcmPayload) payloads
```
gromakovsky: `bcmPayload` is a partial function, it will fail here if payload is `ComponentBlockGenesis`.

core/Pos/Core/Block/Union/Types.hs
```diff
@@ -30,6 +33,15 @@ type BlockHeader = Either GenesisBlockHeader MainBlockHeader
...
 type Block = Either GenesisBlock MainBlock
 
+-- | Representation of 'Block' passed to a component.
+data ComponentBlock payload =
+    ComponentBlockGenesis (Some IsGenesisHeader)
+    | ComponentBlockMain
+       { bcmHeader  :: !(Some IsMainHeader)
+       , bcmPayload :: !payload }
+
+type UpdateBlock = ComponentBlock UpdatePayload
```
gromakovsky: `UpdateBlock` should be defined in `cardano-sl-update` package. For instance, in the same place as before.

ssc/Pos/Ssc/Logic/VAR.hs
```diff
@@ -224,7 +224,10 @@ sscRollbackU
 sscRollbackU blocks = tossToUpdate $ rollbackSsc oldestEOS payloads
   where
     oldestEOS = blocks ^. _Wrapped . _neLast . epochOrSlotG
-    payloads = over _Wrapped (map snd . rights . toList) blocks
+    payloads = NewestFirst $ map bcmPayload $ toList blocks
+    -- payloads = case toList blocks of
+    --     [ComponentBlockMain _ a]  -> NewestFirst $ map bcmPayload $ toList blocks
+    --     [ComponentBlockGenesis _] -> NewestFirst $ _ -- ???
```
v0d1ch: @gromakovsky this compiles but I know I need to pattern match in order not to blow on runtime. What is appropriate way to handle second case ? Is there a way to construct `SscPayload` from thin air ?

gromakovsky: You can write something like (I didn't type check it)
```
    payloads = NewestFirst $ mapMaybe extractPayload $ toList blocks
    extractPayload (ComponentBlockMain _ a) = Just a
    extractPayload (ComponentBlockGenesis _) = Nothing
```

v0d1ch: It compiles, thanks

core/Pos/Core/Block/Union/Instances.hs
```diff
@@ -58,4 +65,40 @@ instance HasDifficulty Block where
...
+
+instance HasPrevBlock (ComponentBlock a) where
+    prevBlockL = lens getter setter
+        where
+            getter (ComponentBlockGenesis genesisHeader) = genesisHeader ^. prevBlockL
+            getter (ComponentBlockMain mainHeader _)     = mainHeader ^. prevBlockL
+            setter (ComponentBlockGenesis genesisHeader) e =
+                ComponentBlockGenesis (genesisHeader & prevBlockL .~ e)
+            setter (ComponentBlockMain mainHeader payload) e =
+                ComponentBlockMain (mainHeader & prevBlockL .~ e) payload
```
gromakovsky: Hm, we use a bit different indentation for `where` clauses: https://github.com/serokell/serokell-util/blob/master/serokell-style.md#indentation

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -439,10 +439,10 @@ dlgApplyBlocks dlgBlunds = do
                 SomeBatchOp $ map GS.DelPostedThisEpoch $
                 HS.toList duPrevEpochPosted
         pure $ edgeOp <> transCorrections <> postedOp
-    applyBlock ((Right block), _) = do
+    applyBlock ((ComponentBlockMain header block), _) = do
```
gromakovsky: Can you rename `block` to `payload`, please (because it's actually payload, not whole block)?

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -472,11 +472,11 @@ dlgRollbackBlocks dlgBlunds = do
     getNewestFirst <$> mapM rollbackBlund dlgBlunds
   where
     rollbackBlund :: DlgBlund -> m SomeBatchOp
-    rollbackBlund (Left _, DlgUndo{..}) =
+    rollbackBlund (ComponentBlockGenesis _, DlgUndo{..}) =
         -- We should restore "this epoch posted" set to one from the undo
         pure $ SomeBatchOp $ map GS.AddPostedThisEpoch $ HS.toList duPrevEpochPosted
-    rollbackBlund (Right block, DlgUndo{..}) = do
-        let proxySKs = getDlgPayload $ snd block
+    rollbackBlund (ComponentBlockMain _ block, DlgUndo{..}) = do
```
gromakovsky: Here also please rename `block` to `payload`.

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -472,11 +472,11 @@ dlgRollbackBlocks dlgBlunds = do
...
+        let proxySKs = getDlgPayload $ block
```
gromakovsky: `$` is now redundant.

explorer/src/Pos/Explorer/Txp/Global.hs
```diff
@@ -65,16 +65,16 @@ applyBlund txpBlund = do
...
     let txpBlock = txpBlund ^. _1
     let slotId   = case txpBlock of
-            Left genesisBlock -> SlotId
+            ComponentBlockGenesis genesisBlock -> SlotId
                                   { siEpoch = genesisBlock ^. epochIndexL
                                   , siSlot  = minBound
                                   -- ^ Genesis block doesn't have a slot, set to minBound
                                   }
-            Right mainBlock   -> mainBlock ^. _1 . headerSlotL
+            ComponentBlockMain mainBlock _  -> mainBlock ^. headerSlotL
```
gromakovsky: AFAIU, `mainBlock` is now `Some IsMainHeader`, so I think it should be called `mainHeader`?

txp/Pos/Txp/Base.hs
```diff
@@ -68,8 +68,8 @@ txOutStake TxOut {..} =
 
 -- | Convert 'TxPayload' into a flat list of `TxAux`s.
 flattenTxPayload :: TxPayload -> [TxAux]
-flattenTxPayload UnsafeTxPayload {..} =
-    zipWith TxAux (toList _txpTxs) _txpWitnesses
+flattenTxPayload payload =
+    zipWith TxAux (toList (_txpTxs payload)) (_txpWitnesses payload)
```
gromakovsky: It looks like unnecessary change, but ok.

txp/Pos/Txp/Logic/Global.hs
```diff
@@ -51,13 +52,13 @@ verifyBlocks
...
-    let epoch = NE.last (getOldestFirst newChain) ^. choosing epochIndexL (_1 . epochIndexL)
+    let epoch = NE.last (getOldestFirst newChain) ^. epochIndexL
     fst <$> runToilAction @_ @() (mapM (verifyDo epoch) newChain)
   where
     verifyDo epoch = verifyToil epoch verifyAllIsKnown . convertPayload
     convertPayload :: TxpBlock -> [TxAux]
-    convertPayload (Left _)             = []
-    convertPayload (Right (_, payload)) = flattenTxPayload payload
+    convertPayload (ComponentBlockMain _ payload) = flattenTxPayload payload
+    convertPayload (ComponentBlockGenesis _ )     = flattenTxPayload $ UnsafeTxPayload MerkleEmpty []
```
gromakovsky: Ah, this is just `[]`, as it was before.

gromakovsky: I mean, `flattenTxPayload $ UnsafeTxPayload MerkleEmpty [] = []`.

PR #2156, opened 2017-12-20
wallet-new: Remove duplicate entry in cabal file

erikd:
The main file for this build target is called `Spec.hs` so the module `Spec` doesn't need to be listed in `other-modules`.

### DIFF
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -190,7 +190,6 @@ test-suite wallet-new-specs
   other-modules:    SwaggerSpec
                     APISpec
                     MarshallingSpec
-                    Spec
 
                     Cardano.Wallet.API.V0.Handlers
                     Cardano.Wallet.API.V1.Handlers


erikd:
Yeah, cabal 2.0 actually gives a build error with `Spec` listed in other modules.

I'm not using stack or Nix to build this, so that why I'm  probably the first to find this.

adinapoli-iohk:
> At my current employer we have a build tool called mafia which I'm rather keen on.

Ah, I think this cropped on /r/haskell a while ago and I have mentally bookmarked it but never explored it. Interesting name choice.. 😁

erikd:
Cabal -> mafia.

Nearly all of our project names are word plays like that :).

erikd:
@gromakovsky Should I submit PRs against develop rather than master?

gromakovsky:
@erikd yes, last night we adopted git flow, so now people should submit PRs against `develop` by default.

manpages:
BTW, regarding reviewers, sadly, @flyingleafe doesn't work with us on Cardano project anymore.

erikd:
@manpages I'm pretty sure that reviewer was added automatically,  possibly from entries in the `CODEREVIEWERS` file.

PR #2157, opened 2017-12-20
Introduce stylish haskell config

volhovm:
It's hard to modify content of cardano-sl-1.0 branch
since it doesn't have stylish-haskell config and by default
styler uses root config which is (already) very different
from what is used in cardano-sl-1.0 branch. So I add
explicit config having style we need.


volhovm:
It's not a logic change, so I will merge it without extra approval.

PR #2159, opened 2017-12-20
[CSL-2043] Fix block generator in cardano-sl-1.0

volhovm:
Yes, blockgen works in master already, but using it in this branch
still does make sense:
1. Switching branches (master/release) is expensive and time consuming.
2. Blockgen interface in cardano-sl-1.0 is somewhat more straight-forward.


volhovm:
Not related to logic, merging w/o extra approval.

PR #2162, opened 2017-12-21
[CSL-2063] Describe ways to establish trust in some snapshot

gromakovsky:


docs/proposals/fast-sync/trust.md
```diff
@@ -0,0 +1,499 @@
...
+blocks and at least one of them must be produced by an honest
+stakeholder according to Common Prefix property. It implies that at
+least one signature in the tiny block for epoch `e + 1` is produced by
+an honest stakeholder, so the whole list must be correct. Verification
+of subsequent tiny blocks is the same.
+
+If all tiny blocks are correct, headers received at step 8 are
+guaranteed to be created by real slot leaders. According to Common
+Prefix property, at least the oldest block must be created by an
+honest stakeholder, hence it must contain the hash of the correct
```
georgeee: Problem is it's not guaranteed by described algorithm that snapshot hash will appear in oldest block. By described algorithm we enforce only inclusion of leader list signatures

docs/proposals/fast-sync/trust.md
```diff
@@ -0,0 +1,499 @@
...
+  block. If some slots don't have blocks, it's probably not enough to
+  put only `k` signatures, but `2 · l` is not necessary. More profound
+  security analysis should be done to determine lower and upper
+  boundaries. In the best case, if `k` signatures is always enough and
+  the previous optimization is done, the size of a tiny block can be
+  about **674** kilobytes (**49** Mb per year).
+* Another option is to make `l` much smaller than `k`. It depends on
+  how much trust we want to have. `k = 2160` is quite a lot. It's much
+  greater than number of confirmations after a transaction is
+  considered stable. Probably having `l = 500` will be enough for
```
georgeee: Don't think we can go that far

docs/proposals/fast-sync/trust.md
```diff
@@ -0,0 +1,499 @@
...
+If we get rid of lightweight delegation (which might happen, but is
+not decided yet), the size of a tiny block in the worst case will be
+approximately 1.8 times smaller. It should be less than 10 megabytes
+of data to download for one-year-long blockchain if `l = 500`.
+
+All these optimizations are relatively easy to do, but some of them
+require more research or involve a certain trade-off. We can go even
+further with a bit more complicated optimizations.
+
+* Aggregate signature. Instead of putting `2 · l` or approximately `k`
```
georgeee: You didn't mention another straightforward optimization. Ed25519 allows you to have public keys, SIGs just 32 byte. We have 64kb just because implementation is slightly different. It sure can be adjusted

docs/proposals/fast-sync/trust.md
```diff
@@ -0,0 +1,499 @@
...
+
+* More compact serialization. Instead of using CBOR for everything, we
+  can serialize tiny blocks more compactly. Then a hash will take 28
+  bytes, a simple signature and a public key will take 64 bytes. It
+  can make a tiny block a bit smaller, approximately **1231**
+  kilobytes (**90** Mb per year).
+* Put less signatures into a tiny block. For instance, if all slots
+  are filled, it's enough to put first `k` signatures into a tiny
+  block. If some slots don't have blocks, it's probably not enough to
+  put only `k` signatures, but `2 · l` is not necessary. More profound
```
georgeee: Why `k` signatures might not be enough?

georgeee:
Also, I understood that we can easily have `l == k` and with no enforcement to include signature into block.
Both decreasing `l` and enforcement are potentially hazardous actions (in former, after you fast-sync you have formally speaking less secure block chain, in latter you decrease chain growth because some nodes on edge of `8k` slot won't be able to fulfill their obligations, which is possible even if `l` is much smaller).

It's possible that protocol analysis can be done to proof `k = 2100` is safe enough. But this is already a heavy ground.

What if we allow some epochs not to produce a valid tiny block? Say in that situation we won't be able to fast-sync on epoch `e`. Then we would need to download  snapshot for `e - 1`, which is last proven and then all blocks for `e`. This is nasty, right. But this way we'll stand on just exactly same security as is in Ouroboros.

Also, practically we don't expect such situation of tiny block not generated to ever occur in practice (i.e. there is high probability that more than `k` blocks will be generated in last `2k` slots and there will be `k` signatures).

Only question if we go with this suggestion is how we enforce block creators to compute leader list in advance and include signature into block I it's not enforced by block validation.
Also, how to enforce that somebody includes snapshot hash in one of early blocks from these `2k` last (this question is same as for current proposal)

docs/proposals/fast-sync/trust.md
gromakovsky: It depends on who we are targeting. If this mechanism is intended to be used only by some services like shops and exchanges and also end users, I don't see anything bad in having small `l`. These services most likely consider a transaction confirmed after less than `500` blocks. For instance, AFAIU, currently exchanges require about 20 confirmations. So they assume that it's impossible to revert more than 20 blocks. Maybe it will be increased in Shelley, but probably not beyond 500 blocks, because 500 blocks is almost 3 hours in the best case.
Full nodes which maintain the system will mostly stay online and they won't need this mechanism probably.
However, the problem is that we should choose `l` suitable for everybody, so even if 100500 services rely on 20 confirmations but there is a service which needs 1000 confirmations, we should either set `l = 1000` or require that service to use full syncing.

docs/proposals/fast-sync/trust.md
gromakovsky: I googled a bit and I think that signatures are 64B in standard Ed25519. Public keys are 32 bytes indeed, I will mention that.

gromakovsky:
>  Also, how to enforce that somebody includes snapshot hash in one of early blocks from these 2k last (this question is same as for current proposal)

What's the problem? Just enforce it by protocol rules. If you are concerned about the fact that it may take a lot of time and a slow node may not do it on time before time comes to create a block, then it's probably not a big problem, because I proposed the following:
> For example, in each epoch we can put snapshot corresponding to the state before `6 · k`-th slot into the first block after `8 · k`-th slot.

So in normal cases you'll have approximately a day to compute hash of snapshot, which should be enough even for slow nodes. By normal cases I mean those when a deep rollback doesn't occur. If rollback of 2000 blocks happens (which won't happen, but anyway), you'll have less time, but still quite a lot.

Moreover, we can require even older snapshot, e. g. from `4 · k`-th slot.

PR #2163, opened 2017-12-22
[CSL-1607] Use TemplateHaskell to get current OS at compile-time

rockbmb:


update/Pos/Update/Core/Types.hs
```diff
@@ -151,6 +154,9 @@ mkSystemTag tag | T.length tag > systemTagMaxLength
                 | otherwise
                     = pure $ SystemTag tag
 
+currentSystemTag :: SystemTag
+currentSystemTag = $(runQ [| SystemTag (T.pack (os ++ "_" ++ arch)) |])
```
georgeee: Very good. But let's have backwards-compatibility. I.e. systemTag `win64` should still remain `win64`. Not `minigw_x86_64`

BTW, can we log current system tag in compile-time?

domenkozar:
Yes! Very excited.

update/Pos/Update/Core/Types.hs
```diff
@@ -141,7 +144,7 @@ newtype SystemTag = SystemTag { getSystemTag :: Text }
   deriving (Eq, Ord, Show, Generic, Buildable, Hashable, Lift, Typeable)
 
 systemTagMaxLength :: Integral i => i
-systemTagMaxLength = 10
+systemTagMaxLength = 15
```
gromakovsky: No, revert it please. This change is basically a hardfork.
If you apply @georgeee's suggestion to retain backwards compatibility, this change won't be necessary (existing tags are less than 10 characters long).

rockbmb: Thanks for pointing that out, this would have been a costly distraction.

rockbmb:
@georgeee using hardcoded strings in Haskell isn't ideal, so I researched for a while and found this: https://downloads.haskell.org/~ghc/8.0.1/docs/html/libraries/Cabal-1.24.0.0/Distribution-System.html

This way we don't have to manipulate the strings directly and we won't have to rely on a particular implementation of Haskell doing one thing or another (in Windows's case, because there seems to be a confusing situation with `mingw32/cygwin32/win`).

Do you think this is worth the effort?

core/Pos/Core/Update/Types.hs
```diff
@@ -296,6 +299,23 @@ mkSystemTag tag | T.length tag > systemTagMaxLength
...
+-- compiled in.
+-- The @System.Info@ module from @base@ was used
+-- (https://hackage.haskell.org/package/base-4.10.1.0/docs/src/System.Info.html#os)
+-- but the @os@ and @arch@ values are @String@'s. For an exhaustive list of possible
+-- platforms, see
+-- https://downloads.haskell.org/~ghc/8.0.1/docs/html/libraries/Cabal-1.24.0.0/Distribution-System.html.
+currentSystemTag :: SystemTag
+currentSystemTag =
+    let f sys | sys == "mingw32" = "win"
+              | sys == "osx"     = "mac"
```
gromakovsky: It should be `macos`, not `mac`, see `configuration.yaml`.

core/Pos/Core/Update/Types.hs
```diff
@@ -296,6 +299,23 @@ mkSystemTag tag | T.length tag > systemTagMaxLength
...
+              | otherwise      = sys
+        g archt | archt == "x86_64" = "64"
+                | archt == "i386"   = "32"
+                | otherwise         = archt
+    in $(runQ [| SystemTag (toText (f os ++ g arch)) |])
```
gromakovsky: Currently we expect it to be `macos64` or `win64` or something for linux. I think it'd be good to add a test that `currentSystemTag` is one of these values. In this case we'll notice if it changes for some reason.

gromakovsky:
> Do you think this is worth the effort?

I think it's worth the effort.

update/Pos/Update/Core/Types.hs
rockbmb: @georgeee Only got around to this now. What do you mean compile-time, just a `putStrLn` or using `log-warper` here (which I don't think is feasible)?

gromakovsky: `putStrLn` should be fine. Is it doable?

core/Pos/Core/Update/Types.hs
```diff
@@ -296,6 +300,20 @@ mkSystemTag tag | T.length tag > systemTagMaxLength
...
+    Windows -> "win"
+    OSX     -> "macos"
+    Linux   -> "linux"
+    _       -> display sys
+
+archHelper :: Arch -> String
+archHelper archt = case archt of
+    I386    -> "32"
+    X86_64  -> "64"
+    _       -> display archt
```
gromakovsky: I think it's better to put these helpers into `Pos.Core.Update.Util`, not `Pos.Core.Update.Types`. `Types` is primarily for type definitions (unsurprisingly).

update/Pos/Update/Configuration.hs
```diff
@@ -70,5 +85,24 @@ lastKnownBlockVersion = ccLastKnownBlockVersion updateConfiguration
...
 
+-- | @SystemTag@ corresponding to the operating system/architecture pair the program was
+-- compiled in.
+-- The @Distribution.System@ module from @Cabal@ was used
+-- (https://hackage.haskell.org/package/Cabal-2.0.1.1/docs/Distribution-System.html)
+currentSystemTag :: SystemTag
+currentSystemTag =
+    $(do let s :: Either String SystemTag
+             s = mkSystemTag (toText (osHelper buildOS ++ archHelper buildArch))
+             color col t =
```
gromakovsky: Please use functions from `Serokell.Util.ANSI` to colorize text.

update/Pos/Update/Configuration.hs
```diff
@@ -9,22 +10,31 @@ module Pos.Update.Configuration
        , withUpdateConfiguration
 
        , ourAppName
+       , currentSystemTag
```
gromakovsky: I think it's better not to export `currentSystemTag`, because it doesn't consider `UpdateConfiguration`, so it shouldn't be used.

update/Pos/Update/Configuration.hs
rockbmb: @gromakovsky the last time I worked with this was in `log-warper` so I was unsure there were functions in `serokell-util`. I'll do that.

rockbmb:
@gromakovsky I spoke to @jmitchell about the CI failure, this is not a critical issue so I asked him to take a look only when possible. I am unsure why running `./pkgs/generate-sh` fails, but it seems to be unrelated to this PR.

PR #2164, opened 2017-12-22
Remove rounding from end users' public logs

gromakovsky:
This rounding sometimes makes debugging too hard and it's not important
with regards to privacy.
Note that currently used log-warper version doesn't support more granular
timestamps (1 second is maximal precision).


volhovm:
Since it's already approved by @dcoutts and gromak rebased, I'll merge this PR.

PR #2166, opened 2017-12-22
[DEVOPS-61] appveyor cache

domenkozar:
Depends on https://github.com/input-output-hk/iohk-ops/pull/221

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,9 @@ environment:
     STACK_ROOT: "c:\\s"
     STACK_WORK: ".w"
     WORK_DIR: "c:\\w"
-    BUCKET: appveyor-ci-cache              
+    CACHE_S3_VERSION: v0.1.1
+    AWS_REGION: us-west-1
+    S3_BUCKET: appveyor-ci-cache              
     AWS_ACCESS_KEY_ID: 
       secure: 0OtMCUmJ4V7fNRPvKjzYxSO19I200rPG8WRW/qn1QPk=
     AWS_SECRET_ACCESS_KEY:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,11 @@ environment:
     STACK_WORK: ".w"
     CACHE_DIR: "c:\\c"
     WORK_DIR: "c:\\w"
+    BUCKET: appveyor-ci-cache              
+    AWS_ACCESS_KEY_ID: 
+      secure: 0OtMCUmJ4V7fNRPvKjzYxSO19I200rPG8WRW/qn1QPk=
+    AWS_SECRET_ACCESS_KEY:
+      secure: EmPAMpga8gjXxPMfJSsdq2sof8SLwyga1rrYuZDlPoriMS1iEFLvCNLA70JqHSna
 
 init:
   - SET BRANCH_CACHE_DIR=%CACHE_DIR%\%APPVEYOR_REPO_BRANCH:/=_%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,13 +15,9 @@ environment:
     AWS_SECRET_ACCESS_KEY:
       secure: EmPAMpga8gjXxPMfJSsdq2sof8SLwyga1rrYuZDlPoriMS1iEFLvCNLA70JqHSna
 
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -8,49 +8,32 @@ environment:
     # Avoid long paths on Windows
     STACK_ROOT: "c:\\s"
     STACK_WORK: ".w"
-    CACHE_DIR: "c:\\c"
     WORK_DIR: "c:\\w"
+    CACHE_S3_VERSION: v0.1.0
+    AWS_REGION: us-west-1
+    BUCKET: appveyor-ci-cache
```
lehins: `cache-s3` expects this variable's name to be `S3_BUCKET` instead of `BUCKET`

appveyor.yml
```diff
@@ -8,49 +8,32 @@ environment:
     # Avoid long paths on Windows
     STACK_ROOT: "c:\\s"
     STACK_WORK: ".w"
-    CACHE_DIR: "c:\\c"
     WORK_DIR: "c:\\w"
+    CACHE_S3_VERSION: v0.1.0
```
lehins: There is newer release of `cache-s3`: `v0.1.1`, that has an additional feature `--public`, which has no effect cardano-sl, so leaving it at `v0.1.0` is fine as well.

lehins:
@domenkozar Besides the variable name this looks great.

lehins:
Changes in 416d95f look good!. :+1:

appveyor.yml
```diff
@@ -8,49 +8,32 @@ environment:
     # Avoid long paths on Windows
     STACK_ROOT: "c:\\s"
     STACK_WORK: ".w"
-    CACHE_DIR: "c:\\c"
     WORK_DIR: "c:\\w"
+    CACHE_S3_VERSION: v0.1.1
+    AWS_REGION: us-west-1
```
deepfire: @domenkozar, do we really want to use `us-west-1`?  I think something EU-based like `eu-central-1`, or alternatively `eu-west-1` (for older auth protocol version support, if needed) would make more sense.

appveyor.yml
```diff
@@ -8,49 +8,32 @@ environment:
...
+    S3_BUCKET: appveyor-ci-cache              
+    AWS_ACCESS_KEY_ID: 
+      secure: 0OtMCUmJ4V7fNRPvKjzYxSO19I200rPG8WRW/qn1QPk=
```
deepfire: @domenkozar, do we have it documented somewhere, what user we're using for accessing the cache bucket?

appveyor.yml
lehins: @deepfire It's better to place the bucket in the region closest to AppVeyor servers, so the cache can be uploaded/downloaded quicker. I don't know if @domenkozar decided on `us-west-1`, experimentally, but considering that AppVeyor is registered in Canada, I'd expect placing S3 bucket in Europe would be slower.

appveyor.yml
lehins: @deepfire Don't know about the documentation part, but the user is created by terraform here: https://github.com/input-output-hk/iohk-ops/pull/221/files#diff-fefc5952bb2ab240eaa7c2245608fc79R7 and it's default name will be `appveyor-ci-cache-user`

appveyor.yml
deepfire: Understood, makes sense, thanks!

appveyor.yml
deepfire: Thanks again!

deepfire:
@domenkozar, I've hit the "rebuild PR" button on both the PR and branch builds, and yet both came at 40+ minutes.

domenkozar:
@deepfire this won't take effect until we merge it. If you follow the logic through, it will never save state for a PR

deepfire:
Oh, so it's not meant to reduce latency for branch merging -- only latency to first results, I see.

lehins:
@deepfire it will reduce the latency for merging as well, except for PRs it will use cache from master, which is yet not available.

domenkozar:
I've changed this logic to also support `develop` and `release` branches.

PR #2169, opened 2017-12-23
[CSL-1025] Block storage spec

neongreen:
[Rendered](https://github.com/input-output-hk/cardano-sl/blob/0ea5efa3654f3173cae3df9e4709619fe71aa197/docs/proposals/block-storage.md)

### DIFF
--- a/docs/proposals/block-storage.md
+++ b/docs/proposals/block-storage.md
@@ -1,0 +1,76 @@
+# Optimized block storage
+
+## The problem
+
+Currently we store blocks in a very inefficient way (space-wise). This is
+tracked in [CSL-1025][]. The main concerns are:
+
+[CSL-1025]: https://iohk.myjetbrains.com/youtrack/issue/CSL-1025
+
+  * On most systems, files are stored in 4kB blocks, which means that any
+    file smaller than 4kB will take at least 4kB on the disk. Since our
+    blocks are very small right now, they take about **eight times** more
+    space than they would if they were all stored in one file.
+
+  * Our blocks contain a lot of repetitive data (e.g. hashes of empty
+    payload) and some data that can easily be derived from the rest of the
+    block (e.g. headers). By applying compression and removing the parts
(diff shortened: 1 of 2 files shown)


docs/proposals/block-storage.md
```diff
@@ -0,0 +1,76 @@
...
+database that we are using). Indeed, a long time ago we had all blocks
+stored in RocksDB – but we switched from that to storing each block in a
+different file. Storing blocks in RocksDB has several related disadvantages:
+
+  * If you want to back the database up you would have to copy the blocks as
+    well, while in reality the blocks can always be redownloaded and we
+    don't care much about losing them.
+
+  * A hard drive failure has a much bigger chance to corrupt the database,
+    if the database contains lots of data (blocks).
```
gromakovsky: Is it essential? How possible are hard drive failures? I mean, if it happens for one node once per year, we can just take database from another node and that's it (for example).

To summarize: I still don't have a good understanding of why we decided to stop storing blocks and undos in RocksDB 🤷‍♂️ . The only good reason I see is that Bitcoin stores blocks in files (but differently). But I don't know whether it's a good reason actually.

docs/proposals/block-storage.md
```diff
@@ -0,0 +1,76 @@
...
+
+So, here's a list of requirements for a good solution:
+
+  * We'd like to store blocks separately from the main database.
+
+  * We want fast lookup by header hash.
+
+  * We'd like to store them in batches (e.g. one batch = one epoch), so that
```
gromakovsky: I'd like to understand the importance of this requirement.

neongreen:
^ Addressed the comments

georgeee:
How many files is typical RocksDB database?

gromakovsky:
It's very hard to say, because it highly depends on options we use and of course on the size of the database. In short: we can configure it as we want.
What exactly do you want to know and why?

PR #2171, opened 2017-12-24
[CSL-2114] Update log configuration files for new log-warper version

hithroc:
Latest log-warper version uses a different structure for its configuration files.

Note: I removed `termSeverity` from all configuration files because the default log-warper behavior writes everything except errors to stdout and errors to stderr. If that needs to be changed, please comment on it, and I'll change that.

### DIFF
--- a/scripts/log-templates/log-config-greppable.yaml
+++ b/scripts/log-templates/log-config-greppable.yaml
@@ -1,3 +1,3 @@
 # This template is used to grep the output.
 
-termSeverity: Debug
+termSeveritiesOut: All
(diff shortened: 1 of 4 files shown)


scripts/log-templates/log-config-greppable.yaml
```diff
@@ -1,3 +1,3 @@
 # This template is used to grep the output.
 
-termSeverity: Debug
+termSeveritiesOut: Debug+
```
ChShersh: You can write just `termSeveritiesOut: All` to not forget update this config in future if we add more severities in `log-warper`.

PR #2175, opened 2017-12-26
stack.yaml: Drop directory extra-dep

erikd:
The `directory` extra dependency seems to have been added in commit 8d9a444f1a to make it build with GHC 7.4. Since the current stack resolver version is `lts-9.1` which specifies GHC 8.0, this seems completely un-ncessary.


erikd:
Seems stack has `directory == 1.3.0.0` and hence the `extra-dep` version is needed.

Closing this.

PR #2182, opened 2017-12-28
[CSL-2122] Allow to specify def tx inputs selection policy in config

Martoon-00:


node/src/Pos/Configuration.hs
```diff
@@ -76,11 +80,27 @@ data NodeConfiguration = NodeConfiguration
...
     , ccPendingTxResubmissionPeriod :: !Int
       -- ^ Minimal delay between pending transactions resubmission
+    , ccTxDefInputSelectionPolicy   :: !InputSelectionPolicyConf
+      -- ^ Default policy to pick inputs when forming transaction
     } deriving (Show, Generic)
 
 instance FromJSON NodeConfiguration where
     parseJSON = genericParseJSON defaultOptions
 
+-- | Specifies the way Uxtos are going to be grouped.
```
gromakovsky: Typo: Utxos.

Martoon-00: Right, thanks

Martoon-00:
We can't reuse `InputSelectionPolicy` as is, because it deserializes from `{ "groupingPolicy": "<value>" }`, and for config we need to deserialize from plain `"<value>"` 🙂

adinapoli-iohk:
> We can't reuse InputSelectionPolicy as is

Ah, I see, that's because we deserialise from the yaml file which uses the same `Aeson` instances, and thus we need an extra layer of indirection for the config :)

node/src/Pos/Client/Txp/Util.hs
```diff
@@ -147,7 +150,7 @@ isCheckedTxError = \case
 -- Tx creation
 -----------------------------------------------------------------------------
 
--- | Specifies the way Uxtos are going to be grouped.
+-- | Specifies the way Uxto is going to be grouped.
```
gromakovsky: There's still a typo :/

gromakovsky: Fixed by myself, because Konstantin is on vacation.

Martoon-00:
@adinapoli-iohk Since it's not getting merged, can I close this PR?

adinapoli-iohk:
@Martoon-00 It all boils down to whether or not we have a nice migration path for exchanges using the old API vs exchanges using the new (if you remember, the JSON payload for `newPayment` changed a little bit).

Would you do a bit of investigation and get back to me? My gut feeling is that we might need to rebase this against `master`, review it and make sure it's actually buying us something.

Let's start with baby step first: would you close this PR and open a new one against `master`, describing what this would buy us, and what are the consequences if we don't merge it?
Then we can kickoff the discussion and decide once for all if we want to merge it or not.

TL;DR Close this PR, open a new one, let's discuss 😉

Martoon-00:
Rebased to master, see ^

Martoon-00:
Now we can assess which work exactly should be done :eyes:

PR #2183, opened 2017-12-28
[CSL-1868] Write the guidelines for monadic effects

int-index:


docs/monads.md
```diff
@@ -0,0 +1,325 @@
+# Guidelines for Effect Handling in Cardano SL
+
+A monad comes with two basic operations, ``return`` and ``>>=``, but this
```
erikd: There is a proposal (which will be implemented in GHC 8.4 I think) for removing `return` from the `Monad` type class and using `pure` from the `Applicative` superclass in its place. Might be worth proactively using `pure` here now which should work with GHC 7.10 and later.

int-index: I am calling these operations "basic" because there are two ways to define a monad:

* via `return` + `>>=`
* via `return` + `fmap` + `join`

Alternatively, I could name `return`, `fmap`, and `join` as basic monad operations, although `join` is define outside the class.

Even after the "Monad of no return" proposal, these will be the basic monad operations.

int-index: If it's a naming issue, I don't think that `pure` vs `return` makes any difference, but people normally associate `return` with `Monad`, so I'd use it out of familiarity.

erikd: Maybe its just my biases showing. I'm currently working on a number of codebases which favour `pure` over `return`.

PR #2188, opened 2017-12-28
[CSL-2127] Make getHistoryCache return empty history instead of Nothing

Martoon-00:
This has already been done before (see https://github.com/input-output-hk/cardano-sl/blob/9e38d21cbeef62c7bb64169bba84c22ee7d5d687/node/src/Pos/Wallet/Web/State/Storage.hs#L450 for evidence), but was lost during rebase.

It results in unpleasant false warning appearing in logs.

### DIFF
--- a/node/src/Pos/Wallet/Web/State/State.hs
+++ b/node/src/Pos/Wallet/Web/State/State.hs
@@ -181,7 +181,7 @@ getUpdates = queryDisk A.GetUpdates
 getNextUpdate :: WebWalletModeDB ctx m => m (Maybe CUpdateInfo)
 getNextUpdate = queryDisk A.GetNextUpdate
 
-getHistoryCache :: WebWalletModeDB ctx m => CId Wal -> m (Maybe (Map TxId TxHistoryEntry))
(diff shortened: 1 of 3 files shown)


Martoon-00:
My bad, you answered a month ago and I noticed it only now :pensive: 
I suppose now I have to rebase it somewhere.

Some bits of explanation:
Previously we accessed history on following scheme:
* No history for wallet - invalid state
* Empty history - ok empty history
* Nonempty history - ok nonempty history
and correspondingly, getters returned `Nothing`, `Just mempty`, `Just smth`, the former just to throw error or warning later.

After some change, acid state switched on another scheme:
* No history for wallet - ok, corresponds to empty history
* Empty history - ok, but setters never produce this state, using "no history" instead
* Nonempty history - ok nonempty history
In some sense we merged 1 and 2, so that we can return `Just mempty` instead of `Nothing` - and now there is no need in `Maybe` at all.

Despite this fact acid state getters remained to be working according to previous scheme, along with code which uses them, due to invalid rebase.
For example, if there was a warning on "no history" state, now it is going to fire in normal situations because "no history" is about empty history actually - that is the one I was talking about.

Martoon-00:
Rebased, see ^

PR #2193, opened 2017-12-29
[CSL-2129] Add blockchain dumping and loading the dump

Pastafarianist:
This is a WIP implementation of an `auxx` command which traverses the blockchain starting from the tip and dumps each epoch to a separate CBOR file.

Currently there is a problem which may imply that I misunderstand how we store blocks. The algorithm this command executes is the following:

1. Set `start` = current tip.
2. Traverse blockchain from `start` to the nearest ancestor `G` which is a genesis block.
3. Dump all blocks from `start` to `G` to a file.
4. If `G` matches `genesisHash`, stop. Otherwise, set `start` = parent of G and goto 2.

This assumes that if a block is in the DB, then either it is the very first genesis block, or its parent is in the DB too. However, I've tested this patch and it seems that this is not the case.

```
$ scripts/launch/auxx.sh --mode=with-node repl
... lots of output ...
auxx> dump /tmp/cardano-blockchain
Command failed: 
DB contains block but not its parent
  error, called at src/Debug.hs:43:11 in universum-0.9.0-7P9sY0yY1Bo8ubfcEspjCA:Debug
...
```

Apparently, my understanding is incorrect and help would be appreciated.

### DIFF
--- a/auxx/src/Command/DumpBlockchain.hs
+++ b/auxx/src/Command/DumpBlockchain.hs
@@ -1,5 +1,5 @@
-module Command.Dump
-    ( dump
+module Command.DumpBlockchain
+    ( dumpBlockchain
     ) where
 
 import           Universum
@@ -86,11 +86,11 @@ stripProof (Right mb) = Right $ GenericBlockNoProof mb
 
 -- | Dump whole blockchain in CBOR format to the specified folder.
 -- Each epoch will be at <outFolder>/epoch<epochIndex>.cbor.
-dump
+dumpBlockchain
     :: (MonadAuxxMode m, MonadIO m)
     => FilePath
     -> m ()
-dump outFolder = withStateLock HighPriority "auxx" $ \_ -> do
+dumpBlockchain outFolder = withStateLock HighPriority "auxx" $ \_ -> do
     printTipDifficulty
     liftIO $ createDirectoryIfMissing True outFolder
     tipHeader <- DB.getTipHeader
--- a/auxx/src/Command/DumpBlockchain.hs
+++ b/auxx/src/Command/DumpBlockchain.hs
@@ -4,85 +4,27 @@ module Command.DumpBlockchain
 
 import           Universum
 
-import           Codec.Compression.Lzma (compress)
+import           Conduit (runConduit, runResourceT, sinkFile, yieldMany, (.|))
 import           Control.Lens (_Wrapped)
-import qualified Data.ByteString.Lazy as BSL
 import           Data.List.NonEmpty (last)
-import qualified Data.Vector as V
 import           Formatting (build, sformat, (%))
 import           System.Directory (createDirectoryIfMissing)
 import           System.FilePath ((</>))
 import           System.Wlog (logInfo, logWarning)
 
-import           Pos.Binary (Bi (..), encodeListLen, enforceSize, serialize)
+import           Pos.Block.Dump (encodeBlockDump)
 import           Pos.Core (Block, BlockHeader, EpochIndex (..), HeaderHash, blockHeaderHash,
                            difficultyL, epochIndexL, genesisHash, getBlockHeader, prevBlockL)
-import qualified Pos.Core as C
-import           Pos.Core.Block (GenericBlock (..), GenericBlockHeader (..))
-import           Pos.Crypto (HasCryptoConfiguration, getProtocolMagic, protocolMagic, shortHashF)
+import           Pos.Crypto (shortHashF)
 import           Pos.DB.Block (getBlock)
 import qualified Pos.DB.BlockIndex as DB
 import           Pos.DB.Class (MonadDBRead)
 import           Pos.DB.Error (DBError (..))
 import           Pos.StateLock (Priority (..), withStateLock)
-import           Pos.Util.Chrono (NewestFirst (..))
-import           Pos.Util.Util (eitherToFail, maybeThrow)
+import           Pos.Util.Chrono (NewestFirst (..), toOldestFirst)
+import           Pos.Util.Util (maybeThrow)
 
 import           Mode (MonadAuxxMode)
-
-newtype GenericBlockNoProof b = GenericBlockNoProof (GenericBlock b)
-
-type BlockNoProof = Either GenesisBlockNoProof MainBlockNoProof
-type GenesisBlockNoProof = GenericBlockNoProof C.GenesisBlockchain
-type MainBlockNoProof = GenericBlockNoProof C.MainBlockchain
-
-instance ( Typeable b
-         , Bi (C.BHeaderHash b)
-         , Bi (C.ConsensusData b)
-         , Bi (C.ExtraHeaderData b)
-         , Bi (C.Body b)
-         , Bi (C.ExtraBodyData b)
-         , C.BlockchainHelpers b
-         , HasCryptoConfiguration
-         ) =>
-         Bi (GenericBlockNoProof b) where
-    encode (GenericBlockNoProof gb) =
-        encodeListLen 3
-        <> encodeHeader (_gbHeader gb)
-        <> encode (_gbBody gb)
-        <> encode (_gbExtra gb)
-      where
-        encodeHeader bh =
-            encodeListLen 4
-            <> encode (getProtocolMagic protocolMagic)
-            <> encode (_gbhPrevBlock bh)
-            <> encode (_gbhConsensus bh)
-            <> encode (_gbhExtra bh)
-
-    decode = do
-        enforceSize "GenericBlock" 3
-        (prevBlock, consensus, headerExtra) <- decodeHeader'
-        body  <- decode
-        extra <- decode
-        let bodyProof = C.mkBodyProof body
-        header <- eitherToFail $
-            C.recreateGenericHeader prevBlock bodyProof consensus headerExtra
-        block <- eitherToFail $ C.recreateGenericBlock header body extra
-        pure $ GenericBlockNoProof block
-      where
-        decodeHeader' = do
-            enforceSize "GenericBlockHeader b" 4
-            blockMagic <- decode
-            when (blockMagic /= getProtocolMagic protocolMagic) $
-                fail $ "GenericBlockHeader failed with wrong magic: " <> show blockMagic
-            prevBlock <- decode
-            consensus <- decode
-            extra     <- decode
-            pure $ (prevBlock, consensus, extra)
-
-stripProof :: Block -> BlockNoProof
-stripProof (Left gb)  = Left $ GenericBlockNoProof gb
-stripProof (Right mb) = Right $ GenericBlockNoProof mb
 
 -- | Dump whole blockchain in CBOR format to the specified folder.
 -- Each epoch will be at <outFolder>/epoch<epochIndex>.cbor.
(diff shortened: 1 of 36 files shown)


block/src/Pos/Block/Dump.hs
```diff
@@ -0,0 +1,273 @@
...
+       -- * Encoding
+         encodeBlockDump
+       , encodeBlockDump'
+
+       -- * Decoding
+       , decodeBlockDump
+       , decodeBlockDump'
+       , BlockDumpDecodeError (..)
+
+       -- * Internals
```
Pastafarianist: Why are so many internals exported? E.g. where are you using `stripProof`?

block/src/Pos/Block/Dump.hs
```diff
@@ -0,0 +1,273 @@
...
+decodeBlockDump
+    :: (Bi BlockNoProof, Conduit.MonadResource m, MonadThrow m)
+    => Conduit ByteString m Block
+decodeBlockDump =
+    Lzma.decompress memlimit .| do
+        ver <- decodeDumpHeader
+        when (ver /= 0) $ throwM (UnsupportedBlockVersion ver (one 0))
+        decodeBlockStream
+  where
+    memlimit = Just (200*1024*1024)  -- 200 MB
```
Pastafarianist: memelimit

block/src/Pos/Block/Dump.hs
```diff
@@ -0,0 +1,273 @@
...
+stripProof (Left gb)  = Left  $ GenericBlockNoProof gb
+stripProof (Right mb) = Right $ GenericBlockNoProof mb
+
+addProof :: BlockNoProof -> Block
+addProof (Left  (GenericBlockNoProof gb)) = Left gb
+addProof (Right (GenericBlockNoProof mb)) = Right mb
+
+-- | Encode the block without including any proofs. When decoding, recreate
+-- the proofs.
+instance ( Typeable b
```
Pastafarianist: Any ideas how to avoid duplicating code here? This is essentially copy-paste from `core/Pos/Binary/Core/Blockchain.hs`.

block/src/Pos/Block/Logic/VAR.hs
```diff
@@ -179,12 +185,34 @@ verifyAndApplyBlocks rollback blocks = runExceptT $ do
...
+    :: forall ctx m. (BlockLrcMode ctx m, MonadMempoolNormalization ctx m)
+    => Bool -> Consumer Block m (Either ApplyBlocksException HeaderHash)
+verifyAndApplyBlocksC rollback = do
+    tip <- lift GS.getTip
+    splitC 1000 .| Conduit.mapC OldestFirst .| processBatches tip
+  where
+    processBatches tip = Conduit.await >>= \case
+        Nothing     -> pure (Right tip)
+        Just blocks -> lift (verifyAndApplyBlocks rollback blocks) >>= \case
+            Left err     -> pure (Left err)
```
Pastafarianist: What state does `verifyAndApplyBlocks` leave if it errors?

crypto/Pos/Binary/Crypto.hs
```diff
@@ -239,3 +240,11 @@ instance Bi EdStandard.Signature where
...
 deriving instance Bi RedeemSecretKey
 deriving instance Typeable a => Bi (RedeemSignature a)
+
+----------------------------------------------------------------------------
+-- Configuration
+----------------------------------------------------------------------------
+
+instance Bi ProtocolMagic where
+    encode (ProtocolMagic magic) = encode magic
+    decode = ProtocolMagic <$> decode
```
Pastafarianist: KPACUBO

crypto/Pos/Crypto/Signing/Redeem.hs
```diff
@@ -46,7 +46,7 @@ redeemDeterministicKeyGen seed =
 
 -- | Encode something with 'Binary' and sign it.
 redeemSign ::
-       (HasCryptoConfiguration, Bi a)
+       (HasCryptoConfiguration, Bi a, Bi ProtocolMagic)
```
Pastafarianist: Why can't you import `Pos.Crypto.Binary` here?

crypto/Pos/Crypto/Signing/Tag.hs
```diff
@@ -5,14 +5,16 @@ module Pos.Crypto.Signing.Tag
...
+import           Pos.Binary.Class (Bi, serialize')
 import           Pos.Crypto.Configuration (HasCryptoConfiguration, ProtocolMagic (..),
                                            protocolMagic)
 import           Pos.Crypto.Signing.Types.Tag
 
 -- | Get magic bytes corresponding to a 'SignTag'. Guaranteed to be different
 -- (and begin with a different byte) for different tags.
-signTag :: HasCryptoConfiguration => SignTag -> ByteString
+signTag
+    :: (Bi ProtocolMagic, HasCryptoConfiguration)
```
Pastafarianist: And here.

block/src/Pos/Block/Dump.hs
```diff
@@ -0,0 +1,273 @@
...
+----------------------------------------------------------------------------
+
+-- | Encode a block dump. The blocks should be coming oldest-first.
+encodeBlockDump
+    :: (Bi BlockNoProof, Conduit.MonadResource m)
+    => Conduit Block m ByteString
+encodeBlockDump = (encodeDumpHeader >> encodeBlockStream)
+               .| Lzma.compress Nothing
+
+-- | Like 'encodeBlockDump', but without conduits.
```
Pastafarianist: Please follow your naming convention everywhere. Earlier you defined `func` and `funcC` for a non-conduit and conduit version respectively. Can you rename these functions to follow this convention?

lib/test/Test/Pos/Block/DumpSpec.hs
```diff
@@ -0,0 +1,51 @@
...
+    dec <- QC.run $ decodeBlockDump' $
+        BSL.fromChunks $ chunkBS chunkSize $ BSL.toStrict enc
+    pure (OldestFirst blocks === dec)
+
+----------------------------------------------------------------------------
+-- Utilities
+----------------------------------------------------------------------------
+
+chunkBS :: Int -> ByteString -> [ByteString]
+chunkBS n = takeWhile (not . null) . map (BS.take n) . iterate (BS.drop n)
```
Pastafarianist: That's pretty elegant, but are you sure there isn't any library function for that?

Pastafarianist:
BTW I can't approve or reject this PR since I'm the original author.

auxx/src/Command/DumpBlockchain.hs
```diff
@@ -164,3 +117,48 @@ dumpBlockchain outFolder = withStateLock HighPriority "auxx" $ \_ -> do
...
+-- | Load blocks from a file or directory and apply them.
+applyBlockchainDump
+    :: (MonadAuxxMode m, MonadIO m)
+    => FilePath -> m ()
+applyBlockchainDump path = withStateLock HighPriority "auxx" $ \_ -> do
+    isDir <- liftIO $ doesDirectoryExist path
+    if isDir then applyDir path else applyEpoch path
+
+-- | Apply all epochs from a directory.
+applyDir
```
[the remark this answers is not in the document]
neongreen: I don't like functions that take more than 1.5 screens, and usually my heuristic is “if I can write a Haddock docstring for it, it probably can be moved out of `where`”. It also helps that when the function is out of `where`, I can jump to its definition :)

block/src/Pos/Block/Dump.hs
neongreen: I'm not using `stripProof`, but I feel that if `BlockNoProof` is exported, then `stripProof` should be exported as well, and exporting `BlockNoProof` is needed to be able to write `Bi BlockNoProof` in type signatures.

block/src/Pos/Block/Dump.hs
neongreen: No ideas, unfortunately. This instance could be autoderived if `deriveBiSimple` was a bit smarter.

block/src/Pos/Block/Logic/VAR.hs
neongreen: If `rollback` is true, it does a rollback. If `rollback` is false, it silently doesn't apply failed blocks and returns the current tip.

crypto/Pos/Binary/Crypto.hs
neongreen: A to.

crypto/Pos/Crypto/Signing/Redeem.hs
neongreen: I think it would lead to an import cycle, but I haven't checked.

lib/test/Test/Pos/Block/DumpSpec.hs
neongreen: Well, I haven't found it. There definitely is one in some library or other, but I've no idea which.

crypto/Pos/Crypto/Signing/Redeem.hs
Pastafarianist: It wouldn't :)

lib/test/Test/Pos/Block/DumpSpec.hs
Pastafarianist: Ok.

crypto/Pos/Crypto/Signing/Tag.hs
Pastafarianist: Here everything works with that import as well.

auxx/src/Command/DumpBlockchain.hs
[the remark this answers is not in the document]
Pastafarianist: I feel that moving everything into a `where` clause is superior here because these functions are auxiliary with respect to `applyBlockchainDump` and aren't used anywhere else, while keeping them outside of `where` hints at the opposite. On the other hand, `where` introduces more visual clutter.

In this case, I'd probably weakly agree with you.

block/src/Pos/Block/Dump.hs
Pastafarianist: You have 5 exports that are never used for one `BlockNoProof`. I think this is too much.

block/src/Pos/Block/Logic/VAR.hs
Pastafarianist: So if `rollback` is `False` and there is an error, we are left with a state where `1000*k` blocks have been applied, where `k` is some integer?

auxx/src/Command/Proc.hs
```diff
@@ -275,6 +277,18 @@ createCommandProcs hasAuxxMode printAction mSendActions = rights . fix $ \comman
...
+    let name = "apply-blockchain-dump" in
+    needsAuxxMode name >>= \Dict ->
+    return CommandProc
+    { cpName = name
+    , cpArgumentPrepare = identity
+    , cpArgumentConsumer = getArg tyFilePath "file/directory"
+    , cpExec = \path -> do
+        DumpBlockchain.applyBlockchainDump path
+        return ValueUnit
+    , cpHelp = ""
```
int-index: add a help message please

auxx/src/Command/Proc.hs
```diff
@@ -467,6 +481,20 @@ createCommandProcs hasAuxxMode printAction mSendActions = rights . fix $ \comman
...
+    return CommandProc
+    { cpName = name
+    , cpArgumentPrepare = identity
+    , cpArgumentConsumer = do
+        dumpOutFolder <- getArg tyFilePath "dump-folder"
+        pure DumpBlockchainParams{..}
+    , cpExec = \DumpBlockchainParams{..} -> do
+        DumpBlockchain.dumpBlockchain dumpOutFolder
+        return ValueUnit
+    , cpHelp = ""
```
int-index: add a help message please

auxx/src/Command/Proc.hs
```diff
@@ -275,6 +277,18 @@ createCommandProcs hasAuxxMode printAction mSendActions = rights . fix $ \comman
                \ using secret key #i"
     },
 
+    let name = "apply-blockchain-dump" in
+    needsAuxxMode name >>= \Dict ->
+    return CommandProc
+    { cpName = name
+    , cpArgumentPrepare = identity
+    , cpArgumentConsumer = getArg tyFilePath "file/directory"
```
int-index: that's a bad name for a keyword argument, I'm not sure a slash will be even parsed appropriately

block/src/Pos/Block/Logic/VAR.hs
neongreen: No, it can be any amount of blocks.

neongreen:
Merging it because it's going to be a PR-on-top-of-PR anyway.

PR #2199, opened 2017-12-29
[CSL-2130] Get logger params from CommonArgs instead of CommonNodeArgs

hithroc:


georgeee:
Closed because we are to deliver `develop` and we have these changes there already

PR #2216, opened 2017-12-31
CSL-1852 Wrap big tuples in launcher into sensible types 

v0d1ch:
Use new types instead of passing three or four element tuples around

### DIFF
--- a/tools/src/launcher/Main.hs
+++ b/tools/src/launcher/Main.hs
@@ -98,7 +98,23 @@ data LauncherOptions = LO
 -- | The concrete monad where everything happens
 type M a = (HasConfigurations, HasCompileInfo) => Log.LoggerNameBox IO a
 
+-- | Executable can be node, wallet or updater
 data Executable = EWallet | ENode | EUpdater
+
+-- | This datatype holds values for either node or wallet
+--   Node/wallet path, args, log path
+data NodeData = NodeData
+    { ndPath    :: !FilePath
+    , ndArgs    :: ![Text]
+    , ndLogPath :: Maybe FilePath }
+
+-- | Updater path, args, windows runner path, archive path
+data UpdaterData = UpdaterData
+    { udPath        :: !FilePath
+    , udArgs        :: ![Text]
+    , udWindowsPath :: Maybe FilePath
+    , udArchivePath :: Maybe FilePath
+    }
 
 optionsParser :: Parser LauncherOptions
 optionsParser = do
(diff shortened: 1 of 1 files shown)


tools/src/launcher/Main.hs
```diff
@@ -100,6 +100,20 @@ type M a = (HasConfigurations, HasCompileInfo) => Log.LoggerNameBox IO a
 
 data Executable = EWallet | ENode | EUpdater
 
+-- Node, its args, node log
+data NodeData = NodeData
```
gromakovsky: I'd write a comment that this type can also carry wallet data (wallet is also a node, but it's not obvious whether `NodeData` corresponds only to wallet or to either node or wallet).

tools/src/launcher/Main.hs
```diff
@@ -100,6 +100,20 @@ type M a = (HasConfigurations, HasCompileInfo) => Log.LoggerNameBox IO a
...
+    { _nodePath    :: FilePath
+    , _nodeArgs    :: [Text]
+    , _nodeLogPath :: Maybe FilePath }
+
+-- Updater Node, its args, node log, archive path
```
gromakovsky: Updater is not a node, it's just a separate executable (which is currently unused, btw).
Can you remove `Node` from field names and this comment?

tools/src/launcher/Main.hs
```diff
@@ -100,6 +100,20 @@ type M a = (HasConfigurations, HasCompileInfo) => Log.LoggerNameBox IO a
 
 data Executable = EWallet | ENode | EUpdater
 
+-- Node, its args, node log
+data NodeData = NodeData
+    { _nodePath    :: FilePath
+    , _nodeArgs    :: [Text]
+    , _nodeLogPath :: Maybe FilePath }
```
gromakovsky: Why did you use `_` prefix? The only reason to use `_` prefix I am aware of is when you also use `makeLenses` (because it strips `_`). But you don't use `_` for this type.
Our common convention is to use abbreviation of the type as prefix, i. e. `nd` in this case, so the fields would be called: `ndPath`, `ndArgs` and `ndLogPath`.
Also our style guides suggest making all fields strict by default (using `!`), i. e. when there are no reasons to make them lazy.

tools/src/launcher/Main.hs
```diff
@@ -100,6 +100,20 @@ type M a = (HasConfigurations, HasCompileInfo) => Log.LoggerNameBox IO a
...
+
+-- Updater Node, its args, node log, archive path
+data UpdaterData = UpdaterData
+    { _uNodePath          :: FilePath
+    , _uNodeArgs          :: [Text]
+    , _uNodeLogPath       :: Maybe FilePath
```
gromakovsky: Are you sure it's the correct name? AFAIU, the third field is the path to windows runner.

tools/src/launcher/Main.hs
gromakovsky: Ah, probably that's because you don't use these names and it leads to warnings.
Well, to be honest, I think it's better to use these names. Explicitly assigning values to different fields makes code clearer, in my opinion.

v0d1ch: > Ah, probably that's because you don't use these names and it leads to warnings.

Right. I used proper assigning instead of pattern matching so I don't get warnings no more.

PR #2229, opened 2018-01-02
WIP CSL-1838 report absence of block with tip hash to reporting server

v0d1ch:


block/src/Pos/DB/Block/Load.hs
```diff
@@ -182,9 +183,11 @@ getHeaderThrow hash =
...
-    :: MonadDBRead m
+    :: forall ctx m. (MonadReporting ctx m, MonadDBRead m)
     => HeaderHash -> m Blund
-getBlundThrow hash =
-    maybeThrow (DBMalformed $ sformat errFmt hash) =<< getBlund hash
+getBlundThrow hash = do
+    mBlund <- getBlund hash
+    whenJust mBlund (\_ -> reportError $ sformat errFmt hash)
+    maybeThrow (DBMalformed $ sformat errFmt hash) mBlund
   where
```
v0d1ch: @gromakovsky Am I on the right path here with adding `MonadReporting` constraint ? After this I get errors in `tools/src/blockchain-analyser/Main.hs`
```
    • Could not deduce (Pos.KnownPeers.MonadFormatPeers Production)
            arising from a use of ‘DB.loadBlundsFromTipWhile’
          from the context: HasConfiguration
            bound by the type signature for:
                       analyseBlockchainLazily :: HasConfiguration =>
                                                  CLIOptions -> BlockchainInspector ()
            at src/blockchain-analyser/Main.hs:67:1-83
        • In the second argument of ‘(<$>)’, namely
            ‘DB.loadBlundsFromTipWhile (const True)’
          In a stmt of a 'do' block:
            allBlocks <- map (bimap identity Just) . getNewestFirst
                         <$> DB.loadBlundsFromTipWhile (const True)
          In the expression:
            do { allBlocks <- map (bimap identity Just) . getNewestFirst
                              <$> DB.loadBlundsFromTipWhile (const True);
                 putText (renderBlocks cli allBlocks) }
```

v0d1ch: I reverted it back since I don't think that is correct way of doing this. Am I supposed to use `reportError` function to report the error to server ? Maybe there is another function that can run in `MonadDBRead` context and have ability to do the same task ?

v0d1ch: @gromak Just to check - I am trying to send error report to server from `getBlundThrow` function from `cardano-sl/block/src/Pos/DB/Block/Load.hs`. Problem with this is that I have to add whole bunch of constraints which feels wrong
```
(MonadIO m, MonadMask m,
     MonadReader ctx m, MonadFormatPeers m,
     HasReportingContext ctx,
     HasNodeType ctx,
     WithLogger m, HasConfiguration,
     HasCompileInfo)
```
The `getBlundThrow` seems to me like a good place to send error report since that is where exception is thrown, maybe I am wrong.

gromakovsky:
@v0d1ch after thinking a bit more thoroughly about this issue, I propose a slightly different approach: let's report all `DBMalformed` exceptions. In fact, `DBMalformed` is something that should never happen. Of course some user can manually modify DB contents (not using our code) and then our code will report non-existing errors, but actually users can report non-existing errors anyway (everyone can report arbitrary errors, it's public). It's also possible that DB will be malformed due to FS error or hardware error, but I guess it won't harm to receive statistics about such cases.

In `Pos.Reporting.Methods` there is `reportOrLog`. We use this function in all exception handlers which catch `SomeException` (they are quite high-level). Currently it reports `CardanoFatalError`s and `ErrorCall`s. I suggest that we also report `DBMalformed` there.
That function does some tricky stuff (not very tricky, just combination of `AllowAmbiguousTypes` and `TypeApplications`), let me know if you get stuck modifying it.

v0d1ch:
@gromakovsky can you check the outdated comments, I still have issues trying to work with around bunch of constraints needed

gromakovsky:
@v0d1ch it's not needed to add `MonadReporting` constraint to `loadBlundsByDepth` and similar functions. It's enough to change `reportOrLog`.
If there is no block with tip hash, `DBMalformed` will be thrown and caught by `reportOrLog` which will report it.
So new I see that ` <|> tryCast @DBError` should be actually enough 🤷‍♂️

v0d1ch:
Oh another unnecessary head-banging session :)

v0d1ch:
> So new I see that <|> tryCast @DBError should be actually enough
@gromakovsky  if this is correct than this PR is ready

v0d1ch:
All good, little headache never killed anyone :)

PR #2231, opened 2018-01-02
[CSL-2135] Add endpoint to remove all canceled transactions from pending list

Martoon-00:
Two endpoints:
* Bulk removal of canceled transactions - for a new version of plan.
* Removal of specific transaction - for old plan if occur to be needed

Tested with `Applying`, `InBlocks` and manually created `WontApply` transactions


jmitchell:
cc @dcoutts

PR #2239, opened 2018-01-03
[CSL-2135] Cancel PTX on resubmission after 1h limit

int-index:


wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -116,11 +118,22 @@ submitAndSavePtx PtxSubmissionHandlers{..} enqueue ptx@PendingTx{..} = do
...
-    addOnlyNewPendingTx ptx
-    when ack $ ptxUpdateMeta _ptxWallet _ptxTxId PtxMarkAcknowledged
+    now <- getCurrentTimestamp
+    if | PtxApplying poolInfo <- _ptxCond,
+         Just creationTime <- poolInfo ^. thTimestamp,
+         diffTimestamp now creationTime > hour 1 -> do
+           let newCond = PtxWontApply "1h limit exceeded" poolInfo
+           void $ casPtxCondition _ptxWallet _ptxTxId _ptxCond newCond
+           logInfo $
+             sformat ("Pending transaction #"%build%" discarded becauce \
```
adinapoli-iohk: Minor typo, that's "because".

Martoon-00:
LGTM!

As far as I understand, following things become broken:
1. Invariant `transaction is in WontApply state => it will never get into blocks`
2. Taking into account known bug when transaction in block is marked as `Applying`, user may get transaction in forever `WontApply` state which is actually in blockchain

AFAIU it was discussed in slack, (2) is to be fixed soon, and we took risk of 1 for now.

Martoon-00:
Regarding question about possibly absent timestamp: when transaction is made by us, `_thTimestamp` is set immediately on creation. And pending states are kept only for those transactions which are made by us.
So `_thTimestamp` will be always `Just _` in this case.

PR #2243, opened 2018-01-03
[CSM-500] Remember pending transaction from the very beginning

neongreen:
This is important, because if we add transaction to pending list once transaction creation is done,
then creation and addition to pending list may complete only after the moment when transaction gets to
blocks, so on this "transaction detected in blocks" nothing effectively happens. (and adding transaction to pending list once it detected in block is problematic, because there are many meta info to be stored along with pending transaction which may be not accessible from block tracker context)

Solution is to add new pending transaction status, `PtxCreating`, in which
transaction is not displayed to frontend, and such transaction may be safely removed if
full cycle of transaction creation finishes with fatal error (unlike
`PtxApplying`).
Transaction remembered with this status still caries all pending meta info,
so blocks tracker can successfully update status of that transaction.

(cherry picked from commit 2c1cfa3290af33e4b34f6f6ebfe46315af1ce2cb)

### DIFF
--- a/node/src/Pos/Wallet/Web/State/Storage.hs
+++ b/node/src/Pos/Wallet/Web/State/Storage.hs
@@ -79,6 +79,7 @@ module Pos.Wallet.Web.State.Storage
        , removeFromHistoryCache
        , setPtxCondition
        , casPtxCondition
+       , removeOnlyCreatingPtx
        , ptxUpdateMeta
        , addOnlyNewPendingTx
        , resetFailedPtxs
@@ -88,16 +89,18 @@ module Pos.Wallet.Web.State.Storage
 
 import           Universum
 
-import           Control.Lens                   (at, ix, makeClassy, makeLenses, non', to,
-                                                 toListOf, traversed, (%=), (+=), (.=),
-                                                 (<<.=), (?=), _Empty, _head)
-import           Control.Monad.State.Class      (put)
+import           Control.Lens                   (at, has, ix, makeClassy, makeLenses, non',
+                                                 to, toListOf, traversed, (%=), (+=), (.=),
+                                                 (<<.=), (?=), _Empty, _Just, _head)
+import           Control.Monad.State.Class      (get, put)
+import qualified Control.Monad.State.Lazy       as LS
 import           Data.Default                   (Default, def)
 import qualified Data.HashMap.Strict            as HM
 import qualified Data.Map                       as M
 import           Data.SafeCopy                  (Migrate (..), base, deriveSafeCopySimple,
                                                  extension)
 import           Data.Time.Clock.POSIX          (POSIXTime)
+import           Serokell.Util                  (zoom')
 
 import           Pos.Client.Txp.History         (TxHistoryEntry, txHistoryListToMap)
 import           Pos.Core.Configuration         (HasConfiguration)
@@ -115,7 +118,7 @@ import           Pos.Wallet.Web.ClientTypes     (AccountId, Addr, CAccountMeta, 
                                                  PassPhraseLU, Wal, addrMetaToAccount)
 import           Pos.Wallet.Web.Pending.Types   (PendingTx (..), PtxCondition,
                                                  PtxSubmitTiming (..), ptxCond,
-                                                 ptxSubmitTiming)
+                                                 ptxSubmitTiming, _PtxCreating)
 import           Pos.Wallet.Web.Pending.Updates (cancelApplyingPtx,
                                                  incPtxSubmitTimingPure,
                                                  mkPtxSubmitTiming,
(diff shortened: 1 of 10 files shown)


wallet/src/Pos/Wallet/Web/Pending/Util.hs
```diff
@@ -53,7 +61,7 @@ mkPendingTx wid _ptxTxId _ptxTxAux th = do
     _ptxCreationSlot <- getCurrentSlotInaccurate
     CWalletMeta{..} <- maybeThrow noWallet =<< getWalletMeta wid
     return PendingTx
-        { _ptxCond = PtxApplying th
+        { _ptxCond = PtxCreating th
```
adinapoli-iohk: Right, that's the default state for a new transaction, after this patch.

wallet/src/Pos/Wallet/Web/Pending/Util.hs
```diff
@@ -32,9 +33,16 @@ import           Pos.Wallet.Web.Pending.Updates (mkPtxSubmitTiming)
...
+ptxPoolInfo (PtxCreating i)     = Just i
+ptxPoolInfo (PtxApplying i)     = Just i
+ptxPoolInfo (PtxWontApply _ i)  = Just i
+ptxPoolInfo PtxInNewestBlocks{} = Nothing
+ptxPoolInfo PtxPersisted{}      = Nothing
+
+-- | Whether transaction is claimed to be once created.
+isPtxActive :: PtxCondition -> Bool
+isPtxActive PtxCreating{} = False
+isPtxActive _             = True
```
adinapoli-iohk: Can a transaction be considered `active` is if in `Applying` as well? What `active` really means in this context? I guess it really means "everything but `Creating`"?

wallet/src/Pos/Wallet/Web/Pending/Util.hs
```diff
@@ -32,9 +33,16 @@ import           Pos.Wallet.Web.Pending.Updates (mkPtxSubmitTiming)
...
 
 ptxPoolInfo :: PtxCondition -> Maybe PtxPoolInfo
-ptxPoolInfo (PtxApplying i)    = Just i
-ptxPoolInfo (PtxWontApply _ i) = Just i
-ptxPoolInfo _                  = Nothing
+ptxPoolInfo (PtxCreating i)     = Just i
+ptxPoolInfo (PtxApplying i)     = Just i
+ptxPoolInfo (PtxWontApply _ i)  = Just i
+ptxPoolInfo PtxInNewestBlocks{} = Nothing
+ptxPoolInfo PtxPersisted{}      = Nothing
```
adinapoli-iohk: I like the fact we unrolled the wildcard 😉

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -116,10 +118,15 @@ submitAndSavePtx PtxSubmissionHandlers{..} enqueue ptx@PendingTx{..} = do
...
         throwM $ InternalError "Transaction creation is disabled by configuration!"
 
+    addOnlyNewPendingTx ptx
     ack <- submitTxRaw enqueue _ptxTxAux
     reportSubmitted ack
-    saveTx (_ptxTxId, _ptxTxAux) `catches` handlers ack
-    addOnlyNewPendingTx ptx
+    (saveTx (_ptxTxId, _ptxTxAux)
+        `catches` handlers ack)
+        `onException` creationFailedHandler
```
adinapoli-iohk: Help me out to untangle this double layer of exception handling: are we really saying here: try to gracefully handle the exception (first `catches`) and if no other handler match, fail with a generic exception message (second `onException` block?)
  
If that's true, wouldn't the 2nd `Handler $ \e@SomeException{}` always match, making the second `onException` effectively unreachable?

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -202,17 +203,19 @@ updateTransaction :: MonadWalletWebMode m => AccountId -> CTxId -> CTxMeta -> m
...
     pendingTxs <- fromMaybe [] <$> getWalletPendingTxs wid
     let conditions = map _ptxCond pendingTxs
     -- show only actually pending transactions in logs
     logTxHistory "Pending" $ mapMaybe (preview _PtxApplying) conditions
     -- but return all transactions which are not yet in blocks
-    let candidatesList = txHistoryListToMap (mapMaybe ptxPoolInfo conditions)
+    let candidatesList = txHistoryListToMap (mapMaybe getPtxTh conditions)
     return $ Map.union currentHistory candidatesList
+  where
+    getPtxTh cond = guard (isPtxActive cond) *> ptxPoolInfo cond
```
adinapoli-iohk: Right, this is effectively filtering the history to only Txs which are anything but in `Creating` state.

node/src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -477,14 +480,29 @@ setPtxCondition :: CId Wal -> TxId -> PtxCondition -> Update ()
...
+    :: CId Wal
+    -> TxId
+    -> (Maybe PtxCondition -> Bool)
+    -> LS.State (Maybe PendingTx) ()
+    -> Update Bool
+checkAndSmthPtx wid txId whetherModify modifier =
+    fmap getAny $ zoom' (wsWalletInfos . ix wid . wsPendingTxs . at txId) $ do
+        matches <- whetherModify . fmap _ptxCond <$> get
+        when matches modifier
+        return (Any matches)
```
adinapoli-iohk: That's some dense code, but here is my interpretation: Zoom to focus on a particular `PendingTx` given its `TxId`, and if the predicate `whetherModify` matches, run `modifier`. Then return true if we had a match in the first place.

A minor nitpick would be that this function is really not enforcing that `modifier` actually _did_ something, just that we had a match in the first place, which is probably fine(?).

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
kantp: Good catch! The `creationFailedHandler` will never fire

kantp: We should change this, to either remove the redundant handler, or make sure it can be reached.

wallet/src/Pos/Wallet/Web/Pending/Util.hs
neongreen: > Whether transaction is claimed to be once created.

Yep.

neongreen:
@kantp fixed.

neongreen:
Rebased.

neongreen:
We're dropping the exchanges branch.

PR #2261, opened 2018-01-05
Remove unnecessary details from README, just links.

denisshevchenko:
The idea is to remove all detailed information from `README.md` and keep it in minimalistic mode with links only. In this case `README.md` never be outdated and we won't keep the same information in two places.


lib/README.md
```diff
@@ -1,86 +1,31 @@
-<!-- THIS IS AUTOGENERATED FILE. DO NOT CHANGE IT MANUALLY! -->
```
gromakovsky: How was this file generated? Should we disable the mechanism which was used?

PR #2265, opened 2018-01-06
CSL-1818 Reject transaction if adopted script version mismatch

v0d1ch:


txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -205,6 +206,8 @@ verifyKnownInputs VTxContext {..} resolvedInputs TxAux {..} = do
                 redVer = scrVersion twRedeemer
             when (valVer /= redVer) $
                 throwError $ WitnessScriptVerMismatch valVer redVer
+            when (valVer /= vtcAdoptedVersion) $
```
gromakovsky: It should be `valVer > vtcAdoptedVersion`. It's ok to use script version less than the adopted one.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -44,6 +44,7 @@ data VTxContext = VTxContext
     { -- | Verify that script versions in tx are known, addresses' and
       -- witnesses' types are known, attributes are known too.
       vtcVerifyAllIsKnown :: !Bool
+    , vtcAdoptedVersion   :: !ScriptVersion
```
gromakovsky: Can you rename it to `vtcAdoptedScriptVersion` please? There are many different versions which can be adopted, not only script version.

v0d1ch:
All done

gromakovsky:
@v0d1ch no, you didn't fix CI. Tests don't compile. See https://ci.appveyor.com/project/jagajaga/cardano-sl/build/1.0.18902

v0d1ch:
In order to fix the test I need to add `ScriptVersion` param to `VTxContext` creation. My question is how do I do it in test context which expects type `Property` as function result and not some `m Property` ?

v0d1ch:
To clarify this is how the function looks
```
verifyTxInUtxo :: HasConfiguration => SmallGenerator GoodTx -> Property
verifyTxInUtxo (SmallGenerator (GoodTx ls)) =
    let txs = fmap (view _1) ls
        witness = V.fromList $ toList $ fmap (view _4) ls
        (ins, outs) = NE.unzip $ map (\(_, tIs, tOs, _) -> (tIs, tOs)) ls
        newTx = UnsafeTx ins (map toaOut outs) (mkAttributes ())
        utxo = M.fromList $ do
            tx@UnsafeTx{..} <- toList txs
            let id = hash tx
            (idx, out) <- zip [0..] (toList _txOutputs)
            pure ((TxInUtxo id idx), TxOutAux out)
        vtxContext = VTxContext False
        txAux = TxAux newTx witness
    in counterexample ("\n"+|nameF "txs" (blockListF' "-" genericF txs)|+""
                           +|nameF "transaction" (B.build txAux)|+"") $
       qcIsRight $ verifyTxUtxoPure vtxContext utxo txAux
```
and in order to get adopted script version I need to do something like  `adoptedBlockData <- gsAdoptedBVData` which introduces monadic context to test.

v0d1ch:
Ok this was very helpful since I lost some time trying to make it work with adopted script version.
> As a good exercise, can you add a test where adopted version is 0, but transaction has a script witness with script version = 1

I will do that

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -459,3 +467,11 @@ txShouldFailWithPlutus res err = case res of
     other -> expectationFailure $
         "expected: Left ...: " <> show (WitnessScriptError err) <> "\n" <>
         " but got: " <> show other
+
+-- | Transaction should fail with a 'ToilInvalidWitness' error.
+txShouldFailWithInvalidWitness :: Either ToilVerFailure () -> Expectation
+txShouldFailWithInvalidWitness = \case
+    Left ToilInvalidWitness{} -> pass
```
v0d1ch: @gromak I am kinda stuck here trying to make test fail with `ToilInvalidWitness` ...

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -266,6 +270,13 @@ scriptTxSpec = describe "script transactions" $ do
                 alwaysSuccessValidator
                 (const witness)
 
+        it "validator script provided in witness is greater \
+           \than adopted script version" $ do
+            let witness = ScriptWitness intValidatorWithVersionOne goodIntRedeemer
+            txShouldFailWithInvalidWitness $ checkScriptTx
+                alwaysSuccessValidator
```
gromakovsky: AFAIU, you should pass `intValidatorWithVersionOne` here.

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
gromakovsky: It'd be good to also check that `WitnessVerFailure` inside `ToilInvalidWitness` is actually `WitnessAdoptedScriptVerMismatch`.

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
v0d1ch: Yes that is correct

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -456,3 +467,19 @@ txShouldFailWithPlutus res err = case res of
...
+        WitnessScriptVerMismatch _ _ -> expectationFailure $
+            "expected: WitnessAdoptedScriptVerMismatch\n" <>
+            " but got: WitnessScriptVerMismatch"
+        other -> expectationFailure $
+            "expected: WitnessAdoptedScriptVerMismatch\n" <>
+            " but got: " <> show other
+
+    other -> expectationFailure $
+        "expected: Left ToilInvalidWitness\n" <>
+        " but got: " <> show other
```
gromakovsky: It's better to use `pretty` instead of `show` (here and above).

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -456,3 +467,19 @@ txShouldFailWithPlutus res err = case res of
     other -> expectationFailure $
         "expected: Left ...: " <> show (WitnessScriptError err) <> "\n" <>
         " but got: " <> show other
+
+-- | Transaction should fail with a 'ToilInvalidWitness' specifically 'WitnessAdoptedScriptVerMismatch' error.
+txShouldFailWithInvalidWitness :: Either ToilVerFailure () -> Expectation
```
gromakovsky: I think it's better to make this name more specific, e. g. `txShouldFailWithInvalidAdoptedScriptVer`.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -203,6 +204,8 @@ verifyKnownInputs VTxContext {..} resolvedInputs TxAux {..} = do
         ScriptWitness{..} -> do
             let valVer = scrVersion twValidator
                 redVer = scrVersion twRedeemer
+            when (valVer > vtcAdoptedScriptVersion) $
+                throwError $ WitnessAdoptedScriptVerMismatch valVer vtcAdoptedScriptVersion
             when (valVer /= redVer) $
                 throwError $ WitnessScriptVerMismatch valVer redVer
             when (vtcVerifyAllIsKnown && not (isKnownScriptVersion valVer)) $
```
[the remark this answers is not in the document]
gromakovsky: Can you clarify? Ideally this order shouldn't matter for your test. Your test should be done in such way that all conditions are satisfied except the one about adopted script version.
AFAIU, currently your test passes (because CI passes). Which permutations of these lines lead to failing tests?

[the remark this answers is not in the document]
v0d1ch: > Ideally this order shouldn't matter for your test
That is why I say it sucks. If you put `when (valVer /= redVer) $` line before `when (valVer > vtcAdoptedScriptVersion) $` then tests will fail.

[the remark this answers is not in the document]
gromakovsky: > If you put when (valVer /= redVer) $ line before when (valVer > vtcAdoptedScriptVersion) $ then tests will fail.

That's exactly what I wanted to know. It should be investigated, I'll take a look.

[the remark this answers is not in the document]
gromakovsky: Oh, well, that's because you use `goodIntRedeemer` as redeemer and it has version 0. Redeemer and validator should have same versions.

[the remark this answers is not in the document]
v0d1ch: Ok I will take a look and fix later today, thanks!

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
v0d1ch: What would be the most correct way to do it since `expectationFailure` expects `String` and pretty produces `Text` so I need `Text -> String`

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -266,6 +270,13 @@ scriptTxSpec = describe "script transactions" $ do
                 alwaysSuccessValidator
                 (const witness)
 
+        it "validator script provided in witness is greater \
+           \than adopted script version" $ do
+            let witness = ScriptWitness intValidatorWithVersionOne goodIntRedeemer
+            txShouldFailWithInvalidAdoptedScriptVer $ checkScriptTx
+                intValidatorWithVersionOne
+                (const witness)
+
```
v0d1ch: Why is this failing to pass ? Aren't the validator and the redeemer here both 1 ?

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
gromakovsky: There is `toString` in `Universum`.

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
gromakovsky: No, redeemer is `goodIntRedeemer` which is 0.

v0d1ch: If I use `badIntRedeemer` I get the same result ¯\_(ツ)_/¯

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
v0d1ch: wtf when using `toString $ pretty other` I get
```
    • No instance for (Buildable (Either ToilVerFailure ()))
            arising from a use of ‘pretty’
        • In the second argument of ‘($)’, namely ‘pretty other’
          In the second argument of ‘(<>)’, namely
            ‘(toString $ pretty other)’
          In the second argument of ‘(<>)’, namely
            ‘" but got: " <> (toString $ pretty other)’
```
but I see the `Buildable` instance defined so why this happens ?

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
gromakovsky: `badIntRedeemer` also has version 0. Once again, both validator and redeem ought to have the same version (1 in this case).

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
gromakovsky: > but I see the `Buildable` instance defined so why this happens ?

It's defined for `ToilVerFailure`, but probably not for `Either`.
You can use `toString $ either pretty show other` here. If it's `ToilVerFailure`, it will apply `pretty`, otherwise it will apply `show` to `()` which will be `"()"`.

v0d1ch:
Should be all green now

txp/Pos/Script/Examples.hs
```diff
@@ -80,6 +81,16 @@ intValidator = fromE $ parseValidator [text|
...
 
+intValidatorWithVersionOne :: Script
+intValidatorWithVersionOne = fromE $ parseValidatorWithVersionOne [text|
+    data Foo = { Foo }
+
+    validator : Int -> Comp Foo {
+        validator x = case !equalsInt x 1 of {
+            True  -> success Foo ;
+            False -> failure } }
+    |]
```
gromakovsky: It's copy-pasted, please avoid this copy-paste. You can make `setVersionOne` a standalone exported function and use it here.

txp/Pos/Script/Examples.hs
```diff
@@ -80,6 +81,9 @@ intValidator = fromE $ parseValidator [text|
             False -> failure } }
     |]
 
+intValidatorWithVersionOne :: Script
+intValidatorWithVersionOne = setVersionOne intValidator
+
```
v0d1ch: @gromakovsky is this what you had in mind ?

gromakovsky: Yes.

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -266,6 +270,13 @@ scriptTxSpec = describe "script transactions" $ do
                 alwaysSuccessValidator
                 (const witness)
 
+        it "validator script provided in witness is greater \
+           \than adopted script version" $ do
+            let witness = ScriptWitness intValidatorWithVersionOne intValidatorWithVersionOne
```
neongreen: This is actually a good test for one more reason: it checks that the version checks happens before running the scripts. It'd be nice to add an explicit comment about that here.

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -435,24 +446,39 @@ scriptTxSpec = describe "script transactions" $ do
...
-        "expected: Left ...: " <> show (WitnessScriptError err) <> "\n" <>
-        " but got: " <> show other
+        | otherwise -> expectationFailure $ toString $
+              "expected: " <> pretty (WitnessScriptError err) <> "\n" <>
+              " but got: " <> pretty tiwReason
+    other -> expectationFailure $ toString $
+        "expected: Left ...: " <> pretty (WitnessScriptError err) <> "\n" <>
+        " but got: " <> either pretty show other
+
+-- | Transaction should fail with a 'ToilInvalidWitness' specifically 'WitnessAdoptedScriptVerMismatch' error.
```
neongreen: minor: wrap this comment

lib/test/Test/Pos/Txp/Toil/UtxoSpec.hs
```diff
@@ -435,24 +446,39 @@ scriptTxSpec = describe "script transactions" $ do
...
+txShouldFailWithInvalidAdoptedScriptVer :: Either ToilVerFailure () -> Expectation
+txShouldFailWithInvalidAdoptedScriptVer = \case
+    Left (ToilInvalidWitness _ _ c) -> case c of
+        WitnessAdoptedScriptVerMismatch _ _ -> pass
+        WitnessScriptVerMismatch _ _ -> expectationFailure $
+            "expected: WitnessAdoptedScriptVerMismatch\n" <>
+            " but got: WitnessScriptVerMismatch"
```
neongreen: Why are you special-casing this branch? I don't like that the two fields of `WitnessScriptVerMismatch` won't be shown when tests fail.

txp/Pos/Script.hs
```diff
@@ -75,6 +76,9 @@ parseValidator t = do
         scrScript = Bi.serialize' scr,
         scrVersion = 0 }
 
+setVersionOne :: Bi Script_v0 => Script -> Script
```
neongreen: Just derive lenses for `Script` and use them here instead of doing this.

neongreen: No, wait, actually it's even simpler. Why do you need `setVersionOne` at all? Just write `intValidatorWithVersionOne = intValidator {scrVersion = 1}` at the call site.

gromakovsky: Hm, strange, I am quite sure it was used in two places earlier, so I asked to move it into a separate function to avoid (minor) duplication. Also moving it into a separate function allowed to write `setVersionOne <$> somethingBig`, but now it's not needed.

jcmincke:

PR #2273, opened 2018-01-08
[CSL-2144] Add system/shutdown endpoint

volhovm:


volhovm:
Exchanges branch is obsolete.

PR #2274, opened 2018-01-08
[CSL-2144] Add system/shutdown endpoint

volhovm:
Same as #2273. Adds endpoint that triggers node shutdown.

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Api.hs
+++ b/wallet/src/Pos/Wallet/Web/Api.hs
@@ -27,10 +27,12 @@ module Pos.Wallet.Web.Api
        , WSettingsApi    , WSettingsApiRecord(..)
        , WBackupApi      , WBackupApiRecord(..)
        , WInfoApi        , WInfoApiRecord(..)
+       , WSystemApi      , WSystemApiRecord(..)
        -- ** Something
        , WalletVerb
 
        -- * Swagger API
+
        , WalletSwaggerApi
        , swaggerWalletApi
        ) where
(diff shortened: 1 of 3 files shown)


wallet/src/Pos/Wallet/Web/Api.hs
```diff
@@ -503,6 +506,21 @@ data WBackupApiRecord route = WBackupApiRecord
...
+--   /system
+-- ~~~~~~~~~~
+
+type WSystemApi = "system" :> ToServant (WSystemApiRecord AsApi)
+
+data WSystemApiRecord route = WSystemApiRecord
+  {
+    _requestShutdown :: route
+    :- "shutdown"
+    :> Summary "Import full information about wallet from a given file."
```
pva701: ahaha, it seems you copied this from import endpoint and forgot to rewrite, :lalka:

volhovm: Lol, nice.

gromakovsky: ahaha loh ppz mda :kaby:

wallet/src/Pos/Wallet/Web/Methods/Misc.hs
```diff
@@ -113,6 +117,22 @@ postponeUpdate = removeNextUpdate >> return NoContent
...
+-- needed in order for http request to succeed.
+requestShutdown ::
+       ( MonadIO m
+       , MonadReader ctx m
+       , WithLogger m
+       , HasShutdownContext ctx
+       , Mockables m [Async, Delay]
+       )
+    => m NoContent
+requestShutdown = NoContent <$ async (delay (sec 1) >> triggerShutdown)
```
avieth: This is dodgy.

Why does `triggerShutdown` cause the web server to fail to send a response?

PR #2278, opened 2018-01-08
[CSL-2124] Rebase GH#2187 onto master

int-index:
Based on #2187

this is a remake of #2264 that had to be done because #2251 was merged before it


adinapoli-iohk:
We don't need the full payload, but just to cherry-pick these two:

- [ ] Search signer by address using map, not list
- [ ] More logs in `getHistory`, remove huge ones

Will take ownership of cherry-picking just those from this PR.

adinapoli-iohk:
This can be closed as we have backported the relevant bits.

PR #2280, opened 2018-01-08
[CSL-2145] fix memory leak in time-warp stats

avieth:
This doesn't appear in actual deployments because the EKG / statsd servers force these values periodically. I noticed it while profiling a special setup which doesn't run these servers.


avieth:
Uncontroversial patch here. Can I get an approval? @dcoutts

avieth:
> Is there not a strict version of modifySharedAtomic like there is for atomicModifyIORef'?

There is not.

PR #2283, opened 2018-01-08
[CSL-2135] Add endpoint to cancel unconfirmed pending transactions

kantp:
Part of the backport of #2213

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Api.hs
+++ b/wallet/src/Pos/Wallet/Web/Api.hs
@@ -41,6 +41,8 @@ module Pos.Wallet.Web.Api
 
        , NewPayment
        , TxFee
+       , CancelApplyingPtxs
+       , CancelSpecificApplyingPtx
        , UpdateTx
        , GetHistory
        , GetPendingTxsSummary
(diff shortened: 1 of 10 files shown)


wallet/src/Pos/Wallet/Web/Api.hs
```diff
@@ -298,6 +300,19 @@ type UpdateTx =
     :> ReqBody '[JSON] CTxMeta
     :> WRes Post ()
 
+type CancelApplyingPtxs =
+       "txs"
+    :> "resubmission"
+    :> "cancel"
+    :> WRes Get ()
```
adinapoli-iohk: This was not very clear from the spreadsheet, but these two endpoints are not RESTful-abiding and really they should be `POST`, not `Get`, as they are not idempotent 😉

kantp: Ah, I see. Let me do that, and resolve the merge conflicts!

kantp:
Ok, turned them into POST, and rebased on `master`, squashing the commits together.

PR #2289, opened 2018-01-09
Hotfix [CSL-1883]

sectore:
Backport of #2272 to `master`

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Methods/Payment.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/Payment.hs
@@ -46,7 +46,7 @@ import           Pos.Util.LogSafe                 (logInfoS)
 import           Pos.Wallet.KeyStorage            (getSecretKeys)
 import           Pos.Wallet.Web.Account           (GenSeed (..), getSKByAddressPure,
                                                    getSKById)
-import           Pos.Wallet.Web.ClientTypes       (AccountId (..), Addr, CAddress (..),
+import           Pos.Wallet.Web.ClientTypes       (AccountId (..), Addr,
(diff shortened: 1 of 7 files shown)


wallet/src/Pos/Wallet/Web/Methods/Payment.hs
```diff
@@ -69,14 +72,30 @@ newPayment
...
 
+newPaymentBatch
+    :: MonadWalletWebMode m
+    => SendActions m
+    -> PassPhrase
+    -> NewBatchPayment
+    -> m CTx
+newPaymentBatch sa passphrase NewBatchPayment {..} = do
+    src <- decodeCTypeOrFail npbFrom
+    notFasterThan (6 :: Second) $
```
dcoutts: Hmm, check that we've not merged this already. I thought that we already had `notFasterThan` in master and were using it for `newPayment`.

adinapoli-iohk: @dcoutts As surprising as it seems, we don't have (it might be introduced by some of the ported outstanding PRs, but not the the current tip):

screen shot 2018-01-10 at 15 30 56

dcoutts:
Yes lets squash as @adinapoli-iohk suggests.

wallet/src/Pos/Wallet/Web/Methods/Payment.hs
dcoutts: Ok, find it in the exchanges branch. We need to include that change.

avieth: Do we really need this? We already have rate limiting in the dequeue policy. @dcoutts

PR #2303, opened 2018-01-10
[CSL-2135] Add a newly created tx to the mempool before submitting it

adinapoli-iohk:
Rebase of #2270 .

Alas I couldn't preserve the original commit as cherry picking would case `git` to fail with `fatal: bad object`. I hope that's ok for everybody.


adinapoli-iohk:
>  you could pass

Oh, TIL, thanks @gromakovsky ! CI is failing, so I'm still on time to give people proper credit 😉

PR #2310, opened 2018-01-11
[CSL-738] Use MonadThrow, MonadCatch, MonadMask

int-index:


lrc/Pos/Lrc/Mode.hs
```diff
@@ -22,7 +22,7 @@ type LrcMode ctx m
        , MonadGState m
        , MonadDB m
        , MonadIO m
-       , Mockables m [Async, Bracket, Concurrently, Delay]
+       , Mockables m [Async, Concurrently, Delay]
                      -- ^ alphabet for the youngest haskellers
```
gromakovsky: You've broken the alphabet :(
Can you delete this comment? ;d

auxx/src/Repl.hs
```diff
@@ -89,10 +88,10 @@ createAuxxRepl = do
     lastResultVar <- newEmptyMVar
     (printActionVar :: MVar (PrintAction IO)) <- newEmptyMVar
     let
-        withCommand :: forall m. (MonadIO m, Mockable Catch m) => (Text -> m ()) -> m ()
+        withCommand :: forall m. (MonadIO m, MonadCatch m) => (Text -> m ()) -> m ()
         withCommand cont = do
             cmd <- takeMVar nextCommandVar
-            res <- handle
+            res <- handleAsync
```
gromakovsky: Is it intentional or you just want to preserve existing logic and change it to `handleAny` later?
_perhaps I can figure that out by reading the code, but I don't want to ;d_

networking/src/Mockable/Concurrent.hs
```diff
@@ -46,10 +46,11 @@ module Mockable.Concurrent (
 
   ) where
 
-import           Control.Exception (AsyncException (ThreadKilled), Exception)
+import           Control.Exception (AsyncException (..))
+import           Control.Exception.Safe hiding (throwTo)
+--import           Control.Exception (AsyncException (ThreadKilled), Exception)
```
gromakovsky: First, I see implicit imports, second, is this commented out line needed?

networking/src/Mockable/Concurrent.hs
```diff
@@ -121,13 +122,14 @@ data Async m t where
     AsyncThreadId :: Promise m t -> Async m (ThreadId m)
     Race :: m t -> m r -> Async m (Either t r)
     Link :: Promise m t -> Async m ()
+    UnsafeUnmask :: m a -> Async m a
```
gromakovsky: What's the story about `UnsafeUnmask`? Why did you add it?

auxx/src/Repl.hs
int-index: It is intentional: when a command is interrupted with an async exception, we do not want the REPL to crash.

networking/src/Mockable/Concurrent.hs
int-index: It was a part of `Mockable Bracket` before, and it is used in some of `Mockable Async` combinators.

gromakovsky: Ok.

networking/src/Mockable/Concurrent.hs
int-index: this is oops

PR #2311, opened 2018-01-11
Cosmetic changes

Pastafarianist:
* Fixed lots of typos.
* Files under `scripts/tls-files` no longer match with `.gitignore`.
* Renamed `Pos.Lrc.Class` to `Pos.Lrc.RichmenComponent`.
* Improved descriptions of some modules.
* Removed commented-out code.


ssc/Pos/Ssc/Error/Seed.hs
```diff
@@ -1,4 +1,4 @@
--- | Error which can arise during seed calculation.
+-- | Possible failures during SSC seed calculation.
```
neongreen: Shared seed calculation seed calculation :trollface: But okay, it's clearer this way.

wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -40,7 +40,7 @@ englishWords = [ "recycle" , "child" , "universe" , "extend" , "edge" , "tourist
                , "color" , "screen" , "birth" , "detect" , "summer" , "palm"
                , "entry" , "swing" , "fit" , "garden" , "trick" , "timber"
                , "toss" , "atom" , "kitten" , "flush" , "master" , "transfer"
-               , "success" , "worry" , "rural" , "silver" , "invest" , "mean "
+               , "success" , "worry" , "rural" , "silver" , "invest" , "mean"
```
Pastafarianist: This was confirmed by @adinapoli-iohk as being correct.

neongreen: This might break something. Not sure.

gromakovsky: Yes, I am also afraid of breakages.

neongreen:
> This was confirmed by @adinapoli-iohk as being correct.

Okay, fine then.

adinapoli-iohk:
> I'm worried about the "mean " thing and would like someone more knowledgeable to confirm that it's alright.

Yep, that was copied & pasted by me when trying to expose a more sensible example for our Swagger V1 API, it won't break anything 😉

ssc/Pos/Ssc/Error/Seed.hs
Pastafarianist: * `shared seed computation seed calculation`, I guess :)

PR #2313, opened 2018-01-11
[CSL-2155] Merge parts of master into develop

sectore:
Part of https://iohk.myjetbrains.com/youtrack/issue/CSL-2155 to merge parts of `master` into `develop` step by step :footprints:


client/src/Pos/Client/Txp/Util.hs
```diff
@@ -145,14 +152,14 @@ isCheckedTxError = \case
 
 -- | Specifies the way Utxos are going to be grouped.
 data InputSelectionPolicy
-    = OptimizeForSecurity  -- ^ Spend everything from the address
-    | OptimizeForSize      -- ^ No grouping
+    = OptimizeForSecurity       -- ^ Spend everything from the address
+    | OptimizeForHighThroughput -- ^ No grouping, prefer confirmed addresses
```
adinapoli-iohk: This is fine, as users are using `OptimizeForSecurity` by default and technically `OptimiseForHighThroughput` is a better policy for exchanges anyway.

wallet/src/Pos/Wallet/Web/Api.hs
```diff
@@ -366,6 +365,12 @@ data WTxsApiRecord route = WTxsApiRecord
     :> QueryParam "skip" ScrollOffset
     :> QueryParam "limit" ScrollLimit
     :> WRes Get ([CTx], Word)
+
+  , _pendingSummary :: route
+    :- "pending"
+    :> "summary"
+    :> Summary "Get the pending tx summeries."
```
adinapoli-iohk: Typo! 😉

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -129,14 +137,19 @@ getHistoryLimited mCWalId mAccId mAddrId mSkip mLimit = do
...
+
+    logCTxs "Total last 20" $ take 20 sortedTxh
     pure (applySkipLimit sortedTxh, n)
   where
     sortByTime :: [(CTx, POSIXTime)] -> [CTx]
     sortByTime thsWTime =
         -- TODO: if we use a (lazy) heap sort here, we can get the
         -- first n values of the m sorted elements in O(m + n log m)
         map fst $ sortWith (Down . snd) thsWTime
+    forceList l = length l `seq` l
```
adinapoli-iohk: I don't recall this being in `master`, why was it added?

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -215,3 +230,11 @@ logTxHistory desc = do
         . sformat (stext%" transactions history: "%listJson) desc
         . map _thTxId
         . toList
+
+logCTxs
+    :: (Container t, Element t ~ CTx, WithLogger m)
+    => Text -> t -> m ()
+logCTxs desc =
```
adinapoli-iohk: Maybe we shouldn't stop the PR from being merged, but I think these kind of functions can backfire really rapidly if we do not limit the number of elements logged _internally_, rather than externally.

wallet/src/Pos/Wallet/Web/Methods/History.hs
gromakovsky: Well, there is the same `forceList` in `master`: https://github.com/input-output-hk/cardano-sl/blob/master/wallet/src/Pos/Wallet/Web/Methods/History.hs#L151
It was merged in this PR (to `master`): https://github.com/input-output-hk/cardano-sl/pull/2287/files#diff-1bfb54b702d6d93310adefeed510d053R150

wallet/src/Pos/Wallet/Web/Methods/History.hs
gromakovsky: It's a valid concern, but not for this PR, because the only goal here is to merge `master` to `develop`.

wallet/src/Pos/Wallet/Web/Methods/History.hs
adinapoli-iohk: ha! And I was the author of the original port nevertheless 😂  I'm not a performance guru but I think this `forceList` is absolutely useless, as it would force to WHNF just the outermost function application, thus we would go from a single thunk to `(:) <thunk> <thunk>`. Anyway, not for this PR I guess..

gromakovsky:
@adinapoli-iohk anything else you want me to fix here? I've fixed the typo you spotted, clarified one concern, another one is not for this PR and lastly there is:
> This is fine, as users are using OptimizeForSecurity by default and technically OptimiseForHighThroughput is a better policy for exchanges anyway.

Which doesn't sound like a concern, but maybe I just didn't get it and it implies that I should do something :)

If everything is ok, please approve 😉

PR #2314, opened 2018-01-11
[CSL-2014] Report misbehaviors to EKG metrics

hithroc:


infra/Pos/Reporting/MemState.hs
```diff
@@ -2,21 +2,39 @@
...
+import           System.Metrics.Counter (Counter)
 import           Control.Lens (makeLenses)
 import           System.Wlog.LoggerConfig (LoggerConfig)
 import           Universum
 
+-- | EKG metric values for misbehaviors
+data MisbehaviorMetrics = MisbehaviorMetrics
+    { _mmRollbacks :: Gauge
+    , _mmSscFailures :: Counter
+    , _mmIgnoredCommitments :: Gauge
```
gromakovsky: Can you document all three fields please?

block/src/Pos/Block/Network/Logic.hs
```diff
@@ -561,6 +563,11 @@ applyWithRollback nodeId enqueue toApply lca toRollback = do
     reportRollback = do
         let rollbackDepth = length toRollback
         let isCritical = rollbackDepth >= criticalForkThreshold
+
+        -- Commit rollback value to EKG
+        whenJustM (view misbehaviorMetrics) $ liftIO .
+            flip Metrics.set (fromIntegral rollbackDepth) . _mmRollbacks
```
gromakovsky: But wait, the issue description suggests that `_mmRollbacks` gauge corresponds to
> the maximal rollback depth in last T (where is configurable time value, e. g. few hours).

infra/Pos/Reporting/Methods.hs
```diff
@@ -343,7 +343,8 @@ reportNode sendLogs extendWithNodeInfo reportType =
 reportMisbehaviour
     :: forall ctx m . (MonadReporting ctx m)
     => Bool -> Text -> m ()
-reportMisbehaviour isCritical message = do
+reportMisbehaviour isCritical message = when False $ do -- TODO: Add a flag wether to report
```
gromakovsky: You could write that famous
```
reportMisbehaviour isCritical message = don't $ do
    bla
  where
    don't _ = pure ()
```

:)

lib/src/Pos/Launcher/Resource.hs
```diff
@@ -293,6 +295,12 @@ allocateNodeContext ancd txpSettings = do
     let slotDuration :: Integer
         slotDuration = toMicroseconds . bvdSlotDuration $ gdBlockVersionData genesisData
     ncSubscriptionKeepAliveTimer <- newTimer $ 3 * fromIntegral slotDuration
+
+    _mmRollbacks <- liftIO $ Metrics.createGauge "cardano.BlockRollbacks" ekgStore
+    _mmSscFailures <- liftIO $ Metrics.createCounter "cardano.SCCComputationFailures" ekgStore
+    _mmIgnoredCommitments <- liftIO $ Metrics.createGauge "cardano.IgnoredCommitments" ekgStore
+    let mm = MisbehaviorMetrics {..}
```
gromakovsky: I'd define `MonadIO m => EKG.Store -> m MisbehaviorMetrics` in `Pos.Reporting.MemState` (near `MisbehaviorMetrics` definition).
Also please use `Pos.System.Metrics.Constants.withCardanoNamespace` for metric names.

lib/src/Pos/Launcher/Resource.hs
```diff
@@ -301,6 +309,7 @@ allocateNodeContext ancd txpSettings = do
             , ncNodeParams = np
             , ncTxpGlobalSettings = txpSettings
             , ncNetworkConfig = networkConfig
+            , ncMisbehaviorMetrics = Just mm
```
gromakovsky: I might be missing something, but it looks like `ncMisbehaviorMetrics` is always `Just`. Why is it `Maybe`?

hithroc: `Pos.Reporting.MemState` has `emptyReportingContext :: ReportingContext`, which is used in some tests and other things that I haven't dived into. That's where `Nothing` for misbehavior metrics is used.

block/src/Pos/Block/Network/Logic.hs
gromakovsky: We decided that we don't need it and this logic is ok.

lib/src/Pos/Launcher/Resource.hs
gromakovsky: Ah, so it's needed for `setter`, ok!

PR #2315, opened 2018-01-11
[CSL-2176] Respect some invariants

neongreen:
I recommend reviewing each commit individually.

### DIFF
--- a/core/Pos/Core/Genesis/Canonical.hs
+++ b/core/Pos/Core/Genesis/Canonical.hs
@@ -145,14 +145,14 @@ instance Monad m => ToJSON m Address where
     toJSON = fmap JSString . toObjectKey
 
 instance Monad m => ToJSON m (ProxySecretKey EpochIndex) where
-    toJSON ProxySecretKey {..} =
+    toJSON psk =
         -- omega is encoded as a number, because in genesis we always
         -- set it to 0.
         mkObject
-            [ ("omega", pure (JSNum . fromIntegral $ pskOmega))
-            , ("issuerPk", toJSON pskIssuerPk)
-            , ("delegatePk", toJSON pskDelegatePk)
-            , ("cert", toJSON pskCert)
+            [ ("omega", pure (JSNum . fromIntegral $ pskOmega psk))
+            , ("issuerPk", toJSON $ pskIssuerPk psk)
+            , ("delegatePk", toJSON $ pskDelegatePk psk)
+            , ("cert", toJSON $ pskCert psk)
             ]
 
 instance Monad m => ToJSON m SoftforkRule where
@@ -386,7 +386,7 @@ instance ReportSchemaErrors m => FromJSON m (ProxySecretKey EpochIndex) where
         pskIssuerPk <- fromJSField obj "issuerPk"
         pskDelegatePk <- fromJSField obj "delegatePk"
         pskCert <- fromJSField obj "cert"
-        return ProxySecretKey {..}
+        pure UnsafeProxySecretKey{..}
 
 instance ReportSchemaErrors m => FromJSON m SoftforkRule where
     fromJSON obj = do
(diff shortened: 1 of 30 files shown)


crypto/Pos/Crypto/Signing/Types/Signing.hs
```diff
@@ -140,7 +140,9 @@ instance B.Buildable (ProxyCert w) where
 
 -- | Convenient wrapper for secret key, that's basically ω plus
 -- certificate.
-data ProxySecretKey w = ProxySecretKey
+--
+-- The invariant is that the certificate is valid.
```
volhovm: What is "valid"? You should add "with respect to `validateProxySecretKey`" I think.

core/Pos/Core/Genesis/Canonical.hs
```diff
@@ -386,7 +386,7 @@ instance ReportSchemaErrors m => FromJSON m (ProxySecretKey EpochIndex) where
         pskIssuerPk <- fromJSField obj "issuerPk"
         pskDelegatePk <- fromJSField obj "delegatePk"
         pskCert <- fromJSField obj "cert"
-        return ProxySecretKey {..}
+        pure UnsafeProxySecretKey{..}
```
volhovm: Why are you sure that this psk is valid?

delegation/src/Pos/Delegation/Logic/Mempool.hs
```diff
@@ -195,7 +200,6 @@ processProxySKHeavyInternal psk = do
 
         let res = if | cached -> PHCached
                      | not coherent -> PHTipMismatch
-                     | not consistent -> PHBroken
```
volhovm: You should delete the related constructor also.

crypto/Pos/Crypto/Signing/Signing.hs
```diff
@@ -8,21 +8,20 @@ module Pos.Crypto.Signing.Signing
...
+       , checkSig                      -- reexport
        , fullSignatureHexF
        , parseFullSignature
-
        , mkSigned
 
        -- * Versions for raw bytestrings
        , signRaw
-       , verifyRaw
+       , checkSigRaw                   -- reexport
```
volhovm: So all of this was duplicated? ..
Does https://iohk.myjetbrains.com/youtrack/issue/CSL-1157 have any relation to it?

update/Pos/Update/Logic/Local.hs
```diff
@@ -395,6 +395,8 @@ finishPrepare badProposals proposals votes = do
...
         | otherwise =
             deactivateProposal upId
-    isVoteValid UpdateVote {..} =
-        (not (HS.member uvProposalId badProposals) &&) . isJust <$>
-        getProposal uvProposalId
+    isVoteValid vote = do
+        let id = uvProposalId vote
+        proposalIsPresent <- isJust <$> getProposal id
+        pure $ not (HS.member id badProposals) && proposalIsPresent
+
```
volhovm: Trailing line.

crypto/Pos/Crypto/Signing/Signing.hs
neongreen: No relation.

I'm not sure what you mean by “duplicated”.

core/Pos/Core/Genesis/Canonical.hs
neongreen: I'm not sure it's valid, but stuff like `VssCertificate` isn't checked in `Pos.Core.Genesis.Canonical` either so I assumed that whoever wrote `Pos.Core.Genesis.Canonical` knew what they were doing. Maybe it was a wrong assumption in hindsight. If you want, I can change it.

volhovm:
I need these changes to implement CSL-2072 inside PR #2324, so I'm merging it.

PR #2316, opened 2018-01-11
Use Nix' openssl in connect-to-cluster

puffnfresh:
I don't have openssl on my PATH, so the Cardano web server wasn't starting for me.


domenkozar:
Thanks! Huh, this was fixed by me before, let me see why it didn't make to develop.

PR #2317, opened 2018-01-11
[CSL-2168] Use combinators from 'safe-exceptions'

int-index:


block/src/Pos/GState/SanityCheck.hs
```diff
@@ -6,7 +6,7 @@ module Pos.GState.SanityCheck
 
 import           Universum
 
-import           Control.Monad.Catch (MonadMask)
+import           Control.Exception.Safe (MonadMask)
```
volhovm: I do not quite get it. `Universum` experts `MonadMask` from `exceptions-safe` already.
https://github.com/serokell/universum/blob/master/src/Universum/Exception.hs#L18

generator/src/Pos/Generator/Block/Mode.hs
```diff
@@ -146,7 +146,7 @@ type BlockGenMode ext m = ReaderT (BlockGenContext ext) m
 type BlockGenRandMode ext g m = RandT g (BlockGenMode ext m)
 
 instance MonadThrow m => MonadThrow (RandT g m) where
-    throwM = lift . throwM
+    throwM = lift . UnsafeExc.throwM
```
volhovm: Wait for it, so you define safe `MonadThrow` instance using lifted unsafe `throwM`. For what reason?

networking/bench/LogReader/Main.hs
```diff
@@ -1,10 +1,9 @@
 {-# LANGUAGE TypeApplications #-}
 
 import           Control.Applicative (empty, (<|>))
-import           Control.Exception (Exception)
+import           Control.Exception.Safe (Exception, handle)
```
volhovm: `Exception` should be imported from `Universum` already.

util/Pos/Util/Filesystem.hs
```diff
@@ -107,13 +107,13 @@ withTempFile :: (MonadIO m, MonadMask m)
...
-  MC.bracket
+  E.bracket
     (liftIO (openTempFile tmpDir template))
     (\(name, handle) -> liftIO (hClose handle >> ignoringIOErrors (removeFile name)))
     (uncurry action)
   where
-     ignoringIOErrors :: MC.MonadCatch m => m () -> m ()
-     ignoringIOErrors ioe = ioe `MC.catch` (\e -> const (return ()) (e :: Prelude.IOError))
+     ignoringIOErrors :: E.MonadCatch m => m () -> m ()
+     ignoringIOErrors ioe = ioe `E.catch` (\e -> const (return ()) (e :: Prelude.IOError))
```
volhovm: Both `catch` and `bracket` from universum are safe, `MonadMask` and `MonadCatch` are too.

generator/src/Pos/Generator/Block/Mode.hs
int-index: I define `MonadThrow.throwM` via `MonadThrow.throwM`. The async exception handling logic is defined separately in `safe-exceptions`.

PR #2323, opened 2018-01-12
[CSL-2154]+ [CSL-2165] 'getHistory' improvements

pva701:
This PR:
1. Speeds up 'getHistoryLimited'. Now we construct `CTx` only for response history entries
2. Adds 'since' parameter to `/api/txs/histories` enpoint, which accepts time in seconds

TESTED

### DIFF
--- a/node/src/Pos/Wallet/Web/ClientTypes/Instances.hs
+++ b/node/src/Pos/Wallet/Web/ClientTypes/Instances.hs
@@ -49,8 +49,8 @@ import           Pos.Wallet.Web.ClientTypes.Types     (AccountId (..), ApiVersio
                                                        CWallet, CWalletAssurance,
                                                        CWalletInit (..), CWalletMeta (..),
                                                        CWalletRedeem (..),
-                                                       ClientInfo (..), ScrollLimit (..),
-                                                       ScrollOffset (..),
+                                                       ClientInfo (..), SinceTime (..),
+                                                       ScrollLimit (..), ScrollOffset (..),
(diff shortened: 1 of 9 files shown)


adinapoli-iohk:
Thanks @pva701 !

Just trying to anticipate the review from @dcoutts: Can we have this PR as two separate ones? the fact this PR is fixing two issues but is being bundled into one makes reviewing, merging and cherry picking harder 😉 

May I suggest:

- One single commit PR which fixes [CSL-2154]
- One single commit PR which fixes [CSL-2165]

Thanks a ton!

adinapoli-iohk:
@pva701 Ping? 😉 

I think having these as separate PRs would make more sense. Also be aware that the work @nc6 is doing on the DB consistency is also touching `getHistory`, you guys might find useful to collaborate on the shared goal 😉

adinapoli-iohk:
> regarding two separate PRs: CSL-2165 must be based on CSL-2154 PR, so does it make sense to split them?

It does in my opinion 😉  You can open the second using the first as the base branch. The idea is to do incremental changes, so we can test them in isolation -- this will be very important going forward with the QA team and the work Anatoli is doing 😉 

Thanks a lot @pva701 !

pva701:
so, I closed this PR, and opened two others:
https://github.com/input-output-hk/cardano-sl/pull/2414
https://github.com/input-output-hk/cardano-sl/pull/2415

the first one should be reviewed and merged at first because the second one is based on the first

PR #2328, opened 2018-01-13
[CSL-2021] Make logs less verbose in recovery mode

hithroc:


block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -343,11 +343,13 @@ getProcessBlocks enqueue nodeId lcaChild newestHash = do
                 logWarning msg
                 throwM $ DialogUnexpected msg
             Right blocks -> do
+                let hashes = getOldestFirst $ map (headerHash . view blockHeader) blocks
```
gromakovsky: Btw, `view blockHeader` is most likely redundant here, because `headerHash` is very polymorphic.

block/src/Pos/Block/Network/Logic.hs
```diff
@@ -491,8 +493,11 @@ applyWithoutRollback
     -> OldestFirst NE Block
     -> m ()
 applyWithoutRollback enqueue blocks = do
-    logInfo $ sformat ("Trying to apply blocks w/o rollback: "%listJson) $
-        fmap (view blockHeader) blocks
+    let headers = getOldestFirst $ map (view blockHeader) blocks
+    logInfo $ sformat ("Trying to apply blocks w/o rollback. First headers: "%listJson
+                      %"\nLast headers: "%listJson)
+        (NE.take 3 headers)
+        (takeLast 3 headers)
```
gromakovsky: It's a bit more complicated. If there are less than 6 headers, some of them will be printed more than once which is not what we want. There's the same problem below.

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -222,8 +223,11 @@ classifyHeaders inRecovery headers = do
     uselessGeneral =
         CHsUseless "Couldn't find lca -- maybe db state updated in the process"
     processClassify tipHeader = runMaybeT $ do
+        let hashes = getNewestFirst $ map (view headerHashG) headers
         lift $ logDebug $
-            sformat ("Classifying headers: "%listJson) $ map (view headerHashG) headers
+            sformat ("Classifying headers: "%buildListBounds)
```
gromakovsky: I think it'd be good to write `Classyfying headers (newest first):`, because newest first order may be a bit confusing (at least I usually expect oldest headers first).

block/src/Pos/Block/Network/Logic.hs
```diff
@@ -491,8 +493,13 @@ applyWithoutRollback
     -> OldestFirst NE Block
     -> m ()
 applyWithoutRollback enqueue blocks = do
-    logInfo $ sformat ("Trying to apply blocks w/o rollback: "%listJson) $
-        fmap (view blockHeader) blocks
+    let headers = getOldestFirst $ map (view blockHeader) blocks
+    let fmtString = "Trying to apply blocks w/o rollback." :: Format r r
+    if length headers <= 6
```
hithroc: Probably should put this into a utility function

gromakovsky: Yes, please at least eliminate copy-paste in this file (move this few lines into a separate function).

util/Pos/Util/Util.hs
```diff
@@ -267,3 +274,7 @@ median l = NE.sort l NE.!! middle
 -}
 sleep :: MonadIO m => NominalDiffTime -> m ()
 sleep n = liftIO (threadDelay (truncate (n * 10^(6::Int))))
+
+-- | Formats two values as first and last elements of a list
+buildListBounds :: (Buildable a, Buildable b) => F.Format r (a -> b -> r)
+buildListBounds = "[" F.% F.build F.% ".." F.% F.build F.% "]"
```
volhovm: I'm not really sure about `b /= a` cases.

block/src/Pos/Block/Network/Logic.hs
```diff
@@ -535,8 +536,7 @@ applyWithRollback
     -> NewestFirst NE Blund
     -> m ()
 applyWithRollback nodeId enqueue toApply lca toRollback = do
-    logInfo $ sformat ("Trying to apply blocks w/ rollback: "%listJson)
-        (map (view blockHeader) toApply)
+    logBounds 6 "Trying to apply blocks w/ rollback." . getOldestFirst . map (view blockHeader) $ toApply
```
gromakovsky: Too long line.

util/Pos/Util/Util.hs
gromakovsky: Do you mean `a == b` case?

gromakovsky: Ah, you mean types, not values. I mean it prints a strange thing when value of `a` equals to the value of `b`.

hithroc: >print single thing only.

I'm fairly sure you can't change amount of things you want to print with `Format` alone. You'd have to have an actual printing function that has such logic inside it.

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -223,7 +223,8 @@ classifyHeaders inRecovery headers = do
         CHsUseless "Couldn't find lca -- maybe db state updated in the process"
     processClassify tipHeader = runMaybeT $ do
         lift $ logDebug $
-            sformat ("Classifying headers: "%listJson) $ map (view headerHashG) headers
+            sformat ("Classifying headers (newest first): "%buildListBounds) $
+                getNewestFirst $ map (view headerHashG) headers
```
ChShersh: Do we still using `formatting` and not `fmt`?

hithroc: *Shrug* The code I was changing clearly uses `formatting`.

PR #2333, opened 2018-01-15
[CSL-2184] Get rid of forks in NTP.Client

gromakovsky:
I slightly modified some logic to make it possible.
I think this code is still not perfect and there are other things
to improve, but within this issue I am only concerned about `forkIO` usages.
Notable changes (other than actual eliminating `forkIO`):
1. I removed stop button, because it's not needed anymore.
2. I removed `ncClosed`, because it seems to be not needed, because
 with current interface it seems to be impossible to access `NtpClient` after
 it's closed.
3. I moved `modifyLoggerName` to `spawnNtpClient` and it allowed to remove
ad-hoc `log` and `log'` functions.

### DIFF
--- a/networking/src/NTP/Client.hs
+++ b/networking/src/NTP/Client.hs
@@ -1,47 +1,45 @@
 {-# LANGUAGE ConstraintKinds     #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE NoImplicitPrelude   #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 -- | This module implements functionality of NTP client.
 
 module NTP.Client
-    ( startNtpClient
+    ( NtpMonad
+    , spawnNtpClient
     , NtpClientSettings (..)
-    , NtpStopButton (..)
     , ntpSingleShot
     , hoistNtpClientSettings
     ) where
 
-import           Control.Concurrent.STM (atomically, modifyTVar')
-import           Control.Concurrent.STM.TVar (TVar, newTVarIO, readTVar, readTVarIO, writeTVar)
-import           Control.Exception.Safe (Exception, MonadMask, bracketOnError, catchAny, handleAny,
-                                         throwM)
+import           Universum
+
+import           Control.Concurrent.STM (modifyTVar')
+import           Control.Concurrent.STM.TVar (TVar, readTVar)
+import           Control.Exception.Safe (Exception, MonadMask, catchAny, handleAny)
 import           Control.Lens ((%=), (.=), _Just)
-import           Control.Monad (forM_, forever, unless, void, when)
+import           Control.Monad (forever)
 import           Control.Monad.State (gets)
-import           Control.Monad.Trans (MonadIO (..))
 import           Control.Monad.Trans.Control (MonadBaseControl)
 import           Data.Binary (decodeOrFail, encode)
 import qualified Data.ByteString.Lazy as LBS
 import           Data.Default (Default (..))
-import           Data.List (sortOn)
+import           Data.List (sortOn, (!!))
 import           Data.Maybe (catMaybes, isNothing)
-import           Data.Monoid ((<>))
-import           Data.Text (Text)
 import           Data.Time.Units (Microsecond, Second, toMicroseconds)
 import           Data.Typeable (Typeable)
 import           Formatting (sformat, shown, (%))
 import           Network.Socket (AddrInfo, SockAddr (..), Socket, addrAddress, addrFamily, close)
 import           Network.Socket.ByteString (recvFrom, sendTo)
-import           Prelude hiding (log)
 import           Serokell.Util.Concurrent (modifyTVarS, threadDelay)
-import           System.Wlog (LoggerName, Severity (..), WithLogger, logMessage, modifyLoggerName)
-import           Universum (whenJust)
+import           System.Wlog (LoggerName, WithLogger, logDebug, logError, logInfo, logWarning,
+                              modifyLoggerName)
 
 import           Mockable.Class (Mockable)
-import           Mockable.Concurrent (Delay, Fork, delay, fork)
+import           Mockable.Concurrent (Async, Concurrently, concurrently, forConcurrently, race)
 import           NTP.Packet (NtpPacket (..), evalClockOffset, mkCliNtpPacket, ntpPacketSize)
 import           NTP.Util (createAndBindSock, resolveNtpHost, selectIPv4, selectIPv6,
                            udpLocalAddresses, withSocketsDoLifted)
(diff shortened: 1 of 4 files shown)


networking/src/NTP/Client.hs
```diff
@@ -280,22 +277,19 @@ stopNtpClient cli = do
...
-            void . fork . withSocketsDoLifted $ startReceive cli
-            void . fork . withSocketsDoLifted $ startSend addrs cli
-            log cli Info "Launched"
-
-        return $ NtpStopButton $ stopNtpClient cli
+        when (null addrs) $ throwM NoHostResolved
+        withAsync (withSocketsDoLifted $ startReceive cli) $ \_ ->
+            withAsync (withSocketsDoLifted $ startSend addrs cli) $ \_ -> do
+                log cli Info "Launched NTP client"
+                callback (NtpStopButton $ stopNtpClient cli)
```
Martoon-00: Since it's now written this way, it seems better to press stop button here (which will close sockets) rather than export button outside

networking/src/NTP/Client.hs
```diff
@@ -280,22 +277,19 @@ stopNtpClient cli = do
...
-        if null addrs then
-            throwM NoHostResolved
-        else do
-            void . fork . withSocketsDoLifted $ startReceive cli
-            void . fork . withSocketsDoLifted $ startSend addrs cli
-            log cli Info "Launched"
-
-        return $ NtpStopButton $ stopNtpClient cli
+        when (null addrs) $ throwM NoHostResolved
+        withAsync (withSocketsDoLifted $ startReceive cli) $ \_ ->
```
Martoon-00: `withSocketsDo` uses global counter to track whether Windows socket library should be attached, so one `withSocketsDoLifted` in the beginning would be enough

networking/src/NTP/Client.hs
```diff
@@ -1,47 +1,46 @@
...
 
-import           Control.Concurrent.STM (atomically, modifyTVar')
-import           Control.Concurrent.STM.TVar (TVar, newTVarIO, readTVar, readTVarIO, writeTVar)
-import           Control.Exception.Safe (Exception, MonadMask, bracketOnError, catchAny, handleAny,
-                                         throwM)
+import           Universum hiding (log)
+
+import           Control.Concurrent.STM (modifyTVar')
+import           Control.Concurrent.STM.TVar (TVar, readTVar, writeTVar)
+import           Control.Exception.Safe (Exception, MonadMask, bracketOnError, catchAny, handleAny)
```
Martoon-00: I didn't notice how we switched to `exceptions-safe` here (which is indeed nice, also my comment isn't related to exactly this PR, but still), looks like migration was done not fully correctly. 

For instance, 
https://github.com/input-output-hk/cardano-sl/blob/d9e2fd91bc9e1a7b8a3d933c55c5a5f37c32c6dd/networking/src/NTP/Client.hs#L278
assumed to be cleanup action, and shouldn't be interrupted by async exceptions

networking/src/NTP/Client.hs
Martoon-00: Also, this pretty code
https://github.com/input-output-hk/cardano-sl/blob/d9e2fd91bc9e1a7b8a3d933c55c5a5f37c32c6dd/networking/src/NTP/Client.hs#L256:L259
becomes unsafe because exception can fly in during `=<<` and created socket will remain unused :confused:

networking/src/NTP/Client.hs
gromakovsky: > `withSocketsDo` uses global counter to track whether Windows socket library should be attached, so one `withSocketsDoLifted` in the beginning would be enough

As you can see, old code was using it more than once.
```
            void . fork . withSocketsDoLifted $ startReceive cli
            void . fork . withSocketsDoLifted $ startSend addrs cli
```

So I decided to preserve old behaviour. Do I understand correctly that in old code usage of these `withSocketsDoLifted` was redundant as well?

networking/src/NTP/Client.hs
gromakovsky: > Also, this pretty code becomes unsafe

Was it safe before? To be honest the whole design of `NTP.Client` looks not perfect, but in this PR I only wanted to spawn threads properly.

gromakovsky:
@Martoon-00 please have a look at my last commit. I think all your concerns are now addressed. I went further simplified `NTP.Client` a bit more. Hopefully I didn't break anything, but please review it carefully.
Notable changes:
1. I removed stop button, because it's not needed anymore.
2. I removed `ncClosed`, because it seems to be not needed, because with current interface it seems to be impossible to access `NtpClient` after it's closed.
3. I moved `modifyLoggerName` to `spawnNtpClient` and it allowed to remove ad-hoc `log` and `log'` functions.
4. I ditched the idea of passing continuation to `withNtpClient`, now we have `spawnNtpClient`. The caller can use `withAsync spawnNtpClient` or use `race` or just run it in main thread, whatever.

>  forM_ sockets $ \s -> liftIO (close s) `catchAny` (const $ pure ()) 

This code doesn't exist anymore. Sockets are always closed as release action of `bracket`.

I also left only one `withSocketsDoLifted`.

gromakovsky:
If everything looks fine, I'll squash two commits and write a bit better commit message.

networking/src/NTP/Client.hs
Martoon-00: In old code you could not know for sure (at least at first sight) which of the threads will finish earlier, and both require socket library attached, so both are wrapped into `withSocketsDo`.
Now you can happily use only single `withSocketsDo` which wraps over all threads.

Martoon-00:
> > Also, this pretty code becomes unsafe
> Was it safe before? To be honest the whole design of NTP.Client looks not perfect, but in this PR I only wanted to spawn threads properly.
Yeees, it was not perfect. Despite this fact, previously `startSend` and `startReceive` were assumed to stop on their own and didn't expect any async exceptions (which was wrong in presence of different `MemoryOverflow` and others, but still ~plausible). But now dropping `ThreadKilled` in these functions is normal case, so async-exceptions safety should be revised at new scale.

Martoon-00:
All other things now look much better than they were, thanks for your big work!

gromakovsky:
> Yeees, it was not perfect. Despite this fact, previously startSend and startReceive were assumed to stop on their own and didn't expect any async exceptions (which was wrong in presence of different MemoryOverflow and others, but still ~plausible). But now dropping ThreadKilled in these functions is normal case, so async-exceptions safety should be revised at new scale.

Okay, I see. Anyway, I think that:
1. Probability of such situation is very small anyway, because first there should be an exception in `doReceive` and second it should be killed during `=<<` which is really unlikely to happen. I am not saying that we shouldn't care at all, but I think that probability is not much higher than the probability of `MemoryOverflow` and others.
2. I think it's an indicator of another flaw in the design on `NTP.Client` and in this PR there is no goal to address all such flaws. So I'd like to leave it for further improvements.

So, @Martoon-00, I squashed two commits, can you approve now or do you want me to do anything else here?

Martoon-00:
No, everything looks good to me.
`<access to green button granted>`

PR #2343, opened 2018-01-18
[CSL-2190] Merge .Lrc and .RichmenComponent modules

neongreen:
(Also renamed some stuff.)


block/src/Pos/Lrc/DB/Richmen.hs
```diff
@@ -11,15 +11,15 @@ module Pos.Lrc.DB.Richmen
        -- * Concrete instances
        -- ** Ssc
        , RCSsc
-       , getRichmenSsc
+       , tryGetSscRichmen
 
        -- ** US
-       , RCUs
-       , getRichmenUS
+       , RCUS
```
volhovm: You can't have more than three capital letters in a row.

PR #2344, opened 2018-01-18
Cosmetic changes

Pastafarianist:
* Import `fromAvvmPk` into `cardano-addr-convert` directly from `Pos.Crypto.Signing`. Remove redundant export from `Pos.Aeson.Genesis`.
* Remove `LockedTVar` and `RWVar` which had not been used anywhere.
* Reorder functions in `crypto/Pos/Crypto/Encryption.hs` and `tools/src/addr-convert/Main.hs`.
* Apply `stylish-haskell` to some files.
* Add/update some comments.
* Fix lots of typos and grammar issues.

If necessary, I will split this PR into smaller ones, but since none of the changes are functional, I believe it is quite easy to review.


neongreen:
Kuda trekat?

PR #2348, opened 2018-01-18
[csl-2147] subscription status

coot:


lib/src/Pos/Configuration.hs
```diff
@@ -43,8 +46,10 @@ data NodeConfiguration = NodeConfiguration
     {
       ccDefaultPeers                 :: ![Text]
       -- ^ List of default peers
-    , ccNetworkConnectionTimeout     :: !Int
-      -- ^ Network connection timeout in milliseconds
+    , ccNetworkConnectionTimeouts    :: !(NonEmpty Int)
+      -- ^ Network connection timeouts in milliseconds.  It must be an
+      -- increasing sequence.  If a timeout will happen to be to small the next
```
gromakovsky: `s/to small/too small` ?

lib/configuration.yaml
```diff
@@ -118,7 +118,7 @@ dev: &dev
 
   node: &dev_node
     defaultPeers: []
-    networkConnectionTimeout: 2000 # ms
+    networkConnectionTimeouts: [15000, 300000, 60000, 120000] #ms
```
avieth: Too many zeros in the second one. Same error below.

coot:
@avieth I updated this PR:
* merged develop branch
* reverted the timouts and added progressive timeouts in `dnsSubscriptionWorker`:
  * Scale timeout by the factor 3/2.
  * When the timeout is greater than 20 seconds error and stop trying
to subscribe.  Now the sequence of timeouts is `5` (slot duration
/ 4), `7.5`, `11.25` and `16.875`.

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -82,8 +84,12 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
         -- Try to subscribe to some peer.
         -- If they all fail, wait a while before trying again.
         subscribeToOne dnsPeersList
-        retryInterval >>= delay
-        subscribeAlts dnsPeersVar (index, alts)
+        interval <- retryInterval retryCounter
+        case interval of
+            Nothing -> logError "dnsSubscriptionWorker: DNS timeout failure"
```
avieth: Here we give up forever? Should the program quit? Should we just have a maximum delay?

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -107,10 +113,21 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
 
     -- How long to wait before retrying in case no alternative can be
     -- subscribed to.
-    retryInterval :: m Millisecond
-    retryInterval = do
+    retryInterval :: SharedAtomicT m Int -> m (Maybe Millisecond)
+    retryInterval retryCounter = do
+        counter <- modifySharedAtomic retryCounter (\c -> return (c + 1, fromIntegral c))
```
avieth: This counter contains the number of times we've retried. Do we ever reset it to 0?

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -107,10 +113,21 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
...
         slotDur <- nextSlotDuration
-        pure $ max (slotDur `div` 4) (convertUnit (5 :: Second))
+        let interval = fromMicrosecondsF
+                $ (toMicrosecondsF slotDur) / 4 * (3 ** counter) / (2 ** counter)
```
avieth: I don't follow this. Maybe put a comment?

lib/src/Pos/Logic/Types.hs
```diff
@@ -104,6 +106,9 @@ data Logic m = Logic
     , recoveryInProgress :: m Bool
 
     , securityParams     :: SecurityParams
+
+      -- action which runs whenever connection status changes
+    , connectionChangeAction :: ConnectionChangeAction m NodeId
```
avieth: This doesn't fit with the logic layer. Instead of a callback-based approach, could we have the diffusion layer offer a `TVar ConnectionStatus`? Then the *control* part of the program (which uses logic and diffusion) can grab it and use it to implement the UI.

avieth:
> When the timeout is greater than 20 seconds error and stop trying to subscribe.

This means the node will forever be disconnected. Either ensure the whole program stops, or never stop trying.

infra/Pos/Diffusion/Subscription/Dns.hs
coot: As we talked on slack, let's reset the counter the same way that the subscription timer get reset.  Is it fine to add a `retryCounter :: SharedAtomicT d Int` to the `Logic`.  It cannot be created in `runDiffusionLayerFull` because there we don't have access to the monad `d`.  Otherwise I can just use an `MVar` instead of a `SharedAtomicT`.

coot: I feel this implementation detail should not be exposed in the logic layer, and anyway we want to get rid of `Mockable` so I prefer to crate an `MVar` in `runDiffusionLayerFull`.

coot:
@avieth I updated the PR regarding all your comments, including the `TVar` which greatly simplified it!

I arbitrary chose `20s` as the max retry interval, but if you think we should use a greater value I am fine with that.  We use `15s` network timeout so maybe we should use `30s` or `45s` here.

lib/src/Pos/Diffusion/Full.hs
```diff
@@ -97,6 +97,10 @@ diffusionLayerFull networkConfig lastKnownBlockVersion transport mEkgNodeMetrics
 
         -- Timer is in microseconds.
         keepaliveTimer :: Timer <- newTimer 20000000
+        -- Retry counter.  It is reset when a new block header arrives in
+        -- @`handleBlockHeaders`@ and it is incremented when
+        -- @`dnsSubscriptionWorker`@ fails to subscribe to any node.
+        retryCounter <- newMVar 0
```
avieth: Is it right to reset it when a block comes in? Unlike the keepalive timer, I think this can be private to the subscription worker: reset it when a subscription is established.

networking/src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -339,16 +344,24 @@ setInFlightFor Packet{..} f var = liftIO $ modifyMVar_ var (return . update)
     updateInnerMap :: Map Precedence Int -> Map Precedence Int
     updateInnerMap = Map.adjust f packetPrec
 
+-- | Connection status
+data ConnectionStatus =
+      Connecting
+    | Connected
+    | ConnectionBroken
```
avieth: Is this one of any use? Isn't it always `NotConnected -> Connecting -> Connected -> Recurse` ? What would be the policy here? i.e. how long would the status remain "broken" before it becomes `NotConnected` or `Connecting`?

networking/src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -787,6 +805,7 @@ intEnqueue outQ@OutQ{..} msgType msg peers = fmap concat $
           return Nothing
         Just alt -> liftIO $ do
           sentVar <- newEmptyMVar
+          atomically $ modifyTVar qConnections $ Map.alter alterConnStatus alt
```
avieth: Use `modifyTVar'`

networking/src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1353,7 +1383,10 @@ updatePeersBucket outQ@OutQ{..} buck f = do
             applyMVar_ qInFlight    $ at nid .~ Nothing
             applyMVar_ qFailures    $ Map.delete nid
             applyMVar_ qRateLimited $ Set.delete nid
+            atomically $ modifyTVar qConnections $ Map.delete nid
             MQ.removeAllIn (KeyByDest nid) qScheduled
+          let newConns = Map.fromList $ map (,NotConnected) $ Set.elems peers
+          atomically $ modifyTVar qConnections $ flip Map.union newConns
```
avieth: `modifyTVar'`

networking/src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1353,7 +1383,10 @@ updatePeersBucket outQ@OutQ{..} buck f = do
             applyMVar_ qInFlight    $ at nid .~ Nothing
             applyMVar_ qFailures    $ Map.delete nid
             applyMVar_ qRateLimited $ Set.delete nid
+            atomically $ modifyTVar qConnections $ Map.delete nid
```
avieth: `modifyTVar'`

networking/src/Network/Broadcast/OutboundQueue.hs
avieth: Also, should do the `Map.delete` and `Map.union` in the same transaction.

networking/src/Network/Broadcast/OutboundQueue.hs
coot: It would stay broken until a request is send in which case it will change into `Connecting` -  how long it takes depends on the send policy / node being rate limited.  If there was just an error an the policy is `SendAll` the two values `ConnectionBroken` and `Connecting` will flicker.   Another strategy is to skip the `Connecting` and keep it `ConnectionBroken` until we get `Connected`.

lib/src/Pos/Diffusion/Full.hs
coot: There's no confirmation message back to the subscriber from `subscriptionListener`. Is there some other way I detect that a subscription was established?

avieth: Aha, that's a shame... there is a confirmation at a lower level: a reverse-direction network-transport connection is established via a little handshake. But the conversation action is run right away rather than waiting for that to happen.

Actually, there is this static peer data convention: a conversation doesn't run until the remote node's "peer data" (in this case, `VerInfo`) has been delivered. The idea is that you can choose what to do based on the version the peer reports. So in fact, the subscription conversation handler won't run until a connection has been established!

avieth: Sorry to appear to contradict myself, I was just thinking through it as I wrote. Resetting the counter as the first action in the subscription conversation would work. If it times out / fails for some other reason, the action won't run.

avieth: But this neglects the case of connecting to a relay which happens to be at capacity. There, the connection will appear to go through, and then close up immediately, so the retry counter would be set back to 0. That could be a problem.

coot: ok, I will check how this works tomorrow morning.  I've seen the handshake, I think it's in `nodeDispatcher`. Maybe there is a way to differentiate the two cases.

avieth: Just throwing an idea around: we could also do it based on time connected. At the start of the conversation, record the current time. When the connection is lost, grab the time and take the difference with the connection start time. The bigger this number, the shorter the delay before reconnecting, probably subject to an arbitrary threshold... if it lasted more than `n` seconds, try again right away. Otherwise, exponential backoff until at least one of them lasts more than `n` seconds.

avieth: This handles the case in which the relay boots the client because it's at capacity. That conversation will last only a few seconds at most.

coot: Yes, this will prevent the case Neil had in mind: spinning through the nodes and failing shortly.

PR #2349, opened 2018-01-18
Remove unused module

gromakovsky:


avieth:
> I guess @avieth should comment on it.

LGTM we indeed don't use store anymore. This file wasn't even in the cabal file nor was store in the dependencies.

PR #2351, opened 2018-01-19
[CSL-2195] Add option to get info about single address

Martoon-00:
Added GET `api/addresses/<accountId>/<addressId>` endpoint so that one can get info about single address.

Example:
`https://127.0.0.1:8090/api/addresses/Ae2tdPwUPEZEK5DvxPMtnTnUfQg8coWAAbNfLEvQ4GqWTe9h8d6AEkBDMce@2147483648/DdzFFzCqrhsg9ngNRhHEa49se7qEMKyubT9tcE13Fkvh8QC82trpTDsNvdQV7mg9SCZiuENkf77zrtwPXrTyGMNznUsSinPC1gb2ZCqK`

```
{
    "Right": {
        "cadId": "DdzFFzCqrhsg9ngNRhHEa49se7qEMKyubT9tcE13Fkvh8QC82trpTDsNvdQV7mg9SCZiuENkf77zrtwPXrTyGMNznUsSinPC1gb2ZCqK",
        "cadAmount": {
            "getCCoin": "37499999999166"
        },
        "cadIsUsed": false,
        "cadIsChange": false
    }
}
```

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Methods/Logic.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/Logic.hs
@@ -7,6 +7,7 @@ module Pos.Wallet.Web.Methods.Logic
        , getWallets
        , getAccount
        , getAccounts
+       , getWAddressSingle
 
        , createWalletSafe
        , newAccount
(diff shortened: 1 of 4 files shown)


wallet/src/Pos/Wallet/Web/Methods/Logic.hs
```diff
@@ -162,6 +166,20 @@ getAccounts
     => Maybe (CId Wal) -> m [CAccount]
 getAccounts = getAccountsIncludeUnready False
 
+getAccountsPaginate
+    :: MonadWalletWebMode m
+    => Maybe (CId Wal) -> Maybe (CId Addr) -> m [CAccount]
+getAccountsPaginate mwid mcid = do
+    accounts <- getAccounts mwid
```
adinapoli-iohk: Cannot we simply access the full list of accounts for an address in the `WalletStorage`, and do a Map lookup rather than getting all the addresses followed by a filter afterwards?

Martoon-00: Oh, to do it correctly we need to check both db and mempool, since where could be addresses which are made by us just recently and not yet in blocks (compare with https://goo.gl/NY3MUJ). Thus it won't be the "most easy solution", as you had said :sweat_smile:.
But I guess, it's worth going the way you mention anyway, right?

adinapoli-iohk: I think that's a trade-off. We could allow this endpoint to return a slight (but self healing) stale view of the world, considering is idempotent and is being accessed via a GET. If we relax this invariant, we can do it "the easy and efficient way". Whether or not we can relax this invariant, I guess it really depends from the use case and the expectations from this endpoint, which we should probe for 😉

Martoon-00: Yes, this makes sense. For me it looks like this endpoint would become primary one for API users along with `api/accounts` and others.
On the other hand, if we have several endpoints which return addresses, returning different views on them (from temporal point of time) may be confusing. :thinking:

adinapoli-iohk: > For me it looks like this endpoint would become primary one for API users along with api/accounts and others.

I think the primary target here are exchanges, and if we instruct them that some addresses might not show up if not yet accepted in blocks, it could be fine (or maybe not, that's why we need to probe :) )

> On the other hand, if we have several endpoints which return addresses, returning different views on them (from temporal point of time) may be confusing.

On the other hand, the state of the blockchain and the state of the wallet changes with time, so maybe that's not so absurd 😉

Martoon-00: Gotcha. I'm on it :+1:

Martoon-00:
@adinapoli-iohk Does it look better now?
I moved logic to another endpoint, in order not to spoil output with `account` details.
Description updated.

Martoon-00:
Not needed right now. Probably will be implemented within CSL-2327.

adinapoli-iohk:
@Martoon-00 It might be nice if you could pick up CSL-2327 as part of your backlog and reuse this PR as a foundation, I think it shouldn't be too much extra work :)

PR #2353, opened 2018-01-19
[CSL-2195] Add option to get info about single address

Martoon-00:
Rebased #2351


adinapoli-iohk:
@Martoon-00 This shouldn't target the exchanges branch, but `master` 😉

Martoon-00:
Ah, okay. My understanding was that we want this change now and here, and maybe even don't care much about whether it will be available in future, but things happen to be vice versa :smile: 

Let's close this PR once our known current problem is resolved

adinapoli-iohk:
@Martoon-00 Spoke with @domenkozar while is Lisbon, we agree we should simply not bother and close this PR, as we really do consider the exchanges branch dead at this point. Let's review your other PR instead 😉

PR #2154, opened 2017-12-19
[CSL-2089] Use diffusion layer everywhere

avieth:
The full diffusion layer has been wired up everywhere. No direct access to `SendActions` anymore.

A stop-gap full logic layer has been implemented in terms of the existing stack of monad constraints. The diffusion layer uses the logic layer though, not those constraints.

Happy to see that deletions outnumber additions. There's probably a bunch of stuff that can now be deleted but was not, so please point it out if you see it.

PR #2354, opened 2018-01-19
[CSL-2188] back port Pos.Subscription to Pos.Diffusion.Subscription

coot:
The PR includes the following changes:
* back port `infra/Pos/Subscription/Common.hs` and `infra/Pos/Subscription/Dns.hs`
* fixed a typo in `Pos.Util.Timer` (`timerSemaphore` instead of `timerSempaphore`)
* added a dynamic timer in `Pos.Util.DynamicTimer` (the timer duration is read from a `MonadIO` action).
* Change `Timer` interface to use `Microsecond` type: otherwise its too easy to pass a wrong value, for example from slot duration which is in `Millisecond`s.


lib/src/Pos/Diffusion/Full.hs
```diff
@@ -289,8 +292,8 @@ diffusionLayerFull networkConfigOpts mEkgStore expectLogic =
 -- | Create kademlia, network-transport, and run the outbound queue's
 -- dequeue thread.
 runDiffusionLayerFull
-    :: forall d x .
-       ( DiffusionWorkMode d, MonadFix d )
+    :: forall d ctx x .
+       ( DiffusionWorkMode d, MonadSlotsData ctx d, MonadFix d )
```
avieth: I've got to reject this. My goal is to keep all logic-layer related stuff out of constraints, and have it come from the `Logic d` value instead. The next slot duration can be derived from the mutable `BlockVersionData` on the `Logic d` interface instead.

lib/src/Pos/Diffusion/Full/Types.hs
```diff
@@ -28,6 +29,7 @@ type DiffusionWorkMode m
     = ( WithLogger m
       , CanJsonLog m
       , MonadMockable m
+      , MonadKnownPeers m
```
avieth: This must go. We don't need it. You can update the known peers by grabbing the `OutboundQueue` directly. Full diffusion layer has one in scope.

lib/src/Pos/Launcher/Resource.hs
```diff
@@ -292,9 +292,9 @@ allocateNodeContext ancd txpSettings ekgStore = do
     -- TODO synchronize the NodeContext peers var with whatever system
     -- populates it.
     peersVar <- newTVarIO mempty
-    let slotDuration :: Integer
-        slotDuration = toMicroseconds . bvdSlotDuration $ gdBlockVersionData genesisData
-    ncSubscriptionKeepAliveTimer <- newTimer $ 3 * fromIntegral slotDuration
+    let slotDuration :: Microsecond
+        slotDuration = convertUnit . bvdSlotDuration $ gdBlockVersionData genesisData
+    ncSubscriptionKeepAliveTimer <- newTimer $ 3 * slotDuration
```
avieth: This is not the same timer that the diffusion layer uses. Fair enough, given that in `develop`, the diffusion layer isn't actually used.

Perhaps this should instead be back-ported on top of https://github.com/input-output-hk/cardano-sl/pull/2154

lib/src/Pos/Diffusion/Full.hs
coot: You're right, I'll fix it.

lib/src/Pos/Diffusion/Full/Types.hs
coot: The `dnsSubscriptionWorker`is also used where we don't have access to the `OutboundQ` ([here](https://github.com/input-output-hk/cardano-sl/blob/develop/lib/src/Pos/Worker.hs#L55)).  I will add the `updatePeersBucket` function to the type signature so we can pass it as the first argument either from the `MonadKnownPeers` constraint or by using `updatePeersBucket` function directly when the `OutboundQ` is in scope).

lib/src/Pos/Launcher/Resource.hs
coot: ok, I'll look into this part on Monday morning. I'll push now fixes for the other two changes that you requested though.

lib/src/Pos/Diffusion/Full.hs
```diff
@@ -322,7 +323,8 @@ runDiffusionLayerFull networkConfig ourVerInfo oq slotDuration listeners action
         return ((>>= either throwM return) <$> itMap)
     subscriptionThread nc sactions = case topologySubscriptionWorker (ncTopology nc) of
         Just (SubscriptionWorkerBehindNAT dnsDomains) -> do
-            dnsSubscriptionWorker oq networkConfig dnsDomains slotDuration sactions
+            timer <- newDynamicTimer (convertUnit <$> slotDuration)
+            dnsSubscriptionWorker (OQ.updatePeersBucket oq) networkConfig dnsDomains timer sactions
```
avieth: Could you put a TODO note here about how we need to wire this timer up somewhere else so that it resets when we receive a block?

lib/src/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -73,20 +73,20 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} retryInterval sendActions = d
...
             let dnsPeers' = M.insert index dnsPeersList dnsPeers
-            void $ OQ.updatePeersBucket oq BucketBehindNatWorker $ \_ ->
+            void $ updatePeersBucket BucketBehindNatWorker $ \_ ->
                 peersFromList mempty ((,) NodeRelay <$> M.elems dnsPeers')
             pure (dnsPeers', ())
         -- Try to subscribe to some peer.
         -- If they all fail, wait a while before trying again.
         subscribeToOne dnsPeersList
-        retryInterval >>= delay
+        dtDuration keepAliveTimer >>= delay
```
avieth: You don't have to do it for this pull request, but I think eventually we'll want this to be different from the keep alive timer. No sense waiting a whole slot duration without any connection before trying again, I think.

lib/src/Pos/Launcher/Resource.hs
avieth: Thanks.
The place to look is `Pos.Block.Network.Retrieval` and `Pos.Block.Worker`, which pass along a timer and reset it whenever a block comes through. Really though, this should all be done via the diffusion layer. The timer can be created in `fullDiffusionLayer` and passed to the subscription feature, and also to the block listeners, to  get the same result.

util/Pos/Util/Timer.hs
```diff
@@ -22,38 +22,45 @@ module Pos.Util.Timer
...
   unless done retry
 
 -- Start the timer. If the function is called before @t + n@, where @t@ is the
 -- time startTimer was called for the last time and @n@ is the number of
 -- microseconds associated with the Timer, the timer is restarted.
 startTimer :: MonadIO m => Timer -> m ()
-startTimer Timer{..} = liftIO (registerDelay =<< readMVar timerDuration)
-    >>= atomically . writeTVar timerSempaphore
+startTimer Timer{..} = liftIO (registerDelay . fromIntegral =<< readMVar timerDuration)
+    >>= atomically . writeTVar timerSemaphore
```
avieth: I think the comment is wrong. `waitTimer` will take the current delay `TVar` from the outer `TVar`, but `startTimer` will replace the contents of the outer `TVar`, so the earlier `waitTimer` will still only wait for the original duration. What do you think? Am I missing something?

Also, would be good to really consolidate this with `DynamicTimer`. I think we only need `DynamicTimer`.

coot: That's true. `threadDelay` (which is used) does not expose any interface to manipulate the timeout.  

I agree that we actually only need the `DynamicTimer`.  This will make the other timers where slot duration is used more elastic on slot duration changes.

lib/src/Pos/Launcher/Resource.hs
coot: At the moment merging #2156 brings quite a few conflicts.  I'll do that in a separate issue/PR.  We can try to solve the TODO for the dns subscription timer in a similar way (thought in this case we'll need to keep a list of timers per dns subscription, since they are not shared across connections, like the `ncSubscriptionKeepAliveTimer` is).

lib/src/Pos/Context/Context.hs
```diff
@@ -107,8 +107,8 @@ data NodeContext = NodeContext
     -- ^ Settings for global Txp.
     , ncConnectedPeers      :: !ConnectedPeers
     -- ^ Set of peers that we're connected to.
-    , ncNetworkConfig       :: !(NetworkConfig KademliaParams)
-    , ncMisbehaviorMetrics  :: Maybe MisbehaviorMetrics
+    , ncNetworkConfig              :: !(NetworkConfig KademliaParams)
```
avieth: Mistake?

lib/src/Pos/Diffusion/Full.hs
```diff
@@ -12,7 +12,9 @@ import           Universum
 
 import           Control.Monad.Fix (MonadFix)
 import qualified Data.Map as M
-import           Data.Time.Units (Millisecond, Second)
+import           Data.Time.Units (Millisecond, convertUnit)
+-- import           Formatting (sformat, shown, (%))
```
avieth: Delete

lib/src/Pos/Diffusion/Full.hs
```diff
@@ -351,10 +354,13 @@ runDiffusionLayerFull networkConfig transport ourVerInfo mEkgNodeMetrics oq keep
         let itMap = M.fromList itList
         return ((>>= either throwM return) <$> itMap)
     subscriptionThread nc sactions = case topologySubscriptionWorker (ncTopology nc) of
-        Just (SubscriptionWorkerBehindNAT dnsDomains) ->
-            dnsSubscriptionWorker oq networkConfig dnsDomains keepaliveTimer slotDuration sactions
+        Just (SubscriptionWorkerBehindNAT dnsDomains) -> do
+            -- TODO wire this timer up so that we can reset it when a new block
+            -- arrives.
+            timer <- newDynamicTimer (convertUnit <$> slotDuration)
```
avieth: Do we not need the timer to be in scope within the diffusion layer? If so, get rid of the `_keepaliveTimer` argument.

avieth: Ah, but we *do* need it to be in scope, so that we can reset it when a new block arrives (as the comment says).

coot:
@avieth this is ready for review.  This PR introduces the `DynamicTimer`.  Its advantage is that it nows the current slot time, unlike the previous timer which didn't had access to the current slot time at some call sites.  I also removed the previous timer, since it is not used anywhere at the moment.

coot:
This is not how the timer is supposed to be used.  You need to (re)start the timer:
```
timer <- newTimer 5e6
let showPOSIXTime = getPOSIXTime >>= putStrLn . show
showPOSIXTime >> startTimer timer 
  >> (async $ atomically (waitTimer timer) >> showPOSIXTime) 
  >> (setTimerDuration timer 10e6 >> startTimer timer)
```
This prints:
```
1522340339.547878236s
1522340349.552934973s
```

The dynamic timer works in the same way.

coot:
I will add a quickcheck test.

util/Pos/Util/DynamicTimer.hs
```diff
@@ -0,0 +1,36 @@
...
+module Pos.Util.DynamicTimer
+  ( DynamicTimer
+  , newDynamicTimer
+  , waitDynamicTimer
+  , startDynamicTimer
+  ) where
+
+import           Data.Time.Units (Microsecond, toMicroseconds)
+import           Control.Concurrent.STM (readTVar, registerDelay, retry)
+import           Universum
```
avieth: Let's use the `STM` definitions, not the `Universum` ones (`newTVarIO`). We ought to divest from `Universum`.

avieth:
@coot I guess I just don't see how the dynamic timer works. The inner `TVar` is always written with `True`, so that `retry` never happens. When I `waitTimer`, I will always finish whenever the most recent `startDynamicTimer` duration finishes, no? So if another thread bumps it (for instance, because a block came in), I'll still wake up after the original duration.

coot:
It works this way, let say  you have a dynamic timer which constant timeout: 100ms, and let say that after 50ms, some other thread calls `startDynamicTimer` (so we have two calls to `startDynamicTimer` the initial one, and one after `50ms`).

So after the first `startDynamicTimer` call one has  `TVar (TVar False)`.  The inner `TVar False` will mutate to `TVar True` after 100ms (by `registerDelay`). But after `50ms` the second call to `startDynamicInterval` mutates the outer `TVar`: it throws away this inner `TVar False` (which in `50ms` will mutate its value), and puts inside a fresh `TVar False` (which  holds the promise that in another `100ms` it will mutate its value to `TVar True`).  Now the `waitDynamicTimer` will watch for the value of the inner `TVar` but when the outer one is mutated it will start watching the new one.

avieth:
@coot Ah, the magic is all done in `registerDelay`! I get it now.

I'd prefer an interface in which the `m Microsecond` timeout is not built-in to the timer itself, but instead the `Microsecond` is given as a parameter to `startTimer`. More modular this way. Is this just the original `Timer`?

coot:
All tests passed, I had to rerun buildkite on linux because of a Cbor test failure:
```
  src/Test/Pos/Helpers.hs:92:
  1) Test.Pos.Cbor.Cbor, Cbor.Bi instances, Types, Bi instances, Core.Types, Attributes ()
       Falsifiable (after 6 tests and 34 shrinks):
       Attributes { data: (), remain: <16 bytes> }
Attributes { data: (), remain: <16 bytes> }
```

coot:
Sure, the history needs to be cleaned a bit.

arybczak:
> @arybczak Could you look into this one?

Yeah, but I don't have access to buildkite and I think unless I get a RNG seed to get the exact value that produces failure, I probably won't be able to reproduce.

avieth:
> Yeah, but I don't have access to buildkite and I think unless I get a RNG seed to get the exact value that produces failure, I probably won't be able to reproduce.

And unfortunately that test includes not only `decode . encode = id` but also the canonicity test.

FWIW the failing item is 

```Haskell
Attributes { attrData = (), attrRemain = <some nonempty map> }
```

The failing piece is almost certainly the `attrRemain :: UnparsedFields ~ Map Word8 LBS.ByteString`

coot:
@avieth I rebased this PR.  I saw today a discussion weather we should use `src` directory or not, which is not yet decided, so I dropped this change. This PR contains three things:

* simplification of the timer
* we used `slotDuration` to restart the timer uniformelly (previously, `3 * slodDuration` was used when a block arrives and `slotDuration` repeatedly when a subscription comes.  There is no point to wait `3` longer in block duration since, it will be reset earlier by the other thread.
* tests for the timer (I hope they will not by flaky, if they'll turned out to be we'd need to adjust times in the tests, but they shouldn't be since usual resolution of system clock can be measured in tens of microseconds).

coot:
I removed the commit that modified the timer timeout when a block is received.  It makes sense to have a longer pause of the keep alive messages when new blocks are arriving.  (This wasn't possible with the dynamic timer).

util/Pos/Util/Timer.hs
```diff
@@ -1,60 +1,31 @@
...
--- activate it 'startTimer' needs to be called.
-newTimer :: MonadIO m => Microsecond -> m Timer
-newTimer n = Timer <$> newMVar n <*> (newTVarIO True >>= newTVarIO)
-
--- | Set the duration of a timer to a specified number of microseconds. Note
--- that if the timer is already started, calling this function wont't restart
--- it.
-setTimerDuration :: MonadIO m => Timer -> Microsecond -> m ()
-setTimerDuration Timer{..} n = liftIO $ modifyMVar_ timerDuration $ \_ -> return $! n
+-- | Construct new `Timer` from an action that returns time interval
```
avieth: Incorrect comment

util/test/Test/Pos/Util/TimerSpec.hs
```diff
@@ -0,0 +1,64 @@
...
+        _ <- startTimer t timer
+        _ <- atomically $ waitTimer timer
+        endTime <- run getPOSIXTime
+        let diffTime = nominalDiffTimeToMicroseconds (endTime - startTime)
+        assert $ t <= diffTime 
+
+-- | A timer should be additive: (re)-starting a timer @t1@ after @t0@
+-- microseconds should wait @t0 + t1@ microseconds.
+prop_additive :: Property
+prop_additive =
```
avieth: This is a nice test but I think we should also include a similar one, in which a single timer is `startTimer`d twice: first with `t0` and then again with `t1` after `d`, _before_ `t0` has elapsed (say, `d = t0 / 2`, using a `threadDelay`). The test then checks that the observed elapsed time by the `waitTimer` thread is at least `t1 + d`

coot: I'll preserve the current implementation, because the test passes more consistently (with `threadDelay` I needed to run the timer for a longer period, which makes the test slower).

avieth:
And now a different CI failure.

```
Command executed with exception: /AppData/Local/Programs/stack/x86_64-windows/ghc-8.0.2\lib\package.conf.d\package.cache
ghc will see an old view of this package db. Use 'ghc-pkg recache' to fix.
```

PR #2355, opened 2018-01-20
[CSL-2170] Remove MonadFail

int-index:


lib/test/Test/Pos/UtilSpec.hs
```diff
@@ -1,8 +1,8 @@
 {-# LANGUAGE DataKinds    #-}
 {-# LANGUAGE RankNTypes   #-}
-{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE Pos #-}
```
neongreen: Wat

client/test/Test/Pos/Client/Txp/UtilSpec.hs
```diff
@@ -316,8 +316,7 @@ getSignerFromList (HM.fromList . map swap . toList -> hm) =
 
 makeManyUtxoTo1Params :: InputSelectionPolicy -> Int -> Integer -> Integer -> Gen CreateMTxParams
 makeManyUtxoTo1Params inputSelectionPolicy numFrom amountEachFrom amountTo = do
-    [skFrom, skTo] <- nonrepeating 2
-
+    ~[skFrom, skTo] <- nonrepeating 2
```
avieth: Is this relevant to the rest of the patch?

int-index: Yes, this changes how pattern matching failure is desugared. With a lazy pattern it's not using `MonadFail`, in case of a mismatch the execution will continue, but each bound variable will be bottom (`skFrom` and `skTo` in this case).

PR #2360, opened 2018-01-23
Explicit DB reads and writes in the wallet

nc6:
This work subsumes #2321 and #2357.

Atop those PRs, we:
- Add explicit threading of the DB handle into functions in `State.hs`.
- Rewrites various areas of functionality to use single transactions, where we judge this to be appropriate.

It does not make any changes to the return types of `Update` functions to the Acid state. Doing so would likely result in a number of changes across the codebase as we are forced to discard return types, and the number of places where we need such a return type seems minimal. If we do this, I think we should address in a separate PR.

Commits are currently split for easy review; after review we can rebase and add YT references etc,

### DIFF
--- a/node/src/Pos/Wallet/Web/State/State.hs
+++ b/node/src/Pos/Wallet/Web/State/State.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Rank2Types   #-}
 {-# LANGUAGE TypeFamilies #-}
 
 module Pos.Wallet.Web.State.State
@@ -13,10 +14,12 @@ module Pos.Wallet.Web.State.State
        , closeState
 
        , AddressLookupMode (..)
-       , CustomAddressType (..)
        , CurrentAndRemoved (..)
+       , CustomAddressType (..)
 
        -- * Getters
+       , WalletSnapshot
+       , getWalletSnapshot
        , getProfile
        , doesAccountExist
        , getAccountIds
@@ -82,7 +85,6 @@ module Pos.Wallet.Web.State.State
        , addOnlyNewPendingTx
        , cancelApplyingPtxs
        , cancelSpecificApplyingPtx
-       , getWalletStorage
        , flushWalletStorage
        ) where
 
@@ -106,9 +108,9 @@ import           Pos.Wallet.Web.Pending.Types (PendingTx (..), PtxCondition)
 import           Pos.Wallet.Web.State.Acidic  (WalletState, closeState, openMemState,
                                                openState)
 import           Pos.Wallet.Web.State.Acidic  as A
+import qualified Pos.Wallet.Web.State.Storage as S
 import           Pos.Wallet.Web.State.Storage (AddressInfo (..), AddressLookupMode (..),
-                                               CurrentAndRemoved (..),
-                                               CustomAddressType (..),
+                                               CAddresses, CurrentAndRemoved(..) , CustomAddressType (..),
                                                PtxMetaUpdate (..), WalletBalances,
                                                WalletStorage, WalletTip (..))
 
@@ -129,11 +131,17 @@ type WebWalletModeDB ctx m =
     , HasConfiguration
     )
 
+type WalletSnapshot = WalletStorage
+
 queryDisk
     :: (EventState event ~ WalletStorage, QueryEvent event,
         MonadWalletWebDB ctx m, MonadIO m)
     => event -> m (EventResult event)
 queryDisk e = getWalletWebState >>= flip A.query e
+
+queryValue
+    :: WalletStorage -> S.Query a -> a
+queryValue ws q = runReader q ws
 
 updateDisk
     :: (EventState event ~ WalletStorage, UpdateEvent event,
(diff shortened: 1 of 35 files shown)


adinapoli-iohk:
Good start @nc6 ! Now that we have reverted the silly `updateDisk'`, we can crack on with the plan we envisioned together with @dcoutts , which I'm summarising just for my own benefit here 😁  : 

- We introduce a `takeSnapshot` (or similar) function to grab the snapshot from the DB ref: `takeSnapshot :: MonadIO m => WalletDB -> WalletSnapshot`

- (Bonus) We rename `WalletStorage` into `WalletDbSnapshot` (or similar) to better enforce the separation between the ref and the value manipulated (but might be overkill at this stage)

- We pass explicitly a `WalletDb` to the update functions in `State.hs` so that now they can drop the `WalletDbReader` constraint and, more importantly, reveal all the updates in our system.

- For each of these updates, we go back to the drawing board asking whether they should be fused into a bigger, atomic `Update` transaction.

- For each `Update` function (in `Storage`), we decide whether or not they should return the updated `WalletDbSnapshot` or whether they should instead return just what changed. I'm on the fence but I tend to prefer returning the updated snapshot, it makes the API a bit consistent, but you guys might have strong opinions.

Sounds about right? 😉

node/src/Pos/Wallet/Web/State/State.hs
```diff
@@ -127,6 +128,10 @@ askWalletDB = view (lensOf @WalletDB)
 -- | The 'WalletDbWriter' constraint encapsulate the fact we can write an updated
 -- copy of the 'WalletStorage', but that we cannot read it back, if not as part
 -- of an atomic update API.
+--
+-- Note that this is not a counterpart to 'WalletDbReader'. 'WalletDbReader'
```
adinapoli-iohk: @nc6 Do you think this is confusing enough for us to want `WalletDbReader` to be renamed into something else? 😉

nc6: Possibly, but I couldn't think of an immediately better name. It will be fairly easy to change if desired in the future; it's only referenced in four files or so.

node/src/Pos/Wallet/Web/State/Transactions.hs
```diff
@@ -0,0 +1,23 @@
...
+
+-- | Create an account with an address.
+createAccountWithAddress
+  :: AccountId
+  -> CAccountMeta
+  -> CWAddressMeta
+  -> Update ()
+createAccountWithAddress accId accMeta addrMeta = do
+  WS.createAccount accId accMeta
+  WS.addWAddress addrMeta
```
adinapoli-iohk: Yay! 🎉

node/src/Pos/Wallet/Web/State/State.hs
```diff
@@ -379,13 +378,20 @@ addOnlyNewTxMeta :: (WalletDbWriter A.AddOnlyNewTxMeta m)
...
+--   - Associated accounts
+--   - Transaction metadata
+--   - History cache
+--
+--   Note that this functionality has changed - the old version of
+--   'removeWallet' did not used to remove the associated data.
+--   This functionality was not used anywhere and was therefore
+--   removed. Should it be needed again, one should add 'removeWallet'
+--   to the set of acidic updates and add a suitable function in this
+--   module to invoke it.
```
adinapoli-iohk: Unless we are both missing something obvious I totally agree with the new behaviour, I cannot envisage a use case where we might want to _not_ delete the associated data for a wallet. @Martoon-00 , are you aware of anything subtle we are missing here?

wallet/src/Pos/Wallet/Web/Methods/Backup.hs
```diff
@@ -46,7 +46,11 @@ restoreWalletFromBackup WalletBackup {..} = do
                           & each . _2 %~ \(AccountMetaBackup am) -> am
 
             addSecretKey wbSecretKey
-            -- XXX Transaction
+            -- My current thinking is that we don't want a transaction here. If we're restoring
```
adinapoli-iohk: A minor style note: I think the style guide is such that we should be talking not directly in first person, but I do appreciate this has been left here just for ease of review, really 😀 

Ps. Minor typo below, that's "backup", not "bacup" 😀

node/src/Pos/Wallet/Web/State/Transactions.hs
```diff
@@ -41,3 +48,27 @@ deleteWallet walId = do
...
+  :: CId Wal
+  -> [CWAddressMeta] -- ^ Wallet addresses to add
+  -> [(WS.CustomAddressType, [(CId Addr, HeaderHash)])] -- ^ Custom addresses to add
+  -> UtxoModifier
+  -> [(CTxId, CTxMeta)] -- ^ Transaction metadata to add
+  -> Map TxId TxHistoryEntry -- ^ Entries for the history cache
+  -> [(TxId, PtxCondition)] -- ^ PTX Conditions
+  -> HeaderHash -- ^ New sync tip
+  -> Update ()
+applyModifierToWallet walId wAddrs custAddrs utxoMod
```
adinapoli-iohk: I guess this one will require a bit more thought (but just because I guess it's a fiddly issue). Let me report an excerpt of a conversation I had with @pva701 a while ago:

> `CAccModifier` is outdated name of such thing which should be actually called *wallet modifier*. previously wallet was called account (so Acc in `CAccModifier` stands for "account"), so now we have this discrepancy. 
as you know, the wallet listens to blockchain, in sense that it reacts on each block (`BListener` class is provide interface for it).
each block may affect some wallet stored in wallet-db, namely, incoming and outgoing transactions may affect wallet-db's utxo, add and remove new addresses and so on and so forth.
so we consider each block from point of view of each stored wallet and try to find which things are affected for each of wallets.
so `CAccModifier` contains all these modifications for *one particular wallet*. for each block we compute `CAccModifier` corresponding to some wallet and then apply it to wallet-db.
yet another thing, that `CAccModifier` is monoid and hence `CAccModifier`s for several blocks (but corresponding to the same wallet) can be mappended and as result we get such `CAccModifier`, application of which gives the same wallet-db as for sequential applications of `CAccModifier`s for these blocks to wallet-db.
(you may see mappending in some places of code and it was done to reduce number of writing to wallet-db, especially when we sync with blockchain and iterate over whole blockchain)

@nc6 Is this excerpt of any help to you? 😀

nc6:
Closing this issue in favour of a branch from the IOHK repo, since otherwise CI will not work.

PR #2362, opened 2018-01-23
[CSL-2073] Cosmetic changes

Pastafarianist:
More of the same. Mostly comments, also very small refactoring and export cleanup.

One concern is the comment I removed from `explorer/src/Pos/Explorer/Web/ClientTypes.hs`. It refers to adding vending addresses to Explorer API (so that genesis addresses could be searched by corresponding vending addresses, not just by Cardano addresses). As far as I understand, we have abandoned this idea (we have not brought it to the table for about four months), but feel free to reject the PR if this is not the case.


block/src/Pos/Lrc/Worker.hs
```diff
@@ -274,7 +274,7 @@ richmenComputationDo epochIdx consumers = unless (null consumers) $ do
 -- It was deleted because of possible deadlock. This worker may start
 -- doing LRC and try to acquire 'StateLock' while another thread may
 -- hold 'StateLock' for block processing and try to start LRC.  So if
--- you are going to uncomment it at some point, please take it into
+-- you are going bring it back at some point, please take it into
```
neongreen: going *to* bring it back

PR #2364, opened 2018-01-23
CSL-1570 resolve pending bi tests refactor

v0d1ch:


block/src/Pos/Arbitrary/Block.hs
```diff
@@ -18,10 +18,13 @@ import           Pos.Arbitrary.Delegation (genDlgPayload)
 import           Pos.Arbitrary.Ssc (SscPayloadDependsOnSlot (..))
 import           Pos.Arbitrary.Txp ()
 import           Pos.Arbitrary.Update ()
+import           Pos.Arbitrary.Update ()
```
gromakovsky: Duplicated line.

crypto/Pos/Arbitrary/Crypto.hs
```diff
@@ -224,3 +224,11 @@ instance Arbitrary HDPassphrase where
 
 instance Arbitrary HDAddressPayload where
     arbitrary = genericArbitrary
+
+----------------------------------------------------------------------------
+-- WithHash
+----------------------------------------------------------------------------
+
+instance (Bi a, Arbitrary a) => Arbitrary (WithHash a) where
+    arbitrary = genericArbitrary
```
gromakovsky: It's better to use `withHash` function to generate something valid.

crypto/Pos/Arbitrary/Crypto.hs
```diff
@@ -224,3 +230,86 @@ instance Arbitrary HDPassphrase where
...
+instance Arbitrary CC.ChainCode where
+    arbitrary = do
+        a <- arbitrary
+        return $ CC.ChainCode a
+
+instance Arbitrary CC.XPrv where
+    arbitrary = do
+        a <- arbitrary
+        return $ CC.xprv b
+
```
v0d1ch: @gromakovsky how do I define arbitrary instance for XPrv which is 
```
newtype XPrv = XPrv EncryptedKey
    deriving (NFData, ByteArrayAccess)
```
The constructor is not exported and it is defined as external git dependency so I should not touch the module.
I can work with `xprv` exported function which returns `Either String XPrv` but I don't know how would the `Left` case look then ?

crypto/Pos/Arbitrary/Crypto.hs
```diff
@@ -224,3 +230,86 @@ instance Arbitrary HDPassphrase where
...
+--------------------------------------------------------------------------
+-- Pos.Crypto.Signing.Types.Safe
+--------------------------------------------------------------------------
+
+instance Arbitrary EncryptedSecretKey where
+    arbitrary = do
+        a <- arbitrary
+        b <- arbitrary
+        return $ EncryptedSecretKey a b
```
gromakovsky: AFAIU, it's the same as `genericArbitrary`.

gromakovsky:
@v0d1ch I've looked at your problem and actually I think we shouldn't define `Bi` and/or `Arbitrary` instances for the types from external packages which are encapsulated by types from our packages. I am talking about: `Pvss.*`, `Ed25519.*`, `CC.*`, `EdStandard.*`.
I see that we have `Bi` instances for those types, but I think we should delete them.
AFAIU, they are only used to define `Bi` instances for our types which wrap those types.
So for example we have
```
instance Bi CC.XPrv where
    encode (CC.unXPrv -> kc) = encode kc
    decode = toCborError . over _Left fromString . CC.xprv =<< decode @ByteString

deriving instance Bi SecretKey
```

I would instead define `encodeXPrv` and `decodeXPrv` and then 
```
instance Bi SecretKey where
    encode (SecretKey xprv) = encodeXPrv xprv
    decode = fmap SecretKey . decodeXPrv
```

Does it make sense? Sorry that I didn't think about it earlier.

v0d1ch:
No problem, sounds good to me ¯\_(ツ)_/¯

lib/src/Pos/Util/UserSecret.hs
```diff
@@ -122,6 +129,15 @@ data UserSecret = UserSecret
 
 makeLenses ''UserSecret
 
+instance Eq UserSecret where
```
int-index: Why is this instance not derived?

lib/src/Pos/Util/UserSecret.hs
```diff
@@ -81,6 +81,13 @@ data WalletUserSecret = WalletUserSecret
 
 makeLenses ''WalletUserSecret
 
+instance Eq WalletUserSecret where
```
int-index: Why is this instance not derived?

crypto/Pos/Crypto/Signing/Types/Safe.hs
```diff
@@ -39,6 +39,13 @@ instance Show EncryptedSecretKey where
 instance B.Buildable EncryptedSecretKey where
     build _ = "<encrypted key>"
 
+instance Eq CC.XPrv where
+    a == b = CC.unXPrv a == CC.unXPrv b
+
+instance Eq EncryptedSecretKey where
```
int-index: Why is this instance not derived?

crypto/Pos/Crypto/Signing/Types/Safe.hs
```diff
@@ -39,6 +39,13 @@ instance Show EncryptedSecretKey where
 instance B.Buildable EncryptedSecretKey where
     build _ = "<encrypted key>"
 
+instance Eq CC.XPrv where
```
int-index: Why did you decide to make an orphan instance instead of defining it in `Cardano.Crypto.Wallet`?

crypto/Pos/Arbitrary/Crypto.hs
```diff
@@ -224,3 +225,20 @@ instance Arbitrary HDPassphrase where
...
     arbitrary = genericArbitrary
+
+----------------------------------------------------------------------------
+-- WithHash
+----------------------------------------------------------------------------
+
+instance (Bi a, Arbitrary a) => Arbitrary (WithHash a) where
+    arbitrary = do
+        a <- arbitrary
+        return $ withHash a
```
int-index: That's just a `fmap`

block/src/Pos/Arbitrary/Block.hs
```diff
@@ -402,3 +404,21 @@ instance (Arbitrary SscPayload, HasConfiguration) =>
...
+------------------------------------------------------------------------
+
+instance Arbitrary SlogUndo where
+    arbitrary = genericArbitrary
+    shrink = genericShrink
+
+instance HasConfiguration => Arbitrary Undo where
+    arbitrary = genericArbitrary
+    shrink = genericShrink
+    -- arbitrary = do
```
int-index: Why is there code that is commented out?

v0d1ch: my mistake

crypto/Pos/Crypto/Signing/Types/Safe.hs
v0d1ch: AFAIK we are not supposed to touch code that comes from external git dependency

crypto/Pos/Crypto/Signing/Types/Safe.hs
v0d1ch: It is now

v0d1ch:
@gromakovsky  || @int-index  -> Ok turns out that we need `Bi` instance for `XPrv` in `cardano-sl/wallet/src/Pos/Wallet/Aeson/WalletBackup.hs:85` for `From/ToJSON` 
```
cardano-sl/wallet/src/Pos/Wallet/Aeson/WalletBackup.hs:85:19: error:
        • No instance for (Bi.Bi
                             cardano-crypto-1.0.0:Cardano.Crypto.Wallet.XPrv)
            arising from a use of ‘decodeBase64’
        • In the first argument of ‘(=<<)’, namely ‘decodeBase64’
          In a stmt of a 'do' block:
            prvKey <- decodeBase64 =<< o .: "walletSecretKey"
          In the expression:
            do { let decodeBase64 x
                       = toAesonError (B64.decode x) >>= toAesonError . Bi.decodeFull
                     collectAccMap = foldlM parseAddAcc HM.empty
                     ....;
                 prvKey <- decodeBase64 =<< o .: "walletSecretKey";
                 passPhraseHash <- decodeBase64 =<< o .: "passwordHash";
                 walletMeta <- o .: "walletMeta";
                 .... }

    /Users/v0d1ch/code/cardano-sl/wallet/src/Pos/Wallet/Aeson/WalletBackup.hs:125:44: error:
        • No instance for (Bi.Bi
                             cardano-crypto-1.0.0:Cardano.Crypto.Wallet.XPrv)
            arising from a use of ‘Bi.serialize'’
        • In the first argument of ‘B64.encode’, namely
            ‘(Bi.serialize' prvKey)’
          In the second argument of ‘(.=)’, namely
            ‘B64.encode (Bi.serialize' prvKey)’
          In the expression:
            "walletSecretKey" .= B64.encode (Bi.serialize' prvKey)
```
Should I try to define encode/decode functions like I did before or is there another solution ?

gromakovsky:
In the first case you can decode it as `SecretKey` and then extract `XPrv` from it. In the second case you can `serialize'` it as a `SecretKey`. Just make sure that `SecretKey` is serialized exactly the same way as `XPrv`.

crypto/Pos/Crypto/Signing/Types/Safe.hs
int-index: I don't see why.

int-index:
@v0d1ch Note that I changed the base branch of this PR, because according to our new workflow individual tasks are merged into user story branches. You might need to rebase.

v0d1ch:
Ok I think this should be ready for final review

crypto/Pos/Crypto/Signing/Types/Safe.hs
int-index: At least change it to `deriving instance Eq CC.XPrv` and write a TODO about adding this instance to `Cardano.Crypto.Wallet` (I don't like unnecessary orphan instances)

wallet/src/Pos/Wallet/Aeson/WalletBackup.hs
```diff
@@ -122,7 +122,7 @@ instance ToJSON IndexedAccountMeta where
 
 instance ToJSON WalletBackup where
     toJSON (WalletBackup skey wMeta wAccounts) = object
-        [ "walletSecretKey" .= B64.encode (Bi.serialize' prvKey)
+        [ "walletSecretKey" .= B64.encode (Bi.serialize' $ SecretKey prvKey)
```
int-index: Why did you have to add this wrapper here? Did it work incorrectly before? I thought this task was about adding tests, but here you're changing the logic, is this a bugfix?

gromakovsky: This task is about adding some instances for types with `Bi` instances. Previously `XPrv` had `Bi` instance, but I find it wrong (because it's implementation detail) and asked to delete that instance. I think `XPrv` shouldn't be used directly, the purpose of `SecretKey` newtype is to hide it.

gromakovsky:
One of commits in this PR shouldn't be in this PR.

crypto/Pos/Binary/Crypto.hs
```diff
@@ -126,45 +128,43 @@ BiMacro(C.EncShare, encShareBytes)
 -- Signing
 ----------------------------------------------------------------------------
 
-instance Bi Ed25519.PointCompressed where
-    encode (Ed25519.unPointCompressed -> k) = encode k
-    decode = Ed25519.pointCompressed <$> decode
+encodeXSignature :: CC.XSignature -> E.Encoding
+encodeXSignature a = E.encodeBytes $ CC.unXSignature a
```
gromakovsky: Previously it was:
```
instance Bi CC.XSignature where
    encode (CC.unXSignature -> bs) = encode bs
```
, but now you are using `E.encodeBytes` instead of `encode`. Why? Even if it's equivalent, I think it's better to leave `encode`.
It also applies to some other places.

crypto/Pos/Crypto/Signing/Types/Safe.hs
v0d1ch: I did that with todo about what needs to be added to `Cardano.Crypto.Wallet`

crypto/Pos/Binary/Crypto.hs
v0d1ch: `encode` it is than

gromakovsky:
Please remove ac4073ebdedbcca7ccbec379245328aaedb9e249 from this PR.

v0d1ch:
Ha ha I just deleted them. I thought they were there just as a reminder that we need those instances defined - I'll bring it back

patch
```diff
@@ -0,0 +1,251 @@
+diff --git a/pkgs/default.nix b/pkgs/default.nix
```
gromakovsky: .

PR #2365, opened 2018-01-24
[CSL-2170] Remove MonadFail, part 2

int-index:


binary/Pos/Binary/Class/TH.hs
```diff
@@ -175,10 +184,6 @@ deriveSimpleBiInternal predsMB headTy constrs = do
     tagType :: TypeQ
     tagType = [t| Word8 |]
 
-    -- Helpers --
-    failText :: MonadFail m => T.Text -> m a
-    failText = fail . toString
```
gromakovsky: Now there are 100500 usages of `toTemplateHaskellError . Left`. Why not leave `(failText :: Text -> Q a)  = toTemplateHaskellError . Left`? Maybe `failText` is not the best name, but other than that such function makes perfect sense to me 🤷‍♂️

crypto/Pos/Crypto/HD.hs
```diff
@@ -125,7 +125,7 @@ packHDAddressAttr (HDPassphrase passphrase) path = do
         CryptoPassed p  -> HDAddressPayload p
 
 -- | Try to decrypt HDAddressPayload using HDPassphrase.
-unpackHDAddressAttr :: MonadFail m => HDPassphrase -> HDAddressPayload -> m [Word32]
+unpackHDAddressAttr :: HDPassphrase -> HDAddressPayload -> Maybe [Word32]
```
gromakovsky: Why not `Either`? We are losing some information here. I think it's better for the caller to decide whether that information is useful.

int-index: When I was instantiating these types, I was studying all call sites. For this function, its type is always instantiated to `Maybe`, and that is how I rewrote it.

binary/Pos/Binary/Class/TH.hs
int-index: Primarily for consistency, but I see your point, I'll restore this function under a different name. I don't like the name `failText` because it makes harder to grep for `fail`, thus to adhere to the policy of no `MonadFail`.

int-index: I replaced it with `templateHaskellError`.

util/Pos/Util/Util.hs
```diff
@@ -110,22 +121,56 @@ logException name = E.handleAsync (\e -> handler e >> E.throw e)
...
+-- return monad is intentionally specialized because we avoid 'MonadFail'.
+toTemplateHaskellError :: Either Text ~> TH.Q
+toTemplateHaskellError = external_api_fail
+
+templateHaskellError :: Text -> TH.Q a
+templateHaskellError = toTemplateHaskellError . Left
+
+-- | Convert an 'Either'-encoded failure to a 'cereal' failure. The
+-- return monad is intentionally specialized because we avoid 'MonadFail'.
+toCerealError :: Either Text ~> Cereal.Get
```
gromakovsky: 🤔

gromakovsky: Prikolno

networking/bench/Sender/SenderOptions.hs
```diff
@@ -94,4 +94,4 @@ argsParser =
     recipient = connection >>= \(h, mp) ->
         case mp of
             Just p -> return (fromString h, p)
-            _      -> fail $ "No port specified for host " <> h
+            _      -> _ $ "No port specified for host " <> h
```
gromakovsky: :cho:

int-index: I missed this one because I didn't build benchmarks locally.

PR #2366, opened 2018-01-24
[CSL-2207] progressive network connection timeouts

coot:


networking/src/Node/Internal.hs
```diff
@@ -111,6 +115,8 @@ data NodeState peerData m = NodeState {
     , _nodeStateClosed             :: !Bool
       -- ^ Indicates whether the Node has been closed and is no longer capable
       --   of establishing or accepting connections (its EndPoint is closed).
+    , _nodeStateTimeouts           :: !(Map NodeId Microsecond)
```
avieth: How about `Map NodeId [Microsecond]`. The head of this list (if nonempty) is the next delay. If the list is empty, it means all delays have run their course, so the connection attempt should fail. This will make `connectToPeer` logic simpler, as it won't need the partial `last :: [a] -> a` function.

lib/src/Pos/Configuration.hs
```diff
@@ -45,8 +48,10 @@ data NodeConfiguration = NodeConfiguration
     {
       ccDefaultPeers                 :: ![Text]
       -- ^ List of default peers
-    , ccNetworkConnectionTimeout     :: !Int
-      -- ^ Network connection timeout in milliseconds
+    , ccNetworkConnectionTimeouts    :: !(NonEmpty Int)
+      -- ^ Network connection timeouts in milliseconds.  It must be an
```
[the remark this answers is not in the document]
avieth: Also let's add a comment saying that the first of them is the network-transport-tcp timeout.

avieth:
If I understand correctly, if the first timeout fails, it won't automatically try to reconnect (and use the next timeout) but instead will use the next timeout at the next connection attempt.

If we connect to a peer once, then never again, the next progressive timeout remains in the node state. This could be considered a memory leak.

networking/src/Node/Internal.hs
coot: The `last` is not partial here because it consumes a non empty list.  What if all attempt failed, but after a while the failing node accepts connections back again.  In this case you will not reach to it (unless you restart the node).

coot:
That's true.  But I don't agree that's a memory leak, since this really depends on the state of other nodes in  the network.  The question is if we can afford to lock `Integer` (or a list of `Integers` as you propose) per `NodeId`.  I don't know how the topology of the real network looks like, so its hard for me to judge.

avieth:
> That's true. But I don't agree that's a memory leak, since this really depends on the state of other nodes in the network. The question is if we can afford to lock Integer (or a list of Integers as you propose) per NodeId. I don't know how the topology of the real network looks like, so its hard for me to judge.

The memory leakage is proportional to the number of peers that a node ever talks to. Once this thing is distributed, that could be a very high number.

coot:
Fair enough. What about keeping `Maybe Millisecond` in the state (as the last timeout) and empty it after some time (it has to be at least the sum of all timeouts)?  `Nothing` has a very minimal memory footprint after all.

avieth:
> Fair enough. What about keeping Maybe Millisecond in the state (as the last timeout) and empty it after some time (it has to be at least the sum of all timeouts)? Nothing has a very minimal memory footprint after all.

The added complexity of clearing up the map doesn't seem worth it to me.

coot:
Another idea: lets keep a quadruple of timeouts and just keep an index of type `data Index = First | Second | Third | Fourth`.  The memory footprint will only come from the `Map` itself.

PR #2367, opened 2018-01-24
Wallet bench gauge

denisshevchenko:
DO NOT MERGE - work in progress.

Related to https://iohk.myjetbrains.com/youtrack/issue/CBR-23

This is a servant-based client for wallet benchmarking. Currently it's just a skeleton.


wallet/bench/Main.hs
```diff
@@ -0,0 +1,31 @@
...
+-- is already running.
+main :: IO ()
+main = do
+    config <- getBenchConfig
+    hSetEncoding stdout utf8
+    putText "Wallet benchmarking begin..."
+    -- Only benchmark with config will be launched.
+    b1 <- async $ maybe (return ()) (runBench getHistoryIO) (extractConfigFor GetHistoryBench config)
+    b2 <- async $ maybe (return ()) (runBench getWalletsIO) (extractConfigFor GetWalletsBench config)
+    b3 <- async $ maybe (return ()) (runBench newPaymentIO) (extractConfigFor NewPaymentBench config)
```
adinapoli-iohk: Are we sure we want to run all the benchmarks concurrently? Perhaps it would be better sequentially, so that we get less variance out of them.

wallet/bench/config/Endpoints.csv
```diff
@@ -0,0 +1,4 @@
+BenchName,BenchDuration,MinDelayBetweenCalls,MaxDelayBetweenCalls,PathToReportFile
+GetHistoryBench,5.0,0.5,2.5,/tmp/GetHistoryBenchReport.csv
```
adinapoli-iohk: This assumes the target environment has a $TMPDIR, but what about Windows? 😉  My recommendation is to create a local directory (which we `gitignore`) that stores these results.

wallet/bench/Client/Pos/Wallet/Web/Api.hs
```diff
@@ -0,0 +1,52 @@
...
+import           Pos.Core.Types             (Coin)
+import           Pos.Wallet.Web.Api         (ApiPrefix, GetHistory, GetWallets,
+                                             NewPayment)
+import           Pos.Wallet.Web.ClientTypes (Addr, CAccountId (..),
+                                             CId (..), CPassPhrase, CTx, CWallet,
+                                             ScrollLimit, ScrollOffset, Wal)
+
+-- | "Benchmarking API" which includes
+-- endpoints we need for benchmarking.
+type WalletBenchApi = ApiPrefix :> (
```
adinapoli-iohk: Do we really need to use `ApiPrefix` here? what does it buy us? (Genuine question).

wallet/bench/Client/Pos/Wallet/Web/Api.hs
```diff
@@ -0,0 +1,52 @@
...
+getWallets
+    :: ClientM [CWallet]
+newPayment
+    :: Maybe CPassPhrase
+    -> CAccountId
+    -> CId Addr
+    -> Coin
+    -> Maybe InputSelectionPolicy
+    -> ClientM CTx
+getHistory :<|> getWallets :<|> newPayment = client (Proxy @WalletBenchApi)
```
adinapoli-iohk: Food for thoughts: I think currently this module is too coupled with the old API, which means that the day we'll decommission it we'll have to rewrite most of this module (which is fine as it's trivial) but then amend the rest of the benchmarks as the old types would have "leaked" everywhere.

I propose, as a  starter, to make the return type of these functions more opaque, i.e. let's change them to return simply a `ClientM Aeson.Value`. This way we can still inspect the blob we get from the server, if we need it (for example via `lens-aeson`) but now the code will be future-proofed. What do you think?

wallet/bench/Main.hs
denisshevchenko: > Are we sure we want to run all the benchmarks concurrently? 

AFAIU, yes. Because in the real life all requests always occurs concurrently... :-)

wallet/bench/config/Endpoints.csv
denisshevchenko: No, please don't look at the content of this file. :-) These lines were added just to keep this file nonempty. Of course these values will change.

wallet/bench/Client/Pos/Wallet/Web/Api.hs
denisshevchenko: Yes, we need it. Because this `ApiPrefix` is using in the complete API, so I want to use it here as well. Yes, we're using just 3 endpoints, but it should be the **same** as if _all_ API contained only these 3 endpoints.

denisshevchenko:
@adinapoli-iohk Thanks for your comments (and for the "food for thoughts").

wallet/bench/Client/Pos/Wallet/Web/Api.hs
denisshevchenko: Agree.

wallet/bench/Main.hs
adinapoli-iohk: Yes, I agree with that, but how do you get consistent measurements out of a benchmark if the starting environment is not stable? In order to get meaningful data out of an experiment (a benchmark in this case) you have to lock in place variance as much as you can, otherwise it's not a bench anymore but a stress test 😉  This is just my view and I'm definitely not the foremost expert, but I'm sure people like Neil can advise here 😀

denisshevchenko:
Moved to https://github.com/input-output-hk/cardano-sl/pull/2426

PR #2361, opened 2018-01-23
Migration to rocksdb-haskell-ng

chrisdone:
Summary:

* Comments out a few RocksDB-related things that aren't currently used.
* Switches all cabal-depends to rocksdb-haskell-ng.
* Adds rocksdb-haskell-ng as an extra-dep to the stack.yaml.

The rocksdb-haskell-ng project is a ground-up rewrite (with some code re-used where possible) of the rocksdb-haskell project, with things that are not used by cardano-sl not included.

The rewrite is motivated by making the binding as safe as possible: it should not ever segfault, and should be safe from memory leaks, double closes, async exceptions, and safe concurrent use, all for both connection handles and for iterators. The worst case scenario is that it should throw a meaningful exception when you have made a mistake.

To backup these aims, included is a fairly comprehensive test suite which attempts to test edge-cases and programmer error. Included are tests made by Niklas and some fixes for Unicode file access on Linux and Windows. It also includes a benchmark suite to track any performance improvements.

It also has Linux and Windows CI.

[Linux build Status](https://travis-ci.org/chrisdone/rocksdb-haskell-ng) [Windows build status](https://ci.appveyor.com/project/chrisdone/rocksdb-haskell-ng)

PR #2368, opened 2018-01-24
[CSL-1701] rocksdb-haskell-ng to fix segfaults

domenkozar:
See https://github.com/input-output-hk/cardano-sl/pull/2361 for context

### DIFF
--- a/block/cardano-sl-block.cabal
+++ b/block/cardano-sl-block.cabal
@@ -106,7 +106,7 @@ library
                       , mtl
                       , random
                       , reflection
-                      , rocksdb-haskell
+                      , rocksdb-haskell-ng
                       , safe-exceptions
                       , serokell-util
                       , stm
--- a/db/Pos/DB/Rocks/Functions.hs
+++ b/db/Pos/DB/Rocks/Functions.hs
@@ -19,10 +19,6 @@ module Pos.DB.Rocks.Functions
        , rocksDelete
        , rocksPutBi
 
-       -- * Snapshot
-       , Snapshot (..)
-       , usingSnapshot
-
        -- * Iteration
        , rocksIterSource
 
@@ -39,7 +35,6 @@ import           Universum
 import           Control.Lens (ASetter')
 import           Control.Monad.Trans.Resource (MonadResource)
 import           Data.Conduit (ConduitM, Source, bracketP, yield)
-import           Data.Default (def)
 import qualified Database.RocksDB as Rocks
 import           Ether.Internal (lensOf)
 import           System.Directory (createDirectoryIfMissing, doesDirectoryExist,
@@ -59,11 +54,12 @@ import qualified Pos.Util.Concurrent.RWLock as RWL
 ----------------------------------------------------------------------------
 
 openRocksDB :: MonadIO m => FilePath -> m DB
-openRocksDB fp = DB def def def
-                   <$> Rocks.open fp def
-                        { Rocks.createIfMissing = True
-                        , Rocks.compression     = Rocks.NoCompression
-                        }
+openRocksDB fp = DB Rocks.defaultReadOptions Rocks.defaultWriteOptions options
+                   <$> Rocks.open options
+  where options = (Rocks.defaultOptions fp)
+          { Rocks.optionsCreateIfMissing = True
+          , Rocks.optionsCompression     = Rocks.NoCompression
+          }
 
 closeRocksDB :: MonadIO m => DB -> m ()
 closeRocksDB = Rocks.close . rocksDB
@@ -150,6 +146,11 @@ rocksPutBi k v = rocksPutBytes k (dbSerializeValue v)
 -- Snapshot
 ----------------------------------------------------------------------------
 
+-- The following is not used in the project yet. So the
+-- rocksdb-haskell-ng binding doesn't include it yet, and we simply
+-- comment it out here, to be added back at a later stage when needed.
+
+{-
 newtype Snapshot = Snapshot Rocks.Snapshot
 
 usingSnapshot
@@ -160,6 +161,7 @@ usingSnapshot DB {..} action =
         (Rocks.createSnapshot rocksDB)
         (Rocks.releaseSnapshot rocksDB)
         (action . Snapshot)
+-}
 
 ----------------------------------------------------------------------------
 -- Iteration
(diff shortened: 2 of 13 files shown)


chrisdone:
@rockbmb Continuing our discussion here. I've RPC'd into the AppVeyor box on a branch, here's an interesting result I've found:

If I use some Chinese (`莊周`, Daoist philosopher) in a filename and try to open it with RocksDB from Haskell, this is what happens:

screenshot 2018-01-24 12 23 29

```
Database.RocksDB Database.RocksDB> open (defaultOptions "databases/made-from-haskell-no-unicode") {optionsCreateIfMissing = True }
Database.RocksDB Database.RocksDB> open (defaultOptions "databases/chinese-\33674\21608-made-from-haskell") {optionsCreateIfMissing = True }
Database.RocksDB Database.RocksDB> :!ls databases
hinese--made-from-haskell  chinese-??????-made-from-haskell  made-from-haskell-no-unicode
Database.RocksDB Database.RocksDB> import System.Directory
Database.RocksDB Database.RocksDB System.Directory> getDirectoryContents "databases"
"made-from-haskell-no-unicode","chinese-\33674\21608-made-from-haskell","chinese--made-from-haskell","..","."]
Database.RocksDB Database.RocksDB System.Directory> getDirectoryContents "databases/chinese--made-from-haskell"
"OPTIONS-000005","MANIFEST-000001","LOG","LOCK","IDENTITY","CURRENT","000003.log","..","."]
Database.RocksDB Database.RocksDB System.Directory> getDirectoryContents "databases/chinese-\33674\21608-made-from-haskell"
```

**Observations**

* You can ignore the `????` output, that's because the terminal or `ls` doesn't support the encoding properly. That's why I switch to `getDirectoryContents`.
* With `getDirectoryContents` you can see that *two* directories are created for the unicode filename. The first, by Haskell to try to create the directory ahead of time. That's the one with *correct* unicode.
* The directory `chinese--made-from-haskell` is created by Rocks, we know this because looking inside it has all the Rocks paraphernalia.

**Conclusions**

So it's clear that the test suite entry for Unicode filenames is hiding the real problem. The FFI function receives a string with those unicode characters stripped out, would be my hypothesis. Presumably because GHC's IO assumes a basic ASCII encoding.

I can go and confirm this by setting the encoding on Windows to be UTF16 and letting GHC take care of the encoding, or by using the `text` package to render the string via `T.encodeUtf16 . T.pack :: String -> ByteString` and pass the bytestring to the FFI function. I can try both in GHCi and report back.

chrisdone:
OK, I believe it should be the `FFI` function `withCString` that I need to encode myself. Going to try that now.

chrisdone:
@rockbmb I believe that we'll have to patch RocksDB to support Unicode filenames on Windows. Here's why:

* The method [`DB::Open`](https://github.com/facebook/rocksdb/blob/master/db/db_impl_open.cc#L1002) actually [calls `createDirIfMissing`](https://github.com/facebook/rocksdb/blob/master/db/db_impl_open.cc#L1002).
* That in turn calls [`CreateDir`](https://github.com/facebook/rocksdb/blob/master/port/win/env_win.cc#L4310) from `port/win/env_win.cc`.
* Finally, that in turn [calls `_mkdir`](https://github.com/facebook/rocksdb/blob/master/port/win/env_win.cc#L434), which is from `Direct.h` and has this type: `int _mkdir(const char* pathname)`.
* To properly support Unicode on Windows you typically use a wide char (`wchar_t` or equivalent Win32 API name). The bytes passed into `_mkdir` are interpreted as some kind of ASCII format where each byte is a character, as opposed to UTF-16 encoded byte pairs. That's why creating files works at all, and why when you try to pass a UTF-16 encoded byte vector, it interprets it as a null-terminated string of one character.

**So my recommendation is not to create filenames with unicode in it for RocksDB.** You could spend time patching RocksDB itself, but that seems out of scope of this particular PR. You should use filenames only and not try to include directory names (which could contain a user's full name in Unicode).

domenkozar:
@chrisdone could you file an issue for https://github.com/facebook/rocksdb please?

rockbmb:
@chrisdone thanks for the detailed explanation! This matches the findings I made exactly, and goes even further.

Do you think using any of the functions from [here](https://github.com/haskell/win32/blob/master/System/Win32/Encoding.hs) on the desired filepath when `open`ing a database would help?

chrisdone:
@rockbmb Seems worth a try. In my case the codepage is the one from DOS, [CP 437](https://en.wikipedia.org/wiki/Code_page_437), so it doesn't have many characters in it. Kind of extended Latin. I suppose that would be the best that we can get if we try that route.

chrisdone:
@rockbmb By the way, what's the use-case for having unicode names for databases? I think if we take this upstream to RocksDB, they're going to give push-back, arguing that your DB names can just be ASCII or something.

domenkozar:
@chrisdone we pass in absolute path on windows, which includes windows username in %APPDATA%.

chrisdone:
I've opened a PR here: https://github.com/facebook/rocksdb/issues/3408

rockbmb:
@domenkozar I don't know how long this `rocksdb` fix might take, if it comes at all, so I'm going to implement @cleverca22's suggested fix to be used in the meantime.

domenkozar:
@rockbmb which suggestion - let's discuss it

rockbmb:
@domenkozar here: https://iohk.myjetbrains.com/youtrack/issue/CSL-1803#comment=93-15877

We want to use the [getShortPathName](https://github.com/haskell/win32/blob/master/System/Win32/Info.hsc#L129) function, which replaces a path with a unique version of itself with all `Unicode` characters stripped.

I'll make a PR for this today or this weekend. This fix will be in `cardano-sl`'s launcher, and not in `rocksdb-haskell-ng` or `rocksdb-haskell`. It's just temporary until `rocksdb` does something (or nothing at all) about Windows' Unicode woes.

domenkozar:
This won't work, because people can disable `getShortPathName` in windows and then our app will break. You'll still have to fallback to not using it.

rockbmb:
In that case, I'm not sure how to fix this in Windows aside from not using absolute paths, which is still not sure to work.

Perhaps waiting for `rocksdb`'s issue outcome and, in the meantime, adding a section to the FAQ seems best.

domenkozar:
> Perhaps waiting for rocksdb's issue outcome and, in the meantime, adding a section to the FAQ seems best.

That's the plan :)

rockbmb:
I forgot to mention I was agreeing with you x)

I'll keep my eye on `rocksdb`'s issue to see if I can be of assistance.

PR #2370, opened 2018-01-24
[CSL-2205] Get rid of Block.Network.Announce module

volhovm:
This module wasn't properly formed -- it had one announcing method (what about `relayBlock` in `Network.Logic` then?) and one server-side listener method, which was there for no good reason.


volhovm:
Fair, also diffusion changes everything a lot. I better help Alex with diffusion.

PR #2373, opened 2018-01-24
CSL-2212 Eliminate `forkIO` usages in wallet 

gromakovsky:
All usages of `forkIO` and similar functions in wallet were replaced by functions from `async`.
It's better to review each commit separately, I provided brief explanations in commit messages.

### DIFF
--- a/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
+++ b/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
@@ -8,6 +8,7 @@ module Cardano.Wallet.Server.Plugins (
     , acidCleanupWorker
     , conversation
     , walletBackend
+    , resubmitterPlugin
     ) where
 
 import           Universum
@@ -87,10 +88,13 @@ walletBackend WalletBackendParams {..} =
       wsConn <- getWalletWebSockets
       ctx <- V0.walletWebModeContext
       syncWalletsWithGState =<< mapM findKey =<< myRootAddresses
-      startPendingTxsResubmitter
       launchNotifier (V0.convertHandler ctx)
       let app = upgradeApplicationWS wsConn $ serve API.walletAPI (API.walletServer (V0.convertHandler ctx))
       return $ withMiddleware walletRunMode app
+
+-- | A @Plugin@ to resubmit pending transactions.
+resubmitterPlugin :: (HasConfigurations, HasCompileInfo) => Plugin WalletWebMode
+resubmitterPlugin = ([ActionSpec $ \_ _ -> startPendingTxsResubmitter], mempty)
 
 -- | "Attaches" the middleware to this 'Application', if any.
 -- When running in debug mode, chances are we want to at least allow CORS to test the API
@@ -107,4 +111,3 @@ corsMiddleware = cors (const $ Just policy)
         { corsRequestHeaders = ["Content-Type"]
         , corsMethods = "PUT" : simpleMethods
         }
-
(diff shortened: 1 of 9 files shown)


wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -87,11 +88,14 @@ walletBackend WalletBackendParams {..} =
...
       ctx <- V0.walletWebModeContext
       syncWalletsWithGState =<< mapM findKey =<< myRootAddresses
-      startPendingTxsResubmitter
       launchNotifier (V0.convertHandler ctx)
       let app = upgradeApplicationWS wsConn $ serve API.walletAPI (API.walletServer (V0.convertHandler ctx))
       return $ withMiddleware walletRunMode app
 
+-- | A @Plugin@ to resubmit pending transactions.
+resubmitterPlugin :: (HasConfigurations, HasCompileInfo) => Plugin WalletWebMode
+resubmitterPlugin = ([ActionSpec $ \_ _ -> startPendingTxsResubmitter], mempty)
```
Martoon-00: Woa, I can't believe it works. Previously I couldn't do so because plugins had `RawRealMode` monad or smth like that.

wallet/src/Pos/Wallet/Web/Sockets/Notifier.hs
```diff
@@ -42,7 +42,7 @@ launchNotifier
     => (forall x. m x -> Handler x)
     -> m ()
 launchNotifier nat =
-    void . liftIO $ mapM startForking
+    void . liftIO $ mapConcurrently startNotifier
```
Martoon-00: `launchNotifier` becomes blocking (forever). Since it's used in wallet startup sequence as is (https://goo.gl/nbnnW9), wallet will never actually start

Martoon-00:
Also wallet has some `forkPingThread`. It contains quite simple logic, maybe you'd like to rewrite it with `async` as well

wallet/src/Pos/Wallet/Web/Sockets/Notifier.hs
gromakovsky: Oh, yes, of course. I should make it another plugin as I did with `startPendingTxsResubmitter`. Good catch.

gromakovsky:
> Also wallet has call of some `forkPingThread`. It contains quite simple logic, maybe you'd like to rewrite it with async as well

This function is provided by `websockets` package, maybe it's not perfect, but it's not my business, I think it's better to just use library function.
It shouldn't do anything bad anyway.

gromakovsky:
@Martoon-00 fixed.

wallet/src/Pos/Wallet/Web/Sockets/Notifier.hs
akegalj: My understnading is that Daedalus UI is not using notifier - so it is just a part in our code nobody is using.

My understanding is that UI is using http for all request nowdays.

akegalj: @Martoon-00 @gromakovsky see here https://input-output-rnd.slack.com/archives/C7PJ29FME/p1516873894000474?thread_ts=1516873660.000197&cid=C7PJ29FME . I believe notifier can be removed alltogether

gromakovsky: Created https://iohk.myjetbrains.com/youtrack/issue/CSL-2218, maybe someone will do it.

PR #2374, opened 2018-01-24
[CSL-2214] Eliminate forkAccompanion

gromakovsky:
This function is very similar to `withAsync`.
I used `withAsync` instead, because it's safer and more conventional.
I also removed `Unless` part from `runPeriodicallyUnless`, because it appeared
useless. That's because accompanion thread will die when main thread dies.


Martoon-00:
Yeah, when I wrote this I assumed we would use `time-warp` without `async`, and it's not even close to truth

PR #2376, opened 2018-01-25
[CSL-2057] Show full git SHA

akegalj:
Adds remainings  of https://github.com/input-output-hk/cardano-sl/pull/2127 into feature/new-wallet branch.

Tested:
look


adinapoli-iohk:
@akegalj LGTM -- have you checked locally that the newly-generated swagger diff shows the full SHA now? 😉

akegalj:
> have you checked locally that the newly-generated swagger diff shows the full SHA now?

not yet - build is in the progress. I will attach the pic here when I do

akegalj:
Recompiling - as it seems `stack` cached the wrong version somehow (ie, it shows short older hash)

akegalj:
For some strange reasons I am getting "Couldn't find git revision" from CompileInfo.hs which is thrown if git exits with error. I am getting this when i try compiling from scratch:
```
[akegalj@notebook cardano-sl]$ stack clean --full
[akegalj@notebook cardano-sl]$ stack build --ghc-options=" -Wwarn +RTS -A256m -n2m -RTS" --test --no-haddock-deps --bench --jobs=2 --no-run-tests --no-run-benchmarks --fast  cardano-sl-wallet-new
```

Currently I am testing against feature/new-wallet branch to see how it behaves.

Ps, instead of using our own implementation to fetch git revision we could have used gitrev from hackage/stackage https://www.stackage.org/package/gitrev

adinapoli-iohk:
> Ps, instead of using our own implementation to fetch git revision we could have used gitrev from hackage/stackage

It's overkill to use a package just for a one-line thing 😉 

> For some strange reasons I am getting "Couldn't find git revision" from CompileInfo.hs which is thrown if git exits with error.

Can you cut and paste the error? I think you would have much more success in pinpointing the error if you change this line to report the actual error (and the status code):

```
[...]
         pure $ case exitCode of
              ExitSuccess -> output
              _           -> "Couldn't fetch git revision"
[...]
```
Then you can debug your way to the solution.

akegalj:
@adinapoli-iohk yes, I am trying to debug what's wrong. Everything seems to be working in ghci:
```
.Timer Pos.Util.Util> $(retrieveCompileTimeInfo)                                                                                                              
CompileTimeInfo {ctiGitRevision = "df64a6d38e3b7259c468d0c41d3ceba5c3592db4"}
```

will ping when I figure it out.

akegalj:
@adinapoli-iohk everything works now. I was unable to reproduce the problem I had yesterday.

adinapoli-iohk:
@akegalj Can you squash those 4 commits into a single one, to preserve a clean git history? Apart from that, we are good.

akegalj:
@adinapoli-iohk ready for review after CI

PR #2377, opened 2018-01-25
[CSE-260] Fix `TypeError: ce.destroy(...) is not a function`

sectore:
Details https://iohk.myjetbrains.com/youtrack/issue/CSE-260

### DIFF
--- a/explorer/frontend/src/Lib/Waypoints/Waypoints.purs
+++ b/explorer/frontend/src/Lib/Waypoints/Waypoints.purs
@@ -1,8 +1,9 @@
 module Waypoints where
 
 import Prelude
+
 import Control.Monad.Eff (kind Effect, Eff)
-import Control.Monad.Eff.Uncurried (EffFn3, runEffFn3)
+import Control.Monad.Eff.Uncurried (EffFn1, EffFn3, runEffFn1, runEffFn3)
 import DOM.Node.Types (ElementId)
 import Data.Function.Uncurried (Fn1)
 import Data.Generic (class Generic, gShow)
@@ -42,4 +43,8 @@ waypoint elemId handler = runEffFn3 waypointImpl elemId handler defaultWaypointO
 waypoint' :: forall eff. ElementId -> (WaypointHandler eff) -> WaypointOffset -> Eff (waypoint :: WAYPOINT | eff) Waypoint
 waypoint' = runEffFn3 waypointImpl
 
-foreign import destroy :: forall eff. Fn1 Waypoint (Eff (waypoint :: WAYPOINT | eff) Unit)
+foreign import destroyImpl :: forall eff. EffFn1 (waypoint :: WAYPOINT | eff) Waypoint Waypoint
+
+-- | Destroys a `Waypoint` with all registered event handlers
+destroy :: forall eff. Waypoint -> Eff (waypoint :: WAYPOINT | eff) Waypoint
+destroy = runEffFn1 destroyImpl
--- a/explorer/frontend/src/Lib/Waypoints/Waypoints.js
+++ b/explorer/frontend/src/Lib/Waypoints/Waypoints.js
@@ -10,6 +10,6 @@ exports.waypointImpl = function (elementId, callback, offset) {
   });
 }
 
-exports.destroy = function (waypoint) {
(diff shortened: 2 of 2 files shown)


explorer/frontend/src/Lib/Waypoints/Waypoints.purs
```diff
@@ -42,4 +43,8 @@ waypoint elemId handler = runEffFn3 waypointImpl elemId handler defaultWaypointO
 waypoint' :: forall eff. ElementId -> (WaypointHandler eff) -> WaypointOffset -> Eff (waypoint :: WAYPOINT | eff) Waypoint
 waypoint' = runEffFn3 waypointImpl
 
-foreign import destroy :: forall eff. Fn1 Waypoint (Eff (waypoint :: WAYPOINT | eff) Unit)
```
akegalj: @sectore I am not clear why the error was occuring at all? Isn't the above correct as well

explorer/frontend/src/Lib/Waypoints/Waypoints.purs
```diff
@@ -42,4 +43,8 @@ waypoint elemId handler = runEffFn3 waypointImpl elemId handler defaultWaypointO
...
+foreign import destroyImpl :: forall eff. EffFn1 (waypoint :: WAYPOINT | eff) Waypoint Waypoint
```
akegalj: @sectore I don't know the details, but shuldn't destroy return `Unit` instead of `Waypoint`.  Something that is destroy shouldn't return reference to it (in general - it makes a little sense). Did you mean:
```haskell
foreign import destroyImpl :: forall eff. EffFn1 (waypoint :: WAYPOINT | eff) Waypoint Unit
```

?

explorer/frontend/src/Lib/Waypoints/Waypoints.purs
akegalj: ps, have you tried with:
```haskell
foreign import destroy :: forall eff. Fn1 Waypoint (Eff (waypoint :: WAYPOINT | eff) Waypoint)
```

?

sectore: The foreign function `destroy` was called directly from `Update.purs` (see https://github.com/input-output-hk/cardano-sl/blob/develop/explorer/frontend/src/Explorer/Update.purs#L487), which occurs that error. Wrapping it around to not using a foreign directly fixed this error.

explorer/frontend/src/Lib/Waypoints/Waypoints.purs
sectore: `destroy` returns an instance of `Waypoint` - see http://imakewebthings.com/waypoints/api/destroy/

sectore: Waypoint is used to animate the size of the header of Dashboard depending on current scroll position.

akegalj: hm, in that case I believe https://github.com/input-output-hk/cardano-sl/pull/2377#discussion_r163841639 should be sufficient. Right?

akegalj: Can you try just with this modification (just to check it out):
```
foreign import destroy :: forall eff. Fn1 Waypoint (Eff (waypoint :: WAYPOINT | eff) Waypoint)
```

ie, that it returns Waypoint instead of Unit. Because that should be the only difference between the old and new implementation.

sectore: Just tried it ^ out as you suggested. It still throws `TypeError: Waypoints.destroy(...) is not a function`. The only way to fix this seems to be wrapping that foreign function.

akegalj: Strange. Ok, lets then proceed with it.

akegalj: Ah no, it makes sense. In fact I was always using EffFn modifiers and they worked (don't have much exp with Fn1 modifiers and should they be unwrapped somhow as well as with EffFn and runEffFn). Approving

PR #2378, opened 2018-01-25
[CSL-2217] Increase networkConnectionTimout to 15s

coot:


avieth:
https://github.com/input-output-hk/cardano-sl/pull/2371

PR #2324, opened 2018-01-12
[CSL-2043] [CSL-2149] Block serialization/deserialization improvements (with benchmarks)

avieth:
Title recently changes to reflect the fact that this should eventually be merged.
All that remains to do is replace the various verifications which were formerly done in the deserialization instance.

# Block serialization benchmarks

## Purpose

To prove or disprove, with evidence, my claim that verification in
deserialization is a significant performance problem, and to discover the
extent to which recent optimizations to `Digest` marshalling will offer any
improvements in deserialization and serialization alike.

All this in the context of increased pressure on our relays due to ADA hype,
and the longstanding issue of a painfully slow initial wallet sync.
Improvements in serialization and deserialization performance could reduce
relay load as well as block sync latency.

## Test setup

See `block/bench/Block.hs`.

The `Arbitrary` instances for `MainBlock` is used to come up with test subjects.
The deterministic `QCGen` from `QuickCheck` is seeded according to the
environment variable `SEED` and the size of the subjects is determined by the
environment variable `SIZE`.

The uses of `makeSmall` from `cardano-sl-util` were banished. This term
takes the square root of a third of the size, and was applied to, among other
things, block bodies. It in essence corrupted the data.

The benchmark suite includes deserialization and serialization of the block,
block header, and each of their components, including all components of the
block body (payloads).

The suite must be compiled against different revisions of the rest of
cardano-sl, corresponding to the three cases:

  - Verification on deserialization, no further optimizations (3451d1b32d6f86a98ca7f1cc9c1de34ac4a32114).
  - No verification on deserialization, no further optimizations (3451d1b32d6f86a98ca7f1cc9c1de34ac4a32114).
  - No verification on deserialization, various optimizations (47652100600aff15da3569b67d520a697315e911).

The "various optimizations" include:

  - More efficient `Digest`/`ByteString` conversion (requires new basement,
    memory, cryptonite (unreleased) packages).
  - Fewer uses of `LBS.toStrict`.
  - Custom `Builder` strategies for serialization.

`criterion` is used to gather data. With my recent patches (tip of mainline
at the moment, but unreleased) it now properly collects RTS numbers and so we
can regress bytes allocated over iterations:

```
SEED=0 SIZE=32 ./block/bench/Block --regress allocated:iters +RTS -T -RTS
```

All exectuables compiled with `-O2`.
All tests were run with SEED=4, SIZE=32, giving a block of size 77,496 bytes.

## Results

- Significant improvements in deserialization time when dropping validation:
[...]

PR #2380, opened 2018-01-25
[CSL-2072] Refined verification framework

volhovm:
This PR:
* Improves some datatypes (delegation mostly) by removing implicit consistency rules to types (changed list to set in several places, added wrappers here and there).
* Introduces pure verification framework: See `Pos.Util.Verification`, it has docs.


avieth:
I want to point out that #2324 includes verification, and indeed makes it "granular" (composites are verified in terms of their components). Reviewing and merging that must be a priority over this patch.

volhovm:
@avieth sure, though this PR is about _secure_ verification granularity. I insist on putting this patch into release anyway, even if if #2324 is merged before.

auxx/src/Command/TyProjection.hs
```diff
@@ -149,10 +149,10 @@ tyProposeUpdateSystem :: TyProjection ProposeUpdateSystem
 tyProposeUpdateSystem = TyProjection "ProposeUpdateSystem" (preview _ValueProposeUpdateSystem)
 
 tySystemTag :: TyProjection SystemTag
-tySystemTag = TyProjection "SystemTag" ((fmap . fmap) (SystemTag . fromString) (preview _ValueString))
+tySystemTag = TyProjection "SystemTag" ((fmap . fmap) (UnsafeSystemTag . fromString) (preview _ValueString))
 
 tyApplicationName :: TyProjection ApplicationName
-tyApplicationName = TyProjection "ApplicationName" ((fmap . fmap) (ApplicationName . fromString) (preview _ValueString))
+tyApplicationName = TyProjection "ApplicationName" ((fmap . fmap) (UnsafeApplicationName . fromString) (preview _ValueString))
```
avieth: We should drop this `Unsafe*` naming convention. There's nothing unsafe about using a constructor. It doesn't risk any segmentation fault or undefined behaviour or breakage of referential transparency.

block/cardano-sl-block.cabal
```diff
@@ -157,22 +157,22 @@ library
...
+--                   , cardano-sl-block
+--                   , cardano-sl-crypto
+--                   , cardano-sl-core
+--                   , cardano-sl-txp
+--                   , cardano-sl-update
+--                   , cardano-sl-ssc
+--                   , cardano-sl-delegation
+--                   , formatting
+--                   , universum
+--                   , deepseq
```
avieth: Why comment this out? Need me to fix something here?

block/src/Pos/Block/Pure.hs
```diff
@@ -59,9 +60,9 @@ data VerifyHeaderParams = VerifyHeaderParams
       -- ^ Check that header has no unknown attributes.
     } deriving (Eq, Show)
 
-verifyFromEither :: Text -> Either a b -> VerificationRes
-verifyFromEither txt (Left _)  = verifyGeneric [(False, txt)]
-verifyFromEither txt (Right _) = verifyGeneric [(True, txt)]
+verifyFromEither :: Text -> Either Text () -> VerificationRes
```
avieth: Doesn't this violate those exception guidelines? @int-index

block/src/Pos/Block/Verification.hs
```diff
@@ -0,0 +1,158 @@
+{-# LANGUAGE TypeFamilies  #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- This module could be in core, but 'verifySscPayload' is kind of
+-- complex and I think it can't be moved there.
+
+-- | Pure verification for block. See "Pos.Util.Verification" for more info.
+
+module Pos.Block.Verification () where
```
avieth: So everything is done by typeclass instances? What's the reason for this? What's the advantage over just using a value (verification function)?

block/src/Pos/Block/Verification.hs
```diff
@@ -0,0 +1,158 @@
+{-# LANGUAGE TypeFamilies  #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- This module could be in core, but 'verifySscPayload' is kind of
```
avieth: Stale comment? Is this module to do with SSC?

core/Pos/Core/Common/Types.hs
```diff
@@ -303,12 +302,16 @@ maxCoinVal = 45000000000000000
...
+          Right () -> c
 
-checkCoin :: MonadError Text m => Coin -> m ()
-checkCoin (Coin c)
-    | c <= maxCoinVal = pure ()
-    | otherwise       = throwError $ "Coin: " <> show c <> " is too large"
+instance PVerifiable Coin where
+    pverifyOne (UnsafeCoin c)
+        | c <= maxCoinVal = pass
+        | otherwise       = pverFail $ "Coin: " <> show c <> " is too large"
```
avieth: Why don't you just define `verifyCoin` top-level? Is there any value in having this `PVerifiable` class? Seems like an overcomplication.

avieth:
> I insist on putting this patch into release anyway, even if if #2324 is merged before.

You may have to wait a while. There are more important things to do. In terms of cost/benefit, this patch scores low.

block/src/Pos/Block/Pure.hs
int-index: an ADT would be definitely preferable to raw `Text`

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,96 @@
...
+
+import           Pos.Util.Util (eitherToThrow)
+
+-- Prefix P is annoying, but we need to distinguish between pure
+-- verification and VAR somehow.
+
+-- | Pure verification monad. It accumulates errors along the
+-- verification (AST) path.
+newtype PVer a = PVer
+    { getPVer :: Either [Text] a
```
int-index: Not `Text`, no.

block/cardano-sl-block.cabal
volhovm: Yes, in your PR. I'll drop this patch when the base branch will work.

auxx/src/Command/TyProjection.hs
volhovm: Let others decide. I find it to be pretty neat. cc @gromakovsky @int-index

block/src/Pos/Block/Verification.hs
volhovm: The comment is highly relevant.

core/Pos/Core/Common/Types.hs
volhovm: It's not. 
* Forces programmer to think about verification in two levels (self + recursive)
* Unifies the standard about how to name these validation functions
* Forbids multiple verifiers for the same type
* Forces intermediate datatypes to be verified as well (not only sub-sub-children, but also children)

block/cardano-sl-block.cabal
volhovm: BTW you can review by commits ;)

georgeee:
> A "refined verification framework" is simply not necessary. We could include this later on but #2324 does not need this and should be expedited.

@avieth please respect the opinion of your colleagues. 

@volhovm aims to solve a real problem which is about maintaining clarity of what is being checked, at which stage, which way. Before we had one rule of thumb "everything being in use is fully verified", now we have these checks being separated, but still, we want to have clarity whenever possible if data was checked or not. I like Mikhail's idea of proposing unified type class serving as framework for verification.

block/src/Pos/Block/Verification.hs
georgeee: Advantage is that it's clear for you, which function to trigger to verify the data.

auxx/src/Command/TyProjection.hs
georgeee: It's certainly good naming convention for such Unchecked constructors. Maybe only we should use `Unchecked` or other prefix instead of `Unsafe`

core/Pos/Binary/Core/Delegation.hs
```diff
@@ -1,16 +1,32 @@
...
-instance HasConfiguration => Bi DlgPayload where
-    encode = encode . getDlgPayload
-    decode = UnsafeDlgPayload <$> decode
+instance Bi DlgPayload where
+    encode = encode . S.toList . getDlgPayload
+    decode = do
+        (psks :: [ProxySKHeavy]) <- decode
+        let asSet :: Set ProxySKHeavy
+            asSet = S.fromList psks
+        when (length psks /= length (ordNub psks)) $
```
gromakovsky: `length` is `O(1)` for `Set`, so I would use `length asSet` instead of `length (ordNub psks)`. In most cases `asSet` will be forced, so it should be more efficient (and clear).

core/Pos/Core/Delegation/Util.hs
```diff
@@ -1,28 +0,0 @@
...
-checkDlgPayload it = do
-    unless (null duplicates) $
-        throwError "Some of block's PSKs have the same issuer, which is prohibited"
-    forM_ proxySKs validateProxySecretKey
-  where
-    proxySKs = getDlgPayload it
-    proxySKsDups psks =
-        filter (\x -> length x > 1) $
-        groupBy ((==) `on` pskIssuerPk) $ sortOn pskIssuerPk psks
-    duplicates = proxySKsDups proxySKs
```
gromakovsky: So previously we were checking that there are no such `psk1` and `psk2` that `pskIssuerPk psk1 == pskIssuerPk psk2`, but now it's not checked. Is it ok?

delegation/src/Pos/Delegation/Helpers.hs
```diff
@@ -38,6 +39,6 @@ isRevokePsk = isSelfSignedPsk
 dlgMemPoolApplyBlock :: MainBlock -> DlgMemPool -> DlgMemPool
 dlgMemPoolApplyBlock block m = flip execState m $ do
     let (toDelete,toReplace) =
-            partition isRevokePsk (getDlgPayload $ block ^. mainBlockDlgPayload)
+            partition isRevokePsk (toList $ getDlgPayload $ block ^. mainBlockDlgPayload)
```
gromakovsky: It would probably make sense to define `ToList` (or `Container` in old `universum`) for `DlgPayload`.

lib/src/Pos/SafeCopy.hs
```diff
@@ -362,8 +362,16 @@ instance SafeCopy (AsBinary a) where
...
     getCopy = getCopyBi
+    putCopy = putCopyBi
+
+instance SafeCopy HeavyDlgIndex where
+    getCopy = contain $ HeavyDlgIndex <$> safeGet
+    putCopy x = contain $ safePut $ getHeavyDlgIndex x
+
+instance SafeCopy LightDlgIndices where
+    getCopy = contain $ LightDlgIndices <$> safeGet
+    putCopy x = contain $ safePut $ getLightDlgIndices x
```
gromakovsky: Why did you write these instances manually instead of generating them (as it's done for 90% types)?

core/Pos/Core/Delegation/Util.hs
volhovm: Yes. Now `DlgPayload` has a `Set` inside, so the property is guaranteed in the decoder.

core/Pos/Binary/Core/Delegation.hs
volhovm: Fixed.

lib/src/Pos/SafeCopy.hs
volhovm: I am not really sure what exactly will `deriveSafeCopySimple` generate.  Some other wrappers (that doesn't have their own binary representation) are also implemented in the similar way.

Example:
```
instance SafeCopy (Signature a) where
    putCopy (Signature sig) = contain $ safePut sig
    getCopy = contain $ Signature <$> safeGet
```

core/Pos/Core/Delegation/Util.hs
gromakovsky: But `Set` will only guarantee that there are no such `psk1` and `psk2` that `psk1 == psk2`. Isn't it still possible to have `psk1` and `psk2` such that `pskIssuerPk psk1 == pskIssuerPk psk2`?

volhovm: Oh, you're right. No idea how did I miss that. Maybe, I should revert this particular refactoring then.

volhovm: @gromakovsky I've reverted these `list -> set` changes.

delegation/src/Pos/Binary/Delegation.hs
```diff
@@ -6,19 +6,34 @@ module Pos.Binary.Delegation
...
+        encode duPsks <>
+        encode duPrevEpochPosted
+    decode = do
+        enforceSize "DlgUndo" 2
+
+        (duPsks :: [ProxySKHeavy]) <- decode
+        let duPsksSet :: Set PublicKey
+            duPsksSet = Set.fromList $ map pskIssuerPk duPsks
+        when (length duPsks /= Set.size duPsksSet) $
+              fail "DlgUndo.duPsks is not a set: it has duplicates in issuers"
```
gromakovsky: AFAIU, changes in this file ought to be reverted.

delegation/src/Pos/Delegation/Logic/Mempool.hs
```diff
@@ -52,7 +53,9 @@ import           Pos.Util.Concurrent.PriorityLock (Priority (..))
 getDlgMempool
     :: (MonadIO m, MonadDBRead m, MonadDelegation ctx m, MonadMask m)
     => m DlgPayload
-getDlgMempool = UnsafeDlgPayload <$> (runDelegationStateAction $ uses dwProxySKPool HM.elems)
+getDlgMempool =
+    fmap (UnsafeDlgPayload . toList) $
+    runDelegationStateAction $ uses dwProxySKPool (S.fromList . HM.elems)
```
gromakovsky: Revert these changes please, you convert to a `Set` and then back to a list.

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -71,10 +71,12 @@ type ReverseTrans = HashMap StakeholderId (HashSet StakeholderId, HashSet Stakeh
 -- Takes a set of dlg edge actions to apply and returns compensations
 -- to dlgTrans and dlgTransRev parts of delegation db. Should be
 -- executed under shared Gstate DB lock.
+--
+-- eActions list passed must be effectively a set.
```
gromakovsky: Why did you change it from a set to a list then?

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -482,12 +483,12 @@ dlgRollbackBlocks dlgBlunds = do
...
-        transCorrections <- calculateTransCorrections $ HS.fromList edgeActions
-        let pskOp = SomeBatchOp (map GS.PskFromEdgeAction edgeActions) <> transCorrections
+            edgeActions =
+                map (DlgEdgeDel . addressHash) backDeleted <> map DlgEdgeAdd duPsks
+        transCorrections <- calculateTransCorrections $ edgeActions
+        let pskOp = SomeBatchOp (map GS.PskFromEdgeAction $ edgeActions) <> transCorrections
         -- we should also delete issuers from "posted this epoch already"
-        let postedOp = SomeBatchOp $ map (GS.DelPostedThisEpoch . addressHash) issuers
+        let postedOp = SomeBatchOp $ map (GS.DelPostedThisEpoch . addressHash) $ toList issuers
         pure $ pskOp <> postedOp
```
gromakovsky: Tbh, I don't understand what is goings ons in this file. Did you forget to revert smth here?
AFAIU, some changes are not necessary and not useful, can you revert them please?

infra/Pos/Arbitrary/Infra.hs
```diff
@@ -42,5 +42,5 @@ instance Arbitrary VerInfo where
     shrink = genericShrink
 
 instance HasConfiguration => Arbitrary (DataMsg ProxySKHeavy) where
-    arbitrary = genericArbitrary
-    shrink = genericShrink
+    arbitrary = DataMsg <$> arbitrary
+    shrink (DataMsg x) = map DataMsg $ shrink x
```
gromakovsky: Can you explain this change?

auxx/src/Command/TyProjection.hs
gromakovsky: After recent changes I don't like `Unsafe` prefix. `Unchecked` is ok.

core/Pos/Core/Delegation.hs
```diff
@@ -1,2 +1,117 @@
...
+-- in).
+newtype HeavyDlgIndex =
+    HeavyDlgIndex { getHeavyDlgIndex :: EpochIndex }
+    deriving (Show, Eq, Ord, Generic)
+
+instance NFData HeavyDlgIndex
+instance Hashable HeavyDlgIndex
+
+instance Buildable HeavyDlgIndex where
+    build (HeavyDlgIndex i) = bprint build i
```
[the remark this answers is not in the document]
volhovm: Fair enough.

delegation/src/Pos/Delegation/Logic/VAR.hs
volhovm: Because there was an extra conversion.

delegation/src/Pos/Binary/Delegation.hs
volhovm: You're right, but I want to leave this one.

crypto/Pos/Crypto/Signing/Signing.hs
```diff
@@ -180,4 +181,4 @@ proxyVerify t ProxySignature{..} omegaPred m =
                  , signTag t
                  , Bi.serialize' m
                  ])
-            psigSig
+            (psigSig psig)
```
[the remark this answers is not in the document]
volhovm: This style was introduced by neongreen, nothing i can do.

infra/Pos/Arbitrary/Infra.hs
volhovm: This is legacy. Sorry, I'll see what I can clean up. This PR is 100 years old :(

update/Pos/Update/Poll/Logic/Apply.hs
```diff
@@ -66,7 +66,7 @@ verifyAndApplyUSPayload ::
     -> m ()
 verifyAndApplyUSPayload lastAdopted verifyAllIsKnown slotOrHeader upp@UpdatePayload {..} = do
     -- First of all, we verify data.
-    either (throwError . PollInvalidUpdatePayload) pure =<< runExceptT (checkUpdatePayload upp)
+    either (throwError . PollInvalidUpdatePayload . show) pure $ runPVerify upp
```
gromakovsky: 1. I don't like that you use `show` in all places where you need to convert `VerError` to `Text`. We want to format error prettily. The right function for it is `pretty`. `show` is not for pretty printing.
2. Why do you `show` here at all? Why not change argument of `PollInvalidUpdatePayload` to `VerError`? It would give more information. It will be more useful when we rewrite error handling in `Pos.Util.Verification` properly.

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,96 @@
+{- |
+
+Machinery to support pure datatype verification.
+
+1. Datatypes can have and not have internal pure consistency. If
+datatype doesn't have any implicit rules, it's called safe. 'Integer'
+is safe, 'Either Text String' is safe too. Otherwise, it's called
+unsafe. 'Coin' is unsafe, because it has implicit upper bound on the
```
gromakovsky: Can we use a different term please? Calling types safe/unsafe is really strange.
Previously some constructors were called unsafe, because calling them could lead to invariants violation.
1. Term unsafe was applied to **constructors**, not to types.
2. Now it's not the case, we can't break any invariants.

We can call types complex and simple, for example. Just a random naming idea, but it's still better than safe/unsafe, imo.

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,96 @@
...
+value, which is lower than 'Word64' bound. Newtype over string that is
+supposed to contain only strings of length less than 30 symbols is
+unsafe.
+
+2. Unsafe types' constructors should have prefix \"Unsafe\" (for
+example, 'UnsafeProxySecretKey'). They should also have description of
```
gromakovsky: George proposed `Unchecked` prefix, I like it much more than `Unsafe`.
`Unsafe`, in my opinion, means something like «don't call me unless you know what u'r doin'». But that's not the case for these constructors.

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,96 @@
...
+why are they unsafe and provide some easy-to-use builders (@mkX@) if
+needed.
+
+3. If datatype @S@ is a part of @T@ (e.g. record field), then @T@
+is not called unsafe.
```
gromakovsky: If I interpret this statement correctly, you forgot to write that `S` is unsafe and that `T` itself doesn't have internal pure consistency apart from consistency of its parts.

util/Pos/Util/Verification.hs
gromakovsky: Shouldn't it be `NonEmpty Text`?

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,96 @@
...
+
+-- | Fail inside the 'PVer' monad.
+pverFail :: Text -> PVer ()
+pverFail t = PVer (Left [t])
+
+-- | Verifies some field, prefixing with the text value in case of
+-- error. Prefix is supposed to be the record field name. Use it when
+-- you want to specify which field/component are you verifying.
+pverField :: Text -> PVer () -> PVer ()
+pverField p (PVer v) = PVer $ first (\x -> p:x) v
```
gromakovsky: Hm, so `Left` in `getPVer` accumulates values which are semantically different, right? It looks like a bad idea :/

gromakovsky: Do we have an idea how to modify `PVer` type to do proper exception handling (using ADTs instead of `Text`)?

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,96 @@
...
+3. If datatype @S@ is a part of @T@ (e.g. record field), then @T@
+is not called unsafe.
+
+4. @T@ is called verifiable if it is unsafe or any of its parts are
+verifiable.
+
+5. Every verifiable type must have 'PVerifiable' instance. See
+description of this class.
+
+-}
```
gromakovsky: I saw that somewhere you wrote why this module is useful. I think you should also write it here, because at first glance it looks like unnecessary overcomplication.

crypto/Pos/Crypto/Signing/Signing.hs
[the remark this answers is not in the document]
gromakovsky: Which style? AFAIU, this code is equivalent to the previous version, you only did some cosmetic changes (which are subjective and I can't say you made this code better). It's better not to do them (or at least do them in a separate commit), because it complicates code review.
If you changed anything else in this module, then reviewing these cosmetic changes would be a bit less annoying.
Maybe that's just a leftover, because you had to revert smth.

delegation/src/Pos/Delegation/Helpers.hs
```diff
@@ -1,43 +0,0 @@
...
-import           Pos.Core (EpochIndex, HeavyDlgIndex (..))
-import           Pos.Core.Block.Main (MainBlock, mainBlockDlgPayload)
-import           Pos.Crypto (ProxySecretKey (..), isSelfSignedPsk)
-import           Pos.Delegation.Types (DlgMemPool, DlgPayload (getDlgPayload))
-
--- | Verify delegation payload without using GState. This function can
--- be used for block verification in isolation, also it can be used
--- for mempool verification.
-dlgVerifyPayload :: MonadError Text m => EpochIndex -> DlgPayload -> m ()
-dlgVerifyPayload epoch (getDlgPayload -> proxySKs) =
```
gromakovsky: Where is this logic now?

txp/Pos/Txp/Logic/Local.hs
```diff
@@ -90,15 +91,18 @@ txProcessTransaction itw =
...
+    => (TxId, TxAux)
+    -> m (Either ToilVerFailure ())
 txProcessTransactionNoLock =
     txProcessTransactionAbstract
         buildProccessTxContext
-        processTx
+        processTxLocal
+  where
+    processTxLocal e t@(_,txAux) = do
+        either (throwError . ToilInconsistentTxAux . show) pure (runPVerify txAux)
```
gromakovsky: No, move it to `Toil.processTx` please. It's not used for VAR, only for local txp.

gromakovsky: Otherwise explorer's txp is kinda broken.

delegation/src/Pos/Delegation/Helpers.hs
volhovm: It is in `PVerifiable` instance.

txp/Pos/Txp/Logic/Local.hs
volhovm: Alright, thank you.

util/Pos/Util/Verification.hs
volhovm: @gromakovsky I do not quite understand what does "semantically different" mean here. What I wanted to express with `PVer` is that user can provide hints along the verification path. The good analogy is "the term we failed to parse". It cannot be simply (without TH) automated. I think it's more or less optimal solution. Number of places this is used in is low. It is mostly needed when `A` has two fields of type `B` and we want to point out which one lead to error.

>Do we have an idea how to modify PVer type to do proper exception handling

I think we could define a type family. But again, despite agreeing this is important, I consider this to be a separate task.

util/Pos/Util/Verification.hs
volhovm: With current design it is impossible to throw `[]`, but yes, I'll better change it. Thanks.

util/Pos/Util/Verification.hs
volhovm: Ok, will add.

util/Pos/Util/Verification.hs
volhovm: It makes sense. Let's use checked/unchecked.

util/Pos/Util/Verification.hs
volhovm: Fixed.

delegation/src/Pos/Delegation/Logic/Mempool.hs
gromakovsky: Once again, revert it please.

delegation/cardano-sl-delegation.cabal
```diff
@@ -52,6 +52,7 @@ library
                      , cardano-sl-networking
                      , cardano-sl-util
                      , conduit
+                     , containers
```
gromakovsky: If you revert all unnecessary changes, this dependency shouldn't be necessary I suppose.

delegation/src/Pos/Arbitrary/Delegation.hs
```diff
@@ -7,27 +7,28 @@ module Pos.Arbitrary.Delegation
...
+    genPSK = createPsk <$> arbitrary <*> arbitrary <*> pure (HeavyDlgIndex epoch)
 
 instance HasConfiguration => Arbitrary DlgPayload where
     arbitrary = arbitrary >>= genDlgPayload
     shrink = genericShrink
 
 instance HasConfiguration => Arbitrary DlgUndo where
-    arbitrary = genericArbitrary
+    arbitrary =
+        genericArbitrary `suchThat` (allDistinct . map pskIssuerPk . duPsks)
```
gromakovsky: Ideally it should be written without `suchThat`, because it's likely inefficient. You can generate a list of PSKs with different issuers quite easily (e. g. create a list and use some kind of `nub` which takes a function).

delegation/src/Pos/Delegation/Logic/VAR.hs
gromakovsky: Okay, then I would prefer seeing it in a separate commit, but it's minor, doesn't really matter.

delegation/src/Pos/Delegation/Logic/VAR.hs
gromakovsky: So apparently all changes in this file are either cosmetic or optimization to avoid extra conversion between `HashSet` and list. Putting them into a separate commit would make review easier.

generator/src/Pos/Generator/Block/Mode.hs
```diff
@@ -21,7 +21,7 @@ module Pos.Generator.Block.Mode
 import           Universum
 
 import           Control.Lens.TH (makeLensesWith)
-import qualified Control.Monad.Catch as UnsafeExc
+import qualified Control.Monad.Catch as UncheckedExc
```
gromakovsky: Seriously?

networking/src/Node/Internal.hs
```diff
@@ -47,7 +47,7 @@ module Node.Internal (
 import           Control.Exception.Safe (Exception, MonadCatch, MonadMask, MonadThrow,
                                          SomeException, bracket, catch, finally, throwM)
 import           Control.Monad (forM, forM_, when)
-import qualified Control.Monad.Catch as UnsafeExc
+import qualified Control.Monad.Catch as UncheckedExc
```
gromakovsky: Seriously?

networking/src/Mockable/Concurrent.hs
```diff
@@ -162,7 +162,7 @@ data Async m t where
     Race :: m t -> m r -> Async m (Either t r)
     -- | This is needed to implement 'withAsyncWithUnmask'. Note that
     -- 'unsafeUnmask' function is not exported.
-    UnsafeUnmask :: m a -> Async m a
+    UncheckedUnmask :: m a -> Async m a
```
gromakovsky: No, it doesn't make sense, the function is called `unsafeUnmask`.

wallet-new/test/unit/UTxO/Verify.hs
```diff
@@ -50,7 +50,7 @@ import           Serokell.Util.Verify
 --
 -- Unsafe constructor since the various fields must be in agreement with
 -- each other. See 'verifyEnv'.
-data VerifyEnv = UnsafeVerifyEnv {
+data VerifyEnv = UncheckedVerifyEnv {
```
gromakovsky: Not sure it should be renamed.

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,101 @@
+{- |
+
+Machinery to support pure datatype verification.
```
gromakovsky: IIRC you were going to enumerate advantages of this machinery (as I requested), but I still don't see them.

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,101 @@
...
+
+import           Universum
+
+import qualified Data.List.NonEmpty as NE
+import qualified Data.Text as T
+import           Formatting (sformat)
+import           Serokell.Util (VerificationRes (..), listJson, verResToMonadError)
+
+-- Prefix P is annoying, but we need to distinguish between pure
+-- verification and VAR somehow.
```
gromakovsky: And that's actually subjective! Previously we indeed had this distinction, and it was for a good reason: we had static information about these types' verification status (actually we statically knew they were always verified). Now we don't know anything statically. Hence it's not clear whether there are real benefits in having this separation.
I mean that we can move these checks into VAR and probably we won't lose anything, will we?

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,101 @@
...
+
+-- | Pure verification monad. It accumulates errors along the
+-- verification (AST) path.
+newtype PVer a = PVer
+    { getPVer :: Either (NonEmpty Text) a
```
gromakovsky: Btw, there is also [Validation](https://www.stackage.org/haddock/nightly-2018-03-07/validation-1/Data-Validation.html).

util/Pos/Util/Verification.hs
```diff
@@ -0,0 +1,101 @@
...
+-- | Things that can be (purely) verified.
+class PVerifiable a where
+
+    -- | Verify the 'a' predicates only, assuming that all subfields
+    -- are correct. Is needed to be implemented for unsafe types only.
+    pverifySelf :: a -> PVer ()
+
+    -- | Verify all subcomponents and the component itself. Basically,
+    -- call 'pverify' for subcomponents and 'pverifySelf' in the end.
+    pverify :: a -> PVer ()
```
gromakovsky: Wouldn't it make more sense to have `pverifySelf` and `pverifyParts` and then `pverify a = pverifyParts a >>= pverifySelf a` (outside this class probably)? This way one wouldn't have to call `pverifySelf` in definition of `pverify`.

generator/src/Pos/Generator/Block/Mode.hs
volhovm: Whoops, I checked the `grep` output, but missed those. Sorry!

delegation/cardano-sl-delegation.cabal
volhovm: Fixed.

delegation/src/Pos/Delegation/Logic/Mempool.hs
volhovm: Done.

delegation/src/Pos/Delegation/Logic/VAR.hs
volhovm: They are indeed.

delegation/src/Pos/Arbitrary/Delegation.hs
volhovm: I think this change is not important anyway, I've reverted it.

util/Pos/Util/Verification.hs
volhovm: Added.

util/Pos/Util/Verification.hs
volhovm: Well, sometimes you want to only build some datatype, like update proposal, without validating it with respect to update gstate.

util/Pos/Util/Verification.hs
volhovm: Ok, fixed.

volhovm: It's not that trivial as you say, but I thought again and I think I have an idea.

volhovm: I introduced a `PVerifiableSub` wrapper which solves the problem you are describing.

util/Pos/Util/Verification.hs
volhovm: See the Pos.Util.Validation description. I've added some motivation.

akegalj:
@avieth 
as you were involved into relevant work - do you think it makes sense to close this one? There are so many conflicts that I don't think it makes sense to keep this open. I propose to close this PR and open a new one to reuse ideas from this PR if still useful?

avieth:
> as you were involved into relevant work - do you think it makes sense to close this one? There are so many conflicts that I don't think it makes sense to keep this open. I propose to close this PR and open a new one to reuse ideas from this PR if still useful?

It makes sense to close it. I'll update https://iohk.myjetbrains.com/youtrack/issue/CSL-2072 as well

PR #2382, opened 2018-01-25
[CSL-2213] Eliminate fork in Pos.Network.CLI.monitorStaticConfig

gromakovsky:
Now monitoring is launched as a separate worker.
I had to make `m` in `StaticPeers` concrete, because otherwise GHC panics.
It should be ok, because we know that it's always IO-based anyway.
Monad abstractions are not essential now, because big refactoring of
monadic approach is coming.

### DIFF
--- a/infra/Pos/Network/Types.hs
+++ b/infra/Pos/Network/Types.hs
@@ -45,10 +45,9 @@ module Pos.Network.Types
        , NodeId (..)
        ) where
 
-import           Universum hiding (show)
+import           Universum
 
 import           Data.IP (IPv4)
-import           GHC.Show (Show (..))
 import           Network.Broadcast.OutboundQueue (OutboundQ)
 import qualified Network.Broadcast.OutboundQueue as OQ
 import           Network.Broadcast.OutboundQueue.Types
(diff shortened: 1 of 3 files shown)


infra/Pos/Network/Types.hs
```diff
@@ -116,12 +116,15 @@ showableNetworkConfig NetworkConfig {..} =
...
 -- because we re-read the file on SIGHUP.
-data StaticPeers = forall m. (MonadIO m, WithLogger m) => StaticPeers {
+data StaticPeers = StaticPeers {
       -- | Register a handler to be invoked whenever the static peers change
       --
       -- The handler will also be called on registration
       -- (with the current value).
-      staticPeersOnChange :: (Peers NodeId -> m ()) -> IO ()
+      staticPeersOnChange   :: (Peers NodeId -> LoggerNameBox IO ()) -> IO ()
+    , -- | Monitoring worker which is supposed to be started in a
```
Pastafarianist: Please provide more detail about what this worker does (as opposed to how to use it) and how it is different from `staticPeersOnChange`.

PR #2387, opened 2018-01-26
CSL-1606 Find better approach for long running actions passed no 'onNewSlot' 

gromakovsky:
`onNewSlot` worker now takes `ActionTerminationPolicy` as an argument.
    Parameters are wrapped in `OnNewSlotParams` type.
    `ActionTerminationPolicy` allows to cancel running action when new slot starts.
    This feature is used in two workers: block creator and `Update.processNewSlot`.
    The detailed analysis was performed here:
    https://github.com/input-output-hk/cardano-sl/pull/1163#issuecomment-318136810
    In particular it allowed us to avoid `fork` in block creator.
    Ideally we also want to separate actual block creation and network stuff which
    happens afterwards, but I think it's minor, so it hasn't been done.

### DIFF
--- a/block/src/Pos/Block/Worker.hs
+++ b/block/src/Pos/Block/Worker.hs
@@ -13,7 +13,7 @@ import           Control.Lens (ix)
 import qualified Data.List.NonEmpty as NE
 import           Data.Time.Units (Microsecond)
 import           Formatting (Format, bprint, build, fixed, int, now, sformat, shown, (%))
-import           Mockable (delay, fork)
+import           Mockable (delay)
 import           Serokell.Util (enumerate, listJson, pairF, sec)
 import qualified System.Metrics.Label as Label
 import           System.Random (randomRIO)
@@ -52,7 +52,8 @@ import           Pos.Recovery.Info (getSyncStatus, getSyncStatusK, needTriggerRe
                                     recoveryCommGuard)
 import           Pos.Reporting (MetricMonitor (..), MetricMonitorState, noReportMonitor,
                                 recordValue, reportOrLogE)
-import           Pos.Slotting (currentTimeSlotting, getSlotStartEmpatically)
+import           Pos.Slotting (ActionTerminationPolicy (..), OnNewSlotParams (..),
+                               currentTimeSlotting, defaultOnNewSlotParams, getSlotStartEmpatically)
 import           Pos.Update.DB (getAdoptedBVData)
 import           Pos.Util (mconcatPair)
 import           Pos.Util.Chrono (OldestFirst (..))
@@ -81,7 +82,7 @@ blkWorkers keepAliveTimer =
 
 informerWorker :: BlockWorkMode ctx m => (WorkerSpec m, OutSpecs)
 informerWorker =
-    onNewSlotWorker True announceBlockOuts $ \slotId _ ->
+    onNewSlotWorker defaultOnNewSlotParams announceBlockOuts $ \slotId _ ->
         recoveryCommGuard "onNewSlot worker, informerWorker" $ do
             tipHeader <- DB.getTipHeader
             -- Printe tip header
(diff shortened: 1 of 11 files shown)


avieth:
I wonder if starting a new thread at each slot is not the best way to do this. How about having one thread, and a way to block until the next slot is available, for instance using a `TVar`:

```Haskell
type CurrentSlot = TVar SlotId

syncWithSlots :: CurrentSlot -> SlotId -> (SlotId -> IO ()) -> IO ()
syncWithSlots currentSlot wakeSlot k = do
    slot <- blockUntil currentSlot wakeSlot
    k slot
    syncWithSlots currentSlot (nextSlot slot) k

blockUntil :: CurrentSlot -> SlotId -> IO SlotId
blockUntil var slot = atomically $ do
    slot' <- readTVar var
    unless (slot' >= slot) retry
```

And there would be one slotting thread which updates that `TVar` at the appropriate time.

Should an action *ever* be killed because it took more than the slot duration? Wouldn't it be better to let it run its course, and then:

- If it finishes before the next slot is over, it can continue normally. It didn't miss a slot, it just didn't finish *within* its slot.
- If it finishes after the next slot is over (its slot, and the slot after it, completed before it finished) then it can decide what to do (the particular action).

gromakovsky:
> Should an action ever be killed because it took more than the slot duration?

Yes, there are two cases where I find it reasonable. Not necessary more than slot duration, but some timeout makes sense. Initially I had another constructor for `ActionTerminationPolicy` which had explicit timeout, but then I decided that waiting until next slot is fine in both cases.
These are the cases where `NewSlotTerminationPolicy` is used.
1. The first one is `Update.processNewSlotWorker`. It updates some `TVar` which stores `SlotId` to the new `SlotId`. When we create a block, we want this `TVar` to have current `SlotId`. If slot `s` finishes, it's useless to wait until it updates `TVar` for slot `s`, it can switch to `s + 1` immediately when `s + 1` starts.
2. The second one is block creation worker. It creates a block and then announces it to the network. If we don't manage to finish this activity for slot `s` before slot `s` finishes, most likely our block will be useless anyway, because we don't accept blocks which arrive after slot for which they were created. So I think such timeout makes sense in this case.

> If it finishes before the next slot is over, it can continue normally. It didn't miss a slot, it just didn't finish within its slot.
> If it finishes after the next slot is over (its slot, and the slot after it, completed before it finished) then it can decide what to do (the particular action).

This is about timeout which should be used. As I said, my approach can be extended by adding another constructor for `ActionTerminationPolicy` (which will wait for one more slot). However, in cases which I mentioned above we do want until the end of the current slot, not more.
Regarding `can decide what to do`: currently this decision is made by users of `onNewSlot` and there are two options (corresponding to constructors of `ActionTerminationPolicy`). Action itself can also decide what to do, of course (it can apply `timeout` internally), but doing it inside `onNewSlot` allows to slightly decrease code duplication (we calculate time till the end of slot and apply `timeout` only in that function, not in all actions which want such logic).

Regarding your code snippet: this `TVar` approach looks like another way to implement `onNewSlot` function, but it's not obvious which benefits it has comparing to existing implementation.

block/src/Pos/Block/Worker.hs
```diff
@@ -81,7 +83,7 @@ blkWorkers keepAliveTimer =
 
 informerWorker :: BlockWorkMode ctx m => (WorkerSpec m, OutSpecs)
 informerWorker =
-    onNewSlotWorker True announceBlockOuts $ \slotId _ ->
+    onNewSlotWorker def announceBlockOuts $ \slotId _ ->
```
adinapoli-iohk: This is just a little fetish of mine (which I know it's debatable) but I largely prefer the use of an explicit top-level binding which alias this `def`, but it contextualises it for the task at hand. By that I mean writing this, essentially:

```haskell
defaultOnNewSlotParams :: OnNewSlotParams
defaultOnNewSlotParams = ...
```

The idea is that I can now `grep` for `defaultOnNewSlowParams`, but the same cannot be said for `def`. Granted, I can grep for `instance Default OnNewSlotParams`, but it generally adds more overhead 😁 

Nothing to stop the PR from being merged, though :)

infra/Pos/Slotting/Util.hs
```diff
@@ -93,41 +97,85 @@ getNextEpochSlotDuration =
...
+-- 2. Thread management will be quite complicated if we want to fork
+-- threads inside `onNewSlot`.
+-- 3. If more than one action is launched, they may use same resources
+-- concurrently, so the code must account for it.
+data ActionTerminationPolicy
+    = NoTerminationPolicy
+    -- ^ Even if action keeps running after current slot finishes,
+    -- we'll just wait and start action again only after the previous
+    -- one finishes.
+    | NewSlotTerminationPolicy !Text
```
adinapoli-iohk: Would passing a `LoggerName` be overkill or not what we want here?

gromakovsky: It's not what we want here, because this `Text` value is not used as a logger name, it's used as a part of the message which is logged (see below).
```
+                  logWarning $ sformat
+                  ("Action "%stext%
+                   " hasn't finished before new slot started") name
```

block/src/Pos/Block/Worker.hs
gromakovsky: I tend to agree, I am not a big fan of `def` either, but for some reason I decided to use it here. I won't be lazy and will change it :)

PR #2388, opened 2018-01-26
[CSL-2208] Add fake utxo to dbgen

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2208

Tested on `mainnet`. Used:
```
stack exec dbgen -- --config ./tools/src/dbgen/config.dhall --nodeDB db-mainnet --walletDB wdb-mainnet --configPath node/configuration.yaml --secretKey secret-mainnet.key --configProf mainnet_full --add-to Ae2tdPwUPEZJHA8wEbVWoT4zgDGuWkXT9vLW6RzLvMt8kYCefkBQ1nixzpX@2147483648 --genFakeUtxo
```

The result(just the last bit, output is too large):
```
        {
          "cadId": "DdzFFzCqrht9pDX2vt4sZe3NxdDBtMN8QL2sKbAG9ztAgZ7iNACCr8FBDPbRv8kAL6hPSCMbRCheWnHxjdqVbWd9CvBHjm29ubGRbiYd",
          "cadAmount": {
            "getCCoin": "10000"
          },
          "cadIsUsed": false,
          "cadIsChange": false
        },
        {
          "cadId": "DdzFFzCqrhsyWHF8kPwR7rRENnH7iifEJkLu6zx4MF5vZVoCJnHpdwuaa9w4rCotxcX5aLgRH47Z6pN4UoWVCzcqTbiAixmPDuTancAs",
          "cadAmount": {
            "getCCoin": "10000"
          },
          "cadIsUsed": false,
          "cadIsChange": false
        },
        {
          "cadId": "DdzFFzCqrht32GLbiAtKEuaTcxFrRCaUj6kAXG2fQaEb2yQURgxDq3iTgQS5gr2KcEgBUifYCo7WxZmL5BUgRtvULHu7dNXsdv2PhhyQ",
          "cadAmount": {
            "getCCoin": "10000"
          },
          "cadIsUsed": false,
          "cadIsChange": false
        }
      ],
      "caAmount": {
        "getCCoin": "1000000"
      }
    }
  ]
}

```

### DIFF
--- a/tools/src/dbgen/CLI.hs
+++ b/tools/src/dbgen/CLI.hs
@@ -30,12 +30,11 @@ data CLI = CLI
     , showStats         :: Bool
     -- ^ If true, print the stats for the `wallet-db`
     , queryMethod       :: Maybe Method
-    -- ^ If true, generate a DB targeting mainnet.
+    -- ^ If a method is defined, try to call it. For example,
+    -- calling `GetWallet` could be useful.
     , addTo             :: Maybe AccountId
     -- ^ If specified, only append addresses to the
     -- given <wallet_id@account_id>
-    , genFakeUtxo       :: Bool
-    -- ^ If true, generate fake UTXO.
     }
 
 
@@ -67,7 +66,6 @@ instance ParseRecord CLI where
                             (long "add-to" <> metavar "walletId@accountId"
                                            <> help "Append to an existing wallet & account."
               )))
-              <*> switch (long "genFakeUtxo" <> help "Generate fake UTXO for the wallet. Fake as-in doesn't exist in node DB.")
 
 
 readAccountId :: String -> Either String AccountId
--- a/tools/src/dbgen/CLI.hs
+++ b/tools/src/dbgen/CLI.hs
@@ -10,7 +10,7 @@ import           Pos.Util.Servant
 import           Pos.Wallet.Web.ClientTypes.Instances ()
 import           Pos.Wallet.Web.ClientTypes.Types
 import           Text.Read                            (readMaybe)
-import           Types                                (Method)
+import           Types                                (Method (..))
 
 data CLI = CLI
     { config            :: FilePath
--- a/tools/src/dbgen/CLI.hs
+++ b/tools/src/dbgen/CLI.hs
@@ -21,9 +21,6 @@ data CLI = CLI
     -- ^ The path to the secret key from the database.
     , walletPath        :: FilePath
     -- ^ The path to a valid acid-state database.
-    , addTo             :: Maybe AccountId
-    -- ^ If specified, only append addresses to the
-    -- given <wallet_id@account_id>
     , configurationPath :: FilePath
     -- ^ The path to a valid cardano-sl configuration.
     , configurationProf :: String
@@ -34,7 +31,11 @@ data CLI = CLI
     -- ^ If true, print the stats for the `wallet-db`
     , queryMethod       :: Maybe Method
     -- ^ If true, generate a DB targeting mainnet.
+    , addTo             :: Maybe AccountId
+    -- ^ If specified, only append addresses to the
+    -- given <wallet_id@account_id>
     , genFakeUtxo       :: Bool
+    -- ^ If true, generate fake UTXO.
     }
 
 
@@ -52,10 +53,6 @@ instance ParseRecord CLI where
               <*> (strOption (long "walletDB" <> metavar "acidstate-path"
                              <> help "A path to a valid acidstate database."
                                       ))
-              <*> (optional (option (eitherReader readAccountId)
-                            (long "add-to" <> metavar "walletId@accountId"
-                                           <> help "Append to an existing wallet & account."
-              )))
               <*> (strOption (long "configPath" <> metavar "configuration-path"
                              <> help "A path to a valid cardano-sl configuration."
                                       ))
@@ -66,6 +63,10 @@ instance ParseRecord CLI where
                              <> help "A valid node system start to use."))
               <*> switch (long "stats" <> help "Show stats for this wallet.")
               <*> ((readMaybe =<<) <$> (optional (strOption (long "query" <> help "Query a predefined endpoint."))))
+              <*> (optional (option (eitherReader readAccountId)
+                            (long "add-to" <> metavar "walletId@accountId"
+                                           <> help "Append to an existing wallet & account."
+              )))
               <*> switch (long "genFakeUtxo" <> help "Generate fake UTXO for the wallet. Fake as-in doesn't exist in node DB.")
 
 
--- a/tools/src/dbgen/Lib.hs
+++ b/tools/src/dbgen/Lib.hs
@@ -17,7 +17,8 @@ import qualified Data.ByteString                      as B
 import           Data.String.Conv
 import           Data.Time
 import           Data.Map                             (union, fromList)
-import           Dhall                                (Interpret, input, auto)
+--import           Dhall                                (Interpret, input, auto)
+import           Data.Aeson                           (FromJSON, ToJSON, eitherDecodeStrict)
 import           GHC.Generics                         (Generic)
 import           Network.Haskoin.Crypto               (Entropy, toMnemonic)
 import           Pos.Crypto.Random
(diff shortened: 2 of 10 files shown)


tools/src/dbgen/CLI.hs
```diff
@@ -34,7 +31,11 @@ data CLI = CLI
     -- ^ If true, print the stats for the `wallet-db`
     , queryMethod       :: Maybe Method
     -- ^ If true, generate a DB targeting mainnet.
+    , addTo             :: Maybe AccountId
+    -- ^ If specified, only append addresses to the
+    -- given <wallet_id@account_id>
     , genFakeUtxo       :: Bool
+    -- ^ If true, generate fake UTXO.
```
adinapoli-iohk: I would prefer we extend the Dhall config file we have, rather than adding another CLI command. What about something like, for example:

```
{ wallet_spec = { account_spec = { addresses = 1 }, accounts = 1, balance = 10000 }
, wallets     = 1
}
```

This would make sure that the total Utxo for this wallet would be 10000, randomly divided into the addresses is being asked to generate.

tools/src/dbgen/CLI.hs
```diff
@@ -34,7 +31,11 @@ data CLI = CLI
     -- ^ If true, print the stats for the `wallet-db`
     , queryMethod       :: Maybe Method
     -- ^ If true, generate a DB targeting mainnet.
```
adinapoli-iohk: This comment seems to be incorrectly placed (it's probably me to blame here, but as we are reviewing the code let's change it 😉 )

tools/src/dbgen/Lib.hs
```diff
@@ -30,38 +35,55 @@ import           Pos.Wallet.Web.ClientTypes
...
 --
 
-data GenSpec = GenSpec {
-    wallets       :: !Integer
+data GenSpec = GenSpec
+    { wallets     :: !Integer
     -- ^ How many wallets to create
     , wallet_spec :: WalletSpec
     -- ^ The specification for each wallet.
+    , fakeUtxo    :: FakeUtxoSpec
```
adinapoli-iohk: It probably makes more sense to make the FakeUtxoSpec nested into a WalletSpec. This will match my suggestion above (of the Dhall config changes).

tools/src/dbgen/Lib.hs
```diff
@@ -30,38 +35,55 @@ import           Pos.Wallet.Web.ClientTypes
...
+-- TODO(ks): The question here is whether we need to support some other
+-- strategies for distributing money, like maybe using a fixed amount
+-- of `toAddress` to cap the distribution - we have examples where we
+-- have around 80 000 addresses which is a lot and can be intensive?
+-- For now, KISS.
+data FakeUtxoSpec = FakeUtxoSpec
+    {  fromAddress  :: !Integer
+    -- ^ How many addresses to contain fake ADA.
+    ,  amount       :: !Integer
+    -- ^ How much ADA do we want to distribute in a single address.
```
adinapoli-iohk: Agree, excellent observation. For now let's add uniform stake distribution, later we can change the dhall config to specify a distribution policy (per wallet, that is).

tools/src/dbgen/config.dhall
```diff
@@ -1,3 +1,4 @@
 { wallet_spec = { account_spec = { addresses = 100 }, accounts = 1 }
+, fakeUtxo    = { fromAddress = 100, amount = 10000 }
```
adinapoli-iohk: Interestingly you took a similar path of the one I am proposing here, that's good, it means we are not too far off to a good solution 😉

tools/src/dbgen/README.md
```diff
@@ -9,37 +9,18 @@ fail with the "wrong magic" error.
...
-dbgen --dbPath ../cardano-sl/run/node-db0
-Faking StateLock syncing...
-Generating 1 wallets...
-Action took 0.13194 seconds.
-Generating 1 accounts for Wallet 1...
-Action took 0.083039 seconds.
-Generating 10 addresses for Account CAccountId "Ae2tdPwUPEZGwUVR8meJ6mgbHeiC16TNyM3wTDYRYwDBMeCHUXdMVLHKupC@793406314"...
-Action took 0.739177 seconds.
-OK.
+stack exec dbgen -- --config ./tools/src/dbgen/config.dhall --nodeDB db-mainnet --walletDB wdb-mainnet --configPath node/configuration.yaml --secretKey secret-mainnet.key --configProf mainnet_full
```
adinapoli-iohk: Thanks for modifying the README, that's always a good practice. 😉

tools/src/dbgen/README.md
```diff
@@ -9,37 +9,18 @@ fail with the "wrong magic" error.
...
+, fakeUtxo    = { fromAddress = 100, amount = 10000 }
+, wallets     = 1
 }
 ```
 
-Once you are ready, you can invoke `dbgen`:
-
+You should run the program using something like (from say `cardano-sl` root):
 ```
-Usage: dbgen [--config CONFIG.DHALL] [--dbPath rocksdb-path]
```
adinapoli-iohk: My 2 cents, but I still think that's nice to show in the README the fact you can also use the tool as a poor man's benchmarking tool.

tools/src/dbgen/Lib.hs
```diff
@@ -30,38 +35,57 @@ import           Pos.Wallet.Web.ClientTypes
...
 
 instance Interpret AccountSpec
 
+-- TODO(ks): The question here is whether we need to support some other
+-- strategies for distributing money, like maybe using a fixed amount
+-- of `toAddress` to cap the distribution - we have examples where we
+-- have around 80 000 addresses which is a lot and can be intensive?
+-- For now, KISS.
+data FakeUtxoSpec = FakeUtxoSpec
+    {  fromAddress  :: Maybe Integer
```
adinapoli-iohk: I find `fromAddress` not very much self-explanatory, and it's hard to evince from the Dhall config. Furthermore, it feels like there is a bit of duplication in the utxo spec, as the overall spec already specify how many addresses to create, we just need to specify how many fill with fake utxo, and with which balance. I propose something like this:

``` haskell
data FakeUtxoSpec = FakeUtxoSpec {
  amount :: !Integer
  -- ^ If set to zero, this effectively corresponds to "do not generate fake utxo".
  , distribution :: CoinDistributionSpec
}

data CoinDistributionSpec = 
    UniformDistribution
    -- ^ All the addresses gets the same amount.
  | RangeDistrubution !Integer
  -- ^ Distributes to only XX addresses.
  | ...
  -- ^ For now we KISS, later we can add more type constructors

instance Interpreted CoinDistributionSpec where
  -- This will need to be written in order to map an object into a proper CoinDistrubitionSpec
```

This way, you can now specify a uniform utxo distribution this way:

```
{ wallet_spec = { account_spec = { addresses = 100 }, accounts = 1, fakeUtxo = { distribution = { type = "uniform" } , amount = 10000 } }
 , wallets     = 1
}
```

But you can now also do the equivalent of the `fromAddresses = [100]` like before by setting `distribution` to be, for example:

```
{ distribution = { type = "range"
                          , addresses = 100
                          }
}
```

It's now easy to add extra policies and we don't need any `Optional`: merely setting the `amount` to 0 or the `addresses` to 0 would entail "Do not distribute stake". What do you think?

ksaric: I tried to avoid that case because it's not as explicit as the `Optional` one - the user must know to set the number to `0` which doesn't make sense to me - imagine if you had to know how to set up a full configuration of a library you don't know.
So that's why my preference is/was:
1. Explicit argument sent to the application (high visibility, low chance of doing something wrong)
2. Explicit configuration parameter (medium visibility, medium chance of doing something wrong)
3. Set the parameter to specific value (low visibility, high chance of doing something wrong)

I guess the other option which would be ok for would be to have (`NoDistribution`):
```
data CoinDistributionSpec = 
    NoDistribution
    -- ^ Do not distribute the coins.
 |  UniformDistribution
    -- ^ All the addresses gets the same amount.
  | RangeDistrubution !Integer
  -- ^ Distributes to only XX addresses.
  | ...
  -- ^ For now we KISS, later we can add more type constructors
```

@adinapoli-iohk Is that ok? 
You can still do something bad with `addresses = 0` but at least you know that you explicitly asked for the generation of the fake utxo.

adinapoli-iohk: I see your point -- I don't think this in practice would be too much of a concern as this tool is mostly for internal use, and the README should cover the ambiguity. However, I like the `NoDistribution` idea, so I'm happy to settle on this middle ground 😉 

Generally speaking, I don't see necessarily setting some values to 0 as being an evil thing -- it's just another declination of "what you see is what you get", in my opinion. Setting the `amount` to 0 explicitly declares you don't want those addresses to have balance, which is fine. Same applies for `addresses = 0`: you are explicitly stating you want to distribute the stake to exactly 0 addresses. Again, that's your choice, and I'm not sure what `Optional` buys you in that case, but this is more design bikeshedding than anything 😉

ksaric:
Squash required :grin:

tools/src/dbgen/Lib.hs
```diff
@@ -30,52 +35,108 @@ import           Pos.Wallet.Web.ClientTypes
...
+-- | A simple example of how the configuration looks like.
+_testSpec :: GenSpec
+_testSpec = GenSpec
+    { walletSpec = WalletSpec
+        { accounts = 1
+        , accountSpec = AccountSpec { addresses = 100 }
+        , fakeUtxo = FakeUtxoSpec { amount = 1000, distribution = RangeDistribution { range=100 } }
+        }
+    , wallets = 1
+    }
```
adinapoli-iohk: Good idea, I like this! I would call it simply `_exampleSpec` as we are not really testing anything, but that's a minor nitpick.

tools/src/dbgen/Lib.hs
```diff
@@ -30,52 +35,108 @@ import           Pos.Wallet.Web.ClientTypes
...
+    deriving (Show, Eq, Generic)
+
+instance FromJSON CoinDistributionSpec
+instance ToJSON CoinDistributionSpec
+
+{-
+λ> encode NoDistribution
+"{\"tag\":\"NoDistribution\"}"
+λ> encode $ RangeDistribution { range=100 }
+"{\"tag\":\"RangeDistribution\",\"range\":100}"
```
adinapoli-iohk: I'm not totally sold on the idea of relying on whichever format the `Generic` deriving dictates. I think writing a `FromJSON` and `ToJSON` manual instances is so quick and simple we should go for it. It will allow us to have a proper-looking `config` file, something like:

```
{ "type" : "range"
, "addresses": 100 
}
```

As per my original proposal. 

I'm sad to see Dhall going away, but I can see how avoiding the extra dep and a foreign config format might be a quick win for the scope of the tool.

tools/src/dbgen/Lib.hs
```diff
@@ -94,25 +155,66 @@ fakeSync = do
...
     -- but append the requested addresses to the input
     -- CAccountId.
+
+    -- TODO(ks): Simplify this?
+    let fakeUtxoSpec = fakeUtxo walletSpec
     case addTo of
-        Just accId -> addAddressesTo accId spec
+        Just accId -> case (distribution fakeUtxoSpec) of
+            NoDistribution -> addAddressesTo accId spec
+            _              -> generateFakeUtxo fakeUtxoSpec accId
```
adinapoli-iohk: Good, much better now.

ksaric:
BTW, I can't merge this. @dcoutts ? Can you merge this?

gromakovsky:
I can merge, but I think our policy is that PRs must be approved to @dcoutts before they can be merged to master or release branches.

denisshevchenko:
So, when will this PR be merged? We need it for Bittrex environment emulation. @gromakovsky @dcoutts

gromakovsky:
Well, since this PR affects only `dbgen` and doesn't make any changes to the code which runs in production, I think it's not strictly necessary to wait for Duncan's approval, so I am merging it.

PR #2393, opened 2018-01-26
[CSL-2122] Allow to specify def tx inputs selection policy in config

Martoon-00:
Rebased #2182


Martoon-00:
Not required for now.
I'll remove branch in a month unless need arises again.

PR #2394, opened 2018-01-26
Explicit DB threading in the wallet

nc6:
This is a resurrection of #2360 using a branch hosted in IOHK github.

### DIFF
--- a/node/src/Pos/Wallet/Web/Util.hs
+++ b/node/src/Pos/Wallet/Web/Util.hs
@@ -33,53 +33,54 @@ import           Pos.Wallet.Web.ClientTypes (AccountId (..), Addr, CId,
 
 import           Pos.Wallet.Web.Error       (WalletError (..))
 import           Pos.Wallet.Web.State       (AddressLookupMode(..), AddressInfo (..),
-                                             CurrentAndRemoved (..),
-                                             WebWalletModeDB, getAccountAddrMaps, getAccountIds,
+                                             CurrentAndRemoved (getCurrent, getRemoved),
+                                             WalletSnapshot,
+                                             getAccountAddrMaps, getAccountIds,
                                              getAccountWAddresses, getWalletMeta,
                                              getAccountMeta)
 
-getAccountMetaOrThrow :: (WebWalletModeDB ctx m, MonadThrow m) => AccountId -> m CAccountMeta
-getAccountMetaOrThrow accId = getAccountMeta accId >>= maybeThrow noAccount
+getAccountMetaOrThrow :: MonadThrow m => WalletSnapshot -> AccountId -> m CAccountMeta
+getAccountMetaOrThrow ws accId = maybeThrow noAccount (getAccountMeta ws accId)
   where
     noAccount =
         RequestError $ sformat ("No account with id "%build%" found") accId
 
-getWalletAccountIds :: WebWalletModeDB ctx m => CId Wal -> m [AccountId]
-getWalletAccountIds cWalId = filter ((== cWalId) . aiWId) <$> getAccountIds
+getWalletAccountIds :: WalletSnapshot -> CId Wal -> [AccountId]
+getWalletAccountIds ws cWalId = filter ((== cWalId) . aiWId) (getAccountIds ws)
 
-getAccountAddrsOrThrow
-    :: (WebWalletModeDB ctx m, MonadThrow m)
-    => AddressLookupMode -> AccountId -> m [AddressInfo]
-getAccountAddrsOrThrow mode accId =
-    getAccountWAddresses mode accId >>= maybeThrow noWallet
+getAccountAddrsOrThrow :: MonadThrow m
+                       => WalletSnapshot
+                       -> AddressLookupMode
+                       -> AccountId
+                       -> m [AddressInfo]
+getAccountAddrsOrThrow ws mode accId = maybeThrow noWallet (getAccountWAddresses ws mode accId)
   where
     noWallet =
         RequestError $
         sformat ("No account with id "%build%" found") accId
 
 getWalletAddrMetas
-    :: (WebWalletModeDB ctx m, MonadThrow m)
-    => AddressLookupMode -> CId Wal -> m [CWAddressMeta]
-getWalletAddrMetas lookupMode cWalId = do
-    accountIds <- getWalletAccountIds cWalId
-    map adiCWAddressMeta <$> concatMapM (getAccountAddrsOrThrow lookupMode) accountIds
+    :: MonadThrow m
+    => WalletSnapshot
+    -> AddressLookupMode
+    -> CId Wal
+    -> m [CWAddressMeta]
+getWalletAddrMetas ws lookupMode cWalId = do
+    let accountIds = getWalletAccountIds ws cWalId
+    map adiCWAddressMeta <$> concatMapM (getAccountAddrsOrThrow ws lookupMode) accountIds
 
-getWalletAddrs
-    :: (WebWalletModeDB ctx m, MonadThrow m)
-    => AddressLookupMode -> CId Wal -> m [CId Addr]
-getWalletAddrs mode wid = cwamId <<$>> getWalletAddrMetas mode wid
+getWalletAddrs :: MonadThrow m => WalletSnapshot -> AddressLookupMode -> CId Wal -> m [CId Addr]
+getWalletAddrs ws mode wid = fmap cwamId <$> getWalletAddrMetas ws mode wid
 
-getWalletAddrsDetector
-    :: (WebWalletModeDB ctx m, MonadThrow m)
-    => AddressLookupMode -> CId Wal -> m (CId Addr -> Bool)
-getWalletAddrsDetector lookupMode cWalId = do
-    accIds <- getWalletAccountIds cWalId
-    accAddrMaps <- mapM getAccountAddrMaps accIds
+getWalletAddrsDetector :: WalletSnapshot -> AddressLookupMode -> CId Wal -> (CId Addr -> Bool)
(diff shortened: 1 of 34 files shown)


node/src/Pos/Wallet/Web/State/Transactions.hs
```diff
@@ -0,0 +1,104 @@
...
+  for_ wAddrs WS.removeWAddress
+  for_ custAddrs $ \(cat, addrs) ->
+    for_ addrs $ WS.removeCustomAddress cat
+  WS.updateWalletBalancesAndUtxo utxoMod
+  WS.removeFromHistoryCache walId historyEntries
+  WS.removeWalletTxMetas walId (encodeCType <$> M.keys historyEntries)
+  for_ ptxConditions $ \(txId, cond, meta) -> do
+    WS.ptxUpdateMeta walId txId meta
+    WS.setPtxCondition walId txId cond
+  WS.setWalletSyncTip walId syncTip
```
gromakovsky: Please reformat this module to use 4-spaces indentation (as it's done in all other code in `cardano-sl`) 🙏

adinapoli-iohk: Haha stylish-haskell police to the rescue! 😁  Don't worry @gromakovsky , the agreement with Nic is that once we are confident this PR delivers what it should, he will rebase everything properly (by adding, for example, the `[CSL-xxxx]` prefix to the commits) as well as fixing stylish issues 😁

nc6: Yeah, I'm aware of the need to do this, but 4-space indentation makes me miserable so I'm putting it off :-P

wallet/src/Pos/Wallet/Web/Methods/Backup.hs
```diff
@@ -45,11 +46,17 @@ restoreWalletFromBackup WalletBackup {..} = do
...
+            -- the bacup is large, there may be a large number of accounts, so forcing atomicity
+            -- may not be what we want.
+            db <- askWalletDB
             for_ accList $ \(idx, meta) -> do
                 let aIdx = fromInteger $ fromIntegral idx
                     seedGen = DeterminedSeed aIdx
-                accId <- genUniqueAccountId seedGen wId
-                createAccount accId meta
+                accId <- genUniqueAccountId ws seedGen wId
+                createAccount db accId meta
```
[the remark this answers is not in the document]
nc6: Oh, good catch. Will do.

adinapoli-iohk:
@nc6 Looking good now! Importing a backup now works:

screen shot 2018-01-31 at 13 36 54

So does creating a new wallet:

screen shot 2018-01-31 at 13 38 17

Finally, also payments seems to be on point:

screen shot 2018-01-31 at 13 38 55

nc6:
One thing we need to consider is that whilst only allowing reading a snapshot works well when we're doing multiple successive reads, it could potentially be a performance overhead where we're having to read it multiple times due to DB updates. Moreover, whilst sometimes we are doing e.g.

```
createAccount ix
foo <- getAccount ix
```
this is not universally the case - often we do an update and then fetch another part of the DB state, which nonetheless may have been affected by the update.

I think there are a few things we should do here:
- Try to eliminate instances of this pattern where possible - e.g. can we generate unique IDs within the context of 'create' transactions, which would be safer anyway.
- Where we do have cases of `createX >> getX`, modify the create transaction to return that being created. I'm slightly opposed to doing this in general, since it would entail doing a bunch of pointless reads in the various places where the results are not used.
- Where this isn't possible, selectively reintroduce specific read methods to avoid re-reading the entire snapshot. Note that we can still keep a certain amount of explicitness here by making these take explicit `WalletDB` parameters, as with `getWalletSnapshot`.

Mentioning this here to keep track of it, though I don't think we should handle in this PR, which is large enough already.
cc @dcoutts

adinapoli-iohk:
> I'm slightly opposed to doing this in general, since it would entail doing a bunch of pointless reads in the various places where the results are not used.

@nc6 Stupid idea: what about a middle ground where we return the new `WalletSnapshot` _without_ re-reading it from disk but simply by calling `get` at the very end of an `Update`? Being `Update` just a constrained monad over `MonadState`, it will of course not be perfect but still better than doing I/O access.

node/src/Pos/Wallet/Web/State/Transactions.hs
nc6: @gromakovsky Speaking of code style, do you have the appropriate `stylish-haskell.yaml` file for master?

nc6:
@adinapoli-iohk Sounds good, actually. I hadn't spotted that acid-state worked that way.

node/src/Pos/Wallet/Web/State/Transactions.hs
gromakovsky: See [this thread](https://input-output-rnd.slack.com/archives/C2VJ41WDP/p1515686971000339) about `stylish-haskell` and `master`.

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -166,22 +169,23 @@ getHistoryLimited mCWalId mAccId mAddrId mSkip mLimit = do
         "Please do not specify both walletId and accountId at the same time"
 
 addHistoryTx
-    :: MonadWalletWebMode m
-    => CId Wal
+    :: (WalletDbWriter A.AddOnlyNewTxMetas m, MonadIO m)
```
dcoutts: Are you sure `WalletDbWriter` is needed? We're not abstracting over the event here.

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -423,19 +424,13 @@ selectOwnTxInsAndOuts encInfo (WithHash tx txId, NE.toList -> undoL) (mDiff, mTs
     (ownInputs, map (first toTxInOut) ownOutputs, th)
 
 applyModifierToWallet
-    :: ( WalletDbReader ctx m
-       , WalletDbWriter A.UpdateWalletBalancesAndUtxo m
-       , HasConfiguration
-       )
-    => CId Wal
+    :: WalletDbWriter A.ApplyModifierToWallet m
+    => WalletDB
```
dcoutts: 👍

dcoutts:
I have a suggestion if we want to try and get this down to a manageable size (bearing in mind we'll eventually need to port these changes to the develop branch): for this PR omit the changes to the transactions (merging many updates into one tx, rearranging getHistory & newPayment. Get it down to basically just the explicit passing of the db, and type sig simplifications, and squash together the patches where we go back and forth and the fixups. So we do that, and we do the audit I mentioned above. Then we do those other improvements in one or more follow-on PRs. Just a thought. Lets discuss tomorrow.

dcoutts:
> I'm slightly opposed to doing this in general, since it would entail doing a bunch of pointless reads in the various places where the results are not used.

It's worth noting that returning results in acid-state transactions or queries is trivially cheap. Reading the entire snapshot is cheap. It's just a read of an in-mem value from a variable. The only cost is the synchronization (concurrent reader/writer stuff).

node/src/Pos/Wallet/Web/Util.hs
```diff
@@ -60,17 +60,15 @@ getAccountAddrsOrThrow ws mode accId = maybeThrow noWallet (getAccountWAddresses
...
+    :: WalletSnapshot
     -> AddressLookupMode
     -> CId Wal
-    -> m [CWAddressMeta]
-getWalletAddrMetas ws lookupMode cWalId = do
-    let accountIds = getWalletAccountIds ws cWalId
-    map adiCWAddressMeta <$> concatMapM (getAccountAddrsOrThrow ws lookupMode) accountIds
+    -> [CWAddressMeta]
+getWalletAddrMetas ws lookupMode cWalId =
+  map adiCWAddressMeta $ getWAddresses ws lookupMode cWalId
```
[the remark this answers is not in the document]
nc6: It was not possible for such a thing to happen (unless by virtue of the non-transactionality). It fetched the keys from a map, then tried to look them up in the same map. Rewrite was deliberately to remove this false notion of partiality.

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -49,7 +49,7 @@ getFullWalletHistory :: MonadWalletWebMode m
 getFullWalletHistory ws cWalId = do
     logDebug "getFullWalletHistory: start"
 
-    cAddrs <- getWalletAddrs ws Ever cWalId
+    let cAddrs = getWalletAddrs ws Ever cWalId
```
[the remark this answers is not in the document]
nc6: See above comment.

wallet/src/Pos/Wallet/Web/Methods/Logic.hs
```diff
@@ -139,7 +138,8 @@ getAccountMod ws accMod accId = do
 getAccount :: MonadWalletWebMode m => AccountId -> m CAccount
 getAccount accId = do
     ws <- getWalletSnapshot
-    fixingCachedAccModifier ws (getAccountMod ws) accId
+    accMod <- txMempoolToModifier ws =<< findKey accId
+    getAccountMod ws accMod accId
```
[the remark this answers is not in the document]
nc6: Addressed after rebase. I don't like to rebase things when being reviewed :-)

adinapoli-iohk:
@nc6 These are the tests I have conducted:

- [X] Importing a wallet works
- [X] Creating a new wallet works
- [X] Sending Txs between two wallet works
- [X] Generating a new Cardano address works
- [X] Deleting a wallet works
- [X] Restoring a wallet works

Looking good! 🎉  cc @dcoutts

PR #2396, opened 2018-01-26
[CSL-2167] Additional logging

gromakovsky:
This logging is needed because in some cases we get
> cardano-node: signalProcess: permission denied (Operation not permitted)

error shortly after start.

### DIFF
--- a/lib/src/Pos/Client/CLI/Util.hs
+++ b/lib/src/Pos/Client/CLI/Util.hs
@@ -2,6 +2,7 @@
 
 module Pos.Client.CLI.Util
        ( printFlags
+       , printInfoOnStart
        , attackTypeParser
        , attackTargetParser
        , defaultLoggerConfig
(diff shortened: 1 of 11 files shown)


lib/src/Pos/Client/CLI/Util.hs
```diff
@@ -14,24 +14,38 @@ import           Universum
 
 import qualified Data.ByteString.Lazy as BSL
 import           Formatting (sformat, shown, (%))
+import           Mockable (CurrentTime, Mockable, currentTime)
 import           System.Wlog (LoggerConfig (..), WithLogger, logInfo, parseLoggerConfig,
                               productionB)
-import           Text.Parsec (try, parserFail)
+import           Text.Parsec (parserFail, try)
```
Pastafarianist: @int-index, y u no `stylish-haskell`?

util/Pos/Util/Util.hs
```diff
@@ -287,6 +281,28 @@ multilineBounds maxSize = F.later formatList
...
+-- | 'bracket' which logs given message after acquiring the resource
+-- and before calling the callback with 'Info' severity.
+bracketWithLogging ::
+       (MonadMask m, WithLogger m)
+    => Text
+    -> m a
+    -> (a -> m b)
+    -> (a -> m c)
+    -> m c
+bracketWithLogging msg acquire release = bracket acquire release . addLogging
```
Pastafarianist: Maybe also log something before (or after) the resource is released?

gromakovsky: It's not essential for CSL-2167, but in general it makes sense. Will do.

gromakovsky:
> Might be too verbose. Realistically, we probably won't need that much logging on context & resources initialization. On the other hand, it's not quite clear which logs can be left out and which ones are more useful. LGTM, but if you have an extra hour, please think about how to make initialization logging less verbose.

Well, yes, it's verbose, but error from CSL-2167 is absolutely strange and I have no idea where it happens. I'd like to figure out the exact place as precisely as possible when we get logs from someone who has this problem.
And I think it won't be a problem to have this logging even after we resolve CSL-2167, because it's printed only once (on start), but the node is supposed to run for a long time, so it's not a big deal. And we may have other issues when something goes wrong during initialization.

util/Pos/Util/Util.hs
gromakovsky: Done.

PR #2397, opened 2018-01-27
[CSL-1918] Update launcher documentation

Pastafarianist:
There is one thing I am not sure about. Both the old documentation for CLI parameters and the newer `docs/launcher.md` state that not specifying `nodeLogPath` cause launcher to create a temporary file where node output is redirected. After reviewing the code I couldn't find anything resembling that. Instead, apparently the output simply gets passed through to `stdout`/`stderr`. See [here](https://github.com/input-output-hk/cardano-sl/blob/2d23da907fa49436dcdd5911ae712f8aee99bc86/tools/src/launcher/Main.hs#L555) for more details.

### DIFF
--- a/docs/launcher.md
+++ b/docs/launcher.md
@@ -18,8 +18,8 @@ and the client scenario. The server scenario is used when the launcher is used
 for the node alone (without wallet), and the client scenario is used when the
 launcher coordinates the node together with the wallet.
 
-The choice of scenario is determined by whether the optional CLI parameter
-`--wallet` was specified.
+The choice of scenario is determined by whether `walletPath` was specified
+in the config or not.
 
 ### Server scenario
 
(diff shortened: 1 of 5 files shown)


gromakovsky:
> There is one thing I am not sure about. Both the old documentation for CLI parameters and the newer docs/launcher.md state that not specifying nodeLogPath cause launcher to create a temporary file where node output is redirected. After reviewing the code I couldn't find anything resembling that. Instead, apparently the output simply gets passed through to stdout/stderr. 

@neongreen @Pastafarianist did you guys resolve it?
This behaviour was changed in #1629.

Pastafarianist:
Yes, @neongreen said that this change was introduced intentionally, but documentation wasn't updated.

PR #2401, opened 2018-01-28
[CSL-2150] Speed up 'getWallet'

pva701:
This PR slightly changes the way which we compute balances, instead of getting wallet's accounts and then summing, the following approach was implemented:
1. take wallet's accounts
2. union their addresses
3. sum coins on these addresses using balance map, which is already presented in `WalletStorage`

### DIFF
--- a/wallet/src/Pos/Wallet/Web/Methods/Logic.hs
+++ b/wallet/src/Pos/Wallet/Web/Methods/Logic.hs
@@ -48,13 +48,12 @@ import           Pos.Wallet.Web.Account     (AddrGenSeed, findKey, genUniqueAcco
                                              genUniqueAddress, getAddrIdx, getSKById)
 import           Pos.Wallet.Web.ClientTypes (AccountId (..), CAccount (..),
                                              CAccountInit (..), CAccountMeta (..),
-                                             CAddress (..), CCoin, CId, CWallet (..),
-                                             CWalletMeta (..), Wal, addressToCId,
-                                             encToCId, mkCCoin)
+                                             CAddress (..), CId, CWallet (..),
+                                             CWalletMeta (..), Wal, encToCId, mkCCoin)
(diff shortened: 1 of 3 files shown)


adinapoli-iohk:
Thanks for that, @pva701 -- Did you test these changes with a local cluster, did they work?

Generally speaking I would like to wait merging any performance work up until @nc6 's completes the first round of open-heart surgery on the `WalletStorage` 😉

nc6:
Biggest surgery in this area is done, so should be good to rebase and we can review.

Martoon-00:
Thanks for notification!

wallet/src/Pos/Wallet/Web/Methods/Logic.hs
```diff
@@ -164,14 +126,21 @@ getAccounts = getAccountsIncludeUnready False
 
 getWalletIncludeUnready :: MonadWalletWebMode m => Bool -> CId Wal -> m CWallet
 getWalletIncludeUnready includeUnready cAddr = do
```
ksaric: Maybe create a new datatype for `includeUnready` and replace `Bool` in order to make it more (type)safe and understandable?

```
data WalletInclusionState = WalletReady | WalletNotReady
```

ksaric: `cAddr = cWalId` ? Not sure what the convention is, but `WalletId` is more clear than `Addr`.

Martoon-00: Ok, remaned

Martoon-00: I agree, but this code was written many months ago and it's out of scope of this PR.
Although would be nice to change it in some way when someone has extra time

PR #2404, opened 2018-01-29
[CSL-2174] Eliminate `forkIO` usages 

gromakovsky:
This branch contains implementation of issues from CSL-2174 grand task.
All issues from this task will be in PRs targeting this branch.


networking/src/Mockable/Concurrent.hs
```diff
@@ -227,6 +231,57 @@ forConcurrently
...
+----------------------------------------------------------------------------
+-- Low-level Async mock and helper functions
+----------------------------------------------------------------------------
+
+data LowLevelAsync m t where
+    Async :: m t -> LowLevelAsync m (Promise m t)
+    Link :: Promise m t -> LowLevelAsync m ()
+    Wait :: Promise m t -> LowLevelAsync m t
+    WaitAny :: [Promise m t] -> LowLevelAsync m (Promise m t, t)
+    CancelWith :: Exception e => Promise m t -> e -> LowLevelAsync m ()
```
avieth: Why `LowLevelAsync`? Was having these in `Async` considered dangerous? If so, why?

gromakovsky: This is one of recommendations from FP-Complete. See https://docs.google.com/document/d/1ueuoII0UIkHbj_8eozJ_3InkLTRc0wTSBvu7g6RcrOk/edit#heading=h.ar36nz4qljge.

Usually these functions are not needed, higher-level abstractions are enough. We don't want to completely prohibit usage of these functions, but we want to make it clearer in which places they are used.

infra/Pos/Slotting/Util.hs
```diff
@@ -137,40 +185,34 @@ onNewSlotImpl withLogging startImmediately action =
...
+        onNewSlotImpl withLogging params action
 
 onNewSlotDo
-    :: OnNewSlot ctx m
-    => Bool -> Maybe SlotId -> Bool -> (SlotId -> m ()) -> m ()
-onNewSlotDo withLogging expectedSlotId startImmediately action = do
+    :: MonadOnNewSlot ctx m
+    => Bool -> Maybe SlotId -> OnNewSlotParams -> (SlotId -> m ()) -> m ()
+onNewSlotDo withLogging expectedSlotId onsp action = do
     curSlot <- waitUntilExpectedSlot
```
volhovm: Wait, so we will wait until current slot even if `onspStartImmediately onsp == True`? It seems that this call will block until new slot comes. Am I wrong?

infra/Pos/Slotting/Util.hs
```diff
@@ -137,40 +185,34 @@ onNewSlotImpl withLogging startImmediately action =
...
     Timestamp nextSlotStart <- getSlotStartEmpatically nextSlot
     let timeToWait = nextSlotStart - curTime
+
+    let applyTimeout a = case onspTerminationPolicy onsp of
+          NoTerminationPolicy -> a
+          NewSlotTerminationPolicy name ->
+              whenNothingM_ (timeout timeToWait a) $
+                  logWarning $ sformat
+                  ("Action "%stext%
+                   " hasn't finished before new slot started") name
```
volhovm: I'd write "was terminated because of timeout" to explicitly emphasize it was caller decision to kill it, not the action's one.

networking/src/Mockable/Concurrent.hs
```diff
@@ -98,6 +120,24 @@ delay time = liftMockable $ Delay time
 sleepForever :: ( Mockable Delay m ) => m ()
 sleepForever = liftMockable SleepForever
 
+----------------------------------------------------------------------------
+-- MyThreadId mock and helper functions
+----------------------------------------------------------------------------
+
+data MyThreadId m t where
+    MyThreadId :: MyThreadId m (ThreadId m)
```
volhovm: It returns current thread id, right? Document it maybe?

infra/Pos/Slotting/Util.hs
gromakovsky: We won't wait until current slot if `expectedSlotId` is `Nothing`. However, we'll wait till the the end of recovery and it's you who introduced this logic 😉

networking/src/Mockable/Concurrent.hs
gromakovsky: This whole module is almost undocumented. I'd probably do it if it was pointed out in my first PR, but the issue is done already, so I won't (since it's not related to the issue anyway).

PR #2408, opened 2018-01-29
CSL-2220 Make it hard to use unsafe functions from Mockable 

gromakovsky:
Some functions from `Mockable` (`async`, `asyncWithUnmask`, `cancelWith`, `cancel`, `wait`, `waitAny`, `link`, `fork`, `throwTo`) were moved into separate mocks which are not part of `MonadMockable`, because [they usually should be avoided](https://docs.google.com/document/d/1ueuoII0UIkHbj_8eozJ_3InkLTRc0wTSBvu7g6RcrOk/edit#heading=h.4ybdz5elh6jm).
Now it's clearer which code uses these functions.

I also reviewed usages of `async` function, they look safe, except two places for which I created https://iohk.myjetbrains.com/youtrack/issue/CSL-2230.

### DIFF
--- a/networking/src/Network/Broadcast/OutboundQueue.hs
+++ b/networking/src/Network/Broadcast/OutboundQueue.hs
@@ -1226,12 +1226,12 @@ type SendMsg m msg nid = forall a. msg a -> nid -> m a
 -- It is the responsibility of the next layer up to fork this thread; this
 -- function does not return unless told to terminate using 'waitShutdown'.
 dequeueThread :: forall m msg nid buck. (
-                   MonadIO              m
-                 , MonadMask            m
-                 , M.Mockable M.Async   m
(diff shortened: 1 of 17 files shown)


networking/src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1367,21 +1367,21 @@ updatePeersBucket outQ@OutQ{..} buck f = do
 -------------------------------------------------------------------------------}
 
 data ThreadRegistry m =
-       ( MonadIO              m
-       , M.Mockable M.Async   m
-       , M.Mockable M.Fork    m
-       , MonadMask            m
-       , Ord (M.ThreadId      m)
+       ( MonadIO                 m
```
pva701: krasivo

PR #2410, opened 2018-01-30
[CBR-48] Fix haddock compilation errors

365andreas:


gromakovsky:
It seems to duplicate #2072.

365andreas:
@gromakovsky Yes it is. Should I update #2072 ?
cc @jcmincke

gromakovsky:
I don't know, to be honest. But whoever will merge this PR, should close #2072 or vice versa.

explorer/src/Pos/Explorer/ExtraContext.hs
```diff
@@ -10,12 +10,12 @@ module Pos.Explorer.ExtraContext
 
     , HasGenesisRedeemAddressInfo (..)
     , GenesisRedeemAddressInfo
-    -- ^ Genesis address info
+    -- * Genesis address info
```
jcmincke: Why a '-- *' here instead of '--'?

365andreas: `-- *` Is acceptable and creates a contents table. I can also replace it with `-- `or put the description above the functions with a `-- |`. In my opinion, the best for the html generated by haddock is the latter and the best for someone reading the source code is the former.
@jcmincke which should I prefer?

domenkozar:
Rocksdb will get further dumped by https://github.com/input-output-hk/cardano-sl/pull/2417

365andreas:
@domenkozar should I delete changes in files (stack.yaml, pkgs/default.nix) ?

domenkozar:
Best to do that, yes :)

PR #2412, opened 2018-01-30
[CSL-2127] Make getHistoryCache return empty history instead of Nothing

Martoon-00:
Rebased #2188


Martoon-00:
@adinapoli-iohk I tested only with Postman, `api/history` with no/some transactions, including case when wallet has been deleted and restored again upon `api/history` calls.
Does it seem enough, or daedalus acceptance tests should rather be checked?

adinapoli-iohk:
> Does it seem enough, or daedalus acceptance tests should rather be checked?

Seems reasonable to me 😉 

Will merge.

PR #2413, opened 2018-01-30
[CSL-2106] Add V1 `fees` endpoint

sectore:
_Issue:_ https://iohk.myjetbrains.com/youtrack/issue/CSL-2106


pkgs/result
```diff
@@ -0,0 +1 @@
+/nix/store/rzqyfil4xmja85w5sm13m107j1zm8d7s-ghc-8.0.2
```
adinapoli-iohk: @sectore Oh! This seems to be an unwanted artifact -- am I right?

PR #2420, opened 2018-01-31
[CSL-2212] Remove duplicated logic in wallet-new

Martoon-00:
It so happened that in old wallet, 1. notifier launch and 2. wallets sync were
moved to some other place (from `walletProd` to Main module); in wallet-new Main
got those changes while corresponding `walletBackend` function didn't, thus we
got 1 and 2 executed twice in wallet-new.

`walletBackend` in wallet-new remain outdated even after CSL-2212 where notifier
launcher became blocking operation in order to have control over the
relative thread, this change prevented wallet startup sequence completion.

### DIFF
--- a/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
+++ b/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
@@ -20,23 +20,22 @@ import           Cardano.Wallet.Server.CLI (RunMode, WalletBackendParams (..), i
                                             walletAcidInterval, walletDbOptions)
 
 import qualified Control.Concurrent.STM as STM
+import           Formatting (build, sformat, (%))
 import           Network.Wai (Application, Middleware)
 import           Network.Wai.Middleware.Cors (cors, corsMethods, corsRequestHeaders,
                                               simpleCorsResourcePolicy, simpleMethods)
 import           Network.Wai.Middleware.RequestLogger (logStdoutDev)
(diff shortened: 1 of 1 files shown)


wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -82,14 +86,11 @@ walletBackend WalletBackendParams {..} =
     -- Gets the Wai `Application` to run.
     getApplication :: SendActions WalletWebMode -> WalletWebMode Application
     getApplication sendActions = do
-      logInfo "DAEDALUS has STARTED!"
+      logInfo "Wallet Web API has STARTED!"
       saVar <- asks wwmcSendActions
       atomically $ STM.putTMVar saVar sendActions
-      when (isDebugMode walletRunMode) $ addInitialRichAccount 0
```
adinapoli-iohk: @Martoon-00 Do we have a counterpart for this somewhere else? (i.e. I'm asking if that's fine to remove this line for good 😁 )

Martoon-00: Oh, exactly this like was once used to create wallet on startup for debugging, but it has not been maintained for a really long time, it can't be it still works :smile:

adinapoli-iohk: I see, fine 😉

adinapoli-iohk:
Sweet! Will merge as soon as CI goes green. Thanks both!

PR #2421, opened 2018-01-31
[CSL-2064] Add a bit more logging to `cardano-launcher'

gromakovsky:
It can help us understand why it doesn't kill the node on MacOS.

### DIFF
--- a/util/Pos/Util/Util.hs
+++ b/util/Pos/Util/Util.hs
@@ -286,7 +286,12 @@ logException :: LoggerName -> IO a -> IO a
 logException name = E.handleAsync (\e -> handler e >> E.throw e)
   where
     handler :: E.SomeException -> IO ()
-    handler = usingLoggerName name . logError . pretty
+    handler exc = do
+        let message = "logException: " <> pretty exc
+        usingLoggerName name (logError message) `E.catchAny` \loggingExc -> do
+            putStrLn message
+            putStrLn $
+                "logException failed to use logging: " <> pretty loggingExc
 
 -- | 'bracket' which logs given message after acquiring the resource
 -- and before calling the callback with 'Info' severity.
--- a/tools/src/launcher/Main.hs
+++ b/tools/src/launcher/Main.hs
@@ -74,7 +74,7 @@ import           Pos.Reporting.Methods (compressLogs, retrieveLogFiles, sendRepo
 import           Pos.ReportServer.Report (ReportType (..))
 import           Pos.Update (installerHash)
 import           Pos.Update.DB.Misc (affirmUpdateInstalled)
-import           Pos.Util (HasLens (..), directory, postfixLFields, sleep)
+import           Pos.Util (HasLens (..), directory, logException, postfixLFields, sleep)
 import           Pos.Util.CompileInfo (HasCompileInfo, retrieveCompileTimeInfo, withCompileInfo)
 
 data LauncherOptions = LO
@@ -316,7 +316,7 @@ main =
                   Just _  ->
                       set Log.ltFiles [Log.HandlerWrap "launcher" Nothing] .
                       set Log.ltSeverity (Just Log.debugPlus)
-    Log.usingLoggerName "launcher" $
+    logException loggerName . Log.usingLoggerName loggerName $
         withConfigurations loConfiguration $
         case loWalletPath of
             Nothing -> do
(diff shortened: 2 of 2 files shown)


util/Pos/Util/Util.hs
```diff
@@ -285,8 +285,15 @@ multilineBounds maxSize = F.later formatList
 logException :: LoggerName -> IO a -> IO a
 logException name = E.handleAsync (\e -> handler e >> E.throw e)
   where
+    -- Two ways of logging are used.
+    -- 1. log-warper is used, because it can log to files, etc.
+    -- 2. Simple 'putStrLn' is used, because something may be wrong
+    -- with log-warper and this function is useful in exceptional situations.
     handler :: E.SomeException -> IO ()
-    handler = usingLoggerName name . logError . pretty
+    handler exc = do
```
Pastafarianist: On MacOS, this will work, but it won't work on Windows since we silence all `stdout` and `stderr` there. To work around that, see `reportErrorDefault` [here](https://github.com/input-output-hk/cardano-sl/blob/604f58270f790a2338d854ea23214640a4b9baba/tools/src/launcher/Main.hs#L174) and consider either using it or rewriting it to avoid duplicating entities.

cleverca22: stdout is visible on windows, if the program is manually ran from a terminal, and doesnt go out of its way to silence its own stdout

Pastafarianist: @cleverca22 are you absolutely sure about this? I remember having to write this function because there was no `stdout` output on Windows, but I may be mistaken.

cleverca22:
```
module Main where

import System.IO

main :: IO ()
main = do
    hPutStrLn stdout "hello world"
    hPutStrLn stderr "and std err"
```

```
C:\stack>stack ghc -- example.hs -o example.exe
[1 of 1] Compiling Main             ( example.hs, example.o )
Linking example.exe ...

C:\stack>example.exe
hello world
and std err
```

cleverca22: https://github.com/input-output-hk/cardano-sl/blob/develop/tools/src/launcher/Main.hs#L296-L301

Pastafarianist: @cleverca22 yes, those lines are exactly what I mean by "silencing all `stdout` and `stderr`". Can you test this code?

```
module Main where

import System.IO
import qualified System.IO.Silently as Silently

main :: IO ()
main =
    Silently.hSilence [stdout, stderr] $ do
        hPutStrLn stdout "hello world"
        hPutStrLn stderr "and std err"
```

cleverca22: cant get that one to compile under windows, since i dont have nix helping out and auto-loading all of hackage

but it does make sense why that would fail, i'm thinking that part of launcher should either be removed or made optional

why was that added originally?

Pastafarianist: @cleverca22 there is a comment in the lines you linked to [here](https://github.com/input-output-hk/cardano-sl/pull/2421#discussion_r165798894) explaining that.

gromakovsky: > On MacOS, this will work, but it won't work on Windows since we silence all stdout and stderr there.

What exactly won't work on Windows? `putStrLn` will do nothing in `cardano-launcher`, that's true. But `logError` will log `message` to a file. That's exactly what we want on Windows, isn't it? For `cardano-node` and other executables we don't silence stdout/stderr, so it's not different from MacOS at all.

> To work around that, see reportErrorDefault here and consider either using it directly

Those are some cardano-launcher-specific hacks, I don't want to have them in `cardano-sl-util`.

> but please avoid duplicating entities.

That's the only problem, but I think it's minor. If you strictly don't like it, I will try to change it. I think the best way is to temporary disable terminal logging for logger name `name` (only for the `logError` call below). I've asked in #log-warper channel how to do it.

gromakovsky: > -- This function is only intended to be used before normal logging
> -- is initialized.

In this PR `logException` is used after logging is initialized. Which is fine for CSL-2064, because there we definitely do initialize logging successfully (strange things happen much later).

gromakovsky: @Pastafarianist okay, I changed this function, please re-review.

PR #2423, opened 2018-01-31
[DEVOPS-424] Simplify nix-based exchange workflow

jmitchell:


sample-wallet-config.nix
```diff
@@ -0,0 +1,24 @@
+# If any customization is required, copy this file to
```
domenkozar: This file needs to be in `/scripts/launch/connect-to-cluster` for this code to work

default.nix
```diff
@@ -1,5 +1,7 @@
 let
   localLib = import ./lib.nix;
+  walletConfigFile = ./custom-wallet-config.nix;
```
domenkozar: (nitpick) I suggest we move this down, inside to `connect` function with a new `let .. in`

scripts/launch/connect-to-cluster/default.nix
```diff
@@ -9,6 +9,8 @@
 , gitrev ? localLib.commitIdFromGitRepo ./../../../.git
 , walletListen ? "127.0.0.1:8090"
 , ekgListen ? "127.0.0.1:8000"
+, ghcRuntimeArgs ? "-N2 -qg -A1m -I0 -T"
+, additionalArgs ? ""
```
domenkozar: (nitpick) additonalArgs -> additionalNodeArgs

domenkozar:
Initially I wasn't happy with this approach, but on second thought it makes more sense to have declarative config over imperative args.

sample-wallet-config.nix
jmitchell: If we were using JSON that would be true, but the path resolution works about when `custom-wallet-config.nix` is in the root of the repo and paths are specified relative to that directory.

domenkozar: Ah, you're right.

PR #2427, opened 2018-02-01
[CSL-2231] Upgrade to conduit-1.3 and resourcet-1.2

gromakovsky:
This PR upgrades `conduit` and `resourcet` to more modern versions and resolves dependency hell and other breakages associated with that.

`MonadBaseControl IO` usages were removed in many places. `MonadUnliftIO` appeared instead. That's due to changes in `resourcet`.

`Source`, `Sink` and friends are deprecated in `conduit-1.3.0`, `ConduitT` is used instead.

`hoist` calls for `ResourceT`/`ConduitT` were replaced by concrete
calls to `transResourceT`/`transPipe` as these types are no longer
instances of `MFunctor`.

Demo works locally.

### DIFF
--- a/util/Pos/Util/Orphans.hs
+++ b/util/Pos/Util/Orphans.hs
@@ -154,9 +154,6 @@ instance (Typeable s, Buildable a) => Buildable (Tagged s a) where
 -- MonadResource/ResourceT
 ----------------------------------------------------------------------------
 
-instance LiftLocal ResourceT where
-    liftLocal _ l f = hoist (l f)
-
 instance {-# OVERLAPPABLE #-}
     (MonadResource m, MonadTrans t, Applicative (t m),
      MonadBase IO (t m), MonadIO (t m), MonadThrow (t m)) =>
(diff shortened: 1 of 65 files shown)


util/Pos/Util/Orphans.hs
```diff
@@ -233,3 +231,15 @@ type instance Distribution (Ether.TaggedTrans tag t m) = Distribution m
 type instance SharedExclusiveT (Ether.TaggedTrans tag t m) = SharedExclusiveT m
 type instance Gauge (Ether.TaggedTrans tag t m) = Gauge m
 type instance ChannelT (Ether.TaggedTrans tag t m) = ChannelT m
+
+instance MonadUnliftIO (t m) => MonadUnliftIO (Ether.TaggedTrans tag t m) where
+    {-# INLINE askUnliftIO #-}
+    askUnliftIO =
+        Ether.TaggedTrans $
+        withUnliftIO $ \u ->
+        return (UnliftIO (unliftIO u . \case Ether.TaggedTrans trans -> trans))
```
volhovm: I can see why this typechecks, but I can't be sure it is _correct_.

gromakovsky: Copied from https://hackage.haskell.org/package/unliftio-core-0.1.1.0/docs/src/Control-Monad-IO-Unlift.html#line-91

PR #2428, opened 2018-02-01
[CSL-2211] Restore Wallet Balance from Utxo only

adinapoli-iohk:
This PR ships the ability of restoring wallet's balance from the Utxo only, without traversing the blockchain. Subsequent PRs will add the ability to restore the history, which at the moment is not fetched.


wallet/src/Pos/Wallet/Web/Methods/Restore.hs
```diff
@@ -63,24 +69,42 @@ newWalletFromBackupPhrase passphrase CWalletInit {..} isReady = do
...
     withStateLockNoMetrics HighPriority $ \tip -> setWalletSyncTip wId tip
     L.getWallet wId
 
-restoreWallet :: L.MonadWalletLogic ctx m => PassPhrase -> CWalletInit -> m CWallet
-restoreWallet passphrase cwInit = do
-    -- Restoring a wallet may take a long time.
-    -- Hence we mark the wallet as "not ready" until `syncWalletOnImport` completes.
-    (sk, wId) <- newWalletFromBackupPhrase passphrase cwInit False
-    -- `syncWalletOnImport` automatically marks a wallet as "ready".
-    syncWalletOnImport sk
```
Martoon-00: Maybe I'm missing it, but where synchronization is initiated now?

adinapoli-iohk: @Martoon-00 This is a set of ongoing patches, that's why I'm merging into `feature/CBR-90` 😛 

This one has no sync, it will be added in a separate PR.

Martoon-00: Ah okay :slightly_smiling_face:

adinapoli-iohk:
@Martoon-00 Merging this one as I have another PR incoming which will actually do something useful 😉

PR #2433, opened 2018-02-01
[CSL-2152] Sort utxo only once during stabilization

Martoon-00:
Rewrite transaction forming logic carefully, so that Utxo is passed and processed only once during tx stabilization.

Tested with `trace`, utxo is actually sorted just once for both inputs grouping policies.

### DIFF
--- a/client/src/Pos/Client/Txp/Util.hs
+++ b/client/src/Pos/Client/Txp/Util.hs
@@ -298,7 +298,7 @@ makeLenses ''InputPickerState
 type InputPicker = StateT InputPickerState (Either TxError)
 
 plainInputPicker :: PendingAddresses -> InputPickingWay
-plainInputPicker (PendingAddresses pendingAddrs) utxo _outputs moneyToSpent =
+plainInputPicker (PendingAddresses pendingAddrs) utxo = \_outputs moneyToSpent ->
     evalStateT (pickInputs []) (InputPickerState moneyToSpent sortedUnspent)
   where
     onlyConfirmedInputs :: Set.Set Address -> (TxIn, TxOutAux) -> Bool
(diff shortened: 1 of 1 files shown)


client/src/Pos/Client/Txp/Util.hs
```diff
@@ -373,20 +373,22 @@ makeLenses ''GroupedInputPickerState
...
 groupedInputPicker :: InputPickingWay
-groupedInputPicker utxo outputs moneyToSpent =
-    evalStateT (pickInputs []) (GroupedInputPickerState moneyToSpent sortedGroups)
+groupedInputPicker utxo = \outputs moneyToSpent ->
+    let outputAddrsSet =
+            foldl' (flip HS.insert) mempty $
+            map (txOutAddress . toaOut) outputs
+        isOutputAddr = flip HS.member outputAddrsSet
+        preparedGroups = filter (not . isOutputAddr . ugAddr) sortedGroups
+        disallowedInputGroups = filter (isOutputAddr . ugAddr) gUtxo
```
nc6: Can use `partition` here to avoid double iterating the list, and then only have to sort the smaller set.

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -399,23 +401,29 @@ groupedInputPicker utxo outputs moneyToSpent =
...
+-- supports that.
 prepareTxRawWithPicker
     :: Monad m
     => InputPickingWay
     -> Utxo
     -> TxOutputs
     -> TxFee
     -> TxCreator m TxRaw
-prepareTxRawWithPicker inputPicker utxo outputs (TxFee fee) = do
+prepareTxRawWithPicker inputPickingWay utxo = \outputs (TxFee fee) -> do
```
nc6: What is the purpose of this rewrite? I could sort of understand earlier where you're constructing an `InputPickingThing`, but here it just seems to make things more opaque.

client/src/Pos/Client/Txp/Util.hs
Martoon-00: Right, didn't notice this thing.

I didn't get point about
> have to sort the smaller set

thought. I have to filter after sorting because filtering requires `outputs` be known.

client/src/Pos/Client/Txp/Util.hs
Martoon-00: Maybe I understand it incorrectly, but with previous version even if compiler understands that `inputPicker` evaluation could be done once when first two arguments are fixed, that's only due to optimizations, and it would be better not to rely on them and write explicitly.

nc6: I'm fairly certain there's no semantic distinction between these concepts: you can see they generate exactly the same Core:

```
Prelude> let foo a b c = a $ b c

==================== Simplified expression ====================
returnIO
  (: ((\ (@ t) (@ t) (@ t) (a :: t -> t1) (b :: t -> t1) (c :: t) ->
         $ a (b c))
      `cast` <Co:15>)                                                                                                                                                      
     [])                                                                                                                                                                   
                                                                                                                                                                           
                                                                                                                                                                           
Prelude> let foo a b = \c -> a $ b c
                                                                                                                                                                           
==================== Simplified expression ====================                                                                                                            
returnIO                                                                                                                                                                   
  (: ((\ (@ t) (@ t) (@ t) (a :: t -> t1) (b :: t -> t1) (c :: t) ->                                                                                                       
         $ a (b c))                                                                                                                                                        
      `cast` <Co:15>)                                                                                                                                                      
     [])     
```

Even if there were a difference here, I think we'd want to be careful about exploiting such things, because it makes it very prone to accidental breaking. So we should have at least warnings in the comments if we're relying on nuances like this.

nc6: (In this case, since there is no difference, we should prefer the most straightforward way)

client/src/Pos/Client/Txp/Util.hs
nc6: ``` sortedGroups = sortOn (Down . ugTotalMoney) . filter (not . isOutputAddr . ugAddr ) $ gUtxo```

Martoon-00: The point of this PR is to do sorting once when `groupedInputPicker`, with fed `utxo`, is executed multiple times;
in ^ way I will get `filter` part executed when `outputs` is given.

client/src/Pos/Client/Txp/Util.hs
Martoon-00: Thanks, nice point.
And I agree about warnings, going to scatter them around.

Martoon-00: Hm, I think there is a difference here after all. Simplified example:
```
meq :: Int -> Int -> Int -> Int
meq a b c = d + c
  where
    d = a + b
```
translates to
```
-- RHS size: {terms: 12, types: 6, coercions: 0}
meq :: Int -> Int -> Int -> Int
meq =
  \ (a :: Int) (b :: Int) (c :: Int) ->
    let {
      d :: Int
      d = + $fNumInt a b } in
    + $fNumInt d c
```

While 
```
meq :: Int -> Int -> Int -> Int
meq a b = \c -> d + c
  where
    d = a + b
```
does to
```
-- RHS size: {terms: 12, types: 6, coercions: 0}
meq :: Int -> Int -> Int -> Int
meq =
  \ (a :: Int) (b :: Int) ->
    let {
      d :: Int
      d = + $fNumInt a b } in
    \ (c :: Int) -> + $fNumInt d c
```
Evaluation on first arguments will be shared for `meq 1 2` calls only for second case.

Martoon-00: Ah, need in this thing in that code eventually disappeared.

Martoon-00:
See rebased version ^

PR #2436, opened 2018-02-02
First version of the pure wallet

edsko:
This introduces module `Wallet.Spec`, which is a fairly direct translation of
Duncan's pure wallet specification. The idea is that we should be able to use
this to unit test the "real" wallet later.

This also makes a few minor additional changes:

* The 'isBootstrapTransaction' now just checks if the set of inputs is empty
  (that's what it should have done in the first place..)
* Few additional operations on the UTxO in the DSL, including a definition
  of equality.
* A 'PreChain' now gets the actual boot transaction rather than its hash,
  so that we inspect it to generate arbitrary valid ledgers. (We may have
  to add a monad parameter to 'PreChain' though.)

### DIFF
--- a/wallet-new/test/unit/UTxO/DSL.hs
+++ b/wallet-new/test/unit/UTxO/DSL.hs
@@ -50,8 +50,11 @@ module UTxO.DSL (
   , utxoFromList
   , utxoToList
   , utxoDomain
+  , utxoRange
   , utxoRemove
   , utxoUnion
+  , utxoUnions
+  , utxoFilterByAddr
     -- ** Chain
   , Block
   , Blocks
@@ -337,7 +340,9 @@ findHash' h l = fromJust err (findHash h l)
 -------------------------------------------------------------------------------}
 
 -- | Unspent transaciton outputs
-data Utxo h a = Utxo { utxoToMap :: Map (Input h a) (Output a) }
+newtype Utxo h a = Utxo { utxoToMap :: Map (Input h a) (Output a) }
+
+deriving instance (Hash h a, Eq a) => Eq (Utxo h a)
 
 utxoEmpty :: Utxo h a
 utxoEmpty = Utxo Map.empty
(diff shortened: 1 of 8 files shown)


wallet-new/test/unit/UTxO/DSL.hs
```diff
@@ -354,12 +359,21 @@ utxoToList = Map.toList . utxoToMap
 utxoDomain :: Utxo h a -> Set (Input h a)
 utxoDomain = Map.keysSet . utxoToMap
 
+utxoRange :: Utxo h a -> [Output a]
```
nc6: Are 'Range' and 'Domain' conventionally used in the context of UTXO accounting? Otherwise these aren't terribly obvious unless you're aware of the Map inside `Utxo`, which somewhat breaks the abstraction.

edsko:
Spec is not yet available online as far as I know, still very much in flux.

wallet-new/test/unit/UTxO/DSL.hs
edsko: This is terminology straight from the formal model.

edsko:
Sorry guys, this is broken. Don't merge yet.

edsko:
Ok, I believe it's correct now. Detailed comments in the code.

edsko:
Just for the record, I'd just to whine once more about the fee calculation model :) It's causing all sorts of trouble.

wallet-new/test/unit/UTxO/PreChain.hs
```diff
@@ -27,7 +27,30 @@ import UTxO.Translate
...
+--
+-- > m (Transaction h Addr -> [[Fee]] -> Blocks h Addr)
+--
+-- instead, then (thinking @m@ = 'Gen' again) we could not generate different
+-- chains for different bootstrap transactions, which would also be no good.
+--
+-- By having the @m@ parameter after the bootstrap transaction but before
+-- the fees we avoid having to run the generator twice. It is still the
+-- responsibility of the 'PreChain' author to make sure that the structure of
+-- the blockchain does not depend on the fees that are passed.
```
parsonsmatt: Great explanation -- thanks!

PR #2418, opened 2018-01-31
[CSL-2225] Prettify `dbgen` code

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2225

@dcoutts We first require https://github.com/input-output-hk/cardano-sl/pull/2388 to be merged!

With this one merged, we can address https://github.com/input-output-hk/cardano-sl/pull/2400.

PR #2438, opened 2018-02-02
[CSL-2210] Add option to dbgen to generate fake history

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2210

We first need https://github.com/input-output-hk/cardano-sl/pull/2418 to be able to merge this.

Tested on `mainnet`, working. The last part of the JSON response when querying `curl -X GET --insecure https://localhost:8090/api/txs/histories?walletId=Ae2tdPwUPEZCFJzRM8mMRULhAnwwEcZWzswh72ieApfKY9imuGMCfqQmQqR | jq .` is:
```
            }
          ],
          [
            "2JZF6DQDSxEBUk2pShHy8qd4oMy6aJB9ePrRoLkagYjEcaBPXwLth9mvdRyk3QqaWorwb2z4PN1TFfBMqjkgpXgNFKG2svb",
            {
              "getCCoin": "27193921288038690"
            }
          ],
          [
            "3JDPKH7dE4D7rXDwuAR6EfEVefSLziXj6HUEhEZjnJoDXfKWcgkjWLywP26qURHcMe1brnH4XCmyFvxPhrSPw3ypMqXKDWCcUqQPeLLnSuVEk6JyJ7N5JPnD8KPCQRd6XNbC4usc5de5TyMmgKoewx3jgCfYJ2ZXk6GibrF4AgxN9gL3rXWVs6cAg8nfsabNkRLTSNuujTReXMK1zQnV3zQHMJWToNy4SW1NSMFzoRhgpGRmS2iejKmPRAJ8vvPY2V64osKG7KmtVXbCY1L9Hqvic8bkVDo4EMsVCfjhE1tUxaMX1SXsaqxS",
            {
              "getCCoin": "30923639785195979"
            }
          ]
        ],
        "ctIsLocal": false,
        "ctIsOutgoing": true,
        "ctCondition": "CPtxNotTracked"
      }
    ],
    1000
  ]
}
```


tools/src/dbgen/Lib.hs
```diff
@@ -172,21 +202,85 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
+    let outputAddresses :: [Address]
+        outputAddresses = take txsNumber accountsAddrs
+
+    let genTxOut :: Gen [TxOut]
+        genTxOut = forM outputAddresses $ \address -> TxOut address <$> genCoins
+
+    let genTxHistoryEntry :: IO TxHistoryEntry
+        genTxHistoryEntry = generate $ THEntry
+                                <$> arbitrary
+                                <*> arbitrary -- TODO(ks): Maybe use txOut as well?
```
Martoon-00: I think using `arbitrary` is completely ok here

ksaric: It's not. 
You can't use addresses you don't have in your wallet and you can't generate huge sums of money, for example.

ksaric: Ah, sorry, I thought you were referring to the whole block above :grin:

tools/src/dbgen/Lib.hs
```diff
@@ -172,21 +202,85 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
+                                <*> arbitrary
+                                <*> genTxOut
+                                <*> pure outputAddresses
+                                <*> arbitrary
+
+    let walletId :: CId Wal
+        walletId = aiWId aId
+
+    -- Generate arbitrary tx ids and txs for the fake history.
+    fakeTxIds <- liftIO $ replicateM txsNumber genTxId
```
Martoon-00: `TxHistoryEntry` already contains `TxId`, no need to generate it again

tools/src/dbgen/Lib.hs
Martoon-00: Although just `arbitrary` seem to generate uniformly between 0 and 100 inputs which is far from real life.

tools/src/dbgen/Lib.hs
```diff
@@ -172,21 +202,85 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
+
+    -- Insert into the @WalletStorage@.
+    insertIntoHistoryCache db walletId fakeMapTxs
+  where
+
+    genTxId :: IO TxId
+    genTxId = generate arbitrary
+
+    genCoins :: Gen Coin
+    genCoins = mkCoin <$> choose (1, 1000)
```
Martoon-00: How do you ensure that sum of inputs is not less than sum of outputs? This is validated when history is displayed.

tools/src/dbgen/Lib.hs
```diff
@@ -172,21 +202,85 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
+
     case addTo of
-        Just accId -> case fakeUtxoSpec of
-            NoDistribution -> addAddressesTo accId spec
-            _              -> generateFakeUtxo fakeUtxoSpec accId
+        Just accId ->
+            if (checkIfAddTo fakeUtxoSpec fakeTxs) then
+                addAddressesTo spec accId
+            else do
+                void $ generateFakeUtxo fakeUtxoSpec accId
```
Martoon-00: Why `void`?

tools/src/dbgen/Lib.hs
ksaric: I'm not sure about this one since it could be an issue, but AFAIK wallet utxo doesn't need to be completely valid - the same principle is used when generating "fake" coins which is done in the previous PR. 
The application is for testing only anyways.

tools/src/dbgen/Lib.hs
ksaric: Will fix.

tools/src/dbgen/Lib.hs
ksaric: Ugh. Because I'm not looking at the function types? Will fix, good catch.

ksaric: Ugh. Because I'm not looking at function types? Will fix, nice catch.

tools/src/dbgen/Lib.hs
Martoon-00: I took another look, seems like `getHistory` won't crash because of that after all, false alert

tools/src/dbgen/Lib.hs
```diff
@@ -257,18 +257,17 @@ generateFakeTxs SimpleTxsHistory{..} aId   = do
...
+        fakeMapTxs = fromList $ zip (map getTxId fakeTxs) fakeTxs
 
     -- Insert into the @WalletStorage@.
     insertIntoHistoryCache db walletId fakeMapTxs
   where
 
-    genTxId :: IO TxId
-    genTxId = generate arbitrary
+    getTxId :: TxHistoryEntry -> TxId
+    getTxId (THEntry txId _ _ _ _ _) = txId
```
Martoon-00: If one day new field appears in `TxHistoryEntry`, this line will be the only one which breaks 😒

ksaric: You are right. Fixed.

tools/src/dbgen/Lib.hs
```diff
@@ -265,6 +277,25 @@ generateFakeTxs SimpleTxsHistory{..} aId   = do
...
     insertIntoHistoryCache db walletId fakeMapTxs
   where
+    -- | Generate sensible @Tx@.
+    genTxs :: Gen Tx
+    genTxs = do
+
+        -- Generate a more realistic distribution. @Martoon said:
+        -- After release we have limit - ~76 inputs in transaction,
+        -- and bittrex worked for a month or two before hitting that bound.
+        -- In other words, it should be pretty rare to see that stuff.
```
Martoon-00: Umm, it's not necessary to refer to my words )
Also I have found that mentions of any concrete partners are strongly discouraged.
Just say that it was an approximate limit one day.

tools/src/dbgen/Lib.hs
```diff
@@ -227,13 +229,23 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
     -- Get the number of txs we need to generate.
     let txsNumber = fromIntegral txsCount
 
-    accounts <- getAccounts Nothing -- aId
+    let batchSize = 100
+    let batches   = txsNumber `mod` batchSize
+    let remainder = txsNumber - (batches * batchSize)
+
+    void $ replicateM batches (generateNFakeTxs batchSize aId)
+    generateNFakeTxs remainder aId
```
Martoon-00: Could you explain this? `generateNFakeTxs` seems unhappy with generating too much entries?

ksaric: Just so we can persist the txs and everything related from memory to disk and clear the memory (GC).

tools/src/dbgen/Lib.hs
ksaric: Well, I had the best intention. It seemed kind of wrong to literally copy your words and not give you any credit.
I thought about "concrete partners" as well, it should probably be visible on explorer anyway, but ok.

Martoon-00: > It seemed kind of wrong to literally copy your words and not give you any credit.
To allow people blame me later :peka: :smile:

tools/src/dbgen/Lib.hs
Martoon-00: Aha, I see it. Then please specify in comment, why we don't generate all transactions at once, and why don't we generate and put 1 transaction at once.

tools/src/dbgen/Lib.hs
```diff
@@ -227,13 +229,23 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
-
-    db <- askWalletDB
-
     -- Get the number of txs we need to generate.
     let txsNumber = fromIntegral txsCount
 
-    accounts <- getAccounts Nothing -- aId
+    let batchSize = 100
+    let batches   = txsNumber `mod` batchSize
+    let remainder = txsNumber - (batches * batchSize)
```
Martoon-00: Also these two lines look very suspicious.

tools/src/dbgen/Lib.hs
```diff
@@ -172,21 +204,109 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
+
+-- | Here we generate fake txs. For now it's a simple arbitrary generation.
+generateFakeTxs :: FakeTxsHistory -> AccountId -> UberMonad ()
+generateFakeTxs NoHistory _                = pure ()
+generateFakeTxs SimpleTxsHistory{..} aId   = do
+    -- Get the number of txs we need to generate.
+    let txsNumber = fromIntegral txsCount
+
+    let batchSize = 100
+    let batches   = txsNumber `mod` batchSize
```
pva701: are you sure it shouldn't be `div`?

tools/src/dbgen/Lib.hs
```diff
@@ -172,21 +204,109 @@ generateWalletDB CLI{..} spec@GenSpec{..} = do
...
+        accountsAddrs = rights . map unwrapCAddress $ concatMap caAddresses accounts
+
+    let outputAddresses :: [Address]
+        outputAddresses = take txsNumber accountsAddrs
+
+    let genTxOut :: Gen [TxOut]
+        genTxOut = forM outputAddresses $ \address -> TxOut address <$> genCoins
+
+    let genTxHistoryEntry :: IO TxHistoryEntry
+        genTxHistoryEntry = generate $ THEntry
```
pva701: I have two notices for this function:
1. Do I understand correctly that `_thTx` isn't consistent with `_thInputs` and `_thOutputAddrs`? They must be, but it seems to be not important for our current cases.
2. Do I understand correctly that we generate only incoming transactions, not outgoing?

tools/src/dbgen/Lib.hs
```diff
@@ -233,13 +233,16 @@ generateFakeTxs SimpleTxsHistory{..} aId   = do
     let txsNumber = fromIntegral txsCount
 
     let batchSize = 100
-    let batches   = txsNumber `mod` batchSize
-    let remainder = txsNumber - (batches * batchSize)
+    let batches   = txsNumber `div` batchSize
+    let remainder = txsNumber `mod` batchSize
```
Martoon-00: There is `divMod` :star:
but ok

tools/src/dbgen/Lib.hs
ksaric: 1. Yes. I found this working so I thought we didn't need it. Maybe we do?
2. We generate transactions coming in the existing addresses of the accounts of the wallet, yes. Do you think we need to generate outgoing addresses?

pva701: 2. dunno, at least I think you should drop comment about it

ksaric:
Awesome. @gromakovsky can you please merge this :pray: ? Squash merge since it's full of garbage.

ksaric:
:+1:

PR #2447, opened 2018-02-04
[DEVOPS-648] quick fix on the cardano side of things

cleverca22:


gromakovsky:
Btw, I don't need this fix for CSL-2231, because new version of `yaml` was uploaded to Hackage, so now it's in `extra-deps`, not in `packages`.

domenkozar:
Ah thanks!

PR #2450, opened 2018-02-04
[CSL-2129] Add HTTP-based block downloading

neongreen:


block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -72,16 +83,67 @@ retrievalWorker keepAliveTimer = worker outs (retrievalWorkerImpl keepAliveTimer
...
+            -- something and we'll have to skip 10000 blocks before we can
+            -- show any progress. In this case let's warn about it.
+            whenRight (unEpochOrSlot slot) $ \slotId ->
+                when (slotId ^. epochIndexL == epoch) $
+                    logDebug ("Going to download blocks and skip until slot "
+                              +|view siSlotL slotId|+", this may take " <>
+                              "some time")
+            let applyDump
+                    = decodeBlockDumpC
+                   .| (C.dropWhileC ((/= tip) . view prevBlockL)
```
Pastafarianist: Is it possible to assert here that not all blocks were `drop`ped?

auxx/src/Command/DumpBlockchain.hs
```diff
@@ -149,9 +149,9 @@ applyEpoch path = do
              "difficulty is "+|difficulty|+"")
     result <- runConduitRes $
            C.sourceFile path
-        .| decodeBlockDumpC                            -- get a stream of blocks
-        .| (C.dropWhileC ((/= tip) . view prevBlockL) -- skip blocks until tip
-        >> verifyAndApplyBlocksC True)                -- apply blocks w/ rollback
+        .| decodeBlockDumpC                               -- get stream of blocks
+        .| (C.dropWhileC ((/= tip) . view prevBlockL)     -- skip blocks until tip
```
volhovm: "Until" looks suspicious here. You will drop `tip` itself too. I'd say "skip blocks `[start..tip]`" or something.

block/cardano-sl-block.cabal
```diff
@@ -100,8 +100,10 @@ library
                       , ether
                       , exceptions
                       , filepath
+                      , fmt
```
volhovm: I can't say I'm happy with the fact we're using two formatting libraries at the same time.

block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -72,16 +83,67 @@ retrievalWorker keepAliveTimer = worker outs (retrievalWorkerImpl keepAliveTimer
...
+        whenJust blockStorageMirror $ \dumpUrl ->
+            tryDownload epoch (toString dumpUrl)
+    handleAll recoveryLoopE $ do
+        logInfo "Starting recovery loop"
+        recoveryLoop
   where
-    mainLoop = do
+    -- | Try to download and apply a blockchain dump, starting from the
+    -- given epoch.
+    tryDownload epoch dumpUrl = whenNothingM_ getCurrentSlot $ do
```
volhovm: Please, move it to the separate function. `where` section of the worker is too big already.

block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -72,16 +83,67 @@ retrievalWorker keepAliveTimer = worker outs (retrievalWorkerImpl keepAliveTimer
...
+        let epochUrl = dumpUrl <//> format "epoch{}.cbor.lzma"
+                                           (toInteger epoch)
+        logInfo ("Downloading blockchain dump for epoch "+|epoch|+
+                 " from "+|epochUrl|+"")
+        isSuccess <- modifyStateLock HighPriority "tryDownload" $ \tip -> do
```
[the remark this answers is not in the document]
volhovm: This should be high priority, otherwise it is possible for the action to be postponed until something more important happens. But recovery is a top priority.

block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -72,16 +83,67 @@ retrievalWorker keepAliveTimer = worker outs (retrievalWorkerImpl keepAliveTimer
...
+                Left err -> do
+                    logError ("Failed to apply the dump: "+|err|+"")
+                    pure (tip, False)
+                Right newTip ->
+                    pure (newTip, True)
+        when isSuccess $ tryDownload (succ epoch) dumpUrl
+    tryDownloadE e =
+        reportOrLogE "retrievalWorker tryDownloadE: error caught " e
+
+    recoveryLoop = do
```
volhovm: It's not a recovery loop, it's a block retrieval worker loop, it doesn't only do recovery, also incremental "continues" synchronization.

block/cardano-sl-block.cabal
neongreen: My hope is that we'll switch to `fmt` at some point in the future. I don't want to write any new code that uses `formatting`.

block/src/Pos/Block/Network/Retrieval.hs
neongreen: @Pastafarianist @volhovm fight

block/src/Pos/Block/Network/Retrieval.hs
neongreen: > also incremental "continues" synchronization

What does it mean? I can't parse it.

block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -72,16 +83,67 @@ retrievalWorker keepAliveTimer = worker outs (retrievalWorkerImpl keepAliveTimer
...
-    => Timer -> SendActions m -> m ()
-retrievalWorkerImpl keepAliveTimer SendActions {..} =
-    handleAll mainLoopE $ do
-        logInfo "Starting retrievalWorker loop"
-        mainLoop
+    => Timer -> Maybe Text -> SendActions m -> m ()
+retrievalWorkerImpl keepAliveTimer blockStorageMirror SendActions {..} = do
+    handleAll tryDownloadE $ do
+        epoch <- view epochIndexL <$> DB.getTipHeader
+        whenJust blockStorageMirror $ \dumpUrl ->
```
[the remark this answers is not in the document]
neongreen: Why?

block/src/Pos/Block/Network/Retrieval.hs
neongreen: It can in theory happen if we're at the penultimate block in epoch N and the last block wasn't generated – in which case there are no blocks left to download in this epoch, but we can't know that and so we'll have to download and discard all blocks from epoch N before moving to the next epoch.

block/src/Pos/Block/Network/Retrieval.hs
volhovm: There's no need to fight, you're moving it to the separate function :peka:

block/src/Pos/Block/Network/Retrieval.hs
volhovm: See `CHContinues` and `brtContinues`. I mean that this loop is also responsible for "handle the newest header that has just been propagated".

block/src/Pos/Block/Network/Retrieval.hs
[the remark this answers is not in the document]
volhovm: It doesn't make any significant difference.

block/src/Pos/Block/Network/Retrieval.hs
neongreen: I want you to fight because @Pastafarianist believes that stuff should be in `where` and I want him to stop believing it. (As for me, I'm obviously going to move it to a separate function.)

volhovm:
Also,

>neongreen wants to merge 1 commit into serokell/csl2129-http-based-block-sync from neongreen/csl-2129-http-block-downloading

block/src/Pos/Block/Network/Retrieval.hs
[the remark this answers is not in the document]
Pastafarianist: You do some work, including querying the DB, and then you may decide you didn't want it based on the value of an in-memory variable. This is a small optimization.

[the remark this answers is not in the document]
neongreen: I didn't do it because it was an optimization, I did it because it looked better this way (at least to me).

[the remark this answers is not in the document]
neongreen: Specifically, I didn't want to have two levels of indentation in that function.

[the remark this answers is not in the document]
Pastafarianist: No, I mean refactoring the code the way I suggested it would be an optimization, as this way we would avoid an extra DB query.

block/src/Pos/Block/Network/Retrieval.hs
Pastafarianist: Good point.

block/src/Pos/Block/Network/Retrieval.hs
Pastafarianist: Ah, the scope locality debate. Why don't we just rewrite the whole thing in C++ so that we could use *namespaces*?

block/src/Pos/Block/Network/Retrieval.hs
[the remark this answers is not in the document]
neongreen: Oh, I see. Okay, sure.

block/src/Pos/Block/Network/Retrieval.hs
neongreen: > Ah, the scope locality debate.

To be honest, I don't see a particularly good reason to spend effort / hurt readability in order to preserve the “no more things in scope than absolutely necessary” thing. Sure it feels *nice* (just like it feels nice to not use `WorkMode` when a smaller set of constraints suffices), but that's not a particularly good reason.

neongreen:
> I suppose we'll download blocks from epoch of b, which is not correct. Maybe you want to do something smarter here.

This is true. I don't know what smarter thing I could do, though. With epochs ranging between 5–15 MB it probably doesn't matter much; and when epochs grow bigger, the right thing to do would be simply creating a server that can serve a suffix of an epoch on demand.

block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -192,19 +224,74 @@ retrievalWorkerImpl keepAliveTimer SendActions {..} =
...
+            when (slotId ^. epochIndexL == epoch) $
+                logDebug ("Going to download blocks and skip until slot "
+                          +|view siSlotL slotId|+", this may take some time")
+        let applyDump
+                = decodeBlockDumpC
+               .| (C.dropWhileC ((/= tip) . view prevBlockL)
+               >> C.transPipe lift (verifyAndApplyBlocksC True))
+        -- We use Data.Conduit.Async to keep a buffer of downloaded data and
+        -- apply blocks in parallel with downloading them. The buffer size
+        -- is 16k * 512 = 8 megabytes.
```
volhovm: Where did you get these numbers from? 😒

neongreen: 16 kB = I tried to remember what was the default chunk size from `Data.ByteString.Lazy` but failed (it's actually 32 kB). I'll change it.

8 MB = approximately one epoch.

volhovm: Just add some comments about it.

PR #2455, opened 2018-02-05
[CSL-2260] Use Microsecond type in Timer

coot:


avieth:
Have you checked all call sites of `setTimerDuration` and `newTimer`?

coot:
Good catch, I pushed the changes.  Indeed there was a typo in the title. I also fixed a warning when compiling with stack about not exposed module from a test in `wallet-new`

avieth:
> I also fixed a warning when compiling with stack about not exposed module from a test in wallet-new

Right I saw this warning too but I just ignored it :) 

Who's in charge of wallet-new? @adinapoli-iohk ?

adinapoli-iohk:
@avieth @coot I _think_ (but I'm toasted so don't take my words for meaningful 😁 ) that this stems from some recent renaming work Mr @edsko did. In particular, now `Handlers.hs` points to a [desolated cliff](https://github.com/input-output-hk/cardano-sl/blob/develop/wallet-new/server/Cardano/Wallet/API/V1/Handlers.hs) and we forgot to update the specs to actually use the handlers in `LegacyHandlers`.

Said that, considering the current specs are not very useful at this stage, I think that's a benign warning: it's really telling us that something changed and that we should address it, but in practice it won't make any substantial difference right now 😁  (I hope to have stated my thoughts clearly enough).

avieth:
@coot I'd say leave the warning be. It's certainly beyond the scope of this patch.

The CI failure is weird. Some missing constraint on a binary spec test. How could these changes have caused it?

adinapoli-iohk:
@avieth As this PR merged `develop` into the feature branch, I bet it's due to [this PR](https://github.com/input-output-hk/cardano-sl/pull/2505) 😀

coot:
@avieth I reverted the change.

gromakovsky:
> The CI failure is weird. Some missing constraint on a binary spec test. How could these changes have caused it?

Fixed in #2514, please rebase/merge.

coot:
@gromakovsky current `develop` branch is broken:
```
    [25 of 29] Compiling Pos.Explorer.Socket.Methods ( src/Pos/Explorer/Socket/Methods.hs, .stack-work/dist/x86_64-linux-tinfo6-nopie/Cabal-1.24.2.0/build/Pos/Explorer/Socket/Methods.o )
    
    /home/marcin/repos/iohk/cardano-sl/explorer/src/Pos/Explorer/Socket/Methods.hs:58:1: warning: [-Wunused-imports]
        The import of ‘Control.Exception.Safe’ is redundant
          except perhaps to import instances from ‘Control.Exception.Safe’
        To import instances alone, use: import Control.Exception.Safe()
```
at the moment.

gromakovsky:
I wouldn't say it's _broken_. Warning is bad of course, but at least it can be built and CI passes.
https://i.imgur.com/pjd0HwP.png

coot:
ok; I though that `-Werror` is used in `CI` as in the local setup. Probably that's why the commit slipped to the `develop` branch.

gromakovsky:
Currently it's used only on Windows, unfortunately. And Windows doesn't build explorer.

PR #2456, opened 2018-02-05
Skeleton new wallet integration

edsko:
Next step is to introduce skeleton handlers.


wallet-new/src/Cardano/Wallet/Kernel/Diffusion.hs
```diff
@@ -0,0 +1,39 @@
...
+    WalletDiffusion(..)
+  , fromDiffusion
+  ) where
+
+import Universum
+
+import Pos.Core
+import Pos.Diffusion.Types
+
+-- | Wallet diffusion layer
```
avieth: @edsko It's feasible to make the existing full diffusion layer work in `IO`. Here are the constraints on its monad that it needs at present:

```Haskell
WithLogger m
MonadFix m
MonadIO m
MonadMask m
MonadRandom m
MonadBaseControl IO m
Mockable Metrics m -- and some equality constraints related to this.
```

If we get rid of `Mockable` and deal with `WithLogger` then we can specialize it. At present you can get a `Diffusion Production` and then use `runProduction` to bring it to `IO`. In fact, there's also `IO t -> Production t` so for this choice you can hoist the whole diffusion layer `Diffusion Production -> Diffusion IO`.

But in practice we may get more constraints on this monad because the full diffusion layer needs a `Logic m`, and the current (first) iteration of the full logic layer still uses the massive cardano-sl `WorkMode` constraint.

edsko: Well we want to be able to instantiate the wallet also in a testing (mock) context, so not having to instantiate a full diffusion layer is actually a bonus. My most important problem was that I am presented with a `Diffusion m` for a specific `m` for the node, but I don't want to use that `m` in the wallet kernel itself. But for now I think the translation is alright, as long as we don't need `requestTip` in the wallet (which I think we shouldn't need to).

Pastafarianist:
This PR contains [`undefined`](https://github.com/input-output-hk/cardano-sl/pull/2456/files#diff-1a1b5132f1b3040e6da512b170869f9cR10) which triggers compile-time warnings:

```
/home/ser/repos/cardano-sl/wallet-new/server/Cardano/Wallet/API/V1/Handlers.hs:10:15: warning: [-Wdeprecations]
    In the use of ‘undefined’
    (imported from Universum, but defined in Debug):
    "'undefined' function remains in code (or use 'error')"
```

How did it even pass CI?

PR #2457, opened 2018-02-05
[CSL-2152] Sort utxo only once during stabilization

Martoon-00:
Rebased #2433, with slight modification.

Rewrite transaction forming logic carefully, so that Utxo is passed and processed only once during tx stabilization.

Tested with trace, utxo is actually sorted just once for both inputs grouping policies.


node/src/Pos/Client/Txp/Util.hs
```diff
@@ -298,9 +310,18 @@ makeLenses ''InputPickerState
...
+-- | Most simple way of picking inputs:
+--
+-- * prefer ones which are not used in this slot
+--
+-- * take ones which have more coins first (in order to decrease number of
+-- picked inputs and thus decrease fee)
 plainInputPicker :: PendingAddresses -> InputPickingWay
-plainInputPicker (PendingAddresses pendingAddrs) utxo _outputs moneyToSpent =
-    evalStateT (pickInputs []) (InputPickerState moneyToSpent sortedUnspent)
+plainInputPicker (PendingAddresses pendingAddrs) utxo =
```
nc6: I think one thing that might make this clearer is rewriting this to use `let-` binding before the lambda: e.g. 
```
let
     sortedUnspent = confirmed ++ unconfirmed
 	 
     (confirmed, unconfirmed) =	...
in \_outputs moneyToSpend  -> ...
```

This makes it more obvious that the purpose here is to share the evaluation of these things.

node/src/Pos/Client/Txp/Util.hs
```diff
@@ -373,21 +394,28 @@ makeLenses ''GroupedInputPickerState
...
+    let outputAddrsSet =
+            foldl' (flip HS.insert) mempty $
+            map (txOutAddress . toaOut) outputs
+        isOutputAddr = flip HS.member outputAddrsSet
+        preparedGroups = filter (not . isOutputAddr . ugAddr) sortedGroups
+        disallowedInputGroups = filter (isOutputAddr . ugAddr) gUtxo
+        disallowedMoney = sumCoins $ map ugTotalMoney disallowedInputGroups
+    in  evalStateT (pickInputs disallowedMoney [])
+                   (GroupedInputPickerState moneyToSpent preparedGroups)
   where
```
nc6: Same here. Can we move these up to emphasize the sharing?

node/src/Pos/Client/Txp/Util.hs
Martoon-00: Right, this makes sence.

ksaric:
:arrow_up: pulled from `master` with a `rebase`.

Martoon-00:
@ksaric looks like you rebased upon some strange branch.
There are much more commits than was before, and conflicts are still here

Martoon-00:
Lol what happened to this PR, how these commits appeared here? :open_mouth:

jcmincke:
Closed because this PR is just too old and is targeting master

PR #2459, opened 2018-02-06
[CSL-2244] Refactor syncWalletOnImport

adinapoli-iohk:
This PR refactors slightly `syncWalletOnImport` and `syncWalletFromGState`, paving the way to finally restore a wallet asynchronously.

Tests conducted:

- [X] Restoring the genesis wallet from backup works
- [X] Restoring from seed works
- [x] Importing from JSON works


wallet/src/Pos/Wallet/Web/Server/Launcher.hs
```diff
@@ -65,7 +65,9 @@ walletServer
     => (forall x. m x -> Handler x)
     -> m (Server WalletSwaggerApi)
 walletServer nat = do
-    syncWalletsWithGState =<< mapM findKey =<< myRootAddresses
+    syncResults <- syncWalletsFromGState =<< mapM findKey =<< myRootAddresses
+    -- TODO(adn): Make sure re-throwing potential failures is not disruptive here.
+    mapM_ processSyncResult syncResults
```
Martoon-00: This one looks duplicate of
https://github.com/input-output-hk/cardano-sl/pull/2459/files#diff-08add64f2bb6347f4c3d0e786086d2b1R86
Dunno how it happened, but we can remove this line completely eventually

adinapoli-iohk: @Martoon-00 Probably a bad merge on my part, I made a little mess with this branch, I'd have to admit, hehe.

Martoon-00: Btw, do you actually throw any error there?

adinapoli-iohk: @Martoon-00 I have decided to just log them (with `logError`), as I'm not sure that throwing an error would be very beneficial as we cannot recover anyway. What do you think?

Martoon-00: > Probably a bad merge on my part, I made a little mess with this branch, I'd have to admit, hehe.
Nope-nope, it seems to be so even in develop. Need to remember to remove sync here one day.

>  I have decided to just log them (with logError), as I'm not sure that throwing an error would be very beneficial as we cannot recover anyway.
I agree, there is a little we can help with if restore failed. So `processSyncResult` may not even require `MonadThrow` if this makes this piece of code look safer.

adinapoli-iohk: Good catch. Let me remove that extra constraint!

PR #2460, opened 2018-02-06
[CSL-2089] Fix diffusion layer

volhovm:
This PR attempts to cleanup some things that became redundant after diffusion layer was merged, also it tries to fix several important bugs that were not noticed at the review stage.

At the time of writing, I've pushed two commits. One removes copy-pasted logic that was moved to the diffusion layer, but wasn't cleaned up from `Block.Networking`. Second fixes exception handling that violate the exceptions guideline (`docs/exceptions.md`). It also renames some things and removes `getChainFrom` (which must be implemented in a completely different way and wasn't used anywhere). 

As for bugs, there are two of them:
1. `getHashesRange` is used without `depthLimit` in diffusion layer. This means everybody can query the whole blockchain at once, which is an easy way to attack the node. This changed wasn't discuss explicitly, so it's a bug unless it's decided to think otherwise.
2. More serious one: block/headers communication is broken in case of rollbacks. It was decided to move all the checks out of the diffusion layer, but one important bit was missed: `classifyHeaders` was returning LCA header which then was helping to form `MsgGetBlocks`. Now, we request the very same headers server returned to us after `MsgGetHeaders` call, which is a mistake, because checkpoints are distributed exponentially over last `k` blocks (so some blocks are not included). Example: client has `A B C D E F G H` (oldest first), server has `A B C D E' F' G' H'`, client sends checkpoints set `{H G E A}` and thus server wil return `B..H'`, then the corresponding blocks will be fetched. But in fact we only need `E'..H'`. This comparison logic was previously made in `classifyHeaders`. I must also admit that there is an extra check (taking LCA in `handleBlocks`) that prevents client to fail when he tries to apply unexpectedly big list of blocks (with already known prefix). But facts are still left: protocol is used incorrectly, leading to non-optimal data transfer in case of forks.

@avieth Please tell your thoughts on this and let's decide what to do. 

IMHO:
* (1) behavior should be restored, it's easy to do so
* (2) should be left as it is, because I think the protocol format will be changed to a more compact one (one request-response instead of two) anyway. Remember though, that neither cheap to increase the size of checkpoints set on the client side as it leads to slower request processing on the server side (every checkpoint must be checked).

### DIFF
--- a/block/src/Pos/Block/Logic/Header.hs
+++ b/block/src/Pos/Block/Logic/Header.hs
@@ -7,9 +7,12 @@ module Pos.Block.Logic.Header
        , classifyNewHeader
        , ClassifyHeadersRes (..)
        , classifyHeaders
+
+       , GetHeadersFromManyToError (..)
        , getHeadersFromManyTo
        , getHeadersOlderExp
-       , getHeadersRange
+       , GetHashesRangeError (..)
+       , getHashesRange
        ) where
 
 import           Universum
@@ -244,6 +247,9 @@ classifyHeaders inRecovery headers = do
                           depthDiff blkSecurityParam
             | otherwise -> CHsValid lcaChild
 
+
+data GetHeadersFromManyToError = GHFBadInput Text deriving (Show,Generic)
+
 -- | Given a set of checkpoints @c@ to stop at and a terminating
 -- header hash @h@, we take @h@ block (or tip if latter is @Nothing@)
 -- and fetch the blocks until one of checkpoints is encountered. In
@@ -255,11 +261,11 @@ getHeadersFromManyTo ::
        , WithLogger m
        , HasConfiguration
        )
-    => Maybe Word -- ^ Optional limit on how many to bring in.
+    => Maybe Word          -- ^ Optional limit on how many to bring in.
     -> NonEmpty HeaderHash -- ^ Checkpoints; not guaranteed to be
                            --   in any particular order
     -> Maybe HeaderHash
-    -> m (Either Text (NewestFirst NE BlockHeader))
+    -> m (Either GetHeadersFromManyToError (NewestFirst NE BlockHeader))
 getHeadersFromManyTo mLimit checkpoints startM = runExceptT $ do
     logDebug $
         sformat ("getHeadersFromManyTo: "%listJson%", start: "%build)
@@ -270,12 +276,12 @@ getHeadersFromManyTo mLimit checkpoints startM = runExceptT $ do
 
     -- This filters out invalid/unknown checkpoints also.
     inMainCheckpoints <-
-        noteM "no checkpoints are in the main chain" $ lift $
-        nonEmpty <$> filterM GS.isBlockInMainChain (toList checkpoints)
+        maybe (throwLocal "no checkpoints are in the main chain") pure =<<
+        lift (nonEmpty <$> filterM GS.isBlockInMainChain (toList checkpoints))
     let inMainCheckpointsHashes = map headerHash inMainCheckpoints
     when (tipHash `elem` inMainCheckpointsHashes) $
-        throwError "found checkpoint that is equal to our tip"
-    logDebug $ "got checkpoints in main chain"
+        throwLocal "found checkpoint that is equal to our tip"
+    logDebug $ "getHeadersFromManyTo: got checkpoints in main chain"
 
     if (tip ^. prevBlockL . headerHashG) `elem` inMainCheckpointsHashes
         -- Optimization for the popular case "just get me the newest
@@ -291,15 +297,16 @@ getHeadersFromManyTo mLimit checkpoints startM = runExceptT $ do
                     curH /= startHash && maybe True ((<) h) mLimitInt
             up <- lift $ GS.loadHeadersUpWhile newestCheckpoint loadUpCond
             res <-
-                note "loadHeadersUpWhile returned empty list" $
-                _NewestFirst nonEmpty (toNewestFirst $ over _OldestFirst (drop 1) up)
+                maybe (throwLocal "loadHeadersUpWhile returned empty list") pure $
+                _NewestFirst nonEmpty $
+                toNewestFirst $ over _OldestFirst (drop 1) up
             logDebug $ "getHeadersFromManyTo: loaded non-empty list of headers, returning"
             pure res
   where
-    noteM :: (MonadError e n) => e -> n (Maybe a) -> n a
-    noteM reason action = note reason =<< action
     mLimitInt :: Maybe Int
     mLimitInt = fromIntegral <$> mLimit
+
+    throwLocal = throwError . GHFBadInput
 
 -- | Given a starting point hash (we take tip if it's not in storage)
 -- it returns not more than 'blkSecurityParam' blocks distributed
@@ -354,29 +361,40 @@ getHeadersOlderExp upto = do
                 | otherwise = selGo es ii $ succ skipped
         in selGo elems ixs 0
 
+
+data GetHashesRangeError = GHRBadInput Text deriving (Show,Generic)
+
+-- Throws 'GetHashesRangeException'.
+throwGHR :: Monad m => Text -> ExceptT GetHashesRangeError m a
+throwGHR = throwError . GHRBadInput
+
 -- | Given optional @depthLimit@, @from@ and @to@ headers where @from@
 -- is older (not strict) than @to@, and valid chain in between can be
 -- found, headers in range @[from..to]@ will be found. If the number
 -- of headers in the chain (which should be returned) is more than
 -- @depthLimit@, error will be thrown.
-getHeadersRange ::
+getHashesRange ::
        forall m. (HasConfiguration, MonadDBRead m)
     => Maybe Word
     -> HeaderHash
     -> HeaderHash
-    -> m (Either Text (OldestFirst NE HeaderHash))
-getHeadersRange depthLimitM older newer | older == newer = runExceptT $ do
+    -> m (Either GetHashesRangeError (OldestFirst NE HeaderHash))
+getHashesRange depthLimitM older newer | older == newer = runExceptT $ do
     unlessM (isJust <$> lift (DB.getHeader newer)) $
-        throwError "getHeadersRange: can't find newer-older header"
+        throwGHR "can't find newer-older header"
     whenJust depthLimitM $ \depthLimit ->
         when (depthLimit < 1) $
-        throwError $
-        sformat ("getHeadersRange: depthLimit is "%int%
+        throwGHR $
+        sformat ("depthLimit is "%int%
                  ", we can't return the single requested header "%build)
                 depthLimit
                 newer
     pure $ OldestFirst $ one newer
-getHeadersRange depthLimitM older newer = runExceptT $ do
+getHashesRange depthLimitM older newer = runExceptT $ do
+    let fromMaybeM :: Text
+                   -> ExceptT GetHashesRangeError m (Maybe x)
+                   -> ExceptT GetHashesRangeError m x
+        fromMaybeM r m = maybe (throwGHR r) pure =<< m
     -- oldest and newest blocks do exist
     newerHd <- fromMaybeM "can't retrieve newer header" $ DB.getHeader newer
     olderHd <- fromMaybeM "can't retrieve older header" $ DB.getHeader older
@@ -386,13 +404,13 @@ getHeadersRange depthLimitM older newer = runExceptT $ do
     -- Proving newerD >= olderD
     let newerOlderF = "newer: "%build%", older: "%build
     when (newerD == olderD) $
-        throwError $
-        sformat ("getHeadersRange: newer and older headers have "%
+        throwGHR $
+        sformat ("newer and older headers have "%
                  "the same difficulty, but are not equal. "%newerOlderF)
                 newerHd olderHd
     when (newerD < olderD) $
-        throwError $
-        sformat ("getHeadersRange: newer header is less dificult than older one. "%
+        throwGHR $
+        sformat ("newer header is less dificult than older one. "%
                  newerOlderF)
                 newerHd olderHd
 
@@ -406,8 +424,8 @@ getHeadersRange depthLimitM older newer = runExceptT $ do
 
     whenJust depthLimitM $ \depthLimit ->
         when (depthDiff + 1 > depthLimit) $
-        throwError $
-        sformat ("getHeadersRange: requested "%int%" headers, but depthLimit is "%
+        throwGHR $
+        sformat ("requested "%int%" headers, but depthLimit is "%
                  int%". Headers: "%newerOlderF)
                 depthDiff
                 depthLimit
@@ -423,15 +441,15 @@ getHeadersRange depthLimitM older newer = runExceptT $ do
     -- branch (after first checks are performed here) and olderHd is
     -- no longer in the main chain.
     -- CSL-1950 We should use snapshots here.
-    when (null $ allExceptNewest ^. _OldestFirst) $ throwError $
-        "getHeadersRange: loaded 0 headers though checks passed. " <>
+    when (null $ allExceptNewest ^. _OldestFirst) $ throwGHR $
+        "loaded 0 headers though checks passed. " <>
         "May be (very rare) concurrency problem, just retry"
 
     -- It's safe to use 'unsafeLast' here after the last check.
     let lastElem = allExceptNewest ^. _OldestFirst . to unsafeLast
     when (newerHd ^. prevBlockL . headerHashG /= lastElem) $
-        throwError $
-        sformat ("getHeadersRange: newest block parent is not "%
+        throwGHR $
+        sformat ("getHashesRange: newest block parent is not "%
                  "equal to the newest one iterated. It may indicate recent fork or "%
                  "inconsistent request. Newest: "%build%
                  ", last list hash: "%build%", already retrieved (w/o last): "%listJson)
@@ -441,9 +459,7 @@ getHeadersRange depthLimitM older newer = runExceptT $ do
 
     -- We append last element and convert to nonempty.
     let conv =
-           fromMaybe (error "getHeadersRange: can't happen") .
+           fromMaybe (error "getHashesRange: can't happen") .
            nonEmpty .
            (++ [newer])
     pure $ allExceptNewest & _OldestFirst %~ conv
-  where
-    fromMaybeM r m = ExceptT $ maybeToRight ("getHeadersRange: " <> r) <$> m
--- a/block/src/Pos/GState/BlockExtra.hs
+++ b/block/src/Pos/GState/BlockExtra.hs
@@ -15,12 +15,10 @@ module Pos.GState.BlockExtra
        , loadHeadersUpWhile
        , loadBlocksUpWhile
        , initGStateBlockExtra
-       , blocksSourceFrom
        ) where
 
 import           Universum
 
-import           Data.Conduit (ConduitT, yield)
 import qualified Data.Text.Buildable
 import qualified Database.RocksDB as Rocks
 import           Formatting (bprint, build, (%))
@@ -176,26 +174,6 @@ loadBlocksUpWhile
     -> (Block -> Int -> Bool)
     -> m (OldestFirst [] Block)
 loadBlocksUpWhile = loadUpWhile getBlock
-
--- | Produce blocks from a database (MonadBlockDBRead) beginning at a hash.
--- Forward links are resolved after each block is yielded, and the source
--- ends when that forward link does not determine another block.
--- Semantics depend upon the MonadBlockDBRead instance, of course. If the
--- underlying database is changed while streaming, what happens?
-blocksSourceFrom
-    :: ( MonadBlockDBRead m )
-    => HeaderHash -> ConduitT () Block m ()
-blocksSourceFrom fromH = do
-    mBlk <- lift $ getBlock fromH
-    case mBlk of
-        Nothing  -> pure ()
-        Just blk -> yield blk >> continue blk
-  where
-    continue blk = do
-        mNextH <- lift $ resolveForwardLink blk
-        case mNextH of
-            Nothing    -> pure ()
-            Just nextH -> blocksSourceFrom nextH
 
 ----------------------------------------------------------------------------
 -- Initialization
--- a/explorer/src/Pos/Explorer/Socket/Methods.hs
+++ b/explorer/src/Pos/Explorer/Socket/Methods.hs
@@ -55,6 +55,7 @@ module Pos.Explorer.Socket.Methods
 
 import           Universum
 
+import           Control.Exception.Safe (tryAny)
 import           Control.Lens (at, ix, lens, non, (.=), _Just)
 import           Control.Monad.State (MonadState)
 import           Data.Aeson (ToJSON)
@@ -358,10 +359,10 @@ getBlundsFromTo
     :: forall ctx m . ExplorerMode ctx m
     => HeaderHash -> HeaderHash -> m (Maybe [Blund])
 getBlundsFromTo recentBlock oldBlock =
-    DB.getHeadersRange Nothing oldBlock recentBlock >>= \case
+    DB.getHashesRange Nothing oldBlock recentBlock >>= \case
         Left _ -> pure Nothing
-        Right (getOldestFirst -> headers) ->
-            Just . catMaybes <$> forM (NE.tail headers) getBlund
+        Right (getOldestFirst -> hashes) ->
+            Just . catMaybes <$> forM (NE.tail hashes) getBlund
 
 addrsTouchedByTx
     :: (MonadDBRead m, WithLogger m)
--- a/lib/src/Pos/Diffusion/Full/Block.hs
+++ b/lib/src/Pos/Diffusion/Full/Block.hs
@@ -21,15 +21,11 @@ import qualified Data.List.NonEmpty as NE
 import qualified Data.Set as S
 import qualified Data.Text.Buildable as B
 import           Data.Time.Units (toMicroseconds)
--- TODO hopefully we can get rid of this import. It's needed for the
--- security workers stuff and peeking into some reader context which contains
--- it (part of WorkMode).
 import           Formatting (bprint, build, int, sformat, shown, stext, (%))
 import qualified Network.Broadcast.OutboundQueue as OQ
 import           Serokell.Util.Text (listJson)
 import           System.Wlog (logDebug, logWarning)
 
--- MsgGetHeaders Bi instance etc.
 import           Pos.Binary.Communication ()
 import           Pos.Block.Configuration (recoveryHeadersMessage)
 import           Pos.Block.Network (MsgBlock (..), MsgGetBlocks (..), MsgGetHeaders (..),
@@ -483,7 +479,8 @@ handleBlockHeaders
     -> OQ.OutboundQ pack NodeId Bucket
     -> Timer
     -> (ListenerSpec m, OutSpecs)
-handleBlockHeaders logic oq keepaliveTimer = listenerConv @MsgGetHeaders oq $ \__ourVerInfo nodeId conv -> do
+handleBlockHeaders logic oq keepaliveTimer =
+  listenerConv @MsgGetHeaders oq $ \__ourVerInfo nodeId conv -> do
     -- The type of the messages we send is set to 'MsgGetHeaders' for
     -- protocol compatibility reasons only. We could use 'Void' here because
     -- we don't really send any messages.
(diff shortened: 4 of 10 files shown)


block/src/Pos/Block/Logic/Header.hs
```diff
@@ -352,29 +357,39 @@ getHeadersOlderExp upto = do
...
     => Maybe Word
     -> HeaderHash
     -> HeaderHash
-    -> m (Either Text (OldestFirst NE HeaderHash))
-getHeadersRange depthLimitM older newer | older == newer = runExceptT $ do
+    -> m (OldestFirst NE HeaderHash)
+getHashesRange depthLimitM older newer | older == newer = do
     unlessM (isJust <$> DB.getHeader newer) $
-        throwError "getHeadersRange: can't find newer-older header"
+        throwGHR "can't find newer-older header"
```
avieth: Is this really an error? Isn't it a completely normal thing that can be handled in a non-exceptional way?

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -352,29 +357,39 @@ getHeadersOlderExp upto = do
...
     whenJust depthLimitM $ \depthLimit ->
         when (depthLimit < 1) $
-        throwError $
-        sformat ("getHeadersRange: depthLimit is "%int%
+        throwGHR $
```
avieth: Again, really an error? Can't we handle it non-exceptionally?

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -384,13 +399,13 @@ getHeadersRange depthLimitM older newer = runExceptT $ do
...
-        throwError $
-        sformat ("getHeadersRange: newer and older headers have "%
+        throwGHR $
+        sformat ("newer and older headers have "%
                  "the same difficulty, but are not equal. "%newerOlderF)
                 newerHd olderHd
     when (newerD < olderD) $
-        throwError $
-        sformat ("getHeadersRange: newer header is less dificult than older one. "%
+        throwGHR $
```
avieth: This and the `throwGHR` on line 402 seem misplaced. Why are doing these checks here? Surely this function is taking on too much responsibility. Why can't it just fetch the headers and leave it at that?

lib/src/Pos/Logic/Full.hs
```diff
@@ -123,23 +117,19 @@ logicLayerFull jsonLogTx k = do
         getBlockHeader :: HeaderHash -> m (Maybe BlockHeader)
         getBlockHeader = DB.getHeader
 
+        getHashesRange
+            :: HeaderHash
+            -> HeaderHash
+            -> m (OldestFirst NE HeaderHash)
+        -- TODO CSL-2089 bug: should be Just k, but not
+        -- Nothing. Probably, listener logic is broken.
+        getHashesRange = Block.getHashesRange Nothing
```
avieth: Should it be `Just k`? We plan to improve the horribly slow block sync by allowing a client to fetch *more* than `k` blocks (i.e. the whole chain) at once.

lib/src/Pos/Logic/Types.hs
```diff
@@ -27,47 +26,33 @@ import           Pos.Util.Chrono (NE, NewestFirst, OldestFirst)
...
+    { -- | The stakeholder id of our node.
       ourStakeholderId   :: StakeholderId
-      -- Get a block, perhaps from a database.
+      -- | Get a block, perhaps from a database.
     , getBlock           :: HeaderHash -> m (Maybe Block)
-      -- Stream blocks from first hash to second hash.
-      -- Conduit is chosen mainly due to precedent: it's already used in
-      -- cardano-sl.
-    , getChainFrom       :: HeaderHash
-                         -> Source m Block
```
avieth: Why did you delete this? Please restore it.

avieth: Oh

> removes getChainFrom (which must be implemented in a completely different way and wasn't used anywhere).

It's going to be used. CSL-2148

volhovm: Yes, my idea is that:
1. All DB traversals should be conduits (it's easy to do it).
2. `getHeadersFromManyTo` and `getHashesRange` should be too, also I think first one should be implemented in terms of the second one. 

So `getChainFrom` is an unused code with implementation that will be totally different, that's why i deleted it. I suggest we keep things simple and move on incrementally. It's 5m to somebody to return it back, but it takes much more time for me to figure out which functionality is actually needed and which is unused.

lib/src/Pos/Logic/Full.hs
volhovm: In fact, it's simpler to get rid of this parameter at all. I just wanted to consult whether it is a mistake or not.

volhovm:
@avieth any comments on the (2) bug?

block/src/Pos/Block/Logic/Header.hs
int-index: The updated code follows the guidelines in https://github.com/input-output-hk/cardano-sl/blob/develop/docs/exceptions.md unlike the old code.

The reasoning behind the current design choice in the guidelines is that we want only a single way to throw exceptions in any part of code - it could be `ExceptT` in pure code, or `MonadThrow` in (potentially) impure code. This guarantees that we can safely use `bracket` everywhere, and that there's never need to use both `catch` and `catchError`.

This code was using `throwError`, but it wasn't pure, so it was rightly rewritten. If you disagree with the use of `throwM` here, the alternative is to return `Either`, but without `runExceptT` which creates a secondary exit path from any part of the code.

Additionally, we don't use `Text` to represent errors (or "normal things", as you refer to them above) anymore, and the updated code at least uses a newtype (ideally it should be an ADT).

block/src/Pos/Block/Logic/Header.hs
int-index: Again, `throwError` that was there before is not "non-exceptionally" either, we should either use `throwM` here or a non-monadic way to handle this codepath.

avieth:
> (1) behavior should be restored, it's easy to do so

I'm not so sure. IIRC the `depthLimit` was in practice determined by the recovery headers limit parameter in the configuration. It was in practice 2200. This also was used to compute the size limit of `MsgHeaders`. Its effectiveness at mitigating a denial of service attack is modest: the attacker can just spam `MsgGetHeaders` requests which it knows will lead to 2200 headers in the response. Is that significantly less bad than allowing an attacker to request the whole blockchain?

avieth:
>  any comments on the (2) bug?

I'm trying to understand it. We give the server our checkpoints and tip. Server responds with a list of hashes corresponding to a chain, where one end is in our checkpoint set, and the other is the tip we requested. We used to follow that chain to get the LCA with our chain, potentially coming up with a smaller chain to request.

Another option is to have the server do this, no? Instead of giving back all of the headers in a list it could give two hahses, oldest and newest, where the oldest lies in-between two of our checkpoints, or coincides with one of them. That's to say, the *server* computes the shortest chain that the client needs. Even better, the server could *not* respond with the two hashes, but instead with the actual shorted blockchain that the client needs.

volhovm:
>Is that significantly less bad than allowing an attacker to request the whole blockchain?

We don't have any streaming style, which means that everything will be first load into the RAM. It means that if all the hashes in the chain are big, it's an easy way to fill up all the server RAM. I guess this is much easier attack than just spamming `MsgGetBlocks` (`getHeaderHashes` is used there), which would target lot of IO, which in turn would slow down the whole system. We use blake2b, so `256` bytes per 400k blocks (this is current chain length I believe) is 100Mb, which is huge. 

>where the oldest lies in-between two of our checkpoints, or coincides with one of them

How would you compute this on the server side? It doesn't have enough info about client side chain. 

One more note: previously, sending headers list to client (instead of two headers as you suggest) was used to check that the server side is not trying to fool us, as we can perform pretty good headers validation. It's not that trivial to tamper, because headers have slot leaders which we already know in advance, and they also have signatures.

avieth:
> > where the oldest lies in-between two of our checkpoints, or coincides with one of them
>
> How would you compute this on the server side? It doesn't have enough info about client side chain.

Right indeed, it would have to give one of the actual checkpoints.

How about a newest-to-oldest approach? Server claims to have a more difficult tip than client's, so client asks for the chain *from the tip, backwards*. Server can stream the blocks one-at-a-time with rate controlled by the client, and client can stop as soon as it meets its existing chain.

> One more note: previously, sending headers list to client (instead of two headers as you suggest) was used to check that the server side is not trying to fool us, as we can perform pretty good headers validation. It's not that trivial to tamper, because headers have slot leaders which we already know in advance, and they also have signatures.

Some checks can be done in the streaming style as well (with a bunch of state-related hazards of course). For each new block, assuming we have enough information about its epoch, we can check that it's signed by the right leader, and that the block we received before it has the appropriate hash, etc. and stop the streaming if this is ever violated.

volhovm:
>How about a newest-to-oldest approach? 

Very controversial. 
* We can't immediately data that we get. We also can't verify it. 
* It's O(n) versus O(logn) in existing approach, and I don't see any advantages. I think assuming that server side has returned correct chain (without verifying) and finding LCA on the client in the unsafe way (meaning that we can fail to find it) is faster and better.

>Some checks can be done in the streaming style as well 

This is not _exactly_ true, but can be in theory. And I'd like it to be true.

What about (1)? Can we bring the check back for now and later rework this when switching to the never protocol version?

avieth:
> It's O(n) versus O(logn) in existing approach, and I don't see any advantages. I think assuming that server side has returned correct chain (without verifying) and finding LCA on the client in the unsafe way (meaning that we can fail to find it) is faster and better.

I don't see the linear v. logarithmic distinction. In either case, we have to transmit the same number of blocks.

> What about (1)? Can we bring the check back for now and later rework this when switching to the never protocol version?

Ok I'll look into this. To be sure I understand: after requesting the headers, find the newest hash in the list such that we have that block, and then request from the next header to the tip?

volhovm:
>I don't see the linear v. logarithmic distinction. In either case, we have to transmit the same number of blocks.

Yes, there's obviously the same number of blocks. We try to minimize the data to get `[b1,bn]` which we should then query. Indeed, my mistake, in your top-to-bottom solution the number of headers sent from server to client is O(n) as it is in worst case in the current solution. What we _could_ do is to send client `[c1..c2]`, where `c1` is the newest checkpoint present on the server and `c2` is the oldest checkpoint (after `c1`) which is not present on it. Then client can find LCA among `[c1..c2]`. This is still O(n) data sent, but much less (n/2 in the worst case). 

Top-to-bottom solution seems fine actually, but i suggest to think about everything in a separate ticket, because we better batch all the protocol changes to make things optimal. I would also be happy to see some simplification of this two-request scheme if it's possible. What i suggest now is to take LCA between querying headers and asking for blocks, doing no checks. 

>after requesting the headers, find the newest hash in the list such that we have that block, and then request from the next header to the tip?

Yes, this is a correct understanding, but it's not relevant to (1). (1) is about restricting traversals. See this comment from me again (first part of it): https://github.com/input-output-hk/cardano-sl/pull/2460#issuecomment-363836042

avieth:
> Yes, this is a correct understanding, but it's not relevant to (1). (1) is about restricting traversals. See this comment from me again (first part of it): #2460 (comment)

Oh right, got my numbers mixed up. I was talking about 2.

block/src/Pos/Block/Logic/Header.hs
avieth: > Additionally, we don't use Text to represent errors (or "normal things", as you refer to them above) anymore, and the updated code at least uses a newtype (ideally it should be an ADT).

Sure I'm all for that. In principle, though, I think many of the `throw*`s seen in this patch aren't really exceptions, they're normal outcomes. This line in particular: a header hash was given, but it's not in the database. Ok, that's an expected outcome, and the caller can deal with it normally. We wouldn't promise the caller a total function from header hash to blocks, so we should't pretend we have by giving such a type.

int-index: The normal outcomes that short-circuit the computation (thus can be, theoretically, represented by `MonadThrow` or `MonadError`) go under the name of "regular errors" in the guidelines. You might argue that it's a misnomer, but it doesn't change the essence of the question.

avieth:
Actually I believe issue 1 isn't a problem. `getHeadersFromManyTo` already has the `recoveryHeadersMessage` built-in:

```Haskell
let loadUpCond (headerHash -> curH) h =
        curH /= startHash && h < recoveryHeadersMessage
GS.loadHeadersUpWhile newestCheckpoint loadUpCond
```

@volhovm Should be fine, right?

volhovm:
@avieth Not really. Indeed, when server is asked about headers (given checkpoints), he returns no more than `recoveryHeadersMessage`. So yes, protocol works as expected next -- client queries about the same number of headers. But again, without `getHeaderHashes` limitation anybody can ask server to retrieve `[b0..bs]` , where `b0` is the very first block, and `bs` is some block having difficulty 10^6+, and node will load everything into memory. Hashes are 256 bytes, so 10^6 hashes is about 250 Mb. That's a very easy OOM, takes malicious client just a bunch of simultaneous calls to achieve node crash.

We can later remove this limit when we switch to conduit-based streaming (which will not load all the data into memory at once).

avieth:
@volhovm There's 2 listeners which load up block (header) data: get headers and get blocks.

The former will use `getBlockHeaders` from the logic layer, which has the depth limit built in (as I showed above) so that's fine, no attacker can make the server fill up its heap beyond some bound in this way.

Maybe the get blocks listener is vulnerable? It uses `getBlockHeaders'`, a variant that takes two particular ends point hashes rather than a list of checkpoints. The server will get all hashes between these two points, so yes indeed we need a limit. Full logic layer uses `getHeadersRange Nothing`, so we have to change that. Is this what you meant?

avieth:
https://github.com/input-output-hk/cardano-sl/pull/2484

volhovm:
> Is this what you meant?

Yes, I'll fix it in this PR.

volhovm:
@avieth I've restored the behaviour with LCA on (2). Please review this PR again.

block/src/Pos/Block/Logic/Header.hs
avieth: It's confusing. Here I have some `m` that could be and probably is `IO`-capable. What does "throwing" mean exactly? It suggests an exception, as in a real true exception (the kind that only makes sense in `IO`) but it could also be a "normal" short-circuit.

Anyway, it's beyond the scope of this patch I guess. I think the true issue is the lack of good separation of pure and impure programs. If we had that, then there would be no ambiguity about what "throwing" means.

lib/src/Pos/Diffusion/Full/Block.hs
```diff
@@ -124,11 +120,27 @@ getBlocks logic enqueue nodeId tipHeader checkpoints = do
...
     -- the block itself.
     blocks <- if singleBlockHeader
-              then requestBlocks (NewestFirst (pure tipHeader))
-              else requestHeaders >>= requestBlocks
+              then requestBlocks (OldestFirst (one tipHeader))
+              else twoPhaseAction
     pure (OldestFirst (reverse (toList blocks)))
   where
 
+    twoPhaseAction = do
```
avieth: Needs a comment and a better name if you can think of one. How about `getAndClassifyHeaders`, then take out `requestBlocks` from the end and put it back up on line 124?

lib/src/Pos/Diffusion/Full/Block.hs
```diff
@@ -124,11 +120,27 @@ getBlocks logic enqueue nodeId tipHeader checkpoints = do
...
+        headers <- toOldestFirst <$> requestHeaders
+        getLcaMainChain logic headers >>= \case
```
avieth: How much work would it be to make something like `getLcaMainChain` that, instead of giving the lca hash, gives the suffix that we compute after this? So that we wouldn't have to `dropUntilLca` again? This could save a traversal and eliminate that error case on line 140.

lib/src/Pos/Logic/Full.hs
```diff
@@ -123,25 +117,22 @@ logicLayerFull jsonLogTx k = do
         getBlockHeader :: HeaderHash -> m (Maybe BlockHeader)
         getBlockHeader = DB.getHeader
 
+        getHashesRange
+            :: Maybe Word -- ^ Optional limit on how many to pull in.
+            -> HeaderHash
+            -> HeaderHash
+            -> m (OldestFirst NE HeaderHash)
+        getHashesRange = Block.getHashesRange
```
avieth: Good this is a better name.

block/src/Pos/Block/Logic/Header.hs
volhovm: @avieth this patch tries to fix problems that were missed in PR 2154: https://github.com/input-output-hk/cardano-sl/pull/2154/#discussion_r163384764

* Code is violating exceptions guideline
* It wasn't fixed, so now it is

My patch may be confusing, but it is only because exceptions handling is hard. In this particular place there is no good separation between request exceptions and everything else, pure vs impure, so I just follow the guideline written by @int-index. You're more than welcomed to read and argue with it, but for now -- standards first.

lib/src/Pos/Diffusion/Full/Block.hs
volhovm: Sure, I will.

lib/src/Pos/Diffusion/Full/Block.hs
volhovm: `dropUntilLca` usually does exactly 0 steps, because usually headers list is already `[c1..cn]`, where `c1` is a child of our tip `c0`, so it's a direct continuation. I don't see any need to optimize here, it's a _really_ rare case that this `dropUntilLca` will actually do something, because rollbacks are pretty rare. And if they are, it's anyway super quick and happens only once (there can't be more than one serious rollback).

block/src/Pos/Block/Logic/Header.hs
volhovm: Well, underlying traversal functions require me to put in some sane data, so this is a perfect place to check check its validity. `getHashesRange` is _actually_ a listener, and it's not used anywhere else, so the only difference between doing checks here and doing them outside is how do we handle errors. Here there's `GetHashesRangeException` already that will carry all the info about bad user request. If we do checks outside, we need yet another exception type. Also these checks are simple/fast.

volhovm:
@avieth Check out https://github.com/input-output-hk/cardano-sl/pull/2460/commits/fb50e112c8ed5cd256e936da252737388fa7e682

block/src/Pos/Block/Logic/Header.hs
avieth: Right I won't let it stand in the way of merging this, it's a separate issue. It's not really about exception handling guidelines, it's about when to use exceptions at all.

block/src/Pos/Block/Logic/Header.hs
avieth: > getHashesRange is actually a listener, and it's not used anywhere else, so the only difference between doing checks here and doing them outside is how do we handle errors

I disagree. Sure its only user at the moment is a listener, but it has no intrinsic relation to networking / listener / diffusion layer. One could argue it's a database component but even that is a bit dubious. `getHashesRange` is what its type says it is: given two hashes, get (somehow) a list of hashes. Doesn't matter where the inputs came from, doesn't matter where the output goes.

explorer/src/Pos/Explorer/Socket/Methods.hs
```diff
@@ -358,10 +359,10 @@ getBlundsFromTo
     :: forall ctx m . ExplorerMode ctx m
     => HeaderHash -> HeaderHash -> m (Maybe [Blund])
 getBlundsFromTo recentBlock oldBlock =
-    DB.getHeadersRange Nothing oldBlock recentBlock >>= \case
+    tryAny (DB.getHashesRange Nothing oldBlock recentBlock) >>= \case
```
avieth: Perhaps it would be better to not catch the exception? But right, `getHashesRange` is given a type that doesn't even admit the possibility of a normal failure. I'll write about that in a separate comment.

I take issue with this one because we're conflating two very different reasons for failure:

1. Something normal/expected, like not knowing one of the endpoints.
2. Something *actually* exceptional, like the database being corrupt or a failure to open some file it uses.

lib/src/Pos/Logic/Types.hs
```diff
@@ -27,53 +25,39 @@ import           Pos.Util.Chrono (NE, NewestFirst, OldestFirst)
...
-                         -> HeaderHash
-                         -> HeaderHash
-                         -> m (Either GetBlockHeadersError (OldestFirst NE HeaderHash))
-      -- Get the current tip of chain.
-      -- It's not in Maybe, as getBlock is, because really there should always
-      -- be a tip, whereas trying to get a block that isn't in the database is
-      -- normal.
+                         -> m (NewestFirst NE BlockHeader)
+      -- | Compute LCA with the main chain.
+    , getLcaMainChain    :: OldestFirst NE BlockHeader -> m (Maybe HeaderHash)
```
avieth: Notice that this type expresses the normal failure case: there's no LCA, so you get `Nothing`.

Compare at `getHashesRange`, which completely denies the normal failure case of not knowing one of the endpoints.

lib/src/Pos/Logic/Types.hs
```diff
@@ -27,53 +25,39 @@ import           Pos.Util.Chrono (NE, NewestFirst, OldestFirst)
...
     , getBlockHeader     :: HeaderHash -> m (Maybe BlockHeader)
-      -- Inspired by 'getHeadersFromManyTo'.
-      -- Included here because that function is quite complicated; it's not
-      -- clear whether it can be expressed simply in terms of getBlockHeader.:q
+      -- TODO CSL-2089 use conduits in this and the following methods
+      -- | Retrieve block header hashes from specified interval.
+    , getHashesRange     :: Maybe Word -- ^ Optional limit on how many to bring in.
+                         -> HeaderHash
+                         -> HeaderHash
+                         -> m (OldestFirst NE HeaderHash)
```
avieth: Change it to `Either GetHashesRangeError (OldestFirst NE HeaderHash)` or `Maybe (OldestFirst NE HeaderHash)`.

lib/src/Pos/Logic/Types.hs
```diff
@@ -27,53 +25,39 @@ import           Pos.Util.Chrono (NE, NewestFirst, OldestFirst)
...
-      -- or similar).
-    , getBlockHeaders'   :: Maybe Word -- ^ Optional limit on how many to bring in.
-                         -> HeaderHash
-                         -> HeaderHash
-                         -> m (Either GetBlockHeadersError (OldestFirst NE HeaderHash))
-      -- Get the current tip of chain.
-      -- It's not in Maybe, as getBlock is, because really there should always
-      -- be a tip, whereas trying to get a block that isn't in the database is
-      -- normal.
+                         -> m (NewestFirst NE BlockHeader)
```
avieth: Change it to `m (Either GetBlockHeadersError (NewestFirst NE BlockHeader))`.

block/src/Pos/Block/Logic/Header.hs
avieth: I take it back, this *is* a problem. Discussion here https://github.com/input-output-hk/cardano-sl/pull/2492

volhovm:
@avieth I've returned back `Either e res`, though `e` is now a datatype. Can you please review this again?

avieth:
> @avieth I've returned back Either e res, though e is now a datatype. Can you please review this again?

Ok looking now.

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -7,9 +7,12 @@ module Pos.Block.Logic.Header
        , classifyNewHeader
        , ClassifyHeadersRes (..)
        , classifyHeaders
+
+       , GetHeadersFromManyToError
```
avieth: Should export constructors too, no?

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -7,9 +7,12 @@ module Pos.Block.Logic.Header
...
        , getHeadersFromManyTo
        , getHeadersOlderExp
-       , getHeadersRange
+       , GetHashesRangeError
```
avieth: `(..)`

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -244,6 +247,9 @@ classifyHeaders inRecovery headers = do
                           depthDiff blkSecurityParam
             | otherwise -> CHsValid lcaChild
 
+
+data GetHeadersFromManyToError = GHFBadRequest Text deriving (Show,Generic)
```
avieth: Minor nit: networking language. `GHFBadInput`, `GHFUnknownHeader` or similar instead? Sure, this is *used* by the diffusion layer / networking but it's intended to be modular.

block/src/Pos/Block/Logic/Header.hs
```diff
@@ -354,29 +361,40 @@ getHeadersOlderExp upto = do
                 | otherwise = selGo es ii $ succ skipped
         in selGo elems ixs 0
 
+
+data GetHashesRangeError = GHRBadRequest Text deriving (Show,Generic)
```
avieth: Similar nit to the naming of `GHFBadRequest`.

block/src/Pos/GState/BlockExtra.hs
```diff
@@ -177,26 +175,6 @@ loadBlocksUpWhile
...
-    mBlk <- lift $ getBlock fromH
-    case mBlk of
-        Nothing  -> pure ()
-        Just blk -> yield blk >> continue blk
-  where
-    continue blk = do
-        mNextH <- lift $ resolveForwardLink blk
-        case mNextH of
-            Nothing    -> pure ()
-            Just nextH -> blocksSourceFrom nextH
```
avieth: @karknu relevant to what we were talking about earlier.

avieth: It's deleted here because it's not used yet, but you will probably want to add something similar. Pipe or conduit or similar, whatever you think is appropriate

explorer/src/Pos/Explorer/Socket/Methods.hs
```diff
@@ -358,10 +359,10 @@ getBlundsFromTo
     :: forall ctx m . ExplorerMode ctx m
     => HeaderHash -> HeaderHash -> m (Maybe [Blund])
 getBlundsFromTo recentBlock oldBlock =
-    DB.getHeadersRange Nothing oldBlock recentBlock >>= \case
+    tDB.getHashesRange Nothing oldBlock recentBlock >>= \case
```
avieth: Typo? `tDB`

lib/src/Pos/Diffusion/Full/Block.hs
```diff
@@ -124,11 +120,30 @@ getBlocks logic enqueue nodeId tipHeader checkpoints = do
...
+                -- Headers list is (oldest to newest)
+                -- [n1,n2,...nj,lca,nj+2,...nk] we drop [n1..lca] and
+                -- return [nj+2..nk], as we already have lca in our
+                -- local db. Usually this function does 1 iterations
+                -- as it's a common case that [n1..nj] is absent and
+                -- lca is the oldest header.
+                let dropUntilLca = NE.dropWhile (\h -> h ^. prevBlockL /= lca)
+                case nonEmpty (dropUntilLca $ getOldestFirst headers) of
+                    Nothing -> throwM $ DialogUnexpected $
+                                   "All headers are older than LCA, nothing to query"
```
avieth: I'm not demanding this change here but I just want to ask, do we agree that this would be better as an explicit `Left (DialogUnexpected ...)`? So that the type of `getBlocks` would end in `d (Either <error type> OldestFirst [] Block)`?

block/src/Pos/Block/Logic/Header.hs
volhovm: Yes, fixed. It wasn't intentional, `BadInput` is better.

lib/src/Pos/Diffusion/Full/Block.hs
volhovm: This should be a completely different story, I think. Currently, `retrievalWorker` expects an exception as a sign of "something went wrong", so we eventually throw even if we get `Left` somewhere. So at this point, no. But I'd took a look at exception handling in retrieval worker. 

The suggestion itself is something to think about. At a logic layer, I don't really care whether it was a "normal" exception like "somebody sent us trash" or "exceptional" like "my computer is on fire". But maybe I'm missing something.

block/src/Pos/GState/BlockExtra.hs
volhovm: Me? Or @karknu ? :)

I'll explain again: i have a vision^{tm} of how conduits should be structured. Namely, all traversals should be conduit. And when they are, the particular deleted function will become redundant. Also it wasn't used and it's always available in git history, just in case.

avieth: Nope, I meant @karknu . We were discussing block streaming earlier today.

volhovm:
@avieth https://github.com/input-output-hk/cardano-sl/pull/2460/commits/f4f926bc1d4c6390368fbc4f0df6aa60f826ab74

PR #2466, opened 2018-02-06
[CSL-2288] QuickCheck generation of blockchains

parsonsmatt:
This PR implements QuickCheck generators for the blockchain.

### DIFF
--- a/wallet-new/test/unit/UTxO/BlockGen.hs
+++ b/wallet-new/test/unit/UTxO/BlockGen.hs
@@ -5,7 +5,7 @@
 
 module UTxO.BlockGen where
 
-import           Universum hiding (use)
+import           Universum hiding (use, (.~))
 
 import           Control.Lens hiding (elements)
 import qualified Data.Map as Map
@@ -114,12 +114,14 @@ selectFromAddr =
 selectToAddress :: Addr -> BlockGen h Addr
 selectToAddress addr = do
     addrs <- uses bgcAddressesToBalances (toList . Set.delete addr . Map.keysSet)
-    liftGen (elements (filter notAvvm addrs))
-  where
-    notAvvm addr =
-        case addrActorIx addr of
-            IxAvvm _ -> False
-            _ -> True
+    liftGen (elements (filter (not . isAvvmAddr) addrs))
+
+-- | Returns true if the 'addrActorIx' is the 'IxAvvm' constructor.
+isAvvmAddr :: Addr -> Bool
+isAvvmAddr addr =
+    case addrActorIx addr of
+        IxAvvm _ -> True
+        _ -> False
 
 -- | Select a random address from the current set of addresses, and
 -- a random value between 0 and that address's total value minus the
@@ -127,16 +129,39 @@ selectToAddress addr = do
 -- the 'Addr's balance.
 selectFromAddrAndAmount :: BlockGen h (Addr, Value)
 selectFromAddrAndAmount = do
-    addrs <- uses bgcAddressesToBalances (toList . Map.keysSet)
-    addr <- liftGen $ elements addrs
-    mamount <- use (bgcAddressesToBalances . at addr)
-    let amount = fromMaybe
-            (error "selectFromAddrAndAmount: address somehow not in map")
-            mamount
-        -- TODO: make this real
-        maxFee = 100
-    value <- liftGen $ choose (1, amount - maxFee)
-    pure (addr, value)
+    addrs <- uses bgcAddressesToBalances Map.toList
+    (addr, amount) <- liftGen $ elements addrs
+    value <- liftGen $ choose (1, amount `safeSubtract` maxFee)
+    if value < 1
+        then localBlockGen
+            (bgcAddressesToBalances . at addr .~ Nothing)
+            selectFromAddrAndAmount
+        else pure (addr, value)
+
+maxFee :: Num a => a
+maxFee = 180000
+
+-- | Run a 'BlockGen' action with a modified context. Changes to the state
+-- from the inner action are discarded.
+localBlockGen
+    :: (BlockGenCtx h -> BlockGenCtx h)
+    -> BlockGen h a
+    -> BlockGen h a
+localBlockGen f action = do
+    st <- get
+    modify f
+    r <- action
+    put st
+    pure r
+
+-- | 'Value' is an alias for 'Word64', which underflows. This detects
+-- underflow and returns @0@ for underflowing values.
+safeSubtract :: Value -> Value -> Value
+safeSubtract x y
+    | z > x     = 0
+    | otherwise = z
+  where
+    z = x - y
 
 -- | Create a fresh transaction that depends on the fee provided to it.
 newTransaction :: Hash h Addr => BlockGen h (Value -> Transaction h Addr)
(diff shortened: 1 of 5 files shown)


edsko:
Nice work, getting there! :) The generated ledgers are currently not validating, I think there are two problems (well, possibly others also):

* Although you filter out AVVM addresses in `selectToAddress`, you still end up generating transactions that output to AVVM addresses because in `newTransaction` you return the _change_ back to the AVVM address. That too is also not allowed; that's why I wrote on slack that "_As a corollary, this means that this withdrawal then needs to withdraw the entire balance on the account_". 
* The validation errors about fees being zero or too low imply that the transaction is transferring too much money. Perhaps this is because the `maxFee` value is way too low. The actual fee values for the transactions you're generating at the moment are all around the `170,000` mark. 

Mind you, that 170k value is more than the 100k initial balance that the genesis AVVM accounts have. I'm not entirely sure what the story is there; perhaps the fee for a single transfer out of an AVVM account (with no change going anywhere) is less than 100k. I'd suggest to experiment a bit (`trace` the generated ledger to see what the actual fees are).

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -0,0 +1,245 @@
...
+        then localBlockGen
+            (bgcAddressesToBalances . at addr .~ Nothing)
+            selectFromAddrAndAmount
+        else pure (addr, value)
+
+maxFee :: Num a => a
+maxFee = 180000
+
+-- | Run a 'BlockGen' action with a modified context. Changes to the state
+-- from the inner action are discarded.
```
parsonsmatt: That I use this function makes me want to give a `MonadReader` instance to the `BlockGen` stack; then programming against `MonadReader (BlockGenCtx h)` would forbid you to run any state-changing functions inside of this. It'd need a somewhat gnarly constraint (`MonadReader (BlockGenCtx h) t, MonadTrans t) => t Gen a` maybe?) without refactoring the types a bit more.

 I recall reading a document in Cardano that strongly advised for programming against concrete monads instead of `mtl`-style constraints, though, so I kept it like this.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -0,0 +1,245 @@
...
+-- | 'Value' is an alias for 'Word64', which underflows. This detects
+-- underflow and returns @0@ for underflowing values.
+safeSubtract :: Value -> Value -> Value
+safeSubtract x y
+    | z > x     = 0
+    | otherwise = z
+  where
+    z = x - y
+
+-- | Create a fresh transaction that depends on the fee provided to it.
```
parsonsmatt: This function is the meat of the generator. The other generators just repeat this one some random amount of times.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -0,0 +1,245 @@
...
+        case createInputsAmountingTo fromAddr txns amount of
+            Nothing ->
+                error $ unlines
+                    [ "Somehow doesn't have enough balance? "
+                    , show amount
+                    , show fromAddr
+                    , show toAddr
+                    ]
+            Just (inputs, change) ->
+                pure (inputs, change)
```
parsonsmatt: I tried this with `let (inputs, change) = fromMaybe` and felt it was less readable. Happy to switch back to that if wanted.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -0,0 +1,250 @@
...
+-- the other settings in the 'BlockGenCtx', this function will initialize
+-- the generator and run the action provided.
+runBlockGenWith
+    :: Hash h Addr
+    => (BlockGenCtx h -> BlockGenCtx h)
+    -> Transaction h Addr
+    -> BlockGen h a
+    -> Gen a
+runBlockGenWith settings boot (BlockGen m) =
+    evalStateT m (settings (initializeCtx boot))
```
parsonsmatt: The `settings` attribute here will likely be useful when we expand to more comprehensive tests. We can start with a "seed" blockchain instead of an empty one and ensure that there's not some weird state in that seed that causes a bad transaction failure. Might be overly defensive though.

parsonsmatt:
OK, I've rebased against `develop` -- let me know what else needs to be done :smile:

edsko:
Great, making progress :) I've pushed a commit to your branch (I hope that's not rude..) that makes a few changes. Most of the changes are in the DSL itself, and make sure that we get a bit more informative error messages back when tests fail: the tests now show the full ledger that was used for the failing test, and when the ledger is considered invalid by the DSL validator we get some information back as to _why_ it was considered invalid. Primarily implemented this to try and figure out why some of the ledgers you were creating were considered invalid, but I'm sure it'll be helpful later on as well. 

Then I made one important change to the DSL validator: rather than checking that the sum of the outputs of a transaction is _equal_ to the sum of the inputs, we merely check that the sum of the outputs is _equal to or less than_ than the sum of the inputs. This is necessary because in your tests you are (rightly) conservative in the maximum fee. 

I made two minor bugfixes in your generator:

* The hash of the first transaction you generate should be 1, not 0 (since 0 is already used by the bootstrap transaction)
* I removed the calls to `reverse` in `toPreChainWith` (because `newChain` already returns them in oldest first order, even though `_bgcCurrentBlockchain` is stored in newest first order).

Then I made one more change: in `selectFromAddrAndAmount`, I filtered out the AVVM addresses, just like you already did in `selectToAddress`. The reason that this is necessary is what I already alluded to above: the initial balance of the AVVM accounts is less than the fee of the transaction to withdraw it :scream: I don't know why things were set up that way, but for now let's just avoid AVVM accounts. 

I made a few other minor modifications to avoid compiler warnings, and included an explicit export list so that we get warnings about dead code (in my initial attempt at filtering out the AVVM addresses I had modified `selectFromAddr`, but then realized that actually `selectFromAddr` doesn't get used at all). 

With these changes the tests now pass! :tada: 

@adinapoli-iohk @dcoutts Once @parsonsmatt OK's my changes here, I'm happy for this PR to be merged.

@parsonsmatt I think next step would be to gather a few statistics about the chains that are being generated: I'm thinking:

* how many blocks,
* how many transactions per block
* how many inputs per transaction (average, standard deviation, min, max)
* how many outputs per transaction (same)

Something along those lines. This would be useful to check how "interesting" the chains are that we are generating. Apart from that, move on to generators for inductive wallets :)

Nice work :)

parsonsmatt:
Thanks! I'm happy with the changes :smile: 

Some of the statistics to calculate are fairly easy to make configurable, as well -- the size of the generated chain + range of transactions per block are both random now, and we can incorporate fixed sizes or size ranges into the state configuration.

Likewise for "how many outputs," right now we're just doing a single output from one address to another + change; we could set it up so that you can have a number of outputs with the value split between some number of accounts.

edsko:
One thing I mentioned to say to you: in your block chain generation, you pick an _address and amount_ to transfer _from_, and then you pick _inputs_ to match the address and amount you chose. The way that you choose inputs to match the address/value is what we call "address selection", and its' a thorny problem in its own right, and we'll be exploring multiple ways on how to do it. What you do at the moment is alright, although we may wish to simplify the code here and just choose, first, how many inputs we want, and second, choose random inputs that belong to this user. That would then sidestep the address selection problem in the chain generation. 

@dcoutts , opinon on this?

parsonsmatt:
`buildkite` seems to be failing on the `cardano-sl-explorer` bit of things that's untouched by my code. Also a lot of `hlint` errors throughout the `wallet-new` bits.

I don't have a `buildkite` account though so I can't access it :open_mouth:

edsko:
Ok, after discussion with @parsonsmatt , this PR should not be merged yet.

parsonsmatt:
Closing this PR in favor of #2526

PR #2467, opened 2018-02-06
[CSL-1994] Move binaryTests to CborSpec

hithroc:


lib/test/Test/Pos/Cbor/CborSpec.hs
```diff
@@ -363,6 +389,270 @@ spec = withDefConfiguration $ do
...
+            binaryTest @Nano
+            binaryTest @Millisecond
+            binaryTest @Microsecond
+            binaryTest @Byte
+            binaryTest @(Map Int Int)
+            binaryTest @(HashMap Int Int)
+            binaryTest @(Set Int)
+            binaryTest @(HashSet Int)
+
+        describe "Types" $ do
```
gromakovsky: AFAIU, all these tests below are now inside `modifyMaxSuccess (const 1000)`, but previously they weren't. I think we shouldn't change these limits in this PR (or at least it should be done carefully, I am not sure it was).

lib/test/Test/Pos/Cbor/CborSpec.hs
```diff
@@ -363,6 +389,270 @@ spec = withDefConfiguration $ do
             pendingNoArbitrary "EdStandard.SecretKey"
             pendingNoArbitrary "EdStandard.Signature"
             pendingNoArbitrary "EncryptedSecretKey"
+        describe "Primitive instances" $ do
```
gromakovsky: And these tests were inside `modifyMaxSuccess (const 10000)`.

int-index:
Also change the base branch, user stories must have separate branches, not `develop`.

gromakovsky:
Imo, it should apply only to features, maybe bugfixes, but not refactorings. I talked about it many times, but didn't get a proper response. I will bring it up again.

gromakovsky:
I ran tests locally and I see that serialization tests for the following types are too slow:
> `MsgHeaders`, `CommitmentsMap`, `DataMsg MCShares`, `USUndo`

So you should decrease max success for them or use `SmallGenerator` wrapper. I don't mind leaving `maxSuccess = 1000` for other binary tests (they are relatively fast).

int-index:
I have realized that this goes directly against Duncan's suggestion:

> actual tests should live with the package they are testing

Reverting.

PR #2468, opened 2018-02-06
[CSL-1980] Add Slog documentation

Pastafarianist:
TODO:
- [x] Document pure verification

### DIFF
--- a/docs/block-processing/slog.md
+++ b/docs/block-processing/slog.md
@@ -16,7 +16,10 @@ Slog is responsible for the following (see explanations below):
     -   `inMainChain` flags
 -   Calling callbacks (`BListener`)
 
-Pure verification is *TODO DESCRIBE*
+Pure verification is a set of checks in `Pos.Block.Pure` that operate on
+arbitrary sequences of blocks and verify in a pure context whether these blocks
+constitute a valid chain. For more information, see Haddock comments in
+`Pos.Block.Pure`
 
 Genesis blocks contain no information that cannot be derived from preceding main
 blocks, but we still store them as physical entities for convenience. Here we
--- a/docs/block-processing/slog.md
+++ b/docs/block-processing/slog.md
@@ -1,4 +1,4 @@
-Slog is a part of block processing that does things which don't fit in any other part (other parts being `Toil` in `txp`, `Poll` in `update`, `Toss` in `ssc`, and `lrc`).
+Slog is a part of block processing that does things which don't fit in any other part (other parts being `Toil` in `txp`, `Poll` in `update`, `Toss` in `ssc`, `Cede` in `delegation` and `lrc`).
 
 ## What it does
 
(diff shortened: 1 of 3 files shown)


avieth:
Why is it called "slog"? Is this an acronym?

Pastafarianist:
@avieth please see [here](http://cardano-dev.askbot.com/question/6/what-do-poll-toss-toil-cede-slog-mean-in-the-code/).

docs/block-processing/slog.md
```diff
@@ -0,0 +1,68 @@
...
+    * Current tip
+    * `lastBlkSlots`, a list of slots of the last `blkSecurityParam` blocks
+    * Max seen difficulty
+    * Forward links
+    * `inMainChain` flags
+* Calling callbacks (`BListener`)
+
+Pure verification is *TODO DESCRIBE*
+
+Genesis blocks do not contain any information that cannot be derived from elsewhere, but we still store them as physical entities for convenience. Here we want to ensure that the list of leaders that is stored within genesis blocks is the same as the one computed by LRC based on the previous epoch.
```
volhovm: I know markdown doesn't care, but why don't you? I talk about line wrapping.

volhovm: >do not contain anything that cannot be derived from elsewhere

"can be derived from the chain state"

docs/block-processing/slog.md
```diff
@@ -0,0 +1,68 @@
...
+
+The list of slots of the last `blkSecurityParam` blocks (the last of which must be the tip) is an optimization for computing chain density.
+
+We maintain maximum seen difficulty to ensure that we never rollback deeper than `blkSecurityParam` from it.
```
volhovm: "Maximum seen difficulty" = "the highest difficulty ever seen in GState", not to be confused with "seen somewhere on the internet", right?

docs/block-processing/slog.md
```diff
@@ -0,0 +1,68 @@
...
+
+A forward link is a link from a block to its immediate child. They are currently only used in Explorer and Wallet.
```
volhovm: >They are currently only used in Explorer and Wallet.

Incorrect. Shouldn't be mentioned at all imho.

docs/block-processing/slog.md
```diff
@@ -0,0 +1,68 @@
...
+
+`inMainChain` is a boolean flag that is set whenever the block is an ancestor of the current tip. It is needed for computing LCA of forks with the main chain.
```
volhovm: _quickly_ computing lca. It's possible to compute it without this information, of course.

docs/block-processing/slog.md
```diff
@@ -0,0 +1,68 @@
...
+
+Finally, we often want to store some extra data whenever we apply or rollback a block, which is what `BListener` is all about.
```
volhovm: Obscure. We want to notify wallet/explorer about gstate changes.

docs/block-processing/slog.md
```diff
@@ -0,0 +1,68 @@
...
+
+## How it works
+
+Slog basically consists of 3 major functions: `slogVerifyBlocks`, `slogApplyBlocks`, and `slogRollbackBlocks`.
+
+`slogVerifyBlocks` takes a list of blocks and either throws or returns a list of `SlogUndo`s (for each block, a `SlogUndo` is the `SlotId` of the block `blkSecurityParam` blocks before it). `slogApplyBlocks` and `slogRollbackBlocks` take a list of blunds and return a list of `SomeBatchOp`s, updating the in-memory state themselves.
+
+### `slogVerifyBlocks`
```
volhovm: Why do you even write this instead of documenting code?

docs/block-processing/slog.md
```diff
@@ -0,0 +1,68 @@
...
+
+Note: it only supports verification of blocks from the same epoch.
```
volhovm: This is a universal property of all VAR logic. Verification is only done in batches.

docs/block-processing/slog.md
Pastafarianist: Ok. I'll be using `pandoc` for autoformatting, specifically

```
pandoc --from=markdown --to=markdown_github+yaml_metadata_block-hard_line_breaks \
            --columns=80 --atx-headers --normalize \
            slog.md -o slog.md
```

(Just a note for myself, mostly.)

Pastafarianist: > "can be derived from the chain state"
I believe "cannot" is the correct word here, since genesis blocks contain *only* information that *can* be derived from the chain state — thus, they contain *no* information that *cannot* be derived from the chain state.

docs/block-processing/slog.md
Pastafarianist: Corrected.

docs/block-processing/slog.md
pva701: >They are currently only used in Explorer and Wallet.

`loadHashesUpWhile` uses them, which are used by node.

>Shouldn't be mentioned at all imho.


same

Pastafarianist: Sorry, inattentive `grep`ping.

docs/block-processing/slog.md
Pastafarianist: Corrected.

docs/block-processing/slog.md
pva701: >A forward link is a link from a block to its immediate child.

not sure "child" is appropriate here, at least it should be said that child from the main chain.

docs/block-processing/slog.md
Pastafarianist: You're right, I'll move this to the code and add a link to the doc.

docs/block-processing/slog.md
Pastafarianist: Corrected.

docs/block-processing/slog.md
Pastafarianist: The latter doesn't contradict the former, does it?

volhovm: I mean that there's no need to explicitly state it, because it's not some kind of special effect of `slogVerifyBlocks`, but a general mechanism.

Pastafarianist: Well, "same epoch" ≠ "any batch". Is it true that all VAR logic operates on batches from the same epoch?

volhovm: Okay, again:

This is a universal property of all VAR logic. Verification is only done in (FIX: the same epoch) batches.

avieth:
> @avieth please see here.

Oh I didn't know this word. Thanks for the English lesson :)

Pastafarianist:
> It's better to add inline comments to every logical block of the function
These comments describe *what* the function does. If I make them inline, they will be interspersed with the comments that describe *how* the function works.

PR #2470, opened 2018-02-07
[CSL-2265] Remove 'fromE' in Pos.Script.Examples

int-index:


txp/Pos/Script.hs
```diff
@@ -154,6 +159,8 @@ stdlib = case PL.loadLibrary PL.emptyDeclContext prelude of
   where
     prelude = $(lift . toString =<< runIO PL.preludeString)
 
+deriving instance TH.Lift Script
```
gromakovsky: Why is it here and not in the place where `Script` is defined?

txp/Pos/Script/Examples.hs
```diff
@@ -230,3 +246,10 @@ sigStressRedeemer n = fromE $ parseRedeemer [text|
 
     keyS = sformat fullPublicKeyHexF pk
     sigS = sformat fullSignatureHexF sig
+
+    unwrap (Right a) = a
+    unwrap (Left s) =
+        -- This case is impossible because the input string for 'parseRedeemer'
+        -- is always a valid script regardless of arguments passed to
+        -- 'sigStressRedeemer'.
+        error ("sigStressRedeemer: impossible, failed to parse: " <> s)
```
gromakovsky: Why did you use `fromE` in some places and `unwrap` in other places?

gromakovsky: AFAIU, in all places we pass something valid to `parseRedeemer` or `parseValidator`.

int-index: The reason is that I'm using `liftE` to force compile-time evaluation, so that errors for `parseRedeemer` and `parseValidator` become compile-time errors.

For `multisigRedeemer`, `multisigValidator`, `sigStressRedeemer`, and `shaStressRedeemer` I cannot do that — these `Script`s are parametrized, and the parameters can be specified at run-time. For instance, for `multisigValidator`, the inputs are generated randomly in the test suite.

txp/Pos/Script.hs
neongreen: this is revenge for all my orphans

txp/Pos/Script/Examples.hs
gromakovsky: 👌

txp/Pos/Script.hs
int-index: I thought it was some type from `plutus-prototype` or whatever, I'll push a fix in a moment.

gromakovsky: If it was from `plutus-prototype`, I'd write something like this 😉 
https://i.imgur.com/ydqOOeD.png

neongreen:
approval granted

neongreen:
(garpoval pranted)

PR #2475, opened 2018-02-07
[CSL-2266] Remove BlockHeaderStub

int-index:
* remove `BlockHeaderStub` — it was an awkward hack to resolve module cycles, I used a more principled technique
* rewrite `BlockHeader` as an ADT — previously it was an `Either`, and I had plans to refactor it for a long time, but now I simply had to, as `BlockHeader` is now a data family (a type family wouldn't work because we need to declare instances for `Hash BlockHeader`)

### DIFF
--- a/core/Pos/Binary/Core/Blockchain.hs
+++ b/core/Pos/Binary/Core/Blockchain.hs
@@ -4,11 +4,16 @@ module Pos.Binary.Core.Blockchain
        (
        ) where
 
+import           Codec.CBOR.Decoding (decodeWordCanonical)
+import           Codec.CBOR.Encoding (encodeWord)
 import           Universum
 
-import           Pos.Binary.Class (Bi (..), encodeListLen, enforceSize)
+import           Pos.Binary.Class (Bi (..), decodeListLenCanonicalOf, encodeListLen, enforceSize)
+import           Pos.Binary.Core.Block ()
 import           Pos.Binary.Core.Common ()
 import qualified Pos.Core.Block.Blockchain as T
+import           Pos.Core.Block.Union.Types (BlockHeader (..), GenesisBlockchain, MainBlockchain)
+import           Pos.Core.Configuration (HasConfiguration)
 import           Pos.Crypto.Configuration (HasCryptoConfiguration, getProtocolMagic, protocolMagic)
 import           Pos.Util.Util (cborError, toCborError)
 
@@ -59,3 +64,26 @@ instance ( Typeable b
         body   <- decode
         extra  <- decode
         toCborError $ T.recreateGenericBlock header body extra
+
+----------------------------------------------------------------------------
+-- BlockHeader
+----------------------------------------------------------------------------
+
+instance ( HasConfiguration
+         , T.BlockchainHelpers MainBlockchain
+         , T.BlockchainHelpers GenesisBlockchain
+         ) =>
+         Bi BlockHeader where
+   encode x = encodeListLen 2 <> encodeWord tag <> body
+     where
+       (tag, body) = case x of
+         BlockHeaderGenesis bh -> (0, encode bh)
+         BlockHeaderMain bh    -> (1, encode bh)
+
+   decode = do
+       decodeListLenCanonicalOf 2
+       t <- decodeWordCanonical
+       case t of
+           0 -> BlockHeaderGenesis <$!> decode
+           1 -> BlockHeaderMain <$!> decode
+           _ -> cborError $ "decode@BlockHeader: unknown tag " <> show t
--- a/core/Pos/Core/Block/Union/Instances.hs
+++ b/core/Pos/Core/Block/Union/Instances.hs
@@ -14,8 +14,8 @@ import           Pos.Binary.Class (Bi)
 import           Pos.Core.Block.Blockchain (GenericBlock (..))
 import           Pos.Core.Block.Genesis ()
 import           Pos.Core.Block.Main ()
-import           Pos.Core.Block.Union.Types (Block, BlockHeader, ComponentBlock (..),
-                                             blockHeaderHash)
+import           Pos.Core.Block.Union.Types (Block, BlockHeader (..), ComponentBlock (..),
+                                             blockHeaderHash, choosingBlockHeader)
 import           Pos.Core.Class (HasDifficulty (..), HasEpochIndex (..), HasEpochOrSlot (..),
                                  HasHeaderHash (..), HasPrevBlock (..), IsHeader, IsMainHeader (..))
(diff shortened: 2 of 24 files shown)


core/Pos/Core/Common/Types.hs
```diff
@@ -214,14 +216,24 @@ newtype ChainDifficulty = ChainDifficulty
...
 
+----------------------------------------------------------------------------
+-- BlockHeader (forward-declaration)
+----------------------------------------------------------------------------
+
+-- We use a data family instead of a data type solely to avoid a module
+-- cycle.
+--
+-- | Forward-declaration of block headers. See the corresponding type instance
+-- for the actual definition.
```
neongreen: Either say in what module the instance is, or how to grep for it.

core/Pos/Core/Common/Types.hs
```diff
@@ -214,14 +216,24 @@ newtype ChainDifficulty = ChainDifficulty
...
+data family BlockHeader
+
 ----------------------------------------------------------------------------
 -- HeaderHash
 ----------------------------------------------------------------------------
 
 -- | 'Hash' of block header. This should be @Hash BlockHeader@
```
neongreen: Update the comment

lib/src/Pos/Communication/Limits.hs
```diff
@@ -310,6 +310,17 @@ instance (HasAdoptedBlockVersionData m, Applicative m) => MessageLimited MainBlo
...
     getMsgLenLimit _ = Limit . fromIntegral <$> maxBlockSize
 
+instance ( HasAdoptedBlockVersionData m
+         , Applicative m
+         )
+         => MessageLimited BlockHeader m where
+    getMsgLenLimit _ = f <$> getMsgLenLimit Proxy <*> getMsgLenLimit Proxy
+      where
+      maxLimit (Limit l1) (Limit l2) = Limit (max l1 l2)
+      f limA limB = 1 +
```
neongreen: Why `1`? The overhead is two bytes – one for the array, another for the tag. See [how `[1, ...]` is encoded](http://cbor.me/?diag=[1,%20%22Genesis%22]).

core/Pos/Binary/Core/Blockchain.hs
```diff
@@ -59,3 +64,26 @@ instance ( Typeable b
...
+     where
+       (tag, body) = case x of
+         BlockHeaderGenesis bh -> (0, encode bh)
+         BlockHeaderMain bh    -> (1, encode bh)
+
+   decode = do
+       decodeListLenCanonicalOf 2
+       t <- decodeWordCanonical
+       case t of
+           0 -> BlockHeaderGenesis <$!> decode
```
ChShersh: Why you used strict fmap here?

core/Pos/Core/Block/Union/Instances.hs
```diff
@@ -26,7 +26,9 @@ import           Pos.Core.Slotting.Types (EpochOrSlot (..))
 
 instance Bi BlockHeader =>
          Buildable BlockHeader where
-    build = either Buildable.build Buildable.build
+    build = \case
+        BlockHeaderGenesis bhg -> Buildable.build bhg
+        BlockHeaderMain    bhm -> Buildable.build bhm
```
ChShersh: Maybe there's sense in having `either`-like combinator for `BlockHeader`? Though, I don't know how to name it...

core/Pos/Core/Block/Union/Instances.hs
```diff
@@ -42,7 +44,9 @@ instance Bi BlockHeader =>
 
 -- | Take 'BlockHeader' from either 'GenesisBlock' or 'MainBlock'.
 getBlockHeader :: Block -> BlockHeader
-getBlockHeader = bimap _gbHeader _gbHeader
+getBlockHeader = \case
+    Left  gb -> BlockHeaderGenesis (_gbHeader gb)
```
ChShersh: Next step is to rewrite `Block` from `Either` to custom data type? 🙂

core/Pos/Core/Block/Union/Instances.hs
```diff
@@ -71,6 +82,9 @@ instance Bi BlockHeader => IsHeader BlockHeader
 -- HasPrevBlock
 ----------------------------------------------------------------------------
 
+instance HasPrevBlock BlockHeader where
+    prevBlockL = choosingBlockHeader prevBlockL prevBlockL
```
ChShersh: Is `choosingBlockHeader` is not like `either` for `BlockHeader`? If so, then why not use it everywhere?

core/Pos/Binary/Core/Blockchain.hs
int-index: Just preserving the semantics, the code was strict before my changes.

core/Pos/Core/Block/Union/Instances.hs
int-index: One option is `view $ choosingBlockHeader (to Buildable.build) (to Buildable.build)`, but I decided not to bother.

core/Pos/Core/Block/Union/Instances.hs
int-index: Yes :)

core/Pos/Core/Block/Union/Instances.hs
int-index: It's like `choosing` from `Lens`, a more powerful combinator.

lib/src/Pos/Communication/Limits.hs
int-index: I based this on the instance for `Either`:

```
instance ( MessageLimited a m
         , MessageLimited b m
         , Applicative m
         )
         => MessageLimited (Either a b) m where
    getMsgLenLimit _ = f <$> getMsgLenLimit Proxy <*> getMsgLenLimit Proxy
      where
      f limA limB = 1 + max (Left <$> limA) (Right <$> limB)
```

See `Pos.Communication.Limits.Types`. If this doesn't sound right to you, both instances should be fixed.

neongreen: cc @martoon what do you think?

int-index:
>  I wonder whether this solution could be extended to deal with things like
> type TxpBlock = Either (Some IsGenesisHeader) (Some IsMainHeader, TxPayload)

Not sure what you mean by extension here, but the type above should be written as:

```
data TxpBlock
    = TxpBlockGenesis (Some IsGenesisHeader)
    | TxpBlockMain (Some IsMainHeader) TxPayload
```

neongreen:
> Not sure what you mean by extension here

Having tags (so that the data family would have more than one instance).

PR #2476, opened 2018-02-07
[CSL-1994] Move Explorer BinarySpec to Cbor, use binaryTests instead of binaryEncodeDecode

hithroc:


int-index:
We don't need this.

PR #2478, opened 2018-02-08
[CSL-2268] Fix restart logic in 'retrievalWorker'

int-index:
Before the bugfix:

1. `retrievalWorker` starts, sets up an error handler `mainLoopE`
2. `mainLoop` starts
3. an error occurs in the `mainLoop`
4. the handler  `mainLoopE` catches it
5. `mainLoop` starts again, this time without an error handler
6. an error occurs in the `mainLoop`
7. thread dies

After the bugfix we restart the main loop with the error handler in place, so subsequent exceptions will be caught.


node/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -106,7 +107,7 @@ retrievalWorkerImpl SendActions {..} =
                 (_, Just (nodeId, rHeader))  ->
                     pure (handleRecoveryWithHandler nodeId rHeader)
         thingToDoNext
-        mainLoop
+        runIfNotShutdown mainLoopImpl -- next iteration
```
avieth: Do we need `runIfNotShutdown`? Can't we just `killThread` to stop this?

int-index:
> Does handleAll let async exception through?

Not in this branch, unfortunately.

> Do we need to back port it to develop?

In some form, yes.

> Do we need runIfNotShutdown? Can't we just killThread to stop this?

I'm not sure how exactly shutdown works, I placed this function where @georgeee suggested.

> Overall it's fine, and does count as a fix. But ideally, an exception would kill the retrieval worker, and by extension the thread which spawned it (the main thread), and by extension the whole program.

I see where you're coming from. Indeed, we could adopt this policy and use exceptions only for scenarios that should outright kill the entire app. The problem is that there are many libraries that throw exceptions in situations that should be considered normal, and this includes most common libraries (including `base`). For example, trying to open a file that doesn't exist will result in an exception.

And the real problem is that the exceptions thrown by other libraries aren't reflected in the types or anywhere, so it's easy to accidentally let some exception through and it will kill the entire node, when we actually would prefer the worker to restart.

So the usual solution to this is to let _async_ exceptions crash the app, but to use sync exceptions for "normal" scenarios since it's the approach used by so many libs.

node/src/Pos/Block/Network/Retrieval.hs
volhovm: Current shutdown machinery uses `runIfNotShutdown` approach. If it's going to be changed, it'll be the separate issue/PR. I think you can ask @gromakovsky about it.

gromakovsky: `runIfNotShutdown` is an extremely obsolete thing. It's dead for months, but this PR is to `master` branch (which is extremely old).

gromakovsky:
> I'm not sure how exactly shutdown works

In `master` branch shutdown doesn't work.

adinapoli-iohk:
This PR:

- Was created months ago, and now is lagging severely behind `develop`;
- It touches delicate part of the logic;
- Since then the wallet and the overall system seem to have stabilised.

For this reason, I'm inclined to follow the British's motto of "if isn't broken, do not fix it", and close this PR without merging it.

PR #2487, opened 2018-02-09
Facilitate testing with wallet-new

Martoon-00:
See commit descriptions for details.

I put those commits in one PR, because second commit doesn't help much without first one. And I suppose that @sectore would like to cherry-pick first commit to his branch, so it will appear in `develop` soon or later anyway. Feel free to reject this PR, I did it primarily for my convenience.

Tested with old and new wallets, both HTTP and HTTPS with and without --wallet-debug option.

### DIFF
--- a/lib/src/Pos/Web/Server.hs
+++ b/lib/src/Pos/Web/Server.hs
@@ -18,10 +18,11 @@ import           Control.Monad.Except (MonadError (throwError))
 import qualified Control.Monad.Reader as Mtl
 import           Data.Aeson.TH (defaultOptions, deriveToJSON)
 import           Data.Default (Default)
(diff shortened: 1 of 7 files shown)


adinapoli-iohk:
Not sure if this relevant anymore at this stage as I think most of the work was backported since then. @KtorZ , mind doing a crosscheck and declare if we can revive or close this one? Thanks! :)

KtorZ:
As far as I know, the `cardano-node` binary rely on the new wallet CLI and CLI options which already include a `--no-tls` flag. Nothing whatsoever seems to have been backported  yet to the legacy wallet CLI (at least on develop). Is this CLI still in use somehow?

adinapoli-iohk:
I think this is safe to close at this stage.

PR #2488, opened 2018-02-09
[CSL-2184] NTP fixes

Martoon-00:
Rebased #2445


adinapoli-iohk:
@Martoon-00 Isn't this a dupe of https://github.com/input-output-hk/cardano-sl/pull/2486 ?

Sorry for not claiming it in YT, but I didn't think somebody would have tackle it as it was discussed I was the owner. My bad 🙈

Martoon-00:
Np, if it's going to appear in release/1.1.0 in normal form, it would be better :+1:

PR #2490, opened 2018-02-09
[CSL-1855] Move exitWith to runServer, refactor lancher exit code handling

hithroc:
The node now exits with exit code 20 after a shutdown as it is supposed to.

Launcher exit code handling was refactored a little bit, we do not restart the node if it didn't exit code 20 or was forcefully killed.

### DIFF
--- a/lib/src/Pos/Launcher/Runner.hs
+++ b/lib/src/Pos/Launcher/Runner.hs
@@ -16,30 +16,33 @@ module Pos.Launcher.Runner
 
 import           Universum
 
-import qualified Control.Monad.Reader as Mtl
 import           Control.Monad.Fix (MonadFix)
+import qualified Control.Monad.Reader as Mtl
 import           Data.Default (Default)
 import           JsonLog (jsonLog)
+import           Mockable (race)
 import           Mockable.Production (Production (..))
+import           System.Exit (ExitCode (..))
 
 import           Pos.Binary ()
 import           Pos.Communication (ActionSpec (..), OutSpecs (..))
 import           Pos.Communication.Limits (HasAdoptedBlockVersionData)
 import           Pos.Configuration (networkConnectionTimeout)
 import           Pos.Context.Context (NodeContext (..))
+import           Pos.Diffusion.Full (diffusionLayerFull)
+import           Pos.Diffusion.Full.Types (DiffusionWorkMode)
+import           Pos.Diffusion.Transport.TCP (bracketTransportTCP)
+import           Pos.Diffusion.Types (Diffusion (..), DiffusionLayer (..))
 import           Pos.Launcher.Configuration (HasConfigurations)
 import           Pos.Launcher.Param (BaseParams (..), LoggingParams (..), NodeParams (..))
 import           Pos.Launcher.Resource (NodeResources (..))
-import           Pos.Diffusion.Transport.TCP (bracketTransportTCP)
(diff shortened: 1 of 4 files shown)


lib/src/Pos/Launcher/Runner.hs
```diff
@@ -111,19 +114,20 @@ elimRealMode NodeResources {..} action = do
...
+       , HasShutdownContext ctx
        , MonadFix m
        )
     => NodeParams
     -> EkgNodeMetrics m
     -> OutSpecs
     -> ActionSpec m t
     -> m t
-runServer NodeParams {..} ekgNodeMetrics _ (ActionSpec act) =
+runServer NodeParams {..} ekgNodeMetrics _ (ActionSpec act) = const (exitWith (ExitFailure 20)) =<< (race waitForShutdown $
```
gromakovsky: I think it would be a bit more understandable (and the line would be not so big) if you moved this logic into a separate function in `where` clause (e. g. `exitOnShutdown action = const (exitWith (ExitFailure 20)) =<< (race waitForShutdown action)`).

tools/src/launcher/Main.hs
```diff
@@ -428,34 +428,36 @@ clientScenario ndbp logConf node wallet updater nodeTimeout report walletLog = d
...
+                 ("Waiting for the node to shut down or killing it in "%int%" seconds otherwise")
+                 nodeTimeout
+             nodeExitCode <- liftIO $ timeout (fromIntegral nodeTimeout * 1000000) (wait nodeAsync)
+             return (exitCode, nodeExitCode)
        | otherwise -> do
              logWarning $ sformat ("The wallet has exited with "%shown) exitCode
              -- TODO: does the wallet have some kind of log?
-             killNode nodeHandle nodeAsync
+             return (exitCode, Nothing)
+    when (walletExitCode == ExitFailure 20) $
```
gromakovsky: We also want to kill node if wallet exits with code other than 20.

hithroc:
Updated `ui-simulator.sh` to call `/api/update/apply` before exiting to shutdown the node to match the behavior of an actual wallet.

tools/src/launcher/Main.hs
```diff
@@ -428,34 +428,36 @@ clientScenario ndbp logConf node wallet updater nodeTimeout report walletLog = d
...
+                 nodeTimeout
+             nodeExitCode <- liftIO $ timeout (fromIntegral nodeTimeout * 1000000) (wait nodeAsync)
+             return (exitCode, nodeExitCode)
        | otherwise -> do
              logWarning $ sformat ("The wallet has exited with "%shown) exitCode
              -- TODO: does the wallet have some kind of log?
-             killNode nodeHandle nodeAsync
+             return (exitCode, Nothing)
+    when (isNothing nodeExitCode) $ do
+        logWarning "The wallet has exited exited, but the node is still up."
```
hithroc: Oops, gonna fix that typo tomorrow.

lib/src/Pos/Launcher/Runner.hs
```diff
@@ -136,17 +140,20 @@ runServer NodeParams {..} ekgNodeMetrics _ (ActionSpec act) =
                     maybeWithStatsd $
                     act (diffusion diffusionLayer)
   where
+    exitOnShutdown action = do
+        _ <- race waitForShutdown action
+        exitWith (ExitFailure 20) -- special exit code to indicate an update
```
gromakovsky: Looks much better now 👍

PR #2492, opened 2018-02-09
Inline comments to the exceptions guideline

avieth:
RFC @int-index @volhovm 

Exceptions have come up in that other PR https://github.com/input-output-hk/cardano-sl/pull/2460 so I did a read-through of the guidelines and found some pretty controversial things in there.

Summary: there's a big misunderstanding about what's normal and what's an exception. This definition of exception ("regular error" in the document) is way too vast and must be reined in before more damage is done by citing this document. Relevant thematic example: `getHashesRange :: HeaderHash -> HeaderHash -> m [HeaderHash]` is a bad type, because it sweeps under the rug every normal case in which a `[HeaderHash]` can't be computed (one or both endpoints unknown, or neither endpoint an ancestor of the other).

### DIFF
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -6,6 +6,9 @@ answer these questions:
 1. Is this a programmer mistake or a regular error?
 2. Is the code 100% pure or potentially impure?
 
+*Comment:* why regular error? What about a non-regular error? I propose simpler
+terminology to help things along: bug or exception.
+
 Depending on the answers, an appropriate approach can be selected.
 
 ## Is this a programmer mistake or a regular error?
@@ -50,6 +53,11 @@ It is always a tradeoff whether to allow programmer mistakes, or do type-level
 trickery to avoid them. The decision process for this is out of scope of this
 document.
 
+*Comment:* calling `zipSameLen` with lists of different length is a mistake
+by the programmer which manifests as a bug. This has nothing to do with
+exception handling. A program should never "catch" this bug and recover from
+it. A program with this bug built-in should crash.
+
 ### Regular error
 
 Consider a function like `readFile`. It is entirely possible that the path
@@ -61,8 +69,17 @@ scenarios.
 Another example is parsing user input. We might expect the user to enter a
 number, but we have to consider the case that the user enters something else.
 
+*Comment:* a user giving input which does not parse isn't an error or an
+exception, it's a completely normal thing, well-described by the typical
+`Either ParseError` type. What we're dealing with here is a partial function
+from the input type to the type of the parsed thing, which we make total via the
+sum type. This has nothing at all to do with exception handling.
+
 In cases when the erroneous scenarios are out of our control, we consider
 these to be regular errors.
+
+*Comment:* this is a good definition of "exception". In cases when the scenario
+is out of our control, we consider these to be exceptions.
 
 ### Identifying error class
 
@@ -79,6 +96,9 @@ verification (more precise types?)
 The code is considered impure when it's an `IO` action, a function that returns
 an `IO` action, or similar. The code can be _potentially_ impure when it is
 written in an abstract monad that can be instantiated to `IO`.
+
+*Comment:* bad definition. `foo = return 42` can be instantiated to `IO Int`
+but it can never do any effects.
 
 Definitely impure:
 
@@ -137,10 +157,16 @@ you're not aware of. Always make this trade-off conciously and responsibly.
 (On the other hand, if static guarantees require GADTs or type families, perhaps
 it's better to not overcomplicate code. Use your judgement.)
 
+*Comment:* this is a good section. But I'd remove the first bullet point from
+"Do not". The advice should be "try to avoid this, if possible" but that's
+already conveyed by the rest of the section.
+
 DISCUSSION: Should we create a synonym `bug = impureThrow` in Universum? This
 would make the intention more clear.
 
 ### Pure code, regular errors
+
+*Comment:* there are no regular errors (exceptions) in pure programs.
 
 Do *not*:
 
@@ -155,6 +181,11 @@ Do:
 * use `MonadError` or `MonadThrow` (methods of these classes). Note
   that if you define `f :: MonadError m => m ()`, it won't be pure
 
+*Comment:* I disagree with the last "Do not". Why is this ruled out? Sometimes
+giving `Either Text` is legit. Giving `Maybe` is ok, so it must also ok to
+give `Either ()`, so why not `Either Text`, which contains even more
+information?
+
 Consider parsing: it is pure, but we cannot make assumptions about the input. In
 this case we might want to use `ExceptT ParseError`. Or consider a lookup in a
 `Map`, where we don't know whether the key is present -- in this case we'd like
@@ -164,12 +195,28 @@ to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
 * `MaybeT`, `ExceptT e`
 * `CatchT`
 
+*Comment:* neither of these examples are of errors or exceptions! They are
+examples of normal values in the images of pure functions. The `Map` does not
+make any guarantee to that any particular key is in the map! Nor does the parser
+guarantee that every input will be recognized! These facts are reflected in the
+types. This has *nothing at all* to do with errors or exceptions. We need to
+strike this section out.
+
 Avoid using `Text` with the error message in place of `e` -- create a
 proper ADT. In case creating a proper ADT feels too cumbersome, use
 `CatchT`, which is equivalent to `ExceptT SomeException`. Note,
 however, that using `SomeException` in pure code is not the best
 practice, because the set of all possible exceptions is statically
 known. Use it only if you are lazy to define yet another ADT.
+
+*Comment:* nor is overusing exceptions the best practice, not the least because
+they do not appear in types, so we're essentially throwing away one of the
+strongest features of Haskell. Instead of conjuring up a new exception type and
+throwing it in the case of a completely normal condition, think about the
+contract that your function gives to its caller. Does the caller have a
+guarantee that this header hash picks out a block in the database? Of course
+not! So don't promise to give back a block for every input. That's not
+exceptional, that's not an error, that's a new point in the image type.
 
 Be careful not to use `MaybeT`, `ExceptT`, and `CatchT` in potentially impure
 code. When in doubt whether the code is potentially impure, use `MonadThrow`.
@@ -192,12 +239,20 @@ Do:
 * create a custom exception type
 * use `throwM` (`MonadThrow`)
 
+*Comment:* the last bullet in "Do not" is a silly rule and should be struck.
+If the `Left` case is a normal outcome, then leave it be! Making it into an
+exception and throwing it is unambiguously worse. The type is less descriptive,
+and the normal case is mixed in with every truly exceptional case. It's simply
+a lot harder for a caller to use the function properly.
+
 If you want to return `m (Either e a)` from a function, it's
 recommended to define `instance TypeError "NOT AN EXC" => Exception e`
 for your type `e`. If the meaning of `e` type is not related to
 exceptional situations at all, it's not needed to define such
 instance. But for example if `e` denotes a `ParseError`, please do
 define it.
+
+*Comment:* why? What's the value of this?
 
 We disallow the use of `throwIO` only because it is redundant in the presence of
 `throwM` and requires a stronger constraint (`MonadIO` rather than
@@ -217,6 +272,9 @@ two reasons for this:
 * when in an abstract (but potentially impure) monad, using `throwM` might
   add an additional constraint
 
+*Comment:* why should there be a distinction between a bug in a pure program
+and a bug in an effectful one?
+
 ## Packages and modules
 
 Do not import `Control.Exception` or `Control.Monad.Catch`! We use the
@@ -229,6 +287,12 @@ Use `bracket` or to guarantee the release of resources. In case of concurrent
 code, avoid `forkIO` or `forkProcess` in favor of the `async` package, as it
 rethrows exceptions from the child threads. (Do not use the function `async`
 itself when you can use `withAsync`, `race`, or `concurrently`).
+
+*Comment:* and when `async` is the only option, be aware of `link` and use it
+if appropriate.
+
+*Comment:* beware useing `withAsync` recursively. It essentially does a bracket,
+so you can't get tail recursion.
 
 When resource usage is non-linear, it's okay to use `ResourceT`, but
 prefer `bracket` whenever possible. Non-linear resource usage is


docs/exceptions.md
```diff
@@ -155,6 +181,11 @@ Do:
 * use `MonadError` or `MonadThrow` (methods of these classes). Note
   that if you define `f :: MonadError m => m ()`, it won't be pure
 
+*Comment:* I disagree with the last "Do not". Why is this ruled out? Sometimes
+giving `Either Text` is legit. Giving `Maybe` is ok, so it must also ok to
+give `Either ()`, so why not `Either Text`, which contains even more
+information?
+
 Consider parsing: it is pure, but we cannot make assumptions about the input. In
```
avieth: We *do* make assumptions about the input, namely its type. We do *not* assume that the input is recognised by the parser. But of course we don't! We'd have to define the parser in order to do that.

docs/exceptions.md
```diff
@@ -155,6 +181,11 @@ Do:
 * use `MonadError` or `MonadThrow` (methods of these classes). Note
   that if you define `f :: MonadError m => m ()`, it won't be pure
 
+*Comment:* I disagree with the last "Do not". Why is this ruled out? Sometimes
+giving `Either Text` is legit. Giving `Maybe` is ok, so it must also ok to
+give `Either ()`, so why not `Either Text`, which contains even more
```
int-index: The point is that if you're going to give more information, do it via a dedicated AST, not via `Text`, because it will simplify the life of anyone who might ever need to catch this error.

docs/exceptions.md
int-index: > We do make assumptions about the input, namely its type.

We don't make any assumptions about the input, the input is whatever passed to our programs, and the type we choose to represent it (`String`, `Text`, `ByteString`, `Vector Char`, or whatever) isn't an assumption, it's an engineering decision.

docs/exceptions.md
```diff
@@ -164,13 +195,29 @@ to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
 * `MaybeT`, `ExceptT e`
 * `CatchT`
 
+*Comment:* neither of these examples are of errors or exceptions! They are
+examples of normal values in the images of pure functions. The `Map` does not
+make any guarantee to that any particular key is in the map! Nor does the parser
+guarantee that every input will be recognized! These facts are reflected in the
+types. This has *nothing at all* to do with errors or exceptions. We need to
```
int-index: I'd say that "parse errors" do have something to do with "errors", they are called "parse errors" for a reason. And they might as well have something to do with "exceptions", because it'd be valid to represent them as `ExceptT`, which has got `Except` in its name. (IO exceptions would work too, but parsers are normally pure).

docs/exceptions.md
```diff
@@ -164,13 +195,29 @@ to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
...
 practice, because the set of all possible exceptions is statically
 known. Use it only if you are lazy to define yet another ADT.
 
+*Comment:* nor is overusing exceptions the best practice, not the least because
+they do not appear in types, so we're essentially throwing away one of the
+strongest features of Haskell. Instead of conjuring up a new exception type and
+throwing it in the case of a completely normal condition, think about the
+contract that your function gives to its caller. Does the caller have a
+guarantee that this header hash picks out a block in the database? Of course
+not! So don't promise to give back a block for every input. That's not
```
int-index: A type signature like `IO HeaderHash` communicates clearly that the header hash might not be in the database, because `IO` might throw, that's how `IO` works. In case you're worried that the possible exceptions are not listed in the type, well, I don't like it either, but we have to live with that or use a different base monad rather than IO (for example, https://hackage.haskell.org/package/unexceptionalio)

The issue is that in `ExceptT HeaderNotFound IO Header` we really have two ways to throw: `throwIO HeaderNotFound` and `throwError HeaderNotFound`. So at call site we must account for both exit paths when catching.

docs/exceptions.md
```diff
@@ -192,13 +239,21 @@ Do:
 * create a custom exception type
 * use `throwM` (`MonadThrow`)
 
+*Comment:* the last bullet in "Do not" is a silly rule and should be struck.
+If the `Left` case is a normal outcome, then leave it be! Making it into an
+exception and throwing it is unambiguously worse. The type is less descriptive,
+and the normal case is mixed in with every truly exceptional case. It's simply
+a lot harder for a caller to use the function properly.
```
int-index: It is easier to define and use this function properly:

* the caller does not have to account for two ways of exception throwing
* the implementor does not have the freedom to make a mistake and use `throwM` or `throwIO` instead of `throwError`

The assumption you seem to make is that `IO` exceptions aren't used for regular errors, this implies we never want to catch and handle them. That's simply not how other libraries are structured, even `base` throws exceptions for normal outcomes (see `readFile`), we have to live with that and the guidelines are written accordingly.

If we used some cunning technique to represent all regular errors differently, I wouldn't mind (I already cited `unexceptional-io`), but at the moment we don't.

docs/exceptions.md
```diff
@@ -192,13 +239,21 @@ Do:
...
+
 If you want to return `m (Either e a)` from a function, it's
 recommended to define `instance TypeError "NOT AN EXC" => Exception e`
 for your type `e`. If the meaning of `e` type is not related to
 exceptional situations at all, it's not needed to define such
 instance. But for example if `e` denotes a `ParseError`, please do
 define it.
 
+*Comment:* why? What's the value of this?
```
int-index: The value of adding the `instance TypeError "NOT AN EXC" => Exception e` is that it disallows statically throwing `e` as an `IO` exception. It means that in the following type signature:

```
IO (Either e a)
```

I do not have to handle both the `e` returned in `Either` and the `e` that might have been thrown with `throwIO`.

docs/exceptions.md
```diff
@@ -217,6 +272,9 @@ two reasons for this:
 * when in an abstract (but potentially impure) monad, using `throwM` might
   add an additional constraint
 
+*Comment:* why should there be a distinction between a bug in a pure program
```
int-index: There isn't a distinction between a bug in a pure program and in an effectful one, that's why the guidelines recommend to use the same method to represent both. But in impure code we have other techniques available (`throwM`, `throwIO`), so the document explicitly states not to use them to represent bugs, that's why we have separate sections for bugs in pure and effectful code.

int-index:
@avieth Thanks for writing these comments. Most of the disagreement appears to be on terminology, but I think you will find the guidelines sensible if you try to interpret the material as it laid out rather than by using your own definitions.

The actual disagreement seems to be whether to represent regular errors as IO exceptions, and here the choice is unfortunately already made for us: other libraries do that, and we have to live with that. The alternative is to come up with some technique to convert all normal outcomes represented as IO exceptions to the way we prefer — this would cause severe friction with the Haskell ecosystem, but I can see the long-term value of doing that. This would mean that any `IO` exception is meant to crash the app, and caught only to run some finalizers (`bracket`), not to be handled. Is that what you propose, or did I misinterpret your position?

docs/exceptions.md
parsonsmatt: to illustrate, 
```haskell
case someValue of
  Left "this error happened" -> ...
```
is not a good way to handle this

docs/exceptions.md
```diff
@@ -164,13 +195,29 @@ to return `Maybe v`. In 100% pure code, use one of these ways to handle errors:
 * `MaybeT`, `ExceptT e`
 * `CatchT`
 
+*Comment:* neither of these examples are of errors or exceptions! They are
+examples of normal values in the images of pure functions. The `Map` does not
+make any guarantee to that any particular key is in the map! Nor does the parser
+guarantee that every input will be recognized! These facts are reflected in the
+types. This has *nothing at all* to do with errors or exceptions. We need to
+strike this section out.
```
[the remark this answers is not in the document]
int-index: As the primary author of the document I want to make something clear: `lookup :: k -> Map k v -> Maybe v` is the *right* type for this operation, in full compliance with the guidelines. What does it have to do with Python exceptions?

docs/exceptions.md
avieth: It lies between two points:

```Haskell
case soemValue of
  Left MyExpressiveErrorConstructor -> ...
```

and

```Haskell
case someValue of
    Nothing -> ...
```

Why are these 2 sanctioned, but using `Text` is not?

avieth: For the record: I agree that using a custom type is better than `Text`, I just think we're looking at dogma here.

docs/exceptions.md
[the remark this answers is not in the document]
parsonsmatt: What about `Database.lookup :: DbConn -> k -> IO ???`? I would think that `??? ~ Maybe a`. Even though you *can* throw an exception in `IO` for the missing record, it seems like less of an exceptional situation that eg the database connection being refused, or the network connection screwing up. 

The type `IO (Either DatabaseError (Maybe a))` would be cumbersome, because we're generally just going to want to rethrow the `DatabaseError` when we're dealing with code that wants to handle `a`.

I think the way I have approached this in  the past is to consider: "if this operation could be done purely, in some absurd way, how would I type it?" For a database, we can assume we have some `Datatabase` type, that is the entire database that we're passing around. The resulting type is `lookup :: Database -> k -> Maybe a`. When we add impurity, we get `lookup :: DatabaseConn -> k -> IO (Maybe a)`. 

If I were to reimagine `readFile`, I'd say: "Imagine we have the entire file system in memory", giving us `readFile :: FileSystem -> FileName -> Maybe FileContents`. Impurifying it, we get `readFile :: FileName -> IO (Maybe FileContents)`.  This handles the exceedingly common "file not found" error, while hiding details like "no more RAM available", "no more open file handles available," etc.

docs/exceptions.md
parsonsmatt: `Nothing` implies "there is exactly one way this could have failed," and that's IMO a suitable type for something like `Map.lookup`

`Either` implies "there are multiple ways this could fail, *and* you want to be able to handle all the possible failures." That suggests to me that you should document these possible failure modes you're expecting, and allow programmers to handle them.

docs/exceptions.md
avieth: No, it's an assumption too. If we choose `Text`, we know we don't have to handle a `True` as input.

docs/exceptions.md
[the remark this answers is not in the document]
avieth: > As the primary author of the document I want to make something clear: lookup :: k -> Map k v -> Maybe v is the right type for this operation, in full compliance with the guidelines. We use Maybe to represent this unwanted codepath, classified as "regular error" by the guidelines. What does it have to do with Python exceptions?

My point is that this has nothing at all to do with exceptions and so shouldn't even be mentioned in the document. There's no "unwanted codepath" here.

docs/exceptions.md
```diff
@@ -6,6 +6,9 @@ answer these questions:
 1. Is this a programmer mistake or a regular error?
 2. Is the code 100% pure or potentially impure?
 
+*Comment:* why regular error? What about a non-regular error? I propose simpler
+terminology to help things along: bug or exception.
```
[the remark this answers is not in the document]
avieth: I don't propose to call exceptions bugs, they surely are not. How an exception is or is not dealt with could be a bug, but "exception" is within the domain of the language, whereas "bug" is a meta thing. I propose to call programmer mistakes bugs but this is a minor point.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: @parsonsmatt I'm fine with having `IO (Either DatabaseError (Maybe a))` as long as `DatabaseError` doesn't have an `Exception` instance. If it has an `Exception` instance, it means that it might be both thrown as `IO` exception and returned in `Either`, and it's a headache at each call site. I'm also opposed to `ExceptT DatabaseError IO a` because it means we have two exit paths from the monad (`throwM` and `throwError`) and there's no formal criteria which one to use for what.

docs/exceptions.md
```diff
@@ -50,6 +53,11 @@ It is always a tradeoff whether to allow programmer mistakes, or do type-level
 trickery to avoid them. The decision process for this is out of scope of this
 document.
 
+*Comment:* calling `zipSameLen` with lists of different length is a mistake
```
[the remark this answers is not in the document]
avieth: My point is on lines 57/58: this has nothing to do with exception handling (or exceptions in general).

docs/exceptions.md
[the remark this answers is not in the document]
parsonsmatt: That makes sense to me. Thanks for clarifying!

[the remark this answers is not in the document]
int-index: @avieth Exceptions *are* unwanted codepaths. "Exception" can be read as "exceptional situation". You're using a different definition, apparently, but I already asked you to treat the document on its own terms, because otherwise this discussion is unproductive.

docs/exceptions.md
```diff
@@ -61,9 +69,18 @@ scenarios.
 Another example is parsing user input. We might expect the user to enter a
 number, but we have to consider the case that the user enters something else.
 
+*Comment:* a user giving input which does not parse isn't an error or an
+exception, it's a completely normal thing, well-described by the typical
+`Either ParseError` type. What we're dealing with here is a partial function
+from the input type to the type of the parsed thing, which we make total via the
+sum type. This has nothing at all to do with exception handling.
```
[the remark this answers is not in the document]
avieth: > Call it what you want — exception handling, error handling, "completely normal thing" handling, but it's still a secondary codepath that is normally implemented with ExceptT, CPS-d ExceptT, IO exceptions, or similar. This is the sort of thing that this document is concerned with.

What we call it is important, apparently. I object to how liberal the characterisation of "exception" or "regular error" is within this document. It makes truly normal things seem convoluted when they're not. And what, exactly, is a "secondary codepath"?

docs/exceptions.md
```diff
@@ -80,6 +97,9 @@ The code is considered impure when it's an `IO` action, a function that returns
 an `IO` action, or similar. The code can be _potentially_ impure when it is
 written in an abstract monad that can be instantiated to `IO`.
 
+*Comment:* bad definition. `foo = return 42` can be instantiated to `IO Int`
```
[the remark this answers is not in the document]
avieth: I'd say some applicative term `x` is pure iff

```
fmap (const ()) x == pure ()
```

and impure otherwise.

For instance, every `Reader r t` term is pure, but not every `State s t`.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: > "exception" is within the domain of the language

"Exception" is a term generally used in programming, it's not necessarily within the domain of the language. Even in the context of Haskell it is an overloaded term that can either mean (1) "IO exception" or (2) "imprecise exception" (bottom in pure code), which are technically different.

And "regular error" here is a more broad thing than either  "IO exception" or "imprecise exception".

docs/exceptions.md
avieth: > Nothing implies "there is exactly one way this could have failed," and that's IMO a suitable type for something like Map.lookup

Careful: the lookup didn't fail. It succeeded. Nothing went wrong. There was no key at that value.

> Either implies "there are multiple ways this could fail, and you want to be able to handle all the possible failures." That suggests to me that you should document these possible failure modes you're expecting, and allow programmers to handle them.

Does it? What if there's precisely one way it could fail, and it comes with a `Text`?

docs/exceptions.md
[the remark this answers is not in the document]
int-index: According to your definition of exception.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: > What we call it is important, apparently

Well, yes, given that most of your criticism is based on a misinterpretation of what's being said.

> And what, exactly, is a "secondary codepath"?

Do you expect me to come up with a formal definition? I think you have an intuitive understanding of what this means, it's something that is secondary to the logic of the computation and interrupts its normal execution. We usually represent these with `ExceptT` or `IO` because that's the easiest way to interrupt the computation and reach the handler for it.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: This looks plausible. Mind submitting a separate PR with this improvement?

avieth:
> Most of the disagreement appears to be on terminology, but I think you will find the guidelines sensible if you try to interpret the material as it is laid out rather than by using your own definitions.

Yes the disagreement is in meaning, and it's important. There's heavy baggage behind "error" and "exception", and we must not let these terms envelop normal functional program definitions. The examples of map lookup and parsing demonstrate this: neither the map lookup nor the parser make a value judgement of the outcome, they just give it. `Nothing` from a lookup is not a good thing or a bad thing, a primary or a secondary codepath, it's just a value. Nothing wen't wrong, we got the value.

> The actual disagreement seems to be whether to represent regular errors as IO exceptions

No, it's whether to treat normal, non-errors as exceptions.

docs/exceptions.md
int-index: `True` as a Haskell isn't something a user can enter, parsing input normally comes from a file or from REPL, so the input is necessarily a sequence of characters. We could also run a lexer as a preprocessing pass, then the input would be a sequence of tokens. Or we could be dealing with a binary format, in which case the input would be a sequence of bytes.

In any case, we're not making any assumptions about the input to the program.

int-index:
> Nothing from a lookup is not a good thing or a bad thing, a primary or a secondary codepath

This heavily depends on context. Let's say I want to look up certain things in the map, but if any isn't present, I can't really proceed with my computation:

```
do
  a <- Map.lookup "thing1" m
  b <- Map.lookup "thing2" m
  c <- doSomethingWith a b
  e <- Map.lookup "thing3" m
  doSomethingElseWith e c
```

It's clear here in which sense `Nothing` creates a secondary codepath — we interrupt the computation if any of the `Map.lookup` fail (that is, return `Nothing`).

docs/exceptions.md
int-index: > Careful: the lookup didn't fail. It succeeded. Nothing went wrong. There was no key at that value.

If the goal of the lookup is to retrieve the value, not having the value in the map is a failure. Of course, this failure is represented as a normal value, `Nothing`, and it's not something that would crash our program, it's a different sort of failure.

avieth:
> This heavily depends on context.

Granted. But the actual implementation of `lookup` should make no decision about it. This is relevant to ongoing review of the logic layer. It shouldn't decide whether for instance, giving a header hash that's not in the database is a secondary or primary code path, a bad or a good thing.

int-index:
> Granted. But the actual implementation of lookup should make no decision about it. This is relevant to ongoing review of the logic layer. It shouldn't decide whether for instance, giving a header hash that's not in the database is a secondary or primary code path

The current type is:

```
getHashesRange ::
        forall m. (HasConfiguration, MonadDBRead m)
     => Maybe Word
     -> HeaderHash
     -> HeaderHash
    -> m (Either Text (OldestFirst NE HeaderHash))
```

There are two violations of the guidelines here:

* the returned error is represented as `Text`
* the implementation uses `ExceptT` on top of `IO`

We can address these by creating an ADT and not using `ExceptT`. If you want to keep `ExceptT`, this is evidence that this codepath is secondary. If you think that this scenario is not secondary, then it shouldn't be a problem to remove `ExceptT` — in this case I see why @volhovm's decision to represent this failure as an `IO` exception seems suboptimal.

I approved @volhovm changes three days ago when reviewing that PR because his change followed the guidelines, but I did mention the alternative option, quoting myself:

> This code was using throwError, but it wasn't pure, so it was rightly rewritten. If you disagree with the use of throwM here, the alternative is to return Either, but without runExceptT which creates a secondary exit path from any part of the code.

docs/exceptions.md
volhovm: >What if there's precisely one way it could fail, and it comes with a Text?

I guess that's a bad interface design then. Document your `foo :: a -> Maybe b` so that it'd be clear: if you get `Nothing` then "exactly X went wrong".

docs/exceptions.md
volhovm: @avieth I think that you're doing more work than it is truly optimal. We need to agree on the definitions first :)

"error" means "something that is not default", so "parse errors" are "errors" just because it's too long to call them "completely normal execution paths that didn't return the parse result though". No finding the key in hashmap is completely normal too, though it's called a "lookup fail" or "lookup error" because "not finding the key in hashmap" is too long.

docs/exceptions.md
```diff
@@ -230,6 +288,12 @@ code, avoid `forkIO` or `forkProcess` in favor of the `async` package, as it
 rethrows exceptions from the child threads. (Do not use the function `async`
 itself when you can use `withAsync`, `race`, or `concurrently`).
 
+*Comment:* and when `async` is the only option, be aware of `link` and use it
+if appropriate.
+
+*Comment:* beware useing `withAsync` recursively. It essentially does a bracket,
```
volhovm: Doesn't have to do anything with the exceptions/errors guideline.

docs/exceptions.md
avieth: > In any case, we're not making any assumptions about the input to the program.

But you just admitted that we do make assumptions.

> the input is necessarily a sequence of characters

docs/exceptions.md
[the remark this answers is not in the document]
avieth: Not *unwanted*, just *exceptional*. Exceptions are not "code paths", they're terms in the object language. Further, a parse error, or a map lookup in which the key is not present, are neither unwanted nor exceptional. 

> I already asked you to treat the document on its own terms, because otherwise this discussion is unproductive.

I have objections to the document itself. It's not immune to criticism. And if it's too be taken as some sort of law that can be cited in other pull request discussions, then it's very important that it be criticised.

docs/exceptions.md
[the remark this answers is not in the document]
avieth: > According to your definition of exception.

Yes. It's more faithful to reality.

docs/exceptions.md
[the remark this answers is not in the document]
avieth: > Do you expect me to come up with a formal definition?

No, I'm challenging the document. "Secondary codepath" is vague and dubious, and I really don't expect a formal definition. Maybe I have an intuitive understanding of what it means, but does it match up with yours and everybody else's? Probably not.

Dealing with a `Nothing` from a map lookup is not a "secondary" thing, it's exactly as important as the  case branch for `Just`. Neither branch should be considered "secondary". Another hole to poke: what is "normal execution"? Surely, in a pure function, *everything* in its definition is "normal", no?

docs/exceptions.md
int-index: I propose the following clarification:

```
- but we cannot make assumptions about the input
+ but we cannot make assumptions about the input string
```

docs/exceptions.md
```diff
@@ -137,11 +157,17 @@ you're not aware of. Always make this trade-off conciously and responsibly.
...
+*Comment:* this is a good section. But I'd remove the first bullet point from
+"Do not". The advice should be "try to avoid this, if possible" but that's
+already conveyed by the rest of the section.
+
 DISCUSSION: Should we create a synonym `bug = impureThrow` in Universum? This
 would make the intention more clear.
 
 ### Pure code, regular errors
 
+*Comment:* there are no regular errors (exceptions) in pure programs.
```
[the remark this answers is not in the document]
avieth: The "impure examples" section is good, although I want to insist again that "regular error" should be struck from the record and "exception" used instead.

The "pure examples", however, are incorrect, unhelpful, and serve to confuse. Again, *nothing went wrong* in either of these cases, and exception throwing/catching support from the language is not required in order to deal with them.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: I am confused, are your objections to the meaning, the form, or both? Let's tackle these separately, because otherwise the discussion of meaning drowns in a sea of nitpicks to form. So if you have better phrasing in mind, that's one thing, and if you want the sentence to say something completely different, it's another. What is it, then?

docs/exceptions.md
avieth: > If the goal of the lookup is to retrieve the value, not having the value in the map is a failure. Of course, this failure is represented as a normal value, Nothing, and it's not something that would crash our program, it's a different sort of failure.

There's no "goal" of lookup. What's the "goal" of `(+) 1`? What's the "goal" of an integral? What's the "goal" of the expected value of a random variable? These are mathematical functions. They don't have objectives, they just *are*.

avieth: > I guess that's a bad interface design then. Document your foo :: a -> Maybe b so that it'd be clear: if you get Nothing then "exactly X went wrong".

And if I have a function which returns `Either Text Foo`, where the `Left txt` does *not* subjectively reflect an "unwanted" "code path"? Is that OK?

docs/exceptions.md
avieth: > "parse errors" are "errors" just because it's too long to call them "completely normal execution paths that didn't return the parse result though"

You don't have to call them that. How about "the image of `parse x`". A parse failure is just as much in the image of `parse` as a parse success. 

> I think that you're doing more work than it is truly optimal. We need to agree on the definitions first :)

Sure, I'm showing where/how/why I disagree with the definitions.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: > Yes. It's more faithful to reality.

To your perception of reality. This is leading us nowhere, my understanding is that "exception" is a heavily overloaded term. You find it used with one meaning (an exceptional situation) and then insist that the meaning is different (a term in Haskell with an `Exception` instance, I guess?). Let's be constructive here. How should this document be named, if we keep the current contents?

* exception handling guidelines
* error handling guidelines
* ...

The goal of the document is to help the programmer decide how to structure his/her code when he/she needs to do something vaguely related to bugs/exceptions/errors/throwing/catching/etc. If we narrow it down to dealing with IO exceptions only, for example, it will no longer serve its purpose.

docs/exceptions.md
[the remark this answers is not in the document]
int-index: We already covered this: https://github.com/input-output-hk/cardano-sl/pull/2492#issuecomment-364615107

docs/exceptions.md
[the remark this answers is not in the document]
int-index: We already covered this: https://github.com/input-output-hk/cardano-sl/pull/2492#issuecomment-364615107

docs/exceptions.md
avieth: > A type signature like IO HeaderHash communicates clearly that the header hash might not be in the database, because IO might throw, that's how IO works.

This type signature communicates that you can *always* get a `HeaderHash` using `IO`, unless there was an exception. Whereas, `HeaderHash -> IO (Either Oops Header)` communicates that even if the `IO` went smoothly with no exceptions, there still may not be a `Header` for that hash. The latter is what we want, when looking up a header hash.

> the possible exceptions are not listed in the type, well, I don't like it either, but we have to live with that or use a different base monad rather than IO 

No, there's a very simple solution: `HeaderHash -> IO (Either Oops Header)` where `Oops` includes all of the non-exceptional reasons why the `Header` could not be produced.

docs/exceptions.md
avieth: > the caller does not have to account for two ways of exception throwing

I'm not asking for 2 ways of exception throwing. I'm asking that we don't throw "exceptions" that aren't actually exceptional. Maximize purity. Not every `Left` or `Nothing` should be treated like an exception or a failure.

> the implementor does not have the freedom to make a mistake and use throwM or throwIO instead of throwError

And the implementor wouldn't do that anyway.

> The assumption you seem to make is that IO exceptions aren't used for regular errors, this implies we never want to catch and handle them. 

I don't understand what you mean. Isn't a "regular error" the same as an "exception"? Please, "regular error" has to go. My assumption is that `IO` exceptions are (sometimes) used to express exceptions. My advice is that `IO` exceptions should be used sparingly.

docs/exceptions.md
avieth: Sure it does. It's a caveat for the advice about `withAsync`. It can't be used in some cases, due to its properties with respect to exception handling.

avieth:
> the implementation uses ExceptT on top of IO

No, this type doesn't make any mention of `ExceptT`, nor even of `IO`.

> but I did mention the alternative option, quoting myself:
>
> > This code was using throwError, but it wasn't pure, so it was rightly rewritten. If you disagree with the use of throwM here, the alternative is to return Either, but without runExceptT which creates a secondary exit path from any part of the code.

It's not secondary. No `throwError` or `throwM` should be used in case one of the input hashes couldn't be found in the database. Such a scenario is not exceptional nor "secondary" nor "unwanted". The type *must* express reality: that not every pair of `HeaderHash`es determines a list of intermediate `HeaderHash`es, even when there are no `IO` exceptions.

docs/exceptions.md
int-index: > There's no "goal" of lookup. What's the "goal" of (+) 1? What's the "goal" of an integral? What's the "goal" of the expected value of a random variable? These are mathematical functions. They don't have objectives, they just are.

Programmers have objectives when they write code. The code itself doesn't have any goals of course, but it reflects the goals the programmer had. Maybe this is a bad choice of words on my part, I should've said "purpose", not "goal". 

> And if I have a function which returns Either Text Foo, where the Left txt does not subjectively reflect an "unwanted" "code path"? Is that OK?

This is ok because

* `Text` has no `Exception` instance (this holds)
* `Text` is not an error message of any sort (otherwise I'd say it does reflect an "unwanted" "code path")
* the use of `Either` is symmetric here, so if we do `Either Foo Text`, nothing will change

But assume we use the `Monad` instance for `Either` that gives a special treatment to the `Left` constructor, using it to short-circuit the computation with a "failure". In this case using `Text` wouldn't be OK.

docs/exceptions.md
avieth: Sure, but why make this observation at all? We can't assume that the input string is recognised by our parser, but that's why we're parsing it: we want to determine whether it's recognised by our parser. Would you define

```Haskell
-- We can't assume the input is 42, so we throw an exception if it isn't.
is42 :: Int -> ExceptT Not42 ()`
is42 42 = pure ()
is42 _ = throwError Not42
```

Can we agree that this is a silly thing to write?

docs/exceptions.md
int-index: >  includes all of the non-exceptional reasons 

* how do you decide which reasons are exceptional and which ones aren't?
* how do you deal with other libraries using `IO` exceptions to represent non-exceptional reasons something couldn't be done, including core libraries?

I'm ok with `Either Oops Header` as long as

* `Oops` doesn't have an `Exception` instance
* we're not using `ExceptT` internally to `throwError` this `Oops`, because `ExceptT` is for exceptional cases

docs/exceptions.md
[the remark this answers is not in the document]
avieth: Objections to meaning and to what it prescribes. Some definitions are vague and ultimately unhelpful (unwanted, secondary) and the prescriptions favour throwing exceptions too liberally.

docs/exceptions.md
int-index: > I don't understand what you mean. Isn't a "regular error" the same as an "exception"? Please, "regular error" has to go. 

The document gives a decision procedure for whether an error should be classified as a regular error (not necessarily an exception) or a programmer mistake (bug).

>  I'm asking that we don't throw "exceptions" that aren't actually exceptional. Maximize purity. 

I assume something can be classified as exceptional when we never want to `catch` it to affect the logic of our app, only to release some resources and rethrow (`bracket`). Based on this understanding, I have only one objection:

* existing libraries don't follow this principle, at all

If you're using a different definition of "exceptional", then it's not clear to me what you mean.

docs/exceptions.md
avieth: > how do you decide which reasons are exceptional and which ones aren't?

It's a good question. Imagine a perfect world where the definition could be given without any `IO`/side effects. What would the type be? 

> how do you deal with other libraries using IO exceptions to represent non-exceptional reasons something couldn't be done, including core libraries?

This would be dealt with internally. If you need to catch exceptions, catch them.

> we're not using ExceptT internally to throwError this Oops, because ExceptT is for exceptional cases

What does the caller care whether I use `ExceptT` internally? Is this property even observable by the caller? I don't think so. I guess this is just a code style thing?

int-index:
> No, this type doesn't make any mention of ExceptT, nor even of IO.

Since this was in the context of the logic layer review, I was talking about the implementation of the function with this signature, not about the signature itself. Of course, for the signature itself the condition is that the left case does not have the `Exception` instance. As to IO, as long as the polymorphic monad can be instantiated to something with `IO` as the base monad, this reasoning applies.

> It's not secondary. No throwError or throwM should be used in case one of the input hashes couldn't be found in the database. 

I remember seeing `throwError` in the implementation. Removing it will resolve the problem.

docs/exceptions.md
int-index: > We can't assume that the input string is recognised by our parser, but that's why we're parsing it: we want to determine whether it's recognised by our parser.

No, that's not why we're parsing it. We're parsing it because we want to extract data from it. We can assume it is recognized by the parser: if our nodes communicate via JSON, the bytes sent over the network must adhere to the JSON grammar, otherwise it's an error (or an attack, but clearly not normal operation).

avieth: > We're parsing it because we want to extract data from it. 

Right, in order to do so, it must be well-formed, recognised by our parser, so that we can map it to the structure we want.

> We can assume it is recognized by the parser: if our nodes communicate via JSON, the bytes sent over the network must adhere to the JSON grammar, otherwise it's an error (or an attack, but clearly not normal operation).

Ok let's keep with this example of a particular application of a parser. What is the advantage of throwing an exception on unrecognised input (impure), versus returning a value (pure)?

docs/exceptions.md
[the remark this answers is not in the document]
avieth: > To your perception of reality.

I meant objective reality.

> my understanding is that "exception" is a heavily overloaded term.

It is.

> The purpose of the document is to help the programmer decide how to structure his/her code when he/she needs to do something vaguely related to bugs/exceptions/errors/throwing/catching/etc. If we narrow it down to dealing with IO exceptions only, for example, it will no longer serve its purpose.

`IO` exceptions are the real deal, though. They are substantively different from pure "exceptions":

```Haskell
throwIO e >> pure () = _|_

(throwError e :: ExceptT e Identity x) >> pure () /= _|_
```

If you're writing a pure function, you don't throw exceptions. You ought to be allowed to use `ExceptT`, because "throwing an exception" in this monad is actually pure, despite the name. No user of your pure function will try to catch exceptions when using it, because it's a pure function. It doesn't throw exceptions, it just is.

When a pure partial function like `zipSameLen` blows up due to `error`, it's not an exception, it's a bug. The programmer misused it, but could have avoided doing so. Contrast with `readFile`, in which the programmer could not possibly rule out disk errors, file permission issues, etc.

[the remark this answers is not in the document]
int-index: > `throwIO e >> pure () = _|_`

Well, no. 

```
Prelude Control.Exception> (throwIO (ErrorCall "err") >> pure ()) `seq` pure ()
```

`throwIO` is a non-bottom operation.

docs/exceptions.md
int-index: > Ok let's keep with this example of a particular application of a parser. What is the advantage of throwing an exception on unrecognised input (impure), versus returning a value (pure)?

Disadvantages of exceptions as values are described in "A Semantics for Imprecise Exceptions", section 2.2, "Inadequacies of exceptions as values". Screenshot attached

image

int-index: Of course, with parsers we often need to backtrack on failure, and we don't make the assumption that the input adheres to the grammar, and parsing is a pure operation, so the best option is `ExceptT`. This is why the document refers to parsing as an example where we want to represent errors as values, not throw them.

docs/exceptions.md
int-index: > This would be dealt with internally. If you need to catch exceptions, catch them.

But where do I catch them, every time I call some library function? There are really two options here:

* catch exceptions that represent normal scenarios every time we use a function from an external library; this basically leads us to having a ton of wrappers for every library function we might want to use

* catch exceptions when we want to handle them; this means that we end up with a mix of error handling techniques in our code, where we catch both `IO` exceptions and `ExceptT`-exceptions

Neither is satisfactory.

int-index: > I guess this is just a code style thing?

We discussed this with @gromakovsky. We're going to amend the guidelines to allow `ExceptT E` over `IO` (or an abstract monad) as long as the `E` type has an instance that disallows throwing it as an `IO` exception:

```
instance TypeError "Not an exception" => Exception E
```

int-index: @avieth https://github.com/input-output-hk/cardano-sl/pull/2507

Does this look better? You can keep your usage of `ExceptT`, just replace `Text` with an ADT.

PR #2493, opened 2018-02-10
CBR-100 Reliable shutdown: improvements 

gromakovsky:
The branch from this PR was used as a base branch for all tasks in CBR-100.


georgeee:
Most improvements on CBR-100 are implemented, story will be deferred for a while. Strongly advise to merge this PR

lib/src/Pos/Launcher/Scenario.hs
```diff
@@ -89,15 +87,10 @@ runNode' NodeResources {..} workers' plugins' = ActionSpec $ \diffusion -> ntpCh
...
-    -- Either all the plugins are cancelled in the case one of them
-    -- throws an error, or otherwise when the shutdown signal comes,
-    -- they are killed automatically.
-    void
-      (race
-           (void (mapConcurrently (unpackPlugin) $ workers' ++ plugins'))
-           waitForShutdown)
+    void (mapConcurrently (unpackPlugin) $ workers' ++ plugins')
+
+    exitFailure
```
volhovm: Why is it `exitFailure`?

gromakovsky: Currently we expect at least some workers and plugins to work indefinitely, so this `exitFailure` shouldn't be reachable.
If we reach this `exitFailure`, it likely indicates that something went wrong (because it means that all workers and plugins stopped and not because of exceptions).
Maybe it makes more sense to `reportFatalError` (with explanation) in this case (which will eventually lead to `exitFailure` anyway).

volhovm: Got it. Please add some comment to this line, because it's not obvious that we use `exitOnShutdown` outside.

gromakovsky: @hithroc I am going to merge this PR, but please make sure @volhovm is happy (in another PR to `develop`) 😉

PR #2500, opened 2018-02-12
[CSL-2274] Extend error table with error code and example

akegalj:
Issue: https://iohk.myjetbrains.com/youtrack/issue/CSL-2274

look


wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -88,6 +83,25 @@ instance Arbitrary WalletError where
...
+--  * consider using genExample from V1/Swagger.hs
+allErrorsExamples :: [WalletError]
+allErrorsExamples = map genExample allErrorsConstructors
+  where genExample e = (unGen (resize 3 e)) (mkQCGen 42) 42
+
+-- | List of all existing error constructors
+allErrorsConstructors :: [Gen WalletError]
+allErrorsConstructors =
+    [ NotEnoughMoney <$> arbitrary
+    , OutputIsRedeem <$> pure "address"
```
adinapoli-iohk: Why not putting a real Cardano address here?

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -88,6 +83,25 @@ instance Arbitrary WalletError where
...
+    , SomeOtherError <$> pure "blah" <*> arbitrary
+    , MigrationFailed <$> pure "migration"
```
adinapoli-iohk: This is an internal error, arguably we should expose this as an `Arbitrary` but not as an `Example`.

akegalj: Can you be more specific here - as I am not sure I have understood what you meant ?

In the example above `MigrationFailed` is an `Arbitrary` instance and is "generated" from arbitrary, although there is nothing to generate in the example above (as everything is predefined).

If I understood you correctly you would like to have `Arbitrary` instance of `WalletError` and also `Example` instance of it - and use `Example` in this swagger use case instead of arbitrary ? (I don't think this is what you meant though :/ )

adinapoli-iohk: @akegalj Apologies for misleading you!

What I meant is that `allErrorConstructors` , in its current shape, doesn't really fit our needs for documentation purposes, because:

* It's not automatically generated. This means that every time you add a new `WalletError` constructor you have to remember to add it to `allErrorConstructors`, which is error prone.

* It reveals internal errors we might not want to expose. Case in point `MigrationFailed` is an internal error that we might (or might not) want to exclude from a top-level handler as it leaks an internal implementation detail, not to mention it's a temporary error which will stay here only for the take it takes to implement the new wallet.

Solving this properly would mean writing a function using generic programming to automatically derive those `[Gen WalletError]` also taking into account filter rules (i.e. do not expose this error), but I can see that this is tracked by CSL-2275, so I'm happy with the fact this PR is not addressing this just yet.

**TL;DR** In order to get this PR merged:

* Let's reference CSL-2275 in the `TODO` comment of `allErrorConstructors`, so that's clear we are tracking it.

* Let's leave the `MigrationFailed` error in `allErrorConstructors` for now, but let's remove `SomeOtherError` as that is just a placeholder, no need for users to see that.

* If easy (like a 10 minute job) let's use `encodePretty` from `aeson-pretty` to nicely format our JSON-rendered `WalletError`.

* Let's not tackle CSL-2275 up until post-MVP, as that could be a time sink.

akegalj: Got it!

Yes, it makes sense to do everything like you said. Also, I did this in a hurry (half automated) so that we can fully automate it in CSL-2275 - as you noticed.

akegalj:
@adinapoli-iohk could you approve so I can squash merge.

KtorZ:
I am closing this one in favor of the upcoming PR for CSL-2277.

PR #2503, opened 2018-02-12
[CBR-53] Refactor exception handling

gromakovsky:


adinapoli-iohk:
Considering the severe conflicts this PR accumulated and that @erikd and his squad may want to do things differently in the future, I'm inclined to close this PR unmerged and re-open only if really needed.

PR #2511, opened 2018-02-13
[CSL-2276] Add a CLI flag for dumping runtime configuration

gromakovsky:
It's needed for testing.


gromakovsky:
Yes (before my last commit, but that one is so small that I don't thing it can break something and I will test it anyway very soon, since I am assigned to the related test case).

PR #2512, opened 2018-02-13
[DEVOPS-652] remove all usage of import-from-derivation in release.nix when nix 1.12 is present

cleverca22:
this has been tested on both nixpkgs `61fbdb47a69f78998f55207e64122a0798047b5d`

and also the `5c09cdc187b014143302551e62d4973b5bf52814` where the problem originally occured

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -94,7 +94,7 @@ let
       args: import ./scripts/launch/connect-to-cluster (args // walletConfig // { inherit gitrev; });
   other = rec {
     mkDocker = { environment, connectArgs ? {} }: import ./docker.nix { inherit environment connect gitrev pkgs connectArgs; };
-    stack2nix = import (pkgs.fetchFromGitHub {
+    stack2nix = import (localLib.fetchFromGitHub {
       owner = "input-output-hk";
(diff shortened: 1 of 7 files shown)


cleverca22:
and the usage of `builtins.fetchTarball` requires nix-daemon to be running version 1.12 to solve
```
imported archive of ‘/nix/store/pwyqgvdswd4yaffsdsg09vv99a2swx45-61fbdb47a69f78998f55207e64122a0798047b5d.tar.gz’ lacks a signature
```

domenkozar:
I'm not super keen on this solution. We could just bail out of hydra-eval-jobset and use nix-instantiate and switch to Nix 2.0 fully once it's ready.

domenkozar:
The overhead of supporting both will bring too much overhead. I think we'll have to make the transition big-bang once it's out there.

domenkozar:
We decided to use older nixpkgs for now to get hydra - matching Hydra in nixpkgs.

akegalj:
@cleverca22 can this be closed?

jcmincke:

PR #2515, opened 2018-02-14
[CSL-2173] Explain some 'error's

volhovm:


crypto/Pos/Crypto/Signing/Types/Safe.hs
```diff
@@ -70,7 +70,11 @@ data SafeSigner = SafeSigner EncryptedSecretKey PassPhrase
 -- | Parameters used to evaluate hash of passphrase.
 passScryptParam :: S.ScryptParams
 passScryptParam =
-    -- TODO [CSL-2173]: Clarify
+    -- This returns valid if parameters are not valid, but 'scrypt'
+    -- doesn't define "valid" in docs. Though, default is valid, and
+    -- for bufLen the only requirement is to be positive and less
```
int-index: ✔️

delegation/src/Pos/Arbitrary/Delegation.hs
```diff
@@ -19,8 +19,7 @@ import           Pos.Util.Util (leftToPanic)
 
 genDlgPayload :: HasConfiguration => EpochIndex -> Gen DlgPayload
 genDlgPayload epoch =
-    -- TODO [CSL-2173]: Clarify
-    leftToPanic "genDlgPayload: " .
+    leftToPanic "genDlgPayload: generated payload is broken: " .
```
int-index: How do we know this can't happen?

delegation/src/Pos/Delegation/DB.hs
```diff
@@ -150,8 +150,7 @@ data DelegationOp
 instance HasConfiguration => RocksBatchOp DelegationOp where
     toBatchOp (PskFromEdgeAction (DlgEdgeAdd psk))
         | isRevokePsk psk =
-          -- TODO [CSL-2173]: Clarify
-          error $ "RocksBatchOp DelegationOp: malformed " <>
+          error $ "RocksBatchOp DelegationOp: " <>
```
int-index: How do we know this can't happen?

delegation/src/Pos/Delegation/Logic/Mempool.hs
```diff
@@ -55,7 +55,7 @@ getDlgMempool
 getDlgMempool = do
     sks <- runDelegationStateAction $ uses dwProxySKPool HM.elems
     pure $
-        -- TODO [CSL-2173]: Clarify
+        -- Mempool must be a consistent dlg payload.
```
int-index: How do we know that it is? Describe the preconditions, etc.

delegation/src/Pos/Delegation/Logic/Mempool.hs
```diff
@@ -173,10 +173,12 @@ processProxySKHeavyInternal psk = do
         -- by the database.
     let _cmHasPostedThisEpoch = mempty
     let cedeModifier = CedeModifier {..}
+    -- verificationError is only something sane if pskValid is false.
+    -- 'Maybe' could be used here, but it'll make if-case in the end less beautiful.
```
int-index: I believe you refer to this line:

```
not pskValid -> PHInvalid verificationError
```

What if we keep the result as `Either` and do

```
(Left verErr <- verResult) -> PHInvalid verErr
```
This simple change will allow us to remove:

* the `error` call
* the comment that explains why this `error` call is safe
* `fmap`, `either`, `(, False)`

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -190,22 +190,27 @@ calculateTransCorrections eActions = do
...
-                fromMaybe (error $ "transChangeset shouldn't happen but happened: " <> pretty k) .
+                fromMaybe (error $ "transChangeset.lookupUnsafe failed: " <> pretty k) .
                 HM.lookup k
             toTheseUnsafe :: StakeholderId
                           -> (Maybe StakeholderId, Maybe StakeholderId)
                           -> These StakeholderId StakeholderId
             toTheseUnsafe a = \case
                 (Nothing,Nothing) ->
-                    -- TODO [CSL-2173]: Clarify
+                    -- Algorithm can't come up with (N,N) because it means
```
int-index: ✔️

delegation/src/Pos/Delegation/Logic/VAR.hs
```diff
@@ -190,22 +190,27 @@ calculateTransCorrections eActions = do
...
+                    -- "user didn't have cert and doesn't have now", but
+                    -- algo should not consider these users at all.
                     error $ "Tried to convert (N,N) to These with affected user: " <> pretty a
                 (Just x, Nothing) -> This x
                 (Nothing, Just x) -> That x
                 (Just x, Just y)  -> These x y
+
         let dlgFin = flip HM.mapWithKey affected $ \a dOld ->
+                         -- lookupUnafe here won't fail because 'dlgNew' must
```
int-index: ✔️

core/Pos/Core/Common/Coin.hs
```diff
@@ -50,7 +50,7 @@ unsafeAddCoin :: Coin -> Coin -> Coin
 unsafeAddCoin (unsafeGetCoin -> a) (unsafeGetCoin -> b)
     | res >= a && res >= b && res <= unsafeGetCoin (maxBound @Coin) = mkCoin res
     | otherwise =
-      -- TODO [CSL-2173]: Clarify
+      -- This function is unsafe
```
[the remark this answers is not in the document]
volhovm: >We're basically doing compiler's job here (with something like LiquidHaskell we wouldn't have to), so let's be as formal as possible.

That's not an argument. You see 'addCoin' function and "it can panic" keyword, you read the code => voila. No need for duplicating the check condition twice.

core/Pos/Core/Configuration.hs
```diff
@@ -170,7 +170,7 @@ withGenesisSpec theSystemStart conf@CoreConfiguration{..} val = case ccGenesis o
                 -- Unite with generated
                 finalHeavyDelegation :: GenesisDelegation
                 finalHeavyDelegation =
-                    -- TODO [CSL-2173]: Clarify
+                    -- We couldn't create the genesis delegation, so we panic.
```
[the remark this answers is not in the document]
volhovm: We generate genesis delegation once only, in the beginning of the program. If it can't be generated, we can't start. Thus, we panic.

delegation/src/Pos/Arbitrary/Delegation.hs
volhovm: It can. If it happens, we crash.

core/Pos/Core/Configuration.hs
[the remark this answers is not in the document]
int-index: Then it shouldn't be represented with `error` if that's something that can theoretically happen.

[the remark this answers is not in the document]
volhovm: No, it's alright, you just do not understand how to use `error` properly. It's a programmer bug to create a malformed configuration, so there we have it.

delegation/src/Pos/Arbitrary/Delegation.hs
int-index: Don't crash with `error` if it's something you expect can happen. Just change the "TODO" from "Clarify" to "Refactor" and I'll take it from there.

jcmincke:
Related to a closed issue

PR #2516, opened 2018-02-15
[CSL-198] Revise exception handling in 'onNewSlot'

int-index:


block/src/Pos/Block/Worker.hs
```diff
@@ -82,6 +82,8 @@ blkWorkers =
 
 informerWorker :: BlockWorkMode ctx m => (WorkerSpec m, OutSpecs)
 informerWorker =
+    -- TODO: [CSL-198] Catch and report non-critical errors, as
+    -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: I don't see why this worker may fail in normal cases, so let's stop program if it does (i. e. not catch anything here).

block/src/Pos/Block/Worker.hs
```diff
@@ -105,6 +107,8 @@ informerWorker =
 
 blkCreatorWorker :: BlockWorkMode ctx m => (WorkerSpec m, OutSpecs)
 blkCreatorWorker =
+    -- TODO: [CSL-198] Catch and report non-critical errors, as
+    -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: This worker calls `Diffusion.announceBlockHeader`, so I guess it may fail when something goes wrong with network. I propose to catch exceptions from `announceBlockHeader` and retry later if they occur. On other cases let's stop the program.

infra/Pos/DHT/Workers.hs
```diff
@@ -56,6 +56,8 @@ dumpKademliaStateWorker
     => KademliaDHTInstance
     -> (WorkerSpec m, OutSpecs)
 dumpKademliaStateWorker kademliaInst = localOnNewSlotWorker onsp $ \slotId ->
+    -- TODO: [CSL-198] Catch and report non-critical errors, as
+    -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: I don't see why this worker may fail in normal cases, so let's stop program if it does (i. e. not catch anything here).

ssc/Pos/Ssc/Worker.hs
```diff
@@ -395,6 +397,8 @@ checkForIgnoredCommitmentsWorker
     => (WorkerSpec m, OutSpecs)
 checkForIgnoredCommitmentsWorker = localWorker $ do
     counter <- newTVarIO 0
+    -- TODO: [CSL-198] Catch and report non-critical errors, as
+    -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: I don't see why this worker may fail in normal cases, so let's stop program if it does (i. e. not catch anything here).

ssc/Pos/Ssc/Worker.hs
```diff
@@ -87,6 +87,8 @@ onNewSlotSsc
     :: (SscMode ctx m)
     => (WorkerSpec m, OutSpecs)
 onNewSlotSsc = onNewSlotWorker defaultOnNewSlotParams mempty $ \slotId diffusion ->
+    -- TODO: [CSL-198] Catch and report non-critical errors, as
+    -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: This worker may call `sendSscCommitment`, `sendSscOpening`, etc. Let's catch exceptions from these functions and restart. Other exceptions shouldn't be caught.

update/Pos/Update/Worker.hs
```diff
@@ -40,11 +40,15 @@ usWorkers = (map fst [processNewSlotWorker, checkForUpdateWorker], mempty)
               NewSlotTerminationPolicy "Update.processNewSlot"
         }
     processNewSlotWorker =
+        -- TODO: [CSL-198] Catch and report non-critical errors, as
+        -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: I don't see why this worker may fail in normal cases, so let's stop program if it does (i. e. not catch anything here).

update/Pos/Update/Worker.hs
```diff
@@ -40,11 +40,15 @@ usWorkers = (map fst [processNewSlotWorker, checkForUpdateWorker], mempty)
...
         localOnNewSlotWorker processNewSlotParams $ \s ->
             recoveryCommGuard "processNewSlot in US" $ do
                 logDebug "Updating slot for US..."
                 processNewSlot s
     checkForUpdateWorker =
+        -- TODO: [CSL-198] Catch and report non-critical errors, as
+        -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: AFAIU, this worker already catches exceptions which may occur during update download, so I think nothing should be changed.

wallet/src/Pos/Wallet/Web/Pending/Worker.hs
```diff
@@ -167,6 +167,8 @@ startPendingTxsResubmitter
     => (TxAux -> m Bool)
     -> m ()
 startPendingTxsResubmitter submitTx =
+    -- TODO: [CSL-198] Catch and report non-critical errors, as
+    -- 'onNewSlot' no longer swallows all exception and rethrows them instead.
```
gromakovsky: It seems that this worker already catches exceptions which may occur during tx submission (see code in `submitAndSavePtx`), so I think nothing should be changed.

util/Pos/Util/Util.hs
```diff
@@ -170,6 +172,11 @@ parsecError = toParsecError . Left
 liftE :: TH.Lift a => Either Text a -> TH.ExpQ
 liftE = TH.lift <=< toTemplateHaskellError
 
+pattern Exc :: Exception e => e -> SomeException
```
gromakovsky: Adding a Haddock comment and probably a usage example would be very useful, because this feature is not well-known to majority of people.

lib/src/Pos/Launcher/Runner.hs
```diff
@@ -71,13 +73,14 @@ runRealMode
     -> (ActionSpec (RealMode ext) a, OutSpecs)
     -> Production a
 runRealMode nr@NodeResources {..} (actionSpec, outSpecs) =
-    elimRealMode nr $ runServer
+    elimRealMode nr $ reportExc $ runServer
```
gromakovsky: This is wrong place for this logic, because it's not used by wallet.

block/src/Pos/Block/Worker.hs
```diff
@@ -107,9 +112,20 @@ blkCreatorWorker :: BlockWorkMode ctx m => (WorkerSpec m, OutSpecs)
...
     onNewSlotWorker onsp mempty $ \slotId diffusion ->
         recoveryCommGuard "onNewSlot worker, blkCreatorWorker" $
-        blockCreator slotId diffusion `catchAny` onBlockCreatorException
+        catchReportExc $
+        blockCreator slotId diffusion
   where
-    onBlockCreatorException = reportOrLogE "blockCreator failed: "
+    -- Detect errors that should not crash the worker (and the node with it).
+    isNonCriticalExc = \case
+        -- TODO: [CSL-198] Revise this list of errors.
```
gromakovsky: Are you going to do it in near time? I mean, I except CSL-198 to be closed in next few days, and if it's closed, there shouldn't be such comment.
I would test scenario when network connection disappears (dunno how to do it on localhost, but I guess there is a way) to know which exception is thrown in this case (and not sure if it will be `DialogUnexpected`) and I think it will be enough.

ssc/Pos/Ssc/Worker.hs
```diff
@@ -365,14 +380,19 @@ waitUntilSend msgTag epoch slMultiplier = do
...
-        let delta = maxToSend - minToSend
+    timeDiff <- do
+        -- Compute an approximation for the time difference between the current
+        -- slot and the next slot. The result might be inaccurate, i tak soydet.
+        curSlot <- getCurrentSlotInaccurate
+        slotDur <- getNextEpochSlotDuration
+        let
+            targetSlot = SlotId {siEpoch = epoch, siSlot = slot}
+            subzero a b = if a > b then a - b else 0
+            slotDiff = (subzero `on` flattenSlotId) curSlot targetSlot
```
gromakovsky: Shouldn't `targetSlot` be before `curSlot`?

int-index: Right, I used `subtract` initially, but then forgot to flip the arguments back.

ssc/Pos/Ssc/Worker.hs
```diff
@@ -365,14 +380,19 @@ waitUntilSend msgTag epoch slMultiplier = do
...
+        curSlot <- getCurrentSlotInaccurate
+        slotDur <- getNextEpochSlotDuration
+        let
+            targetSlot = SlotId {siEpoch = epoch, siSlot = slot}
+            subzero a b = if a > b then a - b else 0
+            slotDiff = (subzero `on` flattenSlotId) targetSlot curSlot
+        return $ fromMicroseconds $
+            fromIntegral slotDiff * toMicroseconds slotDur
+    let delta = timeDiff + mpcSendInterval
+    when (delta > 0) $ do
```
gromakovsky: Now I see that semantics of this code changed in a bad way. `delta` will always be `mpcSendInterval`, because `waitUntilSend` is always called after `targetSlot` (which is probably not the best name) starts.
`targetSlot` is the slot when some phase starts (e. g. opening phase starts at `4k`-th slot). When this phase starts, we'll call `onNewSlotOpening` which will call `waitUntilSend`.
So in fact `slotDiff` should be `curSlot - targetSlot`, `delta` should be `mpcSendInterval - timeDiff` and some variables should be renamed.
`targetSlot` can be called `phaseStartSlot`, `timeDiff` can still be called `timeDiff`, but it should be commented (that it's the different between beginning of the current slot and `phaseStartSlot`).

PR #2523, opened 2018-02-16
Script to run the new wallet with new impl

edsko:
The existing demo-with-new-wallet-api script was starting the new wallet in "legacy mode", where it's no more than a thin wrapper around the old wallet code. The new wallet in legacy mode was just there as a proof of concept. I've renamed that script now to demo-with-legacy-new-wallet-api. The demo-with-new-wallet-api script now runs the new wallet starting the _actual_ new wallet. In order to do this I needed to add the possibility to demo.sh to pass additional arguments to the wallet executable.

This also adds some logging to the BListener hooks so that we can verify that they get called; the log will contain entries such as

```
node3.log:[*production*:ERROR:ThreadId 68] [2018-02-16 16:22:32.01 UTC] walletApplyBlocks not implemented
```

@georgeee , requesting review from you just to make sure that the change I made to `demo.sh` is benign and won't break other users.

### DIFF
--- a/scripts/launch/demo-with-legacy-new-wallet-api.sh
+++ b/scripts/launch/demo-with-legacy-new-wallet-api.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+base=$(dirname "$0")
+
+WALLET_EXE_NAME='cardano-node-new' WALLET_TEST=1 "$base"/demo.sh $@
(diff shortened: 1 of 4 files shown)


adinapoli-iohk:
Thank you very much for this, @edsko ! Once @georgeee approves and the conflicts are addresses, I'm OK to merge this and inform the rest of the team about the semantic change in the script(s).

edsko:
Rebased on latest `develop`.

PR #2525, opened 2018-02-16
[CSL-2294] V1 API style changes

adinapoli-iohk:
This PR delivers part of the style changes discussed as part of CSL-2294 (cc @DominikGuzei ).

It still doesn't tackle proper filtering of the `/transactions` endpoint, but that can be delivered separately to not yield a monstrous PR.


wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -60,27 +59,152 @@ import           Universum
...
+    parseJSON x = typeMismatch "parseJSON failed for Address" x
+
+instance Arbitrary (V1 Core.Address) where
+    arbitrary = fmap V1 arbitrary
+
+instance FromHttpApiData (V1 Core.Address) where
+    parseQueryParam = fmap (fmap V1) Core.decodeTextAddress
+
+instance ToHttpApiData (V1 Core.Address) where
+    toQueryParam (V1 a) = sformat build a
```
gromakovsky: You could have used `addressF` instead of `build` here as it's done few lines above (in `instance ToJSON (V1 Core.Address) where`) for consistency.
`addressF` is `build` with more concrete type.
Anyway, it's not important at all :)

adinapoli-iohk: Ha, thanks for the tip! Lazyness, developer lazyness it was :) Will amend in subsequent PRs.

PR #2528, opened 2018-02-16
[CSL-2134] Block verification optimizations

volhovm:
Verification for an average mainnet 2100b batch now takes about 0.5-0.6s, which was about 1s before. I've lowered the delegation component overhead by adding the cache into `CedeModifier`, but other optimizations were helpful as well. Update component should be a little bit quicker also now, but it's something like 10%, not much. The bottleneck in update component is still there (every time we apply a block, we call `applyImplicitAgreement`, `applyDepthCheck` and also record block issuance, this is very expensive), and it's really hard to fix (more of an update system design drawback), so I doubt we need to do it right now.

### DIFF
--- a/delegation/src/Pos/Delegation/Cede/Holders.hs
+++ b/delegation/src/Pos/Delegation/Cede/Holders.hs
@@ -13,7 +13,7 @@ module Pos.Delegation.Cede.Holders
 import           Universum
 
 import           Control.Lens (at, (%=), (.=))
-import qualified Control.Monad.State as Mtl
+import qualified Control.Monad.State.Strict as Mtl
 import           Control.Monad.Trans.Identity (IdentityT (..))
 import           Data.Coerce (coerce)
 import qualified Data.HashMap.Strict as HM
@@ -49,13 +49,15 @@ instance (MonadDBRead m, MonadUnliftIO m) => MonadCedeRead (DBCede m) where
 
 -- | Monad transformer that holds extra layer of modifications to the
 -- underlying set of PSKs (which can be empty if you want).
-type MapCede = Mtl.StateT CedeModifier
+newtype MapCede m a = MapCede
+    { unMapCede :: StateT CedeModifier m a
+    } deriving (Functor,Applicative,Monad,MonadIO,MonadState CedeModifier,MonadThrow,MonadTrans)
 
 runMapCede :: CedeModifier -> MapCede m a -> m (a, CedeModifier)
-runMapCede = flip Mtl.runStateT
+runMapCede c m = Mtl.runStateT (unMapCede m) c
 
 evalMapCede :: Monad m => CedeModifier -> MapCede m a -> m a
-evalMapCede = flip Mtl.evalStateT
+evalMapCede c m = Mtl.evalStateT (unMapCede m) c
 
 instance (MonadDBRead m, MonadUnliftIO m) => MonadCedeRead (MapCede m) where
     getPsk i =
--- a/delegation/src/Pos/Delegation/Cede/Holders.hs
+++ b/delegation/src/Pos/Delegation/Cede/Holders.hs
@@ -12,28 +12,25 @@ module Pos.Delegation.Cede.Holders
 
 import           Universum
 
-import           Control.Lens (at, (%=))
+import           Control.Lens (at, (%=), (.=))
+import qualified Control.Monad.State as Mtl
 import           Control.Monad.Trans.Identity (IdentityT (..))
 import           Data.Coerce (coerce)
 import qualified Data.HashMap.Strict as HM
 import           Data.HashSet as HS
-import qualified Ether
 import           UnliftIO (MonadUnliftIO)
 
 import           Pos.DB.Class (MonadDBRead)
 import           Pos.Delegation.Cede.Class (MonadCede (..), MonadCedeRead (..))
 import           Pos.Delegation.Cede.Types (CedeModifier, DlgEdgeAction (..), cmHasPostedThisEpoch,
-                                            cmPskMods, dlgEdgeActionIssuer)
+                                            cmLookupCache, cmPskMods, dlgEdgeActionIssuer)
 import qualified Pos.Delegation.DB as DB
-import           Pos.Util.Util (ether)
 
 ----------------------------------------------------------------------------
 -- Pure database-only holder
 ----------------------------------------------------------------------------
 
-data DBCedeTag
-
-type DBCede = Ether.TaggedTrans DBCedeTag IdentityT
+type DBCede = IdentityT
 
 runDBCede :: DBCede m a -> m a
 runDBCede = coerce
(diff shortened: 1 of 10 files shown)


delegation/src/Pos/Delegation/Cede/Logic.hs
```diff
@@ -52,6 +53,8 @@ getPskChainInternal toIgnore issuer =
     trav x | HS.member x toIgnore = pass
     trav x = do
         whenM (uses _2 $ HS.member x) $
+            -- FIXME @volhovm We should throw something else, we don't
```
georgeee: FixME?

volhovm: It's for the future, not related to this particular PR. Probably will be fixed when we refining exceptions. In 2019 or something.

delegation/src/Pos/Delegation/Cede/Holders.hs
```diff
@@ -52,25 +49,31 @@ instance (MonadDBRead m, MonadUnliftIO m) => MonadCedeRead (DBCede m) where
 
 -- | Monad transformer that holds extra layer of modifications to the
 -- underlying set of PSKs (which can be empty if you want).
-type MapCede = Ether.LazyStateT' CedeModifier
+type MapCede = Mtl.StateT CedeModifier
```
georgeee: Are we sure we want lazy state?

volhovm: Switched to strict.

update/Pos/Update/Logic/Global.hs
```diff
@@ -158,9 +160,11 @@ usVerifyBlocks ::
 usVerifyBlocks verifyAllIsKnown blocks =
     withUSLogger $
     reportUnexpectedError $
-    processRes <$> run (runExceptT $ mapM (verifyBlock verifyAllIsKnown) blocks)
+    processRes <$> run (runExceptT action)
   where
-    run :: forall m a . PollT (DBPoll m) a -> m (a, PollModifier)
```
gromakovsky: Why did you delete the type? It was rather helpful.

update/Pos/Update/Poll/Logic/Base.hs
```diff
@@ -81,12 +81,12 @@ confirmBlockVersion confirmedEpoch bv =
 -- Specifically, one of the following conditions must be true.
 -- • Given block version is equal to last adopted block version.
 -- • Given block version is competing
-canCreateBlockBV :: MonadPollRead m => BlockVersion -> m Bool
-canCreateBlockBV bv = do
-    lastAdopted <- getAdoptedBV
+canCreateBlockBV :: MonadPollRead m => BlockVersion -> BlockVersion -> m Bool
```
gromakovsky: It would be good to introduce newtype wrappers or use `Tagged` here to make it harder to confuse the order of `BlockVersion`s, but it's minor.

delegation/src/Pos/Delegation/Cede/Types.hs
```diff
@@ -45,19 +47,25 @@ dlgEdgeActionIssuer = \case
...
 -- Cede-related typeclasses.
 data CedeModifier = CedeModifier
     { _cmPskMods            :: !(HashMap StakeholderId DlgEdgeAction)
       -- ^ PSK modifications (added/overwritten or deleted)
     , _cmHasPostedThisEpoch :: !(HashMap StakeholderId Bool)
       -- ^ 'True' means user has posted psk, 'False' means he didn't.
+    , _cmLookupCache        :: !(HashMap StakeholderId (Maybe ProxySKHeavy))
+      -- ^ This is a hacky cache to make things faster (it actually
+      -- does). It can also make things slower in the future. Cache
+      -- invalidation is hard, so we'll just `empty` it every time
```
gromakovsky: What do you mean by `empty` here? My interpretation was that whenever something is written, cache becomes `mempty`. But it's not the case: you only set `Nothing` for one key. So this sentence sounds confusing to me.

delegation/src/Pos/Delegation/Cede/Logic.hs
```diff
@@ -86,73 +89,61 @@ detectCycleOnAddition toAdd
...
+-- | Returns the last certificate in the chain, starting with given
+-- issuer.
+--
+-- Delegate 'd' has right to issue block instead of issuer 'i' if
+-- there's a delegation chain:
+--
+-- i → x₁ → x₂ → … xₖ → d → ∅
+--
+-- where every arrow is resolved psk, and the last one xₖ → d
+-- equals to the passed one.
```
gromakovsky: > equals to the passed one

PSK is not passed anymore.

delegation/src/Pos/Delegation/Cede/Logic.hs
```diff
@@ -86,73 +89,61 @@ detectCycleOnAddition toAdd
...
 
 -- | Verifies a header from delegation perspective (signature checks).
 dlgVerifyHeader ::
        (MonadCedeRead m)
     => MainBlockHeader
     -> m (Either Text ())
 dlgVerifyHeader h = runExceptT $ do
-    -- Issuer didn't delegate the right to issue to elseone.
-    let issuer = h ^. mainHeaderLeaderKey
+    let bIssuer = h ^. mainHeaderLeaderKey
```
gromakovsky: Perhaps it'd be better to call this variable `leader` or `leaderPK`. Issuer is a bit ambiguous word: it can be block issuer or PSK issuer. Apparently `b` means block, but the leader doesn't necessary issue a block.
Moreover it's possible that the leader didn't issue anything. So calling her `bIssuer` is something I don't fully understand.

delegation/src/Pos/Delegation/Cede/Logic.hs
```diff
@@ -86,73 +89,61 @@ detectCycleOnAddition toAdd
...
+                throwError $
+                sformat ("issuer "%build%" has delegated issuance right, "%
+                         "so he can't issue the block himself, psk: "%build%", sig: "%build)
+                    bIssuer psk sig
+        -- We check that the cert is related to block issuer. This
+        -- also checks that the delegate didn't delegate himself.
         (BlockPSignatureHeavy pSig) -> do
-            let psk = psigPsk pSig
-            let delegate = pskDelegatePk psk
-            canIssue <- dlgReachesIssuance issuer delegate psk
```
gromakovsky: Hm, did the old version check that `delegate` didn't re-delegate anything further?

delegation/src/Pos/Delegation/Cede/Logic.hs
```diff
@@ -86,73 +89,61 @@ detectCycleOnAddition toAdd
...
+            lastPskM <- dlgLastPsk (addressHash bIssuer)
+            let canIssue = lastPskM == (Just $ psigPsk pSig)
             unless canIssue $ throwError $
                 sformat ("heavy proxy signature's "%build%" "%
                          "related proxy cert can't be found/doesn't "%
                          "match the one in current allowed heavy psks set")
                         pSig
+        -- We check that the light signature
```
gromakovsky: We check that the light signature does what?

delegation/src/Pos/Delegation/Logic/Common.hs
```diff
@@ -79,21 +77,9 @@ runDelegationStateAction action = do
...
-                find (\psk -> addressHash (pskDelegatePk psk) == dPk) chain
-        let iPk = pskIssuerPk <$>
-                  find (\psk -> addressHash (pskIssuerPk psk) == issuer) chain
-        let throwEmpty = throwM $ DBMalformed $
-                sformat ("getDlgTransPk: couldn't find psk with dlgPk "%build%
-                         " and issuer "%build)
-                        dPk issuer
-        maybe throwEmpty (pure . Just) $ (,) <$> iPk <*> finalPsk
+getDlgTransPsk issuer =
+    fmap (\psk -> (pskIssuerPk psk, psk)) <$> runDBCede (dlgLastPsk issuer)
```
gromakovsky: I suspect you changed the semantics of this function. Previouisly it was returning a `PublicKey` such that its `addressHash` was equal to `issuer`. Now it returns `pskIssuerPk` from the last PSK in the chain.
Is my understanding correct? If it is, is this change intentional?

update/Pos/Update/Logic/Global.hs
volhovm: Restored.

delegation/src/Pos/Delegation/Cede/Types.hs
volhovm: I don't set `Nothing`, i set "nothing", this is how `at` works. I've updated the comment.

delegation/src/Pos/Delegation/Cede/Logic.hs
volhovm: Fixed.

delegation/src/Pos/Delegation/Cede/Logic.hs
volhovm: Valid, fixed.

delegation/src/Pos/Delegation/Logic/Common.hs
volhovm: It was not.

PR #2531, opened 2018-02-16
[CSL-2295] Add an option to continuously dump blocks to cardano-auxx

neongreen:
This will probably be needed for devops.

Note: before this is merged, I'd like someone to help me with a small issue. When I run `dump-blockchain-loop` right after auxx starts, it works but the block retrieval worker never starts for some reason :/ When I wait a bit before running `dump-blockchain-loop`, everything works fine. Does anybody know why that might happen?


neongreen:
Update: it probably was happening because I tested with a small delay (5s instead of 1h) and slot duration is a multiple of 5s so it was always blocked. With a bigger delay everything is fine.

auxx/src/Command/DumpBlockchain.hs
```diff
@@ -39,84 +43,171 @@ import           Mode (MonadAuxxMode)
...
-        errFmt = "getBlockThrow: no block with HeaderHash: "%shortHashF
+                    C.sinkFile outPath
+                case maybePrev of
+                    Nothing -> pure Nothing
+                    Just prev -> DB.getHeader prev >>= \case
+                        Just header -> pure (Just header)
+                        Nothing     -> do
+                            let anchor = headerHash $
+                                         last (getNewestFirst blocks)
+                            logWarning ("DB contains block "+|anchor|+
```
Pastafarianist: I think this should qualify as an error, not a logged warning.

auxx/src/Command/DumpBlockchain.hs
```diff
@@ -39,84 +43,171 @@ import           Mode (MonadAuxxMode)
 -- Each epoch will be at <outFolder>/epoch<epochIndex>.cbor.
 dumpBlockchain
     :: (MonadAuxxMode m, MonadIO m)
-    => FilePath
+    => Maybe EpochIndex                 -- ^ Start epoch (Nothing = 0)
```
Pastafarianist: It would probably be better to comment that both ends of the range are inclusive.

auxx/src/Command/DumpBlockchain.hs
```diff
@@ -39,84 +43,171 @@ import           Mode (MonadAuxxMode)
...
+                           "already dumped")
+            | otherwise -> do
+                  logInfo ("The last stable epoch is "+|stableEpoch|+"")
+                  dumpBlockchain (Just (lastDumpedEpoch + 1))
+                                 (Just stableEpoch)
+                                 outFolder
+    -- Wait before repeating. Without the delay, we'll spend all the time
+    -- doing IO. (A one-hour delay before a fresh epoch becomes available
+    -- seems fine, given that epochs last several days.)
+    delay (hour 1)
```
Pastafarianist: Instead of polling, this could probably be hooked into `BListener`. However, I don't think the benefit of not having to wait for half an hour (on average) for the new epoch to be served is worth rewriting all this code.

Pastafarianist:
Merging because issues are minor and fixing them would trigger CI which would block merging for a significant period of time.

PR #2541, opened 2018-02-19
[CBR-89] HTTP block sync

Pastafarianist:
In this PR we present functionality of HTTP mirror based blockchain syncing.

This functionality will allow us to dramatically improve block retrieval experience via use of common technology for file distribution (HTTP) which is more robust than in-built data exchange layer in Cardano and comes to us along with excellent infrastructure at our disposal (S3 and alternatives).

This HTTP download comes at very low complexity cost (block retrieval logic remains almost unchanged), may serve as good short- and mid-term solution for Cardano SL block syncing (on node recovery after being offline).

In particular this PR includes:
 * By-epoch block dump preparation via auxx command
     * Dumps are cbor encoding of epochs, compressed with LZMA
 * By-epoch block retrieval from HTTP mirror as part of node recovery
 * By-epoch dump application via auxx command

Example PR of changes needed for Daedalus: https://github.com/input-output-hk/daedalus/pull/730


georgeee:
Dmitry benchmarked HTTP downloading vs peer downloading on the same machine using the same internet connection. One caveat is that I served epoch dumps over HTTP from `localhost`. This should not affect the results much, since the total download size is about 189M.

The total time to sync using HTTP was 43 minutes. Syncing from peers took 1 hour 45 minutes. The graph shows local difficulty (essentially, the number of applied blocks) over time. Missing labels on x-axis are likely a [Pandas bug](https://github.com/pandas-dev/pandas/issues/19965). I decided not to spend any time on a workaround.

If someone wants to reproduce these results, be aware that Cardano needs to be built with `-O2 --no-asserts` (like `./b -O2 --no-asserts sl`), otherwise the performance will be much worse.

Pastafarianist:
I benchmarked HTTP downloading vs peer downloading on the same machine using the same internet connection. One caveat is that I served epoch dumps over HTTP from localhost. This should not affect the results much, since the total download size is about 189M.

The total time to sync using HTTP was 43 minutes. Syncing from peers took 1 hour 45 minutes. The graph shows local difficulty (essentially, the number of applied blocks) over time.

If someone wants to reproduce these results, be aware that Cardano needs to be built with -O2 --no-asserts (like ./b -O2 --no-asserts sl), otherwise the performance will be much worse.

image

PR #2543, opened 2018-02-19
Clearing a TODO Comment: File to use for JSON logging

omefire:
Just a minor cleanup I found as I'm perusing the source code.


ChShersh:
@omefire Thanks for your contribution! But did you manage to figure out where this JSON logging is performed and why we need it? We use `log-warper` library for logging but `log-warper` knows nothing about JSON at the moment. So this `npJLFile` is a mystery for everyone.

PR #2544, opened 2018-02-19
[csl-2147] subscription status

coot:
This PR implements:
* subscribtion status
* changes retryInterval of dnsSubscriptionWorker


infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -67,10 +78,17 @@ subscribeTo keepAliveTimer sendActions peer = do
         ]
     let reason = either Exceptional (maybe Normal absurd) outcome
     logNotice $ msgSubscriptionTerminated peer reason
+    -- Change subscription state
+    atomically $ writeTVar subStatus NotSubscribed
+    subEnded <- liftIO getPOSIXTime
```
avieth: Minor point but I think it's best to do this immediately after `outcome <-`

infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -56,9 +59,17 @@ data SubscriptionTerminationReason =
 -- giving the reason. Notices will be logged before and after the subscription.
 subscribeTo
     :: forall m. (SubscriptionMode m)
-    => Timer -> SendActions m -> NodeId -> m SubscriptionTerminationReason
-subscribeTo keepAliveTimer sendActions peer = do
+    => Timer
+    -> TVar SubscriptionStatus
+    -> MVar POSIXTime -- ^ Subscription duration.
```
avieth: Should use `NominalDiffTime` or `Microsecond` or similar; something that better indicates "duration".

infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -67,10 +78,17 @@ subscribeTo keepAliveTimer sendActions peer = do
         ]
     let reason = either Exceptional (maybe Normal absurd) outcome
     logNotice $ msgSubscriptionTerminated peer reason
+    -- Change subscription state
+    atomically $ writeTVar subStatus NotSubscribed
+    subEnded <- liftIO getPOSIXTime
+    liftIO $ modifyMVar_ subDuration $ return . max (subEnded - subStarted)
```
avieth: Might as well be strict here. `\x -> return $! max (subEnded - subStarted) x`

infra/Pos/Diffusion/Subscription/Common.hs
coot: `POSIXTime` is just a type alias for `NominalDiffTime`, but let's use `Millisecond` which we use everywhere else.

coot:
@avieth done

infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -56,21 +61,37 @@ data SubscriptionTerminationReason =
...
-subscribeTo keepAliveTimer sendActions peer = do
+    => Timer
+    -> TVar SubscriptionStatus
+    -> MVar Millisecond -- ^ Subscription duration.
+    -> SendActions m
+    -> NodeId
+    -> m SubscriptionTerminationReason
+subscribeTo keepAliveTimer subStatus subDuration sendActions peer = do
+    -- Change subscription status as we begin a new subscription
+    atomically $ writeTVar subStatus Subscribing
```
avieth: Ah, this needs to be slightly different: if it's already subscribed, it should remain subscribed. It should only become subscribing if it's current "not connected".

infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -56,21 +61,37 @@ data SubscriptionTerminationReason =
...
     logNotice $ msgSubscribingTo peer
+    subStarted <- liftIO getPOSIXTime
     outcome <- try $ withConnectionTo sendActions peer $ \_peerData -> NE.fromList
         -- Sort conversations in descending order based on their version so that
         -- the highest available version of the conversation is picked.
         [ Conversation convMsgSubscribe
         , Conversation convMsgSubscribe1
         ]
+    -- Change subscription state
+    atomically $ writeTVar subStatus NotSubscribed
```
avieth: This is a bit trickier: if there's some other subscription active, don't go to NotSubscribed.

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -107,10 +115,24 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
...
+    retryInterval d = do
         slotDur <- nextSlotDuration
-        pure $ max (slotDur `div` 4) (convertUnit (5 :: Second))
+        let -- slot duration in microseconds
+            slotDurF :: Float
+            slotDurF = fromIntegral $ toMicroseconds slotDur
+            -- Number of slots that passed in the longest subscription duration
+            slotsF :: Float
+            slotsF = (fromIntegral $ toMicroseconds d) / slotDurF
+        return $ fromMicroseconds $ floor $ slotDurF / (2 ** slotsF)
```
avieth: From meeting today: @karknu pointed out that `getPOSIXTime` is not necessarily monotonic. It's possible that the system clock changes mid-subscription and gives a negative `d`, leading to a negative `slotsF`, which could give a massive value here. Either we use a monotonic clock, or just throw in a `max 1` here.

infra/Pos/Diffusion/Subscription/Common.hs
coot: You're right, `subscribeAlts` runs concurrently, so we might be subscribing to multiple nodes at the same time.

infra/Pos/Diffusion/Subscription/Dns.hs
coot: The `System.Clock.Monotonic` module provides platform independent monotonic clock with this two issues (which in my opinion will not hurt our software):
 
>  * On Windows XP, this uses GetTickCount, which has a 49.7 day wraparound. Clock works around this problem, but the workaround only works if clockGetTime is called at least once every 24.8 days.
> * On Linux, this uses clock_gettime with CLOCK_MONOTONIC, which (unfortunately) stops when the computer is suspended. Thus, clockGetTime will not include time spent sleeping. Do not rely on this behavior, as it may be fixed in a future version of this library.

coot:
@avieth I update the PR.  The subscription `TVar` holds now a map `Map NodeId SubscriptionStatus` and I removed the `NotSubscribed` constructor.  When we are not subscribed the  map will just not contain it.  I also added a `Semigroup` instance for `SubscriptionStatus`.  But using `liftA2 @(Maybe SubscriptionStatus) (<>)` might not be what the wallet team will expect:
```
Just Subscribed `liftA2 (<>)` Nothing == Nothing
```
If you think I should get rid of it I am fine with that.

infra/Pos/Diffusion/Subscription/Dns.hs
avieth: Good to know. I can only imagine the problems we might have if the node is kept running through a suspend... would be interesting to try it out.

avieth:
> If you think I should get rid of it I am fine with that.

It makes sense... So we have a `Map NodeId (Either Subscribed Subscribing)`, we can fold it using the semigroup instance into `Maybe (Either Subscribed Subscribing)`, and the UI component can use this, taking `Nothing` to mean not connected. Right, I think exposing `TVar (Map NodeId (Either Subscribed Subscribing))` is fine (where the `Either` is of course replaced by the status datatype).

infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -94,6 +116,19 @@ subscribeTo keepAliveTimer sendActions peer = do
...
+
+    alterPeerSubStatus :: Maybe SubscriptionStatus -> m ()
+    alterPeerSubStatus s = atomically $ do
+        stats <- readTVar subStatus
+        let !stats' = Map.alter fn peer stats
+        writeTVar subStatus stats'
+        where
+            fn :: Maybe SubscriptionStatus -> Maybe SubscriptionStatus
+            fn Nothing = s
+            fn (Just t) = (t <>) <$> s
```
avieth: I suppose it is possible that two subscription threads try to subscribe to the same `NodeId` (their DNS names resolve to intersecting sets) but it's a corner case and doesn't/shouldn't happen in practice. I think it's safe to just replace the value with the new one / delete if it's `Nothing`.

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -81,16 +88,17 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
             pure (dnsPeers', ())
         -- Try to subscribe to some peer.
         -- If they all fail, wait a while before trying again.
-        subscribeToOne dnsPeersList
-        retryInterval >>= delay
-        subscribeAlts dnsPeersVar (index, alts)
+        subscribeToOne dnsPeersList subDuration
+        d <- swapMVar subDuration 0
+        retryInterval d >>= delay
```
avieth: It is technically out of the scope of this patch, but I think it'd be good to put a `logNotice` here which includes the actual delay. `dnsSubscriptionWorker: waiting <n>ms before trying again`.

infra/cardano-sl-infra.cabal
```diff
@@ -176,6 +176,7 @@ library
                       , safe-exceptions
                       , serokell-util >= 0.1.3.4
                       , stm
+                      , clock
```
avieth: I believe you'll have to `./pkgs/generate.sh` and then commit the resulting `./pkgs/default.nix` in order for CI to pass.

coot: Indeed, that's what the logs advice too :). It worked locally, so it should work in CI as well.

infra/Pos/Diffusion/Subscription/Common.hs
coot: I agree, this shouldn't be a problem.

coot:
It's `Maybe SubscriptionStatus` rather than `Either _ _`, and the thing is that `Maybe a` semigroup instance is not the one which one would like to use for folding, one would need a wrapper to get `Nothing <> s = s` rahter than `Nothing <> _ = Nothing`, `First` is also wrong one. The right one is something like:
```
newtype SubStat = SubState (Maybe SubscriptionStatus)

instance Semigroup SubState where
  (Nothing _) <> s = s
  (Just s) <> (Just s') = Just $ s <> s'
```

But if we don't provide the `NotSubscribed` we save a potential memory leak - the subscription state map would store all the nodes that we tried to subscribed at any time...

avieth:
> But if we don't provide the NotSubscribed we save a potential memory leak - the subscription state map would store all the nodes that we tried to subscribed at any time...

Right, I agree on that. I wrote `Either Subscribed NotSubscribed` for explicitness, I realize we'd be using an isomorphic datatype. `Map k` is foldable, so we can use `foldr (<>) (Option Nothing) (fmap (Option . Just) map)`.

infra/Pos/Diffusion/Subscription/Dns.hs
coot: The same happens with `System.Clock.getTime Monotonic` from the `clock` library, at least on linux. It stops running when I suspend my laptop.

avieth:
@coot to fix CI you can either look at the stack2nix failure logs and take the patch it provides you, or run `./pkgs/generate.sh`.

coot:
Indeed, now it should work.

coot:
@avieth all tests passed :)

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -54,52 +60,57 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
...
-    void $ forConcurrently allOf (subscribeAlts dnsPeersVar)
+    void $ forConcurrently allOf (subscribeAlts dnsPeersVar subDuration)
     logNotice $ sformat ("dnsSubscriptionWorker: all "%int%" threads finished") (length allOf)
   where
 
     allOf :: [(Int, Alts (NodeAddr DNS.Domain))]
     allOf = zip [1..] dnsDomains
 
-    -- Resolve all of the names and try to subscribe to one.
+    -- Resolve all of the namtas and try to subscribe to one.
```
avieth: Typo

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -54,52 +60,57 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
...
         -- It does handle the case in which there's no internet connection, or
         -- a bad configuration, so that the subscription thread will keep on
         -- retrying.
-        findAndSubscribe dnsPeersVar index alts
+        findAndSubscribe dnsPeersVar subDuration index alts
             `catch` logDNSError
             `catch` logIOException
-        retryInterval >>= delay
-        subscribeAlts dnsPeersVar (index, alts)
+        d <- swapMVar subDuration 0 >>= retryInterval
```
avieth: Why `swapMVar` to `0`? Didn't we take a `max` before?

coot: Here at this point we tried to subscribe to every body so we can zero the counter and measure the max time for the next round.  This way we don't accumulate maximal time if something worked well and then it stopped.

We compute max time for each round, i.e. per `subscribeAlts` call. The `max` you are referring to is [here](https://github.com/input-output-hk/cardano-sl/pull/2544/files#diff-555a6bf9a7a1fb7e3438987fd51919aeR86).

infra/Pos/Diffusion/Subscription/Dns.hs
coot: Fixed.

infra/Pos/Diffusion/Subscription/Dns.hs
avieth: Ah but the `subDuration` `MVar` is shared between potentially more than 1 thread. It's a conjunction of disjunctions. `subscribeAlts` runs through a single conjunct, but there could be others.

infra/Pos/Diffusion/Types.hs
```diff
@@ -78,14 +95,16 @@ data DiffusionLayer m = DiffusionLayer
     }
 
 -- | A diffusion layer that does nothing.
-dummyDiffusionLayer :: Applicative m => DiffusionLayer m
-dummyDiffusionLayer = DiffusionLayer
-    { runDiffusionLayer = identity
-    , diffusion         = dummyDiffusion
-    }
+dummyDiffusionLayer :: (Monad m, MonadIO m, Monad d) => m (DiffusionLayer d)
```
coot: I should release constraint of `Monad d` to `Applicative d` here.

infra/Pos/Diffusion/Subscription/Common.hs
coot: Indeed, they are isomorphic.

infra/Pos/Diffusion/Subscription/Dns.hs
coot: You are right, this is cleared to often. Instead we should clear it once the outer list (conjuctions) are consumed.

coot: After a pause of thinking, I think it is just better to create the `subDuration` `MVar` in each thread.  This way it will control how fast `subscribeAlts` spins through a list of alternatives.

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -74,25 +72,28 @@ dnsSubscriptionWorker oq networkCfg DnsDomains{..} keepaliveTimer nextSlotDurati
...
-    subscribeAlts dnsPeersVar subDuration (index, alts) = do
+    subscribeAlts dnsPeersVar mSubDuration (index, alts) = do
         -- Any DNSError is squelched. So are IOExceptions, for good measure.
         -- This does not include async exceptions.
         -- It does handle the case in which there's no internet connection, or
         -- a bad configuration, so that the subscription thread will keep on
         -- retrying.
+        subDuration <- case mSubDuration of
+            Just sd -> return sd
+            Nothing -> newMVar (0 :: Millisecond)
```
coot: @avieth if you think creating a single mutable cell for each list of alternatives will allocate too much memory, we can do this the other way around, i.e. reset the `subDuration` when all threads have run at least once. This would requires to store a list of `[Bool]` (hence it will only be lineary more memory efficient).

infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -57,22 +61,38 @@ data SubscriptionTerminationReason =
...
+        (\x -> return $! max x (timeSpecToMilliseconds $ subEnded - subStarted))
     let reason = either Exceptional (maybe Normal absurd) outcome
     logNotice $ msgSubscriptionTerminated peer reason
     return reason
   where
     convMsgSubscribe :: ConversationActions MsgSubscribe Void m -> m t
     convMsgSubscribe conv = do
+        -- We are now subscribed, in rare cases when the connection will be
+        -- dropped this will result in a missleading subscription state.
+        alterPeerSubStatus (Just Subscribed)
```
avieth: I wonder if it would make any difference if we did this only after the `send conv MsgSubscribe`. What do you think? Maybe not... if this continuation is being run then we definitely have a functioning TCP socket. The chances that the send would fail are slim... Just thinking aloud.

infra/Pos/Diffusion/Subscription/Common.hs
```diff
@@ -96,6 +117,18 @@ subscribeTo keepAliveTimer sendActions peer = do
...
+    timeSpecToMilliseconds = fromMicroseconds . div 1000 . toNanoSecs
+
+    alterPeerSubStatus :: Maybe SubscriptionStatus -> m ()
+    alterPeerSubStatus s = atomically $ do
+        stats <- readTVar subStatus
+        let !stats' = Map.alter fn peer stats
+        writeTVar subStatus stats'
+        where
+            fn :: Maybe SubscriptionStatus -> Maybe SubscriptionStatus
+            fn x = getOption (Option x <> Option s)
```
avieth: Nice.

`alterPeerSubStatus Nothing` should go in a `finally` on `subscribeTo` I think, so that if the thing is killed even by an async exception we always bring the state back to consistency.

infra/Pos/Diffusion/Subscription/Dns.hs
avieth: Yes that sounds reasonable. Also `subscribeAlts` should just take an `MVar Millisecond` rather than a `Maybe (MVar Millisecond)`.

infra/Pos/Diffusion/Types.hs
```diff
@@ -78,14 +95,16 @@ data DiffusionLayer m = DiffusionLayer
     }
 
 -- | A diffusion layer that does nothing.
-dummyDiffusionLayer :: Applicative m => DiffusionLayer m
-dummyDiffusionLayer = DiffusionLayer
-    { runDiffusionLayer = identity
-    , diffusion         = dummyDiffusion
-    }
+dummyDiffusionLayer :: (Monad m, MonadIO m, Applicative d) => m (DiffusionLayer d)
```
avieth: Nit: redundant `Monad` constraint.

infra/Pos/Diffusion/Subscription/Dns.hs
coot: So you prefer that the context provides it, here the `MVar` is created by `subscribeAlts`. But ok, let me changes this, it will be more intuitive.

avieth:
@coot anything left to be done on this one or can we merge it?

I've been setting up test/benchmark environments with full diffusion layers and stub logic layers in these past few days and now I'm coming up against some trouble with subscriptions. We haven't made it modular enough. It's not possible to control the subscription. I want to have the client subscribe to a particular known peer (command-line argument) and to block until the subscription is up, and to also take care of the necessary outbound queue book-keeping stuff.

So I'll be doing some refactoring in this area to make that possible, but I don't want to do it before this is merged.

coot:
@avieth I believe we are done with this one and its ready to be merged.

Good idea for refactoring the subscriptions to run experiments.

Marcin Szamotulski, PhD
Haskell Developer | Input Output HK

Website: www.iohk.io
Skype: m.szamotulski
Twitter: @me_coot
PGP Key ID: 0x115ACC1B

Input Output

Twitter Github LinkedIn


This e-mail and any file transmitted with it are confidential and intended
solely for the use of the recipient(s) to whom it is addressed.
Dissemination, distribution, and/or copying of the transmission by anyone
other than the intended recipient(s) is prohibited. If you have received
this transmission in error please notify IOHK immediately and delete it
from your system. E-mail transmissions cannot be guaranteed to be secure or
error free. We do not accept liability for any loss, damage, or error
arising from this transmission

adinapoli-iohk:
@parsonsmatt Just to put this one on your radar -- we can now use this to our advantage to expose even more useful info as part of `/api/v1/node-info` 😀

PR #2546, opened 2018-02-20
[CSL-2312] Generate inductive wallets

parsonsmatt:
This PR implements QuickCheck generators for inductive wallets.

### DIFF
--- a/wallet-new/test/unit/UTxO/BlockGen.hs
+++ b/wallet-new/test/unit/UTxO/BlockGen.hs
@@ -188,7 +188,7 @@ newBlock = do
 
 newChain :: Hash h Addr => BlockGen h [[Value -> Transaction h Addr]]
 newChain = do
-    blockCount <- liftGen $ choose (10, 100)
+    blockCount <- liftGen $ choose (10, 50)
     replicateM blockCount newBlock
 
 zipFees
--- a/wallet-new/test/unit/UTxO/BlockGen.hs
+++ b/wallet-new/test/unit/UTxO/BlockGen.hs
@@ -49,26 +49,25 @@ data BlockGenCtx h
 
 makeLenses ''BlockGenCtx
 
-genValidBlockchain :: Hash h Addr => PreChain h Gen
+genValidBlockchain :: Hash h Addr => PreChain h Gen ()
 genValidBlockchain = toPreChain newChain
 
 toPreChain
     :: Hash h Addr
     => BlockGen h [[Value -> Transaction h Addr]]
-    -> PreChain h Gen
+    -> PreChain h Gen ()
 toPreChain = toPreChainWith identity
 
 toPreChainWith
     :: Hash h Addr
     => (BlockGenCtx h -> BlockGenCtx h)
     -> BlockGen h [[Value -> Transaction h Addr]]
-    -> PreChain h Gen
+    -> PreChain h Gen ()
 toPreChainWith settings bg = DepIndep $ \boot -> do
     ks <- runBlockGenWith settings boot bg
-    return
-        $ OldestFirst
-        . fmap OldestFirst
-        . zipFees ks
+    return $ \fees -> (markOldestFirst (zipFees ks fees), ())
+  where
+   markOldestFirst = OldestFirst . fmap OldestFirst
 
 -- | Given an initial bootstrap 'Transaction' and a function to customize
 -- the other settings in the 'BlockGenCtx', this function will initialize
--- a/wallet-new/test/unit/UTxO/BlockGen.hs
+++ b/wallet-new/test/unit/UTxO/BlockGen.hs
@@ -15,7 +15,7 @@ import           Control.Lens hiding (elements)
 import           Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.Map as Map
 import qualified Data.Set as Set
-import           Pos.Util.Chrono (OldestFirst (..))
+import           Pos.Util.Chrono
 import           Test.QuickCheck
 
 import           Util.DepIndep
@@ -40,8 +40,6 @@ data BlockGenCtx h
     -- ^ The mapping of current addresses and their current account values.
     , _bgcFreshHash              :: !Int
     -- ^ A fresh hash value for each new transaction.
-    , _bgcCurrentBlockchain      :: !(Ledger h Addr)
-    -- ^ The accumulated blockchain thus far.
     , _bgcInputPartiesUpperLimit :: !Int
     -- ^ The upper limit on the number of parties that may be selected as
     -- inputs to a transaction
@@ -83,11 +81,10 @@ runBlockGenWith settings boot m =
 
 -- | Create an initial context from the boot transaction.
 initializeCtx :: Hash h Addr => Transaction h Addr -> BlockGenCtx h
-initializeCtx boot@Transaction{..} = BlockGenCtx {..}
+initializeCtx boot = BlockGenCtx {..}
   where
-    _bgcCurrentUtxo = ledgerUtxo _bgcCurrentBlockchain
+    _bgcCurrentUtxo = trUtxo boot
     _bgcFreshHash = 1
-    _bgcCurrentBlockchain = ledgerSingleton boot
     _bgcInputPartiesUpperLimit = 1
 
 -- | Lift a 'Gen' action into the 'BlockGen' monad.
@@ -158,9 +155,8 @@ newTransaction = do
                 }
 
     -- we assume that the fee is 0 for initializing these transactions
-    bgcCurrentBlockchain %= ledgerAdd (txn 0)
-    ledger <- use bgcCurrentBlockchain
-    bgcCurrentUtxo .= ledgerUtxo ledger
+    -- TODO: This means that the UTxO we maintain is actually inaccurate :/
+    bgcCurrentUtxo %= utxoApply (txn 0)
     pure txn
 
 divvyUp
--- a/wallet-new/test/unit/UTxO/BlockGen.hs
+++ b/wallet-new/test/unit/UTxO/BlockGen.hs
@@ -5,6 +5,8 @@
 
 module UTxO.BlockGen
     ( genValidBlockchain
+    , divvyUp
+    , selectDestinations'
     ) where
 
 import           Universum hiding (use)
@@ -124,10 +126,19 @@ selectSomeInputs = do
             pure (inp : rest)
 
 selectDestinations :: Hash h Addr => Set (Input h Addr) -> BlockGen h (NonEmpty Addr)
-selectDestinations notThese = do
-    utxo <- uses bgcNonAvvmUtxo (utxoRemoveInputs notThese)
-    addr1 <- liftGen $ elements (map (outAddr . snd) (utxoToList utxo))
-    pure (addr1 :| [])
+selectDestinations notThese =
+    liftGen . selectDestinations' notThese =<< use bgcCurrentUtxo
+
+selectDestinations'
+    :: Hash h Addr
+    => Set (Input h Addr)
+    -> Utxo h Addr
+    -> Gen (NonEmpty Addr)
+selectDestinations' notThese =
+    fmap pure . elements
+        . map (outAddr . snd) . utxoToList
+        . utxoRestrictToAddr (not . isAvvmAddr)
+        . utxoRemoveInputs notThese
 
 -- | Create a fresh transaction that depends on the fee provided to it.
 newTransaction :: Hash h Addr => BlockGen h (Value -> Transaction h Addr)
--- a/wallet-new/test/unit/UTxO/BlockGen.hs
+++ b/wallet-new/test/unit/UTxO/BlockGen.hs
@@ -18,6 +18,7 @@ import qualified Data.Set as Set
 import           Pos.Util.Chrono (OldestFirst (..))
 import           Test.QuickCheck
 
+import           Util.DepIndep
 import           UTxO.Context
 import           UTxO.DSL
 import           UTxO.PreChain
@@ -62,7 +63,7 @@ toPreChainWith
     => (BlockGenCtx h -> BlockGenCtx h)
     -> BlockGen h [[Value -> Transaction h Addr]]
     -> PreChain h Gen
-toPreChainWith settings bg = PreChain $ \boot -> do
+toPreChainWith settings bg = DepIndep $ \boot -> do
     ks <- runBlockGenWith settings boot bg
     return
         $ OldestFirst
--- a/wallet-new/test/unit/UTxO/BlockGen.hs
+++ b/wallet-new/test/unit/UTxO/BlockGen.hs
@@ -144,35 +144,40 @@ newTransaction = do
     destinations <- selectDestinations (foldMap Set.singleton (map fst inputs'outputs))
     hash' <- freshHash
 
-    let txn fee =
-            let (inputs, outputs) = divvyUp inputs'outputs destinations fee
-             in Transaction
-                { trFresh = 0
-                , trFee = fee
-                , trHash = hash'
-                , trIns = inputs
-                , trOuts = outputs
-                }
+    let txn = divvyUp hash' inputs'outputs destinations
 
-    -- we assume that the fee is 0 for initializing these transactions
-    -- TODO: This means that the UTxO we maintain is actually inaccurate :/
-    bgcCurrentUtxo %= utxoApply (txn 0)
+    -- We don't know the fee yet, but /do/ need to make it possible to
+    -- generate different kinds of transactions (i.e., different kinds of
+    -- monadic effects) depending on the UTxO. This means that we must be
+    -- conversative here.
+    bgcCurrentUtxo %= utxoApply (withEstimatedFee txn)
     pure txn
 
+-- | Given a set of inputs, tagged with their output values, and a set of output
+-- addresses, construct a transaction by dividing the sum total of the inputs
+-- evenly over the output addresses.
 divvyUp
     :: Hash h Addr
-    => NonEmpty (Input h Addr, Output Addr)
+    => Int
+    -> NonEmpty (Input h Addr, Output Addr)
     -> NonEmpty Addr
     -> Value
-    -> (Set (Input h Addr), [Output Addr])
-divvyUp inputs'outputs destinations fee = (inputs, outputs)
+    -> Transaction h Addr
+divvyUp h inputs'outputs destinations fee = tx
   where
+    tx = Transaction {
+             trFresh = 0
+           , trFee   = fee
+           , trHash  = h
+           , trIns   = inputs
+           , trOuts  = outputs
+           }
     inputs = foldMap (Set.singleton . fst) inputs'outputs
     destLen = fromIntegral (length destinations)
     -- if we don't know what the fee is yet (eg a 0), then we want to use
     -- the max fee for safety's sake
     totalValue = sum (map (outVal . snd) inputs'outputs)
-        `safeSubtract` if fee == 0 then maxFee else fee
+        `safeSubtract` if fee == 0 then estimateFee tx else fee
     valPerOutput = totalValue `div` destLen
     outputs = toList (map (\addr -> Output addr valPerOutput) destinations)
 
@@ -185,8 +190,18 @@ safeSubtract x y
   where
     z = x - y
 
-maxFee :: Num a => a
-maxFee = 180000
+-- | Conversatively estimate the fee for this transaction
+--
+-- Result may be larger than the minimum fee, but not smaller.
+-- TODO: Right now this does not take the transaction structure into account.
+-- We should come up with a more precise model here.
+estimateFee :: Transaction h a -> Value
+estimateFee _ = maxFee
+  where
+    maxFee = 180000
+
+withEstimatedFee :: (Value -> Transaction h a) -> Transaction h a
+withEstimatedFee tx = let tx0 = tx 0 in tx0 { trFee = estimateFee tx0 }
 
 newBlock :: Hash h Addr => BlockGen h [Value -> Transaction h Addr]
 newBlock = do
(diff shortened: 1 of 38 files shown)


wallet-new/test/unit/Wallet/Abstract.hs
```diff
@@ -119,6 +119,22 @@ interpret es p = go
     verify :: Wallets ws h a -> Either Text (Wallets ws h a)
     verify ws = p ws >> return ws
 
+inductiveGen 
+    :: Gen (Block h a)
+    -> Gen (Transaction h a)
```
parsonsmatt: Naturally, the "hard part" here is coming up with these generators :slightly_smiling_face:

wallet-new/test/unit/Wallet/Abstract.hs
```diff
@@ -230,3 +240,187 @@ utxoRestrictToOurs p = utxoRestrictToAddr (isJust . p)
...
+-- list.
+dissect :: [a] -> IntMap [a]
+dissect = IntMap.fromList . zip [0..] . map pure
+
+-- | Reverse the operation of 'dissect'. Given an 'IntMap' originally
+-- representing the original index in the list pointing to the list of new
+-- items at that starting index, collapse that into a single list of
+-- elements.
+conssect :: IntMap [a] -> [a]
+conssect = concatMap snd . IntMap.toList
```
parsonsmatt: `dissect` seemed fitting -- reifying the index of the item in the list to make it easy to insert things in the right place, and then `conssect` seemed like a fun pun on that. Definitely happy to have a more reasonable name though :)

wallet-new/test/unit/Wallet/Abstract.hs
```diff
@@ -230,3 +240,187 @@ utxoRestrictToOurs p = utxoRestrictToAddr (isJust . p)
...
+-- | Once we've created our initial @['Action' h 'Addr']@, we want to
+-- insert some 'Transaction's in appropriate locations in the list. There
+-- are some properties that the inserted events must satisfy:
+--
+-- * The transaction must be after all of the blocks that confirm inputs to
+--   the transaction.
+-- * The transaction must be before the block that confirms it, if any
+--   blocks confirm it. It is not necessary that the transaction gets
+--   confirmed eventually!
+intersperseTransactions
```
parsonsmatt: The bulk of the logic is performed in this function.

edsko:
Looks nice. So if I get this right, the code currently

* Finds all `n` "our" transactions
* Picks a number `m` between `1 .. n` of transactions to insert as `NewPending` into the `Inductive`, with higher numbers more probable than lower numbers
* Choose `m` transactions to be inserted
* For each of those `m`, choose a minimum index _lo_ (all its inputs must have been confirmed) and a maximum index _hi_ (given by its current location in the chain)
* Then pick an index _i_ between _lo_ and _hi_ for each transaction
* Finally, insert all transactions at the chosen locations.

Two questions about this:

* It's not clear to me quite how these indices work. When we find the minimum and maximum indices, and then those indices, they are all in terms of the _original_ chain. As soon as we even insert a single transaciton into that chain (in the final `conssect .. insertWith .. dissect` pipeline), those indices we picked are no longer in sync with the thing that we're actually inserting into. This means that if we're inserting at indices _i_, all indices _j > i_ would now be shifted up by one, which could (I think) invalidate some of the min/max calculations you did, right?
* We're not currently inserting any transactions into the `Inductive` that never get confirmed, or? (I'm not sure.. I don't see where this would happen, but the comment you wrote does mention this case.)

Minor technical question: right now the code doesn't seem like it uses the StateT at all, could do with a ReaderT. Do you envision that this will change?

wallet-new/test/unit/Wallet/Abstract.hs
edsko: Works for me :)

parsonsmatt:
> It's not clear to me quite how these indices work. When we find the minimum and maximum indices, and then those indices, they are all in terms of the original chain. As soon as we even insert a single transaciton into that chain (in the final conssect .. insertWith .. dissect pipeline), those indices we picked are no longer in sync with the thing that we're actually inserting into. This means that if we're inserting at indices i, all indices j > i would now be shifted up by one, which could (I think) invalidate some of the min/max calculations you did, right?

I think this suggests that I should use a `newtype` over `IntMap` to make it more clear what I am doing :slightly_smiling_face: . I'm taking the list of blocks:

```
[ applyBlock_0, applyBlock_1, applyBlock_2, applyBlock_3, applyBlock_4 ]
```

and making an `IntMap` out of them where the index in the intmap is the original index of that block in the chain:

```
{ 0 -> [applyBlock_0]
, 1 -> [applyBlock_1]
, 2 -> [applyBlock_2]
, 3 -> [applyBlock_3]
, 4 -> [applyBlock_4]
}
```

Then, when we select an index between `lo` (the latest block to confirm an input in the transaction) and `hi` (the index of the block that confirms the transaction itself), we can `insertWith` at that index. Suppose we have a transaction `t` where the input was provided in block 1 and is confirmed in block 3. That means we can have `NewPending t` inserted into either index 1 or 2:

```
{ 0 -> [applyBlock_0]
, 1 -> [applyBlock_1] <> [newPending t] = [applyBlock_1, newPending t]
, 2 -> [applyBlock_2]
, 3 -> [applyBlock_3]
, 4 -> [applyBlock_4]
}
-- or,
{ 0 -> [applyBlock_0]
, 1 -> [applyBlock_1] 
, 2 -> [applyBlock_2] <> [newPending t] = [applyBlock_2, newPending t]
, 3 -> [applyBlock_3]
, 4 -> [applyBlock_4]
}
```

Then, when we finally go to `conssect` the `IntMap [Action h a]` back into a `[Action h a]`, we get:

```
{ 0 -> [applyBlock_0]
, 1 -> [applyBlock_1] 
, 2 -> [applyBlock_2] <> [newPending t] = [applyBlock_2, newPending t]
, 3 -> [applyBlock_3]
, 4 -> [applyBlock_4]
}
=>
[ (0, [applyBlock_0]
, (1, [applyBlock_1] 
, (2, [applyBlock_2, newPending t]
, (3, [applyBlock_3]
, (4, [applyBlock_4]
]
=>
[ applyBlock_0
, applyBlock_1 
, applyBlock_2, newPending t
, applyBlock_3
, applyBlock_4
]
```

Which is the order that we want this to be in :smile: 

I've just realized though that `foldr k EmptyWallet` will produce the wallet in *reverse order* as the `[Action h a]` is in oldest-first ordering. Switching to `foldl'` will fix that up :)

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -99,13 +99,6 @@ freshHash = do
     bgcFreshHash += 1
     pure i
 
--- | Returns true if the 'addrActorIx' is the 'IxAvvm' constructor.
```
parsonsmatt: Relocated to `UTxO.Context`

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -195,7 +188,7 @@ newBlock = do
 
 newChain :: Hash h Addr => BlockGen h [[Value -> Transaction h Addr]]
 newChain = do
-    blockCount <- liftGen $ choose (1, 10)
+    blockCount <- liftGen $ choose (10, 50)
```
parsonsmatt: Larger chains are more interesting, but also take longer to test.

wallet-new/test/unit/Wallet/Abstract.hs
```diff
@@ -100,6 +115,17 @@ data Inductive h a =
...
+  deriving Eq
+
+instance
+    ( Buildable (Transaction h a)
+    , Buildable (Block h a)
+    ) => Show (Inductive h a) where
+    show i = toString $ case i of
+      WalletEmpty    -> "WalletEmpty"
+      ApplyBlock b n -> sformat ("ApplyBlock (" % build % ") (" % shown % ")") b n
+      NewPending t n -> sformat ("NewPending (" % build % ") (" % shown % ")") t n
```
parsonsmatt: Is there a better way of writing `Show` instances for QuickCheck purposes?

edsko:
Thanks @parsonsmatt for the detailed explanation! I had indeed not quite understood the code. I've pushed a commit that adds that explanation as a note in the code.

The commit also addresses your question above

> Is there a better way of writing Show instances for QuickCheck purposes?

I was annoyed by this as well; I've created two new modules `Util.Buildable.Hspec` and `Util.Buildable.QuickCheck` to solve this once and for all, otherwise we were just going to run into this problem time and again.

Also pushed a commit that runs `stylish` on some of these files to bring the code in line with the style guidelines.

edsko:
@parsonsmatt Oh, by the way, made one tiny change to your test; you had 

``` haskell
         forAll genInductive $ \wallet -> conjoin
          [ walletInvariants specEmpty wallet `shouldBe` Right ()
          , walletInvariants incrEmpty wallet `shouldBe` Right ()
          , walletEquivalent specEmpty incrEmpty WalletEmpty `shouldBe` Right ()
          ]
```

this `shouldBy` here comes from hspec, and returns a hspec `Expectation`, which is really just a HUnit `Assertion`, which is really just an `IO ()`. This use of `IO` is entirely unnecessary here of course, so I think it's nicer if we avoid IO in QuickCheck where possible. I've also introduced a module `Validated` (isomorphic to `Either e a`, but with some utilities and a `Buildable` instance), so I changed your test to

``` haskell
         forAll genInductive $ \wallet -> conjoin
          [ isValidated $ walletInvariants specEmpty wallet
          , isValidated $ walletInvariants incrEmpty wallet
          , isValidated $ walletEquivalent specEmpty incrEmpty WalletEmpty
          ]
```

edsko:
Rebased on latest `develop`.

edsko:
@parsonsmatt Alright, so, I pushed an implementation of rollback also, and cleaned up some of the existing wallet code. I got that cleanup code wrong at first try, and the unit tests immediately caught that. Nice! :grin: My suggested next steps for you:

1. The way we currently use `conjoin` means that when one of the tests we conjoin fails, we get a failure but we have no idea _which_ test failed. I don't know what the best solution to this is; we can of course just avoid `conjoin` but then we might needlessly generate more chains rather than just reuse the ones we already generated. Perhaps we can use some QuickCheck features to label them?
2. Modify your generator for `Inductive` so that it also inserts transactions that never become pending (either by creating new ones or by removing existing ones from the ledger, whichever you feel is better)
3. Once that is done, start looking at rollback. First, read the wallet spec, the section I wrote on rollback (it's in the repo, I also uploaded a rendered version to slack). I already modified the existing unit tests so that the existing invariants are also verified for the wallet-with-rollback implementation, as well as equivalence to the spec, _but_ rollback comes with its own invariants (Lemmas 1, 2, 3 in section 7.1, "Specification"). Those invariants need to be translated to Haskell; in fact, you may want to generalize Lemma 3 to have an arbtirary number of transactions in between the `applyBlock` and the `rollback`. So, translate those invariants to Haskell, and then
4. Define a variation on `Inductive` that also has `Rollback` calls in there, and then verify that the properties from step (3) actually hold. You'll need to think a bit about what kind of traces with rollback  would make sense, what kind of edge cases would be useful to check in the wallet, etc. Not sure if you want to extend the existing `Inductive`, define a completely new one, or define some kind of wrapper; it doesn't really matter, use your judgement here I suppose.

If you finish with all that today and run out of stuff to do, implement section 6 from the spec (prefiltering) as a separate wallet specification, and after that section 7.5, define yet another wallet implmentation that supports all of rollback, incremental updates, and rollback. However, I suspect (1)-(4) will be sufficient for one day; if that is the case, I might continue with the wallet implementations themselves on Monday.

edsko:
@parsonsmatt actually I just realized that I didn't quite get the implementation of rollback correct; I didn't implement the change I describe in section 7.3. But I'd actually say, let's leave that bug in for the time being, because it would be great if the unit tests could actually _catch_ that bug. So let's make sure we can do that before fixing it.

edsko:
I will continue with efficient-wallet-with-support-for-rollback definition.

wallet-new/test/unit/WalletUnitTest.hs
```diff
@@ -105,15 +106,19 @@ testPureWallet = do
...
+          shouldBeValidated . walletInvariants specEmpty
+        it "incrEmpty" $
+          shouldBeValidated . walletInvariants incrEmpty
+        it "rollEmpty" $
+          shouldBeValidated . walletInvariants rollEmpty
+      describe "Wallets are equivalent after interpretation" $ do
+        it "spec and incr" $
+          shouldBeValidated . walletEquivalent specEmpty incrEmpty
+        it "spec and roll" $
+          shouldBeValidated . walletEquivalent specEmpty rollEmpty
```
parsonsmatt: hspec's `beforeAll` runs the IO action once and the result is made available to all the `it`'s beforehand. These are point-free, but it looks like `it "foo" $ \wallet -> shouldBeValidated ...` when eta expanded

edsko:
@parsonsmatt , I'm still working on reviewing this PR; my primary concern is to address some performance problems. In `BlockGen`, after each new transaction, we compute the UTxO all over again starting from the genesis, leading to `O(n^2)` computations; and then we do the same thing _again_ when we add in never-to-be-confirmed transactions when constructing the wallet. I've pushed some commits that 

1. Generalize `PreChain` a bit so that we can return more information from `BlockGen`; my plan is to use this to return the UTxO after each block, so that we don't have to recompute that later. (This part of the plan I haven't done yet.)
2. I changed BlockGen so that we don't recompute the UTxO after each transaction but simply update the running UTxO with each new transaction. (This required factoring out some functionality from `utxoLedger` in the DSL itself.)
3. I also changed `divvyUp` so that it returned a transaction rather than a set of inputs and outputs; the reason I changed this is that in `BlockGen` you were assuming a fee of 0 when updating the UTxO. I don't think that's quite correct, because that may leave the UTxO "too large". Perhaps it's not a problem in practice but I was a bit concerned about it regardless.
4. I also removed `maxFee` and instead introduced `estimateFee`; at the moment this simply returns what used to be the `maxFee`, but we should modify this because that fee depends on how many inputs and outputs we have.

But before I continue with this, I want to ask you something, because I'm confused (again) about how `intersperseTransactions` works (I know, you already explained it to me once, I apologize :grin: ). So here's what I'm confused about. We start with a bunch of blocks `b0, b1,b2, ..`. We then call dissect to get an IntMap

`[0 -> b0, 1 -> b1, 2 -> b2, ...]`

So far so good. Then we call `findOurTransactions`. This returns the set of transactions (annotated with the index `hi` of the block in which they appear) that have at least one output to an address that the wallet owns. For each of those transactions `t` then call `transactionFullyConfirmedAt` to find the index `lo` of the block in which the (last) input of those tranasctions gets confirmed, and then finally pick an index `lo <= i <= hi` and finally "insert" it into the intmap, ending up with something like

`[0 -> b0, 1 -> b1 <> [NewPending t], 2 -> b2, ...]`

I have a number of questions about this; let me state them in order of low importance to high importance (but in reverse order of how likely I am to just have misunderstood your code :slightly_smiling_face: ):

1. Suppose `hi` is 4, i.e., transaction `t` got included in the block chain in block `b4`. Then the maximum index at which we want to insert `t` is 3, right, not 4? Otherwise we'd end up with

`[..., 4 -> b4 <> NewPending t, ...]`

where `t` would be made pending _after_ it has already been included into the blockchain. (Mind that `choose` from QuickCheck uses _inclusive_ bounds; `choose (0, 5)` returns an `0 <= i <= 5`.)

2. It is possible that the inputs of the transaction don't appear in the blockchain _at all_ (but are instead in the genesis block). I suspect that this is why you needed to have `transactionFullyConfirmedAt` return a `Maybe` and then call `mapMaybe` in  `intersperseTransactions`; but conceptually speaking this should _never_ be `Nothing` right? 

3. Closely related to (2), at what initially set me off thinking about all of this -- your current representation makes it impossible to insert pending transaction _before_ the first block. But this is not a case that we want to rule out.

4. Finally, and most importantly, are `findOurTransactions` even the right transactions to be inserted as pending? We want to insert transactions that the wallet "creates"; but `findOurTransactions` finds transactions that transfer funds _to_ the wallet, which will mostly be created by _other_ actors (with the sole exception of "change"). Shouldn't we look for transactions where all the _inputs_ are (all) "signed" by the wallet? (There is no notion of signing in the DSL of course; but the equivalent is: where all the inputs _refer_ to outputs which go to an address that belongs to the wallet.)

parsonsmatt:
> Suppose hi is 4, i.e., transaction t got included in the block chain in block b4. Then the maximum index at which we want to insert t is 3, right, not 4? Otherwise we'd end up with ...

Correct! I mistakenly thought that `choose` was exclusive on the upper end. I probably had some other library in mind :\

> It is possible that the inputs of the transaction don't appear in the blockchain at all (but are instead in the genesis block). I suspect that this is why you needed to have transactionFullyConfirmedAt return a Maybe and then call mapMaybe in intersperseTransactions; but conceptually speaking this should never be Nothing right?

That's right --we could use the non-Maybe functions; after all, we are plucking the transactions *out* of the blockchain, so it might make sense to ignore the partiality here.

> Closely related to (2), at what initially set me off thinking about all of this -- your current representation makes it impossible to insert pending transaction before the first block. But this is not a case that we want to rule out.

True! Is that case making a transaction using the boot as input?

> Finally, and most importantly, are findOurTransactions even the right transactions to be inserted as pending?

Good catch! You're right, we are currently finding transactions that fund our wallet. So we'd instead be looking at `trIns`, iterating over the Set, doing a `inpSpentOutput`, and collecting transactions where our address is a member of the resulting set.

edsko:
@parsonsmatt thanks for the feedback and the changes. I think it's not entirely correct yet but I'm not sure actually why the tests aren't catching it. I will need to investigate more. 

Mind: I've rebased this (and force-pushed) on latest `develop`.

edsko:
@parsonsmatt So I've isolated and fixed the following problems with the `Inductive` generation:

* New pending transactions were inserted using `Map.insertWith (<>)`, but that inserts them _before_ the block, rather than after. Changed to `Map.insertWith (flip (<>))`.
* After your previous fix, pending transactions were inserted "between" the blocks in which their (last) inputs became available and the block in which they were confirmed; but this would break for transactions in which those two blocks happen to coincide (i.e., transactions that are confirmed in the same block as their inputs). This would lead to transactions being inserted in `(lo, hi - 1)`, for `lo == hi`, i.e., in `lo - 1`, which is not correct.  We now skip such transactions (which is the only thing we _can_ do; such transactions could never be pending from the point of view of this wallet).
* Also in your previous fix the generation of pending transactions was fixed to look for transactions _from_ addresses owned by the wallet rather than transactions _to_ these addresses, but it was only checking whether _some_ of the addresses of the transactions were owned by the wallet rather than _all_.

The 'Inductive' generation is still broken even with these fixes however. Since there is no connection between the pending transactions inserted in the first pass, and the "never to be confirmed" transactions in the second pass (in `synthesizeTransactions`), the latter pass may be using inputs that are
unavailable because they are used by pending transactions inserted previously. 

Let me know if any of the above rings incorrect to you. If not, I will continue and try to fix  `synthesizeTransactions` (I know you're busy with the V1 release), as well as try to address the performance problems.

edsko:
@parsonsmatt Pushed a commit that fixes the problem with `synthesizeTransactions`. Somewhat shockingly, we now (very occassionally) get that one of the wallet invariants gets violated! :-o Don't know why yet, must investigate.

parsonsmatt:
> New pending transactions were inserted using Map.insertWith (<>), but that inserts them before the block, rather than after. Changed to Map.insertWith (flip (<>)).

Too funny, I'd just posted on reddit asking why `Map.insertWith` was confusing :)

> Also in your previous fix the generation of pending transactions was fixed to look for transactions from addresses owned by the wallet rather than transactions to these addresses, but it was only checking whether some of the addresses of the transactions were owned by the wallet rather than all.

I'm confused about this point (obviously), so I'm going to rubber duck my thought:

```
in0 --\                   /-- out0
in1 --|\__[transaction]__/|-- out1 
in2 --|/       |         \|-- out3
in3 --/        +-- change \-- out4
```

Here's a stupid diagram of a transaction: it consists of (among other things) a couple of inputs that feed into it and a bunch of outputs that use the value in the inputs.
The change is equal to the sum of inputs minus the sum of outputs, such that `change + sum outputs == sum outputs`.

We are looking for transactions that we could have feasibly submitted.
For us to be able to participate in a transaction, then at least one of the inputs (and, potentially/trivially, all of the inputs) must originate with one of our addresses.

In the case where we have a transaction like:

```
inp1 (ours) --+-[transaction]
              |
inp2 (nope) --+
```

where it has two inputs, from differing accounts:

- Is this possible/allowed? 
    - If not, then it should be the case that any transaction with at least one input belonging to us must have *all* inputs belonging to us.
    - If so, then how are they created, if we must own all of the address in order to propose the transaction?

edsko:
> (..) where it has two inputs, from differing accounts:

> Is this possible/allowed?
> * If not, then it should be the case that any transaction with at least one input belonging to us must have all inputs belonging to us.
> * If so, then how are they created, if we must own all of the address in order to propose the transaction?

In _principle_ such a transaction would be conceivable, as wallets could "cooperate"  to sign inputs. As far as the blockchain is concerned, this is indeed a valid transaction. In practice however this doesn't happen, at least, not for the current wallet. The wallet spec states the invariant that all inputs for submitted pending transaction must be in the wallet's UTxO; that's another way of saying that all "input addresses" (bit of a misnomer) must be owned by the wallet.

edsko:
Rebased on latest `develop`.

edsko:
@adinapoli-iohk Ready for merging if and when CI turns green.

parsonsmatt:
Thanks @edsko :) I appreciate the corrections!

edsko:
Rebased on `develop`.

parsonsmatt:
Finally, the CI is green. Merging in before dark magic undoes this glorious moment

adinapoli-iohk:
> Finally, the CI is green. Merging in before dark magic undoes this glorious moment

Hah, yay, well done @parsonsmatt!

PR #2547, opened 2018-02-20
[CT-55] Test "Pending transactions aren't removed with block application

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CT-55

Testing the basic conditions under which the code should work, refactored the code in order to be able to test it (easily?).

Should be easy/easier to test other states now.

Please check that I didn't do something wrong while moving the code around :pray:

As discussed with @adinapoli-iohk, this PR was opened on `develop` so the PR on `release` can be closed - https://github.com/input-output-hk/cardano-sl/pull/2534


wallet/src/Pos/Wallet/Web/Pending/Types.hs
```diff
@@ -115,7 +115,7 @@ instance Buildable (SecureLog PtxCondition) where
 data PtxSubmitTiming = PtxSubmitTiming
     { _pstNextSlot  :: SlotId
     , _pstNextDelay :: FlatSlotId
-    } deriving (Eq, Show)
+    } deriving (Eq, Show, Generic)
```
adinapoli-iohk: Why we need the `Generic` instance here?

scripts/clean/db.sh
```diff
@@ -19,6 +19,6 @@ rm -rf wdb-testnet-public/
 rm -rf db-mainnet/
 rm -rf wdb-mainnet/
 rm -rf node-*.*key*
-rm -rf kademlia-abc.dump
-rm kademlia.dump
+#rm -rf kademlia-abc.dump
+#rm kademlia.dump
```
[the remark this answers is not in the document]
ksaric: It was because I was constantly getting "kademlia.dump deosn't exist". I can move it back, but are we even using this?

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -102,6 +104,18 @@ type TxSubmissionMode ctx m =
     , MonadWalletDB ctx m
     )
 
+
+-- | Covering the `submitAndSave` code paths (end states) with types so we can test it.
+data TxSubmissionResult
+    = TxApplying
```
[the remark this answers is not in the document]
ksaric: I'll add them.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -116,41 +130,35 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
-               `catches` handlers)
-               `onException` creationFailedHandler
-           ack <- submitTx _ptxTxAux
-           reportSubmitted ack
-
-           poolInfo <- badInitPtxCondition `maybeThrow` ptxPoolInfo _ptxCond
-           _ <- usingPtxCoords casPtxCondition ptx _ptxCond (PtxApplying poolInfo)
-           when ack $ ptxUpdateMeta _ptxWallet _ptxTxId PtxMarkAcknowledged
+
+    submissionResult <- submitAndSavePtxMocked ptx now saveTxAndResult
```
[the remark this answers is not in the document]
ksaric: Not sure if you saw, but there was a newer commit, fixing your comment from previous PR - https://github.com/input-output-hk/cardano-sl/pull/2547/commits/a0ec5863ab399cfae51b50924e61de4f026959f9

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -166,9 +174,73 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
+
+    case _ptxCond of
+
+      PtxApplying poolInfo -> do
+          let Just creationTime = poolInfo ^. thTimestamp
+          if diffTimestamp now creationTime > hour 1
+              then pure $ TxTimeoutWhenApplying poolInfo ptx
+              else pure $ TxStillApplying
+
+      _ -> do
```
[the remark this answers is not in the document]
ksaric: Erm, kind of hard to handle all the cases since there literally are two:
- `PtxApplying`
- the rest

I can list them all out, but I'm not sure how useful is that since I'll copy the last part on all the other cases. I fear we might awake the HLint monster, "reduce duplication". If you think it's more readable, I can add it.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -166,9 +174,73 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
+    pure TxApplying
+  where
+    handlers =
+        [ Handler $ \e ->
+            if isReclaimableFailure e
+                then minorError "reclaimable" (SomeException e)
+                else nonReclaimableError (SomeException e)
+
+        , Handler $ \e@SomeException{} ->
+            -- I don't know where this error can came from,
```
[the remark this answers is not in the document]
ksaric: Ok, I'll add a more readable comment.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -166,9 +174,73 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
+            -- but it's better to try with tx again than to regret, right?
+            minorError "unknown error" e
+        ]
+
+    minorError desc e = pure $ TxMinorError desc e
```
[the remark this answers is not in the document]
ksaric: Erm, what? What do you mean?

wallet/src/Pos/Wallet/Web/Pending/Types.hs
ksaric: For the `Arbitrary` instance use using `genericArbitrary`, and because it's a very useful instance in _general_.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -166,9 +174,63 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
+    nonReclaimableError e = pure $ TxNonReclaimableError e
+    creationFailedHandler = pure $ TxCreationFailed
+
+
+submitAndSaveTxApplying
```
Martoon-00: This function semantics doesn't correspond to its name )

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -166,9 +174,63 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
+    => PendingTx
+    -> Timestamp
+    -> ((TxId, TxAux) -> m TxSubmissionResult)
+    -> m TxSubmissionResult
+submitAndSavePtxStates ptx@PendingTx{..} now mSaveTx =
+
+    case _ptxCond of
+
+      PtxApplying poolInfo -> do
+          let Just creationTime = poolInfo ^. thTimestamp
```
Martoon-00: You can't write in this way, it fails in `Nothing` case

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -166,9 +174,63 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
+          if diffTimestamp now creationTime > hour 1
```
Martoon-00: Same for this condition. Previously when this is `False`, second branch of `case` was executed. Now you assign special `TxStillApplying` state which corresponds to doing nothing.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
Martoon-00: Hint: `case` supports guards as well:
```
case _ptxCond of
   PtxApplying poolInfo | 
   Just creationTime = poolInfo ^. thTimestamp,
   diffTimestamp new creationTime  > hour
```
I'm sorry in advance if referencing obvious thing

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
```diff
@@ -116,41 +130,35 @@ submitAndSavePtx submitTx PtxSubmissionHandlers{..} ptx@PendingTx{..} = do
...
-           logInfo $
-             sformat ("Pending transaction #"%build%" discarded becauce \
-                      \the 1h time limit was exceeded")
-                      _ptxTxId
-       | otherwise -> do
-           addOnlyNewPendingTx ptx
-           (saveTx (_ptxTxId, _ptxTxAux)
-               `catches` handlers)
-               `onException` creationFailedHandler
-           ack <- submitTx _ptxTxAux
```
Martoon-00: Looks like you lost everything in this function starting from this line

wallet/test/Test/Pos/Wallet/Web/Pending/SubmissionSpec.hs
```diff
@@ -0,0 +1,194 @@
...
+            describe "ToilKnown exception occurs " toilKnownLargeSpec
+            describe "ToilTooLarge exception occurs " toilTooLargeSpec
+
+-- testTime :: UTCTime
+-- testTime = parseTime True defaultTimeLocale "%Y-%m-%dT%H:%M:%S%z" "2017-03-29T11:02:57+00:00 "
+
+
+normalApplicationSpec :: (HasCompileInfo, HasConfigurations) => Spec
+normalApplicationSpec = walletPropertySpec decription $ do
+    let timestamp = Timestamp $ 1518968949 * 1000000 -- In microseconds
```
Martoon-00: As far s I understand, actual time doesn't matter.
Would be nice to set just `0` here or extract this `timestamp` somewhere in order not to call the eyes

Martoon-00:
In this PR you fully covered `submitAndSavePtx` with tests.
However, in order to close CT-55 that's not enough. What you eventually need to check is that pending states replace each other appropriately over time, so what still remains unchecked:
* Which pending states are assigned by `submitAndSavePtxStates`;
* What happens on multiple calls of `submitAndSavePtxStates` (maybe with various outcomes) over time?; are resubmissions stopped after fatal error and continued after noncritical one?;
* How does `submitAndSavePtxStates` interact with state changes in blocks tracker (https://goo.gl/mnKDbP)? Whether all scenarios of interleaving events considered? Pull request which CT-55 is intended to test resolves exactly the behaviour in one corner scenario.

I want to say, you went in a way of full testing, but you're considering only small part of resubmission logic and there still are plenty of factors which may lead things in wrong direction which should be checked.
So I suppose they should be checked, at least manually, in order to complete CT-55.

wallet/test/Test/Pos/Wallet/Web/Pending/SubmissionSpec.hs
ksaric: Yes, actual time doesn't matter, but a real value adds value to the test since the user can have an example of how the code actually works, and we test a more realistic scenario.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
ksaric: Right, but will we ever have a case where we have `PtxApplying` and no `Timestamp`?
I thought about that and it seems that if a transaction is applying, it should have a timestamp, right? Otherwise, we have a transaction that doesn't have `creationTime` which is weird, to say the least, and it should fail?

I can change it, though.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
ksaric: Propose a better name, then, and let's get this one over with.
I have no intention of changing the name 5 times guessing what you had in mind.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
ksaric: Right, but won't the transaction still be in the original state `PtxApplying` and will be picked up by the re-submission mechanism, right?

ksaric:
> I want to say, you went in a way of full testing, but you're considering only small part of resubmission logic and there still are plenty of factors which may lead things in the wrong direction which should be checked.

@Martoon-00 Not really a new information. As I explained to you in DM via Slack, the other states are hard to test and were ignored. Also, as I added in the issue itself - "One should note that if the tests cannot be written easily, the code is probably missing something".
I have no intention of rewriting this part of the code, and this is the best I can do right now. 
Writing code and testing go hand-in-hand, and when testing is ignored, code suffers.
I will test the last condition manually.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
Martoon-00: It does not submit or save anything.
How about `cancelPtxWithTimeout`?

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
Martoon-00: That's correct, any transaction tracked by re-submitter should have timestamp.
But this code is unsafe - if assumption ^ would stop holding one day or there is a bug, submission will totally break with uncertain error.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
Martoon-00: Yes, but you still will do nothing if condition is `False`, because you treat `StillApplying` differently. Or not?

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
ksaric: Ok, will throw if `Nothing`. 
If that assumption stops holding one day, we really are in trouble either way.

wallet/src/Pos/Wallet/Web/Pending/Submission.hs
ksaric: Ah, I see. You are correct, nice catch. 
Thanks for catching this, I would have been in trouble if this went through :sweat_smile:

wallet/test/Test/Pos/Wallet/Web/Pending/SubmissionSpec.hs
```diff
@@ -0,0 +1,194 @@
...
+-- stack test cardano-sl-wallet --fast --test-arguments "-m Wallet.Web.Pending.SubmissionSpec"
+spec :: Spec
+spec = withCompileInfo def $
+       withDefConfigurations $
+    describe "Wallet.Web.Pending.SubmissionSpec" $ modifyMaxSuccess (const 100) $ do
+        describe "submitAndSaveTx" $ do
+            describe "Normal application" normalApplicationSpec
+            describe "Tx timeout after an hour " txTimeoutSpec
+            describe "ToilKnown exception occurs " toilKnownLargeSpec
+            describe "ToilTooLarge exception occurs " toilTooLargeSpec
```
Martoon-00: You get e.g. `Wallet.Web.Pending.SubmissionSpec.Normal application.Normal application: 'PtxApplying -> TxApplying'`, is duplication intentional?

Martoon-00: I just loved the idea of passing description right in function which contains the test

ksaric: It's not duplication, one is `PtxApplying`, the other is `TxApplying`. `TxApplying` is the one I added (for internal state and testing), while `PtxApplying` is already existing.

Martoon-00: I mean, `describe "Normal application"` and later `walletPropertySpec "Normal application ..."`. Do you intentionally duplicate it in `describe`?

ksaric: Yes? I don't know, it seems to be following the same convention as others.

Martoon-00: Lool, okay :bug:

jcmincke:
Old wallet code

Issue #2443, opened 2018-02-03
Error in "Stack with Nix for system libraries (mixed mode)"


gmanroney:
Tried to do stack build following instructions

> nix-shell
> [nix-shell:~/cardano-sl]$ ./scripts/build/cardano-sl.sh
> 

I get the error below. I'm guessing stack should be installed on the host system but no documentation or instruction to do so. Is this correct?

> [nix-shell:/opt/cardano-sl]$ ./scripts/build/cardano-sl.sh
> Going to build:  cardano-sl-core cardano-sl-db cardano-sl-lrc cardano-sl-infra cardano-sl-update cardano-sl-ssc cardano-sl-godtossing cardano-sl-txp cardano-sl cardano-sl-auxx cardano-sl-tools cardano-sl-wallet cardano-sl-explorer
> 'wallet' flag: true
> 'explorer' flag: true
> Building cardano-sl-core
> 
> stack build --ghc-options="-DGITREV=0d010e21e3af8ff6213d7877d9bb8132764900ec -Wwarn" --test --no-haddock-deps --bench --jobs=4 --flag cardano-sl:with-wallet --no-run-tests --no-run-benchmarks --dependencies-only  cardano-sl-core
> 
> No compiler found, expected minor version match with ghc-8.0.2 (x86_64) (based on resolver setting in /opt/cardano-sl/stack.yaml).
> To install the correct GHC into /home/appadmin/.stack/programs/x86_64-linux-nix/, try running "stack setup" or use the "--install-ghc" flag. To use your system GHC installation, run "stack config set system-ghc --global true", or use the "--system-ghc" flag.
>

gmanroney:
Ok I ran 

> stack setup

 from within the nix console that seems to have worked

gmanroney:
Ok I spoke to soon. After I ran 

> [nix-shell:~/cardano-sl]$ stack setup
> [nix-shell:~/cardano-sl]$ ./scripts/build/cardano-sl.sh
> 

I got the following (some logs omitted for brevity):

> ...
> ...
> log-warper-1.2.3.1: download
> log-warper-1.2.3.1: configure
> log-warper-1.2.3.1: build
> statistics-0.13.3.0: copy/register
> Progress: 153/156
> --  While building package log-warper-1.2.3.1 using:
[...]

Issue #2223, opened 2018-01-02
Installation with scripts/build/cardano-sl.sh fails 


p-alik:
Unfortunately installation of cardano-sl with regard to [Stack with Nix for system libraries (mixed mode)](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/how-to/build-cardano-sl-and-daedalus-from-source-code.md#stack-with-nix-for-system-libraries-mixed-mode) Instruction fails.

System Information:
```
Ubuntu 16.04.3 LTS
$ nix-shell --version
nix-shell (Nix) 1.11.16
```
The same issue is in a daedalus repository: https://github.com/input-output-hk/daedalus/issues/573

```
$ nix-shell
 $ ./scripts/build/cardano-sl.sh
Going to build:  cardano-sl-core cardano-sl-db cardano-sl-lrc cardano-sl-infra cardano-sl-update cardano-sl-ssc cardano-sl-godtossing cardano-sl-txp cardano-sl cardano-sl-auxx cardano-sl-tools cardano-sl-wallet cardano-sl-explorer
'wallet' flag: true
'explorer' flag: true
Building cardano-sl-core

stack build --ghc-options="-DGITREV=57e2710549ff9079714f5b7d89c940cc38e8b6c1 -Wwarn +RTS -A256m -n2m -RTS" --test --no-haddock-deps --bench --jobs=4 --flag cardano-sl:with-wallet --no-run-tests --no-run-benchmarks --dependencies-only  cardano-sl-core

optparse-applicative-0.13.2.0: configure
optparse-applicative-0.13.2.0: build
aeson-1.1.2.0: configure
aeson-1.1.2.0: build
Progress: 2/8
--  While building package aeson-1.1.2.0 using:
      /home/palik/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2 --builddir=.stack-work/dist/x86_64-linux-nix/Cabal-1.24.2.0 build --ghc-options " -ddump-hi -ddump-to-file"
    Process exited with code: ExitFailure 1
    Logs have been written to: /opt/devel/cardano/cardano-sl/.stack-work/logs/aeson-1.1.2.0.log

    Configuring aeson-1.1.2.0...
    Building aeson-1.1.2.0...
    Preprocessing library aeson-1.1.2.0...
[...]

PR #2236, opened 2018-01-03
add nix enable flag into stack config file

p-alik:
without the flag setting build process doesn't work properly.
The PR solves the issue #2223

PR #2548, opened 2018-02-20
Stack.yaml now enables Nix to fix build of node

SebastienGllmt:
If you don't have this setting, you run into the problem mention in
#2223 #2443 #1784 (and maybe more)

This fixed the issue for me also (building scripts/build/cardano-sl.sh)-. Is there any reason not to have this setting?

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -137,6 +137,7 @@ packages:
 
 nix:
   shell-file: shell.nix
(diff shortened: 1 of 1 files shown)


gromakovsky:
See https://github.com/input-output-hk/cardano-sl/pull/2236

PR #2549, opened 2018-02-20
[CSL-2093] + [CSL-2094] V1 `/development/x` endpoints

sectore:
This PR includes following two V1 test endpoints
* `DELETE /api/development/secret-keys` - YT issue https://iohk.myjetbrains.com/youtrack/issue/CSL-2093
* `GET /api/development/dump-wallet-state` - YT issue https://iohk.myjetbrains.com/youtrack/issue/CSL-2094

In `production` mode both test endpoints are not available and hided in generated Swagger's docs.


wallet-new/server/Cardano/Wallet/LegacyServer.hs
```diff
@@ -4,7 +4,8 @@ module Cardano.Wallet.LegacyServer where
 import           Cardano.Wallet.API
 import           Cardano.Wallet.API.V1.Migration as Migration
 
-import qualified Cardano.Wallet.API.V0.Handlers       as V0
+import qualified Cardano.Wallet.API.V0.Handlers as V0
+import qualified Cardano.Wallet.API.Dev.LegacyHandlers as Dev
```
adinapoli-iohk: What's wrong with `Development` instead ? I like long (well, reasonably long 😁 ) and explicit names.

adinapoli-iohk: So that would be `import qualified Cardano.Wallet.API.Development.LegacyHandlers as Development`

wallet-new/server/Cardano/Wallet/LegacyServer.hs
```diff
@@ -23,3 +24,4 @@ walletServer :: ( Migration.HasConfigurations
 walletServer natV0 diffusion =
          V0.handlers natV0 diffusion
     :<|> V1.handlers natV0 diffusion
+    :<|> Dev.handlers natV0
```
adinapoli-iohk: We should add a combinator like `developmentOnly` which would wrap handlers so that we cannot call them if we are not running in debug mode? Something like:

```
:<|> developmentOnly (Development.handlers natV0)
```

wallet-new/server/Cardano/Wallet/Server.hs
```diff
@@ -5,15 +5,16 @@ import Servant
...
+import qualified Cardano.Wallet.API.V0 as V0
 import qualified Cardano.Wallet.API.V1.Handlers as V1
 
 -- | Serve the REST interface to the wallet
 --
 -- NOTE: Unlike the legacy server, the handlers will not run in a special
 -- Cardano monad because they just interfact with the Wallet object.
 walletServer :: ActiveWallet -> Server WalletAPI
-walletServer w = v0Handler :<|> V1.handlers w
+walletServer w = v0Handler :<|> V1.handlers w :<|> Dev.handlers
```
adinapoli-iohk: Ditto here.

wallet-new/src/Cardano/Wallet/API.hs
```diff
@@ -24,13 +27,37 @@ import qualified Cardano.Wallet.API.V1 as V1
...
+
 type WalletAPI
-    =    "api" :> Tags '["V0 (Deprecated)"]
-               :> V0.API
-    :<|> "api" :> "v1"
-               :> Tags '["V1"]
-               :> V1.API
+    =     V0_API
+    :<|>  V1_API
+    :<|>  DEV_API
```
adinapoli-iohk: This seems like a gratuitous refactoring -- what was wrong with the code before? It was nice and descriptive, now we added another layer of type alias a programmer needs to unwrap to understand what the API is doing (my 2 cents anyway).

wallet-new/src/Cardano/Wallet/API.hs
```diff
@@ -24,13 +27,37 @@ import qualified Cardano.Wallet.API.V1 as V1
...
+
+type WalletSwaggerApi
+    =     V0_API
+    :<|>  V1_API
 
 walletAPI :: Proxy WalletAPI
 walletAPI = Proxy
+
+swaggerWalletApi :: Proxy WalletSwaggerApi
+swaggerWalletApi = Proxy
```
adinapoli-iohk: This shouldn't belong here. There is a reason why `Swagger` stuff is in `server`, and that's because I've tried to separate the documentation concerns (which lives in executable code) vs purely API concerns (which live in `src`) 😉

wallet-new/src/Cardano/Wallet/API/Dev.hs
```diff
@@ -0,0 +1,9 @@
+module Cardano.Wallet.API.Dev where
+
+
+import           Servant ((:>))
+
+import           Cardano.Wallet.API.Types
+import qualified Cardano.Wallet.API.Dev.Test as Test
+
+type API = Tags '["Test"] :> Test.API
```
adinapoli-iohk: We need to be consistent with names. If this is a development API, we shouldn't also throw the word "test" in, as that conveys a different endeavour.

wallet-new/src/Cardano/Wallet/API/Dev/Test.hs
```diff
@@ -0,0 +1,10 @@
+module Cardano.Wallet.API.Dev.Test where
+
+import           Servant.API.ContentTypes (OctetStream)
+import           Cardano.Wallet.API.Response (WalletResponse)
+import           Pos.Wallet.Web.Methods.Misc (WalletStateSnapshot)
+
+import           Servant
+
+type API = "dump-wallet-state" :> Summary "Dump wallet state."
+                               :> Get '[OctetStream] (WalletResponse WalletStateSnapshot)
```
adinapoli-iohk: I thought this returned a JSON -- isn't this the case? I _think_ (but I might be wrong) that simply returning a JSON is nicer and more development friendly than dumping a file on disk.

wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs
```diff
@@ -25,3 +28,30 @@ instance Arbitrary CWalletRedeem
 instance Arbitrary CPaperVendWalletRedeem
 instance Arbitrary CInitialized
 instance Arbitrary NewBatchPayment
+
+instance Arbitrary WalletStateSnapshot where
+    arbitrary = WalletStateSnapshot <$> arbitrary
```
adinapoli-iohk: My gut feeling is that if we do not use `Swagger` in `src`, we don't need this `Arbitrary` instance at all.

wallet/src/Pos/Wallet/Web/Methods/Misc.hs
```diff
@@ -192,6 +194,9 @@ instance MimeRender OctetStream WalletStateSnapshot where
 instance Buildable WalletStateSnapshot where
     build _ = "<wallet-state-snapshot>"
 
+instance Default WalletStateSnapshot where
+    def = WalletStateSnapshot { wssWalletStorage = def }
```
adinapoli-iohk: Why do we need a `Default` instance for `WalletStateSnapshot`?

wallet-new/server/Cardano/Wallet/LegacyServer.hs
sectore: Oh maaaan, what's wrong with `Dev` instead of `Development`? `Dev` is self-explaining, too. And we do use short modules names everywhere (e.g. `V0` instead of `Version0`). Anyway, I don't care about names and will change it if you really want.

wallet-new/server/Cardano/Wallet/LegacyServer.hs
sectore: Good idea!

wallet-new/src/Cardano/Wallet/API.hs
sectore: The first idea behind was to split API into different parts, so that we can hide endpoints in development mode. But I think your idea https://github.com/input-output-hk/cardano-sl/pull/2549/files#r169258990 is more elegant.

wallet-new/src/Cardano/Wallet/API.hs
sectore: We might not need it anymore...

wallet-new/src/Cardano/Wallet/API/Dev.hs
sectore: If you care about it, please make a suggestion how to name it.

wallet-new/src/Cardano/Wallet/API/Dev/Test.hs
sectore: V0 endpoint returns a `OctetStream` (see https://github.com/input-output-hk/cardano-sl/blob/develop/wallet/src/Pos/Wallet/Web/Api.hs#L153-L156), so I'm not sure why we should change it.

wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs
sectore: Where would you define instance of `Arbitrary WalletStateSnapshot` instead?

wallet/src/Pos/Wallet/Web/Methods/Misc.hs
sectore: We don't need it anymore. The first idea was to use it to create a `Arbitrary WalletStateSnapshot` from it (similar as you suggested today https://input-output-rnd.slack.com/archives/C819S481Y/p1519120198000571?thread_ts=1519115554.000460&cid=C819S481Y )

wallet-new/server/Cardano/Wallet/LegacyServer.hs
adinapoli-iohk: >  Anyway, I don't care about names and will change it if you really want.

Not really a dictator, just thinking out loud here 😁  Maybe the middle ground would be to at least name the module (that you import once) `Development` but use `Dev` to qualify the imports (which is something done frequently)?

wallet-new/src/Cardano/Wallet/API/Dev/Test.hs
adinapoli-iohk: > V0 endpoint returns a OctetStream

I suspect because the `WalletStorage` could be huge and they didn't want to display it all as a JSON. But if we go down the development route, we shouldn't really call this in production (where the state stored in the `WalletStorage` can be huge), so I think we can afford the luxury of showing just a nice JSON. What do you think?

wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs
adinapoli-iohk: That was a red herring, didn't realise `Example` has an explicit constrain on `Arbitrary` (for good reasons). My bad!

wallet-new/src/Cardano/Wallet/API/Dev.hs
adinapoli-iohk: I suggest we simply have the entire API defined at the top-level here (so we move the content of `Dev.Test` into this module, and we get rid of the artificial subsplit. Finally, we change the `Tags` content to be just `Development`, for the sake of consistency. What do you think?

wallet-new/src/Cardano/Wallet/API/Dev/Test.hs
sectore: Let's ask the Daedalus team as an user of this endpoint ^:
@darko-mijic Would your team still happy, if the `V1` endpoint `/api/development/dump-wallet-state` (corresponding `V0` endpoint is `/api/test/state`) will return a `JSON` and not a `OctetStream` we had before?

adinapoli-iohk: Great idea -- probably we also want to page @DominikGuzei.

wallet-new/server/Cardano/Wallet/LegacyServer.hs
```diff
@@ -4,8 +4,11 @@ module Cardano.Wallet.LegacyServer where
 import           Cardano.Wallet.API
 import           Cardano.Wallet.API.V1.Migration as Migration
 
-import qualified Cardano.Wallet.API.V0.Handlers       as V0
+import qualified Cardano.Wallet.API.V0.Handlers as V0
+-- import qualified Cardano.Wallet.API.Development.Helpers as Dev
```
adinapoli-iohk: We should remove this commented-out code.

wallet-new/test/DevelopmentSpec.hs
```diff
@@ -0,0 +1,78 @@
...
+import           Test.Pos.Wallet.Web.Mode (walletPropertySpec)
+import           Test.Hspec (Spec, describe)
+import           Test.Hspec.QuickCheck (modifyMaxSuccess)
+import           Test.Pos.Configuration (withDefConfigurations)
+
+import           Cardano.Wallet.API.Development.LegacyHandlers (deleteSecretKeys)
+import           Cardano.Wallet.Server.CLI (RunMode (..))
+import           Servant
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
```
adinapoli-iohk: No need to write this -- CI won't fail on this as this is just a warning, AFAIK.

wallet-new/test/DevelopmentSpec.hs
```diff
@@ -0,0 +1,78 @@
...
+
+spec :: Spec
+spec =
+    withCompileInfo def $
+    withDefConfigurations $
+    describe "development endpoint" $
+        describe "secret-keys" $ modifyMaxSuccess (const 10) deleteAllSecretKeysSpec
+
+deleteAllSecretKeysSpec :: (HasCompileInfo, HasConfigurations) => Spec
```
adinapoli-iohk: These tests are fine for now but notice how they are not exerting the API externally, but relying on internal functions from the old wallet, which we are in the process of removing.

wallet/cardano-sl-wallet.cabal
```diff
@@ -98,6 +98,9 @@ library
                       Pos.Util.BackupPhrase
                       Pos.Util.Mnemonics
 
+                      -- shared test sources needed by tests of wallet-new
+                      Test.Pos.Wallet.Web.Mode
```
adinapoli-iohk: I think that - considering what I have said above about the `DevelopmentSpec` relying too much on the internals of the old wallet, we should move the `DevelopmentSpec` here and just reuse the code in `wallet-new`. This way we clearly separate the concerns and the day the `wallet` project will be decommissioned we will be forced to revisit our development specs anyway, and rewrite them with the new wallet code. What do you think?

wallet-new/test/DevelopmentSpec.hs
```diff
@@ -0,0 +1,78 @@
...
+    -- TODO: Use an arbitrary instance of `BackupPhrase` if available
+    let phrase = BackupPhrase [ "truly", "enact", "setup", "session"
+                              , "near", "film", "walk", "hard"
+                              , "ginger", "audit", "regular", "any"
+                              ]
+    walletPropertySpec "does remove all secret keys in debug mode mode" $ do
+        sKey <- either  (error "creating a secret key failed")
+                        (pure . fst)
+                        $ safeKeysFromPhrase mempty phrase
+        lift $ addSecretKey sKey
```
adinapoli-iohk: @sectore Have you checked that running these specs will write and read secrets from a file which won't conflict with a production one? What I mean by that is that we should make sure that the `HasKeysContext` defined for the `WalletWebTestMode` monad is not using as keyfile `secrets.key`, as that is the same filename for "production" keys. Prudence is never enough..

wallet-new/test/DevelopmentSpec.hs
sectore: You do need that, because the CI was failing yesterday.

wallet-new/test/DevelopmentSpec.hs
sectore: > relying on internal functions from the old wallet
Sure, because this ^ and most of others `V1` endpoints are currently based on V0 implementation. They are mostly just wrapper around `V0` functions. I don't see a way to test those in other way as long we do have these `V0` dependencies.

wallet/cardano-sl-wallet.cabal
sectore: > relying too much on the internals of the old wallet

Again, that's we currently do with most of all `V1` endpoints. As I've understood that's ok, because the _first_ step is to wrap `V0` endpoints with an `V1` interface (API) and the _second_ step is to re-write the internal thing to be independent from `V0`. Or do I miss something here? Feel free to discuss that in Slack if there are more information to share.

wallet-new/test/DevelopmentSpec.hs
sectore: The endpoint is removing the `secrets.key` in `debug` mode only and not in `production` mode. That's why we do have this test to test this "feature".

wallet-new/test/DevelopmentSpec.hs
adinapoli-iohk: You're right, my bad, I thought it was a different warning. But now that you make me think about it, where is the code duplication stemming from? Any way we can make hlint happy by reducing it?

wallet/cardano-sl-wallet.cabal
adinapoli-iohk: > Or do I miss something here?

Well, there is a substantial difference in using old code within some internal handler vs writing test code which is coupled with it, as in the former case you can test the V1 endpoint as a blackbox (thus completely independent from how the internal implementation is delivered) whereas in the second case you will have to tear down the entire test code once you get rid of the old `wallet` code. But for now let's be pragmatic, as it would be overkill spend more time on these internal endpoints.

> Please note that DevelopmentSpec is testing functions of wallet-new, especially deleteSecretKeys

You're right, I didn't appreciate we were querying the handler directly -- in which case, you made the right decision and I'm happy with the status quo, apologies 😉

wallet-new/test/DevelopmentSpec.hs
adinapoli-iohk: What I mean is that I'm not very comfortable in having test code which is deleting a file named exactly the same way it's named in production. It's easy to accidentally screw up a check for "production vs debug" and accidentally delete your own wallet's `secret.key`. It's probably paranoia (as you would never run a production node on the same folder you run the tests), but all I'm saying is that I would have been much happier if our tests would have run in a monad where the key file was named differently (dunno, `secrets-test.key`). Not a show-stopper, but more an out loud thought :)

wallet-new/test/DevelopmentSpec.hs
sectore: `hlint` did care about following duplicated lines of code:
```
sKey <- either  (error "creating a secret key failed")
                         (pure . fst)
                         $ safeKeysFromPhrase mempty phrase
```

Here https://github.com/input-output-hk/cardano-sl/blob/938a4faca8a15e77ece01c03bdee82a89adb74b2/wallet-new/test/DevelopmentSpec.hs#L53-L55
and here https://github.com/input-output-hk/cardano-sl/blob/938a4faca8a15e77ece01c03bdee82a89adb74b2/wallet-new/test/DevelopmentSpec.hs#L67-L69

I would suggest to ignore that duplication and just using that annotation to make `hlint` happy.

wallet-new/test/DevelopmentSpec.hs
sectore: It might be a valid point ^. However, I'm not sure why `secret.key` could be deleted by running this `DevelopmentSpec`. At least I can't see any reference of `secret.key` in test code (either in `V0` nor in `V1` tests).

PR #2550, opened 2018-02-20
[CSL-2313] Bump report server to 0.4.7

volhovm:


rvl:
Builds and works for me locally.

Please reference `[CSL-2313]` when merging.

PR #2551, opened 2018-02-20
[CSL-2193] Fix misplaced configuration options

volhovm:
Some options were misplaced, some were obsolete. I recommend reviewing per-commit.

@avieth do you know anything about why `propagationQueueSize` and `defaultPeers` are unused? Is it a bug?

### DIFF
--- a/block/src/Pos/Block/Logic/Util.hs
+++ b/block/src/Pos/Block/Logic/Util.hs
@@ -22,10 +22,11 @@ import qualified Data.List.NonEmpty as NE
 import           Formatting (int, sformat, (%))
 import           System.Wlog (WithLogger)
 
+import           Pos.Block.Configuration (HasBlockConfiguration, fixedTimeCQ)
 import           Pos.Block.Slog.Context (slogGetLastSlots)
 import           Pos.Block.Slog.Types (HasSlogGState)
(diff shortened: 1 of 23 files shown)


block/src/Pos/Block/Logic/Util.hs
```diff
@@ -133,7 +134,13 @@ calcOverallChainQuality =
 -- restrictive at all.
 -- 3. We are able to determine which slot started 'fixedTimeCQ' ago.
 calcChainQualityFixedTime ::
-       forall ctx m res. (Fractional res, MonadSlots ctx m, HasConfiguration, HasSlogGState ctx)
+       forall ctx m res.
+       ( Fractional res
+       , MonadSlots ctx m
+       , HasConfiguration
```
gromakovsky: Is `HasConfiguration` still needed?

gromakovsky:
`defaultPeers` was added by me in 2016 in 42f1156bdc94e8bad5da37b93f35b10970af7a9b (CSL-83). Networking code was significantly changed since then, so I am not surprised that it's not used anymore.

block/src/Pos/Block/Logic/Util.hs
volhovm: `HasGeneratedSecrets` is, so yes.

PR #2555, opened 2018-02-21
Update monads.md

nahern:
minor style edits.

### DIFF
--- a/docs/monads.md
+++ b/docs/monads.md
@@ -1,7 +1,7 @@
-# Guidelines for Effect Handling in Cardano SL
+# Guidelines for Monadic Effect Handling in Cardano SL
 
 A monad comes with two basic operations, ``return`` and ``>>=``, but this
-interface is barely useful on its own. Monadic effects extend the ``Monad``
+interface is rarely used on its own. Monadic effects extend the ``Monad``
 interface with additional operations (such as ``ask``, ``local``, ``get``,
 ``put``, ``throwError``, ``liftIO``, ``tell``, etc).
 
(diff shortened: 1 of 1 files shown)


docs/monads.md
```diff
@@ -29,11 +29,11 @@ article.
 
 ## Intrinsic effects and capabilities
 
-We classify the effects into two categories:
+We classify effects into two categories:
 
 * _intrinsic effects_ provide operations enabled by the additional structure of
   the monad itself. For example, the intrinsic effect of `StateT` is adding
-  state to a computation, `ExceptT` adds alternative exit path, `ConduitM` adds
+  state to a computation, `ExceptT` adds the alternative exit path, `ConduitM` adds
```
Pastafarianist: I believe the article should be indefinite here, since this is where this path is introduced for the first time.

docs/monads.md
```diff
@@ -70,7 +70,7 @@ transformer that adds another intrinsic effect (`StateT`), but reuse the
 existing `IO` base monad and represent the mutable state by passing an `IORef`
 in the `ReaderT` context.
 
-### Identifying effect class
+### Identifying the effect class
```
Pastafarianist: I don't believe the article should be added here at all.

docs/monads.md
```diff
@@ -306,7 +306,7 @@ use.
 
 ## Potentially impure code
 
-This is the code that can be instantiated to be either pure or impure:
+The following code can be instantiated to be either pure or impure:
```
int-index: The original meaning was to clarify what the term "potentially impure code" means. Potentially impure code is the code that can be instantiated to be either pure or impure. I guess ending this sentence with a colon was confusing.

adinapoli-iohk:
There is no `docs/monads.md` file anymore.

PR #2556, opened 2018-02-21
make block benchmarks compile

avieth:


avieth:
Already done in #2554 oops

PR #2560, opened 2018-02-22
[CSL-2167] another one bump for acid-state

DanielRrr:


gromakovsky:
The only change here is this hotfix: https://github.com/serokell/acid-state/pull/7/files

PR #2561, opened 2018-02-22
[CSL-2322] Merge release/1.1.0

gromakovsky:


auxx/src/Command/Rollback.hs
```diff
@@ -52,7 +52,7 @@ rollbackAndDump numToRollback outFile = withStateLock HighPriority "auxx" $ \_ -
                     flattenTxPayload $ mainBlock ^. mainBlockTxPayload
             let allTxs :: [TxAux]
                 allTxs = concatMap extractTxs blunds
-            liftIO $ BSL.writeFile outFile (serialize allTxs)
+            liftIO $ BSL.writeFile outFile (encode allTxs)
```
Pastafarianist: I was surprised to find out that `aeson` does not export anything similar to `yaml`'s [`encodeFile`](https://hackage.haskell.org/package/yaml-0.8.2/docs/Data-Yaml.html#v:encodeFile).

Issue I file, of course, will not.

PR #2566, opened 2018-02-22
use log-warper with important bug fix

avieth:
The bug could cause asynchronous exceptions to be squelched and, for
instance, a fatal exception to not kill the program.

The fix was merged into log-warper-1.8.10, but that includes many new
dependency versions which are not compatible with cardano-sl's current
stack configuration. This patch points to a backport of the log-waprer
fix.


avieth:
@gromakovsky ping for review

PR #2567, opened 2018-02-22
[CSL-2245] Async worker to restore & sync a wallet

adinapoli-iohk:
This (untested) PR ships into `feature/CBR-90` the ability of perform fully asynchronous wallet restorations.

Edit: probably needs rebasing since it merges changes from `develop`.


adinapoli-iohk:
I think this is ready to merge into `feature/CBR-90` and I will pick up the CI fallouts as part of CSL-2246.

PR #2568, opened 2018-02-22
[CSL-2292]  Add supplimentary functionality for testing

hithroc:
This adds two thnigs:
* `send-with-fake-input` command in auxx, to be able to send transactions with nonexistant inputs
* Make node create blocks signed by a wrong slot leader

### DIFF
--- a/auxx/src/Command/Tx.hs
+++ b/auxx/src/Command/Tx.hs
@@ -39,7 +39,7 @@ import           Pos.Communication (SendActions, immediateConcurrentConversation
 import           Pos.Core (BlockVersionData (bvdSlotDuration), IsBootstrapEraAddr (..),
                            Timestamp (..), deriveFirstHDAddress, makePubKeyAddress, mkCoin)
 import           Pos.Core.Configuration (genesisBlockVersionData, genesisSecretKeys)
-import           Pos.Core.Txp (TxAux, TxOut (..), TxOutAux (..), txaF)
+import           Pos.Core.Txp (TxAux (..), TxIn (..), TxOut (..), TxOutAux (..), txaF, txInputs)
 import           Pos.Crypto (EncryptedSecretKey, emptyPassphrase, encToPublic, fakeSigner,
                              safeToPublic, toPublic, withSafeSigners)
 import           Pos.Txp (topsortTxAuxes)
@@ -183,8 +183,9 @@ send
     => SendActions m
     -> Int
     -> NonEmpty TxOut
+    -> Bool
     -> m ()
-send sendActions idx outputs = do
+send sendActions idx outputs fakeInputs = do
     CmdCtx{ccPeers} <- getCmdCtx
     skey <- takeSecret
     let curPk = encToPublic skey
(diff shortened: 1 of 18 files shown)


auxx/src/Command/Tx.hs
```diff
@@ -199,7 +202,10 @@ send sendActions idx outputs = do
         let getSigner = fromMaybe (error "Couldn't get SafeSigner") . flip HM.lookup addrSig
         -- BE CAREFUL: We create remain address using our pk, wallet doesn't show such addresses
         (txAux,_) <- lift $ prepareMTx getSigner mempty def (NE.fromList allAddresses) (map TxOutAux outputs) curPk
-        txAux <$ (ExceptT $ try $ submitTxRaw (immediateConcurrentConversations sendActions ccPeers) txAux)
+        let txAux' =
+                if not fakeInputs then txAux
+                else txAux { taTx = taTx txAux & txInputs %~ fmap (\x -> x { txInIndex = 100500 }) }
```
gromakovsky: Btw, what will `x { txInIndex = 100500 }` do if `x` is `TxInUnknown`?

auxx/src/Command/Tx.hs
```diff
@@ -211,6 +217,18 @@ send sendActions idx outputs = do
...
+    :: forall m. MonadAuxxMode m
+    => SendActions m
+    -> TxOut -> m ()
+sendWithoutWitnesses sendActions output = do
+  let fakeTxIn = TxInUtxo (unsafeHash ()) 1
+      fakeTx = UnsafeTx (fakeTxIn:|[]) (output:|[]) def
+      fakeTxAux = TxAux fakeTx V.empty
+
+  CmdCtx {ccPeers} <- getCmdCtx
+  void $ submitTxRaw (immediateConcurrentConversations sendActions ccPeers) fakeTxAux
```
gromakovsky: Hm, I think it makes more sense to make exactly 1 intentional error in 1 command.
`send-with-fake-input` sends a transaction with fake input, ok.
And then `send-without-witness` sends a transaction with fake input and without witnesses. We shouldn't rely on the fact that witnesses correctness is checked before input correctness (moreover, I guess it's not the case). And if input correctness is checked first, this command will fail the same way as `send-with-fake-input`.

P. S. Also it probably would be better to use invalid witness instead of using no witnesses at all.

block/src/Pos/Block/Behavior.hs
```diff
@@ -0,0 +1,73 @@
...
+data BlockBehavior = BlockBehavior
+    { -- | Block header forging settings
+      bbForgeHeader :: !ForgeHeaderParams
+    }
+    deriving (Eq, Show, Generic)
+
+data ForgeHeaderParams
+    = HeaderNormal      -- ^ Do not forge anything
+    | HeaderWrongLeader -- ^ Create blocks header with a wrong slot leader
+                        --    signature
```
gromakovsky: This comment is ambiguous. It can mean that we put wrong leader or wrong signature (or maybe both).

block/src/Pos/Block/Behavior.hs
```diff
@@ -0,0 +1,73 @@
...
+    deriving (Eq, Show)
+
+type HasBlockBehavior = Given BlockBehavior
```
gromakovsky: Hm, why did you decide to use `reflection` instead of passing it in `ReaderT` context? :/
We mostly use `ReaderT`. For example, `BehaviorConfig` is currently stored in `ReaderT` context.
AFAIU, there are only two good reasons to use `reflection`:
1. If there is a lot of existing code which uses some constant `x` and you want to make `x` configurable.
2. If you use `x` in some instance (e. g. `Bi`) where you want to know value of this `x`, but can't use any external information and `x` is not constant.

gromakovsky: So, AFAIU, now there are two ways to access `BlockBehavior`: take it from reader context or take it using `reflection`. And it's not guaranteed that they will be the same (one can use `local` to modify one of them, but the one in `reflection` can not be modified). Sounds like a bad design.

gromakovsky: In #2562 I added `UpdateBehavior` which is stored only in reader and it was quite easy.

auxx/src/Command/Tx.hs
hithroc: It will crash. I was unaware that there's TxInUnknown.

PR #2571, opened 2018-02-23
[CSM-551] Split request and response in servant logs

Martoon-00:
Rebased #1874


Martoon-00:
Ah, it's really exactly the same as in #1874 description, with only difference that numeration starts from `1`, not `0` :slightly_smiling_face:

PR #2573, opened 2018-02-23
[CSL-2043] Add tmeasure to critical places

volhovm:
I was using `tMeasure` function in my local tests and decided it's actually pretty nice to have it in master. I've also added usages to some critical places -- LRC, recovery, headers/blocks/txp listeners.

### DIFF
--- a/util/Pos/Util/Util.hs
+++ b/util/Pos/Util/Util.hs
@@ -64,6 +64,8 @@ module Pos.Util.Util
        , divRoundUp
        , sleep
 
+       , tMeasureLog
+       , tMeasureIO
        ) where
 
 import           Universum
@@ -83,7 +85,7 @@ import qualified Data.Map as M
 import           Data.Ratio ((%))
 import qualified Data.Semigroup as Smg
 import qualified Data.Serialize as Cereal
-import           Data.Time.Clock (NominalDiffTime, UTCTime)
+import           Data.Time.Clock (NominalDiffTime, UTCTime, diffUTCTime, getCurrentTime)
 import           Data.Time.Clock.POSIX (posixSecondsToUTCTime)
 import           Data.Time.Units (Microsecond, toMicroseconds)
 import qualified Ether
@@ -93,7 +95,7 @@ import qualified Language.Haskell.TH as TH
 import qualified Prelude
 import           Serokell.Util (listJson)
 import           Serokell.Util.Exceptions ()
-import           System.Wlog (LoggerName, WithLogger, logError, logInfo, usingLoggerName)
+import           System.Wlog (LoggerName, WithLogger, logDebug, logError, logInfo, usingLoggerName)
 import qualified Text.Megaparsec as P
 
 ----------------------------------------------------------------------------
@@ -388,3 +390,26 @@ median l = NE.sort l NE.!! middle
 -}
 sleep :: MonadIO m => NominalDiffTime -> m ()
 sleep n = liftIO (threadDelay (truncate (n * 10^(6::Int))))
+
+-- | 'tMeasure' with 'logDebug'.
+tMeasureLog :: (MonadIO m, WithLogger m) => Text -> m a -> m a
+tMeasureLog = tMeasure logDebug
+
+-- | 'tMeasure' with 'putText'. For places you don't have
+-- 'WithLogger' constraint.
+tMeasureIO :: (MonadIO m) => Text -> m a -> m a
+tMeasureIO = tMeasure putText
+
+-- | Takes the first time sample, executes action (forcing its
+-- result), takes the second time sample, logs it.
+tMeasure :: (MonadIO m) => (Text -> m ()) -> Text -> m a -> m a
+tMeasure logAction label action = do
+    before <- liftIO getCurrentTime
+    !x <- action
+    after <- liftIO getCurrentTime
+    let d0 :: Integer
+        d0 = round $ 10000 * toRational (after `diffUTCTime` before)
+    let d1 = d0 `div` 10
+    let d2 = d0 `mod` 10
+    logAction $ "tMeasure " <> label <> ": " <> show d1 <> "." <> show d2 <> "ms"
+    pure x
--- a/core/Pos/Util/Chrono.hs
+++ b/core/Pos/Util/Chrono.hs
@@ -36,13 +36,13 @@ newtype NewestFirst f a = NewestFirst {getNewestFirst :: f a}
             Functor, Foldable, Traversable,
             ToList, Container,
             Binary, Bi,
-            Arbitrary)
+            Arbitrary, NFData)
 newtype OldestFirst f a = OldestFirst {getOldestFirst :: f a}
   deriving (Eq, Ord, Show,
             Functor, Foldable, Traversable,
             ToList, Container,
             Binary, Bi,
-            Arbitrary)
+            Arbitrary, NFData)
 
 makePrisms  ''NewestFirst
 makeWrapped ''NewestFirst


util/Pos/Util/Util.hs
```diff
@@ -388,3 +390,26 @@ median l = NE.sort l NE.!! middle
...
+tMeasureLog :: (MonadIO m, WithLogger m) => Text -> m a -> m a
+tMeasureLog = tMeasure logDebug
+
+-- | 'tMeasure' with 'putText'. For places you don't have
+-- 'WithLogger' constraint.
+tMeasureIO :: (MonadIO m) => Text -> m a -> m a
+tMeasureIO = tMeasure putText
+
+-- | Measures time executed by the particular action, logs it before
+-- returning the result.
```
avieth: Should add: also forces the result before taking the second time sample, and before logging.

volhovm:
@avieth There's a very huge difference between "how does the thing work in benchmark" and "why does client's node lag", it is obvious. I test feature X but for some reason everything is slow, so `getBlocks` info tells me that it's a network part which lags -- i choose another relay and things go better. Same with the server nodes. We also have a set of metrics that are sent to the datadog, which serve similar purpose.

util/Pos/Util/Util.hs
volhovm: Fixed.

avieth:
> There's a very huge difference between "how does the thing work in benchmark" and "why does client's node lag", it is obvious.

Of course there's a difference. But there's also a connection: if it's slow in a benchmark, then it's at least that slow in production.

Do not pepper production software with crude pseudo-benchmarks like this. What do you hope to achieve? The user complains that it's slow; you inspect the logs; you say "oh, yeah, can confirm!". And then what? Then you do the actual work of trying to understand *why* it's slow, of which doing proper benchmarks is probably a good option.

util/Pos/Util/Util.hs
```diff
@@ -388,3 +390,26 @@ median l = NE.sort l NE.!! middle
...
+-- 'WithLogger' constraint.
+tMeasureIO :: (MonadIO m) => Text -> m a -> m a
+tMeasureIO = tMeasure putText
+
+-- | Takes the first time sample, executes action (forcing its
+-- result), takes the second time sample, logs it.
+tMeasure :: (MonadIO m) => (Text -> m ()) -> Text -> m a -> m a
+tMeasure logAction label action = do
+    before <- liftIO getCurrentTime
+    !x <- action
```
avieth: Do you want `deepseq` here?

avieth:
I can approve of including `tMeasure` in here, so that developers can use it locally. Just don't include all of those uses.

volhovm:
It's not a pseudo-benchmark. It's not even a benchmark, not related at all. These are different tools. Sometime I need to know how my software performs on particular run. Maybe client's code uses asserts or is compiled with `-O0`. Maybe he uses slow network machine or his HDD is dying.

>What do you hope to achieve?

_Obviously_ there are a lot of things I can achieve:
1. Monitor (using papertrail) that functionality doesn't get too slow (because of CPU/network/disk IO). Again, remember, we have datadog, it's the same.
2. In some places we can actually quickly reason about what is going on. Again -- I connect to cluster and i see that `diffusion.getBlocks` takes 30s to send 1.5mb. Should I panic? I go to the server and validate that `getHeaderHashes` is alright, still takes 0.5s (arbitrary number) as it was before, so it's networking.

avieth:
> _Obviously_

Bear with me, I must be a little slow.

> Sometime I need to know how my software performs on particular run

Fair enough. That's no reason to throw these annotation into production software. End users will not be using it.

> Maybe client's code uses asserts or is compiled with -O0. Maybe he uses slow network machine or his HDD is dying.

All of these things can be determined in more accurate ways. I fail to see how the measurements in this patch are good indicators at all of the optimization level, link, or HDD performance.

> Monitor (using papertrail) that functionality doesn't get too slow (because of CPU/network/disk IO). Again, remember, we have datadog, it's the same.

Is it the same? I don't see datadog annotations around these key functions. And if we already have it, isn't that enough? Can we leave it at that?

> I connect to cluster and i see that diffusion.getBlocks takes 30s to send 1.5mb. Should I panic? I go to the server and validate that getHeaderHashes is alright, still takes 0.5s (arbitrary number) as it was before, so it's networking.

But your conclusion is not valid. You can only see that `getBlocks` takes `n` seconds, _wholistically_. There's more going on in there.

avieth:
If you find this useful, good, keep on using it. It doesn't belong in `develop`, though (except the definition of `tMeasure` itself, that's fine).

volhovm:
>End users will not be using it.

End users also don't need debug logs, it doesn't mean they don't want to have them. 

>I fail to see how the measurements in this patch are good indicators at all of the optimization level, link, or HDD performance.

You can tell a difference between verification that takes 1.5 seconds (which is current level) and 5 seconds (which it can be), etc.

>But your conclusion is not valid. 

My conclusion is "I've decided who will be an assignee for the particular issue", so it's more or less valid :)

>I don't see datadog annotations around these key functions. And if we already have it, isn't that enough?

We have other application metrics. cc @cleverca22 Maybe you can comment on this? I think that having "average tx processing time" metric is pretty neat. I could implement these measurements as ekg metrics, i guess.

avieth:
> You can tell a difference between verification that takes 1.5 seconds (which is current level) and 5 seconds (which it can be), etc.

And what can you conclude?

Aside: if you need these tests in order to determine whether you built the thing with asserts or optimizations, you have other problems.

> My conclusion is "I've decided who will be an assignee for the particular issue", so it's more or less valid :)

What do you mean?

volhovm:
>And what can you conclude?

Probably, wrong build, or a slow machine. We can ask user to check that.

>What do you mean?

It was more of a joke, but anyway: sometimes QA results in issues created, and usually assignee is chosen by "how related the problem is to the code part that has a problem". So I can't say "no" if there are problems in delegation, because asking anybody else would be suboptimal. Same -- regarding recovery. If it's exclusively slow for somebody, it's more optimal for me to spend 3 minutes and say "user seems to have bad network connection, can he check that?" then read logs for 20 minutes and understand the same thing from some other data (since logs provide timestamps, we can usually explain the issue).

avieth:
> It barely relates to benchmarks because benchmarks are always to be run in some artificial emulated environment, whereas in real environment a lot of deviations may happen (e.g. some component abnormally takes 60% of whole getBlock execution, though in benchmarks it's just 5%) and to analyze behavior of particular run (either done for pure experiment or taken from cluster logs) this kind of logging will be of great use.

Those deviations make your measurements useless. This is why we run benchmarks in an "artificial emulated environment".

Maybe we can come to a compromise on this? The logging facility already includes a timestamp on every message. How about instead of putting a `tMeasure` around all of these things, just put start- and end- debug log messages around them?

volhovm:
>just put start- and end- debug log messages around them?

This is two times more logs _and_ it also invalidates half of the profit we try to achieve here. I'd like to do `cat logs | grep "tMeasure x"` and see how some particular call performed. In the same manner we noticed problems with non-equal slots and worker delays several weeks ago.

>Those deviations make your measurements useless. 

Doesn't it make benchmarks somewhat secondary? If you've measured some logic and it behaves in a completely different way on prod, then (I guess) we've missed something. This PR is a very simple way to be sure we didn't :man_shrugging: Also it's a relatively easy way to control critical points and performance if something goes wrong. We're not losing anything at all, these are just 2 lines of logs every 30+ lines or so.

Somewhat offtop: we have logs of this type in master:
```
[node.worker.block:DEBUG:ThreadId 89] [2018-02-26 11:14:06.03 UTC] node3: inFlight = fromList []                                                                                                       
[node.worker.block:DEBUG:ThreadId 89] [2018-02-26 11:14:06.03 UTC] node3: inFlight = fromList []                                                                                                   
[node.worker.block:DEBUG:ThreadId 89] [2018-02-26 11:14:06.03 UTC] node3: inFlight = fromList []        
```
Not sure what it is, but the point is -- logs are not perfect. I don't think it's a good time now to figure out what debug logs are actually super useful and which are not.

avieth:
> Doesn't it make benchmarks somewhat secondary? If you've measured some logic and it behaves in a completely different way on prod, then (I guess) we've missed something. 

Not at all. With proper benchmarks, we have measurements that we can be confident in. A controlled environment is good. If it behaves differently in the real world, and it probably does, that's because there's a whole bunch of other things going on. It doesn't mean we missed anything in the analysis of the particular component.

> This PR is a very simple way to be sure we didn't man_shrugging Also it's a relatively easy way to control critical points and performance if something goes wrong.

I don't believe this patch will make us sure of anything. It will show us symptoms: this particular thing that you happened to choose to `tMeasure` performs more slowly once-in-a-while. But that's for from revealing any useful information. It doesn't give you any indication as to _why_ it performs slowly, once-in-a-while. And how exactly does it offer any _control_?

I'm not saying it's isn't useful to take measurements of these things, just don't do it by default in every end-user build.

> Somewhat offtop: we have logs of this type in master:
> ...
> Not sure what it is, but the point is -- logs are not perfect. I don't think it's a good time now to figure out what debug logs are actually super useful and which are not.

Oh there are plenty of useless log statements already in master/develop.

volhovm:
Ok, I guess this discussion is already taking more time than it would take me to manually analyze logs :)

Should I remove all `tempMeasure` usages or it's alright to leave at least some of them? (maybe client part)

volhovm:
cc @avieth  ^

avieth:
> Should I remove all tMeasure usages or it's alright to leave at least some of them? (maybe client part)

What's the client part?

volhovm:
@avieth block retrieval/recovery. Everything else is measurements of (more like a) server part handlers. We can thus at least tell user "your daedalus is slow because your computer/internet is, not our fault".

avieth:
> We can thus at least tell user "your daedalus is slow because your computer/internet is, not our fault".

I don't think we'd be able to say so.

Supposing all of the measurements are abnormally high (above some known average), we might want to conclude that the user's computer is too slow. But it could be that they're simply running under high load. And besides, telling the user their computer is too slow should be based on actual knowledge about their computer, not some observed running time numbers. That's to say, a user should not expect to be able to run daedalus on a pentium 2 with 128mb ram. But that's _not_ to say that a user should not expect to be able to run it on a mid-range Core 2 with 1gb ram, even if the numbers on such a machine are abnormally high. See the difference? We can't conclude from high numbers that the machine is too slow, it could be (and probably is the case) that we've just done a poor job optimizing the software.

As for the network: it could always be our fault so long as we're centralized. They're connecting to our relays.

volhovm:
@avieth I've removed everything except for the `tMeasure` function itself.

PR #2576, opened 2018-02-26
crypto: Break Arbitrary instances out into a separate package 

erikd:
This means that the compiled binaries don't need to include Arbitrary instances, the QuickCheck libraries and associcated code.

Other packages that need these instances can depend on the cardano-sl-crypto-test package.


erikd:
I don't understand the AppVeyor build failure. It looks like ghc failed when it tried to run `gcc.exe`.

erikd:
Tomorrow I should get access to push a branch to this repo. Closing this for now.

PR #2579, opened 2018-02-26
Update build-cardano-sl-and-daedalus-from-source-code.md

calbrecht:
On recent NixOS master within nix-shell, i need to add the --system-ghc flag to get past the error when running scripts/build/cardano-sl.sh

No compiler found, expected minor version match with ghc-8.0.2 (x86_64) (based on resolver setting in /home/alab/ws/cardano-sl/stack.yaml).
To install the correct GHC into /home/alab/.stack/programs/x86_64-linux-nix/, try running "stack setup" or use the "--install-ghc" flag. To use your system GHC installation, run "stack config set system-ghc --global true", or use the "--system-ghc" flag.

### DIFF
--- a/docs/how-to/build-cardano-sl-and-daedalus-from-source-code.md
+++ b/docs/how-to/build-cardano-sl-and-daedalus-from-source-code.md
@@ -97,7 +97,7 @@ Enter `nix-shell`:
 
 After that, in order to build Cardano SL with wallet capabilities, run the following script:
 
-    [nix-shell:~/cardano-sl]$ ./scripts/build/cardano-sl.sh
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
My understanding is that we specifically don't want to "use your system GHC installation".

akegalj:
I also don't think we want to use your systems GHC installation. 

@calbrecht Note that if you are using nixOS there are better ways to install cardano-sl. Have a look at `docs/how-to/build-cardano-sl-and-daedalus-from-source-code.md`.

Closing this PR. @jmitchell as I am not a nixOS user, please reopen if you think closing this doesn't make sense.

PR #2582, opened 2018-02-26
[CSL-2089] shorten InfraConfiguration

avieth:
This values were always the same in practice and I really don't see why
they should be configurable.

The Kademlia stuff could be configurable, as I don't think mismatches between peers would cause any trouble, but it doesn't need to be delivered by `Data.Reflection`.


infra/Pos/DHT/Configuration.hs
```diff
@@ -7,23 +7,18 @@ module Pos.DHT.Configuration
 
 import           Universum
 
-import           Pos.Infra.Configuration (HasInfraConfiguration, ccEnhancedMessageBroadcast,
-                                          ccEnhancedMessageTimeout, ccKademliaDumpInterval,
-                                          ccNeighboursSendThreshold, infraConfiguration)
-
 -- | See 'Pos.CompileConfig.ccNeighboursSendThreshold'.
```
gromakovsky: This comment should be updated as well.

avieth:
> You forgot to remove these values from lib/configuration.yaml.

Will do

> I think it makes sense to rename infra/Pos/{Communication,DHT}/Configuration.hs to infra/Pos/{Communication,DHT}/Constants.hs. I think we had a convention that Configuration means values which are passed using reflection mechanism (and can be changed without recompilation), but hardcoded values are called Constants.

Yeah I agree, I'll make the change.

avieth:
@gromakovsky Changes made.

PR #2583, opened 2018-02-26
[CSE-261] Fix typo

sectore:
YT issue: https://iohk.myjetbrains.com/youtrack/issue/CSE-261


adinapoli-iohk:
@sectore Is this one for `master` or for `develop`? By the look of it looks like the latter, but I don't have the full context so that's why I'm asking 😉

sectore:
@adinapoli-iohk It's a hot fix, so we do need it in `master`.

adinapoli-iohk:
> @adinapoli-iohk It's a hot fix, so we do need it in master.

Right, but at the same time we also need to backport this to `develop`, if we stick to git-flow.

Anyway, I don't have permissions to merge this, you will have to ping on of the other guys ;)

Issue #2575, opened 2018-02-24
cache.nixos.org issue when building nix-shell


nsticco:
I'm getting the below error when trying to run nix-shell while building Cardano-SL from source. It seems like cache.nixos.org may be down. If this is the issue, is there a way around this?

I am getting this error with Nix 2.0 on both Ubuntu 16.04 and Ubuntu 17.10. I have built Cardano-SL in the past with the same commands without getting it.

```
nsticco@cardano1:~/cardano-sl$ nix-shell
these paths will be fetched (185.99 MiB download, 1787.85 MiB unpacked):
  /nix/store/02p4s2mb64psgr47vz6461nr641ps20f-perl-LocaleGettext-1.05
  /nix/store/069g827lh3hrhp4vkcq3rsh5jh65pm3l-ncurses-6.0-20171125
  /nix/store/0i27lrh6gxj6aiqfz001w6l5cgvmnivy-openssl-1.0.2l-man
  /nix/store/15axh70yqanvivgi69ns8sx7cjydqdx3-pkg-config-0.29.2
  /nix/store/1fvizcbh34fsrj7vivan3jfmhqla4aiy-hook
  /nix/store/1snk2wkpv97an87pk1842fgskl1vqhkr-python-2.7.14
  /nix/store/1yfivvr84q4i7lppp7r61b217g49ixj1-openssl-1.0.2l-debug
  /nix/store/2d2in2wqdgxr0p210lv17dvxibpnfw49-gmp-6.1.2-dev
  /nix/store/2kcrj1ksd2a14bm5sky182fv2xwfhfap-glibc-2.26-131
  /nix/store/2kj173rmvhmif82qg55v7kgy18m379y8-old-locale-1.0.0.7
  /nix/store/2xqlcc9yw8nmgiglc9w9dmcigy2mif6l-openssh-7.5p1
  /nix/store/3hprjz6f0lq1xs9wixijmq5hdfl6xh2c-autoconf-2.69
  /nix/store/3pb86cjaxmqsvar0ka3sx5qyfj2xjqxa-stack-1.5.1
  /nix/store/44dbwrwfy6lj45jy02d0rjbj9930nwf3-patch-2.7.5
  /nix/store/4c7k6js4b8slis635ydcs29pczrwhnr1-linux-pam-1.2.1
  /nix/store/4vkcmravylrzablnpz26fgvknf43bxqa-zlib-1.2.11-static
[...]

PR #2587, opened 2018-02-26
Update binary-cache-public-keys to support Nix v2.0

nsticco:
Nix v2.0 fails to build nix-shell if cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= is not listed as a binary-cache-public-key. We should add this to the compilation documentation.


rvl:
Thanks, I was just going to submit the change myself. This is coming from #2575.

I think the relevant [Nix 2.0 change](https://nixos.org/nix/manual/#ssec-relnotes-2.0) is this:
> Substitutions from binary caches now require signatures by default. This was already the case on NixOS.

jcmincke:

PR #2588, opened 2018-02-27
crypto: Break Arbitrary instances out into a separate package

erikd:
This means that the compiled binaries don't need to include `Arbitrary` instances, the QuickCheck libraries and associcated code. Other packages that need these instances can depend on the `cardano-sl-crypto-test` package.


erikd:
The AppVeyor failures seem unrelated to changes in this PR.
```
    Preprocessing library cardano-sl-wallet-new-0.1.0.0...
    Preprocessing executable 'cardano-node-new' for
    cardano-sl-wallet-new-0.1.0.0...
    Linking .w\dist\ca59d0ab\build\cardano-node-new\cardano-node-new.exe ...
    realgcc.exe: error: CreateProcess: No such file or dir
```

Pastafarianist:
@erikd that error is tracked in [DEVOPS-691](https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-691). Perhaps we need to raise its priority since it affects multiple PRs.

PR #2589, opened 2018-02-27
[CSL-2287] Generate real UTxo and txs history using dbgen

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2287

This is missing documentation. 

And we need a way to make it work with payments due to issues with stopping the nodes and starting them up again, which is required if you want to use `dbgen`.

[asciicast](https://asciinema.org/a/awaBROv3LaFtn1jKVDhC6lsV5)


jcmincke:
@ksaric can i close this PR?

jcmincke:
Merge to master: never do that

PR #2590, opened 2018-02-27
[CSL-2328] no reflection configurations for diffusion layer

avieth:
## The story so far

In the beginning, template Haskell was used to bake in a bunch of values from a configuration file at compile time, including the protocol magic number and the parameter `k`, but also a bundle of other things that we found we wanted to change without having to rebuild. So as a relatively quick solution, we moved to `Data.Reflection`, and read the configuration at run-time. Since many of these values are used in typeclass instances, this was a good stop-gap measure: we could switch values without a re-build, and we could keep the illusion that these values were constants and therefore usable within typeclass instances (see `Pos.Core.Slotting.Util`, where protocol constants are produced from thin air to give `Enum` and `Bounded` instances).

This led to the proliferation of `Has*Configuration` constraints. Often times, the constraints were (and probably still are) overstated: one would write `HasConfigurations` to take all of them, rather than taking the time to figure out which configurations it actually needs. As a result, one would have to come up with a complete configuration in order to do anything at all, really. In particular, if I want to make a couple diffusion layers and have them communication in some little simulation/experiment, I need a complete configuration.

## Changes made

This patch gets rid of the reflection-backed configuration motif for the diffusion layer and for the `Arbitrary Block` instance. Unfortunately, the full diffusion layer still requires `ProtocolConstants`, the `k` parameter in particular, because it's used in a message limit (block security parameter).

The `MessageLimited` class is removed, in favour of term-level limits. Hopefully we can soon get rid of these message limits altogether; the `Bi` instances should implicitly limit the lengths (i.e. fail if too many bytes are given), and this is in fact the case for most types. The issue only arises when we accept a list of something, which we just shouldn't do. That's probably why the `recoveryHeadersMessage` value came to be. I also noticed some message limits were broken to begin with: some messages types have a `NonEmpty` stores as a value in a `Map`. The `Map` itself contributes to the limit, but those `NonEmpty`s did not.

~~The hspec automatic `Spec` generator can't be used anymore, since our `Spec`s now have `HasConfigurations` in their contexts. That's because instead of taking a `ProtocolMagic` from a `Data.Reflection` `given`, any instance which needs such a thing will instead use an `Arbitrary ProtocolMagic`. Now the `Spec`s need to have that instance, so we make one that gives `pure protocolMagic` using a `HasProtocolConstants`. So I had to write out the big `Spec` manually. We should make the `Spec`s not need any configurations, and then we can go back to using the auto generator. I imagine, but I'm not sure, that if we used a generic `Arbitrary` instance for `ProtocolMagic`, tests would fail, because they probably assume a global constants for this.~~

## Diffusion layer interface

In order to get testing/benchmarking setups of multiple diffusion layers in a single process, over an arbitrary network-transport, the full diffusion layer interface was opened up: you don't have to give a `NetworkConfig`, and you can get access to the underlying `Node`.

The old full diffusion layer interface remains, which uses the rougher new interface, filling in many values using the `NetworkConfig` and its `Topology`.


lib/src/Pos/Diffusion/Full/Block.hs
```diff
@@ -261,32 +267,34 @@ getBlocks logic enqueue nodeId tipHeader checkpoints = do
...
     -> (BlockHeader -> NodeId -> d t)
     -> d (Map NodeId (d t))
-requestTip enqueue k = enqueue (MsgRequestBlockHeaders Nothing) $ \nodeId _ -> pure . Conversation $
+requestTip logic enqueue k = enqueue (MsgRequestBlockHeaders Nothing) $ \nodeId _ -> pure . Conversation $
     \(conv :: ConversationActions MsgGetHeaders MsgHeaders m) -> do
         logDebug "Requesting tip..."
+        bvd <- getAdoptedBVData logic
         send conv (MsgGetHeaders [] Nothing)
-        received <- recvLimited conv
+        received <- recvLimited conv (mlMsgHeaders bvd 2200)
```
karknu: 2200?
Did you miss to use recoveryHeaderMessage?

lib/src/Pos/Diffusion/Full/Block.hs
```diff
@@ -501,7 +517,8 @@ handleBlockHeaders logic oq keepaliveTimer =
     -- protocol compatibility reasons only. We could use 'Void' here because
     -- we don't really send any messages.
     logDebug "handleBlockHeaders: got some unsolicited block header(s)"
-    mHeaders <- recvLimited conv
+    bvd <- getAdoptedBVData logic
+    mHeaders <- recvLimited conv (mlMsgHeaders bvd 2200)
```
karknu: see other 2200 comment

lib/src/Pos/Diffusion/Full/Block.hs
avieth: Oh yes, because we don't have the configuration value here. Which one did it come from? Infra, node, block? I think `HasBlockConfiguration`...  I suppose for the time being, this value is a configuration parameter for the full diffusion layer, so I'll put it back in.

avieth: Oh, looks like I *do* already take it as a parameter. Just forgot to wire it up....

avieth:
Review please @dcoutts @coot @karknu

coot:
There is an unused import in `Pos.Arbitrary.Crypto` (`serialize'`):
```
/home/marcin/repos/iohk/cardano-sl/crypto/Pos/Arbitrary/Crypto.hs:21:1: warning: [-Wunused-imports]
    The import of ‘serialize'’
    from module ‘Pos.Binary.Class’ is redundant
[24 of 24] Compiling Pos.Aeson.Crypto ( Pos/Aeson/Crypto.hs, .stack-work/dist/x86_64-linux-tinfo6-nopie/Cabal-1.24.2.0/build/Pos/Aeson/Crypto.o ) [Pos.Crypto changed]
Installing library in
/home/marcin/repos/iohk/cardano-sl/.stack-work/install/x86_64-linux-tinfo6-nopie/lts-9.1/8.0.2/lib/x86_64-linux-ghc-8.0.2/cardano-sl-crypto-1.1.0-JVQA1cPkT55APSpVKAU5lk
Registering cardano-sl-crypto-1.1.0...
```

avieth:
@coot @karknu I'm begging for a review on this one. I know it's big but it's important and just about all of my current work uses it.

avieth:
Will split this up

PR #2592, opened 2018-02-27
[CSL-2318] Added --no-ntp-checks

coot:


networking/src/NTP/Client.hs
```diff
@@ -45,26 +46,29 @@ import           NTP.Packet (NtpPacket (..), evalClockOffset, mkCliNtpPacket, nt
...
+        , ntpResponseTimeout :: Microsecond
+        -- ^ delay between making requests and response collection
+        , ntpPollDelay       :: Microsecond
+        -- ^ how often to send responses to server
+        , ntpMeanSelection   :: [(Microsecond, Microsecond)] -> (Microsecond, Microsecond)
+        -- ^ way to sumarize results received from different servers.
+        -- this may accept list of lesser size than @length ntpServers@ in case some servers
+        -- failed to respond in time, but never an empty list
+        }
+    | NtpSyncUnavailable
```
avieth: Would prefer to just use a `Maybe NtpClientSettings`. `NtpSyncUnavailable` should be given to the clients rather than `NtpSyncOk` in case there's no syncing.

coot: ok

coot: Do you want to disable `getNtpStatusOnce`? It is used by the wallet.  For that we would need to make `NodeParams` be somehow available in the context of `getNtpStatusOnce`.

avieth: I spoke with @dcoutts  about this briefly and we agreed it would be a good idea to have a daemon thread periodically polling NTP (if enabled) and putting the latest sync information into some mutable cell. To serve the UI endpoint, we can just sample it, and give `NtpUnavailable` if it's disabled or otherwise failed.

avieth: I'm unsure of how the NTP thread works now but I do know it does a synchronous check at startup. I was wondering whether a failure here should crash the program but now I can say for sure that it shouldn't; it should be backgrounded and keep on trying to recover in case of IO errors.

coot: There NTP feature looks like this:
* There is low level `spawnNtpClient` which runs in a new thread and receives ntp data. It accepts a callback through `NtpClientSettings` that runs when new data arrives.

Then there are two higher level functions:
* `withNtpCheck` which runs the `spawnNtpClient` and using a configuration (which has the callback), right now this is only used for logging.
* `getNtpStatusOnce`: it runs the `spawnNtpClient` and waits for the first data that arrives, runs the callback and exists.  It is synchronious. It is used by the wallet, one of the apis just runs it and returns the data to the client (`Cardano.Wallet.API.V1.LegacyHandlers.Info` in `new-wallet`).

I think we should:

* use `withNtpCheck` in the diffusion layer, update a mutable cell which we can expose in the diffusion layer, this way the wallet can poll and receive the most recent data.it is not 
* remove `getNtpStatusOnce`.
* Currently `getNtpStatusOnce` creates `NtpClientSettings` from a thin air (except for the callback), so for example one cannot configure which ntp servers are used and also wheather to use ntp at all. This might be out of the scope for this feature, but this might be useful in Shelley. It's probably better to use an ntp server that's geographically close rather than down under ;).

coot: We could extend `NetworkConfig`, first by just adding boolean flag if NTP should be used, and later to add there more ntp stuff from `NtpClientSettings` (like list of servers, logger name, response timeout, poll delay).

avieth: > Currently getNtpStatusOnce creates NtpClientSettings from a thin air (except for the callback), so for example one cannot configure which ntp servers are used and also wheather to use ntp at all. Making ntp fully configurable might be out of the scope for this feature, but this might be useful in Shelley. It's probably better to use an ntp server that's geographically close rather than down under ;).

Ah it probably uses the `Data.Reflection` magical configuration. Yes this is bad. I sliced out a big chunk of that here https://github.com/input-output-hk/cardano-sl/pull/2590 .

> We could extend NetworkConfig, first by just adding boolean flag if NTP should be used, and later to add there more ntp stuff from NtpClientSettings (like list of servers, logger name, response timeout, poll delay).

Good idea.

> use withNtpCheck in the diffusion layer, update a mutable cell which we can expose in the diffusion layer, this way the wallet can poll and receive the most recent data.

Let's not put it on the diffusion layer type itself. I think it's better expressed as a standalone component.

networking/src/NTP/Client.hs
```diff
@@ -45,26 +46,29 @@ import           NTP.Packet (NtpPacket (..), evalClockOffset, mkCliNtpPacket, nt
...
-    , ntpResponseTimeout :: Microsecond
-      -- ^ delay between making requests and response collection
-    , ntpPollDelay       :: Microsecond
-      -- ^ how often to send responses to server
-    , ntpMeanSelection   :: [(Microsecond, Microsecond)] -> (Microsecond, Microsecond)
-      -- ^ way to sumarize results received from different servers.
-      -- this may accept list of lesser size than @length ntpServers@ in case some servers
-      -- failed to respond in time, but never an empty list
-    }
+data NtpClientSettings m =
```
avieth: While we're here, let's make `NtpClientSettings` monomorphic in `IO`. I don't anticipate any difficulties here, and we can get rid of the hoisting functions, the `NtpMonad` constraint, use of `MonadMask`, etc.

PR #2594, opened 2018-02-28
[CSL-2329] Fix reporting of NTP time drifts

adinapoli-iohk:
This commit allows /api/settings/time/difference and /api/v1/node-info
to correctly report _positive_ time drifts with the global NTP servers by
modifying the logic which was computing such information.

It also fixes an incorrect description in the old V0 api, which was
claiming such time difference was returned in milliseconds, rather than
microseconds.

In case of the local time lagging _behind_ the NTP global time, a positive integer is returned:

screen shot 2018-02-28 at 14 48 27

Whereas negative values indicate the local time ahead (i.e. in the future) with regards the NTP time:

screen shot 2018-02-28 at 14 51 04


wallet/src/Pos/Wallet/Web/Api.hs
```diff
@@ -483,8 +483,8 @@ data WSettingsApiRecord route = WSettingsApiRecord
   , _localTimeDifference :: route
     :- "time"
     :> "difference"
-    :> Summary "Get local time difference in milliseconds."
-    :> WRes Get Word
+    :> Summary "Get local time difference in microseconds."
+    :> WRes Get Integer
```
adinapoli-iohk: I've also accidentally spotted that the documentation for the v0 api was incorrectly claiming this were milliseconds. Also, the returned time is now signed, thus the type was switched from `Word` to `Integer`.

wallet/src/Pos/Wallet/Web/Methods/Misc.hs
```diff
@@ -155,16 +156,16 @@ syncProgress = do
...
     diff <$> getNtpStatusOnce
   where
-    diff :: NtpStatus -> Word
+    diff :: NtpStatus -> Integer
     diff = \case
         NtpSyncOk -> 0
         -- `NtpSyncOk` considered already a `timeDifferenceWarnThreshold`
         -- so that we can return 0 here to show there is no difference in time
-        NtpDesync diff' -> fromIntegral diff'
+        NtpDesync diff' -> toMicroseconds diff'
```
adinapoli-iohk: Probably just a `fromIntegral` like before would have suffice, but this one is slightly more self-explanatory.

infra/Pos/NtpCheck.hs
```diff
@@ -49,9 +49,12 @@ ntpCheckHandler cont (newMargin, transmitTime) = do
...
     localTime <- Timestamp <$> currentTime
     let timeDiff = diffTimestamp ntpTime localTime
+    -- If the @absolute@ time difference between the NTP time and the local time is
+    -- bigger than the given threshold, it effectively means we are not synced, as we are
+    -- either behind or ahead of the NTP time.
     let ntpStatus
-            | timeDiff <= timeDifferenceWarnThreshold = NtpSyncOk
-            | otherwise = NtpDesync timeDiff
+            | abs timeDiff > timeDifferenceWarnThreshold = NtpDesync timeDiff
+            | otherwise = NtpSyncOk
```
adinapoli-iohk: I have flipped the logic: if we have an absolute value greater than the threshold level it means we are not synced as we are either in the past or in the future.

adinapoli-iohk:
@darko-mijic @nikolaglumac I have took the liberty of merging this PR, which means you can now proceed in handling the case where the value is positive 😉

darko-mijic:
Issue created. Thanks @adinapoli-iohk

PR #2598, opened 2018-02-28
[CSL-2331] improve synchronisation / communication between outbound queue threads and client threads

avieth:
Previously: when enqueueing the caller gets a `Map NodeId (m (Either SomeException t))` as soon as the outbound queue has completed _enqueueing_. The `m`s in that map will block until the relevant conversation (to the given `NodeId`) has been dequeued and completed.

1. It's impossible to abort an enqueued conversation before it is dequeued.
2. It's impossible to get a hold of the conversation's thread id and, for instance, throw exceptions to it, or link it to another thread in the `Async` style.

This implements the improved interface:

```
enqueue
    :: OutboundQ msg nid buck
    -> MsgType nid
    -> msg a
    -> IO [(nid, TVar (PacketStatus a))]

data PacketStatus a = InFlight (Async a) | Pending | Aborted

-- Abort a pending enqueued thing. No-op if it's already in flight.
-- [Not implemented in this ticket but easy to do]
abortPacket :: TVar (PacketStatus a) -> IO ()

-- Block until the packet is in flight. Give Nothing if it's aborted (outbound queue itself never aborts them).
-- [Not implemented in this ticket but easy to do]
threadForPacket :: TVar (PacketStatus a) -> IO (Maybe (Async a))
```

This will help @karknu with the implementation of CSL-2148

Part of this change is to make the outbound queue use `IO` explicitly, rather than `Mockable`/`MonadIO`/`MonadThrow`/`WithLogger` etc. Turns out it still fits nicely into cardano-sl. I think we ought to make the full diffusion layer use `IO` as well, and ditch our `Network.Transport.Abstract` shim. Fact is: whenever we need a `forall y . hugeMonadTransformer y -> IO y` we have one.

### DIFF
--- a/networking/src/Network/Broadcast/OutboundQueue.hs
+++ b/networking/src/Network/Broadcast/OutboundQueue.hs
@@ -26,6 +26,7 @@
 {-# LANGUAGE ScopedTypeVariables       #-}
 {-# LANGUAGE StandaloneDeriving        #-}
 {-# LANGUAGE TupleSections             #-}
+{-# LANGUAGE RecursiveDo               #-}
 {-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
 
 module Network.Broadcast.OutboundQueue (
@@ -53,6 +54,7 @@ module Network.Broadcast.OutboundQueue (
     -- * Enqueueing
   , Origin(..)
   , EnqueueTo (..)
+  , PacketStatus (..)
   , enqueue
   , enqueueSync'
   , enqueueSync
@@ -80,21 +82,23 @@ module Network.Broadcast.OutboundQueue (
   ) where
 
 import           Control.Concurrent
-import           Control.Exception.Safe (MonadMask, SomeException, displayException, finally, mask_,
-                                         try)
+import           Control.Concurrent.Async
+import           Control.Concurrent.STM
+import           Control.Exception (SomeException, catch, throwIO, displayException,
+                                    finally, mask_)
 import           Control.Lens
 import           Control.Monad
-import           Control.Monad.IO.Class
 import           Data.Either (rights)
 import           Data.Foldable (fold)
 import           Data.List (intercalate, sortBy)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (fromMaybe, maybeToList)
+import           Data.Maybe (fromMaybe, maybeToList, mapMaybe)
 import           Data.Monoid ((<>))
 import           Data.Ord (comparing)
 import           Data.Set (Set)
 import qualified Data.Set as Set
+import           Data.String (fromString)
 import           Data.Text (Text)
 import qualified Data.Text as T
 import           Data.Time
@@ -103,11 +107,10 @@ import           Formatting (Format, sformat, shown, string, (%))
 import qualified System.Metrics as Monitoring
 import           System.Metrics.Counter (Counter)
 import qualified System.Metrics.Counter as Counter
-import           System.Wlog.CanLog (WithLogger, logDebug)
+import           System.Wlog (usingLoggerName, logDebug)
 import qualified System.Wlog.CanLog as Log
 import           System.Wlog.Severity (Severity (..))
 
-import qualified Mockable as M
 import           Network.Broadcast.OutboundQueue.ConcurrentMultiQueue (MultiQueue)
 import qualified Network.Broadcast.OutboundQueue.ConcurrentMultiQueue as MQ
 import           Network.Broadcast.OutboundQueue.Types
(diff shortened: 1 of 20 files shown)


pkgs/default.nix
```diff
@@ -6922,32 +6922,32 @@ inherit (pkgs) mesa;};
...
+         , filepath, fmt, formatting, generic-arbitrary, half, hashable
+         , hspec, kademlia, lens, log-warper, lrucache, memory, mmorph
+         , monad-control, MonadRandom, mtl, neat-interpolation
+         , network-transport, network-transport-tcp, optparse-applicative
+         , parsec, plutus-prototype, pvss, QuickCheck, random, reflection
+         , resourcet, rocksdb-haskell-ng, safe-exceptions, safecopy
+         , serokell-util, servant, servant-server, servant-swagger, stdenv
+         , stm, systemd, tagged, template-haskell, text, text-format, time
+         , time-units, transformers, transformers-base, universum, unix
+         , unliftio, unordered-containers, vector, wai, warp, warp-tls, yaml
```
karknu: How about one packet per line so that the diff will be cleaner the next time we need to add an entry?

erikd: Isn't `pkgs/default.nix` a generated file (generated by `pkgs/generate.sh`)?

karknu: If it is generated then it shouldn't be in git.

avieth: Indeed it is generated, by `pkgs/generate.sh`. It has to be in git in order for CI to pass. Maybe there's a better way, I don't know, but so be it for now.

networking/src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -959,50 +977,81 @@ intDequeue outQ@OutQ{..} threadRegistry@TR{} sendMsg = do
...
+      -- instead we lazily fill it will the new status, using the old status
+      -- to determine what the new status shall be. In case the old status
+      -- is PacketEnqueued, computing the new status requires forking a
+      -- thread.
+      -- Should be fine I think.
+      oldStatus <- atomically $ do
+        status <- readTVar (packetStatus p)
+        writeTVar (packetStatus p) newStatus
+        return status
+      newStatus <- case oldStatus of
```
avieth: This is a bit dodgy. I wasn't even sure it would work. The STM transaction is lazy in `newStatus`, so that's OK. But what happens if another thread reads that `packetStatus p` `TVar` and tries to force the value, in-between this transaction and this thread's determination of `newStatus`? I assume it'll just block until this thread has finished, but honestly I'm not sure. Seems to work though!

coot: That's interesting edge case. It seems it should be guaranteed by the recursive do rather than atomicity of STM.

avieth:
Need a review please.

@dcoutts @coot @karknu

avieth:
Pushed a new commit to make the logging aspect of outbound queue a bit more sane.

I wanted it to be in `IO`, so the log-warper stuff was problematic: we need a `HasLoggerName m` instance, which `IO` does not have. So now we can't use `IO`, simply because of our choice of logging system.

Solution: outbound queue uses a `Trace ~ Op (IO ()) (Severity, Text)` value to do logging. One can be made from a `LoggerName` and `String` queue name. In practice for the full diffusion layer we use `"diffusion.outboundqueue"` and the queue name comes from the network configuration.

networking/src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -326,19 +348,41 @@ inFlightWithPrec nid prec = inFlightTo nid . at prec . anon 0 (== 0)
...
+setInFlightFor Packet{..} f var = modifyMVar_ var (return . update)
   where
     update :: InFlight nid -> InFlight nid
     update = Map.adjust updateInnerMap packetDestId
 
     updateInnerMap :: Map Precedence Int -> Map Precedence Int
     updateInnerMap = Map.adjust f packetPrec
 
+-- | Abstracts logging.
+newtype Trace m s = Trace
```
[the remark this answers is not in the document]
avieth: I was shy to put too much of an emphasis on this, because I worry that people are extremely fond of monad-transformer-based logging (that seems to be the way that everybody does it). Lately I've come to not like it so much. Why should I not be able to use `IO`, just because of logging?
Anyway, yes this does deserve a better place. Do bring it out to its own module and use it in your NTP work if you think it would be useful.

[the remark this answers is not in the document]
coot: +1 I like it, it's a simple abstraction that let both camps to live alongside and indeed logging shouldn't decide on the monad one wants to use.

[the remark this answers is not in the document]
coot: And if one provides profanctor instance instead of the covariant one, one could also lift the monad.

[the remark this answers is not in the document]
avieth: Profunctor wouldn't be appropriate, since the monadic term should always have a `()`. And a profunctor-like thing to change the monad would need a different kind:

```
class WeirdProfunctor (f :: * -> (* -> *) -> *) where
  weirdRmap :: (forall x . m x -> n x) -> f a m -> f a n
  normalLmap :: (b -> a) -> f a m -> f b m
```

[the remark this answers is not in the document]
coot: I know; this is what I had in mind. The `weiredRmap` is just application of a natural transformation. It's kind of a skewed profunctor, rather than just a functor: `C^op x C -> Set`, a functor of this sort: `A^op x B -> Set`

PR #2601, opened 2018-03-01
[CSL-2332] Silence cardano-node

gromakovsky:
That's another attempt to fix CSL-2332. Not tested yet.


gromakovsky:
Doesn't work. The workaround is https://github.com/input-output-hk/daedalus/pull/766.
It should be fixed in `develop` now. For more details see CSL-2332.

PR #2602, opened 2018-03-01
[CSL-2165] Add timestamp to output of `GET api/v1/transactions` endpoint

Martoon-00:
First, `Transaction` type now contains timestamp field. Time is represented
as fractional number of seconds, with precision up to microseconds.
Example: `1519926196.536234`
Full example of response: https://pastebin.com/jcRYcNDi

Second, this allowed to enable filtering and sorting by time via
general existing mechanisms.
E.g. one can pass `?timestamp=LTE[1519926196.536234]` in request
and see only transactions made after mentioned point in time. Both POSIX timestamp and UTC time are acceptable.

Also, since `Timestamp` type began to be used in `wallet-new`, I rewrote its `instance Arbitrary` in more plausible way.

Tests:
I made 4 transactions, and then tested queries.
* `https://127.0.0.1:8090/api/v1/transactions/<wallet>?sort_by=DSC[timestamp]` returned me transactions in newest-at-top order.
* `https://127.0.0.1:8090/api/v1/transactions/<wallet>?timestamp=GTE[<time of transaction X>]` returned all transactions newer than `X` including `X` itself.
`GT[time of X]` and `GTE[floored time of X]` also worked as expected.
* I can do
  * `https://127.0.0.1:8090/api/v1/transactions/<wallet_id>?created_at=LTE[1520447456.42]` or
  * `https://127.0.0.1:8090/api/v1/transactions/<wallet_id>?created_at=LTE[2018-03-07T18:30:56.456855]`
both will work.


core/Pos/Core/Slotting/Timestamp.hs
```diff
@@ -60,8 +66,14 @@ diffTimestamp t1 t2 = getTimestamp t1 - getTimestamp t2
...
 
+-- | Lens to convert timestamp to fractional seconds and vice versa.
+timestampSeconds
+    :: (Eq a, RealFrac a)
+    => Iso' Timestamp a
+timestampSeconds = _Timestamp . iso fromIntegral round . dividing 1e6
+
 timestampToPosix :: Timestamp -> POSIXTime
-timestampToPosix (Timestamp ts) = fromIntegral ts / 1000000
+timestampToPosix = view timestampSeconds
```
Martoon-00: Checked in `ghci`, works exactly in the same way

Martoon-00:
Thanks, all this makes sense.

I have chosen fractional timestamp, because daedalus displays transactions in descendant order. And if user is able to make two transactions during the same second then sorting on POSIX timestamp doesn't give us true descendant order. What do you think about it?
Generally I agree that displaying timestamp in seconds would better.

adinapoli-iohk:
> And if user is able to make two transactions during the same second then sorting on POSIX timestamp doesn't give us true descendant order.

I think that's a minor concern, as probably that second different would be easily spent by the user by clicking the "send" button, so I don't think it will be humanly possible for a user to issue a transaction at exactly the same second 😀 But if @nikolaglumac or @darko-mijic think this is a real concern, we can maybe switch to UTC dates, where we can have more-than-seconds precision (which is possibly even a better model than using POSIX timestamps 🤔 )

wallet-new/src/Cardano/Wallet/API/Indices.hs
```diff
@@ -33,6 +35,15 @@ instance ToIndex Wallet Core.Coin where
         Just c  -> Just (Core.mkCoin c)
     accessIx Wallet{..} = let (V1 balance) = walBalance in balance
 
+instance ToIndex Transaction Core.TxId where
+    toIndex _ = rightToMaybe . decodeHash
+    accessIx Transaction{..} = let V1 ti = txId in ti
```
parsonsmatt: Is there a reason we don't have an `unV1 :: V1 a -> a` ?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -541,10 +556,12 @@ data Transaction = Transaction
     -- ^ The input money distribution.
   , txOutputs       :: !(NonEmpty PaymentDistribution)
     -- ^ The output money distribution.
-  , txType          :: TransactionType
+  , txType          :: !TransactionType
     -- ^ The type for this transaction (e.g local, foreign, etc).
-  , txDirection     :: TransactionDirection
+  , txDirection     :: !TransactionDirection
     -- ^ The direction for this transaction (e.g incoming, outgoing).
+  , txCreationTime  :: !(V1 Core.Timestamp)
```
parsonsmatt: I like having microsecond precision here. However, a unix timestamp is an awfully non-human-friendly way to specify this -- using a UTCTime would be more "human-readable" while offering the same level of precision. On the other hand, a UTCTime is a lot less friendly to manipulate for computers, and has to deal with the vagaries of leap seconds etc.

wallet-new/src/Cardano/Wallet/API/Indices.hs
adinapoli-iohk: No reason 😛

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: FWIW UTCTime can represent sub-second resolutions.

A datatype like

```haskell
data ApiTime = UTC UTCTime | Timestamp Timestamp
```

with a non-tagged parse attempt would be fine, IMO.

wallet-new/src/Cardano/Wallet/API/Request/Sort.hs
```diff
@@ -74,6 +74,7 @@ flattenSortOperations (SortOp f fs) = show f : flattenSortOperations fs
 -- much sense, so we restrict ourselves.
 type family SortParams (syms :: [Symbol]) (r :: *) :: [*] where
     SortParams '["balance"] Wallet = '[Core.Coin]
+    SortParams '["timestamp"] Transaction = '[Core.Timestamp]
```
adinapoli-iohk: @Martoon-00 Shouldn't we change the sorting key to be `created_at` to be consistent with the filtering? 😉

parsonsmatt: tfw stringly kinded programming

adinapoli-iohk: > tfw stringly kinded programming

Lift your semantic errors at the kind level with this _one weird trick_ -- Haskell consultants hate it!

Martoon-00: Thanks, missed it :+1:

Martoon-00: I extracted those `balance`, `id` etc names to type aliases, hope it makes sense.

wallet-new/src/Cardano/Wallet/API/V1/Wallets.hs
```diff
@@ -15,8 +16,8 @@ type API =
                    :> PostCreated '[ValidJSON] (WalletResponse Wallet)
     :<|> "wallets" :> Summary "Returns all the available wallets."
                    :> WalletRequestParams
-                   :> FilterBy '["wallet_id", "balance"] Wallet
-                   :> SortBy   '["balance"] Wallet
+                   :> FilterBy '[P.WalletId, P.Balance] Wallet
+                   :> SortBy   '[P.Balance] Wallet
```
adinapoli-iohk: @Martoon-00 My 2 cents, and maybe @parsonsmatt can disprove: I don't like too much these type aliases, because:

- Doesn't buy you any extra "type safety" (after all, those are already types)
- They introduce another layer of indirection you need to unpack (as in "Hey, what's this type is for? let me check")
- They don't allow you to see _at glance_ the final filter format. An advantage of having a `Symbol` here is that you have a WYSIWYG kind of interaction: the types you see there maps directly to filter in the final API, as if you were using them in the browser. Now we lost this ability 😭

wallet-new/src/Cardano/Wallet/API/Request/Parameters.hs
```diff
@@ -0,0 +1,11 @@
+-- | Enlists names of request parameters.
+-- To import only in /qualified/ way.
+
+module Cardano.Wallet.API.Request.Parameters where
+
+type WalletId = "wallet_id"
+type TransactionId = "id"  -- it's quite unambigous for transactions
+type CreatedAt = "created_at"
+type Balance = "balance"
```
adinapoli-iohk: To clarify what I meant by ambiguity (to not come across as a grumpy man which changes his ideas every other day 😁 ):

As this is a REST API and we are manipulating resources, my rule of thumb would be:

- If we are filtering over the manipulated resource, we use simply `id`. What I mean by "manipulated resource"? If we are querying `/api/wallets` or `api/transactions` it's clear that the manipulated resource is a `Wallet` is the first case and a `Transaction` in the latter, therefore for both we should have just "id" (and yes, on line 6 we should change `wallet_id` to simply wallet, me thinks.

This way our filter will be nice and clean `/api/v1/wallets?id=BlaBla` and `api/v1/transactions?id=1ad2...`

- If there is ambiguity because we are filtering on a resource related to the one manipulated, than we need to qualify things. For example, let's assume the fictional case where we want to filter the wallet list by transaction ids, we would write `/api/v1/wallets?transaction_id=blabla`.

I was initially concerned (by self-reviewing my code) that having `id` being used in more than 1 place would have created conflicts but I think it should be fine, as we have the following type family:

```
type family FilterParams (syms :: [Symbol]) (r :: *) :: [*] where
    FilterParams '["wallet_id", "balance"] Wallet = IndicesOf Wallet
```

This should ensure, thanks to the fact it's parametrised over `r` that things wouldn't conflicts:

```
type family FilterParams (syms :: [Symbol]) (r :: *) :: [*] where
    FilterParams '["id", "balance"] Wallet = IndicesOf Wallet
    FilterParams '["id", "created_at"] Transaction = IndicesOf Transaction
```

@Martoon-00 , @parsonsmatt thoughts?

wallet-new/src/Cardano/Wallet/API/V1/Wallets.hs
Martoon-00: That's reasonable, but as far as I understand, arguments about extra layer of indirection and clearness at glance usually succumb when highly duplication arises, and that's exactly the same case. It doesn't really buy us type safety, but if I provide incorrect name I'll get some totally unclear error from GHC, which doesn't make development easier.
My impression is that in current machinery it's quite easy to make a mistake (inconsistent parameter names `FilterBy` and `SortBy`, inconsistent `toIndex` and `parseJSON`), plus there is no errors processing in it yet. Once we implement errors processing, and remove boilerplate (one parameter name mention for `SortBy`, one for `FilterBy`), then type aliases won't be needed much anymore.

Tl;dr I wish we have these names extracted to type aliases at least as temporal measure, till the moment we facilitate changes in all this stuff.

wallet-new/src/Cardano/Wallet/API/Request/Parameters.hs
Martoon-00: Looks very sane for me.

adinapoli-iohk:
> Tl;dr I wish we have these names extracted to type aliases at least as temporal measure, till the moment we facilitate changes in all this stuff.

I see what you mean, and that's an understandable point of view, @Martoon-00 . What about this sweet spot: you promise to create a tech-debt ticket to remember ourselves we really ought to scrap these type alias at some point, and I will be willing to accept the current status quo 😁

Martoon-00:
Thanks! :star2: 
I created CSL-2360, hope we will rewrite this all in pretty way very soon :slightly_smiling_face:

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -205,6 +209,30 @@ instance FromHttpApiData (V1 Core.Address) where
...
+instance ToJSON (V1 Core.Timestamp) where
+    toJSON timestamp =
+        let utcTime = timestamp ^. _V1 . Core.timestampToUTCTimeL
+        in  String $ showApiUtcTime utcTime
+
+-- | Parses from both UTC time in 'apiTimeFormat' format
+-- and POSIX (fractional) timestamp in seconds.
+instance FromJSON (V1 Core.Timestamp) where
+    parseJSON v = V1 <$> asum
+        [ utcTimeParser v
```
adinapoli-iohk: Silly question: if our `ToJSON` formats using `UTCTime`, what are we gaining in supporting the `timePosixParser` here?

wallet-new/src/Cardano/Wallet/Util.hs
```diff
@@ -23,3 +29,26 @@ stripFieldPrefix = dropWhile (not . isUpper)
...
+-- current LTS-9.1.
+apiTimeFormat :: String
+apiTimeFormat = Time.iso8601DateFormat (Just "%H:%M:%S%Q")
+
+-- | Parse UTC time from API.
+parseApiUtcTime :: MonadFail m => Text -> m Time.UTCTime
+parseApiUtcTime = Time.parseTimeM False defaultApiTimeLocale apiTimeFormat . toString
+
+-- | Encode UTC time for API.
+showApiUtcTime :: Time.UTCTime -> Text
```
adinapoli-iohk: That's probably trivial, but if we have time it would be nice to write a QuickCheck property to assess that `showApiUtcTime` and `parseApiUtcTime` roundtrip.

core/Pos/Core/Slotting/Timestamp.hs
```diff
@@ -60,8 +68,21 @@ diffTimestamp t1 t2 = getTimestamp t1 - getTimestamp t2
 addMicrosecondsToTimestamp :: Microsecond -> Timestamp -> Timestamp
 addMicrosecondsToTimestamp m t = Timestamp { getTimestamp = (getTimestamp t) + m }
 
+-- | Lens to convert timestamp to fractional seconds and vice versa.
+timestampSeconds
+    :: (Eq a, RealFrac a)
+    => Iso' Timestamp a
+timestampSeconds = _Timestamp . iso fromIntegral round . dividing 1e6
```
adinapoli-iohk: Am I the only one who finds the usage of the term `second` without the quantifying precision (as in `micro`) misleading? As we are dividing by `1e6`, should we call this `timestampMicroseconds`?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
Martoon-00: Um right, it's actually useless

core/Pos/Core/Slotting/Timestamp.hs
Martoon-00: As to me, it should be `second`, because
* looking at `timestampMicroseconds`, I don't see at glance which measure it returns time in. Maybe it returns time in `weeks` with up to `microseconds` precision :smile:
Perhaps more qualified name would be more telling.
* it actually returns time with precision which I'd expect when converting from `Timestamp`, i.e, exactly the same `up to mcs` as in `Timestamp`.

What do you think?

adinapoli-iohk: Uhm, not totally convinced about the first point, I think 😁 `microseconds` is not a subjective time of measure, imho. I sort of see what you are saying, but for the _principle of least surprise_ `timestampMicroseconds` does exactly what it says in the title: it returns a timestamp expressed in microseconds. I'm curious to hear what @parsonsmatt thinks here 😉

wallet-new/src/Cardano/Wallet/Util.hs
Martoon-00: Right, that is must-have, added.

core/Pos/Core/Slotting/Timestamp.hs
Martoon-00: It's hard to me to agree, since if I had `t :: Timestamp` corresponding to `12.3 sec` (`Timestamp 12300000` inside), then this "function", applied to `t`, would return `12.3` and that would mean `12.3 sec`, not `12.3 mcs` :slightly_smiling_face: 
Anyway thanks for involving another couple of eyes into it, hope we'll reach a consensus :eyes:

Martoon-00:
Several updates since last review:
* Made `parseApiUtcTime` return `Either` instead of `MonadFail m => m` according to convention
* Added roundtrip test
* Added `instance SchemaParam (V1 Core.Timestamp)` - was required by new code in `develop`

core/Pos/Core/Slotting/Timestamp.hs
Martoon-00: Yeah, that `Timestamp` and lens coupling is a bit tricky, perhaps I did a wrong thing when didn't insert an explaining example in haddock.

adinapoli-iohk:
@KtorZ Just to put this on your radar, this might be a good candidate as your next step -- I think @Martoon-00 brought it almost to completion but we just need to do the final mile and make sure everything works as expected 😉

core/Pos/Core/Slotting/Timestamp.hs
```diff
@@ -50,6 +60,28 @@ instance NFData Timestamp where
...
+        parseTimeM True defaultTimeLocale fmt str
+    utcTimeParser =
+        view (from timestampToUTCTimeL)
+        <$> asum
+            [ parseFmt (iso8601DateFormat (Just "%H:%M:%S%Q"))
+            , parseFmt (iso8601DateFormat Nothing)
+            ]
+    timePosixParser =
+        view (from timestampSeconds)
+        <$> readMaybe @Double str
```
parsonsmatt: I have factored this to a more central location for use in both the `FromJSON` instance and the `ToIndex` instance.

wallet-new/src/Cardano/Wallet/API/V1/Wallets.hs
parsonsmatt: These type aliases may not be useful if we go with the `"query_param" ?= ParamType` design in #2658.

wallet-new/test/MarshallingSpec.hs
```diff
@@ -67,6 +78,48 @@ spec = describe "Marshalling & Unmarshalling" $ do
                 -- currently passphrase should be either empty or of length 32
                 decodingFails @(V1 Crypto.PassPhrase) "aabbcc" Proxy
 
+    describe "Timestamp Parsing" $ do
```
parsonsmatt: Unit tests <3

adinapoli-iohk:
@domenkozar , @rvl Sob, do you guys know why Appveyor is failing here?

parsonsmatt:
I'll fix up whatever PR gets merged second to not have the aliases. No point in doing it preemptively a there will be a conflict.

PR #2612, opened 2018-03-03
[CSL-1786] Implement download timeout

Dolgakov:
Implemented withTimeOut according to https://iohk.myjetbrains.com/youtrack/issue/CSL-1786.
If maxLatency time (30 min for now) exceded Timeout error returned as (Left <Text> ) and logged with other errors.

### DIFF
--- a/update/Pos/Update/Download.hs
+++ b/update/Pos/Update/Download.hs
@@ -11,7 +11,7 @@ import           Universum
 
 import           Control.Exception.Safe (handleAny)
 import           Control.Lens (views)
-import           Control.Monad.Except (ExceptT (..), throwError)
+import           Control.Monad.Except (ExceptT (..))
 import qualified Data.ByteArray as BA
 import qualified Data.ByteString.Lazy as BSL
 import qualified Data.HashMap.Strict as HM
@@ -22,7 +22,7 @@ import           Network.HTTP.Simple (getResponseBody, getResponseStatus, getRes
                                       httpLBS, parseRequest, setRequestManager)
 import qualified Serokell.Util.Base16 as B16
 import           Serokell.Util.Text (listJsonIndent, mapJson)
-import           System.Directory (doesFileExist)
+import           System.Directory (doesFileExist, removeFile)
 import           System.Wlog (WithLogger, logDebug, logInfo, logWarning)
 
 import           Pos.Binary.Class (Raw)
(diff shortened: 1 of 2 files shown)


update/Pos/Update/Worker.hs
```diff
@@ -12,6 +12,9 @@ import           Formatting (build, sformat, (%))
 import           Serokell.Util.Text (listJsonIndent)
 import           System.Wlog (logDebug, logInfo)
 
+import           Control.Lens (views)
```
gromakovsky: This import should be in the group above, it's not special in any sense, so it shouldn't be separated.

update/Pos/Update/Worker.hs
```diff
@@ -67,14 +71,22 @@ checkForUpdate = do
...
+                    %listJsonIndent 2%
+                    "\n The newest one is: "%build%" and we want to download it")
+                    (cpsUpdateProposal <$> confirmedProposals)
+                    (cpsUpdateProposal newestCPS)
+            downloadUpdate newestCPS
+
+    noServersProvided :: ConfirmedProposalState -> m Bool
+    noServersProvided ConfirmedProposalState {..} = do
+        updateServers <- views (lensOf @UpdateParams) upUpdateServers
+        return (updateServers == [])
```
gromakovsky: `== []` is `null` (`hlint` also complained about that).

update/Pos/Update/Download.hs
```diff
@@ -210,6 +223,6 @@ downloadUri manager uri h = do
 * check timeouts?
```
gromakovsky: You can (and should!) remove this line now.

update/Pos/Update/Download.hs
```diff
@@ -191,6 +195,15 @@ downloadHash updateServers h = do
...
     showHash = toString . B16.encode . BA.convert
 
+    withTimeOut :: Int 
+                -> IO (Either Text LByteString)
+                -> IO (Either Text LByteString)
+    withTimeOut maxLatency action = do
+        timeout maxLatency action >>= \case 
+            Nothing -> return $ Left "Timeout occured while downloading an update"
+            Just (Left e) -> return $ Left e
+            Just (Right r) -> return $ Right r
```
gromakovsky: You could've used `<&>` instead of `>>=` above to avoid these three `return`s.

update/Pos/Update/Download.hs
```diff
@@ -171,10 +174,11 @@ downloadHash updateServers h = do
     manager <- liftIO $ newManager tlsManagerSettings
 
     let -- try all servers in turn until there's a Right
+        maxLatency = 2 * 10 ^ (9 :: Int) -- 30 min is enough?
```
gromakovsky: We tend to avoid using types like `Int` for time units as much as possible. We are currently using [time-units](https://hackage.haskell.org/package/time-units-1.0.0/docs/Data-Time-Units.html) library instead (and later we'll switch to `o-clock`, but that's another story).
It would be better to write `maxLatency = 30 :: Minute` and then convert it to `Int` right before passing to `timeout`.
[o-clock](https://github.com/serokell/o-clock) has `timeout` function with better type, so when we switch to it, conversion to `Int` won't be necessary.

> 30 min is enough?

Typical installer size is about 100 Mb. If download speed is 64 kilobytes per second (I think that's a reasonable lower bound), they will download installer in 27 minutes. So 30 minutes should be enough.
But actually in 99% cases the limit should be about 5 minutes (assuming more or less decent internet connection).
So I would use a bit more complex logic: try with limit ≈ 5 min and then ≈ 30 minutes.

update/Pos/Update/Download.hs
```diff
@@ -171,10 +174,11 @@ downloadHash updateServers h = do
     manager <- liftIO $ newManager tlsManagerSettings
 
     let -- try all servers in turn until there's a Right
```
gromakovsky: This comment is a bit misplaced now.

update/Pos/Update/Download.hs
```diff
@@ -170,11 +174,13 @@ downloadHash ::
...
 
-    let -- try all servers in turn until there's a Right
+    let
+        maxLatency = 30 :: Minute
+        -- try all servers in turn until there's a Right
         go errs (serv:rest) = do
             let uri = toString serv <//> showHash h
             logDebug $ "Trying url " <> show uri
-            liftIO (downloadUri manager uri h) >>= \case
+            liftIO (withTimeOut ((fromInteger . toMicroseconds) maxLatency) (downloadUri manager uri h)) >>= \case
```
gromakovsky: It would be even better to pass `Minute` to `withTimeOut` and convert it to `Int` there. Also you can put `maxLatency` inside `where` block and then you won't need to pass this value (but that's subjective).

update/Pos/Update/Download.hs
```diff
@@ -191,6 +197,14 @@ downloadHash updateServers h = do
     showHash :: Hash a -> FilePath
     showHash = toString . B16.encode . BA.convert
 
+    withTimeOut :: Int
+                -> IO (Either Text LByteString)
+                -> IO (Either Text LByteString)
+    withTimeOut maxLatency action = do
```
gromakovsky: To be honest I've never seen capital `O` in the word `timeout` :thinking:

update/Pos/Update/Worker.hs
```diff
@@ -67,14 +69,22 @@ checkForUpdate = do
...
+                sformat
+                    ("There are new confirmed update proposals for our application: "
+                    %listJsonIndent 2%
+                    "\n The newest one is: "%build%" and we want to download it")
+                    (cpsUpdateProposal <$> confirmedProposals)
+                    (cpsUpdateProposal newestCPS)
+            downloadUpdate newestCPS
+
+    noServersProvided :: ConfirmedProposalState -> m Bool
+    noServersProvided ConfirmedProposalState {..} = do
```
gromakovsky: Maybe I don't understand something, but it seems to me that `ConfirmedProposalState` is not used in this function.

gromakovsky: And btw, I don't understand why this change is part of this PR, I expected it to be part of CSL-1602 (and it is, but for some reason it's in both PRs).

jcmincke:

PR #2615, opened 2018-03-05
[CSL-2341] Write logical integration tests.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2341

An example of the idea I had of how to use randomized integration testing.

### DIFF
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -225,7 +225,8 @@ executable cardano-integration-test
                        -Wall
                        -fno-warn-orphans
                        -O2
-  other-modules:       Types
+  other-modules:       CLI
+                       Types
                        Functions
                        Error
   build-depends:       base
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -265,7 +265,6 @@ executable cardano-integration-test
                      , servant-swagger
                      , serokell-util
                      , memory
-                     , MonadRandom
                      , stm
                      , string-conv
                      , swagger2
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -218,6 +218,85 @@ executable cardano-node
                       UndecidableInstances
                       MonadFailDesugaring
 
+executable cardano-integration-test
+  hs-source-dirs:      integration
+  main-is:             Main.hs
+  ghc-options:         -threaded -rtsopts
+                       -Wall
+                       -fno-warn-orphans
+                       -O2
+  other-modules:       Types
+                       Functions
+                       Error
+  build-depends:       base
+                     , QuickCheck
+                     , aeson
+                     , aeson-pretty
+                     , bytestring
+                     , cardano-sl
+                     , cardano-sl-client
+                     , cardano-sl-core
+                     , cardano-sl-crypto
+                     , cardano-sl-client
+                     , cardano-sl-util
+                     , cardano-sl-update
+                     , cardano-sl-infra
+                     , cardano-sl-networking
+                     , cardano-sl-ssc
+                     , cardano-sl-txp
+                     , cardano-sl-wallet
+                     , cardano-sl-wallet-new
+                     , containers
+                     , data-default
+                     , exceptions
+                     , formatting
+                     , formatting
+                     , http-api-data
+                     , http-types
+                     , insert-ordered-containers
+                     , ixset-typed
+                     , lens
+                     , log-warper
+                     , mtl
+                     , neat-interpolation
+                     , optparse-applicative
+                     , servant
+                     , servant-server
+                     , servant-swagger
+                     , serokell-util
+                     , memory
+                     , MonadRandom
+                     , stm
+                     , string-conv
+                     , swagger2
+                     , text
+                     , text-format
+                     , time-units
+                     , universum
+                     , unordered-containers
+                     , wai
+                     , wai-cors
+                     , wai-extra
+                     , warp
+  default-language:    Haskell2010
+  default-extensions: TypeOperators
+                      DataKinds
+                      DefaultSignatures
+                      NoImplicitPrelude
+                      MultiParamTypeClasses
+                      OverloadedStrings
+                      ScopedTypeVariables
+                      FlexibleInstances
+                      FlexibleContexts
+                      TypeFamilies
+                      TypeApplications
+                      TypeOperators
+                      TemplateHaskell
+                      RecordWildCards
+                      ScopedTypeVariables
+                      UndecidableInstances
+                      MonadFailDesugaring
+
 test-suite wallet-unit-tests
   ghc-options:        -Wall
   type:               exitcode-stdio-1.0
--- a/wallet-new/integration/Error.hs
+++ b/wallet-new/integration/Error.hs
@@ -31,6 +31,8 @@ data WalletTestError
     | AddressBalanceNotZero WalletAddress
 
     | InvalidTransactionState Transaction
+    | LocalTransactionsDiffer [Transaction]
+    | LocalTransactionMissing Transaction [Transaction]
 
     deriving (Show, Eq, Generic)
 
@@ -39,16 +41,18 @@ instance Exception WalletTestError
 
 
 instance Buildable WalletTestError where
-    build (HttpClientError _        )  = bprint "Http client error"
+    build (HttpClientError _        )     = bprint "Http client error"
     -- ^ TODO (ks): A proper instance
-    build (WalletBalanceNotZero    w)  = bprint ("Wallet balance is not zero. Wallet - ("%stext%")") (show w)
-    build (LocalWalletDiffers      w)  = bprint ("Local wallet differs. Wallet - ("%stext%")") (show w)
-    build (LocalWalletsDiffers     w)  = bprint ("Local wallets differs. Wallet - ("%stext%")") (show w)
+    build (WalletBalanceNotZero    w)     = bprint ("Wallet balance is not zero. Wallet - ("%stext%")") (show w)
+    build (LocalWalletDiffers      w)     = bprint ("Local wallet differs. Wallet - ("%stext%")") (show w)
+    build (LocalWalletsDiffers     w)     = bprint ("Local wallets differs. Wallet - ("%stext%")") (show w)
 
-    build (AccountBalanceNotZero   a)  = bprint ("Acccount balance is not zero. Account - ("%stext%")") (show a)
-    build (LocalAccountDiffers     a)  = bprint ("Local account differs. Account - ("%stext%")") (show a)
-    build (LocalAccountsDiffers    a)  = bprint ("Local accounts differs. Account - ("%stext%")") (show a)
-    build (AddressBalanceNotZero   a)  = bprint ("Address balance is not zero. Address - ("%stext%")") (show a)
-    build (InvalidTransactionState t)  = bprint ("Transaction state is invalid. Transaction - ("%stext%")") (show t)
+    build (AccountBalanceNotZero   a)     = bprint ("Acccount balance is not zero. Account - ("%stext%")") (show a)
+    build (LocalAccountDiffers     a)     = bprint ("Local account differs. Account - ("%stext%")") (show a)
+    build (LocalAccountsDiffers    a)     = bprint ("Local accounts differs. Account - ("%stext%")") (show a)
+    build (AddressBalanceNotZero   a)     = bprint ("Address balance is not zero. Address - ("%stext%")") (show a)
+    build (InvalidTransactionState t)     = bprint ("Transaction state is invalid. Transaction - ("%stext%")") (show t)
+    build (LocalTransactionsDiffer t)     = bprint ("Local transactions differs. Transactions - ("%stext%")") (show t)
+    build (LocalTransactionMissing t ts)  = bprint ("Local transaction ("%stext%") missing from txs history ("%stext%")") (show t) (show ts)
 
 
--- a/wallet-new/integration/Error.hs
+++ b/wallet-new/integration/Error.hs
@@ -12,12 +12,22 @@ import           Universum
 import qualified Data.Text.Buildable
 import           Formatting (bprint, stext, (%))
 
+import           Cardano.Wallet.API.V1.Types (Account (..), Wallet (..))
+
 import           Cardano.Wallet.Client (ClientError)
 
+
 data WalletTestError
-    = Internal Text
-    | HttpClientError ClientError
-    | ServerConnectionFailed
+    = HttpClientError ClientError
+
+    | WalletBalanceNotZero Wallet
+    | LocalWalletDiffers Wallet
+    | LocalWalletsDiffers [Wallet]
+
+    | AccountBalanceNotZero Account
+    | LocalAccountDiffers Account
+    | LocalAccountsDiffers [Account]
+
     deriving (Show, Eq, Generic)
 
 
@@ -25,7 +35,14 @@ instance Exception WalletTestError
 
 
 instance Buildable WalletTestError where
-    build (Internal msg)           = bprint ("Wallet test error ("%stext%")") msg
-    build (ServerConnectionFailed) = bprint "Server connection failed"
-    -- TODO (ks): A proper instance
-    build (HttpClientError _     ) = bprint "Http client error"
+    build (HttpClientError _      )  = bprint "Http client error"
+    -- ^ TODO (ks): A proper instance
+    build (WalletBalanceNotZero  w)  = bprint ("Wallet balance is not zero. Wallet - ("%stext%")") (show w)
+    build (LocalWalletDiffers    w)  = bprint ("Local wallet differs. Wallet - ("%stext%")") (show w)
+    build (LocalWalletsDiffers   w)  = bprint ("Local wallets differs. Wallet - ("%stext%")") (show w)
+
+    build (AccountBalanceNotZero a)  = bprint ("Acccount balance is not zero. Account - ("%stext%")") (show a)
+    build (LocalAccountDiffers   a)  = bprint ("Local account differs. Account - ("%stext%")") (show a)
+    build (LocalAccountsDiffers  a)  = bprint ("Local accounts differs. Account - ("%stext%")") (show a)
+
+
--- a/wallet-new/integration/Error.hs
+++ b/wallet-new/integration/Error.hs
@@ -24,6 +24,7 @@ data WalletTestError
     | InvalidProbabilityDistr
 
     | WalletBalanceNotZero Wallet
+    | WalletPassMissing Wallet
     | LocalWalletDiffers Wallet
     | LocalWalletsDiffers [Wallet]
 
@@ -51,6 +52,7 @@ instance Buildable WalletTestError where
     -- ^ TODO (ks): A proper instance
     build InvalidProbabilityDistr         = bprint "The probability distribution should be between 1 - 100."
     build (WalletBalanceNotZero    w)     = bprint ("Wallet balance is not zero - ("%stext%")") (show w)
+    build (WalletPassMissing       w)     = bprint ("Missing wallet pass - ("%stext%")") (show w)
     build (LocalWalletDiffers      w)     = bprint ("Local wallet differs - ("%stext%")") (show w)
     build (LocalWalletsDiffers     w)     = bprint ("Local wallets differs - ("%stext%")") (show w)
 
--- a/wallet-new/integration/Error.hs
+++ b/wallet-new/integration/Error.hs
@@ -7,13 +7,16 @@ module Error
     ( WalletTestError (..)
     ) where
 
-import qualified Data.Text.Buildable
-import           Formatting (bprint, stext, (%))
 import           Universum
 
+import qualified Data.Text.Buildable
+import           Formatting (bprint, stext, (%))
+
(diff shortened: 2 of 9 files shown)


wallet-new/integration/Error.hs
```diff
@@ -0,0 +1,22 @@
...
+    ( IntegrationTestError (..)
+    ) where
+
+import qualified Data.Text.Buildable
+import           Formatting (bprint, stext, (%))
+import           Universum
+
+newtype IntegrationTestError =
+    -- | Some internal error.
+    Internal Text
```
parsonsmatt: I would prefer to see constructors rather than `Text` errors. We can easily add constructors when we need them, but factoring out from a `Text` representation is more work.

wallet-new/integration/Types.hs
```diff
@@ -0,0 +1,88 @@
...
+data Probability = Probability { getProbability :: Int }
+    deriving (Show, Eq)
+
+
+-- | Safe constructor that checks the values.
+createProbability :: Int -> Probability
+createProbability prob
+    -- TODO(ks): Maybe we should @throwM@ or @either@ here.
+    | prob <= 0 || prob > 100 =
+        error "Probability should be between 1 - 100."
```
parsonsmatt: Smart constructors should use `Maybe`, not `error`.

wallet-new/integration/Types.hs
```diff
@@ -0,0 +1,88 @@
...
+    | CreateTransaction
+    | GetTransaction
+    deriving (Show, Eq)
+
+
+-- | The type that defines the probabilites of the actions
+-- TODO(ks): We could create a custom constructor with valid
+-- values only.
+-- Add invariant?
+type ActionProbabilities = [(Action, Probability)]
```
parsonsmatt: This is a partial lookup-table -- why not `type ActionProbabilities = Action -> Probability`? Constructing them is nearly as simple, with a guarantee of totality. 

```haskell
foo :: ActionProbabilities 
foo = Probability . \case
  CreateWallet -> 3
  -- etc..
```

wallet-new/integration/Error.hs
ksaric: Fair enough, will fix.

wallet-new/integration/Types.hs
ksaric: Interesting, but AFAIK we don't require a guarantee of totality, we just need to check if the probabilities add up to 100 (comment above).

parsonsmatt: Another thing to note is `QuickCheck`'s `frequency` function -- it uses something like:

```
frequency :: [(Int, Gen a)] -> Gen a
frequency = join . elements . concatMap (\(i, g) -> replicate i g)
```

which essentially uses the sum of the weights in the list to normalize. This eliminates partiality *and* the requirement that the probability sum to 100.

parsonsmatt:
I've made #2624 for the client DSL.

wallet-new/integration/Types.hs
```diff
@@ -0,0 +1,88 @@
...
+    , ActionWalletState
+    , ActionProbabilities
+    , WIntTestMode
+    ) where
+
+import Universum
+
+import Cardano.Wallet.API.V1.Types (Wallet)
+
+type WIntTestMode m =
```
adinapoli-iohk: I'm not very sold to this "url-shortened" names: I would personally prefer `IntegrationTestMode` or simply `WalletTestMode`, (or even just `TestMode`), but `WIntTestMode` doesn't communicate a great deal more and that's just noise at this stage (my 2 cents anyway).

wallet-new/integration/Functions.hs
```diff
@@ -0,0 +1,92 @@
...
+    result <- runAction action
+
+    validateInvariants result walletState
+
+-- | Get action randomly, depending on the action distribution
+chooseAction
+    :: (WIntTestMode m)
+    => ActionProbabilities
+    -> m Action
+chooseAction aProb = do
```
adinapoli-iohk: I'm not sure I understand this code entirely. It seems like it's starting with a probability distribution and it's generating a big list with a number of elements proportional to the probability. Example:

```haskell
[GetWallet, GetWallet, GetWallet, GetAccount, CreateTransaction, ...]
```

So that the more frequent the action the more the probability we pick it with `distributedActions ^? ix prob`.

Am I right or I'm still caffeine deprived?

I'm wondering if we are not overcomplicating this -- cannot we simply reify our probability map into a `Gen` and then use `frequency` from QuickCheck to achieve the same thing?

wallet-new/integration/Functions.hs
```diff
@@ -0,0 +1,92 @@
...
+    prob <- liftIO chooseProb
+
+    -- @choice@ from random-extras-0.2?
+    let distributedActions :: [Action]
+        distributedActions = concatMap replDistribution aProb
+
+    maybe actionNotFound pure (distributedActions ^? ix prob)
```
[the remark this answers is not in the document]
ksaric: Yeah, I used it before and can switch to it, not really that much difference.

wallet-new/integration/Functions.hs
```diff
@@ -0,0 +1,92 @@
...
+  where
+    chooseProb :: IO Int
+    chooseProb = getStdRandom (randomR (1,100))
```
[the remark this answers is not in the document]
ksaric: Ok, will fix.

wallet-new/integration/Types.hs
ksaric: I agree, but the name I came up with was a bit too long. The problem is how to pack information in few words. Will fix.

wallet-new/integration/Functions.hs
ksaric: You are correct.

I don't think we are overcomplicating this, the idea is to have an action distribution close to what a user has and to leave it running. After a while, the test will steer into some random state space that will be close to something a "real user" does.

Also, with those probabilities, we can include different "user profiles" (different action distributions) which can then test the application with the similar actions user from that profile would do. For example, we can have an "exchange user" and a "normal user" and test the application from those perspectives.
Not to mention that we can "translate" this into a Daedalus integration test (just an idea, but interesting nonetheless).

adinapoli-iohk: My gripe is not with passing as input the `ActionProbabilities` (which can probably be renamed later into something which better reflect this idea of user profiles) but it should be totally possible to derive a `Gen` out of such data type (with a function of type `ActionProbabilities -> Gen Action`) which will compose way better with the way QuickCheck works.

ksaric: Oh, I see. I guess. I mean it really depends what we plan to use. I still think that using `IO` is the way to go since we do need the external interaction when communicating with the node, but we can use `Gen` if QuickCheck is going to be our "main weapon".
I don't see much difference myself, TBH, except the fact we are using something a bit heavier (and more powerful) like `IO`. 
I would go for this approach until we have a working version and then we can optimize.

edsko: IO is almost never the way go for functions that don't need it. After all, we program in Haskell because we like things like purity, referential transparency, etc. Ideally a program is a thin side-effect-full (IO) layer on top of an otherwise pure program. QuickCheck's `Gen` monad is designed specifically for generating arbitrary data; it seems an obvious choice here -- not to mention that we may want to run these tests without any IO involved at all (if not right now, then in the future). Taking pure code and using in a stateful context is easy; starting with effectful code and attempting to use it in a pure context can often be very difficult. `IO` should never be the default starting point.

ksaric: Ok guys, will fix :smile: 

Regarding `IO` - I still feel it's very useful when you are prototyping something and don't want to "optimize prematurely". I don't know about you guys, but I rarely have everything "in my head". The task of programming for me is exploring the domain I'm in, and the speed at which I can do that is critical - not having to think about which Monad (domain) I'm in is a very useful thing. 
I rarely look at the code I put in the editor and go "clearly this is a MonadState with MonadV1 witch a pinch of MonadThatCantBeNeverFullyUnderstood". 
Then again, that's just me.

edsko: Let me answer that question in earnest. Definitely, I agree both that premature optimization should be avoided and that when prototyping something we don't have all the details in our head and we don't yet fully know what the full context of this function will be. But here's the thing: types can _help_ with simplifying things. As soon as you declare your function to be in IO, basically all bets are off. _IO is as powerful as every other monad out there_ (including "`MonadState` with `MonadV1` and a pinch of `MonadThatCanNeverBeFullyUnderstood` :slightly_smiling_face:). When I prototype something, I start with the simplest type I can possibly imagine for this function, _and IO isn't it_. If I say `a -> a`, my function can only do one thing (must be the identity); if I say `Int -> Int`, I can do any operation on integers but nothing else; when I write `Gen Int`, I have a generator available. The point being that picking a simple type helps with making decisions. A type like `IO a` gives me no guidance at all because I can do _anything_. 

In fact, not only do I agree that premature optimization should be avoided, I agree that complicated types should be avoided whenever possible (this is an area where the Cardano codebase can be significantly improved). Understanding a function in `MonadThatCanNeverBeFullyUnderstood` indeed _is_ hard. It's one of the things that makes perusing the Cardano code base hard to newcomers (as well as making it hard to refactor it, or apply existing functions in different contexts, such as a pure unit testing context).

Keeping things simple is important; and _IO is not simple_.

parsonsmatt:
@ksaric Can you revert the changes that implement parts of the Servant API, and that change the API type? This PR's scope should be limited to implementing logical tests using the Client interface defined in #2624 . Any changes to the API or other parts of the codebase will introduce conflicts with other PRs that work with the API (eg #2630 and the related documentation work; actually implementing the client; etc)

ksaric:
> Can you revert the changes that implement parts of the Servant API, and that change the API type?

@parsonsmatt Fine.
I will continue work with https://github.com/input-output-hk/cardano-sl/pull/2624 then, but it would be nice if I could commit the changes I have in some other PR.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -371,7 +371,7 @@ data WalletAddress = WalletAddress
   , addrBalance       :: !(V1 Core.Coin)
   , addrUsed          :: !Bool
   , addrChangeAddress :: !Bool
-  } deriving (Show, Generic)
+  } deriving (Show, Eq, Generic)
```
akegalj: I am not sure does it makes sense to add `Eq` here ? Some might think that `WalletAddress` should implement `Eq` in a way where `walletId`s are the same. It might be confusing as the name suggests that it is an address, and two addresses should be the same depending on their IDs, excluding other metadata.

But I might be wrong.

ksaric: I'm not sure I understand. The datatype will be equal if all the fields inside are equal. Not really sure why that's not valid.

wallet-new/integration/Functions.hs
```diff
@@ -162,6 +176,69 @@ runAction wc ws CreateAddress = do
...
+    guard (length localAddresses      >= 2)
+    guard (length localAccsWithMoney  >= 1)
+
+    -- From which source to pay.
+    accountSource <- pickRandomElement localAccsWithMoney
+
+    let _accountSourceMoney = accAmount accountSource
+
+    -- We should probably have a sensible minimum value.
+    -- moneyAmount <- liftIO $ mkCoin $ generate $ choose (0, getCoin accountSourceMoney)
```
akegalj: yes, with
```haskell
 moneyAmount <- liftIO $ generate arbitrary
```
creating transaction will probably fail due too fees. I believe your intuition is right here to  generate something minimum for fee to be included.

edsko:
@ksaric I'm not following all the details of this PR (so this is not in any way meant as a criticism on anything you wrote as such) -- but just be aware that as part of the new wallet design we have

* QuickCheck generators for valid blockchains based on the formal definition of UTxO style accounting
* QuickCheck generators for "inductive wallets" (basically saying "apply this block .. submit this transaction .. apply that block .. submit that transaction .." etc) based on the formal wallet specification.

(@parsonsmatt is fully aware of this of course as he wrote them :slightly_smiling_face: ) . So I'd just say, don't try to be too too clever in how you test things (in terms of generating valid sequences of blocks and transactions); just do the minimum possible here, and then we can later plug in the generators I mentioned above.

ksaric:
>  just do the minimum possible here, and then we can later plug in the generators I mentioned above.

@edsko Right, we have that in mind, that was the plan (at least I thought it was), this will be just the first naive implementation and the rest will follow (the generators).

adinapoli-iohk:
I'm with @edsko here -- let's build up the minimum necessary for the scaffolding to be in place, but without descending into the rabbit hole of generating via UTXO & transactions etc. Then we will revisit integration tests properly after we land the new data layer.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: I think that the `WalletAddress` type represents both the address + metadata about it. If you wanted to compare two `WalletAddresses` and only focus on a field, there's always ```(==) `on` addrId```

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -149,3 +162,18 @@ data ClientError
     -- ^ This constructor is used when the API client reports an error that
     -- isn't represented in either the 'ServantError' HTTP errors or the
     -- 'WalletError' for API errors.
+    deriving (Show)
+
+-- | General (and naive) equality instance.
+instance Eq ClientError where
+    (==) (ClientWalletError e1) (ClientWalletError e2) = e1 == e2
+    (==) (ClientHttpError   e1) (ClientHttpError   e2) = e1 == e2
+    (==) (UnknownError      _ ) (UnknownError      _ ) = True
```
parsonsmatt: it just occured to me that you could write `UnknownError e1 == UnknownError e2 = displayException e1 == displayException e2` but that's a little dastardly

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -78,7 +88,7 @@ data WalletClient m
     -- transactions endpoints
     , postTransaction
          :: Payment -> Resp m Transaction
-    , getTransactionIndex
+    , getTransactionHistory
```
parsonsmatt: Please revert the cosmetic changes to the API. This is going to change a bit anyway when the client is merged; I'd prefer not to have more conflicts than necessary.

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -62,13 +66,19 @@ data WalletClient m
...
          :: WalletId -> Resp m Wallet
     , updateWallet
          :: WalletId -> Update Wallet -> Resp m Wallet
     -- account endpoints
+    , postAccount
+         :: New Account -> Resp m Account
     , deleteAccount
          :: WalletId -> AccountIndex -> Resp m ()
+    , getAccounts
+         :: WalletId -> Resp m [Account]
```
parsonsmatt: `getAccounts` will need to be a top level alias; the actual function will look something like

```haskell
getAccountsPaged
    :: WalletId -> Maybe Page -> Maybe PerPage -> Resp m [Account]
```

parsonsmatt: You can change this now or I can change it when I merge the client implementation in.

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -44,10 +45,13 @@ import qualified Pos.Core as Core
 --
 -- Other functions may be defined in terms of this 'WalletClient' -- see
 -- 'getWalletIndex' as a convenience helper for 'getWalletIndexPaged'.
+-- TODO(ks): I don't think that it's important to preserve paging as
+-- an important detail, we should remove paging and return the full set
+-- of results.
```
parsonsmatt: The `WalletClient m` should contain the primitives of the API. We can easily implement something like `streamAddresses :: WalletClient m -> Producer m Address` that pages through on top if we need.

wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -225,6 +225,85 @@ executable cardano-node
                       UndecidableInstances
                       MonadFailDesugaring
 
+executable cardano-integration-test
```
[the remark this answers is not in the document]
ksaric: Because it requires the full wallet node to be running, and currently we don't have any easy way to run it from the code. Nothing stops us from changing this in the future, and this is indeed what I had in mind myself.
The current issue is the lack of `cardano-runner` or similar which is required for `dbgen` as well, not to mention that it would be very beneficial to replace the bash scripts with an actual configurable and flexible Haskell implementation.

wallet-new/integration/Error.hs
```diff
@@ -0,0 +1,67 @@
...
+    | AddressBalanceNotZero WalletAddress
+    | LocalAddressesDiffer Address [Address]
+    | LocalAddressDiffer Address
+
+    | InvalidTransactionState Transaction
+    | InvalidTransactionFee EstimatedFees
+    | LocalTransactionsDiffer [Transaction]
+    | LocalTransactionMissing Transaction [Transaction]
+
+    deriving (Show, Eq, Generic)
```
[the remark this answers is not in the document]
ksaric: Yes, we could also print `WalletState`, but I guess that's optional since it will probably be clear from the current context what the error/issue is.

wallet-new/integration/Error.hs
```diff
@@ -0,0 +1,67 @@
...
+    build (AddressBalanceNotZero   a)     = bprint ("Address balance is not zero - ("%stext%")") (show a)
+    build (LocalAddressesDiffer a as)     = bprint ("Local address ("%stext%") missing from addresses ("%stext%")") (show a) (show as)
+    build (LocalAddressDiffer      a)     = bprint ("Local address differs - ("%stext%")") (show a)
+
+    build (InvalidTransactionState t)     = bprint ("Transaction state is invalid. Transaction - ("%stext%")") (show t)
+    build (InvalidTransactionFee   f)     = bprint ("Transaction fees are invalid - ("%stext%")") (show f)
+    build (LocalTransactionsDiffer t)     = bprint ("Local transactions differs - ("%stext%")") (show t)
+    build (LocalTransactionMissing t ts)  = bprint ("Local transaction ("%stext%") missing from txs history ("%stext%")") (show t) (show ts)
+
+
```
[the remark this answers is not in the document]
ksaric: Heh, fixed.

wallet-new/integration/Functions.hs
```diff
@@ -0,0 +1,503 @@
...
+    newWall <-  liftIO $ generate arbitrary
+    result  <-  respToRes $ postWallet wc newWall
+
+    checkInvariant
+        (walBalance result == minBound)
+        (WalletBalanceNotZero result)
+
+    -- Modify wallet state accordingly.
+    pure $ ws
+        & wallets    .~ ws ^. wallets <> [result]
```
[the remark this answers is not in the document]
ksaric: Thanks, fixed.

wallet-new/integration/Types.hs
```diff
@@ -0,0 +1,104 @@
...
+-- our random action selection choosing a specific @Action@.
+-- The number is restricted to 1 - 100 - see @createProbability@.
+data Probability = Probability { getProbability :: Int }
+    deriving (Show, Eq)
+
+
+-- | Safe constructor that checks the values.
+createProbability :: Int -> Probability
+createProbability prob
+    | prob <= 0 || prob > 100 = error "Invalid range. 0 - 100."
```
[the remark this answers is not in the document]
ksaric: Regarding this, I do have something to comment. I really don't like the idea of using `Maybe` in places where we can have an error. 
In this scenario, it's much more natural to use `Either` and I used it. Running a function and getting `Nothing` pisses me off since I have to look at "Why?", while `Either` can immediately tell you what's the issue.
All that aside, I don't really see the benefit of using some additional constructs besides `error` here since the end result in this case will be the same - the user will see an error pop up and the error will be "Wrong probability ...". Sometimes less is more, and I think this is one of these cases. I know we are all pure and shiny, but if I have a choice to do something more complicated with no change in code semantics, I don't see the point :hear_no_evil:

wallet-new/src/Cardano/Wallet/Client.hs
ksaric: Done. I do think the naming is a bit off.

wallet-new/src/Cardano/Wallet/Client.hs
ksaric: You can change it, I would prefer if we drop this and use your `streamA :: WalletClient m -> Producer m A`.

parsonsmatt: I can implement that on top of the paginated version, but the paginated version will be in the client record as a base.

wallet-new/cardano-sl-wallet-new.cabal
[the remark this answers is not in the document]
parsonsmatt: Sounds great :smile:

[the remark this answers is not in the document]
adinapoli-iohk: @ksaric Can you please log this as a ticket, so we don't forget? (i.e. turning this into a `test-suite` so CI can run it). Not entirely sure if we have at this stage a parent user story to reference this for, but better logged as an orphan task that not logged at all, imho 😉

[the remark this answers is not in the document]
ksaric: Here we go - https://iohk.myjetbrains.com/youtrack/issue/CSL-2401

PR #2620, opened 2018-03-05
[DEVOPS-660] relay enqueue policy for edge node block req

avieth:
~~A relay yaml policy was included just for the record. It matches the old
relay policy (also changed in this commit) but it seems that the yaml
file can't express everything that the Haskell datatype can: in
particular, a different policy for block header requests depending on
whether we know the NodeIds of peers to request from.~~

Default relay policy updated so that they will make requests from edge node for block headers and blocks.

### DIFF
--- a/infra/Pos/Network/Policy.hs
+++ b/infra/Pos/Network/Policy.hs
@@ -61,13 +61,14 @@ defaultEnqueuePolicyRelay = go
         EnqueueAll NodeCore  (MaxAhead 2) PHigh
       , EnqueueAll NodeRelay (MaxAhead 2) PHigh
       ]
+    -- We can request for headers from edge nodes, if we received one
+    -- unsolicited from them. In that case, we'll have 'Just' here.
     go (MsgRequestBlockHeaders (Just _)) = [
-        -- We never ask for data from edge nodes
-        EnqueueOne [NodeRelay, NodeCore] (MaxAhead 3) PHigh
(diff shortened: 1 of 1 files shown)


policy_relay.yaml
```diff
@@ -0,0 +1,104 @@
...
+        maxAhead: 2
+        precedence: 'high'
+    - all:
+        nodeType: 'relay'
+        maxAhead: 2
+        precedence: 'high'
+
+  requestBlocks:
+    - one:
+        nodeTypes: ['relay', 'core']
```
kantp: Shouldn't this be ['relay', 'core', 'edge']?

kantp: Good idea.

avieth: The reason I didn't bother is that the current yaml parser for policies can't express different enqueue policies for `MsgRequestBlockHeaders` depending on whether or not we have a particular set of `NodeId`s to restrict it to.

kantp:
Two more questions -- just because I'm not 100% up to speed with the current code, and want to make sure I'm not missing anything:
- For transactions, there are `send` and `forward` entries in the `yaml` file, but there's only one entry in the default policy in the code. Is that correct?
- Should the `go (MsgRequestBlockHeaders (Just _))` entry in the code have something corresponding to it in the `yaml` file?

avieth:
@dcoutts @kantp I've removed the yaml changes

PR #2624, opened 2018-03-05
[CSL-2340] Implement a client DSL for testing the wallet

parsonsmatt:
This PR implements a client abstraction for the wallet.

The implementation is left as a stub to be picked up in `CSL-2342`.

The ticket `CSL-2341` can now be implemented in terms of this wallet abstraction, while the actual implementation of the HTTP client can be done in parallel. Testing functions can be written as:

```haskell
testSomeInvariant :: (MonadIO m) => WalletClient m -> m ()
testSomeInvariant wc = ...
```

### DIFF
--- a/wallet-new/src/Cardano/Wallet/Client.hs
+++ b/wallet-new/src/Cardano/Wallet/Client.hs
@@ -1,0 +1,76 @@
+-- | This module defines a client for the Cardano new-wallet.
+module Cardano.Wallet.Client
+    ( -- * The abstract client
+      WalletClient(..)
+    , getWalletIndex
+    , Resp
+    -- * The type of errors that the client might return
+    , WalletError(..)
+    -- * Reexports
+    , module Cardano.Wallet.API.V1.Types
+    , module Cardano.Wallet.API.V1.Parameters
+    , module Cardano.Wallet.API.Request.Pagination
+    ) where
+
+import           Universum
+
+import           Cardano.Wallet.API.Response
+
+import           Cardano.Wallet.API.Request.Pagination
+import           Cardano.Wallet.API.V1.Errors
+import           Cardano.Wallet.API.V1.Parameters
+import           Cardano.Wallet.API.V1.Types
+import qualified Pos.Core as Core
+
+-- | A representation of a wallet client parameterized over some effect
+-- type @m@.
+--
+-- The record fields serve as the API to the wallet client. Note that the
+-- field 'getAddressIndex' has the type:
+--
+-- @
+-- 'getAddressIndex'
+--     :: 'WalletClient' m
+--     -> m ('Either' 'WalletError' ('WalletResponse' ['Address']))
+-- @
+--
+-- Other functions may be defined in terms of this 'WalletClient' -- see
+-- 'getWalletIndex' as a convenience helper for 'getWalletIndexPaged'.
+data WalletClient m
+    = WalletClient
+    { -- address endpoints
+      getAddressIndex :: Resp m [Address]
+    , postAddress :: NewAddress -> Resp m WalletAddress
+    , getAddressValidity :: Text -> Resp m AddressValidity
+    -- wallets endpoints
+    , postWallet :: New Wallet -> Resp m Wallet
+    , getWalletIndexPaged :: Maybe Page -> Maybe PerPage -> Resp m [Wallet]
+    , updateWalletPassword :: WalletId -> PasswordUpdate -> Resp m Wallet
+    , deleteWallet :: WalletId -> Resp m ()
+    , getWallet :: WalletId -> Resp m Wallet
+    , updateWallet :: WalletId -> Update Wallet -> Resp m Wallet
+    -- account endpoints
+    , deleteAccount :: WalletId -> AccountIndex -> Resp m ()
+    , getAccount :: WalletId -> AccountIndex -> Resp m Account
+    , getAccountIndexPaged :: WalletId -> Maybe Page -> Maybe PerPage -> Resp m [Account]
+    , updateAccount :: WalletId -> AccountIndex -> Update Account -> Resp m Account
+    -- transactions endpoints
+    , postTransaction :: Payment -> Resp m Transaction
+    , getTranasactionIndex :: WalletId -> Maybe AccountIndex -> Maybe (V1 Core.Address) -> Maybe Page -> Maybe PerPage -> Resp m [Transaction]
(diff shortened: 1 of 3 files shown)


wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -0,0 +1,126 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE StrictData #-}
```
adinapoli-iohk: First time I see this pragma used in the wild -- nice! (I remember when I was a naive Haskeller and wanted it badly to strictify-all-the-things 😁 )

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -0,0 +1,126 @@
...
+
+-- | This module defines a client for the Cardano new-wallet.
```
adinapoli-iohk: Bikeshedding: we might want to simply call it "wallet" dropping the `new` prefix, to legitimate it a bit more 😀

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -0,0 +1,126 @@
...
+--
+-- @
+-- 'getAddressIndex'
+--     :: 'WalletClient' m
+--     -> m ('Either' 'WalletError' ('WalletResponse' ['Address']))
+-- @
+--
+-- Other functions may be defined in terms of this 'WalletClient' -- see
+-- 'getWalletIndex' as a convenience helper for 'getWalletIndexPaged'.
+data WalletClient m
```
adinapoli-iohk: Ah, interesting! Eventually you settled for "Option 1" 😛

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -0,0 +1,126 @@
...
+    , updateWalletPassword = \x -> phi . updateWalletPassword wc x
+    , deleteWallet         = phi . deleteWallet wc
+    , getWallet            = phi . getWallet wc
+    , updateWallet         = \x -> phi . updateWallet wc x
+    , deleteAccount        = \x -> phi . deleteAccount wc x
+    , getAccount           = \x -> phi . getAccount wc x
+    , getAccountIndexPaged = \x mp -> phi . getAccountIndexPaged wc x mp
+    , updateAccount        = \x y -> phi . updateAccount wc x y
+    , postTransaction      = phi . postTransaction wc
+    , getTranasactionIndex = \wid maid maddr mp -> phi . getTranasactionIndex wc wid maid maddr mp
```
adinapoli-iohk: Typo police to the rescue: shouldn't this be _getTransactionIndex_? 😁

wallet-new/src/Cardano/Wallet/Client/Http.hs
```diff
@@ -0,0 +1,22 @@
+module Cardano.Wallet.Client.Http
+    ( module Cardano.Wallet.Client.Http
+      -- * Absract Client export
```
ksaric: Small typo.

wallet-new/src/Cardano/Wallet/Client/Http.hs
```diff
@@ -0,0 +1,22 @@
...
+
+import Cardano.Wallet.Client
+
+-- | Given a 'BaseUrl' and an @http-client@ 'Manager', this returns
+-- a 'WalletClient' that operates in 'IO'.
+mkHttpClient
+    :: BaseUrl
+    -> Manager
+    -> WalletClient IO
+mkHttpClient _baseUrl _manager = error "implement me" -- WalletClient{..}
```
ksaric: We should probably wait for this https://github.com/input-output-hk/cardano-sl/pull/2625 so we don't duplicate things. Yes, I know that's for V0, but we require a couple of mind-blowing functions in order for the client to work (https://github.com/input-output-hk/cardano-sl/pull/2625/files#diff-66efa8b5a7226f01ae4cb3aaee278af2R289).

wallet-new/src/Cardano/Wallet/Client.hs
parsonsmatt: Ah!! Yes :smile: Thanks

wallet-new/src/Cardano/Wallet/Client.hs
parsonsmatt: Good call :)

wallet-new/src/Cardano/Wallet/Client.hs
parsonsmatt: Strict record fields = no space leaks, in my experience :)

I mostly want them because then you et a compile time error if you don't initialize all the fields in a record. Doing:

```haskell
data X = X { x :: Int, y :: Int }

oops = y (X { x = 3 })
```

gets a runtime error, while

```haskell
data X = X { x :: !Int, y :: !Int }

oops = X { x = 3 }
```

gets you a compile time error.

wallet-new/src/Cardano/Wallet/Client/Http.hs
parsonsmatt: Depending on the time scale, we can possibly just cherry pick those instances for the implementation PR.

I suspect that we'll need more fancy magic like that in the implementation. I can tackle those bits and/or pair with someone to help implementing them if desired.

wallet-new/src/Cardano/Wallet/Client.hs
parsonsmatt: I have used this design before and it is *very* easy to swap out implementations without requiring a ton of `mtl` boilerplate -- passing values is easier than wrangling type classes sometimes :D

wallet-new/src/Cardano/Wallet/Client.hs
adinapoli-iohk: Although it has the word `ana` in them, maybe your brain was thinking "anamorphism" :trollface:

wallet-new/src/Cardano/Wallet/Client.hs
adinapoli-iohk: oh, TIL regarding the runtime/compile error, thanks!

> Strict record fields = no space leaks, in my experience :)

Yea, pretty much my experience as well. There might be good reasons (sometimes) to not "bang" fields in records, but it's 90% what you want, in practice, at least if your record has very simply "primitive" types.

PR #2625, opened 2018-03-06
[CBR-23] Wallet bench, merge to latest develop.

denisshevchenko:
This is wallet bench solution, prepared to merge to the latest `develop`.


denisshevchenko:
I didn't change any functionality, but I plan to test all benchmarks, just in case.

wallet-new/bench/Bench/Cardano/Wallet/Config/Endpoints.hs
```diff
@@ -0,0 +1,88 @@
...
+            <*> r .! 5
+        | otherwise = mzero
+
+-- | Extracts configuration for a single endpoint's benchmark,
+-- if it is presented in the complete configuration.
+extractEndpointConfigFor
+    :: BenchEndpoint
+    -> CompleteConfig
+    -> Maybe EndpointConfig
+extractEndpointConfigFor bench CompleteConfig {..} = find (byName bench) endpointsConfig
```
ksaric: Is it possible that the `CompleteConfig` contains multiple `EndpointConfig`? I didn't see any validation for the uniqueness of `EndpointConfig`.

wallet-new/bench/Bench/Cardano/Wallet/Random.hs
```diff
@@ -0,0 +1,55 @@
+-- | Functions for random values.
+
+module Bench.Cardano.Wallet.Random
```
ksaric: https://hackage.haskell.org/package/random-extras-0.2/docs/Control-Monad-Random-Extras.html

wallet-new/bench/report/pdf/template.tex
```diff
@@ -0,0 +1,166 @@
+% Строим книгу.
+\documentclass[11pt,oneside]{book}
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% Подгружаем необходимые пакеты
```
ksaric: Is it possible to have comments in English? Or just remove them?

wallet-new/bench/Client/Cardano/Wallet/Web/Analyze.hs
```diff
@@ -0,0 +1,38 @@
...
+                                                        ResponseReport (..))
+
+-- | Sometimes we don't want to analyze response from the server.
+analyzeResponseIfNeeded
+    :: BenchEndpoint
+    -> CompleteConfig
+    -> ResponseReport
+    -> IO ()
+analyzeResponseIfNeeded endpoint
+                        conf@CompleteConfig {..}
```
ksaric: Maybe format this better?

wallet-new/bench/Client/Cardano/Wallet/Web/Endpoint/GetHistory.hs
```diff
@@ -0,0 +1,50 @@
...
+                                                    Wallet (..), WalletAccount (..),
+                                                    WalletsConfig (..), Response,
+                                                    ResponseReport (..))
+import           Bench.Cardano.Wallet.Random       (pickRandomElementFrom)
+
+import           Pos.Wallet.Web.ClientTypes        (Addr, CHash (..), CId (..), CTx (..))
+
+-- | Run 'GetHistory' client. As a result we will get
+-- a list of transactions and size of a full history.
+getHistoryIO :: CompleteConfig -> IO ()
```
ksaric: Generally, it seems that there is an interface hidden in all these client calls. Not saying it's necessary, just that it's there.
```
class ClientCall ... where
    callEndpoint :: CompleteConfig -> IO ()  
    analyze :: Needs to be parametrized
```
These individual modules seem kind of... verbose? Not really sure, but something feels off :man_shrugging:

wallet-new/bench/Client/Cardano/Wallet/Web/Run.hs
```diff
@@ -0,0 +1,60 @@
...
+    -> ClientM a
+    -> IO (Either Text a)
+runEndpointClient CompleteConfig {..} realClient = do
+    manager <- makeClientManager tlsPubCert tlsPrivKey
+    runClientM realClient (ClientEnv manager nodeURL) >>= \case
+        Left problem   -> return . Left  $ toText (show problem :: String)
+        Right response -> return . Right $ response
+  where
+    nodeURL  = BaseUrl Https nodeHost nodePort ""
+    nodeHost = "localhost"
```
ksaric: Configurable?

wallet-new/bench/Bench/Cardano/Wallet/Config/Endpoints.hs
denisshevchenko: Yes, theoretically it's possible. But it's not problem: if we have two `GetWallets` benchmarks - they will be launched one after other, or at the same time in different threads. :-)

wallet-new/bench/Client/Cardano/Wallet/Web/Run.hs
denisshevchenko: I don't know if it should be configurable. Benchmarking is always running on localhost, right?

adinapoli-iohk:
From a quick skim looks good, thanks @denisshevchenko !

Before we pull the trigger, I would like to engage two more folks into the review:

- @gromakovsky , as he might have something to say as this PR touches some bits of the `lib` code (although just in "append" mode, we are not really modifying anything);

- @nc6 (mostly to put this on his radar) as he's also working on merging the remedial work from `master` to `develop` and this piece of work will surely come handy once the merge is complete.

A note on gauge vs criterion: In the long run we should decide if we want to switch to one or the other, as having two different benchmarking tools in the dependency stack feels weird (I know I was the guy who proposed _gauge_ in the first place 😁 ). For now I don't think it matters too much as the dependencies of gauge should really be a subset of the ones of _criterion_, but yet it would be nice if we could move everything to gauge if it's lighter and give us the same features of criterion.

denisshevchenko:
> we should decide if we want to switch to one or the other, as having two different benchmarking tools in the dependency stack feels weird

Agree. I think we should choose `gauge` and use it in all our benchmarks.

wallet-new/bench/Client/Cardano/Wallet/Web/Run.hs
ksaric: What about the port? Should that be configurable?

wallet-new/bench/Bench/Cardano/Wallet/Config/Endpoints.hs
ksaric: I have to admit I don't understand. What will be the result of calling `extractEndpointConfigFor` then? Will it call the same `GetWallets` benchmark twice since it's searching by name?

wallet-new/bench/Client/Cardano/Wallet/Web/Endpoint/NewPayment.hs
```diff
@@ -0,0 +1,61 @@
...
+import           Pos.Core                          (mkCoin)
+
+-- | Run 'NewPayment' client. As a result
+-- we will get a newly created transaction.
+newPaymentIO :: CompleteConfig -> IO ()
+newPaymentIO conf@CompleteConfig {..} = do
+    let passPhrase = Nothing
+        policy     = OptimizeForSecurity
+        minAmount  = 300000 :: Word64 -- In Lovelaces
+        maxAmount  = 900000 :: Word64 -- In Lovelaces
```
akegalj: My understanding is that `newPayment` benchmark will include both:
 * successfull payments
 * unsuccessfull payments because we are spending money that we don't have in our account

I believe this might give us a false results for some edge cases. If account has sufficient money - everything will be fine though (I mean - inpact of unsuccessfull payments will be small). 

So my question is: if benchmark is trying to send money from an account that doesn't have money - will this result be included or excluded from `newPayment` end results? (ps, maybe it doesn't matter at all - if average time for successfull payment is similar to average time of unsuccessful sending - but I am not sure if that's true.)

denisshevchenko: > So my question is: if benchmark is trying to send money from an account that doesn't have money - will this result be included or excluded from newPayment end results? 

Yes, it will be included in the result, because it's just IO-action, `gauge` knows nothing about successful or unsuccessful payments. ;-)

Of course, by default it's assumed that we _have_ money we want to send, because we want to measure correct full payment.

> (ps, maybe it doesn't matter at all - if average time for successfull payment is similar to average time of unsuccessful sending - but I am not sure if that's true.)

No, average time for successful payment takes more time than unsuccessful one.

lib/src/Pos/Util/Servant.hs
```diff
@@ -280,6 +282,25 @@ instance HasSwagger (apiType a :> res) =>
     HasSwagger (WithDefaultApiArg apiType a :> res) where
     toSwagger _ = toSwagger (Proxy @(apiType a :> res))
 
+-------------------------------------------------------------------------
+-- HasClient instances we need for benchmarking.
```
gromakovsky: Just a random thought: if they are needed only for wallet benchmarking, maybe it's better to move them to benchmarks or at least to `wallet-new` library? This way `cardano-sl` library will have less dependencies.
Not sure what's the right place for these instances, to be honest :thinking:

denisshevchenko: Yes, these instances are for benchmarking only. But we (@Martoon-00 and me) thought that if `HasServer` is here, so `HasClient` can be here as well... :-)

PR #2630, opened 2018-03-06
[CSL-2339] Get the Swagger/Aeson tests working again

parsonsmatt:
This PR gets the Swagger/Aeson tests working again. The `ValidJSON` type was causing it to miss the tests, and there are some regressions.

### DIFF
--- a/wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
@@ -363,9 +363,6 @@ instance ToSchema AddressValidity where
 instance ToSchema (V1 Core.Address) where
   declareNamedSchema = annotate fromExampleJSON
 
-instance ToSchema WalletId where
-  declareNamedSchema = annotate fromExampleJSON
-
 instance ToSchema Metadata where
   declareNamedSchema = annotate fromExampleJSON
 
@@ -378,9 +375,6 @@ instance ToSchema NewWallet where
             { fieldLabelModifier =
                 over (ix 0) Char.toLower . drop 6 {- length "newwal" -}
             }
-
-instance ToSchema NewAddress where
-  declareNamedSchema = annotate fromExampleJSON
 
 instance ToSchema WalletUpdate where
   declareNamedSchema = annotate fromExampleJSON
--- a/wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
@@ -27,7 +27,8 @@ import           Pos.Util.CompileInfo (HasCompileInfo, compileInfo, ctiGitRevisi
 import           Pos.Wallet.Web.Methods.Misc (WalletStateSnapshot)
 import           Pos.Wallet.Web.Swagger.Instances.Schema ()
 
-import           Control.Lens ((?~))
+import qualified Data.Char as Char
+import           Control.Lens ((?~), ix)
 import           Data.Aeson (ToJSON (..), Value (Number, Object))
 import           Data.Aeson.Encode.Pretty
 import qualified Data.HashMap.Strict as HM
@@ -372,7 +373,11 @@ instance ToSchema Wallet where
   declareNamedSchema = annotate fromExampleJSON
 
 instance ToSchema NewWallet where
-  declareNamedSchema = annotate fromExampleJSON
+    declareNamedSchema =
+        genericDeclareNamedSchema defaultSchemaOptions
+            { fieldLabelModifier =
+                over (ix 0) Char.toLower . drop 6 {- length "newwal" -}
+            }
 
 instance ToSchema NewAddress where
   declareNamedSchema = annotate fromExampleJSON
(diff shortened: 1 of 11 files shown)


parsonsmatt:
`hlint` is failing because of the `pattern` function, which is *also* a keyword, but only if a language extension is enabled (it isn't).

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -231,21 +231,11 @@ The number of entries to display for each page. The minimum is **1**, whereas th
...
-    & type_ .~ SwaggerInteger
-    & default_ ?~ (Number $ fromIntegral defaultPerPageEntries)
-    & minimum_ ?~ 1
-    & maximum_ ?~ (fromIntegral maxPerPageEntries)
-
-instance ToParamSchema Page where
-  toParamSchema _ = mempty
-    & type_ .~ SwaggerInteger
-    & default_ ?~ (Number 1) -- Always show the first page by default.
-    & minimum_ ?~ 1
```
parsonsmatt: I might be knocking out a Schelling fence here, but I can't see a reason for these to be orphaned when the library is in the same package.

wallet-new/src/Cardano/Wallet/API/Response.hs
```diff
@@ -64,6 +75,22 @@ instance Arbitrary a => Arbitrary (WalletResponse a) where
...
+        respRef <- declareSchemaRef (Proxy @ResponseStatus)
+        metaRef <- declareSchemaRef (Proxy @Metadata)
+        pure $ NamedSchema (Just $ "WalletResponse<" <> tyName <> ">") $ mempty
+            & type_ .~ SwaggerObject
+            & required .~ ["data", "status", "meta"]
+            & properties .~
+                [ ("data", aRef)
+                , ("status", respRef)
+                , ("meta", metaRef)
+                ]
```
parsonsmatt: This is a trick I've learned at a previous place. Response envelopes get *weird* if you don't name the things differently, since they get repeated, and Swagger doesn't support parameterized templates.

KtorZ:
Well, thanks for pulling me in @adinapoli-iohk. This PR contains pretty much the work I started today (keeping however the various schema as Orphan instances in the swagger module, though I admit to favour @parsonsmatt approach to keep the schema close to the underlying data-structures). 

Descriptions are still missing from the schemas and I have already reorganized a bit tags and groups, but this is definitely very helpful to move forward. 

How do we proceed from here? I can rebase my work on top of this PR and continue working from there, reviewing at the same time the work achieved so far.

parsonsmatt:
This PR is only about fixing the tests, which it does -- I think we should have this merged before we start writing potentially erroneous documentation :open_mouth: 

Validating that the Swagger spec is actually, well, valid, is probably something I should do :smile: I didn't know that the library would generate invalid specs!

KtorZ:
Sure thing ^^, but I was planning on doing pretty much the same thing at first (defining ToSchema instances for each datatype present in the API). 

The library can't actually generate invalid specs, or more exactly, they're always syntactically correct. Nevertheless, because we have on the one hand, instances describing the types of our API (ToSchema from Swagger) and on the other hand, actual serialization / unserialization functions (ToJSON and FromJSON from Aeson). There's, a priori, nothing proving that what we return actually corresponds to what we say we return (actual json representation of our types vs specifications). 

Hence the need for QuickCheck ^.^

Anyway, thanks for doing this bit of work, very helpful for me.

KtorZ:
The specs are generated on start-up. So far, I've been following the instructions in the README(s) to do so, thus:

```
stack exec cardano-new -- --topology=wallet-new/topology-examples/testnet.yaml --configuration-key mainnet_staging_short_epoch_full
```

seems enough to generate new swagger file every time :+1:

adinapoli-iohk:
> seems enough to generate new swagger file every time

Which is is puzzling, actually 🤔 By design those should be generated only when the `--wallet-debug` option is passed (which runs the wallet in "debug mode"). I'm wondering if we got a regression here.

KtorZ:
My bad indeed. Or least, I can't tell whether it is generated without that flag. Oversight from my side.

adinapoli-iohk:
> My bad indeed. Or least, I can't tell whether it is generated without that flag. Oversight from my side.

Not at all -- indeed it was a regression which slipped code review, alas 😅

KtorZ:
@parsonsmatt Hey, I've just remarked that some schemas are actually refer to as simple `objects`. For instance:

```json
"BlockchainHeight": {
    "required": [
        "quantity"
    ],
    "type": "object"
},
```
whereas the code seems to define details of those properties:

```hs
& properties .~ (mempty
  & ix "quantity" .~ (Inline $ mempty -- ...
  & ix "unit" .~ (Inline $ mempty -- ...
```

This issue comes from the usage of `ix` which provides a lens to access properties and modify properties on the underlying hashmap only if they already exist. Hence the empty map as a result.
I've switch to use `at` instead that will also set an element if the associated key isn't present: 

```hs
& properties .~ (mempty
  & at "quantity" ?~ (Inline $ mempty -- ...
  & at "unit" ?~ (Inline $ mempty -- ...
```

I've made an isolated commit on my branch to fix that: 356ca58 as well as naming of some of those schemas usually named `V1 Something`. Since the name is used as URL path to reference some of those schemas, it leads to not so pretty %20 encoding of the space, so I've just removed it :+1:

parsonsmatt:
Ahahaha! Thanks for catching that :smile: That trips me up every time I spend more than a week away from `lens`

PR #2633, opened 2018-03-07
[CSL-2353] Merge `release/1.1.0` to `develop`

hithroc:


block/src/Pos/Block/Network/Retrieval.hs
```diff
@@ -92,25 +91,21 @@ retrievalWorkerImpl diffusion =
                     pure (handleRecoveryWithHandler nodeId rHeader)
         () <- thingToDoNext
```
gromakovsky: In #2619 I see a comment before this line which you somehow lost :pensive:

PR #2634, opened 2018-03-07
[DEVOPS-712] removing --web parameter from shell scripts and default.nix

disassembler:
Building `connectScripts.mainnetWallet` using `nix-build` results in a binary that won't start because of invalid parameter `--web`. This pull request removes it from the default.nix and all other places I saw it being used with a `git grep "\-\-web" command.

### DIFF
--- a/scripts/common-functions.sh
+++ b/scripts/common-functions.sh
@@ -208,11 +208,6 @@ function node_cmd {
   if [[ "$REPORT_SERVER" != "" ]]; then
     report_server=" --report-server $REPORT_SERVER "
   fi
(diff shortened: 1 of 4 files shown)


domenkozar:
I hope @adinapoli-iohk can sign this off and merge :)

gromakovsky:
@adinapoli-iohk `--web` is an old flag which enables web API from `Pos.Web.Api` (in old wallet). It was added in Dec 2016 and was occasionally used (mostly `utxo` and `confirmed_proposals` options).
Requirements changed 100500 times since then, so now this `Pos.Web.Api` looks weird and probably should be revised.

PR #2636, opened 2018-03-07
[CSL-2346][CSL-2318] Refactoring of Ntp

coot:
~~Things left to do:~~
* ~~simplify `NtpMonad` - we should be able to leave only `WithLogger m`, so we can explicitly run the ntp client in `Production` monad.~~
* ~~`--no-ntp-checks` will disable the ntp check on startup, but will not disable the wallet endpoint.  @avieth, I don't think this is necessary for the purpose you opened the original issue so it should be fine, otherwise we can add an option to `NtpConfiguration` (previously `InfraConfiguration`)~~


avieth:
> simplify NtpMonad - we should be able to leave only WithLogger m, so we can explicitly run the ntp client in Production monad.

Might as well make it `IO` then.

> --no-ntp-checks will disable the ntp check on startup, but will not disable the wallet endpoint. @avieth, I don't think this is necessary for the purpose you opened the original issue so it should be fine, otherwise we can add an option to NtpConfiguration (previously InfraConfiguration)

My understanding was that it shouldn't disable the wallet endpoint, it should just imply that the system will always give "unavailable" as the answer.

auxx/src/Mode.hs
```diff
@@ -147,7 +147,7 @@ instance HasSlogGState AuxxContext where
 instance HasJsonLogConfig AuxxContext where
     jsonLogConfig = acRealModeContext_L . jsonLogConfig
 
-instance (HasConfiguration, HasInfraConfiguration, MonadSlotsData ctx AuxxMode)
+instance (HasConfiguration, HasNtpConfiguration, MonadSlotsData ctx AuxxMode)
```
avieth: Do you think it would be feasible to stop using the `HasNtpConfiguration` style and just take a record instead? Typically the problem with doing such a change is that the configuration is used in instances, but if we could make it not needed in the instances...

lib/src/Pos/Launcher/Scenario.hs
```diff
@@ -101,7 +105,18 @@ runNode' NodeResources {..} workers' plugins' = ActionSpec $ \diffusion -> ntpCh
...
+    ntpCheck False action = action
+    ntpCheck True action = do
+        tvar <- newTVarIO Nothing
+        let settings = ntpSettings (ncNtpConfig nrContext) tvar
+        withAsync
+            (do
+                st <- atomically $ readTVar tvar >>= \case
+                    Nothing -> retry
+                    Just st -> return st
+                onNtpStatusLogWarning st)
```
avieth: This is here to log in case NTP takes too long?

networking/src/NTP/Client.hs
```diff
@@ -39,16 +38,19 @@ import           System.Wlog (LoggerName, WithLogger, logDebug, logError, logInf
...
 import           Mockable.Class (Mockable)
-import           Mockable.Concurrent (Async, Concurrently, Delay, concurrently, forConcurrently,
+import           Mockable (Async, Concurrently, CurrentTime, Delay, concurrently, currentTime, forConcurrently,
                                       timeout, withAsync)
 import           NTP.Packet (NtpPacket (..), evalClockOffset, mkCliNtpPacket, ntpPacketSize)
 import           NTP.Util (createAndBindSock, resolveNtpHost, selectIPv4, selectIPv6,
                            udpLocalAddresses, withSocketsDoLifted)
 
-data NtpClientSettings m = NtpClientSettings
+data NtpStatus = NtpSyncOk | NtpDesync Microsecond
```
avieth: This is I guess where we'd want to add `NtpSyncUnavailable`, right?

coot:
| Might as well make it IO then.

Is there a way to run the logger in IO? I am not familiar with the `log-warper` serokell's logging library.

adinapoli-iohk:
> Is there a way to run the logger in IO? I am not familiar with the log-warper serokell's logging library.

@coot If this is for some throw-away testing code, I think the easiest way is to define orphan instances for `CanLog`. You have to essentially defined two functions: `askLoggerName`, which can really be anything you like, and `modifyLoggerName`, which can be something as simple as a "noop" action, to get started quickly.

lib/src/Pos/Launcher/Scenario.hs
coot: No, rather this will just log the `NtpStatus`, i.e. weather we are within the threshold configured in `NtpClientSettings` (the `ntpTimeDifferenceWarnThreshold` which is copied from `NtpConfiguration`). This run asynchronously, as it was before.

networking/src/NTP/Client.hs
coot: Yes indeed, but we might not need it after all.  I think you agreed that we only need to disable/bypass the startup ntp check rather than disable wallet server to query for ntp status, hence there is no need for additional status message.

auxx/src/Mode.hs
coot: I think that's feasible :). Previously the `HasNtpConfiguration` was used to provide the configuration for ntp slotting that we removed.

coot: @avieth it is possible to eliminate it from `cardano-sl`, but then there is `HasConfigurations` which is using it all over the place in the wallet and explorer.
Furthermore, `withConfigurations` is the function that reads the yaml config, including `ntpConfiguration` part, so the ntp config is coupled with all other configs.

I advanced to the point where ntp configuration is passed explicitly in `cardano-sl` and the `cardano-sl-explorer` compiles, but still the `HasNtpConfiguration` is used in `HasConfigurations` (beacuse of `withConfigurations`).  This means that the functions `runNodeReal`, `bracketNodeResources` (and its friends: `allocateNodeResources`, `allocateNodeContext`), get `NtpConfiguration` explicitly.  I was able to remove the `HasNtpConfiguration` constraint from various instances in `cardano-sl` and `cardano-sl-networking`.  The only place `HasNtpConfiguration` is really needed is in the wallet.

To conclude: I don't think we can simply remove `HasNtpConfiguration` without refactoring `withConfigurations` which I didn't want to touch for now.  The question I have, should I intrude into wallet so that the ntp configuration is passed explicitly?

Another suggestion I have is that we probably should have `NtpNotConfigured` constructor for `NtpConfiguration` (making `NtpConfiguration ~ Maybe NtpConfiguration`).  Simply because ntp is not required in all cases (mostly experiments/tests or the explorer app).  This will force us to add `NtpSyncUnavailable` to `NtpStatus` (to keep all the functions total).

coot:
@adinapoli-iohk unfortunatelly its not, ntp client is used in the wallet. Otherwise I think we will need to keep also `MonadIO m` and `MonadBaseControl IO m` along with `CanLog`.

auxx/src/Mode.hs
coot: I pushed the changes in a separate branch (last commit): https://github.com/input-output-hk/cardano-sl/tree/coot/csl-2346-remove-HasNtpConfiguration (I updated the `wallet` and `new-wallet`, with some more `HasNtpConfiguration` so that it compiles :/)

avieth:
> @coot If this is for some throw-away testing code, I think the easiest way is to define orphan instances for CanLog. You have to essentially defined two functions: askLoggerName, which can really be anything you like, and modifyLoggerName, which can be something as simple as a "noop" action, to get started quickly.

You could use `usingLoggerName :: IO t -> LoggerNameBox IO t` around your logging functions. This way, you don't induce a `CanLog` or `HasLoggerName` constraint. You also don't inherit higher logger names, but that seems a small price to pay.

auxx/src/Mode.hs
avieth: Relevant refactoring here https://github.com/input-output-hk/cardano-sl/pull/2590
That patch removes all `HasXConfiguration` from the diffusion layer, but of course leaves it in place for other areas because it's heavily entrenched, particularly the dependences of slotting definitions on the `k` parameter from `HasProtocolConstants`, which determines the number of slots in an epoch and thereby makes it possible to get `Num`, `Enum`, `Bounded` instances for local slot indices.

Anyway, you're right it would be a bit of a job to fix this now, so I've made a ticket and we can do it later https://iohk.myjetbrains.com/youtrack/issue/CSL-2370

infra/Pos/Infra/Configuration.hs
```diff
@@ -1,10 +1,9 @@
 {-# LANGUAGE Rank2Types #-}
 module Pos.Infra.Configuration
```
avieth: Should move this file to `infra/Pos/Ntp/Configuration.hs`

infra/Pos/Slotting/Impl/Simple.hs
```diff
@@ -32,7 +34,14 @@ type SimpleSlottingMode ctx m
     = ( Mockable CurrentTime m
       , MonadSlotsData ctx m
       , MonadIO m
-      , HasConfiguration)
+      , HasConfiguration
```
avieth: This seems dubious. I wonder what it actually needs... probably only `ProtocolConstants`, because slotting stuff depends upon it. OK to leave it for now, but after https://github.com/input-output-hk/cardano-sl/pull/2590 is merged it can probably be loosened to `HasProtocolConstants`.

infra/Pos/Slotting/Impl/Simple.hs
```diff
@@ -42,49 +51,49 @@ data SimpleSlottingState = SimpleSlottingState
     { _sssLastSlot :: !SlotId
     }
 
-type SimpleSlottingVar = TVar SimpleSlottingState
+type SimpleSlottingStateVar = TVar SimpleSlottingState
 
-mkSimpleSlottingVar :: (MonadIO m, HasConfiguration) => m SimpleSlottingVar
-mkSimpleSlottingVar = atomically $ newTVar $ SimpleSlottingState $ unflattenSlotId 0
+mkSimpleSlottingStateVar :: (MonadIO m, HasConfiguration) => m SimpleSlottingStateVar
```
avieth: Right I see, indeed the `HasConfiguration` is needed because `unflattenSlotId` uses the `k` parameter from `protocolConstants`. No action required, I'm just thinking aloud.

infra/Pos/Slotting/Impl/Simple.hs
```diff
@@ -42,49 +51,49 @@ data SimpleSlottingState = SimpleSlottingState
...
 
 getCurrentSlotInaccurateSimple
-    :: (SimpleSlottingMode ctx m)
-    => SimpleSlottingVar
-    -> m SlotId
-getCurrentSlotInaccurateSimple var =
-    getCurrentSlotSimple var >>= \case
+    :: (MonadSimpleSlotting ctx m)
+    => m SlotId
+getCurrentSlotInaccurateSimple =
```
avieth: These previous 3 functions have all changed to take their slotting var from a typeclass rather than as a parameter. Any specific advantage to this?

infra/cardano-sl-infra.cabal
```diff
@@ -39,11 +39,8 @@ library
                         -- Slotting
                         Pos.Slotting
                         Pos.Slotting.Class
-                        Pos.Slotting.Configuration
                         Pos.Slotting.Impl
-                        Pos.Slotting.Impl.Ntp
                         Pos.Slotting.Impl.Simple
-                        Pos.Slotting.Impl.Sum
```
avieth: Happy to see these deletions.

lib/src/Pos/Launcher/Scenario.hs
avieth: Oh this was being done before this patch?

networking/src/NTP/Client.hs
avieth: I can't remember. I thought the idea was to either disable it completely (in which case UI client should get a `Unavailable` variant) or to enable it (in which case checks run in the background and fill the status var).

networking/src/NTP/Client.hs
```diff
@@ -39,16 +38,19 @@ import           System.Wlog (LoggerName, WithLogger, logDebug, logError, logInf
...
+    deriving (Eq, Show)
+
+data NtpClientSettings = NtpClientSettings
     { ntpServers         :: [String]
       -- ^ list of servers addresses
-    , ntpHandler         :: (Microsecond, Microsecond) -> m ()
+    , ntpStatus          :: TVar (Maybe NtpStatus)
```
avieth: Ah, well I suppose `Nothing` basically means `NtpStatusUnavailable`.

networking/src/NTP/Client.hs
```diff
@@ -39,16 +38,19 @@ import           System.Wlog (LoggerName, WithLogger, logDebug, logError, logInf
...
                                       timeout, withAsync)
 import           NTP.Packet (NtpPacket (..), evalClockOffset, mkCliNtpPacket, ntpPacketSize)
 import           NTP.Util (createAndBindSock, resolveNtpHost, selectIPv4, selectIPv6,
                            udpLocalAddresses, withSocketsDoLifted)
 
-data NtpClientSettings m = NtpClientSettings
+data NtpStatus = NtpSyncOk | NtpDesync Microsecond
+    deriving (Eq, Show)
+
+data NtpClientSettings = NtpClientSettings
```
avieth: What is the difference between `NtpClientSettings` and `NtpConfiguration`? Seems weird that we have a `TVar` in a "settings" record.

networking/src/NTP/Client.hs
```diff
@@ -58,40 +60,26 @@ data NtpClientSettings m = NtpClientSettings
...
+      -- ^ Maxium tolerable difference between NTP time and locak time.
     }
 
-hoistNtpClientSettings
-    :: (m () -> n ()) -> NtpClientSettings m -> NtpClientSettings n
-hoistNtpClientSettings f settings =
-    settings { ntpHandler = f . ntpHandler settings }
-
-data NtpClient m = NtpClient
+data NtpClient = NtpClient
```
avieth: The status `TVar` should go in here I think, rather than in the settings.

lib/src/Pos/Launcher/Scenario.hs
coot: Yes

networking/src/NTP/Client.hs
coot: As I just explained in slack (so just for completeness sake), there is no background ntp server running, every time wallet asks for ntp status, the backend spawns a new ntp client.  Thus there is not need to have `NtpSyncUnavailable`.

But I am open for changing the ntp architecture, just not in this PR :)

infra/Pos/Infra/Configuration.hs
coot: Let's also move `infra/Pos/NtpCheck.hs` to `infra/Pos/Ntp/NtpCheck.hs`.

coot: Or even better, just `Pos.Ntp.Check`

networking/src/NTP/Client.hs
coot: Not really, I use `Nothing` to wait until ntp client fills it with a value. If we had `NtpStatusUnavailable` it would be filled with `Just NtpStatusUnavailable` at the start.

infra/Pos/Slotting/Impl/Simple.hs
coot: That's true, but it now plays the role of `getCurrentSlotSum` where the argument was read from a typeclass.  I haven't figure out if we can change this. But since we want to remove `MonadSlots` this requirement should go away as well.

infra/cardano-sl-infra.cabal
coot: Me too :)

networking/src/Ntp/Client.hs
```diff
@@ -7,114 +8,112 @@
...
+-- |
+-- Handle results.  It is either when all ntp servers responded or when
+-- `ntpResponseTimeout` has passed since the request where send.  If none of the servers responded `ncState`
+handleCollectedResponses :: NtpClient -> IO ()
 handleCollectedResponses cli = do
-    mres <- liftIO $ readTVarIO (ncState cli)
+    mres <- readTVarIO (ncState cli)
     let selection = ntpMeanSelection (ncSettings cli)
-        handler   = ntpHandler (ncSettings cli)
     case mres of
```
coot: Since we now have `NtpSyncUnavailable` we can use it when none of the server responded.

parsonsmatt:
Is there anything holding this PR up?

avieth:
> Is there anything holding this PR up?

@coot Is the PR description TODO list still actual?

avieth:
> > @coot Is the PR description TODO list still actual?
>
> No it's not, the PR is ready. There are some CI problems though.

@coot Did you mean to edit my comment?

The fact that this PR has to touch so many files is alarming. The NTP feature really ought to be isolated. But hey that's not your fault. I have one more request: could you look into removing the `HasNtpConfiguration` constraints? I suspect it might be pretty easy, since I don't think any NTP stuff is used in typeclass instances.

coot:
Ups, no! Sorry for that!

Yes, lets remove `HasNtpConfiguration` (and touch even more files :/).

It seems I broke some tests in `cardano-sl-generator`. There are lots of errors with `block is from slot which hasn't happened yet;`.

avieth:
> Yes, lets remove HasNtpConfiguration (and touch even more files :/).

Don't worry about touching more files, I'm not squeamish about that.

networking/src/NTP/Client.hs
avieth: Ok, so `Nothing` means "pending". Perhaps we want `Maybe (TVar NtpStatus)` instead? Where `NtpStatusPending` is one of the status constructors?

Then we'd define

```Haskell
currentNtpStatus :: Maybe (TVar NtpStatus) -> UINtpStatus
currentNtpStatus Nothing = UINtpUnavailable
currentNtpStatus (Just tvar) = do
    st <- atomically (readTVar tvar)
    case st of
        NtpDrift (updatedAt :: Time) (drift :: Int) -> UINtpDrift drift
        NtpPending -> UINtpPending
```

What do you think?

coot: @avieth I broke the branch with a rebase, and you commented before I restored it.  We have `TVar NtpStatus`, but we  adding `NtpPending` to the `NtpStatus` is a good idea anyway.  Then we can use `NtpStatusUnavailable` when no respond arrived from any of the ntp servers (within the timeout which is now set to `30sec` in `lib/configuration.yaml`). Now we use a preconfigured threshold to deduce `NtpSyncOk` or `NtpDesync Milliseconds`, which is set to `5sec`. It just should be `<< slotDuration` - I think it is better to compute the threshold depending on slot duration and then it's better to use `NtpSyncOK` rather than expose the drift in `Milliseconds` (the wallet might not now the current slot duration).

coot:
@avieth it's ready for a review, though here are a few things which still needs be done:

- [ ] research the syscall that @karknu mentioned
- [ ] use `withoutNtpClient` in cardano-sl-tools where we don't need to use a ntp client
- [ ] resolve conflicts with develop

coot:
@avieth I think this is ready for a final review. If we want to use `adjtimex` sys call, this can go in a new PR.

coot:
Note it's based on top of `develop` branch, mostly because the wallet guys are awaiting for the changes. But we should pull this into `networking/develop` too. I am not sure how it intersects with the other changes you did to get rid of `Has*Configuration` constraints.

avieth:
> @avieth I think this is ready for a final review. If we want to use adjtimex sys call, this can go in a new PR.

Ok.

> Note it's based on top of develop branch, mostly because the wallet guys are awaiting for the changes. But we should pull this into networking/develop too. I am not sure how it intersects with the other changes you did to get rid of Has*Configuration constraints.

I'll give it another pass this afternoon. I'll take care of merging to `networking/develop`. I doubt it will be so bad.

generator/src/Test/Pos/Block/Logic/Mode.hs
```diff
@@ -173,10 +175,11 @@ withTestParams TestParams {..} = withGenesisSpec _tpStartTime coreConfiguration
...
 data TestInitModeContext = TestInitModeContext
-    { timcDBPureVar   :: DBPureVar
-    , timcSlottingVar :: TVar SlottingData
-    , timcSystemStart :: !Timestamp
-    , timcLrcContext  :: LrcContext
+    { timcDBPureVar        :: DBPureVar
+    , timcSlottingVar      :: TVar SlottingData
+    , timcSlottingStateVar :: SimpleSlottingStateVar
+    , timcSystemStart      :: !Timestamp
+    , timcLrcContext       :: LrcContext
```
parsonsmatt: Is there a reason why the timestamp is the only strict field in the record?

infra/Pos/Infra/Configuration.hs
```diff
@@ -1,59 +0,0 @@
-{-# LANGUAGE Rank2Types #-}
```
parsonsmatt: <3 to see files deleted <3

infra/Pos/Slotting/Impl/Simple.hs
```diff
@@ -42,49 +53,68 @@ data SimpleSlottingState = SimpleSlottingState
...
 
-getCurrentSlotSimple
+getCurrentSlotSimple'
     :: (SimpleSlottingMode ctx m)
-    => SimpleSlottingVar
+    => SimpleSlottingStateVar
     -> m (Maybe SlotId)
-getCurrentSlotSimple var = traverse (updateLastSlot var) =<< (currentTimeSlottingSimple >>= slotFromTimestamp)
+getCurrentSlotSimple' var = do
+    -- var <- view (lensOf @SimpleSlottingStateVar)
```
parsonsmatt: Comment to delete?

infra/Pos/Slotting/Impl/Simple.hs
```diff
@@ -42,49 +53,68 @@ data SimpleSlottingState = SimpleSlottingState
...
+    traverse (updateLastSlot var) =<< (currentTimeSlottingSimple >>= slotFromTimestamp)
```
parsonsmatt: I would prefer for everything to read in a consistent direction -- either

```haskell
traverse (updateLastSlot var) =<< slotFromTimestamp =<< currentTimeSlottingSimple

--or,

slotFromTimestamp
    >>= currentTimeSlottingSimple
    >>= traverse (updateLastSlot var)
```

Switching directions mid-line makes it hard to read.

infra/Pos/Slotting/Impl/Simple.hs
coot: Yes!

infra/Pos/Slotting/Impl/Simple.hs
coot: Yes, I shouldn't copy paste it. It's indeed error prone (there is one in your comment with wrong order, which proves the statement :) )

generator/src/Test/Pos/Block/Logic/Mode.hs
coot: I only added `timcSlottingStateVar`, the other lines were only modified to comply with styling rules. Looking at [Test.Pos.Block.Logic.Mode](https://github.com/input-output-hk/cardano-sl/blob/coot/csl-2346/generator/src/Test/Pos/Block/Logic/Mode.hs#L249) only the `timcSlottingVar` and `timcLrcContext` needs to be lazy.

PR #2639, opened 2018-03-08
[CSL-2336] Catch V1 endpoints errors

sectore:
Part 1 of https://iohk.myjetbrains.com/youtrack/issue/CSL-2336  to “catch all” errors (except `ServantErr`) at top level (`Warp`)

Example (using `httpie`):
```
http :8090/api/development/fail     
 
HTTP/1.0 400 Bad Request
Content-Type: application/json
Date: Thu, 08 Mar 2018 13:35:22 GMT
Server: Warp/3.2.13

{
    "diagnostic": {
        "msg": "Something went wrong."
    }, 
    "message": "GenericError", 
    "status": "error"
}
```

_Note:_ Endpoint `fail` ^ has been added for debugging purposes only and is **not** accessible in production mode:
```
http --verify=no https://127.0.0.1:8090/api/development/fail

HTTP/1.1 403 Forbidden
Date: Thu, 08 Mar 2018 13:34:45 GMT
Server: Warp/3.2.13
Transfer-Encoding: chunked
```

_Note 2:_ This PR includes a fix for Swaggers scheme of `WalletStateSnapshot` (part of `CSL-2094`) :
```
Schema error at definitions['WalletResponse WalletStateSnapshot'].properties['data'].properties['wssWalletStorage'].properties['_wsReadyUpdates'].items
should be object
```
To save some time I didn't want to create a new PR for this simple thing.


wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -103,6 +110,11 @@ legacyWalletBackend WalletBackendParams {..} =
               LegacyServer.walletServer (V0.convertHandler ctx) diffusion walletRunMode
       return $ withMiddleware walletRunMode app
 
+    exceptionHandler :: SomeException -> Response
+    exceptionHandler _ =
+        responseLBS badRequest400 [(hContentType, "application/json")] .
+            encode . toJSON $ GenericError "Something went wrong."
```
parsonsmatt: `toJSON` is redundant here -- `encode :: ToJSON a => a -> LByteString` :)

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -52,6 +52,7 @@ data WalletError =
     | SomeOtherError { weFoo :: !Text, weBar :: !Int }
     | MigrationFailed { weDescription :: !Text }
     | JSONValidationFailed { weValidationError :: !Text }
+    | GenericError { weMsg :: !Text }
```
parsonsmatt: I would prefer to see `UnkownError SomeException` instead -- that would let us retain more information about the error up until the serialization boundary. I strongly dislike `Text` and friends for errors, as they are completely opaque and any other error can be serialized to them later.

wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs
```diff
@@ -28,5 +29,20 @@ instance Arbitrary CPaperVendWalletRedeem
...
+                            <*> arbitrary
+                            <*> arbitrary
+                            <*> arbitrary
+                            <*> arbitrary
+
 instance Arbitrary WalletStateSnapshot where
-    arbitrary = WalletStateSnapshot <$> pure def
+    arbitrary = do
+        v <- arbitrary
+        WalletStateSnapshot <$> pure (def {_wsReadyUpdates = [v]})
```
parsonsmatt: Nice! This fixes the swagger spec problem?

Why not generate a list of `v`s instead of just a single `v`?

sectore: > This fixes the swagger spec problem?

It seems that Swagger does not like an empty `[]` in that case, so we have to fill it up. There is a comment in [ `Example.hs`](https://github.com/input-output-hk/cardano-sl/blob/8c2c240406957c9fb688cd6a0d7bf569e2f8d48e/wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs#L111-L115) about how to deal w/ such an issue 

> Why not generate a list of vs instead of just a single v?

At the moment the only place we do use an arbitrary instance of `WalletStateSnapshot` is for generating an example for Swaggers documentation. And a single item is all what we need in that example. If we might need a "real" arbitrary instance without using [`Default` of `WalletStorage`](https://github.com/input-output-hk/cardano-sl/blob/8c2c240406957c9fb688cd6a0d7bf569e2f8d48e/wallet/src/Pos/Wallet/Web/State/Storage.hs#L228-L241) (which provides more `mempty` then any arbitrary values), then we can change it. But for now I would suggest just to add just a single item in that list of `wsReadyUpdates`.

wallet-new/server/Cardano/Wallet/Server/Plugins.hs
sectore: Good catch!

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
sectore: I see you point. However, we are using `WalletError` at _the serialization boundary_ . That means we have to serialize / deserialize `WalletError` (incl. `UnkownError SomeException`) in that case. Not least to deal with `JSend`. How would you do that? 

We have already discussed about that issue ^ in our team today. Please check https://input-output-rnd.slack.com/archives/C819S481Y/p1520592997000412 and all following comments there.

wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs
parsonsmatt: Sounds good :+1:

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
parsonsmatt: Serializing works much the same:

```haskell
Prelude> import Control.Exception
Prelude Control.Exception> :t toException (userError "oops")
toException (userError "oops") :: SomeException
Prelude Control.Exception> show (toException (userError "oops"))
"user error (oops)"
```

When we have a `SomeException` we can always show it, and it'll delegate to the `displayException` (typically `show`) function on the `Exception` class.

---

But there's more! Suppose we have a `GenericError` on our hands -- we can't do anything with it. Our hands are tied. Alas. We can print the message. If we wanted to parse it, we could, but that's a ton of work. If we have an `UnknownError`, then we can use the `fromException :: Exception e => SomeException -> Maybe e` method to test potential causes and handle them separately. This makes `WalletError` here useful at the API boundary *and also* at other points in the stack.

adinapoli-iohk: @parsonsmatt I finally had the chance of playing with the code and I can now see the lost in translation with @sectore : when we originally spoke, I was under the impression that generating a valid generic instance for `SomeException` was not possible, but turns out that [deriveGeneric](http://hackage.haskell.org/package/generics-sop-0.3.2.0/docs/Generics-SOP-TH.html#v:deriveGeneric) from `Generics.SOP.TH` can do that just as well. I think the compilation errors Jens was experiencing are completely fixable if we:

* We remove the `Eq` constraint on `WalletError` or define an orphan for `SomeException` (which would be better to avoid, in which case we might want to rewrap `SomeException` in a `newtype`)

* We remove completely `FromJSON WalletError`, as I think that originally we added this extra instance to be able to write roundtrips in the specs, but practically we should be ever in a position to parse a `WalletError` as input. At least that's a use case I cannot anticipate on top of my head.

Addressing the two above should be enough to have our cake and eat it too: I was, in fact, able to get a clean build relatively quickly. Let me know if I'm being mental.

parsonsmatt: We'd need to parse a `WalletError` in the servant-client. Alternatively we can just rely on the `ServantError` type and not have fine-grained error handling on clients.

parsonsmatt: The missing Eq and FromJSON instances are pretty big deals. I'm happy to leave this as text for now.

wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs
```diff
@@ -28,5 +29,20 @@ instance Arbitrary CPaperVendWalletRedeem
 instance Arbitrary CInitialized
 instance Arbitrary NewBatchPayment
 
+instance Arbitrary CCoin where
+    arbitrary = CCoin <$> arbitrary
+
+instance Arbitrary CUpdateInfo where
```
adinapoli-iohk: Weird, why was this `Arbitrary` instance needed as part of this PR?

wallet-new/src/Cardano/Wallet/API/Development.hs
```diff
@@ -11,3 +11,5 @@ type API
                               :> Get '[OctetStream] (WalletResponse WalletStateSnapshot)
     :<|> "secret-keys"        :> Summary "Clear wallet state and delete all the secret keys."
                               :> DeleteNoContent '[ValidJSON] NoContent
+    :<|> "fail"               :> Summary "Throw a generic error"
```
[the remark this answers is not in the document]
sectore: See PR description above: 
```Note: Endpoint fail ^ has been added for debugging purposes only and is not accessible in production mode:``` 
Why not leave that endpoint here, we will use it for sure to improve error handling of V1 endpoints in future?

File: wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs (hunk 4 above)
Comment (sectore) [reply wallet-new/src/Cardano/Wallet/Orphans/Arbitrary.hs]: It was needed to fix Swagger issues - see PR description above 
> Note 2: This PR includes a fix for Swaggers scheme of WalletStateSnapshot (part of CSL-2094) :

File: wallet-new/server/Cardano/Wallet/API/Development/LegacyHandlers.hs [hunk 6]
```diff
@@ -65,3 +65,7 @@ deleteSecretKeys :: ( V0.HasNodeConfiguration
                  -> m NoContent
 deleteSecretKeys runMode =
     developmentOnly runMode (V0.deleteAllSecretKeys >> V0.testReset >> return NoContent)
+
+throwSomething :: MonadThrow m => RunMode -> m NoContent
```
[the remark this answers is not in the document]
sectore: It's a handler of `fail` endpoint, see my comment below to similar question https://github.com/input-output-hk/cardano-sl/pull/2639#discussion_r174073368

wallet-new/src/Cardano/Wallet/API/Development.hs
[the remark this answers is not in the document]
adinapoli-iohk: > Why not leave that endpoint here, we will use it for sure to improve error handling of V1 endpoints in future?

That's not an unreasonable point actually -- I totally expect us to come back to more sophisticated error handlings in the future. Let's leave it there for now, with the caveat of removing as soon as CBR-26 completes.

sectore:
Note: `Hydra` is still failing. `Devops` team has been informed about it https://input-output-rnd.slack.com/archives/G3TLHGM18/p1520937015000142

adinapoli-iohk:
Thanks, @sectore !

PR #2641, opened 2018-03-08
[CSL-2343] Add missing logException to cardano-node

gromakovsky:
`logException` was used in old `cardano-node`, but for some reason is not used in new `cardano-node`.

### DIFF
--- a/wallet-new/server/Main.hs
+++ b/wallet-new/server/Main.hs
@@ -12,8 +12,8 @@ import           Data.Aeson.Encode.Pretty (encodePretty)
 import qualified Data.ByteString.Lazy.Char8 as BL8
 import           Data.Maybe (fromJust)
 import           Mockable (Production (..), runProduction)
-import           Pos.Communication (ActionSpec (..))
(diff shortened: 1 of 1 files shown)


wallet-new/server/Main.hs
```diff
@@ -181,9 +182,9 @@ generateSwaggerDocumentation api = liftIO $ do
 -- | The main entrypoint for the Wallet.
 main :: IO ()
 main = withCompileInfo $(retrieveCompileTimeInfo) $ do
-  cfg <- getWalletNodeOptions
-  putText "Wallet is starting..."
-  let loggingParams = CLI.loggingParams defaultLoggerName (wsoNodeArgs cfg)
-  loggerBracket loggingParams . runProduction $ do
-    logInfo "[Attention] Software is built with the wallet backend"
-    startEdgeNode cfg
+    cfg <- getWalletNodeOptions
```
adinapoli-iohk: 4 spaces indent police! 😁

PR #2642, opened 2018-03-08
[CSL-2326] killNode on async exception

avieth:
This should means that a SIGINT will kill all handlers and you won't have to wait.

### DIFF
--- a/networking/src/Node.hs
+++ b/networking/src/Node.hs
@@ -306,11 +306,8 @@ node mkEndPoint mkReceiveDelay mkConnectDelay prng packing peerData nodeEnv k = 
               converse :: Converse packing peerData m
               converse = nodeConverse llnode packing
               unexceptional :: m t
-              unexceptional = do
-                  t <- act converse
-                  logNormalShutdown
-                  (LL.stopNode llnode `catch` logNodeException)
-                  return t
+              unexceptional =
+                  act converse <* (logNormalShutdown >> LL.stopNode llnode)
         ; llnode <- LL.startNode
               packing
               peerData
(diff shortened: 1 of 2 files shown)


avieth:
@gromakovsky Check it out

avieth:
NB It could still deadlock on `SIGINT` if one of the handlers doesn't deal properly with async exceptions, or has some releaser which deadlocks.

networking/src/Node.hs
```diff
@@ -307,10 +308,9 @@ node mkEndPoint mkReceiveDelay mkConnectDelay prng packing peerData nodeEnv k =
               converse = nodeConverse llnode packing
               unexceptional :: m t
               unexceptional = do
-                  t <- act converse
-                  logNormalShutdown
-                  (LL.stopNode llnode `catch` logNodeException)
-                  return t
+                  (act converse <* logNormalShutdown)
+                  `finally`
+                  (LL.stopNode llnode)
```
gromakovsky: Is it correct? AFAIU, when an async exception is thrown, `LL.stopNode llnode` will be executed first (due to `finally`) and then `LL.killNode llnode` will be executed as well (due to `onException`). But I suppose we only need to `killNode`.
Also I think it would be ok to leave `logNodeException` here, but I'll leave this decision to you.
And also brackets on this line are redundant.

avieth: Right my bad, it should be `act converse <* (logNormalShutdown >> LL.stopNode llnode)`

gromakovsky:
I also changed base branch to `cbr98-reliable-shutdown`.

PR #2643, opened 2018-03-08
[CSL-2343] Remove wallet/node directory

gromakovsky:
It's not used anymore, so I suppose it should be removed as part of CSL-2343.


hithroc:
Should `scripts/launch/demo-with-new-wallet-api.sh` also be removed? Asking here because seems like relevant pull request.

gromakovsky:
Hm, yes, I think so.

gromakovsky:
Actually no, I don't think so, because there is `--new-wallet` flag in it. So probably it should be left for now.

adinapoli-iohk:
@KtorZ @parsonsmatt I think this PR is worthwhile reviving, as we removed the `executable` flag from `wallet.cabal` but we are still keeping the content inside `wallet/node/*` alive for no good reason.

Would you resurrect this one and give it another spin?

akegalj:
Closed in favour of https://github.com/input-output-hk/cardano-sl/pull/3104

PR #2645, opened 2018-03-08
WIP: Document coinbase transactions validation

erikd:
This is work in progress. Putting it up as a PR so that others can keep an eye on the progress being made and comment on the direction this is taking, things that need to be clarified etc.

I will just be pushing new commits on top of this one and rebase it into a single commit before merging.


erikd:
Closing this and moving it to my own fork of the repo so i can push changes without triggering CI.

Will repon when the document is done.

PR #2646, opened 2018-03-08
WIP: Document coinbase transaction validation

erikd:
This is work in progress. Putting it up as a PR so that others can keep an eye on the progress being made and comment on the direction this is taking, things that need to be clarified etc. I'm pushing to my clone of the IOHK repo so I don't trigger CI every time I push a change.

I will just be pushing new commits on top of this one. When its done, I'll rebase it into a single commit and raise a PR on the IOHK repo so CI runs.


erikd:
Still seems to be triggering CI. Closing this.

PR #2647, opened 2018-03-09
[CSL-2221] Fix OS X installer path check

Dolgakov:
Now UpdaterData is algebraic data type with different constructors for Windows and OS X. I think that approach brings more clarity to the purpose of OS dependent files. All required paths check for that config have been Implemented.
Also, a hash check of downloaded (but yet not installed) update has been implemented in the launcher.

I don`t know what to do if pattern matching on OS name or fromJust failed (It should not). In a current implementation, it throws an error, but it should not be in production, of course. Universum 1.0.1 has a bug function for that case, but It seems you want 0.9.0.


cleverca22:
@Dolgakov what will this code do if ran on linux?

Dolgakov:
@cleverca22 Now It will just skip an update routine. It seems there is no updater implementation for Linux (At least for Daedalus).

update/Pos/Update/DB/Misc.hs
```diff
@@ -22,3 +25,12 @@ affirmUpdateInstalled h = miscPutBi (updateTrackKey h) ()
 
 updateTrackKey :: Hash Raw -> ByteString
 updateTrackKey h = "updinst/" <> encodeUtf8 (sformat hashHexF h)
+
+lastInstallerKey :: ByteString
+lastInstallerKey = "updlast/lastDownloaded" 
+
+getLastInstallerHash :: MonadDBRead m => m (Hash Raw)
+getLastInstallerHash = fromJust <$> (miscGetBi lastInstallerKey)
```
gromakovsky: I think `fromJust` here is a bad idea, I would return `Maybe` instead, you can't know for sure that there is a value for this key. And in `launcher` if `getLastInstallerHash` returns `Nothing` you can assume that your installer is definitely invalid (maybe someone copied some file there).

tools/src/launcher/Main.hs
```diff
@@ -128,12 +129,20 @@ data NodeData = NodeData
     , ndArgs    :: ![Text]
     , ndLogPath :: Maybe FilePath }
 
--- | Updater path, args, windows runner path, archive path
-data UpdaterData = UpdaterData
-    { udPath        :: !FilePath
-    , udArgs        :: ![Text]
-    , udWindowsPath :: Maybe FilePath
-    , udArchivePath :: Maybe FilePath
+data UpdaterData = WindowsUpdaterData WinUD | OSXUpdaterData OSXUD | UnsupportedOSUpdateData UnsupOSUD
```
gromakovsky: I think `UnsupOSUD` is useless. You can write `data UpdaterData = WindowsUpdaterData WinUD | OSXUpdaterData OSXUD | UnsupportedOSUpdateData`.

tools/src/launcher/Main.hs
```diff
@@ -128,12 +129,20 @@ data NodeData = NodeData
...
+
+data UnsupOSUD = UnsupOSUD
+
+data WinUD = WinUD 
+    { wudExePath :: !FilePath -- Installer.exe
+    , wudArgs    :: ![Text]
+    , wudBatPath :: !FilePath -- Installer.bat
+    }
+
+data OSXUD = OSXUD
```
gromakovsky: We tend to avoid so long abbreviations, because it's hard to distinguish parts of it. You can call it `OsxUD` or maybe `MacUD` or `MacOsUD`.

tools/src/launcher/Main.hs
```diff
@@ -128,12 +129,20 @@ data NodeData = NodeData
...
-    , udWindowsPath :: Maybe FilePath
-    , udArchivePath :: Maybe FilePath
+data UpdaterData = WindowsUpdaterData WinUD | OSXUpdaterData OSXUD | UnsupportedOSUpdateData UnsupOSUD
+
+data UnsupOSUD = UnsupOSUD
+
+data WinUD = WinUD 
+    { wudExePath :: !FilePath -- Installer.exe
+    , wudArgs    :: ![Text]
+    , wudBatPath :: !FilePath -- Installer.bat
```
gromakovsky: To be honest I don't like that more knowledge about update process is added to launcher. This knowledge should be mostly in installers ([daedalus/installers](https://github.com/input-output-hk/daedalus/tree/develop/installers)).
Some knowledge was here for a long time (about bat usage on windows), so I am not asking to eliminate. But let's describe these fields using a bit more abstract terms.
`wudExePath` – installer executable (e. g. Installer.exe).
`wudArgs` – command line arguments passed to the installer executable.
`wudBatPath` – path where bat file with installer instructions will be written.

tools/src/launcher/Main.hs
```diff
@@ -128,12 +129,20 @@ data NodeData = NodeData
...
+    }
+
+data OSXUD = OSXUD
+    { oudOpenPath :: !FilePath -- /usr/bin/open
+    , oudArgs     :: ![Text] -- FW
+    , oudPkgPath  :: !FilePath -- Installer.pkg
     }
```
gromakovsky: Some comment about more abstract terms applies here.

`oudOpenPath` – executable to launch the installer (e. g. /usr/bin/open).
`oudArgs` – command line arguments passed to `oudOpenPath` executable.
`oudPkgPath` – path to installer package, passed to `oudOpenPath` (e. g. Installer.pkg).

tools/src/launcher/Main.hs
```diff
@@ -338,13 +346,31 @@ main =
...
                     (NodeData wpath loWalletArgs loWalletLogPath)
-                    (UpdaterData
-                        loUpdaterPath loUpdaterArgs loUpdateWindowsRunner loUpdateArchive)
+                    (setUpdaterData lo)
                     loNodeTimeoutSec
                     loReportServer
                     loWalletLogging
                 logNotice "Finished clientScenario"
   where
+    setUpdaterData :: LauncherOptions -> UpdaterData
```
gromakovsky: Why is it called `set`? :thinking: 
I associate `set` with `Control.Lens.set`. Why not `createUpdaterData` or `constructUpdaterData` or `toUpdaterData` or `mkUpdaterData`?

tools/src/launcher/Main.hs
```diff
@@ -338,13 +346,31 @@ main =
...
+            }
+        Windows -> WindowsUpdaterData WinUD
+            { wudExePath = loUpdaterPath
+            , wudArgs = loUpdaterArgs
+            -- It would be better to use `bug` function from universum-1.0.1 but you want 0.9.0
+            -- , wudBatPath = fromMaybe (bug Not_set_Windows_loUpdateWindowsRunner) loUpdateWindowsRunner
+            , wudBatPath = fromJust loUpdateWindowsRunner            
+            }
+        -- It would be better to use `bug` function from universum-1.0.1 but you want 0.9.0
+        _ -> UnsupportedOSUpdateData UnsupOSUD
```
gromakovsky: This line is not a bug, it's absolutely normal, there shouldn't be any errors, it's ok to run launcher on linux, for example.

tools/src/launcher/Main.hs
```diff
@@ -338,13 +346,31 @@ main =
...
                 logNotice "Finished clientScenario"
   where
+    setUpdaterData :: LauncherOptions -> UpdaterData
+    setUpdaterData LO {..} = case buildOS of 
+        OSX -> OSXUpdaterData OSXUD
+            { oudOpenPath = loUpdaterPath
+            , oudArgs = loUpdaterArgs
+            -- It would be better to use `bug` function from universum-1.0.1 but you want 0.9.0
+            -- , oudPkgPath = fromMaybe (bug Not_set_OSX_loUpdateArchive) loUpdateArchive
+            , oudPkgPath = fromJust loUpdateArchive
```
gromakovsky: See [exception handling guidelines](https://github.com/input-output-hk/cardano-sl/blob/37ab33599470b30c608528d5a844a3a59ba9843f/docs/exceptions.md). Functions like `fromJust` and `bug` should be used only if their failure implies a programmer mistake. But here programmer can't guarantee that launcher config will have `loUpdateArchive` when launcher is launched on OSX. The config is passed as a command line option, so anything can be passed there.
Consider changing `setUpdaterData` type to `LauncherOptions -> IO UpdaterData` (and use `throwM` if something goes wrong).

tools/src/launcher/Main.hs
```diff
@@ -479,46 +505,88 @@ clientScenario ndbp logConf node wallet updater nodeTimeout report walletLog = d
...
+    logIfM predicate message = ifM predicate (return True) ((logWarning message) >> return False)
+
+    logIf predicate message = if predicate 
+        then return True
+        else (logWarning message) >> return False
+
+    validUD :: UpdaterData -> M Bool
+    validUD (WindowsUpdaterData WinUD {..}) = logIfM (liftIO $ doesFileExist wudExePath) "Installer.exe not found"
+    validUD (OSXUpdaterData OSXUD {..}) = logIfM (liftIO $ doesFileExist oudPkgPath) "Installer.pkg not found"
+    validUD (UnsupportedOSUpdateData UnsupOSUD) = logIf True "Current OS is not supported by updater"
```
gromakovsky: 1. `logIf True` doesn't make much sense you can just inline it here. And defining `logIf` is not really useful, because it's used in only one place and `True` is unconditionally passed to it.
2. Let's not hardcode file paths here (I mean `Installer.exe` and `Installer.pkg`) please. They come from the outside, and nobody guarantees they will always be called this way.
3. Actually you don't need to distinguish `WindowsUpdaterData` and `OSXUpdaterData` cases here (if you don't hardcode paths). You can use `getInstaller` in both cases.
4. I suppose it should be `logIf False` here, not `True`.

tools/src/launcher/Main.hs
```diff
@@ -479,46 +505,88 @@ clientScenario ndbp logConf node wallet updater nodeTimeout report walletLog = d
...
+                -- will launch it again and again.
+                affirmInstalled `finally` removeInstaller
+        ExitFailure code ->
+            logWarning $ sformat ("The updater has failed (exit code "%int%")") code
+
+    logIfM predicate message = ifM predicate (return True) ((logWarning message) >> return False)
+
+    logIf predicate message = if predicate 
+        then return True
+        else (logWarning message) >> return False
```
gromakovsky: Btw, `>> return` is `$>` (i. e. this line can be written as `logWarning message $> False`), but it's up to you to decide which version is better.

tools/src/launcher/Main.hs
```diff
@@ -479,46 +505,88 @@ clientScenario ndbp logConf node wallet updater nodeTimeout report walletLog = d
...
+        return $ currentHash == validHash 
+
+    getInstaller :: UpdaterData -> FilePath
+    getInstaller (WindowsUpdaterData WinUD {..}) = wudExePath
+    getInstaller (OSXUpdaterData OSXUD {..}) = oudPkgPath
+    -- Should not happen due to whenM in runUpdater
+    getInstaller _ = error "Unsupported OS UpdateConfig has no Installer"
+
+    archiveAndRemove :: ExitCode -> FilePath -> M ()
+    archiveAndRemove exitCode installerPath = case exitCode of
```
gromakovsky: Archive what?

cleverca22:
@Dolgakov i just finished testing the update installing process on linux, and now need to test downloading

tools/src/launcher/Main.hs
```diff
@@ -128,13 +126,23 @@ data NodeData = NodeData
...
-    { udPath        :: !FilePath
-    , udArgs        :: ![Text]
-    , udWindowsPath :: Maybe FilePath
-    , udArchivePath :: Maybe FilePath
-    }
+data UpdaterData = WindowsUpdaterData WinUD | OSXUpdaterData OsxUD | UnsupportedOSUpdateData deriving (Generic)
+
+instance FromJSON UpdaterData
+instance FromJSON WinUD
+instance FromJSON OsxUD
```
gromakovsky: Can you use `genericParseJSON defaultOptions` in these instances please? `Serokell.Util.defaultOptions` are different from default options in `aeson` and they are more suitable for our code.

gromakovsky: Also previously `updaterArgs` were optional, but now they are not. You can fix it using the trick which was used for `LauncherOptions`.

tools/src/launcher/Main.hs
```diff
@@ -220,38 +228,58 @@ getLauncherOptions = do
...
+            appdata <- toText <$> getEnv "APPDATA"
+            daedalusDir <- (toText . takeDirectory) <$> getExecutablePath
+            let replaceDaedalusDir = replace "%DAEDALUS_DIR%" daedalusDir
+
+                replaceAppdata = replace "%APPDATA%" appdata 
+                replaceUpdaterAppdata (WindowsUpdaterData wud@WinUD {..}) = WindowsUpdaterData wud 
+                    { wudExePath = replaceAppdata wudExePath
+                    , wudBatPath = replaceAppdata wudBatPath
+                    }
+                replaceUpdaterAppdata _ = error "Can not expand %APPDATA%"
```
gromakovsky: No, don't use `error` here please, as I said earlier `error` means a programmer mistake, but here we don't which options will be passed.

tools/src/launcher/Main.hs
Dolgakov: Did you mean Serokell.Aeson.Options (defaultOptions)? Because I can`t see defaultOptions 
 in Serokell.Util. 
>Also previously updaterArgs were optional

What do you mean by optional? It was a [Text]: [] for Windows and ["--FW"] for OS X. I thought it stayed untouched. And I don`t understand, why launcher-config.yaml looks like:
```
# Arguments to be passed to the updater.
updaterArgs:
  - "dir"
  - "binaries"
```

Dolgakov: I think I already got it. In case you want to answer.

jcmincke:

PR #2648, opened 2018-03-09
lib: Remove unused module Pos.Crypto.HDDiscovery

erikd:


erikd:
CI failures were both in the `cardano-wallet-test` testsuite which is unrelated to this PR.

pva701:
@erikd this module was used for tracking blockchain, then we stop using it, but there is CSM-360 which implies using this function

erikd:
@pva701 Even if the code in this file is going to be used at some point to me it seems this file does not belong in this package. It uses functionality from `core`, `db` and `txp`. If this code is needed it should probably be moved to the `txp` package.

erikd:
@adinapoli-iohk Pinging you on this because I think its related to your work on #2795.

Will revisit at a later date.

PR #2649, opened 2018-03-09
[CSL-2354] Merge master to develop

nc6:


wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -132,7 +133,8 @@ walletBackend (NewWalletBackendParams WalletBackendParams{..}) passive =
 
 -- | A @Plugin@ to resubmit pending transactions.
 resubmitterPlugin :: (HasConfigurations, HasCompileInfo) => Plugin WalletWebMode
-resubmitterPlugin = ([ActionSpec $ \diffusion -> startPendingTxsResubmitter (sendTx diffusion)], mempty)
+resubmitterPlugin = ([ActionSpec $ \diffusion -> askWalletDB >>= \db ->
+                         startPendingTxsResubmitter db (sendTx diffusion)], mempty)
```
adinapoli-iohk: Yes, here `askWalletDB` yields the `acid-state` DB handle, so that's obviously necessary to pass it along and have the async worker re-read the snapshot every time, to be able to resubmit transactions.

wallet/src/Pos/Wallet/Web/Methods/Backup.hs
```diff
@@ -52,18 +53,18 @@ restoreWalletFromBackup WalletBackup {..} = do
...
                     () <$ L.newAccountIncludeUnready True defaultAccAddrIdx emptyPassphrase accInit
-                else for_ accList $ \(idx, meta) -> do
-                    let aIdx = fromInteger $ fromIntegral idx
-                        seedGen = DeterminedSeed aIdx
-                    accId <- genUniqueAccountId seedGen wId
-                    createAccount accId meta
-
+                else do
+                    for_ accList $ \(idx, meta) -> do
+                        ws <- getWalletSnapshot db
```
adinapoli-iohk: Yep, here is necessary to re-read the state at each pass of the loop as we need the updated snapshot to be able to compute the next random index whilst avoiding collisions.

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -65,43 +66,49 @@ walletHistorySize =
...
 
 getFullWalletHistory
-    :: MonadWalletHistory ctx m
-    => CId Wal -> m (WalletHistory, WalletHistorySize)
-getFullWalletHistory cWalId = do
+    :: (MonadWalletHistory ctx m)
+    => WalletDB
+    -> CId Wal
+    -> m (WalletHistory, WalletHistorySize)
+getFullWalletHistory db cWalId = do
```
adinapoli-iohk: Minor observation: we probably could have been even stricter and not pass `WalletDB` at all, but directly the `WalletSnapshot`, as I think we are not re-reading it twice.

adinapoli-iohk: Ah, I have just noticed that in the call to `getHistory` we are asking the `WalletDB` directly, as if `MonadWalletHistory` had already an implicit `WalletDbReader`defined. Are we sure we want that?

wallet/src/Pos/Wallet/Web/Methods/History.hs
```diff
@@ -175,53 +182,58 @@ getHistoryLimited mCWalId mAccId mAddrId mSkip mLimit = do
         "Please do not specify both walletId and accountId at the same time"
 
 addHistoryTxMeta
-    :: MonadWalletDB ctx m
-    => CId Wal
+    :: (MonadIO m, HasConfiguration)
+    => WalletDB
+    -> CId Wal
```
adinapoli-iohk: Nice, ideally that's the type of monad we want to work in for non-toplevel functions 😀

adinapoli-iohk:
> Will also try to test this locally with a dev cluster, and report back.

@nc6 Tested with a local cluster, so far so good:

- [x] Import Genesis Wallet
- [x] Restore a wallet from seed
- [x] Send/receive a transaction
- [x] Create a new wallet
- [x] Generate new addresses

wallet/src/Pos/Wallet/Web/State/State.hs
```diff
@@ -81,282 +82,448 @@ module Pos.Wallet.Web.State.State
...
+  -> m ()
+rollbackModifierFromWallet db walId wAddrs custAddrs utxoMod
+                           historyEntries ptxConditions
+                           syncTip =
+    updateDisk
+      ( A.RollbackModifierFromWallet
+          walId wAddrs custAddrs utxoMod
+          historyEntries' ptxConditions syncTip
+      )
+      db
```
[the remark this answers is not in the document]
nc6: Agree, but don't think it should be part of this PR!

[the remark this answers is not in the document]
adinapoli-iohk: Maybe I can cram it as part of CBR-90 😉

adinapoli-iohk:
Finally merging this bad body, @nc6 !

As clarified with @domenkozar , the hydra failures are genuine but they affect only Mac devs building cardano via `nix` (using `stack test` worked fine on my machine).

Provided we open a follow-up PR using `weeder` to get rid of the unused deps, we are now ready to go. 🎉

PR #2650, opened 2018-03-09
[CBR-102] - Document the new Wallet V1 API

KtorZ:
As dicussed, tackles the following tasks from CBR-102

*CSL-2273* | Expand introduction with install instructions and curl example
*CSL-2274* | Extend error handling table
*CSL-2277* | Document V1 exchange endpoints
*CSL-2359* | Serve API documentation through dedicated endpoints


wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -231,150 +181,223 @@ The number of entries to display for each page. The minimum is **1**, whereas th
...
+```
+curl -X POST https://localhost:8090/api/v1/wallets                     \
+     -H "Content-Type: application/json; charset=utf-8"                \
+     -H "Accept: application/json; charset=utf-8"                      \
+     --cacert ./scripts/tls-files/ca.crt                               \
+     -d '{                                                             \
+  "operation": "create",                                               \
+  "backupPhrase": ["squirrel", "material", "silly", "twice", "direct", \
+    "slush", "pistol", "razor", "become", "junk", "kingdom", "flee"],  \
+  "assuranceLevel": "normal",                                          \
```
adinapoli-iohk: A note we should add (just recalled a conversation I had with @jmitchell ): We need to make clear that those 12 words mnemonic are given just as an example and that is not secure by any means reusing it for a production wallet, and that doing so might cause funds to be lost.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -231,150 +181,223 @@ The number of entries to display for each page. The minimum is **1**, whereas th
...
+  "name": "MyFirstWallet"                                              \
+}'
+```
 
-possibleValuesOf :: (Show a, Enum a, Bounded a) => Proxy a -> T.Text
-possibleValuesOf (Proxy :: Proxy a) = T.intercalate "," . map show $ ([minBound..maxBound] :: [a])
+As a response, the API provides you with a wallet `id` used in subsequent requests to uniquely
```
adinapoli-iohk: Maybe something to add to subsequent PRs, but would be nice to include:

* Two sentences on the response payload (i.e. JSend, but maybe that's documented later) and specifically about the fact we send automatically pagination information;

* Spend two words on the `assuranceLevel`, even just saying we are going to describe it in more detail later.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -231,150 +181,223 @@ The number of entries to display for each page. The minimum is **1**, whereas th
...
-  annotate f p = do
-    s <- f p
-    return $ s & (schema . description ?~ "A list of " <> renderType p <> ".")
+Receiving Money
+===============
 
---
--- The API
---
+To receive money from other user you should provide your address. This address can be obtained
```
adinapoli-iohk: `from other user` or `from other users`? 😉

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -231,150 +181,223 @@ The number of entries to display for each page. The minimum is **1**, whereas th
...
+  "source": {                                           \
+    "accountIndex": 0,                                  \
+    "walletId": "Ae2tdPwUPE...8V3AVTnqGZ"               \
+  }                                                     \
+}'
+```
 
-This means that _omitting_ the version number would call the old version of the API. Examples:
+Note that, in order to perform a transaction, you need to have some existing coins on the
+source account! When the transaction succeeds, funds are _moved_ from an address to another.
```
adinapoli-iohk: A couple of things here:

* We should somehow mention our API supports out-of-the-box multiple recipients payments, which means you can pay more than 1 entity with a single transaction;

* I'm wondering if saying that funds are "moved" is so generic is better to be slightly more specific;

* We should mention that, due to the fact we cannot "preview" a transaction the only way to be aware of the transaction fees for this payment is to call `/api/v1/transactions/fees` which serves exactly this purpose: "previewing" the transaction fees for a payment.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -231,150 +181,223 @@ The number of entries to display for each page. The minimum is **1**, whereas th
...
+        }
+    }
+}
+```
 
-## Response format (V1 onwards)
+Pagination
+----------
 
 **All GET requests of the API are paginated by default**. Whilst this can be a source of surprise, is
```
adinapoli-iohk: Perfect, as we are already piggybacking on the old skeleton about pagination, we should at least say hint to the reader, at the beginning (see "Creating a new wallet"), this section.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -386,41 +409,42 @@ used to modify the shape of the response. In particular, those are:
...
 
-### Filter operators
+Filter operators
+================
 
 | Operator | Description                                                               | Example                |
 |----------|---------------------------------------------------------------------------|------------------------|
 | -        | If **no operator** is passed, this is equivalent to `EQ` (see below).     | `balance=10`           |
-| `EQ`     | Retrieves the resources with index _equal to the one provided.            | `balance=EQ[10]`       |
+| `EQ`     | Retrieves the resources with index _equal_ to the one provided.           | `balance=EQ[10]`       |
```
adinapoli-iohk: nice catch!

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -431,54 +455,75 @@ application. For example, here's a typical error which might be issued:
...
+======================
 
 $deWalletErrorTable
 
+
+Mnemonic Codes
+--------------
+
+The full list of accepted mnemonic codes to secure a wallet is defined by the BIP-39
+specifications and available [here](https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt).
```
adinapoli-iohk: Let's add a caveat here: it's not _enough_ to pick 12 random words from that list, but obviously the entire BIP-39 algorithm must be followed to generate entropy such that 12 random words from that list can be derived. I was myself guilt in the past for this false belief and I have reasons to believe other users will be hit by the same pitfall. Better be explicit 😁

wallet-new/server/Cardano/Wallet/LegacyServer.hs
```diff
@@ -4,27 +4,41 @@ module Cardano.Wallet.LegacyServer where
...
-walletServer :: ( Migration.HasConfigurations
-                , Migration.HasCompileInfo
-                )
+walletServer :: (Migration.HasConfigurations, Migration.HasCompileInfo)
              => (forall a. WalletWebMode a -> Handler a)
              -> Diffusion WalletWebMode
-             -> RunMode
              -> Server WalletAPI
-walletServer natV0 diffusion runMode = externalAPI :<|> internalAPI
+walletServer natV0 diffusion = v0DocHdl :<|> v1DocHdl :<|> v0Hdl :<|> v1Hdl
```
adinapoli-iohk: We lost a bit of terseness here -- do you think slightly longer names can help?

wallet-new/server/Cardano/Wallet/LegacyServer.hs
```diff
@@ -4,27 +4,41 @@ module Cardano.Wallet.LegacyServer where
...
+    v1DocHdl = swaggerSchemaUIServer (Swagger.api (compileInfo, curSoftwareVersion) v1API Swagger.highLevelDescription)
+    v1Hdl    = V1.handlers natV0 diffusion
+
+
+walletDevServer :: (Migration.HasConfigurations, Migration.HasCompileInfo)
+             => (forall a. WalletWebMode a -> Handler a)
+             -> Diffusion WalletWebMode
+             -> RunMode
+             -> Server WalletDevAPI
+walletDevServer natV0 diffusion runMode = devDocHdl :<|> devHdl :<|> walletHdl
```
adinapoli-iohk: Same as above -- as we are using these vars just once, I think that's not the end of the world being slightly more verbose.

wallet-new/server/Cardano/Wallet/Server.hs
```diff
@@ -1,36 +1,47 @@
...
+import           Servant.Swagger.UI (swaggerSchemaUIServer)
 
 -- | Serve the REST interface to the wallet
 --
 -- NOTE: Unlike the legacy server, the handlers will not run in a special
 -- Cardano monad because they just interfact with the Wallet object.
-walletServer :: ActiveWallet
-             -> RunMode
+walletServer :: (HasCompileInfo, HasUpdateConfiguration)
+             => ActiveWallet
```
adinapoli-iohk: As this is @edsko 's territory mainly, I just wanted to check if he's happy with the (even if temporary) signature change, as I know he's working hard to have as less "legacy baggage" as possible for the new handlers 😁

wallet-new/src/Cardano/Wallet/API/V1/Generic.hs
```diff
@@ -159,4 +157,11 @@ gconsInfos pa = case datatypeInfo pa of
...
     => Proxy a -> [Text]
-gconsNames = map toText . hcollapse . hliftA gconsName' . gconsInfos
+gconsNames =
+  map toText . hcollapse . hliftA (K . constructorName) . gconsInfos
+
+gconsName
+  :: forall a. (Generic a, HasDatatypeInfo a, SListI (Code a))
+  => a -> Text
+gconsName a =
+  gconsNames (Proxy @a) !! hindex (from a)
```
adinapoli-iohk: Are you confident partiality (as in using `(!!)`) is not a problem here?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -395,7 +450,7 @@ newtype AddressValidity = AddressValidity { isValid :: Bool }
 deriveJSON Serokell.defaultOptions ''AddressValidity
 
 instance ToSchema AddressValidity where
-    declareNamedSchema = genericSchemaDroppingPrefix "is"
+    declareNamedSchema = genericSchemaDroppingPrefix "is" (\_ -> Prelude.id)
```
adinapoli-iohk: You can use `identity` from `Universum` if you want, that's exactly the same 😁

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -661,30 +751,32 @@ instance ToSchema TransactionDirection where
...
-    -- ^ The direction for this transaction (e.g incoming, outgoing).
   } deriving (Show, Ord, Eq, Generic)
 
 deriveToJSON Serokell.defaultOptions ''Transaction
 
 instance ToSchema Transaction where
-    declareNamedSchema = genericSchemaDroppingPrefix "tx"
+  declareNamedSchema =
+    genericSchemaDroppingPrefix "tx" (\(--^) props -> props
+      & ("id"            --^ "Transaction's id")
```
adinapoli-iohk: This would be probably overkill, but I have seen some ideas floating around in the Servant issue tracker where people were thinking of using a type similar to `(--^)` to attach the documentation stuff directly in the types, when you declare the data record. Example:

``` haskell
data Foo = Foo {
  bar :: Int --^ "Documentation for Int"
}
```

It probably makes working with these types madness, in practice, so I'm happy you went for a different approach, but I thought I would share it just for reference. Incidentally, I think also Gabriel Gonzalez uses a similar combinator is his library `optparse-generic`.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Good one. I'll drop a sentence early in the guide and maybe elaborate a bit more in the part about `Mnemonic Codes` :+1:

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Wallets.hs
```diff
@@ -23,15 +23,17 @@ handlers :: ( HasConfigurations
...
         :<|> listWallets
         :<|> (\walletId ->
                      updatePassword walletId
                 :<|> deleteWallet walletId
                 :<|> getWallet walletId
                 :<|> updateWallet walletId
-                :<|> Accounts.handlers walletId
              )
+        )
+        :<|> (\walletId -> Accounts.handlers walletId)
```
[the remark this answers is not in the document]
adinapoli-iohk: Ah, true -- you told me about it but I (obviously) forgot 😁 What about adding a comment in the code unraveling the arcane? I can imagine how the next developer refactoring this endpoints might think that was a mistake and accidentally undo your work 😉

wallet-new/src/Cardano/Wallet/API/V1/Generic.hs
KtorZ: I am not yet really familiar with the *generics-sop* module but as far as I understand, this should be fine since we would call it on types represented as a sum (or a singleton) of constructors. As far as I know, we cannot create new types that have no constructor in Haskell? (`newtype` or `data` or extensions that boil down to these two...). Hence, using their corresponding indexes to access constructors' names should work fine; I didn't find any other way to do that with `generics-sop`.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: Ah! Thanks. Not used to `Universum` yet.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: I've tried this approach on a previous project. Very neat to do generic stuff with the types after that, yet a bit painful indeed because each field has to be unwrap after that; compile-times are also significantly slower. 

Note that unfortunately, I didn't find any way to have this statically checked without altering or annotating the original data-type. It will however throw explicitly should we try to document a non-existing field.

KtorZ:
I'll fix the fist few comments you had and look into those lost types :fearful:  Probably something with the `ToSchema` instances we now use that aren't accurate enough.

adinapoli-iohk:
> Probably something with the ToSchema instances we now use that aren't accurate enough.

Now you understand why I opted, in the early days, for `ToDocs` and `Example`: what we did lose in randomness we gained in specificity 😁 That's probably possible to reify the two worlds, with a mallet, an aspirin and goodwill :trollface:

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
adinapoli-iohk: For now let's not worry too much about it, but let it marinade. I'm all for embedding informations in the types, but on this particular topic I don't think it would be beneficial. Servant error messages are already quite daunting, and I'm wondering what would happen if we would throw textual documentation into the verbosity 😱

KtorZ: Haha ^.^ .. from my own experience, it doesn't get much worse (can it though :/ ?); this is actually the same strategy we use for tagging.

What (other) strategy would recommend to document endpoints' descriptions?

adinapoli-iohk: > What (other) strategy would recommend to document endpoints' descriptions?

I don't know if that's a stupid idea or even possible, but the next obvious choice would be to do that in our `Swagger.hs` module and use lenses and the traversals provided by `servant-swagger` to annotate the endpoint that way. This would have the advantage we can still use quasiquotes and embed markdown syntax without worrying about breaking GHC 😁

domenkozar: I'd suggest bringing this up in #servant channel on freenode, the guys are very responsive and would come up with answers they might already resolve in private projects as well.

KtorZ: Digging a bit into servant's issues, I just found a comment referring to: 

https://hackage.haskell.org/package/servant-0.12/docs/Servant-API-Description.html

Looks like what we are looking for :)

adinapoli-iohk: @domenkozar Yes, that would be a good idea, or even hit /r/haskell for some long-term discussion 😁 

@KtorZ Yep, this is also what I have used in the past but I'm not fully sold that embedding so much info into the types is a good idea here, but I'm happy to be proven wrong 😉

KtorZ:
Hey @adinapoli-iohk 

I have addressed the following points in b80def2 

- Longer identifiers in LegacyServer & Server
- Typo user -> users in API description
- Note about keeping "Wallets" & "Accounts" tag separated
- Extend API documentation about mnemonic codes (add clear warning)
- Replace use of 'Prelude.id' with 'identity' from Universum
- Extend description of the assuranceLevel
- Give more details about out-of-the-box multi recipients payments
- Leave note about "previewing" transactions fees
- Reword generic 'moved' to be slightly more specific
- Put error JSON exmaples inside code-blocks
- Adjust schema of PaginationMeta to reflect more appropriate totals value (not negative)
- Adjust Pattern of spending password (Base16, 32-byte long)

---

And repaired schema examples (as for the address) in 8b0eb2531d6c35189856ec02e366cee49316dce3

Let me know :+1:

adinapoli-iohk:
@KtorZ argh, almost there! I think that's just the specs (as in `stack test cardano-sl-wallet-new`) needs to be fixed now.

KtorZ:
There was a typo in one of the Swagger Schemas (LocalTimeDifference, `milliseconds` instead of `microseconds`) and a wrong constraint on the spendingPassword schema. Fixes incoming :+1:

adinapoli-iohk:
Outstanding work, thanks @KtorZ ! (cc @parsonsmatt )

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
parsonsmatt: At my previous job we'd mostly use `servant-swagger` to generate routes and parameter specs -- the documentation aspect was mostly done to the generated spec, precisely so we could do it at the value level. Stuff like quasiquoters for multiline strings and `mconcat` are very nice and unavailable on symbols :\

wallet-new/src/Cardano/Wallet/API/V1/Generic.hs
parsonsmatt: With `{-# LANGUAGE EmptyDataDecls  #-}`, you can write `data Void`.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: TIL you can introduce operators in a lambda -- super cool idiom :smile:

wallet-new/src/Cardano/Wallet/API/V1/Generic.hs
KtorZ: I'd like to propose this new theorem:

> Everytime one thinks something isn't possible in Haskell, one discovers a new extension that does it.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: Haha, I believe that's the first time I do it... but it didn't felt wrong, and it compiled so... ^.^

wallet-new/src/Cardano/Wallet/API/V1/Generic.hs
parsonsmatt: One either discovers a new extension or patches GHC to add the extension :smile:

parsonsmatt: Though, on the safety benefit, you can't actually *construct* a value of type `Void`, so this function *is* safe, in the sense that `gconsName undefined == undefined` -- the error won't come from `[] !! 0` but from `undefined`!

PR #2651, opened 2018-03-09
[DEVOPS-682] build explorer frontend with nix

rvl:
- Explorer frontend builds with pure nix derivation.
 - CI script uses `nix-build` instead of `frontend/scripts/build.sh`.
 - There is an easy process to regenerate nix expressions for bower and npm.
 - If Nix expressions go out of date (bower/npm is updated), CI will fail


explorer/frontend/scripts/regen-shell.nix
```diff
@@ -0,0 +1,14 @@
+# args@{...}: (import ./default.nix (args // { cardano-sl-explorer = null; })).shell
+
+{ pkgs ? import <nixpkgs> {} }:
```
cleverca22: should probably use `locallib.fetchNixPkgs` here, or the version will be in flux and can fail

rvl: Fixed

.buildkite/pipeline.yml
```diff
@@ -23,3 +23,8 @@ steps:
     command: 'scripts/ci/nix-shell.sh --run scripts/ci/check-hydra.sh'
     agents:
       system: x86_64-linux
+
+  - label: 'bower2nix and node2nix'
```
domenkozar: I'd be more specific here: `explorer frontend: bower2nix and node2nix`

.buildkite/pipeline.yml
```diff
@@ -23,3 +23,8 @@ steps:
...
+    command: 'scripts/ci/nix-shell.sh --run "cd explorer/frontend && ./scripts/regen.hs --test"'
```
domenkozar: `regen` is too generic, maybe `regenExplorerFrontend`?

explorer/frontend/default.nix
```diff
@@ -1,26 +1,121 @@
...
+  };
+
+  nodePackages = import ./composition.nix {
+    inherit pkgs system;
+    inherit src;
+  };
+
+  # p-d-l does not build with our main version of nixpkgs.
+  # Needs to use something off 17.03 branch.
+  oldHaskellPackages = (import (fetchTarball https://github.com/NixOS/nixpkgs/archive/cb90e6a0361554d01b7a576af6c6fae4c28d7513.tar.gz) {}).pkgs.haskell.packages.ghc802.override {
```
domenkozar: Can we use `fetch-nixpkgs.nix` here from 17.09? We want to avoid having mulitple nixpkgs versions around.

explorer/frontend/scripts/regen-shell.nix
```diff
@@ -0,0 +1,18 @@
+let
```
domenkozar: Maybe we should move this to a folder and make this file `default.nix`

explorer/frontend/scripts/regen.hs
```diff
@@ -0,0 +1,133 @@
...
+hashFile src = fmap ("# " <>) (inproc "sha256sum" [tt src] empty)
+
+-- | Adds a hash to the top of the file
+stampFile :: FilePath -> Shell Line -> Shell Line
+stampFile ref f = cat [hashFile ref, f]
+
+----------------------------------------------------------------------------
+
+-- | Replace references to uglify version 3 with uglify version 2
+uglify2 :: Pattern Text
```
domenkozar: Why can't this be handled with conversion itself? Maybe we just need a comment :) I fear this will break in a month :)

rvl: Yes it's not very good. I couldn't find a way of forcing npm to always use a certain version (e.g. `yarn.lock` has version 3 in there also).

I have reworked to make the versions less hardcoded.

I have also generated a `package-lock.json`.

Hopefully these will reduce possibility of breakage.

explorer/frontend/default.nix
rvl: `purescript-derive-lenses` doesn't compile with the nixpkgs used everywhere else. So it needs to nixpkgs to stay as is. We could update `purescript-derive-lenses` to fix?

.buildkite/pipeline.yml
rvl: OK, fixed.

.buildkite/pipeline.yml
rvl: OK fixed.

explorer/frontend/default.nix
rvl: I have dug into this again because having multiple nixpkgs isn't good.

 - Changed it to use the ghc802 packageset from 17.09 (same as everything else).
 - When calling `purescript-derive-lenses` package it needs a jailbreak (because of `optparse-generic` version)
 - Then there is a build error
   ```
   src/Main.hs:66:18: error:
       Not in scope: data constructor ‘P.PositionedDeclaration’
       Module ‘Language.PureScript’ does not export ‘PositionedDeclaration’.
   ```

To fix this would probably require either a fork of [purescript-derive-lenses](https://github.com/paf31/purescript-derive-lenses) (the new version is written in purescript now), or trying to override the purescript version to something older than 0.11.6 (maybe 0.10.*).

rvl:
Have squashed and rebase on latest develop.
Explorer works in development cluster.

However there is a display glitch in this build. For some reason, the top search bar is always visible. I need to check that the dependencies are locked properly.

rvl:
I changed from `node2nix` to `yarn2nix` which allows pinning to `yarn.lock`.
This fixes the double search box problem.
Rebased on latest develop branch.

explorer/frontend/nix/generate-frontend-deps.hs
```diff
@@ -0,0 +1,80 @@
...
+hashFile :: [FilePath] -> Shell Line
+hashFile srcs = fmap ("# " <>) (inproc "sha256sum" (map tt srcs) empty)
+
+-- | Adds a hash to the top of the file
+stampFile :: [FilePath] -> Shell Line -> Shell Line
+stampFile ref f = cat [hashFile ref, f]
+
+----------------------------------------------------------------------------
+
+whenM :: Monad f => f Bool -> f () -> f ()
```
domenkozar: Why not use https://hackage.haskell.org/package/base-4.11.0.0/docs/Control-Monad.html#v:when

rvl: OK fixed.

PR #2652, opened 2018-03-09
[CSL-1628] Make ToilInvalidOutputs informative

mitchellvitez:
Added constructors based on checks in `Pos.Txp.Toil.Utxo.Functions.verifyOutputs`


txp/Pos/Txp/Toil/Failure.hs
```diff
@@ -89,8 +91,10 @@ instance Buildable ToilVerFailure where
         tInputSum tOutputSum
     build (ToilInconsistentTxAux msg) =
         bprint ("TxAux is inconsistent: "%stext) msg
-    build (ToilInvalidOutputs msg) =
-        bprint ("outputs are invalid: "%stext) msg
+    build (ToilInvalidOutputs i reason) =
+        bprint ("input #"%int%"'s outputs are invalid:\n'"%
```
gromakovsky: I would say `ords%" output is invalid:\n"%`, see [ords](https://hackage.haskell.org/package/formatting-6.3.1/docs/Formatting-Formatters.html#v:ords). It's not related to _input_.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -127,28 +125,18 @@ verifyConsistency inputs witnesses
...
-                    %"type ("%addressF%"), this is prohibited") i addr
-          )
-        , ( not (isRedeemAddress addr)
-          , sformat ("output #"%int%" sends money to a redeem address ("
-                    %addressF%"), this is prohibited") i addr
-          )
-        ]
+    verifyOutput :: (Word32, TxOut) -> Either ToilVerFailure ()
+    verifyOutput (i, (TxOut {txOutAddress = addr@Address {..}, ..})) = do
+        unless (vtcVerifyAllIsKnown && not (areAttributesKnown addrAttributes)) $
```
gromakovsky: It should be `when (vtcVerifyAllIsKnown && not (areAttributesKnown addrAttributes))`.
If `not vtcVerifyAllIsKnown || areAttributesKnown addrAttributes` is `False`, it's an error (in the previous version of this code).
Straightforward transformation is `unless (not vtcVerifyAllIsKnown || areAttributesKnown addrAttributes) $ throwError …`. Which is equivalent to `when (vtcVerifyAllIsKnown && not (areAttributesKnown addrAttributes))` which is probably a bit more understandable.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -127,28 +125,18 @@ verifyConsistency inputs witnesses
...
+            throwError $ ToilInvalidOutputs i (TxOutUnknownAttributes addr)
+        unless (vtcVerifyAllIsKnown && not (isUnknownAddressType addr)) $
```
gromakovsky: Same as above, it should be `when`, not `unless`.

txp/Pos/Txp/Toil/Utxo/Functions.hs
```diff
@@ -147,28 +145,25 @@ verifyConsistency inputs witnesses
     errFmt = ("length of inputs != length of witnesses "%"("%int%" != "%int%")")
     errMsg = sformat errFmt (length inputs) (length witnesses)
 
-verifyOutputs :: VTxContext -> TxAux -> VerificationRes
+verifyOutputs :: 
+       MonadError ToilVerFailure m 
+    => VTxContext
+    -> TxAux 
+    -> m ()
```
gromakovsky: Oh, previous version (with concrete `Either` monad instead of abstract `m` monad) was better.
Please see [exception handling guidelines](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/exceptions.md) and [monadic effects guidelines](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/monads.md).
Two points:
1. We don't write code in abstract `m` which is `MonadError`, because this code may be potentially impure and then there will be two ways of throwing exceptions.
2. We write [pure code](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/monads.md#pure-code) in concrete monads.

txp/Pos/Txp/Toil/Failure.hs
```diff
@@ -39,7 +41,7 @@ data ToilVerFailure
     | ToilOutGreaterThanIn { tInputSum  :: !Integer
                            , tOutputSum :: !Integer}
     | ToilInconsistentTxAux !Text
-    | ToilInvalidOutputs !Text  -- [CSL-1628] TODO: make it more informative
+    | ToilInvalidOutputs !Word32 !TxOutVerFailure
```
gromakovsky: Probably it makes sense to rename it to `ToilInvalidOutput`, because only one output is invalid in this case :)

txp/Pos/Txp/Toil/Utxo/Functions.hs
mitchellvitez: Thanks, I was trying to copy other `verify` functions but you're right, the previous (concrete) version is better

PR #2654, opened 2018-03-09
[CSL-2072] Delegation tweaks

volhovm:
This is a delegation-related refactoring of CSL-2072.


block/src/Pos/Block/BHelpers.hs
```diff
@@ -99,7 +98,6 @@ verifyMainBlock block@UnsafeGenericBlock {..} = do
         verifySscPayload
             (Right (Some _gbHeader))
             (_mbSscPayload _gbBody)
-    dlgVerifyPayload (_gbHeader ^. epochIndexL) (_mbDlgPayload _gbBody)
```
gromakovsky: Where is it done now?

volhovm: `dlgVerifyPayload` was doing only one check (verifying that psk's epoch index matches current epoch), which was duplicated. The other (now the only) copy is in `dlgVerifyPskHeavy`.

PR #2655, opened 2018-03-09
Restore top level README

erikd:
This used to be symlink to the file `lib/README.md` so when that changed so did this.

The two files are now different.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,31 @@
-lib/README.md
+# Cardano SL
+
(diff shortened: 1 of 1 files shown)


erikd:
The CI test failure is a Mac OS X specific linker issue: https://ghc.haskell.org/trac/ghc/ticket/14444

erikd:
Thanks @gromakovsky !

PR #2656, opened 2018-03-10
[CSL-2347] Add logging of servant requests in wallet new

Martoon-00:
This is attempt to use same logging mechanism as one we have in old wallet.

#### Previously

Old logging was implemented following such principles:
* Automatic logging - no need to print anything explicitly in implementation.
* No secret info should be revealed (like password) - thus logging requires `instance Buildable` for all types used in API, they should mask content where appropriate.
* No sensitive info should go to public logs (like address) - thus logging in addition requires `instance Buildable (SecureLog ...)` for your type, which masks potentially de-anonymazing info.
* Parameters of request are logged to secret/public logs with `Buildable / Buildable (SecureLog)` instances accordingly. However, responses are not logged at all into public logs (we didn't have time to solve this appropriately previously).

You can see how it looked like e.g. here #1874.

#### This pull request

First, I performed a couple of refactorings:
* Introduce `class BuildableSafeGen` which facilitates simultaneous creation of similar `instance Buildable` for secret logs and `instance Buildable (SecureLog ...)` for public logs via template haskell.
* Modify existing servant type-level combinators from `Pos.Util.Servant`, to work not only with types of form `apiType a` (like `Capture "name" a`), but with arbitrary one.
It will play its role if one day someone needs unusual types like `QueryFlag "name"` to appear in API.

Also logging is implemented for wallet-new-specific servant combinators (`Tags`, `FilterBy`, `SortBy`).
I expect resulting logs to look like
```
Request #1
GET
    :> api
    :> v1
    :> transactions
    :> walletId: <id>
    :> filters (id, balance): [FilterByRange, FilterIdentity]
    :> sort: balance desc
...
Response #1
Status OK > [<transactions here>]
```

There is inconsistency in filters and sorts representation. To artist within me, logs for sorting look good, but logs for filtering are cumbersome (numerous `FilterIdentity`s are not cut away) and incomplete (knowing that filtering mode is `FilterByRange` is good, but which exactly range do we sort by?) but making similar thing as for sorting would be a bit hard and require some arguable decisions (add `Buildable ix` [here](https://goo.gl/PN2ewz)?), see implementation for details. Generally I'd like this to be discussed and caught up in separate PR (unless one finds the inconsistency too dramatic).

**What remains:** PR is WIP, to my knowledge it only requires to go through `Cardano.Wallet.API.V1.Types` and to define `instance Buildable` and `instance Buildable (SecureLog ..)` for every type mentioned in API - I have just started over this process. For types which serve as response slightly another instance is required, follow compiler. I didn't manage to do this part within this week.

Also I suppose I took many arguable decisions here. Including one about `class BuildableSafeGen` introduction - it should be approved before work on this PR can be continued.

### DIFF
--- a/wallet-new/spec/swagger.json
+++ b/wallet-new/spec/swagger.json
@@ -8,16 +8,9 @@
             "url": "http://mit.com",
             "name": "MIT"
         },
-        "description": "This is the specification for the Cardano Wallet API, automatically generated\nas a [Swagger](https://swagger.io/) spec from the [Servant](http://haskell-servant.readthedocs.io/en/stable/) API\nof [Cardano](https://github.com/input-output-hk/cardano-sl).\n\nGit revision: **6f69e2e0033d51ff803630469b5749f85a3e2228**\nSoftware version: **cardano-sl:0**\n\n\n## Request format (all versions)\n\nIssuing requests against this API is conceptually not very different from any other web service out there. The API\nis **versioned**, meaning that is possible to access different versions of the API by adding the _version number_\nin the URL.\n\n**For the sake of backward compatibility, we expose the legacy version of the API, available simply as\nunversioned endpoints.**\n\nThis means that _omitting_ the version number would call the old version of the API. Examples:\n\n```\n/api/version\n```\n\nCompatibility between major versions is not _guaranteed_, i.e. the request & response formats might differ.\n\n## Response format (V1 onwards)\n\n**All GET requests of the API are paginated by default**. Whilst this can be a source of surprise, is\nthe best way of ensuring the performance of GET requests is not affected by the size of the data storage.\n\nVersion `V1` introduced a different way of requesting information to the API. In particular, GET requests\nwhich returns a _collection_ (i.e. typically a JSON array of resources) lists extra parameters which can be\nused to modify the shape of the response. In particular, those are:\n\n* `page`: (Default value: **1**).\n* `per_page`: (Default value: **10**)\n\nFor a more accurate description, see the section `Parameters` of each GET request, but as a brief overview\nthe first two control how many results and which results to access in a paginated request.\n\nThis is an example of a typical (successful) response from the API:\n\n``` json\n{\n    \"status\": \"success\",\n    \"data\": [\n        {\n            \"amount\": 29208924123228353,\n            \"addresses\": [\n                \"5FCjkr138i9qs2EEjvmXF37aSLiZdwQsALPMmvVDgx8PQ31nJTh9p35ZfnDFn6hfWdyY6XuFQXY23kfUjSuwBSEd4912knFgJSSZSBCt2yo\"\n            ],\n            \"name\": \"My account\",\n            \"walletId\": \"J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg\",\n            \"index\": 1\n        },\n        {\n            \"amount\": 28089727180413548,\n            \"addresses\": [\n                \"7J1MaQmKydNUyekmvgR3VNZLeLYkTNpDTKz4Z5SHfnJxDVgCN3tjixySxMNKNULKaV5XBBngc1e5BNzrePahs1Tuo7xcnKUyvkUm8HNcHuufC5q\",\n                \"7fiTbReGezWKbx8zhrjdMekrd8kziXqofdzzGPHWKgRXd24pDM7TZSFfnarpCsHy67ZW3AgFRAfFEGuRSd1VSYP5dDyNJJLwKB29vxsFnGq9GPB8x2RittGrrjNXPpgB8b4Y3yBFbpcXeDF7FG13u5gFHrcXwU86MN2CaCpLaaynjRvnitwGnd\"\n            ],\n            \"name\": \"My account\",\n            \"walletId\": \"J7rQqaLLHBFPrgJXwpktaMB1B1kQBXAyc2uRSfRPzNVGiv6TdxBzkPNBUWysZZZdhFG9gRy3sQFfX5wfpLbi4XTFGFxTg\",\n            \"index\": 0\n        }\n    ],\n    \"meta\": {\n        \"pagination\": {\n            \"page\": 1,\n            \"perPage\": 46,\n            \"totalPages\": 1,\n            \"totalEntries\": 2\n        }\n    }\n}\n```\n\n## Filtering and sorting\n\n`GET` endpoints which list collection of resources supports filters & sort operations, which are clearly marked\nin the swagger docs with the `FILTER` or `SORT` labels. The query format is quite simple, and it goes this way:\n\n### Filter operators\n\n| Operator | Description                                                               | Example                |\n|----------|---------------------------------------------------------------------------|------------------------|\n| -        | If **no operator** is passed, this is equivalent to `EQ` (see below).     | `balance=10`           |\n| `EQ`     | Retrieves the resources with index _equal to the one provided.            | `balance=EQ[10]`       |\n| `LT`     | Retrieves the resources with index _less than_ the one provided.          | `balance=LT[10]`       |\n| `LTE`    | Retrieves the resources with index _less than equal_ the one provided.    | `balance=LTE[10]`      |\n| `GT`     | Retrieves the resources with index _greater than_ the one provided.       | `balance=GT[10]`       |\n| `GTE`    | Retrieves the resources with index _greater than equal_ the one provided. | `balance=GTE[10]`      |\n| `RANGE`  | Retrieves the resources with index _within the inclusive range_ [k,k].    | `balance=RANGE[10,20]` |\n\n### Sort operators\n\n| Operator | Description                                                               | Example                |\n|----------|---------------------------------------------------------------------------|------------------------|\n| `ASC`    | Sorts the resources with the given index in _ascending_ order.            | `sort_by=ASC[balance]` |\n| `DES`    | Sorts the resources with the given index in _descending_ order.           | `sort_by=DES[balance]` |\n| -        | If **no operator** is passed, this is equivalent to `DES` (see above).    | `sort_by=balance`      |\n\n## Dealing with errors (V1 onwards)\n\nIn case a request cannot be served by the API, a non-2xx HTTP response will be issue, together with a\n[JSend-compliant](https://labs.omniti.com/labs/jsend) JSON Object describing the error in detail together\nwith a numeric error code which can be used by API consumers to implement proper error handling in their\napplication. For example, here's a typical error which might be issued:\n\n``` json\n{\n    \"status\": \"error\",\n    \"diagnostic\": {},\n    \"message\": \"WalletNotFound\"\n}\n```\n\n### Existing wallet errors\n\nError Name|HTTP Error code|Example\n---|---|---\n`NotEnoughMoney`|-|-\n`OutputIsRedeem`|-|-\n`SomeOtherError`|-|-\n`MigrationFailed`|-|-\n`JSONValidationFailed`|-|-\n`UnkownError`|-|-\n`WalletNotFound`|-|-\n"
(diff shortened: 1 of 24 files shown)


KtorZ:
@adinapoli-iohk @parsonsmatt 

Here you go! PR updated and re-sync with develop, completed with missing `Buildable` instances and Servant magic. 
I have tried to follow the style and steps of @Martoon-00 but haven't tested much anything yet; I assumed it'd work but I'll definitely spend some time tomorrow to verify that statement :)

parsonsmatt:
To fix the `stack2nix` build failure, run `./pkgs/generate.sh`. It'll take a really long time if it's the first time. Alternatively, copy the diff that the CI failure gives you.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -214,8 +216,12 @@ instance Bounded a => Bounded (V1 a) where
 instance Buildable a => Buildable (V1 a) where
     build (V1 x) = bprint build x
 
+instance Buildable a => Buildable [V1 a] where
+    build = bprint listJson
+
 instance Buildable (SecureLog a) => Buildable (SecureLog (V1 a)) where
-    build (SecureLog vx) = bprint build (SecureLog vx)
+    build = bprint build
```
Martoon-00: Let me resurrect for a while and note, that your rewrite reveals me writing a nonsense here.
With same grade of success I could write `build = build` :smile:

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -145,10 +146,12 @@ instance Show (FilterOperation ix a) where
     show (FilterByRange _ _)          = "FilterByRange"
     show                              = formatToString build
 
+deriveSafeBuildableExt $ \v -> [t| FilterOperation $v $v |]
```
Martoon-00: Looks much better right before `BuildableSafeGen` instance :heart:

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -195,6 +222,21 @@ instance ( HasServer subApi ctx
...
 
+-- | Defines name of @FilterBy syms res@ as sum of @syms@,
+-- and specifies parameter 'FilterBy' provides.
+-- Used in e.g. logging.
+instance (KnownSymbols syms, syms ~ ParamNames res params) => ApiHasArgClass (FilterBy params res) where
+    type ApiArg (FilterBy params res) = FilterOperations res
+
+    apiArgName _ =
+        let filterNames = mconcat . intersperse ", " $ symbolVals (Proxy @syms)
+        in  formatToString ("filters: ("%build%")") filterNames
```
parsonsmatt: This will log all of the names of the parameters that *could* be filtered on, not the actual filtering values. Is this what is wanted?

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -195,6 +222,21 @@ instance ( HasServer subApi ctx
...
+
+-- | Defines how 'FilterBy' is logged by just refering to
+-- 'instance Buildable SortOperations'.
```
parsonsmatt: Copy/paste wasn't fixed :)

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: I wonder how it even compiles actually. Could it be some `UndecidableInstances` magic of something that will blow up at runtime :o ?

wallet-new/src/Cardano/Wallet/API/Request/Sort.hs
```diff
@@ -183,6 +178,47 @@ instance ( HasServer subApi ctx
...
+--
+-- Example: @filter: asc id, desc balance@.
+--
+-- Default implementation of this instance for 'SortBy params res'
+-- (note OVERLAPPING) would print logs in format of
+-- "<some name dependant on 'syms'>: <prettified SortOperations>".
+-- With current implementation of 'SortOperations' the best logs we can get
+-- would be like "filter: asc, desc, none",
+-- or "filter (id, created_at, balance): asc, desc, none",
+-- because 'SortOperations' itself doesn't remember name of parameter it sorts
```
parsonsmatt: It does, actually :) In the GADT constructor, we stuff a bunch of constraints. One of them is `KnownSymbol (IndexToQueryParam res ix)`, so we can unpack the `SortOp` constructor and access that type:

```haskell
foo :: SortOperations a -> [String]
foo NoSorts = []
foo (SortOp (sop :: SortOperation ix a) rest) =
  symbolVal (Proxy :: Proxy (IndexToQueryParam a ix)) : foo rest
```

Extracts the list of values that were passed. Eg:

```haskell
"?sort_by=id&sort_by=ASC[timestamp]"
-- parses into
parsed = SortByIndex SortDesc (Proxy :: Proxy WalletId)
  `SortOp` SortByIndex SortAsc (Proxy :: Proxy (V1 Core.Timestamp))
  `SortOp` NoSorts

>>> print (foo parsed)
["id", "timestamp"]
```

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: It will probably just loop.

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
parsonsmatt: This looks like it'll make an instance like:

```haskell
instance SafeBuildable (FilterOperation t t)
```

which won't ever be satisfied :\

Martoon-00: `v` is not a pure value here (surprisingly), but is `newName "x" :: Q Type` (approximately).
I believe that those `$v` will result in different types.

wallet-new/src/Cardano/Wallet/API/Request/Sort.hs
Martoon-00: Oh, good point.
Then I agree with all points above, perhaps it can be implemented without `OVERLAPPING`, just defining `instance ApiHasArgClass` with `apiArgName = "sort"` and `instance ApiCanLogArg` with `toLogParamInfo` printing parameter names and corresponding values, similarly to current logic here https://goo.gl/AvytVJ - or something like that.
And same thing for filtering. Should be cool eventually :)

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
KtorZ: Not sure what you mean here. Not sure what this comment means either actually ^^"...

wallet-new/src/Cardano/Wallet/API/Request/Sort.hs
KtorZ: I believe this comment _was_ true before I rebased on top the latest work done by @parsonsmatt which actually changes the way `FilterBy` and `SortBy` were declared (from `[Symbol]` to `[*]`), allowing more information to be embedded at the type-level.

Martoon-00: I say :tada: then ))

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
Martoon-00: I suppose @parsonsmatt says about `s/SortOperations/FilterOperations`.

I meant here that `toLogParamInfo` does just ~`sformat build` inside, it's really expressed askew though :thinking:

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
KtorZ: I hardly think so. I believe we should log here the filters that are actually submitted with the query. I fail at seeing the value this brings otherwise.

KtorZ:
So, I've spent a bit more time looking at these `FilterBy` & `SortBy` types and their logging. I've completely revised the implementation there to rely more on the `Buildable` instances and avoid the `Overlapping` instance. I did some cleanup in the meantime, removing what was unnecessary and possibly introduced by merges.

It also solves the issue where filters where wrongly logged (available filters were logged, not actual filters submitted). 
Also, the old Warp logger was still there, it's gone now :) 

Log output now looks like this:

```
[wallet.api:INFO:ThreadId 3181] [2018-03-23 17:34:27.50 UTC] 
GET Request #1
    :> api
    :> v1
    :> node-info

[wallet.api:INFO:ThreadId 3181] [2018-03-23 17:34:27.53 UTC] 
    Response #1 OK 0.03477s > 
	status=success
	meta={ pagination=1/1 total=1 per_page=1 }
	data={ syncProgress=0% blockchainHeight= localBlockchainHeight=0 blocks localTimeDifference=0μs }
```

---

```
GET Request #2
    :> api
    :> v1
    :> wallets
    :> pagination: page=1, per_page=10
    :> filter_by: balance=RANGE[100 coin(s),200 coin(s)]
    :> sort_by: ASC[balance]

[wallet.api:INFO:ThreadId 3206] [2018-03-23 17:38:54.41 UTC] 
    Response #2 OK 0.000709s > 
	status=success
	meta={ pagination=1/1 total=0 per_page=10 }
	data=[]
```

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -195,6 +229,15 @@ instance ( HasServer subApi ctx
...
         in route (Proxy :: Proxy subApi) context delayed
 
+-- | Defines name of @FilterBy syms res@
+instance ApiHasArgClass (FilterBy params res) where
+    type ApiArg (FilterBy params res) = FilterOperations res
+    apiArgName _ = "filter_by"
+
+-- | Defines how 'FilterBy' is logged by just refering to instance Buildable FilterOperations.
+instance ApiCanLogArg (FilterBy params res) where
+    toLogParamInfo _ param = \sl -> sformat (buildSafe sl) param
```
parsonsmatt: :heart_eyes:

wallet-new/spec/swagger.json
```diff
@@ -8,16 +8,9 @@
...
...---------------------------------------------------------------|------------------------|\n| `ASC`    | Sorts the resources with the given index in _ascending_ order.            | `sort_by=ASC[balance]` |\n| `DES`    | Sorts the resources with the given index in _descending_ order.           | `sort_by=DES[balance]` |\n| -        | If **no operator** is passed, this is equivalent to `DES` (see above).    | `sort_by=balance`      |\n\n## Dealing with errors (V1 onwards)\n\nIn case a request cannot be served by the API, a non-2xx HTTP response will be issue, together with a\n[JSend-compliant](https://labs.omniti.com/labs/jsend) JSON Object describing the error in detail together\nwith a numeric error code which can be used by API consumers to implement proper error handling in their\napplication. For example, here's a typical error which might be issued:\n\n``` json\n{\n    \"status\": \"error\",\n    \"diagnostic\": {},\n    \"message\": \"WalletNotFound\"\n}\n```\n\n### Existing wallet errors\n\nError Name|HTTP Error code|Example\n---|---|---\n`NotEnoughMoney`|-|-\n`OutputIsRedeem`|-|-\n`SomeOtherError`|-|-\n`MigrationFailed`|-|-\n`JSONValidationFailed`|-|-\n`WalletNotFound`|-|-\n"
```
adinapoli-iohk: Where we expecting changes of `spec/swagger.json` as part of this PR?

wallet-new/src/Cardano/Wallet/API.hs
```diff
@@ -49,7 +51,7 @@ type DevAPI = "api" :> "development" :> Dev.API
 devAPI :: Proxy DevAPI
 devAPI = Proxy
 
-type WalletAPI = V0Doc :<|> V1Doc :<|> V0API :<|> V1API
+type WalletAPI = V0Doc :<|> V1Doc :<|> LoggingApi WalletLoggingConfig (V0API :<|> V1API)
```
adinapoli-iohk: At least that's not too invasive when it comes the time to hook it in 😀 

Probably not needed now, but in the future we might think about renaming this one to better reflect that its purpose is not just the one of logging, but logging information securely. For now let's not bother as there might be further changes we do to this part of the code.

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -35,15 +41,21 @@ import           Cardano.Wallet.API.V1.Types
 -- the inner closure of 'FilterOp'.
 data FilterOperations a where
     NoFilters  :: FilterOperations a
-    FilterOp   :: (IndexRelation a ix, FromHttpApiData ix, ToHttpApiData ix, Eq ix)
+    FilterOp   :: ( IndexRelation a ix
+                  , KnownSymbol (IndexToQueryParam a ix)
+                  , BuildableSafe ix
+                  , FromHttpApiData ix
+                  , ToHttpApiData ix
+                  , Eq ix
```
adinapoli-iohk: Hopefully that's the last bunch of constraints we need to add, this is rapidly turning into a kitchen sink 😁 :trollface:

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -376,8 +388,14 @@ instance ToSchema (V1 Core.Timestamp) where
 -- base16-encoded string.
 type SpendingPassword = V1 Core.PassPhrase
 
+instance Monoid (V1 Core.PassPhrase) where
+    mempty = V1 mempty
+    mappend (V1 a) (V1 b) = V1 (a `mappend` b)
```
adinapoli-iohk: Why do we need to `mappend` two `PassPhrase`(s) together?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -687,6 +849,19 @@ instance Arbitrary PaymentDistribution where
...
+instance BuildableSafeGen PaymentDistribution where
+    buildSafeGen sl PaymentDistribution{..} = bprint ("{"
+        %" address="%buildSafe sl
+        %" amount="%buildSafe sl
+        %" }")
+        pdAddress
+        pdAmount
+
+-- instance Buildable [PaymentDistribution] where
+--     build = bprint listJson
```
adinapoli-iohk: Should we delete this commented-out code?

wallet-new/spec/swagger.json
KtorZ: Nope nope nope... NOPE! I think I was careful not to commit any changes related to `swagger.json` yet this may come from a previous commit on the branch. I believe it doesn't matter much but will conflict with #2696. Once #2696 is merged though, we have to be careful not introducing `swagger.json` back again!

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: Whoops. Will do.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: I haven't seen a case where we need the SemiGroup part of it, but we *require* `mempty` in a couple of places. I use it as a *default value* to get rid of the `Maybe PassPhrase` we have sometimes when securely logging those phasephrase. In the end, the value doesn't matter much and could be anything since we display a placeholder `<passphrase>` anyway. 

Though, I noticed a function `emptyPassphrase` in one of the core type and I wondered why not using well-known abstractions instead (like Monoid).

KtorZ:
*crosses his fingers, bends the knee and prays the CI gods stronger than ever*

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
adinapoli-iohk: Fair enough. I still have in the back of my mind that concatenating two `PassPhrase`(s) makes no sense at all and thus it shouldn't be allowed, but I don't think that's the end of the world, as the type has a well understood semantic.

PR #2658, opened 2018-03-11
[CSL-2342] Implement client for the wallet v1 api

parsonsmatt:
This PR implements a client for the V1 API interface using servant-client.


wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -65,6 +77,13 @@ data FilterOperation ix a =
     | FilterIdentity
```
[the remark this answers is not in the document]
adinapoli-iohk: If I recall correctly I used it as a "base case" to stop the recursion, but if we can achieve the same without it, let's remove it 😉

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -779,6 +789,12 @@ instance ToJSON (V1 Core.SoftwareVersion) where
                , "version" .=  toJSON svNumber
                ]
 
+instance FromJSON (V1 Core.SoftwareVersion) where
```
adinapoli-iohk: Well spotted: before the introduction of the `V1` type I had to step into the landmine of not supporting some `aeson` instance due to the fact some `Core` types had a different JSON representation which was making handling it in V1 clunky.

wallet-new/test/MarshallingSpec.hs
```diff
@@ -29,13 +29,16 @@ spec = describe "Marshalling & Unmarshalling" $ do
         aesonRoundtripProp @(V1 BackupPhrase) Proxy
         aesonRoundtripProp @Account Proxy
         aesonRoundtripProp @AssuranceLevel Proxy
+        aesonRoundtripProp @(V1 Core.SoftwareVersion) Proxy
```
adinapoli-iohk: yay!

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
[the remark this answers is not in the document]
parsonsmatt: Yeah, we have a `NoFilters` constructor for that I think :)

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -89,6 +89,8 @@ import           Cardano.Wallet.Orphans.Aeson ()
 -- V0 logic
 import           Pos.Util.BackupPhrase (BackupPhrase (..))
 
+-- importing for orphan instances for Coin
+import Pos.Wallet.Web.ClientTypes.Instances ()
```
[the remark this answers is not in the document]
parsonsmatt: I agree with that sentiment! But now we are using `wallet` as a library, where before it was an application, and orphan instances bit me in the development of this :) "Overlapping instance declarations for `HttpApiData SomeThing`", oh no, time to grep...

I'm OK with orphans but I really think that it should be gated such that it's impossible to have a type in scope *without* it's instance, which can be tricky to do. If you define the type it's easier to just put the instance next to it.

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
```diff
@@ -8,19 +8,26 @@ import           Cardano.Wallet.API.V1.Types
...
-    :<|> "accounts" :> Capture "accountIndex" AccountIndex
-                    :> Summary "Update an Account for the given Wallet."
-                    :> ReqBody '[ValidJSON] (Update Account)
-                    :> Put '[ValidJSON] (WalletResponse Account)
+    = "accounts" :> (
+        Capture "walletId" WalletId
+          :> Capture "accountIndex" AccountIndex
+          :> Summary "Deletes an Account."
+          :> DeleteNoContent '[ValidJSON] NoContent
+        :<|> Capture "walletId" WalletId
```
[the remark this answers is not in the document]
parsonsmatt: Good call! `type CaptureWalletId = Capture "walletId" WalletId` DRYs things up nicely and removes a potential point of error :smile:

parsonsmatt:
I am pretty confident that the API changes won't affect the way documentation is generated or tags are distributed. Happy to wait on @KtorZ's input though :)

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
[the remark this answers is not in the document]
adinapoli-iohk: Definitely, I'm with you! I also like to define my types and all the related instances in one single place, it reduces the burden of thinking "oh dear, which are the instances this type implements"? Unfortunately there are multiple school of thoughts 😁

adinapoli-iohk:
Right, we seems to be converging 😀 @parsonsmatt I think that in order for us to merge this PR we should:

* Address the review changes (the one to DRY up the Capture params seems a nice one judging from your reaction);

* Make the CI fail (ignore anything Serokell-related as that's an expected failure, but the `Appveyor` failure is not, alas);

* Let's press the merge button :)

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Wallets.hs
```diff
@@ -23,15 +23,14 @@ handlers :: ( HasConfigurations
...
-                :<|> Accounts.handlers walletId
-             )
+handlers = newWallet
+    :<|> listWallets
+    :<|> ( updatePassword
+        :<|> deleteWallet
+        :<|> getWallet
+        :<|> updateWallet
+        :<|> Accounts.handlers
+    )
```
KtorZ: I've moved the Account handlers out, to be virtually separated from the wallet handlers in order to avoid nesting tags. 
Expect a merge-conflict here with CBR-102. Nothing hard to resolve though.

parsonsmatt:
Oh, dear. The implementation of the FilterBy and SortBy stuff was a lot hairier than I thought :(

wallet-new/src/Cardano/Wallet/API/Request/Sort.hs
```diff
@@ -87,25 +116,20 @@ instance ( Indexable' a
...
+        NoSorts
+    toSortOperations req (key:xs) _ =
+        foldr (either (flip const) SortOp) rest
+            . map (parseSortOperation (Proxy @a) (Proxy @ix) . (,) key)
+            . mapMaybe snd
+            . filter (("sort_by" ==) . fst)
+            . parseQueryText
+            $ rawQueryString req
+      where
+        rest = toSortOperations req xs (Proxy @ ixs)
```
parsonsmatt: The previous implementation had a bug, where multiple `sort_by` in the query would be failed:

```haskell
/foo?sort_by=bar&sort_by=baz
```

The `parseQueryText` would turn that into `[("sort_by", Just "bar"), ("sort_by", Just "baz")]`. When you call `List.lookup`, that would find `"bar"`. If `"bar"` does not match the index, then it is omitted. However, `"bar"` will *always* be what is looked up!

This implementation finds all query parameters that have `"sort_by"` at first, and then attempts to parse all of them into a sort operation. If any succeed, then they are added to the `SortOps`. This allows multiple sorts in a resource.

parsonsmatt:
I wrote a blog post on flattening out a nested route in Servant, so that servers and clients are more ordinary: https://www.parsonsmatt.org/2018/03/14/servant_route_smooshing.html

Not sure if this should be incorporated here; it seems easier to have less magic.

wallet-new/src/Cardano/Wallet/API/Indices.hs
```diff
@@ -39,6 +41,12 @@ type family IndicesOf a :: [*] where
     IndicesOf Transaction = TransactionIxs
     IndicesOf Account     = AccountIxs
 
+-- | This type family imposes a bidirectional mapping between symbols and
+-- their index types. This allows you to go from a symbol to the type:
+type family StringToIndex sym = r | r -> sym  where
+    StringToIndex "balance" = Core.Coin
+    StringToIndex "wallet_id" = WalletId
```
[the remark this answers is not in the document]
adinapoli-iohk: @parsonsmatt Unfortunately I'm not sure this is gonna work out without indexing the type family via `r` as well. If we ever want to have overloaded params (something we can avoid but would still be nice), you will have clashes. For example, in this PR Konstantin is now introducing the `id` filter parameter to filter on `Transaction`s:

https://github.com/input-output-hk/cardano-sl/pull/2602/files#diff-8818902a6e1eac32f98a94cdd65d7048R10

One might argue we should fully qualify our query params (and that would be a perfectly reasonable answer), but it would be nice if we could simply say something like this:

```
GET /api/v1/wallets?id=Blabla
GET /api/v1/transactions?id=blabla
```

In the RESTful sense, as we are dealing with well-defined resources (in this example `Wallet` and `Transaction`) there is no ambiguity (at least imho) on which type of `id` we are referring to. But in this new brave world `StringToIndex` won't work, but this will:

```haskell
type family StringToIndex sym r = where
  StringToIndex "id" Wallet = WalletId
  StringToIndex "id" Transaction = TransactionId
```

I'm ready to be told it's a terrible idea, and you might be right 😁 What I think is most important, though, is to choose a schema and stick with it. If we are opting for "fully qualified" query params is in `resource` + `field_identifier` better have this added to the docs for all the developers to abide 😛

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -175,3 +207,19 @@ parseFilterOperation p Proxy txt = case parsePredicateQuery <|> parseIndexQuery
...
+    , HasClient m (DecomposeFilterBy res syms ixs next)
+    , HasClient m next
+    , KnownSymbols syms
+    , SOP.All (ToIndex res) ixs
+    )
+    => HasClient m (FilterBy syms res :> next) where
+    type Client m (FilterBy syms res :> next) =
+        Client m (DecomposeFilterBy res syms (FilterParams syms res) next)
+    clientWithRoute pm _ =
+        clientWithRoute pm (Proxy @(DecomposeFilterBy res syms (FilterParams syms res) next))
```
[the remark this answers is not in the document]
adinapoli-iohk: @parsonsmatt The only downside I see on this approach is that we are now leaking the fact we are using `ixset-typed` internally, which might not become true anymore as we progress along the rebuilding of the new data layer when the new `Kernel` lands. What about a different name for `IxFor`? Conceptually a seemingly silly change, but at least we keep this opaque enough for now 😁

wallet-new/src/Cardano/Wallet/API/Request/Sort.hs
adinapoli-iohk: > This allows multiple sorts in a resource.

@parsonsmatt That's neat! I think that sorting over multiple field is a nice property it would be nice to support. Good catch!

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
```diff
@@ -8,21 +8,27 @@ import           Cardano.Wallet.API.V1.Parameters
...
-                    :> Post '[ValidJSON] (WalletResponse Account)
-    :<|> "accounts" :> Capture "accountIndex" AccountIndex
-                    :> Summary "Update an Account for the given Wallet."
-                    :> ReqBody '[ValidJSON] (Update Account)
-                    :> Put '[ValidJSON] (WalletResponse Account)
-    )
+type API
+    = Tags '["Accounts"] :> "accounts" :> (
+        CaptureWalletId
+          :> Capture "accountIndex" AccountIndex
```
adinapoli-iohk: @parsonsmatt Shouldn't we factor out `Capture "accountIndex" AccountIndex` into its own type alias like we did for `CaptureWalletId`?

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
```diff
@@ -8,21 +8,27 @@ import           Cardano.Wallet.API.V1.Parameters
...
+        :<|> CaptureWalletId
+          :> Summary "Creates a new Account for the given Wallet."
+          :> ReqBody '[ValidJSON] (New Account)
+          :> Post '[ValidJSON] (WalletResponse Account)
+        :<|> CaptureWalletId
+          :> Capture "accountIndex" AccountIndex
+          :> Summary "Update an Account for the given Wallet."
+          :> ReqBody '[ValidJSON] (Update Account)
+          :> Put '[ValidJSON] (WalletResponse Account)
+      )
```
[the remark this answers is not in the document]
adinapoli-iohk: > but I'm not sure throwing more magic at the problem is a good idea :(

Good observation, although this seems to be such a general problem than even the King of Servant, mr @alpmestan, came up with something [similar](https://gist.github.com/alpmestan/f6bb5e31e6241d1a6de57625fb1df5bc) -- I'm wondering if we could borrow ideas from both worlds 😀

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Wallets.hs
```diff
@@ -23,18 +23,13 @@ handlers :: ( HasConfigurations
...
-        )
-        -- NOTE Separation between tags ("Wallets" vs "Accounts"), see Cardano.Wallet.API.V1.Wallets
-        :<|> (\walletId -> Accounts.handlers walletId)
+handlers = (newWallet
+    :<|> listWallets
+    :<|> updatePassword
+    :<|> deleteWallet
+    :<|> getWallet
+    :<|> updateWallet
+    ) :<|> Accounts.handlers
```
ksaric: Why not the simpler version without the parens as in https://github.com/input-output-hk/cardano-sl/commit/4216cdcba2612d95d98a90b014353d50040dcaf1#diff-a6de663b7f7ec63d491ede6773183ddfR32 ?

wallet-new/src/Cardano/Wallet/API/Indices.hs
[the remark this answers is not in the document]
parsonsmatt: Or, we make the tying explicit at the API definition level:

```haskell
data (sym :: Symbol) ?= (a :: *)

data FilterBy params resource

type Api
  = "wallets"
    :> FilterBy '[ "id" ?= WalletId, "balance" ?= Core.Coin ] Wallet
    :> Get '[JSON] [Wallet]
  :<|> "transactions"
    :> FilterBy '[ "id ?= TransactionId ] Transaction
    :> Get '[JSON] [Transaction] 
```

Then the relevant instances have both the `Symbol` and the the type available.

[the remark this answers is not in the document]
parsonsmatt: Fully qualified API names sounds awful, IMO, so you're right, this isn't a good design, except as a stopgap to get this implemented now.

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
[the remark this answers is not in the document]
parsonsmatt: Hmm, if we do the `?=` thing I commented earlier, then the type gets expanded from

```haskell
FilterBy '["foo" ?= Foo, "bar" ?= Bar] A :> ...
```

to

```haskell
type X' 
  = QueryParam "foo" (FilterOperation Foo A)
  :> QueryParam "bar" (FilterOperation Bar A)
  :> ...
```

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
parsonsmatt: Good call. I'll get that done.

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Wallets.hs
parsonsmatt: I want to say that some of the `Tags` or `Summary` stuff get weird with that. I'll try it out and see if it works.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1065,3 +1090,46 @@ type family New (original :: *) :: * where
...
+        ':$$: 'Text "that use the `"
+            ':<>: 'ShowType (?=)
+            ':<>: 'Text "' type constructor."
+        ':$$: 'Text "An example would be: "
+        ':$$: 'Text "    "
+            ':<>: 'ShowType '["hello" ?= Int, "world" ?= String]
+        ':$$: 'Text "Other types, like "
+            ':<>: 'ShowType ty
+            ':<>: 'Text " are incompatible with this function."
+        )
```
parsonsmatt: i love custom type error messages :smile:

wallet-new/src/Cardano/Wallet/API/Indices.hs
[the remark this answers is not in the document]
adinapoli-iohk: > Or, we make the tying explicit at the API definition level:

Sweet combinator! The only little problem might be we are now giving the developer the degree of freedom or reassociating the symbol `id` to `WalletId` or any other index, but I think this is totally fine as the type family we have on mapping symbols to `ixset` indices should catch the bug. I think I like the extra verbosity of specifying the type in the RHS of `?=` as it's the kind of good verbosity imho!

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Wallets.hs
adinapoli-iohk: Probably that's the reason -- I think @KtorZ had a similar problem whilst working on CBR-102.

parsonsmatt:
The AppVeyor build fails with:

> Build exceeded allowed resource quotas. Fix your build to consume less resources or contact AppVeyor support to request quotas increase.


:(

deepfire:
It looks like the S3-based CI caching tool is at fault.

Perhaps it consumes too much disk for the cache -- which should be pruned somehow -- or maybe it already has a cache size limiting knob?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
adinapoli-iohk: Yea, they are pretty great!

parsonsmatt:
come on appveyor pls

wallet-new/src/Cardano/Wallet/API/Request/Filter.hs
```diff
@@ -63,21 +89,64 @@ data FilterOperation ix a =
...
---   *    `balance=GT[10]`
---   *    `balance=RANGE[0,10]`
-data FilterBy (sym :: [Symbol]) (r :: *) deriving Typeable
+--
+-- The first type parameter is a type level list that pairs the query
+-- parameter string with the expected parsed type. The second type
+-- parameter describes the resource that is being filtered.
+--
+-- @
+-- 'FilterBy' '[ "id" ?= WalletId, "balance" ?= Coin ] Wallet
```
adinapoli-iohk: @parsonsmatt am I dreaming or this comment, now that we scrapped `?=` is not quite right? 😁

parsonsmatt: good catch! :smile: fixed

parsonsmatt:
The CI gods needed a bit of prodding :clap: Manually deleting the already-failing builds with pending OSX builds fixed it right up.

PR #2660, opened 2018-03-12
infra: Remove un-needed import

erikd:


erikd:
Looks like that AppVeyor build failed while uploading the final build artifact.

rvl:
The AppVeyor artefact upload failure is quite mysterious! We will note it. I am clicking the rebuild button.

erikd:
The other two failures are on OS X:
```
dyld: malformed mach-o: load commands size (32928) > 32768
```
which is an OS X  linker bug on Sierra and later: https://ghc.haskell.org/trac/ghc/ticket/14444

PR #2670, opened 2018-03-14
[CSL-2371] Write logical integration tests for wallet key import endpoint

akegalj:
Issue: https://iohk.myjetbrains.com/youtrack/issue/CSL-2371

I am not happy because I had to add `MonadMask` constraint into `WalletTestMode`. I am not sure how to avoid that. The idea was to create a temporary file with `withSystemTempFile` dump poor key content into that file and do tests with import key. 

Comments are welcome.

I am also not satisfied I had to pull logging functionality - as functions used for key dumping and key creation depend on logging. Logging can probably be lifted into `MonadThrow` but I didn't spend time on this (maybe I should?)

If it is ok-ish I can create an issue to resolve these ^ and move on - or fix it asap in this issue.

### DIFF
--- a/wallet-new/integration/Error.hs
+++ b/wallet-new/integration/Error.hs
@@ -36,6 +36,8 @@ data WalletTestError
     | LocalTransactionsDiffer [Transaction]
     | LocalTransactionMissing Transaction [Transaction]
 
+    | PoorWalletBalanceIsZero Wallet
+
     deriving (Show, Eq, Generic)
 
 
@@ -60,5 +62,5 @@ instance Buildable WalletTestError where
     build (InvalidTransactionState t)     = bprint ("Transaction state is invalid. Transaction - ("%stext%")") (show t)
     build (LocalTransactionsDiffer t)     = bprint ("Local transactions differs. Transactions - ("%stext%")") (show t)
     build (LocalTransactionMissing t ts)  = bprint ("Local transaction ("%stext%") missing from txs history ("%stext%")") (show t) (show ts)
-
+    build (PoorWalletBalanceIsZero a)  = bprint ("Generated wallet balance is zero. Wallet - ("%stext%")") (show a)
 
--- a/wallet-new/integration/Functions.hs
+++ b/wallet-new/integration/Functions.hs
@@ -27,9 +27,8 @@ import           Pos.Core.Configuration (genesisSecretsPoor)
 import           Pos.Core.Genesis (poorSecretToEncKey)
 import           Pos.Launcher (withConfigurations)
 import           Pos.Util.Filesystem (withSystemTempFile)
-import           Pos.Util.UserSecret (UserSecret, initializeUserSecret, mkGenesisWalletUserSecret,
-                                      takeUserSecret, usKeys, usPrimKey, usVss, usWallet,
-                                      writeUserSecretRelease)
+import           Pos.Util.UserSecret (initializeUserSecret, mkGenesisWalletUserSecret,
+                                      takeUserSecret, usKeys, usWallet, writeUserSecretRelease)
 import qualified Pos.Wallet.Web.ClientTypes.Types as V0
 
 
@@ -339,8 +338,8 @@ runAction wc ws GetTransaction  = do
         & actionsNum +~ 1
 runAction wc ws ImportPoorWallet = do
     poorSecret <- pickRandomElement =<< liftIO unsafePoorSecrets
-    liftIO $ usingLoggerName "integration-test" $ withSystemTempFile "importWalletKey.sk" $ \tempFp _ -> do
-        dumpPoorSecret tempFp poorSecret
+    withSystemTempFile "importWalletKey.sk" $ \tempFp _ -> do
+        liftIO $ usingLoggerName "integration-test" $ dumpPoorSecret tempFp poorSecret
         poorWallet <- respToRes $ importWalletKey wc tempFp
 
         checkInvariant
--- a/wallet-new/integration/Functions.hs
+++ b/wallet-new/integration/Functions.hs
@@ -7,14 +7,15 @@ module Functions where
 import           Universum
 
 import           Data.Coerce (coerce)
+import           Data.Default (def)
 import           Data.List.NonEmpty (fromList)
 
-import           Control.Lens ((+~))
+import           Control.Lens ((+~), (?~))
 import           Test.QuickCheck (Gen, arbitrary, elements, frequency, generate)
 
 import           Cardano.Wallet.API.Response (WalletResponse (..))
-import           Cardano.Wallet.API.V1.Types (Account (..), AddressValidity (..), AccountIndex, NewAccount (..),
-                                              NewAddress (..), Payment (..),
+import           Cardano.Wallet.API.V1.Types (Account (..), AccountIndex, AddressValidity (..),
+                                              NewAccount (..), NewAddress (..), Payment (..),
                                               PaymentDistribution (..), PaymentSource (..),
                                               Transaction (..), V1 (..), Wallet (..),
                                               WalletAddress (..), WalletId)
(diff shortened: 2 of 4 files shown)


wallet-new/integration/Types.hs
```diff
@@ -32,6 +32,7 @@ type WalletTestMode m =
     ( MonadIO m
     , MonadThrow m
     , MonadPlus m
+    , MonadMask m
```
ksaric: Why `MonadMask` here?

wallet-new/integration/Error.hs
```diff
@@ -60,5 +62,5 @@ instance Buildable WalletTestError where
     build (InvalidTransactionState t)     = bprint ("Transaction state is invalid. Transaction - ("%stext%")") (show t)
     build (LocalTransactionsDiffer t)     = bprint ("Local transactions differs. Transactions - ("%stext%")") (show t)
     build (LocalTransactionMissing t ts)  = bprint ("Local transaction ("%stext%") missing from txs history ("%stext%")") (show t) (show ts)
-
+    build (PoorWalletBalanceIsZero a)  = bprint ("Generated wallet balance is zero. Wallet - ("%stext%")") (show a)
```
ksaric: Spaces alignment :eyes:

wallet-new/integration/Functions.hs
```diff
@@ -326,6 +336,32 @@ runAction wc ws GetTransaction  = do
...
+
+        -- Modify wallet state accordingly.
+        pure $ ws
+            & wallets <>~ [poorWallet]
+            & actionsNum +~ 1
+  where
+    unsafePoorSecrets = usingLoggerName "integration-test" $ withConfigurations def $ pure $ fromMaybe (error "GeneratedSecrets are unknown") genesisSecretsPoor
+    dumpPoorSecret fp poorSec = let hdwSk = poorSecretToEncKey poorSec in
+        dumpUserSecret fp $
+        foldl' (.) identity [ usKeys %~ (hdwSk :)
```
ksaric: I really don't understand this. Is this actually a function composition in disguise?

wallet-new/integration/Types.hs
akegalj: See my description:
> I am not happy because I had to add MonadMask constraint into WalletTestMode. I am not sure how to avoid that. The idea was to create a temporary file with withSystemTempFile dump poor key content into that file and do tests with import key.

I would like to remove that. Maybe you can help me remove `MonadMask` here. I didn't figure out how to do it.

wallet-new/integration/Functions.hs
```diff
@@ -326,6 +336,32 @@ runAction wc ws GetTransaction  = do
     -- Modify wallet state accordingly.
     pure $ ws
         & actionsNum +~ 1
+runAction wc ws ImportPoorWallet = do
+    poorSecret <- pickRandomElement =<< liftIO unsafePoorSecrets
+    withSystemTempFile "importWalletKey.sk" $ \tempFp _ -> do
```
ksaric: This seems too complicated. What we need is to import a key to the wallet. 
If we are generating keys every time we run this function this might be too expensive.
We can generate the import wallet before this test, and that would be simpler.
I mean, we can have a single key to import and import it once before the test and that would be good enough.

wallet-new/integration/Types.hs
ksaric: This should do it - https://github.com/input-output-hk/cardano-sl/pull/2670#discussion_r175061591

wallet-new/integration/Functions.hs
akegalj: I took those bits of the code from `tools/src/keygen/Dump.hs` (as they are not exported)

Basically this part is mimicing some small functionality of `cardano-keygen` executable.

wallet-new/integration/Functions.hs
akegalj: Yes - my understanding what @parsonsmatt said here https://iohk.myjetbrains.com/youtrack/issue/CSL-2371#comment=93-19258 is exactly aligned with what you are proposing.

I will ditch this code and expect `FilePath` to keys on file system will be given (and known upfront) when we will have this service ready. 

My understanding is that we don't have mechanism so far to generate keys on system startup, ie on this step:
> Test suite boots up the server, with initial configuration options

so maybe I can generate keys as a part of `wallet-new/integration/Main.hs` when test suite is starting up. What do you think? 

EDIT:

On a second thought I believe what @parsonsmatt whanted to say there is to import wallet using V0 endpoint there - and then this wallet can be used for other integration tests. But I am not sure what he exactly meant (maybe you can help me clear the picture).

How I think we should implement it:
 * generate keys as a part of starting up integration test suite - in `wallet-new/integration/Main.hs`
 * when V1.importKey is implemented - add integration test that imports keys using `FilePath` of previously generated keys

Otherwise I am not sure why we need import key in integration tests at all atm. Can you help me explain the motivation @ksaric ?

ksaric: @akegalj The motivation is that we need money when we are running this integration test so we can send transactions around.

akegalj: Ok, in that case I will just do what I have described:
 * generate keys as a part of starting up integration test suite - in wallet-new/integration/Main.hs
 * when V1.importKey is implemented - add integration test that imports keys using FilePath of previously generated keys

I have wrongly assumed the motivation is to test V0.importKey that will eventually be in place. Ok - will fix it now

ksaric: Personally, I don't see anything wrong with this test (it will lead to more random states), but the key generation should be extracted outside it in any case.

parsonsmatt: > generate keys as a part of starting up integration test suite - in wallet-new/integration/Main.hs

:+1: 

> when V1.importKey is implemented - add integration test that imports keys using FilePath of previously generated keys

Sounds good!

> I have wrongly assumed the motivation is to test V0.importKey that will eventually be in place. Ok - will fix it now

To further clarify: the wallet must be initialized so that it has money we can spend with. *How* that happens should be kept out of the tests themselves -- as you've noted, we can do that in `main` before we actually start running the test suite. We can do an `http-client` request to the `V0` part of the API; or we can do something else (maybe there's a command line option to start the wallet with a key? I'm unfamiliar).

At some point, we'll have implemented the `v1/import-wallet` endpoint, at which point we can write a full integration test like this.

wallet-new/integration/Functions.hs
parsonsmatt: @ksaric Yes -- observe the expansion:

```haskell
foldl' (.) identity [f, g, h]
-- foldl' k z (x : xs) = foldl' k (k z x) xs
foldl' (.) (identity . f) [g, h]
foldl' (.) (identity . f . g) [h]
foldl' (.) (identity . f . g . h) []
-- foldl' k z [] = z
(identity . f . g . h)
```

wallet-new/src/Cardano/Wallet/Client.hs
```diff
@@ -58,6 +58,8 @@ data WalletClient m
     , getAddressValidity
          :: Text -> Resp m AddressValidity
     -- wallets endpoints
+    , importWalletKey
+         :: FilePath -> Resp m Wallet
```
parsonsmatt: let's remove this and leave that to a further ticket for implementing/testing this endpoint specifically

parsonsmatt:
@akegalj Upon further thought:

1. Let's merge #2615 and #2658 to get the initial testing logic incorporated.
2. We'll still need to write a bash script to actually run the test -- [I've created CSL-2398](https://iohk.myjetbrains.com/youtrack/issue/CSL-2398) to track actually running the tests in CI.
3. The proper place to initialize the server state will be in that test script.

The test script will look something like:

```
# fork off server node
stack exec -- cardano-node 
# some request to initialize the wallet with seed data
curl ... 
# run the tests
stack exec -- cardano-integration-tests # with relevant options
```

akegalj:
Closing in favour of https://github.com/input-output-hk/cardano-sl/pull/2670#issuecomment-374407445 .

PR #2671, opened 2018-03-14
[DEVOPS-673] Build Daedalus bridge with Nix

rvl:
Provides a `daedalus-bridge` derivation which can be used when building the Daedalus installer.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,6 +4,7 @@ in
 { system ? builtins.currentSystem
 , config ? {}
 , gitrev ? localLib.commitIdFromGitRepo ./.git
+, buildId ? null
 , pkgs ? (import (localLib.fetchNixPkgs) { inherit system config; })
 # profiling slows down performance by 50% so we don't enable it by default
 , enableProfiling ? false
@@ -111,5 +112,23 @@ let
       mainnetWallet = mkDocker { environment = "mainnet"; };
       stagingWallet = mkDocker { environment = "mainnet-staging"; };
     };
+
+    daedalus-bridge = pkgs.runCommand "cardano-daedalus-bridge" {} ''
+      # Generate daedalus-bridge
+      mkdir -p $out/bin $out/config
+      cd $out
+      ${optionalString (buildId != null) "echo ${buildId} > build-id"}
+      echo ${gitrev} > commit-id
+
+      cp ${./log-configs + "/daedalus.yaml"} config/log-config-prod.yaml
+      cp ${./lib}/configuration.yaml config
+      cp ${./lib}/*genesis*.json config
+      cp ${cardanoPkgs.cardano-sl-tools}/bin/cardano-launcher bin
+      cp ${cardanoPkgs.cardano-sl-wallet-new}/bin/cardano-node bin
+
+      # test that binaries exit with 0
+      ./bin/cardano-node --help > /dev/null
+      HOME=$TMP ./bin/cardano-launcher --help > /dev/null
+    '';
   };
 in cardanoPkgs // other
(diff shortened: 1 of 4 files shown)


scripts/ci/ci.sh
```diff
@@ -75,16 +61,8 @@ export XZ_OPT=-1
...
-
-echo "Packing up ${CARDANO_ARTIFACT} ..."
-APP_NAME=cardano-sl
-mkdir -p ${APP_NAME}
-tar cJf ${APP_NAME}/${CARDANO_ARTIFACT_FULL_NAME}.tar.xz daedalus/
-echo "Uploading.."
-buildkite-agent artifact upload ${APP_NAME}/${CARDANO_ARTIFACT_FULL_NAME}.tar.xz s3://${ARTIFACT_BUCKET} --job ${BUILDKITE_JOB_ID}
-echo "Done."
+
+# fixme: check if upload of explorer to bucket is even required
```
domenkozar: It's not :)

domenkozar:
Nice :)

rvl:
OK, removed the code to upload explorer.

Also added `daedalus-bridge` to the hydra jobset.

I just realised that the `buildId` argument is quite useless when we start using nix-build. We might need to rework `iohk-ops find-installers`.

appveyor.yml
```diff
@@ -107,13 +107,6 @@ test_script:
 
 after_test:
  - cd "%WORK_DIR%" # Get back to where cache-s3.exe is located
```
domenkozar: Why was caching removed? It's crucial, otherwise the build takes 80min instead of 15min :)

rvl: Yeah I will zap that commit later but for now I need a green build from AppVeyor to test the Daedalus side. The AppVeyor build cruelly fails in the cache upload step!

default.nix
```diff
@@ -111,5 +112,23 @@ let
       mainnetWallet = mkDocker { environment = "mainnet"; };
       stagingWallet = mkDocker { environment = "mainnet-staging"; };
     };
+
+    daedalus-bridge = pkgs.runCommand "cardano-daedalus-bridge" {} ''
+      # Generate daedalus-bridge
+      mkdir -p $out/bin $out/config
+      cd $out
+      ${if buildId == null then "" else "echo ${buildId} > build-id"}
```
cleverca22:
```
nix-repl> lib.optionalString true "foo"
"foo"

nix-repl> lib.optionalString false "foo"
""
```

default.nix
```diff
@@ -111,5 +112,23 @@ let
...
+      echo ${gitrev} > commit-id
+
+      cp ${./log-configs}/daedalus.yaml config/log-config-prod.yaml
+      cp ${./lib}/configuration.yaml config
+      cp ${./lib}/*genesis*.json config
```
cleverca22: i prefer doing `cp ${./lib + "/configuration.yaml"} config/configuration.yaml` whenever possible, so it doesnt import all of lib into the store
though the wildcards for genesis sort of defeat it in this case

also, my method fails at eval-time if the file is missing, while yours fails at build time, after you have already left because its doing an hour long compile

rvl: How would I handle the wildcards for genesis without copying `./lib` into the nix store?

cleverca22: thats fairly difficult to handle

rvl: :-) OK I have used the path append method for the log config file but copied the entire `./lib` subdirectory.

domenkozar:
@rvl I'm a bit confused here - windows part is missing?

rvl:
The windows part stays the same here - it just builds `CardanoSL.zip`. Then the Daedalus installer downloads it off AppVeyor in input-output-hk/daedalus#789.

PR #2673, opened 2018-03-15
[DEVOPS-670] Exchanges Documentation

disassembler:
This branch needs rebased with lots of commits squashed before it's ready to be merged. I'm opening this PR as a means to review the content of the documentation.


docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+that to /wallet. Note: if no volume is mounted to `/wallet` the container startup
+script will refused to execute `cardano-node` and the container will exit.
+
+The location of `/wallet` cannot be changed, but you can mount any kind of volume
+you want in that directory that docker supports.
+
+Note that if you give this a different name than is specified above, use the
+name you used for any future docker commands in examples in the document.
+
+# Migrating from V0 to V1 API
```
jmitchell: Please add "TODO".

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+# Usage FAQs
+
+## What are recommended hardware/software requirements for exchange wallets?
+
+RAM: 8 GB for building, 4 GB for running
+CPU: ???
```
jmitchell: "Modern x64" for now.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+Disk: SSD recommended of ??? size
```
jmitchell: 100 GB, with the option to allocate more space in coming years. Sound good, @cleverca22?

@disassembler, please add a FAQ about the expected log rotation/expunging behavior with the documented configuration.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+Operating System: NixOS or CentOS 7 recommended, although any linux distribution should work with nix package manager used for building the binaries
```
jmitchell: Have we tested both nix and docker setups with CentOS 7?

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+Software Requirements: Nix package manager for building standalone and docker containers. On systems using docker, docker > ??? required.
```
jmitchell: > docker > ??? required.

For now just indicate what version of Docker we've tested against.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+## How do I know when the wallet has fetched all the blocks?
+
+You can check the sync progress using the API to get detailed json output:
+
+    curl -X GET "https://127.0.0.1:8090/api/v1/node-info" -H "accept: application/json;charset=utf-8"
+    {"data":{"syncProgress":{"quantity":100,"unit":"percent"},"blockchainHeight":{"quantity":738268,"unit":"blocks"},"localBlockchainHeight":{"quantity":738268,"unit":"blocks"},"localTimeDifference":{"quantity":0,"unit":"microseconds"}},"status":"success","meta":{"pagination":{"totalPages":1,"page":1,"perPage":1,"totalEntries":1}}}
+
+The following command can be used to see the percentage completion of the sync only:
+
+    nix-shell -p jq curl --run 'curl -X GET "https://127.0.0.1:8090/api/v1/node-info" -H "accept: application/json;charset=utf-8" | jq .data.syncProgress.quantity'
```
jmitchell: Can you show sample output like you do for the previous command?

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+         -H "Accept: application/json; charset=utf-8"                      \
+         --cacert ./scripts/tls-files/ca.crt                               \
+         -d '{                                                             \
+      "operation": "create",                                               \
+      "backupPhrase": ["squirrel", "material", "silly", "twice", "direct", \
+        "slush", "pistol", "razor", "become", "junk", "kingdom", "flee"],  \
+      "assuranceLevel": "normal",                                          \
+      "name": "MyFirstWallet"                                              \
+    }'
+Warning: Those 12 mnemonic words given for the backup phrase act as an example. Do not use them on a production system. See the section below about mnemonic codes for more information.
```
jmitchell: Great, thank you for including the warning.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+As a response, the API provides you with a wallet id used in subsequent requests to uniquely identity the wallet. Make sure to store it / write it down. Note that every API response is jsend-compliant; Cardano also augments responses with meta-data specific to pagination. More details in the section below about Pagination.
```
jmitchell: > wallet id used in subsequent requests to uniquely identity the wallet

Slightly less wordy: "unique wallet ID to be used in subsequent requests"

jmitchell: > jsend-compliant

Can you link to reference material for `jsend`? I hadn't heard of it until now.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+        "meta": {
+            "pagination": {
+                "totalPages": 1,
+                "page": 1,
+                "perPage": 1,
+                "totalEntries": 1
+            }
+        }
+    }
+You have just created your first wallet. Information about this wallet can be retrieved using the GET /api/v1/wallets/{walletId} endpoint as follow:
```
jmitchell: s/follow/follows/

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+    curl -X GET https://localhost:8090/api/v1/wallets/{{walletId}} \
+         -H "Accept: application/json; charset=utf-8"              \
+         --cacert ./scripts/tls-files/ca.crt                       \
+
+## Receiving Money
+
+To receive money from other users you should provide your address. This address can be obtained from an account. Each wallet contains at least one account, you can think of account as a pocket inside of your wallet. Besides, you can view all existing accounts of a wallet by using the GET /api/v1/wallets/{{walletId}}/accounts endpoint as follow:
```
jmitchell: s/Besides, you can v/V/

jmitchell: Remove "as follow"

jmitchell: > Each wallet contains at least one account, you can think of account as a pocket inside of your wallet.

How about "Each wallet contains at least one account. An account is like a pocket inside your wallet."

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+            "pagination": {
+                "totalPages": 1,
+                "page": 1,
+                "perPage": 10,
+                "totalEntries": 1
+            }
+        }
+    }
+
+Each account has at least one address, all listed under the addresses field. You can communicate one of these addresses to receive money on the associated account.
```
jmitchell: > Each account has at least one address

This was already mentioned. How about "All the wallet's accounts are listed under the `addresses` field."?

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Sending Money
+
+In order to send money from one of your account to another address, you can create a new payment transaction using the POST /api/v1/transactions endpoint as follow:
```
jmitchell: s/one of your account/one of your accounts/

jmitchell: s/can/must/

s/as follow//

jmitchell: s/money/Ada/

It is of course a form of money, but we might as well be precise, especially with exchanges as our audience.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+    curl -X POST https://localhost:8090/api/v1/transactions \
+         -H "Content-Type: application/json; charset=utf-8" \
+         -H "Accept: application/json; charset=utf-8"       \
+         --cacert ./scripts/tls-files/ca.crt                \
+         -d '{                                              \
+      "destinations": [{                                    \
+        "amount": 14,                                       \
```
jmitchell: Indicate whether the units are Ada or Lovelace.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+        "address": "A7k5bz1QR2...Tx561NNmfF"                \
+      }],                                                   \
+      "source": {                                           \
+        "accountIndex": 0,                                  \
+        "walletId": "Ae2tdPwUPE...8V3AVTnqGZ"               \
+      }                                                     \
+    }'
+
+Note that, in order to perform a transaction, you need to have some existing coins on the source account! Beside, the Cardano API is designed to accomodate multiple recipients payments out-of-the-box; notice how destinations is a list of addresses.
```
jmitchell: s/some existing/enough/

s/Beside,//

> list of addresses

"and amounts"

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+When the transaction succeeds, funds are becomes unavailable from the sources addresses, and available to the destinations in a short delay. Note that, you can at any time see the status of your wallets by using the GET /api/v1/transactions/{{walletId}} endpoint as follow:
```
jmitchell: > funds are becomes unavailable from

funds are no longer available in

> available to the destinations in a short delay

are soon made available to the destinations

s/as follow//

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+    curl -X GET https://localhost:8090/api/v1/wallets/{{walletId}}?account_index=0  \
+         -H "Accept: application/json; charset=utf-8"                               \
+         --cacert ./scripts/tls-files/ca.crt                                        \
+
+We have here constrainted the request to a specific account, with our previous transaction the output should look roughly similar to this:
```
jmitchell: s/We have here/Here we/

Split this into two sentences.

s/with/after/

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+                "page": 1,
+                "perPage": 10,
+                "totalEntries": 1
+            }
+        }
+    }
+
+## Where can I find the API documentation?
+
+Run the latest wallet and go to <https://127.0.0.1:8090/docs/v1/index>.
```
jmitchell: Are the V0 docs no longer available anywhere? Until all exchanges are migrated to V1 we need to keep them available, while simultaneously discouraging their use for new exchanges.

jmitchell: Mention the `walletListen` configuration option in `./custom-wallet-config.nix`.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+## How can I inspect runtime metrics and statistics?
+
+Current metrics and stats are available at <http://127.0.0.1:8000/>.
```
jmitchell: Mention the `ekgListen` configuration option in `./custom-wallet-config.nix`.

docs/wip/exchange-onboarding.md
jmitchell: > More details in the section below about Pagination

I don't see it. Please add one.

jmitchell:
Please add `[DEVOPS-670]` as a prefix to the PR, and plan on squashing when you merge. In the future, try to remember to prefix the issue ID in the first line of all commit messages.

.gitignore
```diff
@@ -81,3 +81,14 @@ custom-wallet-config.nix
 # wallet web API benchmarking results
 wallet-new/bench/results/*.csv
 wallet-new/bench/results/*.txt
+
+# cardano-state-* for wallet data
+cardano-state-*
```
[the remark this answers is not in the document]
disassembler: our wiki page: https://github.com/input-output-hk/internal-documentation/wiki/Exchange-requirements

Happy to change it, but we should update our docs to not use that path as well if I do

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Backup local state
+
+Skip to the next section if this is target machine doesn't yet have
+`cardano-sl` set up.
+
+To avoid catastrophic data loss, stop the wallet and backup the
+local databases, keys, certificates, and logs. By default, the
+local state will be in `./state-wallet-mainnet`, but may be
+elsewhere (see `stateDir` attribute in `./custom-wallet-config.nix`).
```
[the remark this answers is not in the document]
disassembler: Had breakage. See the nix file changes for docker that resolved those.

docs/wip/exchange-onboarding.md
disassembler: completely agree, but should probably be fixed in API docs, this was a straight copy/paste from there.

docs/wip/exchange-onboarding.md
disassembler: same as above

docs/wip/exchange-onboarding.md
disassembler: same as above

docs/wip/exchange-onboarding.md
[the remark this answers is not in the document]
jmitchell: So all the custom attrs work now in this branch with Docker?

docs/wip/exchange-onboarding.md
jmitchell: Can you follow up on making sure the API docs are updated as well? If the source is in this repo, you can just include it in this PR.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+      * [How do I know when the wallet has fetched all the blocks?](#how-do-i-know-when-the-wallet-has-fetched-all-the-blocks)
+      * [Creating a New Wallet](#creating-a-new-wallet)
+      * [Receiving Money](#receiving-money)
+      * [Sending Money](#sending-money)
+      * [Where can I find the API documentation?](#where-can-i-find-the-api-documentation)
+      * [How can I inspect runtime metrics and statistics?](#how-can-i-inspect-runtime-metrics-and-statistics)
+
+# Communication
+
+* An e-mail list needs to be provided by the exchange for IOHK to broadcast issues and upddates.
```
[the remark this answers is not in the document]
disassembler: Agree, changing

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,352 @@
...
+* A slack room will be created for each exchange for direct communication with IOHK team.
```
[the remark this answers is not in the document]
disassembler: changed, reached to Tatyana

docs/wip/exchange-onboarding.md
jmitchell: Nevermind. I realize this is just using `jq` to filter the previous request.

docs/wip/exchange-onboarding.md
disassembler: I have not, just NixOS so far. I can spin up an ec2 CentOS 7 instance and validate though. Although, I will say docker on CentOS 7 is painful with the default device mapper storage backend. I've used btrfs in the past, but I know that's not a stable fs. Next version of centos//RHEL 7 I think is supposed to ship with overlay2.

docs/wip/exchange-onboarding.md
disassembler: thank @KtorZ

jmitchell:
Can we add at least one basic functional tests to CI for NixOS and/or CentOS 7?

Tests I'd eventually like to see. Don't aim to do all this now, we just need to focus on testing one of the leaves for now and make plans to improve coverage later.

- for each supported distro (currently NixOS and CentOS 7)
  - for each support build strategy (currently nix and docker), build and launch
    - `./custom-wallet-config.nix`
      - [ ] without it
      - [ ] as a copy of `./sample-wallet-config.nix` 
      - [ ] all supported attrs set to non-default values and verification that resulting connect script reflects those settings
  - checks once running
    - sane response to curling (HTTP 200 and reasonable body)
      - [ ] API docs endpoint
      - [ ] `/api/v1/node-info`
      - [ ] `/api/info` (a V0 endpoint)
      - [ ] eventually all the endpoints mentioned in the doc, including real staging test Ada transactions.
    - sanity checks of local state files
      - [ ] existence tests
      - [ ] important log lines (e.g. At one point the devs tried to get the git rev logged around startup. Last I recall it doesn't work in practice and we log something like "unknown" instead. It may even require calling an endpoint to cause a logging side-effect. This *should* work and we should have a test to verify it.)

.gitignore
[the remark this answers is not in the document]
jmitchell: Let's at least change the following line from `state-wallet-mainnet` to `state-*` since they'll be much more common going forward.

After this is merged the internal wiki page should refer to this document instead to avoid duplication/divergence. We can keep this ignore line anyway.

docs/wip/exchange-onboarding.md
jmitchell: In fact, if we can persuade the devs to make the units obvious in the key name that would be best.

cc @KtorZ

docs/wip/exchange-onboarding.md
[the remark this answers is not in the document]
disassembler: I just tried changing the ports, and it breaks in docker. Should we try to parse the ports used in ExposePorts, or just specify the ports have to remain the same in the docker container?

docs/wip/exchange-onboarding.md
disassembler: I wonder if I should just remove the mention of pagination, or do you think it's relevant to include that in the FAQ?

docs/wip/exchange-onboarding.md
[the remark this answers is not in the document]
jmitchell: Does it work if you change the port mapping specified in the `docker run` options? If so, let's just mention it near the `docker run` call; otherwise, I think it would be better to extract the appropriate attr from the custom config and parse it.

docs/wip/exchange-onboarding.md
jmitchell: Is it described well enough in the API reference docs? If so, you can just refer the reader to that.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,381 @@
...
+                "totalEntries": 1
+            }
+        }
+    }
+
+## Where can I find the API documentation?
+
+Run the latest wallet and go to <https://127.0.0.1:8090/docs/v1/index>.
+
+For the v0 API documentation (deprecated), go to <https://127.0.0.1:8090/docs/v1/index>.
```
jmitchell: > `/docs/v1/index`

That can't be right.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,381 @@
...
+
+This URL can be customized with the `walletListen` attribute in `./custom-wallet-configuration.nix`.
```
jmitchell: s/This URL/The domain and port/

docs/wip/exchange-onboarding.md
jmitchell: Or is it? Before the endpoint path changes I've seen Swagger output that included both V0 and V1.

docs/wip/exchange-onboarding.md
adinapoli-iohk: I cannot remember on top of my head if `amount` is  a fractional number of an unsigned integer. If the latter than maybe you are right, labeling that `lovelaces` might be better, but I have to say I'm a bit on the fence as if people skipped the section or are not familiar with the distinction between ADA vs lovelace this can be even more confusing :)

docs/wip/exchange-onboarding.md
adinapoli-iohk: @disassembler Indeed we generate that "Getting started" from a Haskell interpolated text, available here: https://github.com/input-output-hk/cardano-sl/blob/develop/wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs#L221

docs/wip/exchange-onboarding.md
adinapoli-iohk: @disassembler Here are the links you are looking for:

* (v1) https://localhost:8090/docs/v1/index/index.html
* (v0) https://localhost:8090/docs/v0/index/index.html

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,381 @@
...
+Operating System: NixOS or CentOS 7 recommended, although any linux
+distribution should work with nix package manager used for building
+the binaries
+
+Software Requirements: Nix package manager for building standalone
+and docker containers. On systems using docker, docker > 17.12 required.
+
+## How do I export the CA certificate for the API?
+
+The certificate is generated inside the wallet in the file `tls/server.cert`
```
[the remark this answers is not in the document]
KtorZ: Hey Alfredo, it shouldn't I believe since the client's identity is embedded in the certificate. But please, see my comment on CBR-82 on that matter, I'd appreciate some inputs here :)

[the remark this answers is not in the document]
disassembler: @adinapoli-iohk @KtorZ 

I can definitely weigh in here coming from a security background in my last job. A client certificate needs to be signed by the CA. What I would recommend is having two separate workflows:

1) Daedalus: Generate an SSL CA on first run. Generate cert/key for server (cardano-sl) and cert/key for daedalus. Daedalus should use the client certificate to authenticate to cardano-sl. Optionally, include in daedalus UI a way to export the client cert/key for expert users that want to hit the API directly.

2) Exchanges: Add option for specifying a CA certificate in custom-wallet-config.nix. On first run, cardano-sl should output a CSR and quit. The exchange should then generate a certificate signed by the CA specified in custom-wallet-config.nix and put that certificate in the wallet directory. Then on subsequent runs, cardano-sl would use that server certificate. The client can then generate as many client certificates as they need to talk to the cardano-sl API.

If you want more details on CA setup and server/client certificate generation using a CA, I highly recommend https://jamielinux.com/docs/openssl-certificate-authority/index-full.html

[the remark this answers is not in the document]
KtorZ: Hey @disassembler, perhaps you can have a look at #2683 & #2684. You inputs would be appreciated on the matter.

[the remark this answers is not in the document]
deepfire: @disassembler, here's some context on how it was implemented in Daedalus: https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-33

In short, the CA is indeed generated at install-time.

disassembler:
@adinapoli-iohk thanks for the review! I don't intend to maintain those separately from the swagger docs. Currently, this doc is more of a WIP of what a client deliverable (probably PDF) would contain. It would be awesome if we could automate pulling the docs out of swagger and injecting them in the source document (maybe using something like docbook or latex to generate the PDF) when we get to that point though so it's not a copy/paste endeavor.

.gitignore
```diff
@@ -81,3 +81,14 @@ custom-wallet-config.nix
...
+state-wallet-mainnet
```
jmitchell: Please change this to `state-*` (see https://github.com/input-output-hk/cardano-sl/pull/2673/files#r174981381)

docs/wip/exchange-onboarding.md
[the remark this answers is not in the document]
jmitchell: Ping

jmitchell:
@disassembler, @KtorZ, @adinapoli-iohk: let's consider a few short-term strategies to deal with the duplicated API documentation:

1. Continue maintaining it, manually if necessary.
2. In depth discussion of API endpoints in this document are changed into links to the Swagger docs.
3. Everything in this document is merged into the Swagger docs.

(1) is a terrible maintenance burden. Sam suggested we may be able to extract portions of the Swagger docs. Although possible, I'm concerned about the amount of effort required to do this well and the ongoing need to test for regression.

(2) is DRY and results in a clean separation between build+run concerns and usage concerns. We'd first have to resolve where the Swagger docs will be published on the web, and how to ensure the links from this document to it don't go stale.

(3) is enticing because it eliminates the external coupling problems in (2). However, exchange integrators still need a way to deliver a definitive, current copy of the docs.

Is there already a strategy in mind for publishing the new Swagger docs on the web, and versioning them for exchange consumption?

docs/wip/exchange-onboarding.md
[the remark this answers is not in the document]
jmitchell: Sam just told me the port mapping didn't work, even when `-p` was manually and explicitly passed to `docker run`.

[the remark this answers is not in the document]
disassembler: Per Jake's comments just now, I'm going to convert these listen's to an attr set with host and port so we can specify exposePort of just the port portion. This also is how nixos handles this in most of it's modules, so this is a good best practice to do regardless.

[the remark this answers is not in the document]
disassembler: Port mappings work, just binding to localhost does not for obvious reasons.

I've resolved this with a NOTE added to sample-wallet-config.nix specifying that bind address should be 0.0.0.0 for docker users and they have to specify `-p` when running to expose on a custom port. We can revisit later if we want with getting the port from the sample config and exposing that, and maybe doing an assertion that IP is 0.0.0.0 when building docker images.

disassembler:
basic NixOS tests are added for v0/v1 API endpoints and v1 docs.

docs/wip/exchange-onboarding.md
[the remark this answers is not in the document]
jmitchell: > I'm going to convert these listen's to an attr set with host and port so we can specify exposePort of just the port portion

Was this done? If so, please update the `sample-wallet-config.nix` to reflect that.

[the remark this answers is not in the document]
jmitchell: Which attrs from the `custom-wallet-config.nix` have you verified are supported in Docker? Please test any remaining and convey the lack of support in the doc.

nixos-tests/cardano-node.nix
```diff
@@ -0,0 +1,40 @@
...
+    $server->waitForUnit("cardano_node_default");
+    $server->waitForOpenPort(8090);
+    $server->succeed("${pkgs.curl}/bin/curl -f -k https://127.0.0.1:8090/docs/v1/index/");
+    $server->succeed("${pkgs.curl}/bin/curl -f -k https://127.0.0.1:8090/api/info");
+    $server->succeed("${pkgs.curl}/bin/curl -f -k https://127.0.0.1:8090/api/v1/node-info");
+    $server->waitForUnit("cardano_node_custom_port");
+    $server->waitForOpenPort(8091);
+    $server->succeed("${pkgs.curl}/bin/curl -f -k https://127.0.0.1:8091/api/v1/node-info");
+  '';
+})
```
jmitchell: Excellent! If you haven't already, please integrate this into CI.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,261 @@
...
+    TODO: private relays should be inserted here in document given to exchanges
+
+## Building and running
+
+We support two different ways to build and run the wallet. You can run
+it standalone on any linux system built in the nix store or you can generate
+a docker container that can be ran on any docker container orchestrator like
+docker swarm or kubernetes.
+
+### Build and run in the nix store
```
rvl: Could you prefix this heading with *Method 1:* and the next heading with *Method 2:*. This makes it clear to the skimmer that only one of the methods needs to be done.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,261 @@
...
+available as a symlink called `./launch_2018-01-30_0d4f79eea`, or
+similar. Run that symlink as a script to start the wallet.
+
+### Build and run a docker image
+
+Follow the above instructions for customization and dependencies. To build a docker
+container and import the image run
+(use `connectScripts.stagingWallet` for testnet):
+
+    docker load < $(nix-build -A dockerImages.mainnetWallet)
```
rvl: Maybe add `--no-out-link` here.

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,261 @@
...
+## Fetch latest code
+
+Clone the [cardano-sl repository](https://github.com/input-output-hk/cardano-sl) or `cd` into a preexisting copy.
+
+    git clone https://github.com/input-output-hk/cardano-sl.git
+    cd cardano-sl
+
+Switch to the `master` branch and pull the latest code.
+
+    git checkout master
```
KtorZ: Is it still the case with the recent changes made to the workflow? 

> rename current master branch to release/1.0.x (this is used only for exchanges)
> rename release-1.1.0 to master (latest production)

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,261 @@
...
+and docker containers. On systems using docker, docker > 17.12 required.
+
+## How do I export the CA certificate for the API?
+
+The certificate is generated inside the wallet in the file `tls/server.cert`
+
+If you are using the docker container, this can be output using the command:
+
+`docker exec -it cardano-mainnet-wallet cat /wallet/state-wallet-mainnet/tls/server.cert`
+
```
KtorZ: Just to clarify here, do we generate new certificates with each build? Meaning that exchanges A will end up with different and incompatible certificates than exchange B (not knowing anything about private keys of A)?

docs/wip/exchange-onboarding.md
```diff
@@ -0,0 +1,261 @@
...
+Examples of tasks that can be done via the API that an exchange would need to do:
+
+* Create/Restore a wallet
+* Make a single transaction
+* Make a batch transaction
+* Estimate transaction fees
+* Generate new addresses to receive funds
+* Get balances of a wallet
+* Get transaction history for a specific wallet
+* Get transaction history for a specific address
```
KtorZ: cf https://github.com/input-output-hk/cardano-sl/pull/2718

docs/wip/exchange-onboarding.md
disassembler: That's a tricky one, because these docs are really going to be for master once v1 is on master, but before that, we're going to be telling people to use specific commit on develop branch.

docs/wip/exchange-onboarding.md
disassembler: We don't generate a CA at all right now, and a new cert/key are only regenerated if the tls directory doesn't exist in the wallet:
```
if [ ! -d ./cardano-state-staging/tls ]; then
  mkdir ./cardano-state-staging/tls/
  /nix/store/jn350na5wcshwsixpq1wbskdp7r3x0r3-openssl-1.0.2l-bin/bin/openssl req -x509 -newkey rsa:2048 -keyout ./cardano-state-staging/tls/server.key -out ./cardano-state-staging/tls/server.cert -days 3650 -nodes -subj "/CN=localhost"
fi
```

docs/wip/exchange-onboarding.md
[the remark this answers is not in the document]
disassembler: All of the attributes are supported. Tested `additionalNodeArgs` with `--json-log` and verified it generated the new file. Tested `ghcRuntimeArgs` with `-s` and verified extra output when node shuts down.

Since I was able to get it working with `-p` and using `0.0.0.0` as the bind IP, I didn't redo the config to split out host and port in the config. This could be done if we wanted to, but I'm not sure it's required anymore. There is a comment in sample wallet config that specifies with docker, you should use `0.0.0.0` as the node IP.

docs/wip/exchange-onboarding.md
jmitchell: Now tracked in DEVOPS-769.

docs/wip/exchange-onboarding.md
adinapoli-iohk: @disassembler @KtorZ Indeed that's tricky -- can we think about a middle ground (i.e. an interim solution) to make sure we make explicit that, up until we finally get 100% git-flow compliant, exchanges should really close the latest available `release/x.x.x` branch?

jmitchell:
I manually verified the AppVeyor builds are green (sadly the AppVeyor -> GH feedback is broken for us at the moment).

Squashing and merging.

PR #2675, opened 2018-03-16
[CSL-2354] Merge master to develop 2018-03-16

nc6:
Second batch of the merge from master. This PR:
- Transitions to using `Address` rather than `CId Addr` in the wallet database. This means we can also drop the cache layer which retains a mapping between these things.
- Speeds up `getWallet`.

Note that currently building of the test for migrations is disabled, since develop includes an `Arbitrary` instance for `HeaderHash` which adds a `HasConfiguration` constraint, thus making this test fail to compile.


adinapoli-iohk:
> Note that currently building of the test for migrations is disabled, since develop includes an Arbitrary instance for HeaderHash which adds a HasConfiguration constraint, thus making this test fail to compile.

I beg your pardon for the silly request, @nc6 , but for the hasty reader: can this impact in any way the operations, such that we might end up not being able to migrate a DB forward during the next Daedalus update?

nc6:
@adinapoli-iohk Good question. The answer is no - this only affects the test, within which we are no longer able to conjure up a 'WalletSnapshot' for QuickCheck purposes. But since the 'WalletSnapshot' has not changed between master and develop, this test has already been run on master.

wallet/cardano-sl-wallet.cabal
```diff
@@ -300,3 +300,27 @@ test-suite cardano-wallet-test
...
+  if os(linux)
+    ghc-options:       -optl-fuse-ld=gold
+    ld-options:        -fuse-ld=gold
+  build-depends:      base
+                    , cardano-sl
+                    , cardano-sl-core
+                    , cardano-sl-txp
+                    , cardano-sl-wallet
+                    , safecopy
+                    , tasty
```
adinapoli-iohk: @nc6 Do you think there would be a value into not pulling yet another dependency with `tasty` here but simply write another spec using the `hspec-autodiscover` functionality? I'm a bit fan of `tasty` as well, but considering all our woes recently with the dependency number, maybe it's worth considering 🙈

nc6:
Tested in daedalus:

- [X] Importing a wallet works
- [X] Creating a new wallet works
- [X] Sending Txs between two wallet works
- [X] Generating a new Cardano address works
- [X] Deleting a wallet works
- [X] Restoring a wallet works

nc6:
Looks like there is a genuine test failure here - replicated with the given seed and unable to replicate on develop. Will look into it.

adinapoli-iohk:
@dcoutts @nc6 Merging as per plan. Thank you both!

PR #2678, opened 2018-03-16
Architecture of Explorer's frontend

sectore:


adinapoli-iohk:
Merging this by the virtue it's 100% documentation, we don't need CI to pass for this.

PR #2679, opened 2018-03-16
[CSL-2331] EnqueueMsg shows the TVar PackStatus

avieth:
This is a follow-up to #2598 which exposes the `TVar (PacketStatus t)` in the cardano-sl-infra types.

### DIFF
--- a/infra/Pos/Communication/Protocol.hs
+++ b/infra/Pos/Communication/Protocol.hs
@@ -25,11 +25,13 @@ module Pos.Communication.Protocol
 
 import           Universum
 
+import qualified Control.Concurrent.STM as STM
 import qualified Data.HashMap.Strict as HM
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Text.Buildable as B
 import           Formatting (bprint, build, sformat, (%))
 import           Mockable (Async, Delay, Mockable, Mockables, SharedAtomic)
+import qualified Network.Broadcast.OutboundQueue as OQ
 import qualified Node as N
 import           Node.Message.Class (Message (..), MessageCode, messageCode)
 import           Serokell.Util.Text (listJson)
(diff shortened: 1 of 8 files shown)


lib/src/Pos/Diffusion/Full/Block.hs
```diff
@@ -84,13 +86,13 @@ instance Exception BlockNetLogicException where
 -- | Expects sending message to exactly one node. Receives result or
 -- fails if no result was obtained (no nodes available, timeout, etc).
 enqueueMsgSingle ::
-       ( MonadThrow m )
-    => (t2 -> (t1 -> t -> NonEmpty x) -> m (Map NodeId (m b)))
+       ( MonadThrow m, MonadIO m )
```
avieth: This is a silly thing. I'd like to make it monomorphic (just use `IO`) but that will be quite a bit of extra work. Will do soon but not now.

avieth: Never mind that, I got rid of the `MonadThrow m` and `liftIO` the `throwIO`s instead.

avieth:
@karknu @coot could I get a quick review on this one? Nothing controversial here I thihnk: I just change the  `EnqueueMsg` type so that it shows the `TVar` implemented in the last CSL-2331 pull request. Other changes restore the original behaviour of existing `EnqueueMsg` users by `waitForDequeues` which blocks until the `TVar` shows a `PacketDequeue (Async a)`.

avieth:
> Do we ever use waitForConversations without waitForDequeues? Maybe it should be waitForConversations = sequenceA . waitForDequeues

We do. `infra/Pos/Communication/Relay/Logic.hs` line 474.

`waitForConversations (try <$> waitForDequeues its)`

infra/Pos/Communication/Protocol.hs
```diff
@@ -164,17 +166,22 @@ alternativeConversations nid ourVerInfo theirVerInfo convs
...
+    -> (forall t . Msg -> (NodeId -> VerInfo -> N.Conversation PackingType m t) -> m (Map NodeId (STM.TVar (OQ.PacketStatus t))))
     -> Converse PackingType PeerData m
     -> SendActions m
 makeSendActions ourVerInfo enqueue converse = SendActions
     { withConnectionTo = \nodeId mkConv -> N.converseWith converse nodeId $ \pVI ->
           alternativeConversations nodeId ourVerInfo pVI (mkConv pVI)
-    , enqueueMsg = makeEnqueueMsg ourVerInfo enqueue
-    }
+      -- NB: here we don't use 'converse'; that's ultimately used by the
+      -- outbound queue's dequeue thread. We just enqueue there.
```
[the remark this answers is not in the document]
avieth: Right. Fixed

avieth:
@coot I've responded to your review.

PR #2682, opened 2018-03-19
Wallet spec input selection

edsko:
This is Duncan's thoughts on input selection, submitted just before going on holidays (cutting it so short he wasn't able to create a PR :grin: ).

### DIFF
--- a/wallet-new/docs/spec.tex
+++ b/wallet-new/docs/spec.tex
@@ -1053,25 +1053,225 @@ should extend to this modified definition without difficulty.
 
 Let us start by identifying the problems.
 
-Selecting transaction inputs is non-trivial because of fees and because there
-are multiple competing goals.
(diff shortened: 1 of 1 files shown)


adinapoli-iohk:
haha, excellent, @edsko ! Will give it a look later today, if that's OK with you 😉

denisshevchenko:
Why TeX? We're using Markdown, right?

edsko:
> Why TeX? We're using Markdown, right?

Markdown? No, we're not.

edsko:
Typesetting all this math in markdown would be painful :)

erikd:
@denisshevchenko Markdown is the right choice for much of the documentation of how things work conceptually, coding guidelines etc, For the formal specification of the system that @edsko and others are working on, Markdown is not sufficient and hence TeX is being used..

PR #2684, opened 2018-03-19
[CBR-145] - Draft end-user documentation about TLS authentication

KtorZ:
Here's a bit of documentation explaining how to generate TLS certificates for client-server authentication, as well as how to use them with cardano-sl.

See #2683

### DIFF
--- a/docs/tls-authentication.md
+++ b/docs/tls-authentication.md
@@ -1,0 +1,248 @@
+# Authentication
+
+To defend against impersonation attacks, as well as to maintain an encrypted communication
+between the wallet's frontend (a.k.a. _[daedalus](https://github.com/input-output-hk/daedalus)_)
+and the wallet's backend (a.k.a _[cardano-sl](https://github.com/input-output-hk/cardano-sl)_) both parties rely on TLS certificates and mututal
+verification of their identity. 
+
+
+## Table of Contents
+
+  * [Motivation and Design](#motivation-and-design)
+
+  * [Prerequisites](#prerequisites)
+
+  * [Generating Certificates](#generating-certificates)
+    + [Default (localhost)](#quick-and-easy)
+    + [Personalized (custom domain)](#step-by-step)
+
+  * [Using Certificates](#using-certificates)
+    + [Setting Up Cardano-SL Backend](#setting-up-cardano-sl-backend)
+    + [Contacting Cardano-SL Backend](#contacting-cardano-sl-backend)
+
+
+## Motivation and Design
+
+Communication between _daedalus_ and _cardano-sl_ is done via http server. There are various
+attacks possible to this kind of environment. In order to secure the connection between these
+two services, we rely on TLS and X.509 public key certificates signed by a known authority. 
+
+Hence, the backend server expects clients to provide valid X.509 v3 certificates signed by an
+approved authority (identified by another certificate, provided on initialization) to verify
+their identity. On the other hand, clients must check the certificate provided by the server to
+make sure they're dealing with the expected party and not a malicious service trying to hijack
+the backend's identity. 
+
+Certificates (and corresponding private keys) are generated off-band, and signed by a same CA's
+(e.g. _Certificate Authority_) private key. The steps below describe how to securely generate
+them. Securing the access to the provided private keys is of vitale importance in order to keep
+the system safe. **Any process or machine with access to the private keys may be able to
+usurp daedalus or cardano-sl's identities**.
+
+
+## Prerequisites
+
+- Install and configure [OpenSSL](https://www.openssl.org/)
+
+
+## Generating Certificates
+
+> **NOTE**: If you're setting up backend nodes and a frontend wallet on a same machine, 
+> you can ignore the _personalized_ installation and fully rely on the _default_ case.
+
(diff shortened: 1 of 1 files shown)


docs/tls-authentication.md
```diff
@@ -0,0 +1,248 @@
...
+language. Make sure to provide the client certificate and key generatd in previous steps and
+signed by the same authority provided to the server.
+
+Additionally, to try things out on Unix systems, you may use `curl` as follows:
+
+```
+$ curl -kv --cert ./tls/client/client.crt --key ./tls/client/client.key <url>
+```
+
+The `k` flag ignores security checks on certificates (as they're done by the backend anyway),
```
disassembler: I'll admit, I use `-k` a lot for debugging purposes, but in documentation we give to the client, we really should recommend using `--cacert` to the ca certificate used to sign the server/client certificates instead, or better yet having them add the CA certificate to ca-bundle on any machines connecting to the API.

docs/tls-authentication.md
```diff
@@ -0,0 +1,248 @@
...
+Certificates (and corresponding private keys) are generated off-band, and signed by a same CA's
+(e.g. _Certificate Authority_) private key. The steps below describe how to securely generate
+them. Securing the access to the provided private keys is of vitale importance in order to keep
+the system safe. **Any process or machine with access to the private keys may be able to
+usurp daedalus or cardano-sl's identities**.
+
+
+## Prerequisites
+
+- Install and configure [OpenSSL](https://www.openssl.org/)
```
disassembler: We already have a requirement of nix to build `cardano-sl` for exchanges. Lets just ship any scripts needed with a shebang to include openssl as a dependency.

docs/tls-authentication.md
```diff
@@ -0,0 +1,248 @@
...
+
+#### Running Script
+
+Run the script to generate the certificates as follows:
+
+```
+$ bash ./build-certificates-unix.sh 
+```
+
+Once done (`Oll Korrect`), a folder `tls` should have been created in your current directory
```
disassembler: One thing to note. I would assume we do, but do we handle intermediate certificates in a server.crt or client.crt file correctly? I could definitely see an organization using their existing CA (maybe in Active Directory) signing an intermediate signing authority cert/key specific for exchange API, and then issuing server/client certificates from that intermediate CA. That's what I would do if I ran an exchange at least.

docs/tls-authentication.md
```diff
@@ -0,0 +1,248 @@
+# Authentication
+
+To defend against impersonation attacks, as well as to maintain an encrypted communication
+between the wallet's frontend (a.k.a. _[daedalus](https://github.com/input-output-hk/daedalus)_)
```
disassembler: I would leave mention of daedalus out of these docs. Typical users that are following these steps are probably not using daedalus, but just accessing the API directly (e.g. exchanges).

docs/tls-authentication.md
```diff
@@ -0,0 +1,248 @@
...
+[ server_alt_names ]
+DNS.1                   = www.example.com
+DNS.2                   = example.com
+```
+
+Then, follow the steps described in [Default | Running Script](#running-script) to generate the certificates.
+
+#### Different Expiry Date
+
+By default, certificates generated are valid for a period of 3 years. If you need a shorter (or
```
disassembler: Do we do 3 years for CA and client/server certs? Typically, CA certs are longer, like 10 or 20 years. I'm also a fan of 1 year or less for server certs (most of mine are 90 days), but if 3 is what we decided to use, that's fine.

docs/tls-authentication.md
```diff
@@ -0,0 +1,248 @@
...
+DNS.1                   = localhost.localdomain
+DNS.2                   = localhost
+DNS.3                   = 127.0.0.1
+DNS.4                   = ::1
+```
+
+and replace DNS references with your actual domain. For instance:
+
+```
+[ server_alt_names ]
```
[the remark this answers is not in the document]
KtorZ: :) ... We kinda had to, because using only the common name results in a very strict name validation and hosts like "127.0.0.1" are treated differently from "localhost" (which is not _wrong_ so to speak, but doesn't properly work for our case). It's more a commodity, but at least it now gives room for future user to easily tweak that, in a right way.

docs/tls-authentication.md
KtorZ: You have a point. Will do :+1:

docs/tls-authentication.md
KtorZ: Maybe have a look at [CBR-154](https://iohk.myjetbrains.com/youtrack/issue/CBR-154) ?

docs/tls-authentication.md
KtorZ: Perhaps I could reformulate to more general point of view, talking about _"clients"_ instead of daedalus for that it is what matters eventually.

docs/tls-authentication.md
KtorZ: I haven't changed this settings from the daedalus installers scripts. It is currently set to `1095` days. I am also in favor of a shorter default expiration period. 1 year seems like a good trade-off. People are still free to change the expiry date as explained if needed.

KtorZ:
Closed now, will update later once we're done with #2742

PR #2685, opened 2018-03-20
[CSL-2328] no reflection configurations for diffusion layer (part 1)

avieth:
## Description

Subset of #2590 intended to make it easier for people to review.

## Linked issue

[CSL-2328]

## Type of change
- [x] 🐞 Bug fix (non-breaking change which fixes an issue)
- [x] 🛠 New feature (non-breaking change which adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] ⛑ git-flow chore (backport, hotfix, etc)

PR #2689, opened 2018-03-20
[CSL-2328] generate block without configurations

avieth:
You can give your ProtocolMagic and ProtocolConstants as regular plain
old vanilla first class values!

The block benchmarks now don't need all of that tedious configuration
stuff.

### DIFF
--- a/block/bench/Block.hs
+++ b/block/bench/Block.hs
@@ -12,7 +12,7 @@ import qualified Data.ByteString.Lazy as LBS
 import           Formatting (build, sformat, shown)
 import           System.Environment (lookupEnv)
 
-import           Pos.Arbitrary.Block.Generate (generateBlock)
+import           Pos.Arbitrary.Block.Generate (generateMainBlock)
 import           Pos.Binary.Class (Bi, serialize, unsafeDeserialize)
 import qualified Pos.Block.BHelpers as Verify
 import           Pos.Core (Block, BlockHeader, BlockVersionData (..), Body, BodyProof,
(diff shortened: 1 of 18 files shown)


block/bench/Block.hs
```diff
@@ -161,13 +99,13 @@ withSerialized a = (a, serialize a)
 
 -- | Make a TestSubject using a seed for a PRNG and size.
 testSubject
-    :: ( HasConfiguration )
+    :: ( )
```
erikd: Worth dropping empty constraint set?

block/bench/Block.hs
```diff
@@ -183,13 +121,10 @@ testSubject seed size =
 
   in  TestSubject {..}
 
-benchMain :: ( HasConfiguration ) => Int -> Int -> IO ()
+benchMain :: ( ) => Int -> Int -> IO ()
```
erikd: As above.

ssc/Pos/Ssc/Base.hs
```diff
@@ -100,32 +108,50 @@ mkSignedCommitment
...
 isOpeningIdx :: HasProtocolConstants => LocalSlotIndex -> Bool
-isOpeningIdx =
-    inRange (toLocalSlotIndex (2 * slotSecurityParam),
-             toLocalSlotIndex (3 * slotSecurityParam - 1))
+isOpeningIdx = isOpeningIdxExplicit protocolConstants
+
+isSharesIdxExplicit :: ProtocolConstants -> LocalSlotIndex -> Bool
+isSharesIdxExplicit pc =
+    inRange (toLocalSlotIndex pc (4 * pcSlotSecurityParam pc),
+             toLocalSlotIndex pc (5 * pcSlotSecurityParam pc - 1))
```
erikd: I know this was in the original code, but any idea what these magic numbers (`4` and 5`) are about?

erikd:
Lie with #2685 I cannot approve this, possibly because it has already been merged to `networking/develop`.

PR #2693, opened 2018-03-20
[CSL-2335] Implement API V1 Feedback

parsonsmatt:
## Description

Addresses the feedback contained in CSL-2335. This spans a couple of concerns, and may be better broken into individual PRs.

- [x] Endpoints which return an `Address` should instead return a `WalletAddress`
- [x] Some docs are misleading (eg `All Foos` vs `All Available Foos`)
- [x] `/api/v1/wallets` needs to add `sort_by=creation_date`
- [x] Feature request: `IN` operator for filters, working like SQL
- [x] Better documentation re: filters and the operations they support
- [x] `Wallet` object needs `assurance`, `hasPassword`, and `passwordLastUpdated` fields.
- [ ] Implement `/api/v1/wallets/import` (only used in development? acceptance tests initialize the wallet? maybe we can get away with this being an initialize option?)
- [x] Clarification re type ("local"/"foreign") in Transactions response
- [x] Transaction object missing `date`, `state` (`pending`/`failed`/`okay`)
- [ ] "Request should include "feesIncluded" flag (if true: amount should include fees) - we currently do not use this flag but it would improve the UI greatly!"
- [x] `/api/v1/transactions` should have `wallet_id` as an *optional* parameter.
- [ ] Document error responses for individual endpoints.

## Linked issue

[CSL-2335 YouTrack Issue](https://iohk.myjetbrains.com/youtrack/issue/CSL-2335)

## Type of change

## Developer checklist

## Testing checklist


wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -51,36 +52,43 @@ newTransaction submitTx Payment {..} = do
...
+                    migrate wh
 
+            -- Paginate result
+            respondWith requestParams fops sops (IxSet.fromList <$> transactions)
+        _ ->
+            -- TODO: we need to implement the case for Nothing
+            -- TODO: should we use the 'FilterBy' machinery instead? that
+            --       let us express RANGE, GT, etc. in addition to EQ. does
+            --       that make sense for this dataset?
+            error "Implement me!"
```
parsonsmatt: Looks like the old code requests *either* a `wallet_id` *or* an `AccountId`, which is a wallet ID paired with an account index:

```haskell
data AccountId = AccountId
    { -- | Address of wallet this account belongs to
      aiWId   :: CId Wal
    , -- | Derivation index of this account key
      aiIndex :: Word32
    } deriving (Eq, Ord, Show, Generic, Typeable)
```

So a `WalletId` is in fact necessary -- and we optionally can take an `AccountIndex`. Which is what the original API supported :\

parsonsmatt: However, we're aiming to support eg `BlockExplorer` use case, which gets 5 most recent transactions that the wallet knows about, regardless of who is doing it.

So we can keep the query param, do a `throwError err400` if they don't pass it, and later un-break that part of the API when we can support it.

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
```diff
@@ -48,14 +48,23 @@ instance Migrate from to => Migrate [from] [to] where
...
-  eitherMigrate a = NE.fromList <$> mapM eitherMigrate a
+instance (Migrate from to, Typeable from, Typeable to) => Migrate [from] (NonEmpty to) where
+    eitherMigrate [] = Left . MigrationFailed $ mconcat
+        [ "Failed migrating "
+        , show (typeRep (Proxy @[from]))
+        , " to a non empty list of "
+        , show (typeRep (Proxy @to))
+        , " because the list was empty."
+        ]
+    eitherMigrate (x:xs) = (:|) <$> eitherMigrate x <*> mapM eitherMigrate xs
```
parsonsmatt: the old instance could blow up with a `Right _|_` if we all `eitherMigrate []` :frowning:

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -482,25 +482,36 @@ instance Arbitrary WalletUpdate where
...
+            --^ "Unique wallet identifier"
+            & "name"
+            --^ "Wallet's name"
+            & "balance"
+            --^ "Current balance, in ADA"
+            & "hasPassphrase"
+            --^ "Whether or not the wallet has a passphrase"
+            & "passphraseUpdatedAt"
+            --^ "The timestamp that the passphrase was last updated."
+        )
```
parsonsmatt: formatting change, later because the old formatting would result in some nastier diffs :(

adinapoli-iohk:
Thank you @parsonsmatt ! I didn't have the time yet to review the feedback document nor this PR, but will get to it and hopefully shed some light on the obscure parts 😉

wallet-new/integration/Functions.hs
```diff
@@ -10,7 +10,7 @@ import           Data.Coerce (coerce)
 import           Data.List (delete)
 import           Data.List.NonEmpty (fromList)
 
-import           Control.Lens ((+~), (<>~), (?~), each, filtered, at)
+import           Control.Lens (at, each, filtered, (+~), (<>~), (?~))
```
adinapoli-iohk: Sorry for the lame question, but have you used `stylish-haskell` (as per the style-guide) to format these imports? (in other terms, I'm trying to understand why this changed 😁 )

wallet-new/integration/Functions.hs
```diff
@@ -21,10 +21,11 @@ import           Cardano.Wallet.API.V1.Types (Account (..), AccountIndex, Accoun
                                               PaymentDistribution (..), PaymentSource (..),
                                               SpendingPassword, Transaction (..), V1 (..),
                                               Wallet (..), WalletAddress (..), WalletId,
-                                              WalletOperation (..), WalletUpdate (..))
+                                              WalletOperation (..), WalletUpdate (..), unV1)
 
 import           Cardano.Wallet.API.V1.Migration.Types (migrate)
-import           Cardano.Wallet.Client (ClientError (..), WalletClient (..), getTransactionIndex, getAddressIndex, getAccounts, getWallets)
+import           Cardano.Wallet.Client (ClientError (..), WalletClient (..), getAccounts,
+                                        getAddressIndex, getTransactionIndex, getWallets)
```
adinapoli-iohk: Judging from these line-wrapped imports, looks like you are -- which means this code was merged without the import formatted, it seems 🤔

wallet-new/integration/Functions.hs
parsonsmatt: Yes, I use it.

Could be a difference in stylish-haskell version or what config it's picking up?

adinapoli-iohk: Possibly. Even more confusingly, the `.stylish-haskell.yml` file changed between `master` and `develop`, with the one used in the latter needing the latest version of `stylish-haskell` to work (> 0.8.2 IIRC)

parsonsmatt:
OK, so sorting of the wallets in V0 happens here:

```haskell
-- wallet/src/Pos/Wallet/Web/State/Storage.hs

-- | Get IDs of all existing and /ready/ wallets.
getWalletAddresses :: Query [CId Wal]
getWalletAddresses =
    map fst . sortOn (view wiCreationTime . snd) . filter (view wiIsReady . snd) . HM.toList <$>
    view wsWalletInfos
```

so `wsWalletInfos` has the creation time, but we drop that from the return information.

nikolaglumac:
@parsonsmatt would it be possible to add this creation time to the wallet object returned by the Api?

parsonsmatt:
@nikolaglumac Yes, but it'll require a bit of refactoring :)

nikolaglumac:
@parsonsmatt this would be very valuable information to us!
cc @adinapoli-iohk @DominikGuzei

parsonsmatt:
@nikolaglumac I've implemented the explicit sort-by and we're getting the creation time of the wallet included as a field :smile:

nikolaglumac:
Brutal stuff @parsonsmatt 🎉 Thanks :)

parsonsmatt:
There are three remaining tasks:

### Implement /api/v1/wallets/import (only used in development? acceptance tests initialize the wallet? maybe we can get away with this being an initialize option?)

We're going to *ideally* sidestep this with the new wallet using some kind of server-side initialization scheme. The V0 endpoint is still available. But reimplementing this is low priority, IIRC.

 ### "Request should include "feesIncluded" flag (if true: amount should include fees) - we currently do not use this flag but it would improve the UI greatly!"

I am going to file this as a feature request for the V1 API -- we're trying to get this verified and finished out, so new feature dev will take a backseat for this sprint.

### Document error responses for individual endpoints.

This is a larger chunk of work that can be done in parallel, so I'm making another story for it.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -822,6 +847,27 @@ instance ToSchema TransactionDirection where
             & type_ .~ SwaggerString
             & enum_ ?~ ["outgoing", "incoming"]
 
+data TransactionStatus
+    = TxPending
+    | TxFailed
+    | TxOk
```
adinapoli-iohk: I think we _might_ need to be a bit more exhaustive here, as historically the transaction status "finite state machine" has been subject of debate and gripe. The crux of the problem is that a `CPtxCondition` is reified into an internal `PtxCondition`, so removing states (like we did here for `TransactionStatus`) is something we might not afford at this stage, I think:

``` haskell
data PtxCondition
    = PtxApplying PtxPoolInfo         -- ^ Is waiting to be applyed
    | PtxInNewestBlocks PtxBlockInfo  -- ^ Recently appeared in block.
    | PtxPersisted                    -- ^ Transaction is ~guaranteed to remain
                                      --   in blockchain
                                      --   (with up to *high* assurance level)
    | PtxWontApply Text PtxPoolInfo   -- ^ Can't be applyed and requires user's
                                      --   input to reform tx
    | PtxCreating PtxPoolInfo         -- ^ Transaction is at its creation cycle
                                      --   (this constructor is at bottom for
                                      --   the sake of SafeCopy instance's
                                      --   backward-compatibility)
    deriving (Eq, Show)
```

I suspect we might need to spend some time thinking how to reify this into V1. Definitely we want things to look and feel cleaner for V1 without giving up useful information V0-users might expect to find or might be relying upon.

wallet-new/test/RequestSpec.hs
```diff
@@ -74,6 +74,12 @@ filterSpec = do
                     `shouldSatisfy`
                         isLeft
 
+            it "supports IN" $ do
+                parseFilterOperation pw pcoin "IN[1,2,3]"
+                    `shouldBe`
+                        Right
+                            (FilterIn (map Core.mkCoin [1,2,3]))
```
adinapoli-iohk: Good lad! 🎉 😁 😍 

What about other draconian cases like `IN[]` etc? Should we test for failures as well?

wallet/src/Pos/Wallet/Web/Methods/Logic.hs
```diff
@@ -219,6 +221,14 @@ getWallets = do
     mps <- withTxpLocalData getMempoolSnapshot
     mapM (getWalletIncludeUnready ws mps False) (getWalletAddresses ws)
 
+getWalletsWithInfo ::  MonadWalletLogicRead ctx m => m [(CWallet, WalletInfo)]
+getWalletsWithInfo = do
+    ws <- askWalletSnapshot
```
adinapoli-iohk: Same deal: if this is not a top level handler we might want to pass both `mps` and `ws` externally.

edsko:
Happened to come across this (during CBR-17 planning work): even after this PR, we still have 

``` haskell
-- | A wallet 'Account'.
data Account = Account
  { accIndex     :: !AccountIndex
  , accAddresses :: [V1 Core.Address]  -- should be WalletAddress
  , accAmount    :: !(V1 Core.Coin)
  , accName      :: !Text
  , accWalletId  :: WalletId
```

Is that intentional?

edsko:
Comment out of date here?

``` haskell
    -- Nest the Accounts API, note that it is left out of the "Wallets" tag purposedly
    :<|> "wallets" :> Accounts.API
```

wallet-new/Makefile
```diff
@@ -0,0 +1,7 @@
+ghcid:
+	ghcid -c "stack ghci cardano-sl-wallet-new"
+
+ghcid-test:
+	ghcid -c "stack ghci cardano-sl-wallet-new:lib cardano-sl-wallet-new:test:wallet-new-specs" --test "main"
+
+.PHONY: ghcid ghcid-test
```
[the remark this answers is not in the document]
parsonsmatt: I LOVE IT

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
```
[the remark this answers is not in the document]
parsonsmatt: Those are both actually already in the default-extensions field :o

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Wallets.hs
```diff
@@ -53,7 +60,8 @@ listWallets :: (MonadThrow m, V0.MonadWalletLogicRead ctx m)
             -> SortOperations Wallet
             -> m (WalletResponse [Wallet])
 listWallets params fops sops = do
-    respondWith params fops sops (IxSet.fromList <$> (V0.getWallets >>= migrate @[V0.CWallet] @[V1.Wallet]))
+    respondWith params fops sops (IxSet.fromList <$> do
+        (V0.getWalletsWithInfo >>= migrate @_ @[V1.Wallet]))
```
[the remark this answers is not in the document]
parsonsmatt: I could make it `mirate @[(V0.CWallet, V0.WalletInfo)] @(V1.Wallet]` if you find that more readable :stuck_out_tongue:

wallet-new/test/RequestSpec.hs
parsonsmatt: mmmm, we don't currently require a `NonEmpty`, so I guess `IN[]` is just "please don't return anything to me" :joy:

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: We could just expose `type TransactionStatus = V1 PtxCondition` ? Then when we've migrated, we can replace the underlying type and keep the representation the same.

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Wallets.hs
```diff
@@ -73,7 +83,19 @@ deleteWallet
...
+    wid' <- migrate wid
+    wallet <- V0.getWallet wid'
+    single <$> addWalletInfo wallet
+
+addWalletInfo :: (MonadThrow m, MonadWalletLogicRead ctx m) => V0.CWallet -> m Wallet
+addWalletInfo wallet = do
+    mwalletInfo <- V0.getWalletInfo (V0.cwId wallet) <$> V0.askWalletSnapshot
+    case mwalletInfo of
+        Nothing ->
+            error "Throw a 500 or something, this is a bug for sure"
```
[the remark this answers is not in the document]
parsonsmatt: Good call :) Fixed both comments.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -482,25 +483,44 @@ instance Arbitrary WalletUpdate where
...
-  declareNamedSchema =
-    genericSchemaDroppingPrefix "wal" (\(--^) props -> props
-      & ("id"      --^ "Unique wallet identifier")
-      & ("name"    --^ "Wallet's name")
-      & ("balance" --^ "Current balance, in ADA")
-    )
+    declareNamedSchema =
+        genericSchemaDroppingPrefix "wal" (\(--^) props -> props
+            & "id"
+            --^ "Unique wallet identifier"
```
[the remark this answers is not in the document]
parsonsmatt: Noted; will change :)

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: Alternatively, I can remove the transaction status from this PR and we can create a new story for it, which provides much more time for detailed discussion and design :smile: 

At the least, I can replace the enum-sum type with a more rich tag-based encoding, like:

```
{
  tag: "transaction status constructor tag",
  obj: {
    // the object's fields according to tag
  }
}
```

For now, `tag` would have the string of the sum type, and `obj` would be empty (or null?) to indicate that we don't have extra fields. When we add new fields, we can do so in a non-breaking manner.

parsonsmatt:
@edsko  Nope, I'll adjust the PR for that :smile:

parsonsmatt:
@edsko Fortunately that was a really easy change -- the old `Account` already had the equivalent, so the `migration` instances did the right thing.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
adinapoli-iohk: @parsonsmatt Uhm, that's a thorny question 🤔 I like the first proposal slightly a bit more, just because tag-based encoding is usually less pleasant for humans, and we are trying to hit that sweet spot between humans & machines 😁 

What about exposing `V1 PtxCondition` as you propose but also analyse the transaction status story as part of next sprint? (I think this might even deserve a dedicated user story in YouTrack).

wallet-new/test/RequestSpec.hs
adinapoli-iohk: Got you! As long as it doesn't blow up in the user face, I'm happy 😁 (More concerned about the possibility that passing an empty list down the guts of the filtering machinery could cause some non-total function to go bonkers).

parsonsmatt: I suppose we could change it to `NonEmpty`, and do an `err400` if they provide nonempty list. That might be the less surprising thing?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: Sounds good! I'll do `V1 PtxCondition` and delegate to it's Aeson instance for now.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -822,6 +845,72 @@ instance ToSchema TransactionDirection where
...
+    WontApply {} ->
+        "wontApply"
+    Creating {} ->
+        "creating"
+
+instance ToJSON TransactionStatus where
+    toJSON x = object
+        [ "tag" .= transactionStatusToText x
+        , "data" .= Object mempty
+        ]
```
parsonsmatt: I've used the `tag/data` representation here to allow maximum flexibility for extensions.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -482,25 +483,44 @@ instance Arbitrary WalletUpdate where
...
 -- | A 'Wallet'.
 data Wallet = Wallet {
-      walId      :: !WalletId
-    , walName    :: !WalletName
-    , walBalance :: !(V1 Core.Coin)
+      walId                  :: !WalletId
+    , walName                :: !WalletName
+    , walBalance             :: !(V1 Core.Coin)
+    , walHasPassphrase       :: !Bool
+    , walPassphraseUpdatedAt :: !(V1 Core.Timestamp)
```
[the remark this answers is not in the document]
parsonsmatt: I'm following convention, TBH.

I think it's wise to use a `newtype` around `UTCTime` or `POSIXTime`, so we can control the representation/parsing. But `Core.Timestamp` seems entirely redundant when `POSIXTime` is essentially the same thing.

I *generally* think it's best to be as precise as possible, and `UTCTime` is the most precise time representation that I know of. `newtype ApiTime = ApiTime UTCTime` would be IMO the best way to go.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -482,25 +483,44 @@ instance Arbitrary WalletUpdate where
...
+    , walCreatedAt           :: !(V1 Core.Timestamp)
+    , walAssurance           :: !AssuranceLevel
```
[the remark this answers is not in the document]
parsonsmatt: Yes, good call! :smile:

wallet-new/test/RequestSpec.hs
parsonsmatt: Otherwise, let's destroy all non-total functions :smile:

wallet/src/Pos/Wallet/Web/Pending/Types.hs
```diff
@@ -89,7 +89,7 @@ data PtxCondition
                                       --   (this constructor is at bottom for
                                       --   the sake of SafeCopy instance's
                                       --   backward-compatibility)
-    deriving (Eq, Show)
+    deriving (Eq, Ord, Show)
```
adinapoli-iohk: Are we sure this `Ord` instance is what we want? I appreciate that's probably needed by our lovely `ixset-typed` filtering to work, but the "stock" derivation will derive ordering based on the order these type constructors are listed. What does it mean for a `PtxCondition` to be comparable? Is it really true that `PtxApplying a` >= `PtxCreating ptxPool` ? I don't have a good answer for you at this stage yet as I can see how we do want to be able to filter info out from the REST API, but it doesn't make sense for all resources to be `ord`erable. Let's tackle this in an another PR, for now let's disregard it (but we should track this into a ticket nevertheless, imho).

wallet-new/test/SwaggerSpec.hs
```diff
@@ -49,7 +49,7 @@ instance ToSchema NoContent where
 spec :: Spec
 spec = modifyMaxSuccess (const 10) $
     describe "Swagger Integration" $ do
-        describe "(V1) ToJSON matches ToSchema" $
+        parallel $ describe "(V1) ToJSON matches ToSchema" $
```
adinapoli-iohk: multicore revolution ftw! 😁

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -822,6 +845,72 @@ instance ToSchema TransactionDirection where
...
+       case tag of
+           "applying" ->
+                pure Applying
+           "inNewestBlocks" ->
+                pure InNewestBlocks
+           "persisted" ->
+                pure Persisted
+           "wontApply" ->
+                pure WontApply
+           "creating" ->
```
adinapoli-iohk: Mental note: we should have some sort of manifesto/guidelines to be sure we are consistent in how we are rendering enumerations in the API, as there is always the choice between snake_case, camelCase etc. Let's pick one style and be consistent 😁

wallet-new/test/MarshallingSpec.hs
```diff
@@ -30,8 +30,8 @@ import qualified Cardano.Wallet.Util as Util
 
 -- | Tests whether or not some instances (JSON, Bi, etc) roundtrips.
 spec :: Spec
-spec = describe "Marshalling & Unmarshalling" $ do
-    describe "Roundtrips" $ do
+spec = parallel $ describe "Marshalling & Unmarshalling" $ do
+    parallel $ describe "Roundtrips" $ do
```
adinapoli-iohk: As we are at it, should we write an `aesonRoundtripProp` for `TransactionStatus`? As we are manipulating freeform-strings, it looks like a good thing to do 😉

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -482,25 +482,44 @@ instance Arbitrary WalletUpdate where
...
 -- | A 'Wallet'.
 data Wallet = Wallet {
-      walId      :: !WalletId
-    , walName    :: !WalletName
-    , walBalance :: !(V1 Core.Coin)
+      walId                         :: !WalletId
+    , walName                       :: !WalletName
+    , walBalance                    :: !(V1 Core.Coin)
+    , walHasSpendingPassword        :: !Bool
+    , walSpendingPasswordLastUpdate :: !(V1 Core.Timestamp)
```
adinapoli-iohk: Thanks for putting up with my idiosyncrasies 😁

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
```diff
@@ -224,9 +238,23 @@ instance Migrate V0.CTx V1.Transaction where
...
+        V0.CPtxCreating{} ->
+            V1.Creating
+        V0.CPtxApplying{} ->
+            V1.Applying
+        V0.CPtxInBlocks{} ->
+            V1.InNewestBlocks
+        V0.CPtxWontApply{} ->
+            V1.WontApply
+        V0.CPtxNotTracked ->
+            V1.Persisted
```
adinapoli-iohk: Cannot really suggest a meaningful mapping for this one (not as any arbitrary than `V1.Persisted` :trollface: ). The documentation for `CPtxNotTracked` is value and I can see it's not even actively used AFAICS, so this is fine I think.

wallet/src/Pos/Wallet/Web/Pending/Types.hs
parsonsmatt: I put it on there because `Transaction` has an `Ord` instance, and it demanded one for this. I'll try removing it and see what happens.

parsonsmatt: Yup, we need an `Ord` instance on `Transaction` in order to put them in an `IxSet`.

parsonsmatt: this is another point in favor of moving to SQL instead of ixset. Though the ordering is entirely arbitrary and only used for constructing the set. Could instead use a hashed set, but that would likely require a rewrite of ixset library, and then you're just trading `Ord a` for `Hashable a, Eq a`.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: I have only seen camel case thus far

wallet-new/test/MarshallingSpec.hs
parsonsmatt: Good call! I wish there was a way to say "For any non-derived instance of ToJSON/FromJSON, ensure that they round trip."

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
parsonsmatt: Yeah :man_shrugging: I think the important thing is to pick a nice representation for the new data layer and force the old one to fit, rather than try and force the new one to conform to the old.

parsonsmatt:
cursed appveyor killed the build after 90 minutes :(

wallet/src/Pos/Wallet/Web/Pending/Types.hs
adinapoli-iohk: Yes, I think the reason you need `Ord` is because under the hood (as you probably have noticed) `ixset` relies on normal sets, thus you need to be able to compare the data.

> this is another point in favor of moving to SQL instead of ixset.

You don't give up easily, do you? 😁 Jokes aside, I'm not particularly attached to `ixset`s and I'm happy to review our options (when @dcoutts is back, hehe). All the "scouting" work we are doing I think is extremely important though, as it means @edsko will be able to take more informative decisions as part of https://iohk.myjetbrains.com/youtrack/issue/CSL-2105 😉

PR #2695, opened 2018-03-21
[CBR-49] Fix CBOR decoding issues

arybczak:
Fixes https://iohk.myjetbrains.com/youtrack/issue/CBR-49.

Details:
- Uses cborg revision with fixes for errors found by auditors
- Completes canonicity checks during decoding of Integer and adds tests for them
- Extends canonicity tests with testing length encoding of collections, strings etc.

### DIFF
--- a/binary/Pos/Binary/Class/Core.hs
+++ b/binary/Pos/Binary/Class/Core.hs
@@ -122,14 +122,14 @@ instance Bi Bool where
 
 instance Bi Char where
     encode c = E.encodeString (Text.singleton c)
-    decode = do t <- D.decodeString
+    decode = do t <- D.decodeStringCanonical
                 toCborError $ if Text.length t == 1
                   then Right (Text.head t)
                   else Left "expected a single char, found a string"
 
     -- For [Char]/String we have a special encoding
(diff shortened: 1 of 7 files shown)


arybczak:
@avieth ping. I pushed fix for the CI.

erikd:
I'm getting a stack build error which is almost certainly related to this:
```
/home/erik/cardano-sl-stack/binary/Test/Pos/Cbor/Canonicity.hs:49:17: warning: [-Wname-shadowing]
    This binding for ‘bs’ shadows the existing binding
      bound at Test/Pos/Cbor/Canonicity.hs:46:17
```

arybczak:
> Seems this warning happened in the Linux CI build as well, but wasn't promoted to an error. Not sure why.

It seems CI compiles without -Werror? It also doesn't fail for me (I'm compiling locally with `stack build --test --no-run-tests`).

Anyway, there is a fix in #2710. @erikd, can you verify that it works for you and merge please?

erikd:
I found this by rebasing my changes against develop and then doing a full `stack test` which is what I usually do before pushing my branch.

PR #2697, opened 2018-03-22
[CSL-2381] WIP: Generate rollbacks in the inductive wallet

nc6:
WIP: please do not merge

## Description

This PR adds rollbacks to the inductive wallet.
It does this by generating, instead of blockchains, block _trees_ - e.g. it encapsulates potential forks. These trees can be tuned to alter the likelihood of forking, how long forks should live etc. We then linearise the tree to create a chain that rolls back in various places.

TODO: rewrite `intersperseTransactions` to not require access to the chain, and instead operate directly on the set of actions. There may be some subtleties with e.g. checking when transactions get confirmed.

## Linked issue
https://iohk.myjetbrains.com/youtrack/issue/CSL-2381

## Type of change

## Developer checklist

## Testing checklist
This is testing code, though could do with some testing of the inductive generator itself.


edsko:
@nc6 looks really good, nice work! Just one minor comment: In `FromPreTree` you write:

```
-- - we do not calculate a ledger, since we have not yet fixed on a single chain.
-- (..)
-- - the boot transaction is included in the chain.
```

Not computing a ledger makes perfect sense, but including the bootstrap transaction in the tree is a bit problematic: it is important that the chain we create in the end does _not_ include the bootstrap transaction. In the DSL, the wallet is initialized (`walletBoot`) using the bootstrap transaction, rather than by having this as an explicit transaction in the ledger. The reason is that on the Cardano side of things, the bootstrap transaction ("genesis block") is hardcoded into the system and so the Cardano wallet kernel will be similarly initialized using the `genesisUtxo`. Crucially, we cannot translate the DSL bootstrap transaction to Cardano.

edsko:
At a slightly higher level: it would be good if the generator would explicitly explore some of the rollback corner cases that are described in the spec. Perhaps the way you do things right now is sufficient for that; we need to think a bit about how likely it is that such scenarios arise. (Things like: pending transaction submitted that depends on an incoming transaction, now we roll back past the point where that incoming transaction was confirmed.) Perhaps it's likely enough that we don't need to pay special attention, but we should verify, ideally both emperically (run some tests and report when this happens) and by making some kind of case as to why this is expected.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -83,33 +89,38 @@ runBlockGenWith settings boot m =
...
+nonAvvmUtxo =
+    currentUtxo . to (utxoRestrictToAddr (not . isAvvmAddr))
 
-selectSomeInputs :: Hash h Addr => BlockGen h (NonEmpty (Input h Addr, Output Addr))
-selectSomeInputs = do
-    utxoMap <- uses bgcNonAvvmUtxo utxoToMap
-    upperLimit <- use bgcInputPartiesUpperLimit
-    input1 <- liftGen $ elements (Map.toList utxoMap)
+selectSomeInputs' :: Hash h Addr
+                  => Int
```
k0001: What are the guidelines for dealing with integer numbers where only positives are expected to succeed?

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -219,3 +236,173 @@ zipFees
...
+    , _treeGenGlobalCtxInputPartiesUpperLimit      :: !Int
+      -- ^ The upper limit on the number of parties that may be selected as
+      -- inputs to a transaction.
+    , _treeGenGlobalCtxForkLikelihood              :: !Double
+      -- ^ Likelihood of generating a forked block at any point in the tree.
+      --   Only the non-integer part of this value is considered.
+    , _treeGenGlobalCtxPruneLikelihood             :: !Double
+      -- ^ Likelihood of the non-principal branch to be terminated at any
+      --   block.
+    , _treeGenGlobalCtxMaxHeight                   :: !Int
```
k0001: What's the policy regarding machine-sized types in situations like this? They make an otherwise readily serializable structure less so. Moreover, shouldn't unsigned `Number` or `Word*` types be preferred in general?

wallet-new/test/unit/UTxO/PreChain.hs
```diff
@@ -64,29 +73,82 @@ preChain = fmap (, ()) . DepIndep
...
+{-------------------------------------------------------------------------------
+  Tree with some information still missing.
+-------------------------------------------------------------------------------}
+
+-- | Blocks organised in a tree structure.
+type BlockTree h a = OldestFirst Tree (UTxO.DSL.Block h a)
+
+-- | Generalisation of a 'PreChain'.
+--   A 'PreChain' is the flattening of a 'PreTree' with a single branch.
+type PreTree h m a = DepIndep (Transaction h Addr) (Tree [Fee]) m (BlockTree h Addr, a)
```
k0001: Are there some guidelines about type synonyms like this? My main concern is that many times the rather cryptic expanded type synonym ends up showing up in error messages from the compiler, and that makes maintaining and refactoring the code much harder.

wallet-new/test/unit/UTxO/PreChain.hs
```diff
@@ -64,29 +73,82 @@ preChain = fmap (, ()) . DepIndep
...
+      fptTree      :: BlockTree h Addr
+
+      -- | The boot transaction
+    , fptBoot      :: Transaction h Addr
+
+      -- | Addresses involved in this block tree
+    , fptAddresses :: Set Addr
+
+      -- | Any additional information that was included in the 'PreChain'.
+    , fptExtra     :: a
```
k0001: Making the fields strict, unless there's a specific reason not to, tends to be a good idea down the road as it save us from causing accidental space leaks.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -83,33 +89,38 @@ runBlockGenWith settings boot m =
...
+    i <- use freshHashSrc
+    freshHashSrc += 1
     pure i
 
-bgcNonAvvmUtxo :: Getter (BlockGenCtx h) (Utxo h Addr)
-bgcNonAvvmUtxo =
-    bgcCurrentUtxo . to (utxoRestrictToAddr (not . isAvvmAddr))
+nonAvvmUtxo :: HasCurrentUtxo src (Utxo h Addr)
+            => Getter src (Utxo h Addr)
+nonAvvmUtxo =
```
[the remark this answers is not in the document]
nc6: `src` _does_ need to be polymorphic; that's the entire point for using `makeFields`

wallet-new/test/unit/UTxO/PreChain.hs
nc6: :+1:

wallet-new/test/unit/UTxO/BlockGen.hs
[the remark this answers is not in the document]
k0001: Oh, I see. That clarifies a lot.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -1,20 +1,26 @@
...
 
+import           Control.Category (id)
 import           Control.Lens hiding (elements)
 import           Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.Map as Map
 import qualified Data.Set as Set
+import           Data.Tree (Tree)
+import qualified Data.Tree as Tree
 import           Pos.Util.Chrono
 import           Test.QuickCheck
```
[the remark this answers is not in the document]
edsko: You can blame me for the the non-explicit imports are my fault (they are so hideous! :). I'd say, fixing imports can be done as part of any kind of PR.. In general I think that changing stuff that's not relevant to the PR distracts when proofreading, but for imports lists there's nothing really to verify anyway.

wallet-new/test/unit/UTxO/PreChain.hs
edsko: There's a minor pull in the other direction here also: we may want all this stuff to stream eventually. But how that will work isn't at all clear to me right now, so being strict is fine for now.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -36,16 +42,16 @@ newtype BlockGen h a
...
+    , _blockGenCtxFreshHashSrc           :: !Int
     -- ^ A fresh hash value for each new transaction.
-    , _bgcInputPartiesUpperLimit :: !Int
+    , _blockGenCtxInputPartiesUpperLimit :: !Int
     -- ^ The upper limit on the number of parties that may be selected as
     -- inputs to a transaction
     }
 
-makeLenses ''BlockGenCtx
+makeFields ''BlockGenCtx
```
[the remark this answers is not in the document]
parsonsmatt: I reached for lenses precisely because I was updating/viewing deeply nested records in a stateful context.

wallet-new/test/unit/UTxO/BlockGen.hs
parsonsmatt: Changing to `Nat` or `Word` is a good call IMO.

wallet-new/test/unit/UTxO/BlockGen.hs
parsonsmatt: Agree on this point

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -219,3 +236,175 @@ zipFees
...
+toPreTreeWith
+    :: Hash h Addr
+    => (TreeGenGlobalCtx h -> TreeGenGlobalCtx h) -- ^ Modify the global settings
+    -> TreeGen h (NoFeeBlockTree h)
+    -> PreTree h Gen ()
+toPreTreeWith settings bg = DepIndep $ \(boot :: Transaction h Addr)-> do
+    ks <- evalStateT (unTreeGen bg) (settings (initTreeGenGlobalCtx boot))
+    return $ \fees ->
+      (OldestFirst (fmap (OldestFirst . dropWhile (== boot))
+                         (zipTreeFees ks fees)), ())
```
parsonsmatt: This formatting is a tiny bit confusing to me; for multi-line tuples I prefer to write:

```haskell
x = 
    ( foo
    , bar
    )
```

to make it clearer what's going on.

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -219,3 +236,175 @@ zipFees
...
+      -- - - we have reached the maximum height - 1, or
+      -- - - with probability equal to the prune likelihood
+      toPrune <- do
+        pl <- use pruneLikelihood
+        maxH <- use maxHeight
+        toss <- liftGenTree $ choose (0,1)
+        return $ (curHeight >= maxH)
+               || ((not $ branchCtx ^. principalBranch)
+                   && ((curHeight >= maxH - 1) || toss < pl)
+                  )
```
parsonsmatt: Tricky boolean logic always messes me up. Would it be useful to factor this out into a separate function and write tests on it?

wallet-new/test/unit/UTxO/BlockGen.hs
```diff
@@ -219,3 +236,175 @@ zipFees
...
+    -- Choose a random height for the blocktree
+    height <- liftGenTree $ choose (10, 50)
+    maxHeight .= height
+    Tree.unfoldTreeM buildTree $ initBranchCtx boot
+  where
+    initBranchCtx boot = TreeGenBranchCtx True 0 (trUtxo boot) [const boot]
+    buildTree :: TreeGenBranchCtx h
+              -> TreeGen h ([Value -> Transaction h Addr], [TreeGenBranchCtx h])
+    buildTree branchCtx = do
+      let curHeight = branchCtx ^. branchHeight
```
parsonsmatt: This reverts to 2 space indent instead of 4

wallet-new/test/unit/Wallet/Abstract.hs
```diff
@@ -525,40 +541,35 @@ newtype InductiveGen h a
...
 
-initializeCtx :: FromPreChain h () -> InductiveCtx h
+initializeCtx :: FromPreTree h () -> InductiveCtx h
 initializeCtx fpc = InductiveCtx{..}
   where
-    icFromPreChain = fpc
-
-getFromPreChain :: InductiveGen h (FromPreChain h ())
-getFromPreChain = asks icFromPreChain
+    icFromPreTree = fpc
```
parsonsmatt: Can be simplified to `initializeCtx icFromPreeTree = InductiveCtx {..}` or even `initializeCtx = InductiveCtx` though the latter will need to change more when the Ctx has more stuff

edsko:
Since this branch was quite bitrotten, I cherry-picked these commits onto a new branch starting from recent `develop`, squashed commits so that we have one from @nc6 and one from @k0001 (renamed to track the new YouTrack organization), and (now) one from me that resolves all the resulting conflicts and updates the code to work with recent `develop`. Tests now failing in the same way that they were before, which is good. Closing this ticket; tracking the new work at https://github.com/input-output-hk/cardano-sl/pull/2880 .

PR #2700, opened 2018-03-23
Minor clean up of follow-the-satoshi related code

erikd:
## Description

Moves the pure version of `followTheSatoshiM` to the same file as
the monadic one and moves `followTheSatoshiUtxo` to the only file
that actually uses it.

## Type of change

## Developer checklist

- I have read the [style guide](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/style-guide.md) document, and my code follows the code style of this project.
- This code does not deal with exceptions.
- I have updated any documentation accordingly, if needed. Documentation changes can be reflected in opening a PR on [cardanodocs.com](https://github.com/input-output-hk/cardanodocs.com), amending the inline [Haddock](https://www.haskell.org/haddock/) comments, any relevant README file or one of the document listed in the [docs](https://github.com/input-output-hk/cardano-sl/tree/develop/docs) directory.

## Testing checklist

- All new and existing tests passed.


lrc/Pos/Lrc/Fts.hs
```diff
@@ -194,6 +198,27 @@ whenever we transition to the next coin, its index bona fide exceeds the
...
+-- random seed passed in as the first parameter.
+--
+-- The probability that a stakeholder will be chosen is proportional to the
+-- number of coins this stakeholder holds. The same stakeholder can be picked
+-- more than once.
+--
+-- How the algorithm works: we sort all unspent outputs in a deterministic
+-- way (lexicographically) and have an ordered sequence of pairs
+-- @(StakeholderId, Coin)@. Then we choose several random 'i's between 1 and
+-- amount of satoshi in the system; to find owner of 'i'th coin we find the
```
aspiwack: I guess the atomic unit is not called "satoshi" in Cardano.

lrc/Pos/Lrc/Fts.hs
```diff
@@ -237,3 +262,19 @@ followTheSatoshiM (SharedSeed seed) totalCoins = do
                 findLeader coinIndex
 
     nextStakeholder = fromMaybe (error "followTheSatoshiM: indices out of range") <$> await
+
+
+-- | A pure version of `followTheSatoshiM` above.
```
aspiwack: How do they relate? What is the purpose of having two functions? Maybe we could have a few words on this?

lrc/Pos/Lrc/Fts.hs
erikd: The smallest indivisible unit of Ada is called the "Lovelace". Not sure that really relevant information here.

lrc/Pos/Lrc/Fts.hs
erikd: Added more comments there.

lrc/Pos/Lrc/Fts.hs
aspiwack: It's just that « between 1 and the amount of satoshi in the system » should really be « between 1 and the amount of lovelace in the system ».

erikd: Fixed, but I wonder at what point do we rename this to `followTheLovelace`? :)

PR #2706, opened 2018-03-24
Add some simple golden tests for Address'

vincenthz:
## Description

Add some golden tests for `Address'` generation. This is effectively commiting to the exact byte representation of Address' given a set of input, so that if the serialization format changes by mistake it will be caught by a test.

This is more about starting the general approach than providing a fulled feature golden tests suite.

## Type of change
- [x] Add tests


cleverca22:
`stack2nix: user error (No such package basement-0.0.7 in the cabal database. Did you run cabal update?)`
`basement-0.0.7` is not in the snapshot `2018-02-06T11:56:04Z`, so the snapshot stack2nix is using will need to be updated, or an older basement used

erikd:
@vincenthz i think if you rebase, this will build.

erikd:
Nearly there @vincenthz . Under nix you need to run the `pkgs/generate.sh` shell script or you can apply the patch that is listed in the [build output](https://buildkite.com/input-output-hk/cardano-sl/builds/4674#11009495-7f04-46ac-bf6e-d1464f0a6f62) starting at line 3219. Yes, that is generated file, but it needs to be in git for pure nix builds.

PR #2712, opened 2018-03-27
Fix /accounts endpoints not accessible because of param inversion

KtorZ:
Accounts endpoints were accessible on:

`/wallets/accounts/{{walletId}}/{{accountId}}`

instead of

`/wallets/{{walletId}}/accounts/{{accountId}}`

which is rather unconventional. I believe it was not intended and happened during a refactoring of the routes when creating the Servant client if I am not mistaken. If it's done on purpose, I am curious to know why ^.^ ?

Note that to keep server & client declaration simple, I've introduced a bit of duplication in the routes declarations. It's a small repetition within a same file for a greater good. 

I recall [a nice blog post](http://www.parsonsmatt.org/2018/03/14/servant_route_smooshing.html) on the topic ;), but in the immediate future, we need a fix.

### DIFF
--- a/wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers.hs
+++ b/wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers.hs
@@ -12,8 +12,10 @@ import           Ntp.Client (NtpStatus)
 import           Pos.Diffusion.Types (Diffusion (sendTx))
 
 import qualified Cardano.Wallet.API.V1 as V1
+import qualified Cardano.Wallet.API.V1.Accounts as Accounts
 import qualified Cardano.Wallet.API.V1.Addresses as Addresses
 import qualified Cardano.Wallet.API.V1.Info as Info
+import qualified Cardano.Wallet.API.V1.LegacyHandlers.Accounts as Accounts
 import qualified Cardano.Wallet.API.V1.LegacyHandlers.Addresses as Addresses
 import qualified Cardano.Wallet.API.V1.LegacyHandlers.Info as Info
 import qualified Cardano.Wallet.API.V1.LegacyHandlers.Settings as Settings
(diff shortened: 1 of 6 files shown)


KtorZ:
@adinapoli-iohk Yeah, I am a bit concerned that no tests actually caught that. I guess that since a lot of things are derived generically, we could also infer the right endpoints in the tests suite based on the API declaration. 

Testing that feels more like a backward compatibility kind of test. No features were broken in the API as long as accessed through the right endpoint, but it was simply different / not compatible with what it used to be. At first glance, it seems rather complex to test (having a bunch of test API calls on a list of hard-coded endpoints will do but I don't feel like going this way....)

parsonsmatt:
That is my fault, great catch!

We could use [`layout`](https://hackage.haskell.org/package/servant-server-0.13/docs/Servant-Server.html#v:layout) to print a Golden Test that's merged into CI that serves as a quick visual inspection on how the route changes. Then PRs which affect the route structure will have a diff in the Servant code *and also* in the simpler route drawing.

adinapoli-iohk:
> We could use layout to print a Golden Test that's merged into CI that serves as a quick visual inspection on how the route changes.

I have never used `layout` in anger before, but the idea of the golden test looks interesting, as anything which would allow us to catch these regressions is great! 😁 

Thinking out loud, if we had a "blackbox, http-client-style" integration testsuite, would have that prevented the regression from happening? To be explicit, I'm talking about a utterly-verbose testsuite where you query our API _as if_ we were an exchange/user, by calling directly into the API via `http-client`. Agreed that's not very sexy, but that's looks also pretty darn useful, at least for "key routes" like this one. In such scenario, our specs would have failed with a `500` or `404`.

@parsonsmatt Do you think it might make sense to have something similar?

PR #2715, opened 2018-03-28
[CSL-2346] remove no-ntp & NtpDrift

coot:
* remove `--no-ntp` from some more scripts
* `NtpDrift` should hold time difference rather than absolute value of it, added some doc strings to `NtpStatuts`.


coot:
Closing, it should be a RP to `develop`

PR #2716, opened 2018-03-28
[CSL-2346] remove no-ntp & NtpDrift

coot:
* remove `--no-ntp` from some more scripts
* `NtpDrift` should hold time difference rather than absolute value of it, added some doc strings to `NtpStatuts`.

### DIFF
--- a/networking/src/Ntp/Client.hs
+++ b/networking/src/Ntp/Client.hs
@@ -42,10 +42,13 @@ import           Ntp.Util (createAndBindSock, resolveNtpHost, selectIPv4, select
                            udpLocalAddresses, withSocketsDoLifted)
 
 data NtpStatus =
+      -- | The difference between ntp time and local system time
       NtpDrift Microsecond
+      -- | NTP client has send requests to the servers
     | NtpSyncPending
-    | NtpSyncUnavailable
-    deriving (Eq, Show)
+      -- | NTP is not available: the client has not received any respond within
+      -- `ntpResponseTimeout` or NTP was not configured.
+    | NtpSyncUnavailable deriving (Eq, Show)
 
 data NtpClientSettings = NtpClientSettings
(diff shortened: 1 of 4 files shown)


networking/src/Ntp/Client.hs
```diff
@@ -42,10 +42,13 @@ import           Ntp.Util (createAndBindSock, resolveNtpHost, selectIPv4, select
                            udpLocalAddresses, withSocketsDoLifted)
 
 data NtpStatus =
+      -- | The difference between ntp time and local system time
       NtpDrift Microsecond
```
avieth: Such that if you add it to the local time, then you get the NTP time?

networking/src/Ntp/Client.hs
```diff
@@ -124,8 +127,7 @@ handleCollectedResponses cli = do
     handler (newMargin, transmitTime) = do
         let ntpTime = transmitTime + newMargin
         localTime <- realTime
-        let timeDiff = abs $ ntpTime - localTime
-        atomically $ writeTVar (ncStatus cli) (NtpDrift timeDiff)
+        atomically $ writeTVar (ncStatus cli) (NtpDrift $ ntpTime - localTime)
```
avieth: Right-o

networking/src/Ntp/Client.hs
coot: Yes

coot:
@avieth the appveyor build fails with a 404 error when trying to download `https://slproweb.com/download/Win64OpenSSL-1_0_2n.exe` which is unavailable (there is a newer version available instead). I asked on `sl-devops` but nobody responded.

coot:
The necessary change was just merged; I rebased, so everything should be ok.

coot:
`ag "no-ntp"` returns no results.

PR #2718, opened 2018-03-28
[CSLD-162] Add few use-cases to the API doc

KtorZ:
- [x] multiple transaction being sent in batch /api/v1/transactions
- [x] generating addresses, listing and checking validity /api/v1/addresses
- [x] fee estimation /api/v1/transactions/fees
- [x] transaction history /api/v1/transactions/{walletId}
- [x] checking sync progress /api/v1/node-info
- [x] getting a specific wallet /api/v1/wallets/{walletId}
- [x] retrieve, create, and list accounts /api/v1/wallets/{walletId}/accounts
- [x] poll deposits / search transactions since a given date


wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -247,38 +247,52 @@ Software Version   | Git Revision
...
 
-> **Warning**: Those 12 mnemonic words given for the backup phrase act as an example. **Do not** use them on a production system. See the section below about mnemonic codes for more information.
+> **Warning**: Those 12 mnemonic words given for the backup phrase act as an example. **Do
+> not** use them on a production system. See the section below about mnemonic codes for more
+> information.
 
-As a response, the API provides you with a wallet `id` used in subsequent requests to uniquely identity the wallet. Make sure to store it / write it down. Note that every API response is [jsend-compliant](https://labs.omniti.com/labs/jsend); Cardano also augments responses with meta-data specific to pagination. More details in the section below about *Pagination*.
+The `spendingPassword` is optional but highly recommended. It is a 64-byte hex-encoded hash of
+a secure passphrase that will be required for sensitive operations on the wallet. You can
+create such hash using a *SHA256* or *BLAKE2b* hashing algorithms.
```
adinapoli-iohk: Good lad, you already suggested two alternatives! ❤️

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -299,23 +313,27 @@ As a response, the API provides you with a wallet `id` used in subsequent reques
...
+     --cacert ./scripts/tls-files/ca.crt
 ```
 
-Receiving Money
----------------
+Receiving Ada
+-------------
 
-To receive money from other users you should provide your address. This address can be obtained from an account. Each wallet contains at least one account, you can think of account as a pocket inside of your wallet. Besides, you can view all existing accounts of a wallet by using the `GET /api/v1/wallets/{{walletId}}/accounts` endpoint as follow:
+To receive _Ada_ from other users you should provide your address. This address can be obtained
```
adinapoli-iohk: Genuine question (I don't know the answer): usually currencies (e.g. USD, EUR, etc) are spelled out in caps-lock. Are you aware of any convention in this space we should abide?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -345,42 +363,52 @@ Since you have, for now, only a single wallet, you'll see something like this:
...
+  "source": {
+    "accountIndex": 0,
+    "walletId": "Ae2tdPwUPE...8V3AVTnqGZ"
+  },
+  "spendingPassword": "5416b2988745725998907addf4613c9b0764f04959030e1b81c603b920a115d0"
 }'
 ```
 
-Note that, in order to perform a transaction, you need to have some existing coins on the source account! Beside, the Cardano API is designed to accomodate multiple recipients payments out-of-the-box; notice how `destinations` is a list of addresses.
+Note that, in order to perform a transaction, you need to have enough existing coins on the
```
adinapoli-iohk: you used `coins` here -- let's stick with ADA 😉

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -461,22 +501,47 @@ Sort operators
 Errors
 ======
 
-In case a request cannot be served by the API, a non-2xx HTTP response will be issue, together with a [JSend-compliant](https://labs.omniti.com/labs/jsend) JSON Object describing the error in detail together with a numeric error code which can be used by API consumers to implement proper error handling in their application. For example, here's a typical error which might be issued:
+In case a request cannot be served by the API, a non-2xx HTTP response will be issue, together
```
adinapoli-iohk: Not a native speaker, but shouldn't this be "will be issued"?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -461,22 +501,47 @@ Sort operators
...
 
 
+Monetary Denomination & Units
+=============================
+
+Cardano's currency is called _Ada_ ( ₳ ). _Ada_ has up to **6** decimal places; hence the
+smallest monetary unit that can be represented in the Cardano's blockhain is: 0.000001₳. This
+is also called a _Lovelace_ (Cardano's currency is named after the mathematician and computer
+scientist [Ada Lovelace](https://en.wikipedia.org/wiki/Ada_Lovelace)). Put in another way, one
+_Ada_ is equal to one million _Lovelace_.
```
adinapoli-iohk: Thanks for adding this section!

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -495,6 +562,363 @@ Disable TLS (Not Recommended)
...
+```
+curl -X GET \
+  https://127.0.0.1:8090/api/v1/wallets/Ae2tdPwUPE...8V3AVTnqGZ/accounts \
+  -H 'Accept: application/json;charset=utf-8' \
+  --cacert ./scripts/tls-files/ca.crt
+```
+
+```json
+{
+    "status": "success",
```
adinapoli-iohk: Idea for the future: we should come up with a way to embed automatically-generated JSON in this documentation, so that if we change the API the changes are reflected automatically in this documentation.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -247,38 +247,52 @@ Software Version   | Git Revision
...
-Please note that wallet web API uses TLS for secure communication. Requests to the API need to send a client CA certificate that was used when launching the node and identifies the client as being permitted to invoke the server API.
+Please note that wallet web API uses TLS for secure communication. Requests to the API need to
+send a client CA certificate that was used when launching the node and identifies the client as
+being permitted to invoke the server API.
 
 Creating a New Wallet
 ---------------------
 
-You can create your first wallet using the `POST /api/v1/wallets` endpoint as follow:
+You can create your first wallet using the [`POST /api/v1/wallets`](#tag/Wallets%2Fpaths%2F~1api~1v1~1wallets%2Fpost) endpoint as follow:
```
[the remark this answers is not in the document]
KtorZ: Yes and no. [This](https://github.com/phadej/servant-swagger-ui/pull/44#issuecomment-378857745) just got merged :tada: ... So, I am waiting for the release and will upgrade `servant-swagger-ui` to rely on ReDoc also locally.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Does this disregard the above comment ^.^ ? Or should I perhaps clarify a bit the wording there because indeed, now that I read it again, 64-byte hex-encoded could be read as "64 (utf-8) bytes **then** encoded to hexadecimal" which will be wrong. So maybe something like:

> It is an original `UTF-8` string of 32 characters, encoded in `base 16`, yielding to an hexadecimal sequence of 64 bytes.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: I believe that this refers to the alpha-3 code from the [ISO-4217](https://www.iso.org/iso-4217-currency-codes.html) standard:

> The alphabetic code is based on another ISO standard, ISO 3166, which lists the codes for country names. The first two letters of the ISO 4217 three-letter code are the same as the code for the country name, and where possible the third letter corresponds to the first letter of the currency name.
>
>For example:
>
>    - The US dollar is represented as USD - the US coming from the ISO 3166 country code and the D for dollar.
>    - The Swiss franc is represented by CHF - the CH being the code for Switzerland in the ISO 3166 code and F for franc.

Even though *ADA* isn't yet in the standard, it'd make sense to align with that I suppose. Good call ^.^

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Ah!

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Are you thinking of using some `Arbitrary` instance to generate examples here (The way we do it with Swagger shemas)?
If so, this makes a lot of sense indeed.

adinapoli-iohk: I was thinking something even simpler: a bunch of "exemplar" dummy types we "splice back" into the documentation. Pseudo-Haskell:

```haskell
data FooBar = Foo {
  bar :: Bar
  baz :: Baz
}

instance JSON a => Documentable a where
    spliceIntoDocumentation :: Proxy a -> Q Expr

ourSwaggerDoc = [text|

This is the example returned by a call to /v1/api/foo-bar:

$(spliceIntoDoc @FooBar Proxy)

```

I mean, you get the idea. There are lots of different ways to accomplish the above, but the idea would be to rely on _types_ not on free form strings to keep our documentation updated if we change either the JSON serialisation format or the type itself :)

KtorZ:
Okay, updated the bit of documentation about generating a spending password to remove the confusion between byte lengths.
I've also enable ReDoc directly on our local build (instead of default swagger-ui) like we have on cardanodocs.com. Why is that:

- To be more consistent between the two world
- The doc is more readable / nicer this way and we actually refer to the local doc in the exchange guide
- I've added a lot of links between sections in the doc as it started to become quite broad with all those new use cases. Those links use the format and anchors created by ReDoc and not the original swagger-ui. 

All in all, it's rather small overhead though I'd prefer relying on [servant-swagger-ui-redoc](http://hackage.haskell.org/package/servant-swagger-ui-redoc) but let's wait for us to upgrade to a more recent LTS as it uses a few package outside of our current LTS.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
parsonsmatt: This is a fantastic idea! 

Before we go *too* much further with homebrewing a document system, though, we might want to consider the `shakespeare` libraries -- they are, at heart, a system for interpolating Haskell values into text like this. The HTML/CSS/JS sublanguages aren't necessary.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -255,38 +257,63 @@ Software Version   | Git Revision
...
 ```
 
-> **Warning**: Those 12 mnemonic words given for the backup phrase act as an example. **Do not** use them on a production system. See the section below about mnemonic codes for more information.
+> **Warning**: Those 12 mnemonic words given for the backup phrase act as an example. **Do
+> not** use them on a production system. See the section below about mnemonic codes for more
+> information.
 
-As a response, the API provides you with a unique wallet `id` to be used in subsequent requests. Make sure to store it / write it down. Note that every API response is [jsend-compliant](https://labs.omniti.com/labs/jsend); Cardano also augments responses with meta-data specific to pagination. More details in the section below about *Pagination*.
+The `spendingPassword` is optional but highly recommended. It is an original UTF-8
+string of 32 characters, encoded in base 16, yielding to an hexadecimal sequence of 64 bytes.
```
parsonsmatt: What's the meaning of a UTF-8 encoded *and* base 16 encoded string?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -255,38 +257,63 @@ Software Version   | Git Revision
...
+This passphrase is required for sensitive operations on the wallet and adds
+therefore an extra security layer to it.
```
parsonsmatt: I would cut `therefore` here; or move it before "adds" -- the wording is somewhat awkward sounding to me.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
adinapoli-iohk: Yup, 100% agree, and we can indeed tackle these bits outside this PR 😉

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -255,38 +257,63 @@ Software Version   | Git Revision
...
+
+To generate a valid `spendingPassword`, please follow the following steps:
+
+    - Pick a strong passphrase (long, from a big set of characters), randomly
+      generated passphrase are best for humans are a poor source of randomness.
```
parsonsmatt: Clarity: Rather than qualifying what is meant by `strong`, we can just state it directly:

> Pick a long (how many characters/words?) passphrase using a wide variety of characters (uppercase, lower case, whitespace, punctuation, etc).

I'd probably edit the sentence to say: "Using a computer to randomly generate a passphrase is best, as humans aren't a good source of randomness," though the only grammar error is "randomly generated passphrase*s* are ..." or "*a* randomly generated passphrase *is* best..."

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -456,7 +509,7 @@ Filter operators
 | `GTE`    | Retrieves the resources with index _greater than equal_ the one provided. | `balance=GTE[10]`      |
 | `RANGE`  | Retrieves the resources with index _within the inclusive range_ [k,k].    | `balance=RANGE[10,20]` |
 
-Sort operators
+Sort Ooperators
```
adinapoli-iohk: @KtorZ Ops, tiny typo 😉

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -353,42 +384,52 @@ Since you have, for now, only a single wallet, you'll see something like this:
...
 ```
 
-Note that, in order to perform a transaction, you need to have enough coins on the source account! The Cardano API is designed to accomodate multiple recipients payments out-of-the-box; notice how `destinations` is a list of addresses (and corresponding amounts).
+Note that, in order to perform a transaction, you need to have enough existing _ADA_ on the
+source account! The Cardano API is designed to accomodate multiple recipients payments
+out-of-the-box; notice how `destinations` is a list of addresses (and corresponding amounts).
 
-When the transaction succeeds, funds are no longer available in the sources addresses, and are soon made available to the destinations. Note that, you can at any time see the status of your wallets by using the `GET /api/v1/transactions/{{walletId}}` endpoint:
+When the transaction succeeds, funds are no longer available in the sources addresses, and are
+soon made available to the destinations in a short delay. Note that, you can at any time see
```
parsonsmatt: "destinations in a short delay" -> "destinations with a short delay"

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -469,22 +522,47 @@ Sort operators
...
+is also called a _Lovelace_ (Cardano's currency is named after the mathematician and computer
+scientist [Ada Lovelace](https://en.wikipedia.org/wiki/Ada_Lovelace)). Put in another way, one
+_ADA_ is equal to one million _Lovelace_.
+
+ADA        | Lovelace
+-----------|----------
+`1`        | `1 000 000`
+`.000 001` | `1`
+
+> **Warning**: All amounts manipulated in the API are given and expected in Lovelace
```
adinapoli-iohk: @KtorZ What about a full stop at the end?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
parsonsmatt: Operators

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+Common Use-Cases
+================
+
+Sending Money to Multiple Recipients
+------------------------------------
+
+As seen in [Sending ADA](#section/Getting-Started/Sending-ADA), you can send _ADA_ to
+another party using the [`POST /api/v1/transactions`](#tag/Transactions%2Fpaths%2F~1api~1v1~1transactions%2Fpost) endpoint.
+Important to notice is the type of the field `destinations`: it's a list, enabling you to provide more
+than one destination. Each destination is composed of:
```
parsonsmatt: hmmm, I think I would rewrite as "Note the type of the field `destinations`: it is a list, enabling you to provide more than one destination."

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+
+- An address
+- A corresponding amount
+
+The overall transaction corresponds to the sum of each outputs. For instance, to send money to
```
parsonsmatt: "sum of each output"

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+```
+
+
+Estimating Transaction Fees
+---------------------------
+
+When you submit a transaction to the network, some fees apply depending on, but not only, the
+selected grouping policy and the available inputs on the source wallet. There's actually a
+trade-off between fees, cryptographic security, throughput and privacy. The more inputs are
+selected, the bigger is the payload, the bigger are the fees.
```
parsonsmatt: > When you submit a transaction to the network, there will be fees charged on the transaction. These fees depend on many factors, including the selected grouping policy and available inputs on the source wallet. (snip). As you select more inputs, the payload size increases, and this results in larger fees.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+}'
+```
+
+
+Estimating Transaction Fees
+---------------------------
+
+When you submit a transaction to the network, some fees apply depending on, but not only, the
+selected grouping policy and the available inputs on the source wallet. There's actually a
+trade-off between fees, cryptographic security, throughput and privacy. The more inputs are
```
parsonsmatt: Do we have a doc or link pointing to an explanation of the trade-offs? Or is this something we'd expect someone reading these docs to know :smile:

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+    "meta": { ... }
+}
+```
+
+
+Managing Accounts
+-----------------
+
+A wallet isn't limited to one account. It can actually be useful to have more than one account
+in order to separate business activities. With the API, you can retrieve a specific account,
```
parsonsmatt: > ~It can actually be useful to have more than one account in order to~ A potential use for multiple accounts is to

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+}
+```
+
+
+Managing Addresses
+------------------
+
+By default, wallets you create are provided with an account which has one default address. It
+is possible (and recommended) for an account to manage multiple addresses. Address reuse
+actually reduces privacy for it tights more transactions to a small set of addresses.
```
parsonsmatt: I think you mean `ties`, not `tights` :)

> Address reuse reduces privacy by associating a large set of transactions with a small set of addresses.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+
+Note that when paying, this is fairly abstracted from you: addresses are selected from an
+account of your wallet based on different strategies and policies.
```
parsonsmatt: I'd reword:

> When paying, the wallet makes many of these choices for you. Addresses are selected from a wallet's account based on several different strategies and policies.

Is there a way to point to these strategies and policies, or give a user a way of knowing/understanding what they are or how to have any control over them?

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+```
+
+If your wallet is protected with a password, this password is also required in order to create
+new addresses for that wallet. In such case, the field `spendingPassword` should match the one
+defined earlier to protect your wallet.
+
+Addresses generated as just described are always valid. However, when dealing with other
+parties, you might want to check whether a given address is a valid base58-encoded address.
+
+For that purpose, use [`GET /api/v1/addresses/{{address}}/validity`](#tag/Addresses%2Fpaths%2F~1api~1v1~1addresses~1{address}~1validity%2Fget):
```
parsonsmatt: :warning:  We removed the `validity` endpoint. The endpoint is now just `api/v1/addresses/#{Text}`, and it returns a `WalletResponse WalletAddress`. If the address is invalid, it returns a 400 error. :warning:

KtorZ:
Typos fixed and wallet's README updated to point to correct urls.
I don't get the buildkite errors however:

```
server/Cardano/Wallet/API/V1/Swagger.hs:48:1: error:
    Failed to load interface for ‘Servant.Swagger.UI’
    It is a member of the hidden package ‘servant-swagger-ui-0.2.4.3.4.0’.
```

It's like it had a different version of the .cabal file. Weird, will see on this build.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+```
+
+
+Managing Accounts
+-----------------
+
+A wallet isn't limited to one account. It can actually be useful to have more than one account
+in order to separate business activities. With the API, you can retrieve a specific account,
+create new ones, list all existing accounts of a wallet or edit a few things on an existing
+account. By default, your wallet comes with a provided account; let's see how to create a fresh
```
parsonsmatt: I'd use a full stop instead of a semicolon here.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+  --cacert ./scripts/tls-files/ca.crt
+```
+
+```json
+{
+    "data": [
+        "YQdiVKZwx...d16oatjqf3",
+        "2iQCuXVhU...S4YoYruxet",
+        "jM8R4tBCP...2ueD7ndDV4",
+        "B9x5woS58...TUDWnYRdfv",
```
parsonsmatt: This isn't right either. We're returning a list of `WalletAddress` which contains more info than just the ID.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -503,6 +583,404 @@ Disable TLS (Not Recommended)
...
+            "unit": "microseconds"
+        }
+    },
+    "status": "success",
+    "meta": { ... }
+}
+```
+
+
+Retrieving Transactions History
```
parsonsmatt: grammar nit: `Retrieving Transaction History`

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
adinapoli-iohk: Well spotted @parsonsmatt -- Did miss this one.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: We read from a utf-8 source and convert that to an hexadecimal sequence. Actually, the input can be pretty much anything, it could be latin-1 encoded initially or ANSI.. Doesn't matter much and perhaps it is confusing here. We'll always be manipulating the hex output anyway.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Pardon my French influence here sir ^.^

adinapoli-iohk:
@KtorZ All in all, I think we made the right thing not rushing into into `release/1.2.0`, as @parsonsmatt found a number of things I previously missed in my review.

As this PR doesn't add any feature but crucially would change the way we expose the live-rendered API from the backend, I think it's a good candidate for inclusion as a hotfix for `release/1.2.0`, after we fix all the pending review points :)

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Hmmm... I kinda disagree with that one. Not sure why, but `in` sounds a bit more accurate here imo.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Not sure whether we have any public document on the topic but we could drop a line about those policies? Maybe as a use-case section "Getting More Control Over the Wallet's strategies"

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: Interesting! Good catch :)

KtorZ:
Agreed. I'll make another round of checks on the API endpoints. There are perhaps more discrepancies with endpoints we reworded or removed.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
adinapoli-iohk: Good point -- historically we introduced these policies to support different use cases, as typically exchanges are interested in different tradeoffs than normal users. This is an interesting article on the topic: https://medium.com/@lopp/the-challenges-of-optimizing-unspent-output-selection-a3e5d05d13ef

Said that, input selection and the policies are going to be reworked in the upcoming data layer, so really the _single_ take-away is that "normal users" shouldn't bother, but exchanges **must** use the `HighThroughput` policy, or they would get too-big transactions. That is definitely worth clarifying :)

KtorZ:
Review wording & re-sync examples with latest API changes

- Removed bits about 'address validity' endpoint that isn't part of the API anymore
- Clarify spending password encoding sentence (removed UTF-8)
- 'adds therefore an extra' -> 'adds an exta'
- Review first step about generating a password
- Reword section about wallet abstracting away some operations

Also, examples of API response are now generated from Arbitrary instances. There were too many discrepancies already to realistically think we can keep that in sync over time.

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: See my last comment. Examples are now generated from `Arbitrary` instances at least for keeping all the examples in sync with the API responses. We can iterate from there to also generate the example cURL commands and / or the associated endpoint :+1:.
For now however, I think it was necessary as the doc was already getting quite a few inconsistencies and it's not even merged :| ...

adinapoli-iohk: > Examples are now generated from Arbitrary instances at least for keeping all the examples in sync with the API responses.

Ah, but be careful there -- I cannot check all the details (in the middle of something else) but the reason why we added the `Example` typeclass was because not all the `Arbitrary` instances we had in the wallet yielded useful/valid data, so we had to use the `Example` typeclass to overcome it.

In this [linked issue](https://iohk.myjetbrains.com/youtrack/issue/CSL-2034) I give a (hopefully) excellent rationale for that 😉

wallet-new/server/Cardano/Wallet/API/V1/Swagger.hs
disassembler: What about `within a short delay`?

parsonsmatt: Ahh, I can't quite say *why* `in` feels wrong to me!

> X will happen in an hour

this feels right

> X will happen with an hour

this feels wrong

> X will happen in a delay

this feels wrong

> X will happen with a delay

this feels right

Hmm, I think it boils down to that `X` will happen after a delay, not during the delay. X can't happen *during* the delay, because if X is happening, then the delay is complete. 

After rereading this, `after` seems more appropriate than `with` or `in` anyway

adinapoli-iohk:
Looks solid, thanks @KtorZ .

Do not hate me but I think we should also remove this now-completely-stale line in our `README.md`:

>  The visualization at those URLs lets you play with the API by the mean of a _Try it out_ button
 made available for each endpoint. This will seemingly contact the node already running on your
 local machine with actual HTTP requests augmented with the parameters you provide!

Up to you if you want to do this now or in another PR (I'm happy with both solutions). 😉 

Will give it a spin locally and approve & merge shortly afterwards.

KtorZ:
*__/me__ kneels down to the CI gods with fingers crossed*

KtorZ:
Fixed a small oversight caught by @disassembler (wrong API response shown in example). Should be fine now, eventually!

adinapoli-iohk:
@KtorZ @parsonsmatt Can I ask you kind folks to add an item in your squad backlog that we need to remember backporting this one to `develop`? Probably we will do that automatically once we will ship 1.2.0 and merge the release back to develop, but just to keep ourselves honest, we shall no forget 😅

PR #2720, opened 2018-03-28
No template Haskell in update configuration

avieth:


update/Pos/Update/Configuration.hs
```diff
@@ -92,18 +89,7 @@ curSoftwareVersion = SoftwareVersion ourAppName (ccApplicationVersion updateConf
...
-             st :: Either Text ()
-             st = checkSystemTag tag
-             color c s = "\n" <> colorize c s <> "\n"
-         case st of Left e -> error . color Red . T.concat $
-                                  ["Current system tag could not be calculated: ", e]
-                    Right () -> do runIO . putStrLn . color Blue . T.concat $
-                                       ["Current system tag is: ", show tag]
-                                   TH.lift tag
-     )
+currentSystemTag = SystemTag (toText (osHelper buildOS ++ archHelper buildArch))
```
avieth: The template Haskell would fail if the system tag is too long (`checkSystemTag`).

Is this really necessary?

Cons:
- it causes linker problems
- it's template Haskell

Pros:
- system tag is guaranteed to be short enough

So what? If the system tag is too long, verification will fail at run time.

avieth: Related note: can anyone tell me why a system tag is even necessary? Who cares what platform the software is running on?

avieth: Is it because update proposals version the _software_, rather than the _protocol_?

domenkozar:
:bowing_man:

update/Pos/Update/Configuration.hs
domenkozar: this was always here for "one day when we package cardano for linux distros" which might be in 3 years so let's have the complexity then :)

edsko:
I don't fully understand what's going on, but when I tried to build with profiling enabled, stack failed on `cardano-sl-update` but `nix` and `cabal new-build` built it just fine, even with profiling (#2772).

coot:
Is there a reason what this PR didn't get merged? It seems that hydra build failed due to connection issues (I don't have rights to restart it).

PR #2725, opened 2018-03-29
[DEVOPS-754] AppVeyor: use cache-s3 --max-size=1600MB

rvl:
The `.stack-work` cache was expanding with stale data.

The `cache-s3 --max-size` option prevents build failures when upload volumes exceed AppVeyor limits.


rvl:
AppVeyor build still works, and I can see on the S3 bucket that it has successfully uploaded a `.stack-work` cache for this branch.

PR #2729, opened 2018-03-29
[CSL-2422] Create initial prototype for the WalletActiveLayer.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2422

Initial prototype.

### DIFF
--- a/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
+++ b/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
@@ -24,8 +24,8 @@ import qualified Cardano.Wallet.Server as Server
 import           Cardano.Wallet.Server.CLI (NewWalletBackendParams (..), RunMode,
                                             WalletBackendParams (..), isDebugMode,
                                             walletAcidInterval, walletDbOptions)
-import           Cardano.Wallet.WalletLayer (ActiveWalletLayer, PassiveWalletLayer,
-                                             bracketActiveWalletLayer)
+import           Cardano.Wallet.WalletLayer (ActiveWalletLayer)
+import           Cardano.Wallet.WalletLayer.Kernel (bracketActiveWallet)
 
 import           Data.Aeson
 import           Formatting (build, sformat, (%))
@@ -125,13 +125,12 @@ legacyWalletBackend WalletBackendParams {..} ntpStatus =
 -- TODO: no web socket support in the new wallet for now
 walletBackend :: (HasConfigurations, HasCompileInfo)
               => NewWalletBackendParams
-              -> PassiveWalletLayer Production
               -> Plugin Kernel.WalletMode
-walletBackend (NewWalletBackendParams WalletBackendParams{..}) passive =
+walletBackend (NewWalletBackendParams WalletBackendParams{..}) =
     first one $ worker walletServerOuts $ \diffusion -> do
       env <- ask
       let diffusion' = Kernel.fromDiffusion (lower env) diffusion
-      bracketActiveWalletLayer passive diffusion' $ \active ->
+      bracketActiveWallet diffusion' $ \active ->
         walletServeImpl
           (getApplication active)
           walletAddress
@@ -139,7 +138,7 @@ walletBackend (NewWalletBackendParams WalletBackendParams{..}) passive =
           (if isDebugMode walletRunMode then Nothing else walletTLSParams)
           Nothing
   where
-    getApplication :: forall m. ActiveWalletLayer m -> Kernel.WalletMode Application
+    getApplication :: ActiveWalletLayer Production -> Kernel.WalletMode Application
     getApplication active = do
       logInfo "New wallet API has STARTED!"
       return $ withMiddleware walletRunMode $
--- a/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
+++ b/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
@@ -124,7 +124,7 @@ legacyWalletBackend WalletBackendParams {..} ntpStatus =
 -- TODO: no web socket support in the new wallet for now
 walletBackend :: (HasConfigurations, HasCompileInfo)
               => NewWalletBackendParams
-              -> Kernel.PassiveWallet
+              -> Kernel.PassiveWallet Production
               -> Plugin Kernel.WalletMode
 walletBackend (NewWalletBackendParams WalletBackendParams{..}) passive =
     first one $ worker walletServerOuts $ \diffusion -> do
@@ -138,7 +138,7 @@ walletBackend (NewWalletBackendParams WalletBackendParams{..}) passive =
           (if isDebugMode walletRunMode then Nothing else walletTLSParams)
           Nothing
   where
-    getApplication :: Kernel.ActiveWallet -> Kernel.WalletMode Application
+    getApplication :: Kernel.ActiveWallet Production -> Kernel.WalletMode Application
     getApplication active = do
       logInfo "New wallet API has STARTED!"
       return $ withMiddleware walletRunMode $
--- a/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
+++ b/wallet-new/server/Cardano/Wallet/Server/Plugins.hs
@@ -17,7 +17,6 @@ import           Universum
 
 import           Cardano.Wallet.API as API
 import           Cardano.Wallet.API.V1.Errors (WalletError (..))
-import qualified Cardano.Wallet.Kernel as Kernel
 import qualified Cardano.Wallet.Kernel.Diffusion as Kernel
 import qualified Cardano.Wallet.Kernel.Mode as Kernel
 import qualified Cardano.Wallet.LegacyServer as LegacyServer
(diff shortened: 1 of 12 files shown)


wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -124,21 +125,20 @@ legacyWalletBackend WalletBackendParams {..} ntpStatus =
...
               => NewWalletBackendParams
-              -> Kernel.PassiveWallet
               -> Plugin Kernel.WalletMode
-walletBackend (NewWalletBackendParams WalletBackendParams{..}) passive =
+walletBackend (NewWalletBackendParams WalletBackendParams{..}) =
     first one $ worker walletServerOuts $ \diffusion -> do
       env <- ask
       let diffusion' = Kernel.fromDiffusion (lower env) diffusion
-      Kernel.bracketActiveWallet passive diffusion' $ \active ->
+      bracketActiveWallet diffusion' $ \active ->
```
edsko: It is important that we initialize the passive wallet only once, so it's not correct to re-initialize it when we initialize the active wallet. That `passive` argument should not be removed here.

wallet-new/server/Main.hs
```diff
@@ -105,20 +106,28 @@ actionWithNewWallet sscParams nodeParams params =
...
+    runNodeWithInit nr =
+        let (ActionSpec f, outs) = runNode nr plugins
+         in (ActionSpec $ \s -> f s, outs)
 
     -- TODO: Don't know if we need any of the other plugins that are used
     -- in the legacy wallet (see 'actionWithWallet').
-    plugins :: Kernel.PassiveWallet -> Plugins.Plugin Kernel.Mode.WalletMode
-    plugins w = mconcat [ Plugins.walletBackend params w ]
+    plugins :: Plugins.Plugin Kernel.Mode.WalletMode
+    plugins = mconcat [ Plugins.walletBackend params ]
```
edsko: This means that here too that `w` argument should not be removed.

wallet-new/src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -0,0 +1,43 @@
...
+            { pwlGetWalletAddresses  = error "Not implemented!"
+            , pwlGetWalletMeta       = error "Not implemented!"
+            }
+
+    -- | Empty log function.
+    logFunction _ = pure $ pure ()
+
+-- | Initialize the active wallet.
+-- The active wallet is allowed all.
+bracketActiveWallet
```
edsko: Should take the passive wallet layer as argument.

wallet-new/src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -0,0 +1,43 @@
...
+    :: forall m n a. (MonadMask m, Monad n)
+    => WalletDiffusion
+    -> (ActiveWalletLayer n -> m a) -> m a
+bracketActiveWallet walletDiffusion =
+    bracket
+        (bracketPassiveWallet $ \walletPassiveLayer -> return ActiveWalletLayer {..})
```
edsko: (... which means that it shouldn't call `bracketPassiveWallet`)

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -0,0 +1,42 @@
...
+        , pwlGetWalletMeta       = \cIdWal -> askWalletSnapshot >>= \ws -> pure $ getWalletMeta ws =<< (migrate cIdWal)
+        }
+
+-- | Initialize the active wallet.
+-- The active wallet is allowed all.
+bracketActiveWallet
+    :: forall ctx m n a. (MonadMask m, WalletDbReader ctx n, MonadIO n, MonadThrow n)
+    => WalletDiffusion
+    -> (ActiveWalletLayer n -> m a) -> m a
+bracketActiveWallet walletDiffusion =
```
edsko: Same here.

wallet-new/src/Cardano/Wallet/WalletLayer/QuickCheck.hs
```diff
@@ -0,0 +1,46 @@
...
+    liftedGen :: forall b. (MonadIO n, Arbitrary b) => n b
+    liftedGen = liftIO . generate $ arbitrary
+
+-- | Initialize the active wallet.
+-- The active wallet is allowed all.
+bracketActiveWallet
+    :: forall m n a. (MonadMask m, MonadIO n)
+    => WalletDiffusion
+    -> (ActiveWalletLayer n -> m a) -> m a
+bracketActiveWallet walletDiffusion =
```
edsko: And here.

wallet-new/server/Main.hs
```diff
@@ -105,19 +106,29 @@ actionWithNewWallet sscParams nodeParams params =
...
   where
-    mainAction w = runNodeWithInit w $
-        liftIO $ Kernel.init w
-
-    runNodeWithInit w init nr =
+    mainAction
+        :: PassiveWalletLayer Production
+        -> NodeResources ext
+        -> (ActionSpec Kernel.Mode.WalletMode (), OutSpecs)
+    mainAction w nr = runNodeWithInit w nr
```
edsko: Nice, thanks for adding the type sig :+1:

wallet-new/src/Cardano/Wallet/WalletLayer/Kernel.hs
```diff
@@ -0,0 +1,46 @@
...
+  where
+    -- | TODO(ks): Currently not implemented!
+    passiveWalletLayer _wallet =
+        pure $ PassiveWalletLayer
+            { pwlGetWalletAddresses  = error "Not implemented!"
+            , pwlGetWalletMeta       = error "Not implemented!"
+            }
+
+    -- | Empty log function.
+    logFunction _ = pure $ pure ()
```
edsko: Let's make this an argument again (where we this bracket function we construct it anyway).

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -0,0 +1,28 @@
...
+import           Cardano.Wallet.API.V1.Types (WalletId)
+import           Pos.Wallet.Web.ClientTypes (CWalletMeta)
+
+import           Cardano.Wallet.Kernel.Diffusion (WalletDiffusion (..))
+
+
+-- | The passive wallet (data) layer. See @PassiveWallet@.
+data PassiveWalletLayer m = PassiveWalletLayer
+    { pwlGetWalletAddresses :: m [WalletId]
+    , pwlGetWalletMeta      :: WalletId -> m (Maybe CWalletMeta)
```
adinapoli-iohk: @ksaric We can postpone this after we merge, but perhaps it might be a nicer pattern to port all the functions which can fail to yield something like `m (Either WalletError a)`.

Follow-up: Why are we using `CWalletMeta` here? `ClientTypes` is effectively deprecated, so we should never leak V0 stuff into the interface. What am I missing?

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -0,0 +1,45 @@
...
+    :: forall ctx m n a. (MonadMask m, WalletDbReader ctx n, MonadIO n, MonadThrow n)
+    => (PassiveWalletLayer n -> m a) -> m a
+bracketPassiveWallet =
+    bracket
+        (pure passiveWalletLayer)
+        (\_ -> return ())
+  where
+    passiveWalletLayer :: PassiveWalletLayer n
+    passiveWalletLayer = PassiveWalletLayer
+        { pwlGetWalletAddresses  = askWalletSnapshot >>= \ws -> migrate $ getWalletAddresses ws
```
adinapoli-iohk: You can probably write this as `askWalletSnapshot >>= migrate . getWalletAddresses`, cannot you?

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -0,0 +1,28 @@
...
+
+import           Cardano.Wallet.API.V1.Types (WalletId)
+import           Pos.Wallet.Web.ClientTypes (CWalletMeta)
+
+import           Cardano.Wallet.Kernel.Diffusion (WalletDiffusion (..))
+
+
+-- | The passive wallet (data) layer. See @PassiveWallet@.
+data PassiveWalletLayer m = PassiveWalletLayer
+    { pwlGetWalletAddresses :: m [WalletId]
```
adinapoli-iohk: I find "walletAddresses" a terribly confusing name. I would go with simply `pwlGetWalletIds`.

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
adinapoli-iohk: > Converting all these to Either will result in the same semantics and I'm not sure what would we gain.

In the general case you would gain an explanation on why the computation failed, which can be useful to take more informative decisions (for example to show better error messages to the user).

adinapoli-iohk:
@jmitchell I'm merging this one by the virtue of the fact that:

- 2 people approved it
- CI actually [built it successfully](https://ci.appveyor.com/project/input-output/cardano-sl/build/1.0.23640).

:shipit:

PR #2732, opened 2018-03-29
[CSL-2247] Extend the V1 API to report restoration progress

adinapoli-iohk:
## Summary

This PR (which is being opened against `feature/CBR-90` is virtually a feature-complete incarnation of CBR-90. *The only thing missing are DB migrations, which will be delivered with CSL-2429. This doesn't hold up the Daedalus team from start integrating the feature, as the V1 API provided here won't change*.

## Description

This PR builds upon CSL-2246 (for which the reader is encouraged to defer review up until a PR for `feature/CBR-90` won't land, to get the full picture). In a nutshell we extend the `WalletInfo` section of the `WalletStorage` to keep around some `SyncStatistics` which can be used by the V1 API to compute interesting numbers about a wallet `syncState`, for example:

- `SyncState` (Syncing? Restoring? etc)
- `SyncPercentage` (0 .. 100%)
- `EstimatedCompletionTime` (in milliseconds)
- `SyncThroughput` (in blocks/s)

Notably, `SyncProgress` has been renamed to `SyncPercentage` and `SyncProgress` is now a richer type.

## Screenshots

screen shot 2018-03-29 at 15 43 44

## QA Steps

Provided as part of `feature/CBR-90`, but for @darko-mijic and @nikolaglumac the best course of action is to attach the `cardano-node` to the staging cluster, like this:

```
stack build cardano-sl-wallet-new
stack exec cardano-node -- --topology=docs/network/example-topologies/mainnet-staging.yaml --configuration-key mainnet_dryrun_full --db-path ../staging-db
```

And finally play with the wallet. It's recommended to import a staging DB so that the restoration is realistic. **Note that transactions cannot be sent during a restoration (due to known limitations we discussed) so doing so will incur in unexpected behaviour.**

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,7 @@ before_test:
     }
 
 # Install OpenSSL 1.0.2 (see https://github.com/appveyor/ci/issues/1665)
-- ps: (New-Object Net.WebClient).DownloadFile('https://slproweb.com/download/Win64OpenSSL-1_0_2n.exe', "$($env:USERPROFILE)\Win64OpenSSL.exe")
+- ps: (New-Object Net.WebClient).DownloadFile('https://slproweb.com/download/Win64OpenSSL-1_0_2o.exe', "$($env:USERPROFILE)\Win64OpenSSL.exe")
 - ps: cmd /c start /wait "$($env:USERPROFILE)\Win64OpenSSL.exe" /silent /verysilent /sp- /suppressmsgboxes /DIR=C:\OpenSSL-Win64-v102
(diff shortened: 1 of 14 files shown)


adinapoli-iohk:
@nikolaglumac merging this one so I can finalise the "big PR in the sky" as part of `feature/CBR-90`.

Fear not, I do not plan to add any new functionality (if not DB migrations), so you can happily and peacefully keep working on this branch, as the V1 API won't change. 😉 

Shout if you have questions/doubts/concerns.

PR #2734, opened 2018-03-29
wallet-new: Add Setup.hs and default.nix files for easing Cabal usage

k0001:
See `wallet-new/default.nix` documentation.

### DIFF
--- a/wallet-new/Setup.hs
+++ b/wallet-new/Setup.hs
@@ -1,2 +1,4 @@
+#! /usr/bin/env nix-shell
+#! nix-shell ./default.nix -i runghc
 import Distribution.Simple
 main = defaultMain
--- a/wallet-new/default.nix
+++ b/wallet-new/default.nix
@@ -1,0 +1,10 @@
+# Running `nix-shell` on this file puts you in an environment where all the
+# dependencies needed to work on `wallet-new` using `Cabal` are available.
+#
+# Running `nix-build` builds the `wallet-new` code.
+let
+  drv = (import ../. {}).cardano-sl-wallet-new;
+in
+  if ((import <nixpkgs> {}).stdenv.lib.inNixShell)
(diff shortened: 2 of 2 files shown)


k0001:
Actually, wait, the passing of `args` is wrong.

k0001:
Due to #2736 profiling and debugging can't be enabled by default without a deeper investigation, so I am submitting this without those features.

adinapoli-iohk:
Thanks @k0001 ! Sorry for the noob question (I'm relatively new to `nix`) but what would happen if the user doesn't have `nix` and the `nix-shell` available? Would the two new lines in `Setup.hs` cause any disruption to folks building with plain `stack` or even `cabal new-build`?

k0001:
@adinapoli-iohk that's a good question. 

Nothing would happen, shebang lines are always ignored by GHC. Although, more generally, the entire `Setup.hs` is currently ignored because it says `build-type: Simple` in the `.cabal` file.

The intention with this change is that I can run stuff like `./Setup.hs repl` or `nix-shell` in this folder and it works.

adinapoli-iohk:
> Nothing bad would happen; shebang lines are always ignored by GHC. Although, more generally, the entire Setup.hs is currently ignored because it says build-type: Simple in the .cabal file.

@k0001 Crystal clear (and TIL!)

I have restarted the build for Appveyor, which seems to have failed irrespective of your work -- hopefully rebuilding the commit will do the trick.

adinapoli-iohk:
Ah, I have just clarified this with the DevOps team, and indeed the Appveyor failure is a known one, fixed as part of this PR: https://github.com/input-output-hk/cardano-sl/pull/2714

Would you mind rebasing your PR against `develop`  @k0001 ? 😉

k0001:
@adinapoli-iohk rebased as requested.

PR #2735, opened 2018-03-29
use stack lts-11.2

avieth:
*Vast* majority of changes are in the generated `pkgs/default.nix`. Any kind soul willing to review this would only be in for ~450 additions and ~450 deletions, all of which are basically trivial.


avieth:
@domenkozar I think I need to bump the hackage snapshot but I don't know how to find the right value for it.

parsonsmatt:
@avieth I just bumped the hackage snapshot to get a newer `swagger2` version, and I just modified the day/month value. It worked fine. But it might, in fact, be Wrong

avieth:
> @avieth I just bumped the hackage snapshot to get a newer swagger2 version, and I just modified the day/month value. It worked fine. But it might, in fact, be Wrong

Wow I think it worked. Thanks!

avieth:
As always, appveyor fails. Merging anyway

PR #2745, opened 2018-04-03
[CBR-51] use stack lts-11.2 (#2735)

avieth:
**The vast majority of the changes are in `default.nix`. This is not actually a big change.**

Also satisfies CSL-2500 (new HLint configuration).


avieth:
@jmitchell What do you think? Is this a good idea from a devops perspective?

avieth:
Some package update has apparently broken the auxx lexer. Or, perhaps it never handled strings properly (it fails on specimens which use numeric escape sequences).

Do we have to maintain this part of the system? Does anybody actually use the auxx DSL?

avieth:
@jmitchell Ok, I'm glad I didn't fix the lexer in vain.

avieth:
@jmitchell It seems there are appveyor problems. Did I break something?

avieth:
To fix the lexer I made it so that `Text` is used rather than `String`, they are quoted (same as before), but internal quotations are expressed as double quotes rather than escaped with `\\` because this is simpler and easier to implement. Perhaps this will break some things.

avieth:
Release cutoff is April 13. Think we could land this one for that release? I'd rather not keep it sitting here for 2 weeks. @dcoutts @jmitchell @jcmincke what do you think?

erikd:
@avieth Did you see Duncan's message in #sl-network about not merging this until after Friday?

parsonsmatt:
Since we have a feature-freeze and a `1.2.0` branch, can we go ahead and merge this into `develop`?

update/Pos/Arbitrary/Update/Poll.hs
```diff
@@ -25,45 +24,45 @@ import           Pos.Update.Poll.Types (BlockVersionState (..), ConfirmedProposa
...
 
-instance (Arbitrary HeaderHash, HasProtocolConstants, HasProtocolMagic) => Arbitrary PollState where
+instance (HasProtocolConstants, HasProtocolMagic) => Arbitrary PollState where
     arbitrary = do
         ps <- genericArbitrary
         return (ps & psActivePropsIdx %~ HM.filter (not . null))
     shrink = genericShrink
 
-instance (Arbitrary HeaderHash, HasProtocolConstants, HasProtocolMagic) => Arbitrary USUndo where
+instance (HasProtocolConstants, HasProtocolMagic) => Arbitrary USUndo where
```
parsonsmatt: It's shocking to see so many instance constraints like `Arbitrary HeaderHash`. All these do is defer the "missing constraint" error to use sites. Is there a reason that they're there?

util/Pos/Util/QuickCheck/Property.hs
```diff
@@ -62,15 +62,20 @@ qcIsRight :: Show a => Either a b -> Property
 qcIsRight (Right _) = property True
 qcIsRight (Left x)  = qcFail ("expected Right, got Left (" <> show x <> ")")
 
+-- TBD would it not be sufficient to just use 'Traversable'? Does the
+-- 'Container' class from 'universum' add any value?
```
parsonsmatt: It's *kinda* like `mono-traversable`'s class in that you can use `fold` and friends on Text, ByteString, etc

update/Pos/Arbitrary/Update/Poll.hs
avieth: I've no idea. They've been caught now thanks to the new GHC version.

avieth:
@iohk-devops I could use some help on this one.

1. Any advice on how to fix the failing builds?
2. What should I put for `sha256unpacked` in `nixpkgs-src.json`? Does my change there break something?

parsonsmatt:
Benchmarks and tests had some compilation errors. I'm pushing a commit soon to fix those.

parsonsmatt:
I don't know how to pin `cabal-install-2.2.0.0` on this guy :\

erikd:
I think you may want to revert that last `pkgs/hfsevents.nix` change. I think that issue was fixed in #2835 which has already been merged.

auxx/Main.hs
```diff
@@ -119,7 +119,7 @@ action opts@AuxxOptions {..} command = do
                         { acRealModeContext = realModeContext
                         , acTempDbUsed = tempDbUsed }
                 lift $ runReaderT auxxAction auxxContext
-        let vssSK = unsafeFromJust $ npUserSecret nodeParams ^. usVss
+        let vssSK = fromMaybe (error "no user secret given") $ npUserSecret nodeParams ^. usVss
         let sscParams = CLI.gtSscParams cArgs vssSK (npBehaviorConfig nodeParams)
```
erikd: :100:

lib/src/Pos/Client/CLI/Options.hs
```diff
@@ -50,6 +50,10 @@ commonArgsParser = do
     configurationOptions <- configurationOptionsParser
     pure CommonArgs{..}
 
+-- To get rid of abiguity.
```
erikd: Would be nice to have just a little more context there about which ambiguity we're getting rid of.

avieth:
CI failing with

`hash mismatch in downloaded path ‘/nix/store/yidw31l5afkbjp8caq63xr2a2qcmmjh2-x509-1.7.3’: expected a74cb28ee3f9fb325ae8dc387f20414f419a89006f592acd7d102b359cdfb20d, got 78cd24f4ac3f48da0968325ca3b38b67bfdd2adc17aa1ca62a28a12f83aaef83`

@iohk-devops Is this my fault? How can I fix it?

avieth:
Now there are purescript compilation errors in CI :(

avieth:
@iohk-devops It could be that the purescript version it's compiling requires a GHC newer than 8.0.2. How can I make the CI script use a newer version? Do we need to go to a later nixpkgs version?

avieth:
@cleverca22 has helped to get around the failing purescript build (just remove it) but now I'm getting a bunch of HLint errors for things I didn't even change.

avieth:
> now I'm getting a bunch of HLint errors for things I didn't even change.

New version of HLint perhaps, with evil new draconian default hints?

parsonsmatt:
The new version of hlint uses a Yaml configuration instead of the `HLint.hs` configuration. We'll either need to port the `HLint.hs` to the new yaml-based config or hold back and use the old hlint.

avieth:
> The new version of hlint uses a Yaml configuration instead of the HLint.hs configuration. We'll either need to port the HLint.hs to the new yaml-based config or hold back and use the old hlint.

Oh no. According to [Neil's release announcement](http://neilmitchell.blogspot.ca/2017/04/hlint-20-with-yaml-configuration.html) the Haskell config file hasn't been dropped yet

> At the moment the Haskell and YAML configuration files are both supported. However, I'll be phasing out the Haskell configuration in the future, and encourage people to move to the YAML. The conversion should be pretty simple.

avieth:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2500

avieth:
So close. Now it fails because it tries to build cabal-install-2.0.0.1, apparently required for cardano-sl-auxx, but which cannot be used with async > 2.2.

avieth:
> So close. Now it fails because it tries to build cabal-install-2.0.0.1, apparently required for cardano-sl-auxx, but which cannot be used with async > 2.2.

Already taken care of :) Thanks @cleverca22

avieth:
Obeying HLint suggestions about unnecessary import hiding caused ambiguity errors! Shame

avieth:
Linking problems now with the hyda stuff

```
<no location info>: error:
    <command line>: can't load .so/.DLL for: /private/tmp/nix-build-cardano-sl-wallet-new-1.1.1.drv-0/wallet-new/dist/build/libHScardano-sl-wallet-new-1.1.1-A8UenEHryjl7cIGMmBwMCj-ghc8.2.2.dylib (dlopen(/private/tmp/nix-build-cardano-sl-wallet-new-1.1.1.drv-0/wallet-new/dist/build/libHScardano-sl-wallet-new-1.1.1-A8UenEHryjl7cIGMmBwMCj-ghc8.2.2.dylib, 5): no suitable image found.  Did find:
	/private/tmp/nix-build-cardano-sl-wallet-new-1.1.1.drv-0/wallet-new/dist/build/libHScardano-sl-wallet-new-1.1.1-A8UenEHryjl7cIGMmBwMCj-ghc8.2.2.dylib: malformed mach-o: load commands size (33224) > 32768)
```

Have I done something wrong?

avieth:
@iohk-devops I take it we're aware of these OS X linker problems

https://github.com/NixOS/nixpkgs/issues/22810
https://github.com/NixOS/nixpkgs/pull/27536
https://github.com/NixOS/nixpkgs/pull/25537

Are those failing hydra builds using nixpkgs which include a fix? Do we need to re-open an issue?

avieth:
Apparently [this](https://github.com/NixOS/nixpkgs/pull/32181) fixes our purescript build problem.

I'd like to try bumping the nixpkgs revision in `nixpkgs-src.json`, but I can't figure out what `sha256unpacked` means, and how it's related to `sha256`. One of them is the digest of a tarball and the other of what?

avieth:
> - sha256 should be the digest of the tarball (nix-prefetch <url>)
> - sha256unpacked the digest of the unpacked tarball (nix-prefetch --unpack <url>)

Ah, makes good sense :)

avieth:
@iohk-devops I tried latest nixpkgs release 18.03 and now

```
<no location info>: error:
  | ghc: panic! (the 'impossible' happened)
  | (GHC version 8.2.2 for x86_64-unknown-linux):
  | Dynamic linker not initialised
```

Possibly related to https://ghc.haskell.org/trac/ghc/ticket/13607

Any ideas?

cleverca22:
i think 18.03 uses static linking for everything now

avieth:
> i think 18.03 uses static linking for everything now

https://github.com/NixOS/nixpkgs/issues/29011

So do we need to flip on `enableSharedLibraries`?

avieth:
@angerman this branch has just been rebased on `develop`.

avieth:
@iohk-devops cassava is failing due to missing dependencies. I checked the listed dependencies of it in `pkgs/default.nix` and they do not match those listed on its [haddock page](https://hackage.haskell.org/package/cassava). Could this be a bug in nix/nixpkgs/cabal2nix?

jmitchell:
@avieth, at this point I'm also surprised by the mismatched dependency version constraints compared to `cassava`'s cabal file. For other's sake, here's the relevant bit of the build log:

```
Configuring cassava-0.5.1.0...
--
  | CallStack (from HasCallStack):
  | die', called at libraries/Cabal/Cabal/Distribution/Simple/Configure.hs:948:20 in Cabal-2.0.1.0:Distribution.Simple.Configure
  | configureFinalizedPackage, called at libraries/Cabal/Cabal/Distribution/Simple/Configure.hs:470:12 in Cabal-2.0.1.0:Distribution.Simple.Configure
  | configure, called at libraries/Cabal/Cabal/Distribution/Simple.hs:570:20 in Cabal-2.0.1.0:Distribution.Simple
  | confHook, called at libraries/Cabal/Cabal/Distribution/Simple/UserHooks.hs:67:5 in Cabal-2.0.1.0:Distribution.Simple.UserHooks
  | configureAction, called at libraries/Cabal/Cabal/Distribution/Simple.hs:174:19 in Cabal-2.0.1.0:Distribution.Simple
  | defaultMainHelper, called at libraries/Cabal/Cabal/Distribution/Simple.hs:119:27 in Cabal-2.0.1.0:Distribution.Simple
  | defaultMain, called at Setup.hs:2:8 in main:Main
  | Setup: Encountered missing dependencies:
  | bytestring >=0.9.2 && <0.10.4, text-short ==0.1.*
```

Unless I'm mistaken the only way `text-short ==0.1.*` applies as a direct dependency is when `bytestring >= 0.10.4` also applies, but that doesn't align with the constraint in the error.

A few possibilities that come to mind, none of which I've had the opportunity to investigate yet:
1. stack2nix was developed with 8.0.2 in mind (see [this](https://github.com/input-output-hk/stack2nix/blob/abd199510eb38c442e14fd82677f9cc1f7430cfb/src/Stack2nix/Render.hs#L40-L69)), and needs to be improved to handle this properly.
2. cabal2nix bug as you suggest.
    - See what cabal2nix emits when run directly on `cassava-0.5.1.0`, and how it compares to the aforementioned constraints.
3. problem in stack / stack2nix where cardano's other dependencies are affecting constraints for `cassava`, and the constraint solving is turning up contradictions.

I hope that helps.

avieth:
> See what cabal2nix emits when run directly on cassava-0.5.1.0, and how it compares to the aforementioned constraints.

Here it is

```
{ mkDerivation, array, attoparsec, base, bytestring, containers
, deepseq, hashable, HUnit, Only, QuickCheck, quickcheck-instances
, scientific, stdenv, test-framework, test-framework-hunit
, test-framework-quickcheck2, text, text-short
, unordered-containers, vector
}:
mkDerivation {
  pname = "cassava";
  version = "0.5.1.0";
  src = ./.;
  configureFlags = [ "-f-bytestring--lt-0_10_4" ];
  libraryHaskellDepends = [
    array attoparsec base bytestring containers deepseq hashable Only
    scientific text text-short unordered-containers vector
  ];
  testHaskellDepends = [
    attoparsec base bytestring hashable HUnit QuickCheck
    quickcheck-instances scientific test-framework test-framework-hunit
    test-framework-quickcheck2 text unordered-containers vector
  ];
  homepage = "https://github.com/hvr/cassava";
  description = "A CSV parsing and encoding library";
  license = stdenv.lib.licenses.bsd3;
}
```

avieth:
Ok I have it using an updated `stack2nix` and `cabal2nix` so that the generated `pkgs/default.nix` is legit. Next issue: Serokell's `o-clock` package (a dependency of their `log-warper`) won't build with ghc 8.0.2, but we're stuck to 8.0.2 due to some mac linker patch.

avieth:
It seems purescript won't build on nixpkgs 17.09. It uses optparse-applicative < 0.14, but purescript 0.11.7 assumes >= 0.14 (`strOption` has a different type). Is this a bug in nixpkgs?

Anyway, I've bumped it up to 18.03 and now things should go. I can't confirm locally, because if I try to build the explorer frontend I either run out of memory, or later run out of disk space.

avieth:
Of course, now I remember why we can't upgrade to 18.03:

```
Building library for cardano-sl-db-1.1.1..
[ 1 of 14] Compiling Pos.DB.Error     ( Pos/DB/Error.hs, dist/build/Pos/DB/Error.o )
[ 2 of 14] Compiling Pos.DB.Class     ( Pos/DB/Class.hs, dist/build/Pos/DB/Class.o )
[ 3 of 14] Compiling Pos.DB.Functions ( Pos/DB/Functions.hs, dist/build/Pos/DB/Functions.o )
[ 4 of 14] Compiling Pos.DB.Misc.Common ( Pos/DB/Misc/Common.hs, dist/build/Pos/DB/Misc/Common.o )
[ 5 of 14] Compiling Pos.DB.Pure      ( Pos/DB/Pure.hs, dist/build/Pos/DB/Pure.o )

<no location info>: error:
    ghc: panic! (the 'impossible' happened)
  (GHC version 8.2.2 for x86_64-unknown-linux):
        Dynamic linker not initialised

Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug

[ 6 of 14] Compiling Pos.DB.Rocks.Types ( Pos/DB/Rocks/Types.hs, dist/build/Pos/DB/Rocks/Types.o )

<no location info>: error:
    <command line>: can't load .so/.DLL for: /nix/store/3qlfrfbyzs4ss3ykz9ls610v0xqbzqkw-rocksdb-haskell-ng-0.0.0/lib/ghc-8.2.2/x86_64-linux-ghc-8.2.2/libHSrocksdb-haskell-ng-0.0.0-8hm3hwqqbawC8aDw223BZf-ghc8.2.2.so (/nix/store/ivjclr8nncmsn4nf8l9wmljmn79ma81g-jemalloc-5.0.1/lib/libjemalloc.so.2: cannot allocate memory in static TLS block)
builder for '/nix/store/4kyxbiqhnqkfdfwgzg8lj5hlfmgy8nmr-cardano-sl-db-1.1.1.drv' failed with exit code 1
cannot build derivation '/nix/store/l44yjm447vxzr2vvq566lzwiic68gdid-cardano-sl-1.1.1.drv': 1 dependencies couldn't be built
[31;1merror:[0m build of '/nix/store/l44yjm447vxzr2vvq566lzwiic68gdid-cardano-sl-1.1.1.drv' failed
```

avieth:
Linker problem was to do with jemalloc. I built cardano-sl-db using `-j1` and the true linker error materialised! `cannot allocate memory in static TLS block` https://github.com/jemalloc/jemalloc/issues/937

So I used my limited knowledge of nix idioms to have rocksdb build with gperftools rather than jemalloc and now cardano-sl-db links. @jmitchell @cleverca22 please critique the way I did this (top of `default.nix`). I'm sure there is a nicer way.

avieth:
```
Building library for cardano-sl-infra-1.1.1..
src/tcmalloc.cc:284] Attempt to free invalid pointer 0x7fffa00876f0
/nix/store/l3nz0167a4xhb78ldcjns5p4dwvlm1cw-stdenv/setup: line 1223:  8774 Aborted                 ./Setup build
builder for '/nix/store/3kq03si7hlrp8slwlpfcp096v75axsrq-cardano-sl-infra-1.1.1.drv' failed with exit code 134
cannot build derivation '/nix/store/gd74v9gfrf11xidzcxpzpdrryg0lir5b-cardano-sl-1.1.1.drv': 1 dependencies couldn't be built
error: build of '/nix/store/gd74v9gfrf11xidzcxpzpdrryg0lir5b-cardano-sl-1.1.1.drv' failed
```

(╯°□°）╯︵ ┻━┻

jemalloc doesn't work, tcmalloc doesn't work

avieth:
Somebody flipped on `-fwarn-redundant-constraints` in `develop`, I rebased, and now it's killing my builds. I build locally using stack, not nix, and it doesn't raise the warnings. It does if I build locally with nix, but that eats all of my 8gb of memory. Woe is me. Nevertheless, I'm confident that buildkite will pass once I sort out all of these redundant constraints. The last remaining obstacle: appveyor, windows, and the ridiculous path length limit.

PR #2748, opened 2018-04-04
[DEVOPS-13] Linux wallet configuration key

deepfire:
_Contents_

0. We now have a Linux wallet -- which needs its own configuration keys.  Add them.
0. ~Mainnet now runs with block minor version 1 -- set that as baseline.~

### DIFF
--- a/lib/configuration.yaml
+++ b/lib/configuration.yaml
@@ -14752,6 +14752,16 @@ mainnet_wallet_macos64: &mainnet_wallet_macos64
       bvMinor: 1
       bvAlt: 0
 
+mainnet_wallet_linux: &mainnet_wallet_linux
+  <<: *mainnet_full
+  update:
+    applicationName: csl-daedalus
+    applicationVersion: 0
+    lastKnownBlockVersion:
+      bvMajor: 0
+      bvMinor: 1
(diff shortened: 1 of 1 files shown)


rvl:
Could the yaml "inheritance" be used so that the `lastKnownBlockVersion` records don't have to be duplicated everywhere? Wouldn't the ideal case be where it only needs to be defined twice -- for mainnet and staging.

deepfire:
@rvl, good question -- I'd like the core team to weigh in on the desired outcome, though.

The discussion is in https://iohk.myjetbrains.com/youtrack/issue/CSL-2436

cleverca22:
the block minor version also appears to impact a lot of the tests, which are now failing across the board

deepfire:
Dropped the minor version bump, as per consensus it's more-or-less irrelevant: https://iohk.myjetbrains.com/youtrack/issue/CSL-2436

jmitchell:
AppVeyor is green. Merging.

- https://ci.appveyor.com/project/input-output/cardano-sl/build/1.0.23664
- https://ci.appveyor.com/project/input-output/cardano-sl/build/1.0.23663

PR #2749, opened 2018-04-04
[CSL-2379] Use OOP in wallet spec

edsko:
This uses the approach describes in http://www.well-typed.com/blog/2018/03/oop-in-haskell/ to model the wallet. I think it worked out very well, the relation between the various models is now much clearer.

This closes https://iohk.myjetbrains.com/youtrack/issue/CSL-2379.


edsko:
Ok, since this hasn't been reviewed yet, I think there's one dimension in which I can improve this, let me do that first.

edsko:
Well, so I'm not totally sure we want to push this through. But I had a thought on how we can actually model that, say, the wallet with rollback supports more methods than the wallet without. See https://gist.github.com/edsko/c001d808f58be4f18c96db6a89302c58 for a sketch of how this would work. For the wallet it would look something like

``` haskell
data Wallet ext h a = Wallet {
      -- MAIN API

      -- | Return the total balance of the wallet (see 'available')
      totalBalance :: Value

      -- | Return the available balance of the wallet (see 'total')
    , availableBalance :: Value

      -- | Notify the wallet of a new block
    , applyBlock :: Block h a -> Wallet ext h a

-- (etc)

      -- | Extensions
    , walletExt :: ext
    }

data Rollback ext h a = Rollback {
      -- | Rollback
      rollback' :: WithRollback ext h a

      -- | Further extensions
    , rollbackExt :: ext
    }

type WithRollback ext h a = Wallet (Rollback ext h a) h a

rollback :: WithRollback ext h a -> WithRollback ext h a
rollback = rollback' . walletExt

data ExpectedUtxo ext h a = ExpectedUtxo {
      -- | The expected Utxo
      expectedUtxo' :: Utxo h a

      -- | Minimum balance
    , minimumBalance' :: Value

      -- | Further extensions
    , withExpectedUtxoExt :: ext
    }

type WithExpectedUtxo ext h a = WithRollback (ExpectedUtxo ext h a) h a

expectedUtxo :: WithExpectedUtxo ext h a -> Utxo h a
expectedUtxo = expectedUtxo' . withExpectedUtxoExt . walletExt

minimumBalance :: WithExpectedUtxo ext h a -> Value
minimumBalance = minimumBalance' . withExpectedUtxoExt . walletExt
```

We can then also express in stuff like invariants what part of the wallet API the invariant looks at. 

I have some WIP pushing this halfway through; I think it may be worth it but it does add yet another type parameter so perhaps others consider it overkill. @adinapoli-iohk or @dcoutts , let me know your thoughts.

edsko:
(Mind, we can certainly merge this as is now and address that later.)

edsko:
I think we should merge this as-is. Added a small sanity check to make sure stuff was working as expected.

adinapoli-iohk:
@edsko We just need to please the CI (I think it's the usual business of regenerating `pkgs/default.nix` and then we're ready to hit that merge button methinks 😉

PR #2750, opened 2018-04-04
[CSL-2347] Implement golden test for API shape

parsonsmatt:
This PR implements a golden test for the API shape. If the API representation changes, then the test fails, and prompts the programmer to overwrite the old file.

The file is checked in to version control, so it becomes extra obvious when the route shape has changed.


wallet-new/test/APISpec.hs
```diff
@@ -157,18 +162,39 @@ testV1Context =
...
-            serverSatisfies (Proxy @V1.API) burl stdArgs predicates
+        xdescribe "Servant API Properties" $ do
+            it "V0 API follows best practices & is RESTful abiding" $ do
+                ddl <- D.dummyDiffusionLayer
+                withServantServer (Proxy @V0.API) (v0Server (D.diffusion ddl)) $ \burl ->
+                    serverSatisfies (Proxy @V0.API) burl stdArgs predicates
+            it "V1 API follows best practices & is RESTful abiding" $ do
+                ddl <- D.dummyDiffusionLayer
+                withServantServer (Proxy @V1.API) (v1Server (D.diffusion ddl)) $ \burl ->
+                    serverSatisfies (Proxy @V1.API) burl stdArgs predicates
```
parsonsmatt: 4 space indent to match convention

wallet-new/test/APISpec.hs
```diff
@@ -157,18 +162,39 @@ testV1Context =
...
+
+    describe "Servant Layout" $ do
+        let strip = reverse . dropWhile isSpace . reverse . dropWhile isSpace
+        root <- fmap strip . runIO $ readProcess "git" ["rev-parse", "--show-toplevel"] []
```
parsonsmatt: this is almost certainly *not* the right way to do this; what is?

wallet-new/test/golden/api-layout.txt
```diff
@@ -0,0 +1,48 @@
...
+         │  ┆
+         │  └─•
+         ┆
+         ├─•
+         ┆
+         ├─•
+         ┆
+         ┆
+         └─ <capture>/
+            └─•
```
parsonsmatt: this looks a lot less useful than i thought it would be, to be honest?

parsonsmatt: I mean, it is still *useful* but the output could render more useful information.

wallet-new/test/APISpec.hs
adinapoli-iohk: @parsonsmatt Have a look at `HasCompileInfo` and more specifically to `Pos.Util.CompileInfo`. It uses `Given` (from the `reflection` package) to carry around the git revision 😉

wallet-new/test/golden/api-layout.txt
adinapoli-iohk: Agreed 😿

wallet-new/test/APISpec.hs
parsonsmatt: I need the repository directory. 

Looks like `git` isn't on CI, so it's failing there :\

adinapoli-iohk: > I need the repository directory.

Ah! Do you think it's worth adding that to `HasCompileInfo` then? It might turn out to be useful not just here, and you would avoid the `runIO` thingy.

> Looks like git isn't on CI

Ops! Maybe is provisioned differently and is not in `$PATH`?

parsonsmatt:
It's not much added complexity, and it will make *very* clear if the route shape changes in a way we don't like. I'm in favor of merging

adinapoli-iohk:
> It's not much added complexity, and it will make very clear if the route shape changes in a way we don't like. I'm in favor of merging

@parsonsmatt Works for me :) I'm not pressing the merge button straight away as I was hoping to get https://github.com/input-output-hk/cardano-sl/pull/2754 merged without dealing with any extra conflict (as that is in the critical path for the code freeze scheduled for this Friday), but I promise I won't forget about your PR!

vincenthz:
In term of golden tests, there's a precedent here:
https://github.com/input-output-hk/cardano-sl/pull/2706

I would recommend inspector for golden tests as it's made for this specific purpose, although it's more for binary than text, but I don't see why it couldn't be improved to handle more textual stuff.

adinapoli-iohk:
Thanks @vincenthz -- TIL about this new shiny `inspector` library 😉 

Do you think it's worthwhile to merge this PR from @parsonsmatt first and switch to your standalone `golden` project or do you think we should seize the chance to do this properly?

parsonsmatt:
I'm a big fan of very small "source-to-sink" distances -- I'd prefer to keep the test with the package that would be responsible for it's failure. I could switch to `inspector` though :)

adinapoli-iohk:
> I'm a big fan of very small "source-to-sink" distances -- I'd prefer to keep the test with the package that would be responsible for it's failure.
I tend to agree with you (I know different folks have different opinions though 😅 ). We can reshuffle the tests later, that's for sure :) 

>  I could switch to inspector though :)

If it doesn't make your life miserable, it would be nice, even just to be consistent to @vincenthz 's work. Sorry for turning this PR into a shed painted with nice fluorescent colours, but I think is one those non-critical PRs where we can afford it 😁

parsonsmatt:
so, unless I am mistaken, `inspector` would actually dramatically alter the output of file, making it much less human-readable, which would defeat the purpose of the test. `inspector` calls `show` on the value, which would escape quotes and insert newline literals.

adinapoli-iohk:
> would actually dramatically alter the output of file, making it much less human-readable, which would defeat the purpose of the test.

Ah, didn't appreciate that whilst skimming the library. I suspect this might make sense for cases when you are writing test vectors of various kinds (sounds the primary use case for the library), but it makes other kind of golden testing a bit more cumbersome.

Would @vincenthz be able to shed some light on the topic? In that case, I might be tempted for now to merge @parsonsmatt 's PR unaltered and reassess at a later stage.

vincenthz:
I'm not sure this the best fit here, as the output of a program is not usually considered a test vector; The printer/parser are configurable so there's might be a way to print this properly in a readable format, but inspector is in early days so there's lots of conceptually missing stuff that could be done in the future like handling multi lines textual output.

there's no problem merging what's already there considering it's been done in parallel and aim for eventual consistency. it would maybe be a waste of time to rework this completely now.

parsonsmatt:
A non-required check is failing due to something unrelated; merging

PR #2751, opened 2018-04-05
[CSL-2433] Wrap actions with MaybeT to enable gaurds to exit gracefully

akegalj:
This PR fixes behaviour of `guard` in integration tests. Previously `guard` was using implementation of `Alternative IO` which would fail : 

```haskell
-- | @since 4.9.0.0
instance Alternative IO where
    empty = failIO "mzero"
    (<|>) = mplusIO
```

and now integration tests are wrapped with `MaybeT` which returns `Nothing` if `guard` fails and thus implements correct behaviour for our use case - which is: "Ignore actions which don't satisfy `guard`". We can simply achieve this with `catMaybes` or similar.

Issue: https://iohk.myjetbrains.com/youtrack/issue/CSL-2433

_Note that PR will be merged to `feature/csl2398-integration-ci` branch and the goal of this PR is to have isolated place to review yt issue and discuss things._

### DIFF
--- a/scripts/test/wallet/integration.sh
+++ b/scripts/test/wallet/integration.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 set -Eeuo pipefail
 
@@ -38,7 +38,7 @@ tmux new-session -s $sessionName -d "WALLET_DEBUG=1 scripts/launch/demo-with-wal
 
 # wait until cluster is fully up and running
 echo "Waiting 140 seconds until local cluster is ready..."
-sleep 140s
+sleep 40s
 
 # import keys
 # TODO: we should import multiple keys for them to be useful as they contain little amount of money. Check does fakeavvm contains more money and if it does use that instead
--- a/scripts/test/wallet/integration.sh
+++ b/scripts/test/wallet/integration.sh
@@ -31,24 +31,20 @@ rm -rf $tmpSecrets
 stack exec -- cardano-keygen --system-start 0 generate-keys-by-spec --genesis-out-dir $tmpSecrets
 
 # run cluster
-# TODO (akegalj): move to integration/Main.hs
-# WALLET_DEBUG flag is needed so that wallet is launched without TLS. We could also use --no-tls, but this involves changes to the underlying script. See https://iohk.myjetbrains.com/youtrack/issue/CT-7#comment=93-17848
 echo "Starting local cardano cluster..."
 tmux new-session -s $sessionName -d "WALLET_DEBUG=1 scripts/launch/demo-with-wallet-api.sh"
 
 # wait until cluster is fully up and running
-echo "Waiting 140 seconds until local cluster is ready..."
+echo "Waiting 40 seconds until local cluster is ready..."
 sleep 40s
 
 # import keys
-# TODO: we should import multiple keys for them to be useful as they contain little amount of money. Check does fakeavvm contains more money and if it does use that instead
 echo "Importing poor HD key/wallet..."
  curl -X POST http://localhost:8090/api/wallets/keys -H 'cache-control: no-cache' -H 'content-type: application/json' -d "\"$tmpSecrets/generated-keys/poor/key0.sk\""
 
 FAILED=0
 
 # run integration tests
-# TODO (akegalj): add tls files to default integration options
 echo "Launching cardano integration tests..."
 stack exec -- cardano-integration-test --help
 stack exec -- cardano-integration-test || {
@@ -56,10 +52,8 @@ stack exec -- cardano-integration-test || {
     cleanState
     FAILED=1
 }
-# --tls-pub-cert scripts/tls-files/server.crt --tls-priv-key scripts/tls-files/server.key
 
 # kill cluster
-# TODO (akegalj): move to integration/Main.hs
 if [[ $FAILED == 0 ]]; then
     echo "Shutting down cardano cluster..."
     cleanState
--- a/scripts/test/wallet/integration.sh
+++ b/scripts/test/wallet/integration.sh
@@ -45,13 +45,22 @@ sleep 40s
 echo "Importing poor HD key/wallet..."
  curl -X POST http://localhost:8090/api/wallets/keys -H 'cache-control: no-cache' -H 'content-type: application/json' -d "\"$tmpSecrets/generated-keys/poor/key0.sk\""
 
+FAILED=0
+
 # run integration tests
 # TODO (akegalj): add tls files to default integration options
 echo "Launching cardano integration tests..."
 stack exec -- cardano-integration-test --help
-stack exec -- cardano-integration-test # --tls-pub-cert scripts/tls-files/server.crt --tls-priv-key scripts/tls-files/server.key
+stack exec -- cardano-integration-test || {
+    echo "Shutting down cardano cluster... it did a fail"
+    cleanState
+    FAILED=1
+}
+# --tls-pub-cert scripts/tls-files/server.crt --tls-priv-key scripts/tls-files/server.key
 
 # kill cluster
 # TODO (akegalj): move to integration/Main.hs
-echo "Shutting down cardano cluster..."
-cleanState
+if [[ $FAILED == 0 ]]; then
+    echo "Shutting down cardano cluster..."
+    cleanState
+fi
--- a/wallet-new/integration/Error.hs
+++ b/wallet-new/integration/Error.hs
@@ -22,20 +22,20 @@ data WalletTestError
 
     | WalletBalanceNotZero Wallet
     | WalletPassMissing Wallet
-    | LocalWalletDiffers Wallet
-    | LocalWalletsDiffers [Wallet]
+    | LocalWalletDiffers Wallet Wallet
+    | LocalWalletsDiffers [Wallet] [Wallet]
 
     | AccountBalanceNotZero Account
-    | LocalAccountDiffers Account
-    | LocalAccountsDiffers [Account]
+    | LocalAccountDiffers Account Account
+    | LocalAccountsDiffers [Account] [Account]
 
     | AddressBalanceNotZero WalletAddress
     | LocalAddressesDiffer WalletAddress [WalletAddress]
-    | LocalAddressDiffer Address
+    | LocalAddressDiffer Address Address
 
     | InvalidTransactionState Transaction
     | InvalidTransactionFee EstimatedFees
-    | LocalTransactionsDiffer [Transaction]
+    | LocalTransactionsDiffer [Transaction] [Transaction]
     | LocalTransactionMissing Transaction [Transaction]
 
     deriving (Show, Eq, Generic)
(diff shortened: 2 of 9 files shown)


ksaric:
Maybe I'm missing something here, but this seems wrong if you plan to do what you commented here - https://iohk.myjetbrains.com/youtrack/issue/CSL-2433

For example, look at the definition of [guard](http://hackage.haskell.org/package/base-4.11.0.0/docs/Control-Monad.html#v:guard).

akegalj:
@ksaric 
It is not obvious to me that I am on a wrong track. Can you elaborate?

> In other words, if the guard condition is not met, the action will retry, and another one will be performed (using action distribution). There are several ways to achieve this.

we can retry actions that failed on guard, when a first pass of actions is over. Or we can just skipp them. Retry seems correct way to achieve this. 

So what I have implemented will return `Maybe action` if `guard` didn't fail or `Nothing` if `guard` failed. In the case `guard` failed, we will run other actions that will be in the queue and retry failed actions later. We can repeat this step until all actions in the actions queue succeed (suecceed in the sense no action will be `guard`ed). 

Let me try to demonstrate the idea :

We are running the action here https://github.com/input-output-hk/cardano-sl/blob/3a2cdef60e87857a76577af4a29042720cd0b8ec/wallet-new/integration/Main.hs#L55 : 
```haskell 
    _ <- runActionCheck
        walletClient
        walletState
        actionDistr
```

Instead of ^ we would have some fold of actions: 
```haskell
runAllActions :: WalletClient -> WalletState -> ActionDistribution -> m WalletState
runAllActions walletClient walletState actionDistr = do
  -- guardedActions are those which failed with `guard`
  (newWalletState, guardedActions) <- foldM (someMoreLogicHere . runActionCheck walletClient) (walletState, mempty) actionDistr
  case nonEmpty guardedActions of
    Just ga -> runAllActions newWalletClient walletState ga
    Nothing -> pure newWalletState
```

Note that ^ is just a pseudo code and custom loop might not be necessary. Also if some action is guarded infinitelly (for example - bug in the code) CI would probably timeout and exit with error. We can probably make it better to build a counter for max number of tries `runAllActions` will retry. When number of tries exceeds we exit with error.

ksaric:
@akegalj Currently, it doesn't contain those things you wrote about above and is not valid. Once you write the code and you think it works I can comment more.

parsonsmatt:
We have a stateful, non-deterministic test suite where each action has some precondition before we can meaningfully do it.

`Alternative` expresses this nicely -- "if the randomly selected action isn't possible, try this other one." `IO`'s `Alternative` instance is ``a <|> b = a `catch` \(_ :: IOException) -> b``, so we can use `IO` directly here. But, we do need to express `<|>`.

If we generate a list of actions to take (eg ~100), we can do `foldr (\action rest -> (action *> rest) <|> rest) (pure ()) :: Alternative m => [m a] -> m a`. This will try `action`, and if it succeeds, do `rest`. 

Since `IOException` is the only one caught by `<|>`, we will be able to still fail tests by throwing other exceptions.

akegalj:
@parsonsmatt 
good point. Then we can get rid of `MaybeT` that I have added.

Note that I have added some variants how we can run tests. I was not sure which could we prefer so I have added few variants. Maybe there will be useful to us - depending how we want to test exactly. They will differ only in some corner cases though.

scripts/test/wallet/integration.sh
```diff
@@ -38,7 +38,7 @@ tmux new-session -s $sessionName -d "WALLET_DEBUG=1 scripts/launch/demo-with-wal
 
 # wait until cluster is fully up and running
 echo "Waiting 140 seconds until local cluster is ready..."
-sleep 140s
+sleep 40s
```
parsonsmatt: I dropped this down because it was taking *a long* time to run tests, and i was getting distracted with twitter and friends between runs! can revert

scripts/test/wallet/integration.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
parsonsmatt: was getting some weird failures so i made it bash

wallet-new/integration/Error.hs
```diff
@@ -21,24 +20,22 @@ import           Cardano.Wallet.Client (ClientError)
...
 
     | AddressBalanceNotZero WalletAddress
     | LocalAddressesDiffer WalletAddress [WalletAddress]
-    | LocalAddressDiffer Address
+    | LocalAddressDiffer Address Address
 
     | InvalidTransactionState Transaction
     | InvalidTransactionFee EstimatedFees
-    | LocalTransactionsDiffer [Transaction]
+    | LocalTransactionsDiffer [Transaction] [Transaction]
```
parsonsmatt: All of the `Differ` exception types now contain the local and remote state.

wallet-new/integration/Functions.hs
```diff
@@ -35,454 +37,446 @@ import qualified Pos.Wallet.Web.ClientTypes.Types as V0
...
-    checkInvariant
-        (transaction `elem` result)
-        (LocalTransactionMissing transaction result)
-
-    -- Modify wallet state accordingly.
-    pure $ ws
-        & actionsNum +~ 1
+    -- increment successful actions
+    log "Success!"
+    actionsNum += 1
```
parsonsmatt: The `actionsNum += 1` modification was common to all actions, so I factored it out.

wallet-new/integration/Functions.hs
```diff
@@ -35,454 +37,446 @@ import qualified Pos.Wallet.Web.ClientTypes.Types as V0
...
-    -> m WalletState
+    -> m ()
 -- Wallets
-runAction wc ws  PostWallet = do
-    newPassword <- liftIO $ generate arbitrary
-    newWall     <- liftIO $ generate $ generateNewWallet newPassword
-    result      <- respToRes $ postWallet wc newWall
+runAction wc action = do
+    log $ "Running: " <> show action
+    case action of
```
parsonsmatt: I switched to an explicit `case` rather than a top level pattern match, because I wanted to log the action that was taken, and also reduce a tiny bit of the boilerplate.

wallet-new/integration/Functions.hs
```diff
@@ -517,8 +522,8 @@ respToRes resp = do
 
 
 -- | Pick a random element using @IO@.
-pickRandomElement :: (MonadIO m, HasCallStack) => [a] -> m a
-pickRandomElement [] = error "called with empty list"
+pickRandomElement :: (MonadIO m, Alternative m) => [a] -> m a
+pickRandomElement [] = empty
```
parsonsmatt: We were getting some `elements: called with empty list` errors. This will turn those showstoppers into test skips.

wallet-new/integration/Main.hs
```diff
@@ -55,18 +51,23 @@ main = do
...
-        getWalletProb  <- createProbability 50
-
-        pure $ (PostWallet, postWalletProb) :|
-             [ (GetWallet, getWalletProb)
-             ]
+        (PostWallet, Weight 1) :| fmap (\x -> (x, Weight 1)) [minBound .. maxBound]
+
+initialWalletState :: WalletClient IO -> IO  WalletState
+initialWalletState wc = do
+    _wallets <- either throwM (pure . wrData) =<< getWallets wc
```
parsonsmatt: I was getting test failures where we'd do `GetWallets`, and there would *gasp* be some wallets already! so I initialize the state of the wallet with whatever the API returns. I'll likely need to get more of the stuff from the wallet initial state as well.

wallet-new/integration/Types.hs
```diff
@@ -39,19 +34,14 @@ type WalletTestMode m =
...
 -- | The probability type that captures the chance of
--- our random action selection choosing a specific @Action@.
--- The number is restricted to 1 - 100 - see @createProbability@.
-data Probability = Probability { getProbability :: Int }
+-- our random action selection choosing a specific 'Action'.
+--
+-- An 'Action' is paired with a 'Weight'. All of the 'Weight's are summed
+-- up, and the eventual probability is the weight divided by the total
+-- weight.
+newtype Weight = Weight { getWeight :: Int }
```
parsonsmatt: I switched from a `0-100` probability to a weight model. This drops all of the partiality involved and makes it simpler to work with.

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -53,7 +53,7 @@ data WalletError =
     | SomeOtherError { weFoo :: !Text, weBar :: !Int }
     | MigrationFailed { weDescription :: !Text }
     | JSONValidationFailed { weValidationError :: !Text }
-    | UnkownError { weMsg :: !Text }
+    | UnknownError { weMsg :: !Text }
```
parsonsmatt: oh god, i typoed this :open_mouth:

wallet-new/integration/Functions.hs
```diff
@@ -35,454 +37,455 @@ import qualified Pos.Wallet.Web.ClientTypes.Types as V0
...
 
-    -- If we don't have any http client errors, the delete was a success.
-    _       <-  either throwM pure =<< deleteWallet wc (walId wallet)
+        DeleteWallet -> do
+            localWallets <- use wallets
 
-    -- Just in case, let's check if it's still there.
-    result  <-  respToRes $ getWallets wc
+            -- The precondition is that we need to have wallets.
+            guard (not (null localWallets))
```
akegalj: we don't even need to have this guard now, because `pickRandomElement` bellow in its current form will be the guard.

wallet-new/integration/Functions.hs
```diff
@@ -35,454 +37,455 @@ import qualified Pos.Wallet.Web.ClientTypes.Types as V0
...
-    checkInvariant
-        (walBalance result == minBound)
-        (LocalWalletDiffers result)
+            let localWallets = ws ^. wallets
 
-    -- No modification required.
-    pure $ ws
-        & actionsNum +~ 1
+            -- The precondition is that we need to have wallets.
+            guard (not (null localWallets))
```
akegalj: I believe `pickRandomElement` will be the guard as well - so maybe this is not needed

wallet-new/integration/Functions.hs
```diff
@@ -35,454 +37,455 @@ import qualified Pos.Wallet.Web.ClientTypes.Types as V0
...
+        UpdateWalletPass -> do
 
-    let walletId = walId wallet
+            localWallets <- use wallets
 
-    -- Get the old wallet password.
-    let oldWalletPass :: Maybe SpendingPassword
-        oldWalletPass = ws ^. walletsPass . at wallet
+            -- The precondition is that we need to have wallets.
+            guard (not (null localWallets))
```
akegalj: probably not needed - `pickRandomElement` bellow will be the guard. 

It is used in many places so won't mention it again

akegalj:
@parsonsmatt 
> The script isn't cleanly restarting the server, so the tests aren't exactly idempotent.

I believe this is due too us runing deterministic tests before (and creating a wallet there). `NewWallet` should be fixed with https://iohk.myjetbrains.com/youtrack/issue/CSL-2432

wallet-new/integration/Main.hs
akegalj: Hm that's strange

wallet-new/integration/Functions.hs
akegalj: with this we can remove all explicit `guard`s

scripts/test/wallet/integration.sh
```diff
@@ -31,27 +31,30 @@ rm -rf $tmpSecrets
...
-# TODO (akegalj): move to integration/Main.hs
-# WALLET_DEBUG flag is needed so that wallet is launched without TLS. We could also use --no-tls, but this involves changes to the underlying script. See https://iohk.myjetbrains.com/youtrack/issue/CT-7#comment=93-17848
 echo "Starting local cardano cluster..."
 tmux new-session -s $sessionName -d "WALLET_DEBUG=1 scripts/launch/demo-with-wallet-api.sh"
 
 # wait until cluster is fully up and running
-echo "Waiting 140 seconds until local cluster is ready..."
-sleep 140s
+echo "Waiting 40 seconds until local cluster is ready..."
+sleep 40s
```
akegalj: 40s doesn't work on my machine (80s does). It will be run by our CI which would probably handle 40 good enough

akegalj:
Merging to the base branch. Lets continue the work there

PR #2753, opened 2018-04-05
[CSL-2429] Add proper acid-state migrations

adinapoli-iohk:
This commit ensures that a version of `wallet-db` produced with the
current version of the wallet can be successfully migrated to this
revision, which incorporates the 'SyncStatistics' as well as the
'WalletSyncState'.

It also extends the 'MigrationSpecs' to make sure we can migrate
backward as well.

I have tested this by restoring a wallet using the version in `develop` and assessing that the node wasn't able to start with the unmigrated code:

```
☁  cardano-sl [feature/CBR-90] stack exec cardano-node -- --topology=docs/network/example-topologies/mainnet-staging.yaml --configuration-key mainnet_dryrun_full --db-path ../staging-db --wallet-db-path ../wallet-db-staging-2018-04-04
Wallet is starting...
[*production*:INFO:ThreadId 4] [2018-04-05 09:51:23.07 UTC] [Attention] Software is built with the wallet backend
[*production*:INFO:ThreadId 4] [2018-04-05 09:51:23.07 UTC] using configurations: ConfigurationOptions {cfoFilePath = "lib/configuration.yaml", cfoKey = "mainnet_dryrun_full", cfoSystemStart = Nothing, cfoSeed = Nothing}
[*production*:INFO:ThreadId 4] [2018-04-05 09:51:24.67 UTC] Asserts are ON
[*production*:INFO:ThreadId 4] [2018-04-05 09:51:24.67 UTC] System start time is 1506450213000000
[*production*:INFO:ThreadId 4] [2018-04-05 09:51:24.67 UTC] Current time is 1522921884674043
[*production*:INFO:ThreadId 4] [2018-04-05 09:51:24.67 UTC] Using configs and genesis:
ConfigurationOptions {cfoFilePath = "lib/configuration.yaml", cfoKey = "mainnet_dryrun_full", cfoSystemStart = Nothing, cfoSeed = Nothing}
[*production*:INFO:ThreadId 4] [2018-04-05 09:51:24.67 UTC] Wallet is enabled!
[node:ERROR:ThreadId 4] [2018-04-05 09:51:24.77 UTC] logException: Could not parse saved checkpoint due to the following error: Failed reading: safecopy: Pos.Core.Common.Types.ChainDifficulty: Cannot find getter associated with this version number: Version {unVersion = 16777216}
From:   Pos.Wallet.Web.State.Storage.WalletStorage:
        Pos.Wallet.Web.State.Storage.WalletInfo:
[...]

PR #2754, opened 2018-04-05
[CBR-90] Asynchronous restoration of a wallet

adinapoli-iohk:
## Summary

This PR allows a `Wallet` to be restored (from seed or from backup) asynchronously, meaning the operation yields immediately an (almost) fully operative wallet where the balance is computed synchronously via the UTXO (fast) and the history asynchronously by a linear scan of the blockchain (slow).

## Linked user story

https://iohk.myjetbrains.com/youtrack/issue/CBR-90

## Description

This PR is the culmination of a couple of months of (very intermittent) work, and thus is quite on the beefy side. However, there are a bunch of ancillary PRs where I have delivered the relevant CSL tickets which can be reviewed separately, if that helps:

- https://github.com/input-output-hk/cardano-sl/pull/2428
- https://github.com/input-output-hk/cardano-sl/pull/2459
- https://github.com/input-output-hk/cardano-sl/pull/2567
- https://github.com/input-output-hk/cardano-sl/pull/2704
- https://github.com/input-output-hk/cardano-sl/pull/2732
- https://github.com/input-output-hk/cardano-sl/pull/2753

I have reshuffled things around a bit, but here is the gist:

- There is now a `TBQueue` used to serialise the `SyncRequest` we get from outside, either as part of an API call (cfr. `restoreWalletFromSeed`) or directly at startup, when we need to resume the restoration/syncing process (cfr. `syncWalletWithBlockchain`).

- The `EncryptedSecretKey`-based API has been dropped in favour of simply the `WalletDecrDecrentials`.

- The code inside `Pos.Wallet.Web.Tracking.Sync` has been refactored in a number of ways. In particular, we now have an incremental, fold-style approach to sync the wallet as opposed as one gulp as before. This allows us to not just resume the syncing process, but also to calculate statistics for the frontend;

- We now actively distinguish between a "Restore" and a "Sync", mainly because during a Restore we compute the balance immediately (from the UTXO) and we do not want to further influence the wallet's balance for transactions we "see" _before_ we pass the blockchain depth at which we restored this wallet;

- A wallet is now immediately put into a `ready` state, so that can be made available to the frontend immediately;

- DB Migrations have been written & tested.

## Frontend test steps

The best course of action is to attach the `cardano-node` to the staging cluster, like this:

```
stack build cardano-sl-wallet-new
stack exec cardano-node -- --topology=docs/network/example-topologies/mainnet-staging.yaml --configuration-key mainnet_dryrun_full --db-path ../staging-db
```

And finally play with the wallet. It's recommended to import a staging DB so that the restoration is realistic. **Note that transactions cannot be sent during a restoration (due to known limitations we discussed in other forums) so doing so will incur in unexpected behaviour.**

## QA Steps and testing tickets covered

- https://iohk.myjetbrains.com/youtrack/issue/CT-149
- https://iohk.myjetbrains.com/youtrack/issue/CT-150

## Final notes

**We probably want this PR to eventually be squash-merged, for easy revert and for a cleaner history.**


util/Pos/Util/Util.hs
```diff
@@ -426,23 +427,28 @@ sleep n = liftIO (threadDelay (truncate (n * 10^(6::Int))))
...
+tMeasureLog label = fmap fst . tMeasure logDebug label
 
 -- | 'tMeasure' with 'putText'. For places you don't have
 -- 'WithLogger' constraint.
 tMeasureIO :: (MonadIO m) => Text -> m a -> m a
-tMeasureIO = tMeasure putText
+tMeasureIO label = fmap fst . tMeasure putText label
+
+timed :: (MonadIO m, WithLogger m) => Text -> m a -> m (a, Microsecond)
+timed = tMeasure logDebug
```
adinapoli-iohk: I have generalised `tMeasure` to be able to return the time it took to execute the passed action.

wallet-new/bench/Bench/Cardano/Wallet/Random.hs
```diff
@@ -44,9 +44,7 @@ pickRandomValueBetween (minValue, maxValue) = liftIO $ randomRIO (minValue, maxV
 -- both are positive and @to@ is greater than @from@.
 waitRandom :: (Double, Double) -> IO ()
 waitRandom (from, to) =
-    if thereIsNoDelay
-        then return ()
-        else randomRIO (fromInMicrosec, toInMicrosec) >>= threadDelay
+    unless thereIsNoDelay $ randomRIO (fromInMicrosec, toInMicrosec) >>= threadDelay
```
adinapoli-iohk: This was actually requested by `hlint`, so I obeyed, go figure.

wallet-new/server/Cardano/Wallet/API/V1/Swagger/Example.hs
```diff
@@ -48,7 +48,7 @@ instance Example WalletId
 instance Example BackupPhrase
 instance Example (V1 BackupPhrase)
 instance Example AssuranceLevel
-instance Example SyncProgress
+instance Example SyncPercentage
```
adinapoli-iohk: I have renamed the `SyncProgress` to just `SyncPercentage`, as the former now is used to capture a much richer set of information about the syncing process.

wallet-new/server/Cardano/Wallet/Server/Plugins.hs
```diff
@@ -89,17 +93,18 @@ legacyWalletBackend :: (HasConfigurations, HasCompileInfo)
...
-      logInfo $ sformat ("Transaction submission disabled: "%build)
-        walletTxCreationDisabled
-
-      walletServeImpl
-        (getApplication diffusion)
-        walletAddress
-        -- Disable TLS if in debug mode.
-        (if isDebugMode walletRunMode then Nothing else walletTLSParams)
-        (Just $ setOnExceptionResponse exceptionHandler defaultSettings)
+      modifyLoggerName (const "legacyServantBackend") $ do
```
adinapoli-iohk: Here I have simply added this custom logger name. You will notice (here and in the rest of the `modifyLoggerName` calls) that I am using `const` and not `mappend`. This is intentional as I have discovered that `log-warper` couldn't be configured (at least, I  failed in doing so) to accept a composite logger name (like `foo.bar.baz`) in the `yaml` config file to set the verbosity for a [specific logger](https://github.com/serokell/log-warper/blob/master/examples/how-to-log-config.yaml). This helped my testing quite a lot, but I'm happy to revert.

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
```diff
@@ -60,16 +62,40 @@ instance (Migrate from to, Typeable from, Typeable to) => Migrate [from] (NonEmp
...
+
+instance Migrate (OldStorage.SyncStatistics, Maybe Core.ChainDifficulty) V1.SyncProgress where
+    eitherMigrate (OldStorage.SyncStatistics{..}, currentBlockchainDepth) =
+        let unknownCompletionTime = Core.Timestamp $ fromMicroseconds (fromIntegral (maxBound :: Int))
+            percentage = case currentBlockchainDepth of
+                Nothing -> 0.0
+                Just nd | wspCurrentBlockchainDepth >= nd -> 100
+                Just nd -> (fromIntegral wspCurrentBlockchainDepth / max 1.0 (fromIntegral nd)) * 100.0
+            toMs (Core.Timestamp microsecs) =
+              V1.mkEstimatedCompletionTime (round @Double $ (realToFrac (toMicroseconds microsecs) / 1000.0))
```
adinapoli-iohk: The acute reader will notice that `mkEstimatedCompletionTime` has `Word` precision whereas we are downcasting from an `Integer`, and thus we could surely overflow. In practice, if such a thing would happen it would mean our `EstimatedCompletionTime` is obscenely high (we are talking about roughly `584942417` years if I did the math correctly 😅 ).

wallet/src/Pos/Wallet/Web/Methods/Backup.hs
```diff
@@ -5,84 +5,29 @@
...
-                                       getWalletSnapshot)
-import           Pos.Wallet.Web.Tracking (syncWalletOnImport)
-import           Pos.Wallet.Web.Util (getWalletAccountIds)
+import           Pos.Wallet.Web.Methods.Restore (restoreWalletFromBackup)
+import           Pos.Wallet.Web.State.State (askWalletSnapshot)
+import           Pos.Wallet.Web.Tracking.Types (SyncQueue)
 import           Servant.API.ContentTypes (NoContent (..))
+import           UnliftIO (MonadUnliftIO)
 
-
```
adinapoli-iohk: This code has been moved from `Backup` to `Restore` -- the intention was to have a more cohesive module with "all-the-thing-restoration".

wallet/src/Pos/Wallet/Web/State/State.hs
```diff
@@ -482,7 +506,7 @@ flushWalletStorage :: (MonadIO m, HasConfiguration)
                    -> m ()
 flushWalletStorage = updateDisk A.FlushWalletStorage
 
-applyModifierToWallet
+applyModifierToWallet2
```
adinapoli-iohk: Very interested to hear about this bit. Essentially, I was concerned that `applyModifierToWallet` could still have been in the transaction log, so I have added another acidic transaction (see below) and stopped exporting the original `applyModifierToWallet`.

wallet/src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -875,4 +974,5 @@ instance Migrate WalletStorage where
         , _wsChangeAddresses = mapAddrKeys _v2_wsChangeAddresses
         }
       where
-        mapAddrKeys = HM.fromList . fmap (first unsafeCIdToAddress) . HM.toList
+        mapAddrKeys  = HM.fromList . fmap (first unsafeCIdToAddress) . HM.toList
+        migrateMapElements = HM.fromList . fmap (second migrate) . HM.toList
```
nc6: You haven't introduced a new version here, instead tacking on an additional migration step between two existing versions. The problem occurs if anywhere has already migrated to the version using `Address`, because then they will not be able to migrate further and will have problems. Maybe the chance of that is sufficiently low (it's only been in develop for a couple of days, albeit in master for longer), but still, it would be better to have separate migrations here.

wallet/src/Pos/Wallet/Web/Tracking/BListener.hs
```diff
@@ -86,6 +94,7 @@ onApplyBlocksWebWallet blunds = setLogger . reportTimeouts "apply" $ do
     -- something a bit more reasonable.
     pure mempty
   where
+    -- TODO(adn): Revisit this bit.
```
nc6: Maybe elaborate on this if it's going to remain past your private branch. What needs revisiting?

wallet/src/Pos/Wallet/Web/Tracking/BListener.hs
```diff
@@ -134,6 +147,7 @@ onRollbackBlocksWebWallet blunds = setLogger . reportTimeouts "rollback" $ do
...
```
nc6: As above

wallet/src/Pos/Wallet/Web/Tracking/Restore.hs
```diff
@@ -0,0 +1,103 @@
...
+                        , CanLog m
+                        , MonadDBRead m
+                        , MonadUnliftIO m
+                        ) => WalletDB -> WalletDecrCredentials -> m ()
+restoreWalletBalance db credentials = do
+    utxo <- filterUtxo walletUtxo
+    -- FIXME(adn): Quite inefficient for now.
+    -- NOTE: Due to the fact that in the current implementation of the wallet the balance
+    -- (as returned in a 'CWallet') is computed by looking at the coins on each account of
+    -- the wallet, we also need to add these addresses to the wallet for the balance
```
nc6: This is a bit confusing. Which addresses? Those corresponding to the accounts in the wallet?

wallet/src/Pos/Wallet/Web/Tracking/Restore.hs
```diff
@@ -0,0 +1,103 @@
...
+    -- computation to consider them.
+    forM_ (M.elems utxo) $ \txOutAux -> do
+        whenJust (decryptAddress credentials . getAddress $ txOutAux) (WS.addWAddress db)
```
nc6: Is this not a total duplication of what we do in `filterUtxo walletUtxo` on line 76? We've already filtered to those outputs where we can decrypt the address.

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -183,147 +274,202 @@ syncWalletsWithGState encSKs = forM_ encSKs $ \encSK -> handleAny (onErr encSK)
...
+    -- Compute the next 'CAccModifier' and tentatively assess the throughput.
+    ((mapModifier, newSyncTip), timeTook) <-
+        timed "syncWalletWithBlockchainUnsafe.computeAccModifier" $
+            computeAccModifier credentials getBlockHeaderTimestamp walletTip usedAddresses mempty 0
+
+    let syncTime = BoundedSyncTime 10000 timeTook
+    WS.updateSyncStatistics db walletId (SyncStatistics (calculateThroughput syncTime) (depthOf newSyncTip))
+
+    -- Apply the 'CAccModifier' to the wallet state.
+    applyModifierToWallet db (srOperation syncRequest) walletId newSyncTip mapModifier
```
nc6: Did you not rename this to `applyModifierToWallet2`? I am a little confused!

nc6: Oh, I see, there is another one _in this module_. Ignore that!

wallet/src/Pos/Wallet/Web/Tracking/Sync.hs
```diff
@@ -406,23 +552,69 @@ trackingRollbackTxs (eskToWalletDecrCredentials -> wdc) dbUsed getDiff getTs txs
...
+            RestoreWallet rbd -> RestoringFrom rbd newTip
+    logDebug $ sformat ("applyModifierToWallet: new SyncState = " % shown) trackingOperation
 
     let cMetas = mapMaybe (\THEntry {..} -> (\mts -> (encodeCType _thTxId
                                                      , CTxMeta . timestampToPosix $ mts)
                                             ) <$> _thTimestamp)
                $ DL.toList camAddedHistory
 
-    WS.applyModifierToWallet
+    WS.applyModifierToWallet2
```
nc6: As observed above, should be no need to rename this function.

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
```diff
@@ -60,16 +62,40 @@ instance (Migrate from to, Typeable from, Typeable to) => Migrate [from] (NonEmp
         ]
     eitherMigrate (x:xs) = (:|) <$> eitherMigrate x <*> mapM eitherMigrate xs
 
-instance Migrate (V0.CWallet, V0.WalletInfo) V1.Wallet where
-    eitherMigrate (V0.CWallet{..}, V0.WalletInfo{..}) =
+instance Migrate (V0.CWallet, OldStorage.WalletInfo, Maybe Core.ChainDifficulty) V1.Wallet where
```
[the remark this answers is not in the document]
adinapoli-iohk: I have a confession to make: that `Migrate` typeclass is _not_ the one from `safecopy`, but so it happens that is a one I've created to migrate from V0 & V1 types, but purely at the boundary of the application, i.e. in the Servant handlers.

In other terms, we never store anything V1-related at this stage, and migrating in this context simply means "adapting the types for the V1 endpoints". I hope it's clearer, maybe I should simply rename that typeclass 🤔

wallet/src/Pos/Wallet/Web/Tracking/Restore.hs
adinapoli-iohk: At this stage, it's unfortunately needed duplication. The reason is the following: `addWAddress` has type `addWAddress :: WAddressMeta -> Update ()`, but alas `WAddressMeta` can be obtained only by decrypting the address within a `txOutAux`, as `decryptAddress` has type
`decryptAddress :: WalletDecrCredentials -> Address -> Maybe WAddressMeta`.

In reality the morally correct way to implement that would be to have something like a `filterMap` variant of my `filterUtxo` function, which could filter and apply a transformation on the matched predicate. I think I even attempted something similar but it wasn't easy to do in the current implementation of the DB.UTXO module. Will give another spin, and if I fail, I suspect we might need to stick with the status quo 😉

wallet/src/Pos/Wallet/Web/State/Migrations.hs
```diff
@@ -0,0 +1,16 @@
+-- | This module contains types used for acid-state migrations
+-- that would conflict if declared in the same module.
```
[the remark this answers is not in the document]
adinapoli-iohk: Yes, and I confess I didn't do that because I feared that this would have confused `safecopy` (this stem mostly from my ignorance when it comes to this library). I feared that changing a data constructors would have confused the library as that meant messing up with its internal representation. I'm ofc very happy to rename this `WalletTip_V0` I don't see any harm in doing that (on top of my head!)

wallet/src/Pos/Wallet/Web/State/State.hs
```diff
@@ -179,8 +184,11 @@ getWalletMetaIncludeUnready ws includeReady wid =
 getWalletPassLU :: WalletSnapshot -> CId Wal -> Maybe PassPhraseLU
 getWalletPassLU ws wid = queryValue ws (S.getWalletPassLU wid)
 
-getWalletSyncTip :: WalletSnapshot -> CId Wal -> Maybe WalletTip
-getWalletSyncTip ws wid = queryValue ws (S.getWalletSyncTip wid)
+-- TODO(adn):  Is it necessary to preserve 'getWalletSyncTip' in case it's
+-- present in some acid-state transaction log?
```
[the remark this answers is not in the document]
adinapoli-iohk: Thanks Nic, TIL 😉

wallet-new/src/Cardano/Wallet/API/V1/Migration/Types.hs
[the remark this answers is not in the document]
nc6: Ah! How wonderfully confusing :-) Yes, if this is not the safecopy class, then this code seems much more reasonable. Maybe a qualified import would clear up the situation of thinking this is from `safecopy`?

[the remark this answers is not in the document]
adinapoli-iohk: let me take care of that :)

PR #2758, opened 2018-04-06
[CSL-2423] Implement WalletActiveLayer for wallets and accounts.

ksaric:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2423

### DIFF
--- a/wallet-new/src/Cardano/Wallet/WalletLayer/Error.hs
+++ b/wallet-new/src/Cardano/Wallet/WalletLayer/Error.hs
@@ -13,17 +13,19 @@ import           Universum
 import qualified Data.Text.Buildable
 import           Formatting (bprint, stext, (%))
 
-import           Cardano.Wallet.API.V1.Types (WalletId)
+import           Cardano.Wallet.API.V1.Types (WalletId, AccountIndex)
 
 
 data WalletLayerError
-    = MissingWallet WalletId
-    | MissingAccount
+    = WalletNotFound WalletId
+    | AccountNotFound WalletId AccountIndex
+    | AddressNotFound WalletId AccountIndex
     deriving (Show, Eq, Generic)
 
 instance Exception WalletLayerError
 
 instance Buildable WalletLayerError where
-    build (MissingWallet wId)   = bprint ("Missing wallet with wallet id ("%stext%").") (show wId)
-    build (MissingAccount)      = bprint ("Missing account.")
+    build (WalletNotFound  wId      ) = bprint ("Wallet not found. Wallet id ("%stext%").") (show wId)
+    build (AccountNotFound wId accIx) = bprint ("Account not found. Wallet id ("%stext%"), accound index ("%stext%").") (show wId) (show accIx)
+    build (AddressNotFound wId accIx) = bprint ("Address not found. Wallet id ("%stext%"), accound index ("%stext%").") (show wId) (show accIx)
 
--- a/wallet-new/src/Cardano/Wallet/WalletLayer/Error.hs
+++ b/wallet-new/src/Cardano/Wallet/WalletLayer/Error.hs
@@ -1,0 +1,29 @@
+{-# LANGUAGE DeriveGeneric #-}
+
+-- | Types describing runtime errors related to
+-- wallet layers. It should be a common interface for
+-- all the errors popping up from the @WalletLayer@.
+
+module Cardano.Wallet.WalletLayer.Error
+    ( WalletLayerError (..)
+    ) where
+
+import           Universum
+
+import qualified Data.Text.Buildable
+import           Formatting (bprint, stext, (%))
+
+import           Cardano.Wallet.API.V1.Types (WalletId)
+
+
+data WalletLayerError
+    = MissingWallet WalletId
+    | MissingAccount
+    deriving (Show, Eq, Generic)
+
+instance Exception WalletLayerError
+
+instance Buildable WalletLayerError where
+    build (MissingWallet wId)   = bprint ("Missing wallet with wallet id ("%stext%").") (show wId)
+    build (MissingAccount)      = bprint ("Missing account.")
+
--- a/wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
+++ b/wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
@@ -1,5 +1,5 @@
-{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ConstraintKinds     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module Cardano.Wallet.WalletLayer.Legacy
     ( MonadLegacyWallet
@@ -18,17 +18,20 @@ import           Cardano.Wallet.Kernel.Diffusion (WalletDiffusion (..))
 
 import           Cardano.Wallet.API.V1.Migration (migrate)
 import           Cardano.Wallet.API.V1.Migration.Types ()
-import           Cardano.Wallet.API.V1.Types (Account, AccountIndex, Address, NewWallet (..),
-                                              V1 (..), Wallet, WalletId, WalletOperation (..),
-                                              WalletUpdate)
+import           Cardano.Wallet.API.V1.Types (Account, AccountIndex, AccountUpdate, Address,
+                                              NewAccount (..), NewWallet (..), V1 (..), Wallet,
+                                              WalletId, WalletOperation (..), WalletUpdate)
 
 import           Pos.Client.KeyStorage (MonadKeys)
 import           Pos.Crypto (PassPhrase)
+
+import           Pos.Wallet.Web.Account (GenSeed (..))
 import           Pos.Wallet.Web.ClientTypes.Types (CWallet (..), CWalletInit (..), CWalletMeta (..))
-import           Pos.Wallet.Web.Methods.Logic (MonadWalletLogicRead, deleteWallet, getAccount,
-                                               getAccounts, getWallet)
+import           Pos.Wallet.Web.Methods.Logic (MonadWalletLogicRead)
+import qualified Pos.Wallet.Web.Methods.Logic as V0
 import           Pos.Wallet.Web.Methods.Restore (newWallet, restoreWallet)
-import           Pos.Wallet.Web.State.State (WalletDbReader, askWalletSnapshot, askWalletDB, getWalletAddresses, setWalletMeta)
+import           Pos.Wallet.Web.State.State (WalletDbReader, askWalletDB, askWalletSnapshot,
+                                             getWalletAddresses, setWalletMeta)
 import           Pos.Wallet.Web.State.Storage (getWalletInfo)
 
 
@@ -59,8 +62,11 @@ bracketPassiveWallet =
         , _pwlUpdateWallet  = pwlUpdateWallet
         , _pwlDeleteWallet  = pwlDeleteWallet
 
+        , _pwlCreateAccount = pwlCreateAccount
         , _pwlGetAccounts   = pwlGetAccounts
         , _pwlGetAccount    = pwlGetAccount
+        , _pwlUpdateAccount = pwlUpdateAccount
+        , _pwlDeleteAccount = pwlDeleteAccount
 
         , _pwlGetAddresses  = pwlGetAddresses
         }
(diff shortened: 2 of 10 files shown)


wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -1,43 +1,233 @@
...
+    newWalletHandler CreateWallet  = newWallet
+    newWalletHandler RestoreWallet = restoreWallet
+
+
+pwlGetWalletIds
+    :: forall ctx m. (MonadLegacyWallet ctx m)
+    => m [WalletId]
+pwlGetWalletIds = do
+    ws          <- askWalletSnapshot
+    migrate $ getWalletAddresses ws
```
KtorZ: Perhaps it's me but the naming really confused me here. I know it is from the legacy wallet but this apparently returns wallet ids. At least the new passive layer fixes the naming :)

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -1,43 +1,233 @@
...
+-- the DB.
+pwlDeleteWallet
+    :: forall ctx m. (MonadLegacyWallet ctx m)
+    => WalletId
+    -> m Bool
+pwlDeleteWallet wId = do
+    cWId        <- migrate wId
+    -- TODO(ks): It would be better to catch specific @Exception@.
+    -- Maybe @try@?
+    catchAll (const True <$> V0.deleteWallet cWId) (const . pure $ False)
```
KtorZ: Does the DB throw upon deleting an non-existing wallet? I don't see why we would swallow db exceptions here.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -1,43 +1,233 @@
...
+        walletInfo  <- getWalletInfo cWId ws
+        migrate (wallet, walletInfo)
+
+pwlUpdateWallet
+    :: forall ctx m. (MonadLegacyWallet ctx m)
+    => WalletId
+    -> WalletUpdate
+    -> m Wallet
+pwlUpdateWallet wId wUpdate = do
+    walletDB    <- askWalletDB
```
KtorZ: Is there any document where we elaborate a bit more on the usage of the db vs a snapshot of it? I'd like to understand the rational behind this :thinking:

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -1,43 +1,233 @@
...
+    accInit     <- migrate (wId, newAcc)
+    cAccount    <- V0.newAccount RandomSeed spendingPassword accInit
+
+    migrate cAccount
+
+pwlGetAccounts
+    :: forall ctx m. (MonadLegacyWallet ctx m)
+    => WalletId
+    -> m [Account]
+pwlGetAccounts wId = do
```
KtorZ: Any reason why we choose to return only wallet ids (`getWalletIds`) but return complete accounts here?

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -1,43 +1,233 @@
...
+
+pwlGetAccount
+    :: forall ctx m. (MonadLegacyWallet ctx m)
+    => WalletId
+    -> AccountIndex
+    -> m (Maybe Account)
+pwlGetAccount wId aId = do
+    accId       <- migrate (wId, aId)
+    account     <- V0.getAccount accId
+    Just <$> migrate account
```
KtorZ: `V0.getAccount` throws whenever the account is not found. Yet, the function signature suggests that we'd return `Nothing` in such case, am I right? 
Perhaps we can recover from the exception thrown by V0 methods `getAccountAddrsOrThrow` (and friends) and leverage the `Maybe`? What do you think :no_mouth: ?

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -1,43 +1,233 @@
...
+    migrate cAccount
+
+pwlDeleteAccount
+    :: forall ctx m. (MonadLegacyWallet ctx m)
+    => WalletId
+    -> AccountIndex
+    -> m Bool
+pwlDeleteAccount wId accIdx = do
+    accId <- migrate (wId, accIdx)
+    catchAll (const True <$> V0.deleteAccount accId) (const . pure $ False)
```
KtorZ: Same remark as above. It feels weird to swallow exceptions here -- especially when we don't do it for other methods.

wallet-new/src/Cardano/Wallet/WalletLayer/README.md
```diff
@@ -0,0 +1,62 @@
+# Wallet layer
```
KtorZ: :heart: Thanks for this :)

wallet-new/src/Cardano/Wallet/WalletLayer/Error.hs
```diff
@@ -0,0 +1,29 @@
...
+import qualified Data.Text.Buildable
+import           Formatting (bprint, stext, (%))
+
+import           Cardano.Wallet.API.V1.Types (WalletId)
+
+
+data WalletLayerError
+    = MissingWallet WalletId
+    | MissingAccount
+    deriving (Show, Eq, Generic)
```
parsonsmatt: I think I would prefer to see a finer-grained error type. If I do `getWallet :: WalletId -> m (Either WalletLayerError Wallet)`, then I have a ton of cases that won't ever arise.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
parsonsmatt: I agree. It's a little confusing. I recall @adinapoli-iohk saying that we acquire a snapshot of the DB, so that a single request returns a consistent answer. It's sort of like a read-lock on a SQL database table for a transaction.

Of course, we must get a *new* snapshot if we perform any updates, otherwise we run the risk of stale reads!

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
parsonsmatt: Let's definitely *not* swallow exceptions here, without at least reporting on them. 

I have seen the `V0` code throw a `WalletError` type defined in `wallet/src/Pos/Wallet/Web/Errors/Types.hs` and not much else.

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -1,20 +1,53 @@
...
+    , _pwlCreateAccount :: WalletId -> NewAccount -> m Account
+    , _pwlGetAccounts   :: WalletId -> m [Account]
+    , _pwlGetAccount    :: WalletId -> AccountIndex -> m (Maybe Account)
+    , _pwlUpdateAccount :: WalletId -> AccountIndex -> AccountUpdate -> m Account
+    , _pwlDeleteAccount :: WalletId -> AccountIndex -> m Bool
+    -- * addresses
+    , _pwlGetAddresses  :: WalletId -> m [Address]
     }
 
+makeLenses ''PassiveWalletLayer
```
parsonsmatt: This might be a good time to use `makeClassy` instead, which lets us use the `pwlCreateWallet` function on an `ActiveWallet` too. We would only need an `instance HasPassiveWallet (ActiveWalletLayer m)`.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
ksaric: Good catch guys, already fixed in https://github.com/input-output-hk/cardano-sl/pull/2778 (the PR following this one).

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
ksaric: There isn't a document AFAIK, but since we are going to change this anyway, not sure if there is anything better to write then checking out https://github.com/input-output-hk/cardano-sl/blob/develop/wallet/src/Pos/Wallet/Web/State/State.hs
If that is confusing as well, I can write a couple of lines in some documentation, but essentially we use the snapshot for reading and the db for writing.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
adinapoli-iohk: Agree, we should fail appropriately, if not with an exception, with a proper `Either` type to be reflected in the type signature of `pwlDeleteWallet`.

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -1,20 +1,53 @@
...
+    , pwlUpdateWallet
+    , pwlDeleteWallet
+
+    , pwlCreateAccount
+    , pwlGetAccounts
+    , pwlGetAccount
+    , pwlUpdateAccount
+    , pwlDeleteAccount
+
+    , pwlGetAddresses
```
adinapoli-iohk: What I like about this layer is that if you squint hard, you will see how is shaping up to essentially be a CRUD layer, because for each resource we do the usual _Create, Read, Update, Delete_. Would be nice to reshuffle the order of the functions (or amend them) to follow this pattern: for each resource, we expose just these 4 operations. I would be curious to see where we get by chasing this trail.

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -1,20 +1,53 @@
...
 
 
 -- | The passive wallet (data) layer. See @PassiveWallet@.
 data PassiveWalletLayer m = PassiveWalletLayer
-    { pwlGetWalletIds       :: m [WalletId]
+    {
+    -- * wallets
+      _pwlCreateWallet  :: NewWallet -> m Wallet
+    , _pwlGetWalletIds  :: m [WalletId]
+    , _pwlGetWallet     :: WalletId -> m (Maybe Wallet)
```
adinapoli-iohk: As said by multiple folks, it would be better to have `Either LayerError a` rather than an opaque maybe.

wallet-new/src/Cardano/Wallet/WalletLayer/Error.hs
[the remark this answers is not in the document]
ksaric: Fair enough.

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
adinapoli-iohk: @ksaric Our data record should really be independent from any concrete monad instantiation -- how do you know that you will have `MonadThrow` for the _new_ data layer? I might be missing something, but that is not something which would be necessarily true, right? That depends from the particular `bracketXXX` implementation.

The less constraints on our monads, and the more explicit semantic in our pure data types, the better, imho 😉

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -1,20 +1,53 @@
 module Cardano.Wallet.WalletLayer.Types
     ( PassiveWalletLayer (..)
     , ActiveWalletLayer (..)
+    -- * Lenses
+    , pwlCreateWallet
```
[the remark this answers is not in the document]
parsonsmatt: IMO the `createWallet` should be a getting function (eg. `createWallet = view pwlCreateWallet`), as we are probably never going to be modifying them, and are mostly using the lenses as polymorphic getters.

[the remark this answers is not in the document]
adinapoli-iohk: > IMO the createWallet should be a getting function

I love this suggestion, @parsonsmatt :)

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
ksaric: I agree.

> how do you know that you will have MonadThrow for the new data layer?

The same way I know we are going to use `Either` here - `Either` is forcing all the data layers to handle exceptions, and we could say the same for `MonadThrow`, I guess :man_shrugging:

adinapoli-iohk:
Merging this one with the gentlemen agreement that @ksaric will address all the different points in a different PR and will make sure a feature branch will be created for features next time rather than opening PRs directly on `develop`.

PR #2759, opened 2018-04-06
[CSL-2432] Generate arbitrary 12 word backup phrase

akegalj:
This issue is fixing:
 * wrong implementation of `Arbitrary BackupPhrase` which picked arbitrary 12 words from bip39 english mnemonic dictionary
      - this is not sufficient to be bip39 compatible as it doesn't take checksum into account
 * `Arbitrary (V1 BackupPhrase)` is using correct 12 word `BackupPhrase` implemented above instead of correct hardcoded BackupPhrase
 * documentation should contain arbitrary `BackupPhrase` now
 * problems with integration tests which are relying on generating arbitrary `BackupPhrase` for generating new wallets - https://iohk.myjetbrains.com/youtrack/issue/CSL-2398

This issue is not fixing:
 * the distinction between paper vend backup phrase (9 words) and normal backup phrase (12 words)
     - this should be handled with https://iohk.myjetbrains.com/youtrack/issue/CSL-2048
     - we gain only stronger type safety
 * documentation would be more correct with the distinction above as some endpoints are using peper vend backup phrase (and documentation is showing 12 words mnemonic instead of 9 words)
     - note that most of the certificates are already redeemed (http://cardanoexplorer.com/genesis) so with time this is less and less important - but still incorrect (should be handled with https://iohk.myjetbrains.com/youtrack/issue/CSL-2048 )


wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -28,21 +29,15 @@ newtype BackupPhrase = BackupPhrase
     { bpToList :: [Text]
     } deriving (Eq, Generic)
 
+-- NOTE: it's guaranteed not to fail
```
parsonsmatt: Can we write a QuickCheck property that verifies this?

wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -28,21 +29,15 @@ newtype BackupPhrase = BackupPhrase
...
 instance Arbitrary BackupPhrase where
-    arbitrary = BackupPhrase <$> vectorOf 12 (elements englishWords)
+    arbitrary = either error identity <$> arbitraryMnemonic 16
```
parsonsmatt: Instead of failing at runtime, we could instead recur until we get a valid value:

```haskell
arbitrary = do
    em <- arbitraryMnemonic 16
    case em of
          Left _ -> arbitrary
          Right a -> pure a
```

akegalj:
>  I'm curious about how the "guaranteed not to fail" bit works out, and if we can implement a test for it.

only by code review. The bits that generate bip39 are from here https://github.com/input-output-hk/cardano-sl/blob/master/wallet/src/Pos/Util/Mnemonics.hs#L51 and it is a pure function that generates failure only if it doesn't satisfies some of the conditions:

> "entropy must be a multiple of 4 bytes"

which it is as we are using 16 bytes for 12 word mnemonics https://github.com/input-output-hk/cardano-sl/pull/2759/files#diff-f78b72ef586b87d206550724ec229577R39

> "maximum entropy is 64 bytes (512 bits)"

the same answer as the above.

> "non-ASCII characters not supported"

Hm - this is a bit tricky to review as I have to dig into `ByteString` and `toMnemonic` internals. And I have just tried to do so for 15 min - but concluded that it is easier in this case to just make a safe loop as you have suggested :) Thanks for a careful review!

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -237,20 +237,7 @@ instance ByteArray.ByteArrayAccess a => ByteArray.ByteArrayAccess (V1 a) where
 
 -- TODO(adinapoli) Rewrite it properly under CSL-2048.
```
adinapoli-iohk: @akegalj One for you: once we close the relevant ticket, we will need to remove this stale comment 😉

wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -29,20 +30,17 @@ newtype BackupPhrase = BackupPhrase
...
-englishWords = [ "recycle" , "child" , "universe" , "extend" , "edge" , "tourist"
-               , "swamp" , "rare" , "enhance" , "rabbit" , "blast" , "plastic" , "attitude"
-               , "name" , "skull" , "merit" , "night" , "idle" , "bone" , "exact"
-               , "inflict" , "legal" , "predict" , "certain" , "napkin" , "blood"
-               , "color" , "screen" , "birth" , "detect" , "summer" , "palm"
-               , "entry" , "swing" , "fit" , "garden" , "trick" , "timber"
-               , "toss" , "atom" , "kitten" , "flush" , "master" , "transfer"
-               , "success" , "worry" , "rural" , "silver" , "invest" , "mean"
-               ]
+arbitraryMnemonic :: Int -> Gen (Either Text BackupPhrase)
```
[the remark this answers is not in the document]
akegalj: Agreed. 

> Now, I understand that my implementation of octets is naive because I think the size of the checksum is proportional to the number of words chosen

for 9 words mnemonic we would need 12 octets - but yes - it is deterministic.

Lets track this under https://iohk.myjetbrains.com/youtrack/issue/CSL-2441

PR #2760, opened 2018-04-06
[CBR-119] Ugly fix to avoid some OOM problems on Explorer

KtorZ:
While waiting for a better design / design revision, this will most likely fix CPU issues of
iterating over huge list of transactions for some addresses. The pagination of those addresses
happens client-side, in browser and forces the server to fetch and return all transactions
summaries on each page load; leading to really heavy load in some cases.

### DIFF
--- a/explorer/src/Pos/Explorer/Web/Server.hs
+++ b/explorer/src/Pos/Explorer/Web/Server.hs
@@ -340,9 +340,10 @@ getAddressSummary cAddr = do
 
     let nTxs = length txIds
 
-    -- FIXME Waiting for design discussion
+    -- FIXME [CBR-119] Waiting for design discussion
     when (nTxs > 1000) $
-        throwM $ Internal $ "Response too large: " <> show nTxs <> " transactions"
+        throwM $ Internal $ "Response too large: no more than 1000 transactions"
+            <> " can be returned at once. This issue is known and being worked on"
 
     transactions <- forM txIds $ \id -> do
         extra <- getTxExtraOrFail id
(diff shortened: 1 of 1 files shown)


KtorZ:
I agree 100%. Though, as @adinapoli-iohk pointed out, pagination would unfortunately require some redesign and breaking changes in the API. 
Let me look into making the maximum a bit more explicit for clients :+1:

ksaric:
> Isn't it already the case though? It seems that error thrown is caught and explicitly returned as Left in the API. Or am I missing something?

You are correct, but I didn't mean that. I mean a more informative message for the client such as: "Response too large. We currently don't allow users to view more than 1000 transactions. We are currently working on it, it will be fixed.".

> I only hope this won't block some of the Exchanges relying on Explorer.

Exactly. Not just that, but for the plain users who want to see their Exchange information.

KtorZ:
I left a reference to the original ticket next to the note and will update the ticket with details about this PR once this is accepted. 
Also, I've reworded a bit the error message to be more aligned with what @ksaric & @adinapoli-iohk suggested :)

adinapoli-iohk:
@KtorZ This looks good to me now, thanks!

I think we can press the merge button on this one -- despite the limitation in Explorer functionality, I still think that's better than causing the system to go OOM.

We might need to exercise caution if one of the partner exchanges are actively relying on Explorer to answer some of the queries (cc @jmitchell and @rvl ), but I think that's a problem easily mitigated in practice by either asking them to downgrade Explorer (they are all running it locally as far as I know) or by shipping an hotfix. Either way, we should be good.

I'm leaving this one open to marinade for 24 hours and then will merge if no objection will arise.

PR #2769, opened 2018-04-10
wallet-new: Remove in-package modules from test-suite build

erikd:
Having them included in the test-suite build is unnecessary and prevents
HPC coverage analysis from working.


erikd:
This is not correct.

The naming of these modules suggests they are part of the library rather than being purely for testing purposes. These modules should have names beginning with `Test` so it is obviously they are part of the tests and not part of the library.

PR #2770, opened 2018-04-10
stack.yaml: Re-enable building of haddocks

erikd:
These were disabled due to problems with the swager2 version that was
being pulled from github. Since we have now moved to a swagger2 version
that comes from Hackage we can re-enable Haddock generation.

The fact that haddocks were not being built was on the list of issues
mentioned by the external auditors.


erikd:
Swagger2 still breaks haddock builds. Will close this for now.

PR #2771, opened 2018-04-10
[CSL-2442] Prevent the backend from making payments whilst a wallet is restoring

adinapoli-iohk:
## Description

This PR builds on top of the work I did as part of `feature/CBR-90` to disallow making new payments whilst the _source_ wallet is being restored. This extends what we have been doing already in the frontend to the backend, to prevent exchanges from shooting on their feets.

It doesn't assume prior knowledge over the work on CBR-90, thus is suitable for reviewing in a black-box fashion by chaps like @KtorZ and @parsonsmatt . 

You will notice how there are no tests worth their name for the V1 API. Reason is twofold:

- We need proper support in the `deterministicTests`, for which I have logged CBR-184.
- The implementation for the V1 piggybacks on V0 (for which we have tests), therefore it should be enough to assess by code audit that the V1 shim does the right thing, and in terms of testing we are covered in the actual logic by the V0 handler extension (I hope 😂 ).

If you have ideas on how to test this one without the integration tests, I'm all ear.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CSL-2442

### DIFF
--- a/wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
+++ b/wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
@@ -5,26 +5,30 @@ module Test.Pos.Wallet.Web.Methods.PaymentSpec
 import           Nub (ordNub)
 import           Universum
 
+import           Control.Exception.Safe (try)
 import           Data.Default (def)
 import           Data.List ((!!), (\\))
 import           Data.List.NonEmpty (fromList)
 import           Formatting (build, sformat, (%))
-import           Test.Hspec (Spec, describe)
+import           Test.Hspec (Spec, describe, shouldBe)
 import           Test.Hspec.QuickCheck (modifyMaxSuccess)
-import           Test.QuickCheck (arbitrary, choose)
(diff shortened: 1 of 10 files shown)


wallet/test/Test/Pos/Wallet/Web/Methods/PaymentSpec.hs
```diff
@@ -36,21 +40,38 @@ import           Pos.Wallet.Web.Util (decodeCTypeOrFail, getAccountAddrsOrThrow)
...
+data PaymentFixture = PaymentFixture {
+      pswd        :: PassPhrase
+    , dstWalIds   :: [CId Wal]
+    , dstCAddrs   :: [CId Addr]
+    , initBalance :: Coin
+    , policy      :: InputSelectionPolicy
+    , batch       :: NewBatchPayment
+    , srcAddr     :: Address
+    , walId       :: CId Wal
+    , coins       :: [Coin]
```
adinapoli-iohk: I know I'm a terrible person, but please appreciate my laziness 😂

PR #2772, opened 2018-04-10
.md file with profiling instructions

edsko:
This document describes how to build and run for profiling. 

Sadly, `stack` doesn't seem to work; the `nix` option does work but there are some questions there too. It would be good to have some other eyes on this.

@adinapoli-iohk can you ping some relevant people?

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -1,0 +1,123 @@
+# Building with support for profiling
+
+## stack build
+
+Although stack has support for building with profiling [1], for some reason
+it is currently unable to build `cardano-sl-update` with profiling enabled.
+
+```
+stack build --profile cardano-sl-wallet-new:wallet-unit-tests
+```
+
+will result in
+
+```
+    ByteCodeLink.lookupCE
+    During interactive linking, GHCi couldn't find the following symbol:
+      textzm1zi2zi2zi2zm9UQZZjEJZZQFSGMffj1ZZ5g00_DataziText_zdfMonoidTextzuzdczlzg_closure
+    This may be due to you not asking GHCi to load extra object files,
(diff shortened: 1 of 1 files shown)


adinapoli-iohk:
Thank you very much @edsko !

> the nix option does work but there are some questions there too. It would be good to have some other eyes on this.

I see -- @jmitchell , @cleverca22  mind helping us here as well? 😉

adinapoli-iohk:
Also, @erikd is rapidly becoming our "code practice & operational knowledge base" overlord and gate keeper, so I would also like to put this on his radar 😉

edsko:
Possibly related `nix` question, although probably entirely orthogonal to this PR. Running

```
nix-build -A cardano-sl-wallet-new --arg enableProfiling true
```

as in the instructions, then CTRL-Cing this at the very end when it runs the tests for the new wallet (i.e., having built all dependencies), and then running _the exact same thing again_, results in a ton of dependencies being rebuilt again. Dunno why that is. May or may not be due to the use of `src = ./../lib` etc in the generated `default.nix`.

erikd:
I needed to enable stack traces recently and because stack traces use the profiling machinery, I need profiling. However, I only use stack and/or nix to do a final test of my PR before I push it to github.

My usual build tool is [mafia](https://github.com/haskell-mafia/mafia) which I use in conjunction with [jenga](https://github.com/erikd/jenga). With mafia, building with profiling was as easy as `mafia build -p` and after it rebuilt to whole world I got my stack trace.

edsko:
Regarding the profiling build failures: #2720

adinapoli-iohk:
Merging this one overriding CI as it's only touching documentation stuff.

We can improve it later to cover the nix stuff with suggestions from the Devops folks.

PR #2775, opened 2018-04-10
[CSL-2441] Make the mnemonic generator a bit safer.

parsonsmatt:
https://iohk.myjetbrains.com/youtrack/issue/CSL-2441

### DIFF
--- a/wallet/src/Pos/Util/BackupPhrase.hs
+++ b/wallet/src/Pos/Util/BackupPhrase.hs
@@ -31,15 +31,17 @@ newtype BackupPhrase = BackupPhrase
 
 instance Arbitrary BackupPhrase where
     arbitrary = do
-        em <- arbitraryMnemonic 16
+        em <- arbitraryMnemonic 4
         case em of
             Left _  -> arbitrary
             Right a -> pure a
     shrink    = genericShrink
 
+-- | Generate an arbitrary mnemonic with the given number of words.
 arbitraryMnemonic :: Int -> Gen (Either Text BackupPhrase)
 arbitraryMnemonic len = do
-    eitherMnemonic <- toMnemonic . BS.pack <$> vectorOf len arbitrary
+    let byteCount = len * 4
+    eitherMnemonic <- toMnemonic . BS.pack <$> vectorOf byteCount arbitrary
     pure . first toText $ BackupPhrase . words <$> eitherMnemonic
 
 -- | Number of words in backup phrase


wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -31,15 +31,17 @@ newtype BackupPhrase = BackupPhrase
...
         case em of
             Left _  -> arbitrary
             Right a -> pure a
     shrink    = genericShrink
 
+-- | Generate an arbitrary mnemonic with the given number of words.
 arbitraryMnemonic :: Int -> Gen (Either Text BackupPhrase)
 arbitraryMnemonic len = do
-    eitherMnemonic <- toMnemonic . BS.pack <$> vectorOf len arbitrary
+    let byteCount = len * 4
```
akegalj: Hm, what should "word" mean in this context ?

`arbitraryMnemonic 4` will create 12 word mnemonic
`arbitraryMnemonic 3` will create 9 word mnemonic

adinapoli-iohk: I'm confused by this one as well -- wasn't the plan having the input for `arbitraryMnemonic` be the number of english words you wanted to get out of your generator and having `arbitraryMnemonic` calculate the correct number of octets ? I suspect is not sufficient to multiply by 4, we should really take a look at the [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) spec and see how the checksum length is calculated, is probably quite a simple calculation.

akegalj: In that link ^ there is a table how this can be calculated:

CS = ENT / 32
MS = (ENT + CS) / 11

|  ENT  | CS | ENT+CS |  MS  |
+-------+----+--------+------+
|  128  |  4 |   132  |  12  |
|  160  |  5 |   165  |  15  |
|  192  |  6 |   198  |  18  |
|  224  |  7 |   231  |  21  |
|  256  |  8 |   264  |  24  |

MS - number of mnemonic words
CS - this is what you are refering as "word"

Note that 9 word mnemonic has been depricated last year but we use it for redemption phase. This should look like:
|  96  |  3 |   99  |  9  |

adinapoli-iohk: Good find @akegalj -- indeed that's the table we need (and `CS` stands for _checksum_, is the number of bytes you need extra on top of the space occupied by the mnemonic words (here called `MS`).

This also exposes another interesting problem, which is that this table represents a finite enumeration, so if we really want to be on the same side, I propose to create a simple bounded enumeration we pass as input as part of `arbitraryMnemonic` which would make impossible states unrepresentable as in:

```
data Mnemonics =
    Nine
  | Twelve
  | Fifteen
  | Eighteen
  | Twentyone
  | Twentyfour
```

Which could yield an API like `arbitraryMnemonicOfLength Twelve` and can later be used to do explicit pattern matching when we calculate the checksums.

I suck at names, but I hope you get the idea. @parsonsmatt do you think is overkill? I think we should limit the degree of freedom of what gets passed into our `arbitraryMnemonic` function.

akegalj: And this ^ proposal can be very useful later for https://iohk.myjetbrains.com/youtrack/issue/CSL-2048 and this part https://github.com/input-output-hk/cardano-sl/blob/akegalj/csl2048-backup-phrase/wallet/src/Pos/Util/BackupPhrase.hs#L30 : 
```haskell

data MnemonicType
    = PaperVendMnemonic -- 9 word mnemonic used for paper vend redemption
    | BackupMnemonic    -- 12 word mnemonic used as a wallet backup
    deriving (Bounded, Enum)
```

which is exactly what I think we need for that task

adinapoli-iohk: yes, your `MnemonicType` in a sense is more domain-specific and would allow us to talk about mnemonics and their different lengths in the context of Daedalus/Cardano, whereas that naive `Mnemonics` is more _primitive_ , in the sense it's independent by any business domain but it taps directly into BIP39, which sits at a lower level of abstraction.

parsonsmatt: Wonderful :) As they say, the best way to get help on the internet is to assert something incorrect :smile: I'll incorporate these fixes.

wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -29,17 +29,36 @@ newtype BackupPhrase = BackupPhrase
...
+data MnemonicWordCount
+    = Nine
+    | Twelve
+    | Fifteen
+    | Eighteen
+    | Twentyone
+    | Twentyfour
+    deriving (Eq, Show)
+
+byteCount :: MnemonicWordCount -> Int
```
adinapoli-iohk: It would be nice to post here the link to the BIP39 proposal, so that people know where we picked this magic numbers from 😉 

We should really call this function _checksumLength_ in my opinion, as this is the size of the checksum, in reality.

wallet/src/Pos/Util/BackupPhrase.hs
```diff
@@ -29,17 +29,36 @@ newtype BackupPhrase = BackupPhrase
...
             Right a -> pure a
     shrink    = genericShrink
 
-arbitraryMnemonic :: Int -> Gen (Either Text BackupPhrase)
-arbitraryMnemonic len = do
-    eitherMnemonic <- toMnemonic . BS.pack <$> vectorOf len arbitrary
+-- | Generate an arbitrary mnemonic with the given number of words.
+arbitraryMnemonic :: MnemonicWordCount -> Gen (Either Text BackupPhrase)
+arbitraryMnemonic wordCount = do
+    eitherMnemonic <- toMnemonic . BS.pack <$> vectorOf (byteCount wordCount) arbitrary
```
adinapoli-iohk: I might be undercaffeinated but this is not quite right, isn't it? You are saying that we want a vector of `byteCount Twelve`, which is 4, but note how `byteCount` gives you back the size of the _checksum_, which are _extra bytes_ which needs to be added to the total size.

Therefore, what you really want is `vectorOf ((byteCount wordCount) + 12))`.

I suggest we write also a function that given a `MnemonicWordCount` gives you back the size in bytes. If you take a look at that BIP39 table:

```
| ENT | CS | ENT+CS | MS |
+-------+----+--------+------+
| 128 | 4 | 132 | 12 |
| 160 | 5 | 165 | 15 |
| 192 | 6 | 198 | 18 |
| 224 | 7 | 231 | 21 |
| 256 | 8 | 264 | 24 |
```

Here `byteCount` gives you back only `CS`, but what we need to feed into `vectorOf` is `CS + MS` 😉 

If you take a look at the original implementation, that was exactly where this 16 was coming from for a 12-word mnemonic: 12 `Word8` octets + 4 bytes for the checksum .

parsonsmatt: Thanks for catching that. I will reproduce the table as a function :)

parsonsmatt:
Alas, it is now 6:05pm GMT, just missing the cutoff :frowning_face:

PR #2776, opened 2018-04-11
[CBR-181] Describe API errors in a developer-friendly way

KtorZ:
And add this description to the current Swagger API doc. I've also reviewed a bit how the examples where displayed because I wasn't really satisfied with this long unreadable line ^.^

> NOTE: I've removed the `SomeOtherError { foo , bar }` cosntructor for it wasn't used anywhere. I believe it was an old dev / debug artifact ?

Preview for your (hopefully) mesmerized eyes:

screenshot from 2018-04-11 12-43-50

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveGeneric     #-}
+{-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell   #-}
 
@@ -8,7 +9,7 @@ import           Universum
 
 import           Cardano.Wallet.API.Response.JSend (ResponseStatus (ErrorStatus))
 import           Data.Aeson
-import           Data.List.NonEmpty (NonEmpty)
+import           Data.List.NonEmpty (NonEmpty ((:|)))
 import           Generics.SOP.TH (deriveGeneric)
 import qualified Network.HTTP.Types.Header as HTTP
 import           Servant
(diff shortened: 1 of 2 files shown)


wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -50,7 +51,6 @@ import           Cardano.Wallet.API.V1.Generic (gparseJsend, gtoJsend)
 data WalletError =
       NotEnoughMoney { weNeedMore :: !Int }
     | OutputIsRedeem { weAddress :: !Text }
-    | SomeOtherError { weFoo :: !Text, weBar :: !Int }
```
adinapoli-iohk: Yes, as you guessed this `SomeOtherError` was used just as a witness it was possible to generate jsend-style errors with more than 1 field in the `diagnostic` section, and encode everything using `generics-sop`. In practice, it's for debugging purposes only and I agree it's better to remove it.

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -91,35 +92,51 @@ type ErrorExample = Value
 sample :: [WalletError]
 sample =
   [ NotEnoughMoney 1400
-  , OutputIsRedeem "b10b24203f1f0cadffcfd16277125cf7f3ad598983bef9123be80d93"
-  , SomeOtherError "foo" 14
-  , MigrationFailed "migration"
+  , OutputIsRedeem "b10b242...be80d93"
```
adinapoli-iohk: Don't you think we should use the long representation? As this is intended to be a "real-world" representation of an error, it would be a shame if somehow somebody could misinterpret that we are rendering addresses in such short notation. What do you think?

KtorZ: This one gave me a hard time thinking actually. The reason why I truncated here is because these samples are used for documentation and, since we render the examples from those samples, the particular one yield to a "wide" example completely messing up the whole formatting of the table. Could be fixed with a bit of css, yet having simply a more human-readable address here seemed more _legit_ to me :/

adinapoli-iohk: hehe ok, fair enough -- definitely not a big deal, but I'm glad it gave you food for thoughts 😉

PR #2780, opened 2018-04-12
[CSL-2457] Document flushing log to disk with cardano-node

KtorZ:
See [CSL-2457](https://iohk.myjetbrains.com/youtrack/issue/CSL-2457)


adinapoli-iohk:
This is touching documentation only, ignoring bonkers CI.

PR #2800, opened 2018-04-19
CSL-2175 make things more polymorphic by removing HasConfiguration

cleverca22:


db/Pos/DB/Class.hs
```diff
@@ -95,7 +95,7 @@ type SerializedBlock = Serialized SerBlock
 type SerializedUndo = Serialized SerUndo
 
 -- | Pure read-only interface to the database.
-class (HasConfiguration, MonadThrow m) => MonadDBRead m where
```
taktoa: Huh, it seems very strange to me that someone would ever add a nullary constraint like `HasConfiguration` to a _class_, rather than to its instances. It doesn't constrain the set of instances at all; all it does is act like an obfuscated constraint synonym.

block/src/Pos/Block/Logic/Internal.hs
```diff
@@ -231,40 +236,40 @@ rollbackBlocksUnsafe bsc scb toRollback = do
...
-toComponentBlock :: HasConfiguration => (MainBlock -> payload) -> Block -> ComponentBlock payload
+toComponentBlock :: (MainBlock -> payload) -> Block -> ComponentBlock payload
 toComponentBlock fnc block = case block of
     Left genBlock   -> ComponentBlockGenesis (convertGenesis genBlock)
     Right mainBlock -> ComponentBlockMain (Some $ mainBlock ^. gbHeader) (fnc mainBlock)
 
 toTxpBlock
-    :: HasConfiguration
+    :: ()
     => Block -> TxpBlock
```
erikd: You should probably drop the constraint on that all together so it becomes:
```
toTxpBlock
    :: Block -> TxpBlock
```
and similar for the rest.

block/src/Pos/Arbitrary/Block/Message.hs
```diff
@@ -12,25 +12,26 @@ import           Pos.Arbitrary.Txp ()
...
+instance Arbitrary T.MsgGetBlocks where
     arbitrary = genericArbitrary
     shrink = genericShrink
 
 instance ( Arbitrary SscPayload
          , Arbitrary SscProof
          , Bi Raw
-         , HasConfiguration
+         , HasProtocolConstants
+         , HasProtocolMagic
```
parsonsmatt: While we're touching these constraints, can we drop the monomorphic `Arbitrary {SscPayload,SscProof}` instances? These defer constraint solving errors to the use-site instead of the definition-site.

parsonsmatt: Unless... oh god... we're relying on orphan instances that aren't in scope :sweat:

avieth:
Nice, happy to see this. I'll go through it now. Ideally, we'd get rid of these constraints altogether and just take the relevant thing as an argument, but this is a fine step in that direction.

block/src/Pos/Arbitrary/Block/Message.hs
erikd: We should work on removing orphan instances.

PR #2806, opened 2018-04-20
[DEVOPS-810] reduce the useless rebuilds in wallet-new due to nix and gitrev

cleverca22:
[DEVOPS-810]

todo:
- [ ] cardano-sl-auxx
- [ ] cardano-sl-explorer-static
- [ ] cardano-sl-tools
- [ ] cardano-sl-node
- [ ] cardano-sl-node-static

### DIFF
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -108,23 +108,38 @@ library
   ghc-options:         -Wall
 
   build-depends:       base
+                     , QuickCheck
                      , acid-state
                      , aeson
                      , aeson-pretty
                      , bytestring
                      , cardano-sl
+                     , cardano-sl
+                     , cardano-sl-block
                      , cardano-sl-block
                      , cardano-sl-client
+                     , cardano-sl-client
+                     , cardano-sl-core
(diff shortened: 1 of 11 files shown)


wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -14,7 +14,10 @@ extra-source-files:  README.md
 cabal-version:       >=1.10
 
 library
-  hs-source-dirs:      src
+  hs-source-dirs:      src server
```
parsonsmatt: If we're going to do this I want to move all the `server` code into `src`. Can happen at a later date/PR.

wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -14,7 +14,10 @@ extra-source-files:  README.md
...
+  ghc-options:         -Wall
+                       -fno-warn-orphans
```
parsonsmatt: :(

I'd rather put this in each file than here. Also can be another PR.

cleverca22: i believe that was in the old cabal file, just moved a bit

PR #2808, opened 2018-04-22
Remove generated file

erikd:
Remove a generated file with paths specific to the user who originally added it to the repo.


erikd:
`*/dist` is already in `.gitignore`. Not sure how this one slipped in.

PR #2811, opened 2018-04-23
[CSL-2425] Implement WalletActiveLayer for transactions

ksaric:
## Description

Implement WalletActiveLayer for transactions.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CSL-2425

## Type of change

- [X] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

## Developer checklist

- [X] I have updated any documentation accordingly if needed. Documentation changes can be reflected in opening a PR on [cardanodocs.com](https://github.com/input-output-hk/cardanodocs.com), amending the inline [Haddock](https://www.haskell.org/haddock/) comments, any relevant README file or one of the documents listed in the [docs](https://github.com/input-output-hk/cardano-sl/tree/develop/docs) directory.

## Testing checklist

- [ ] All new and existing tests pass.

### DIFF
--- a/wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
+++ b/wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
@@ -10,13 +10,13 @@ module Cardano.Wallet.WalletLayer.Legacy
 import           Universum
 
 import           Control.Monad.IO.Unlift (MonadUnliftIO)
+import           Control.Exception.Safe (try)
 import           Data.Coerce (coerce)
 
-import           Cardano.Wallet.WalletLayer.Error (WalletLayerError (..))
+import           Cardano.Wallet.WalletLayer.Error
 import           Cardano.Wallet.WalletLayer.Legacy.Transactions (pwlCreateTx, pwlEstimateFees,
                                                                  pwlGetTxs)
-import           Cardano.Wallet.WalletLayer.Types (ActiveWalletLayer (..), PassiveWalletLayer (..),
-                                                   monadThrowToEither)
+import           Cardano.Wallet.WalletLayer.Types (ActiveWalletLayer (..), PassiveWalletLayer (..))
 
 import           Cardano.Wallet.Kernel.Diffusion (WalletDiffusion (..))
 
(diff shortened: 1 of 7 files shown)


wallet-new/src/Cardano/Wallet/WalletLayer/Legacy/Transactions.hs
```diff
@@ -0,0 +1,113 @@
...
+            -- Create a `[V0.AccountId]` to get txs from it
+            let accIds = case mAccIdx of
+                    Just accIdx -> migrate (walletId, accIdx)
+                    -- ^ Migrate `V1.AccountId` into `V0.AccountId` and put it into a list
+                    Nothing     -> V0.getWalletAccountIds ws cIdWallet
+                    -- ^ Or get all `V0.AccountId`s of a wallet
+
+            let v0Addr = case mAddr of
+                    Nothing        -> Nothing
+                    Just (V1 addr) -> Just $ V0.encodeCType addr
```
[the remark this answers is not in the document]
ksaric: This part of the code was literally copied. This should be the same as existing code so we can confirm it works the same. I know it's not a great answer, but that's the way it is for transactions for now.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -128,21 +150,21 @@ pwlGetWalletIds = do
...
-    -> m (Maybe Wallet)
+    -> m Wallet
 pwlGetWallet wId = do
     ws          <- askWalletSnapshot
 
     cWId        <- migrate wId
     wallet      <- V0.getWallet cWId
 
-    pure $ do
+    maybeThrow (WalletNotFound wId) $ do
```
parsonsmatt: I would prefer to see this written as:

```haskell
pwlGetWallet :: WalletId -> m (Either GetWalletError Wallet)
```

(provided that `GetWalletError` contains more information and is more valuable than `Nothing`)

We should really be trying to avoid `MonadThrow` as much as possible and representing things as values whenever possible. `MonadThrow` is a way of evading the type system, and we have no need to do that in simple cases like `pwlGetWallet`.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -159,24 +181,24 @@ pwlUpdateWallet wId wUpdate = do
     setWalletMeta walletDB cWId cWMeta
 
     -- Get wallet or throw if missing.
-    maybeThrow (WalletNotFound wId) =<< pwlGetWallet wId
+    pwlGetWallet wId
```
parsonsmatt: If `pwlGetWallet` returns `Maybe`, then this logic also needs to change.

If we update a wallet that doesn't exist, what happens? Seems like the same problem: it should return a value indicating `WalletNotFound` (whether that is `Nothing` or a `Left (WalletNotFound ...)`. Let's reflect that in the types.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -159,24 +181,24 @@ pwlUpdateWallet wId wUpdate = do
...
     => WalletId
-    -> m Bool
+    -> m ()
 pwlDeleteWallet wId = do
     cWId        <- migrate wId
-    -- TODO(ks): It would be better to catch specific @Exception@.
-    -- Maybe @try@?
-    catchAll (const True <$> V0.deleteWallet cWId) (const . pure $ False)
+    _           <- V0.deleteWallet cWId
+    pure ()
```
parsonsmatt: I'd write this as `void $ v0.deleteWallet cWId`.

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -200,15 +223,17 @@ pwlGetAccounts wId = do
     cAccounts   <- V0.getAccounts $ Just cWId
     migrate cAccounts
 
+
 pwlGetAccount
     :: forall ctx m. (MonadLegacyWallet ctx m)
     => WalletId
     -> AccountIndex
-    -> m (Maybe Account)
+    -> m Account
```
parsonsmatt: let's keep the `Maybe` here

wallet-new/src/Cardano/Wallet/WalletLayer/Legacy.hs
```diff
@@ -222,19 +247,72 @@ pwlUpdateAccount wId accIdx accUpdate = do
...
     => WalletId
     -> AccountIndex
-    -> m Bool
+    -> m ()
 pwlDeleteAccount wId accIdx = do
-    accId <- migrate (wId, accIdx)
-    catchAll (const True <$> V0.deleteAccount accId) (const . pure $ False)
+    accId       <- migrate (wId, accIdx)
+    _           <- V0.deleteAccount accId
+    pure ()
```
parsonsmatt: `void $ ...`

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -14,39 +18,65 @@ module Cardano.Wallet.WalletLayer.Types
...
+    , _pwlDeleteAccount  :: WalletId -> AccountIndex -> WalletLayerResponse m ()
     -- * addresses
-    , _pwlGetAddresses  :: WalletId -> m [Address]
+    , _pwlCreateAddress  :: NewAddress -> WalletLayerResponse m WalletAddress
+    , _pwlGetAddresses   :: WalletId -> Maybe AccountIndex -> WalletLayerResponse m [WalletAddress]
+    , _pwlIsAddressValid :: WalletAddress -> WalletLayerResponse m AddressValidity
+    -- * transactions
+    , _pwlCreateTx       :: (TxAux -> m Bool) -> Payment -> WalletLayerResponse m Transaction
+    , _pwlGetTxs         :: Maybe WalletId -> Maybe AccountIndex -> Maybe (V1 Address) -> WalletLayerResponse m [Transaction]
+    , _pwlEstimateFees   :: Payment -> WalletLayerResponse m EstimatedFees
```
parsonsmatt: style note: putting the `::` on a newline will provide for cleaner diffs and make it easier to know what actually changed. eg:

```haskell
data X = X
    { foo
        :: Int -> IO ()
    , bar
        :: Double -> IO Char
    }
```

wallet-new/src/Cardano/Wallet/WalletLayer/Types.hs
```diff
@@ -14,39 +18,65 @@ module Cardano.Wallet.WalletLayer.Types
...
 
 import           Cardano.Wallet.Kernel.Diffusion (WalletDiffusion (..))
+import           Cardano.Wallet.WalletLayer.Error (WalletLayerError (..))
 
 ------------------------------------------------------------
 -- Passive wallet layer
 ------------------------------------------------------------
 
+-- The response for the @WalletPassiveLayer@.
+type WalletLayerResponse m a = m (Either WalletLayerError a)
```
parsonsmatt: I know that this is how the HTTP client does things, but the reason I chose that representation is because of the HTTP-nature of the program -- HTTP stuff can always fail (eg with a `ServantErr`), so it makes sense to wrap all of the endpoints with that. However, the `WalletLayer` is operating locally on a much more trusted system -- the local database. Because of that, we should *expect* that a catastrophic, transport-level failure is much more rare, and we can use things like "The database disappeared off of the disk" as an exceptional situation.

For the wallet layer, I would prefer to see the possible failures for an action represented in the return type of the action. If getting wallets has an error, then the return of `pwlGetWallets` should have `Either GetWalletError Wallet`. Same for accounts, etc. 

This would mean breaking up the `WalletLayerError` into smaller, finer types, and returning those from relevant functions instead of the monolithic error type.

ksaric:
Requires https://github.com/input-output-hk/cardano-sl/pull/2778 to be merged first.

PR #2812, opened 2018-04-23
[CSL-2493] Fee estimation endpoint returns an error if the account has insufficient funds

parsonsmatt:
## Description

This PR fixes the fee estimation endpoint. When the underlying transaction layer returns a `Left (NotEnoughMoney n)`, we convert this into an `EstimatedFee` as a lower bound now.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CSL-2493

## Type of change

## Developer checklist

## Testing checklist

This PR fixes a previously failing test case.

## QA Steps

Run the integration tests.

## Screenshots (if available)

image


wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -2,44 +2,48 @@ module Cardano.Wallet.API.V1.LegacyHandlers.Transactions where
 
 import           Universum
 
-import           Cardano.Wallet.API.Request
```
parsonsmatt: I reorganized the imports to have `Prelude -> Libraries -> IOHK packages -> Internal modules` ordering

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -2,44 +2,48 @@ module Cardano.Wallet.API.V1.LegacyHandlers.Transactions where
...
+    :: (HasConfigurations, HasCompileInfo)
+    => (TxAux -> MonadV1 Bool)
+    -> ServerT Transactions.API MonadV1
 handlers submitTx =
-             newTransaction submitTx
-        :<|> allTransactions
-        :<|> estimateFees
+         newTransaction submitTx
+    :<|> allTransactions
+    :<|> estimateFees
```
parsonsmatt: regularized indentation and formatting

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -2,44 +2,48 @@ module Cardano.Wallet.API.V1.LegacyHandlers.Transactions where
...
 
 newTransaction
-    :: forall ctx m . (V0.MonadWalletTxFull ctx m)
-    => (TxAux -> m Bool) -> Payment -> m (WalletResponse Transaction)
+    :: forall ctx m. (V0.MonadWalletTxFull ctx m)
+    => (TxAux -> m Bool)
+    -> Payment
+    -> m (WalletResponse Transaction)
```
parsonsmatt: just formatting, again

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1018,12 +1019,29 @@ instance BuildableSafeGen PasswordUpdate where
...
-data EstimatedFees = EstimatedFees {
-    feeEstimatedAmount :: !(V1 Core.Coin)
-  } deriving (Show, Eq, Generic)
+data EstimatedFees = EstimatedFees
+    { feeEstimatedAmount :: !(V1 Core.Coin)
+    -- ^ The estimated amount.
+    , feeAccuracy :: !Accuracy
+    -- ^ The accuracy of this estimate. If an account has insufficient
+    -- funds to perform a transaction, then we return 'LowerBound', as
+    -- adding additional inputs to a transaction will increase the fee.
```
parsonsmatt: I know that I talked with @adinapoli-iohk about keeping the error response, but this ended up being an easier way to go. It also preserves the "successful" semantics vs an error condition.

We say *estimated* fees, which is why I am okay with this change.

If we decide to go and do an HTTP error, then I can change this.

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -118,6 +122,46 @@ estimateFees Payment{..} = do
...
+        -> Core.Coin
+        -- ^ The amount of coins necessary to complete the transaction.
+        -> EstimatedFees
+    mkLowerBound acct amountUnder =
+        EstimatedFees (V1 (Core.mkCoin (fromIntegral feeLowerBound))) LowerBound
+      where
+        feeLowerBound =
+            totalToSpend - txnTotal
+        totalToSpend =
+            Core.getCoin amountUnder + acctAmount
```
parsonsmatt: i am absolutely terrible at math, so someone please check me on this:

example:

Attempt to pay 90 coins
I have 100 coins in balance
I get the error back "Need 10 more coins"
The assumed total of transaction is (balance + amount needed = 110 coins)
The assumed fee is the assumed total minus the amount we try to pay (20 coins).

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1018,30 +1019,68 @@ instance BuildableSafeGen PasswordUpdate where
         pwdOld
         pwdNew
 
+data Accuracy
+    = Accurate
+    | LowerBound
```
adinapoli-iohk: What about some comments for the lost soul trying to find purpose into this type? 😀 I know it's self descriptive, but let's try to be good open source citizens 😁

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1018,30 +1019,68 @@ instance BuildableSafeGen PasswordUpdate where
...
+    buildSafeGen _ = show
+
+instance ToJSON Accuracy where
+    toJSON = String . show
+
+instance FromJSON Accuracy where
+    parseJSON = withText "Accuracy" $ \str ->
+        case str of
+            "Accurate" -> pure Accurate
+            "LowerBound" -> pure LowerBound
```
adinapoli-iohk: I might be wrong, but I'm decently sure the style we picked was camel case, but _non-capitalised_, for enumerations. If I'm right, we should strive to keep everything uniform and maybe for the future even write a Swagger test to check that 😀

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -164,6 +165,23 @@ toHttpStatus :: WalletError -> HTTP.Status
...
+    TxError.NotEnoughMoney coin ->
+        NotEnoughMoney (fromIntegral (Core.getCoin coin))
+    TxError.NotEnoughAllowedMoney coin ->
+        NotEnoughMoney (fromIntegral (Core.getCoin coin))
+    TxError.FailedToStabilize ->
+        UnknownError "Transaction process failed to stabilize."
+    TxError.OutputIsRedeem addr ->
+        OutputIsRedeem (show addr)
+    TxError.RedemptionDepleted ->
+        UnknownError "Redemption Depleted"
```
adinapoli-iohk: At the cost of sounding pedagogical (sorry, not my intention) I'm wondering if we should apply the same style advocated here:

https://github.com/input-output-hk/cardano-sl/pull/2816#discussion_r183640924

Coalescing everything into a stringly-typed `UnknownError` looks like a missed opportunity here 😉

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -164,6 +165,23 @@ toHttpStatus :: WalletError -> HTTP.Status
...
+transactionErrorToWalletError :: TxError.TxError -> WalletError
+transactionErrorToWalletError = \case
+    TxError.NotEnoughMoney coin ->
+        NotEnoughMoney (fromIntegral (Core.getCoin coin))
+    TxError.NotEnoughAllowedMoney coin ->
+        NotEnoughMoney (fromIntegral (Core.getCoin coin))
+    TxError.FailedToStabilize ->
+        UnknownError "Transaction process failed to stabilize."
+    TxError.OutputIsRedeem addr ->
+        OutputIsRedeem (show addr)
```
adinapoli-iohk: The type of `OutputIsRedeem` is `OutputIsRedeem !Address`, which means we are being passed a nice Cardano's `Address` but we are now flattening it into a raw string.

This wouldn't be in principle a problem here, if not that showing it via the `Show` instance will (AFAIK) bump the raw internal structure as an output -- what you probably want is to use the `Buildable` instance, which will render this as a nice base58 string 😉 

I'm wondering if indeed our `WalletError.OutputIsRedeem` is a poor constructor -- what about:

1. Changing its type to take a `V1 Address` (basically almost _exactly_ as `TxError.OutputIsRedeem`)?

If I'm not mistaken, this will also ensure that rendering such type constructor as a `JSON` will trigger `ToJSON (V1 Address)` which should correctly render it, as `ToJSON (V1 Address)` is defined as:

```haskell
instance ToJSON (V1 Core.Address) where
    toJSON (V1 c) = String $ sformat addressF c
```

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -118,6 +122,46 @@ estimateFees Payment{..} = do
...
     outputs <- V0.coinDistrToOutputs =<< mapM migrate pmtDestinations
-    fee <- V0.rewrapTxError "Cannot compute transaction fee" $
-        eitherToThrow =<< V0.runTxCreator policy (V0.computeTxFee pendingAddrs utxo outputs)
-    single <$> migrate fee
+    efee <- V0.runTxCreator policy (V0.computeTxFee pendingAddrs utxo outputs)
+    single <$> case efee of
+        Left txError ->
+            case txError of
+                V0.NotEnoughMoney coin -> do
+                    acct <- V0.getAccount cAccountId
```
adinapoli-iohk: Possibly not relevant here as this is a "benign" endpoint, but this doesn't look fully  _data-race free_, if I'm not mistaken. You could potentially, in a very fast-paced environment, yield an incorrect calculation if, by the time you grab the account via `getAccount`, the utxo of that account has changed.

But look, you already grabbed the `utxo` for that `accountId` via `getMoneySourceUtxo`, so you should be able to compute this light-hearted if you pass the `Utxo` as input to `mkLowerBound` and fold over it to grab the total `Coin`s for that account.

If the semantic is the same I think it is, `getMoneySourceUtxo` should give you only the _account-relevant_ portion of the available Utxo.

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -118,6 +122,46 @@ estimateFees Payment{..} = do
...
+        EstimatedFees (V1 (Core.mkCoin (fromIntegral feeLowerBound))) LowerBound
+      where
+        feeLowerBound =
+            totalToSpend - txnTotal
+        totalToSpend =
+            Core.getCoin amountUnder + acctAmount
+        txnTotal =
+            sum (fmap (Core.getCoin . unV1 . pdAmount) pmtDestinations)
+        acctAmount =
+            case V0.caAmount acct of
```
adinapoli-iohk: I might be wrong, but if you use the `Utxo` approach as I was suggesting, you now get a proper `Coin` type, not an awful `CCoin`, and all this code goes away.

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
parsonsmatt: > At the cost of sounding pedagogical

At the cost of sounding pedantic, do you mean "pedantic"? :trollface: 

This is a good point :smile:

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
parsonsmatt: Good call :)

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
adinapoli-iohk: Let's translate your example into code, but let's assume something different: _we need to pay 100 but we only have 90 in the balance_ . What you posted in your example is a perfectly valid transaction as I do have enough money to pay for it -- what am I missing? 😀 . We have:

```haskell
let amountUnder = 10 -- let's assume I have 90 coins, I need 10 more to fullfill 100.
let fee= EstimatedFees (V1 (Core.mkCoin (fromIntegral feeLowerBound))) LowerBound
let feeLowerBound = totalToSpend - txnTotal
let totalToSpend = Core.getCoin amountUnder + acctAmount
let txnTotal = sum (fmap (Core.getCoin . unV1 . pdAmount) pmtDestinations)
let acctAmount = sum (fmap (Core.getCoin . Txp.txOutValue . Txp.toaOut) utxo)
```

reducing:

```
let amountUnder = 10
let fee= EstimatedFees (V1 (Core.mkCoin (fromIntegral feeLowerBound))) LowerBound
let feeLowerBound = totalToSpend - txnTotal
let totalToSpend = 10 + acctAmount
let txnTotal = 100 -- we need to pay 100 coins to all the various destinations.
let acctAmount = 90 -- we have 90 coins in our "bank"

=> 

let amountUnder = 10
let fee= EstimatedFees (V1 (Core.mkCoin (fromIntegral feeLowerBound))) LowerBound
let feeLowerBound = 100 - 100
let totalToSpend = 10 + 90 = 100
let txnTotal = 100
let acctAmount = 90
```

This shows that essentially `totalToSpend == txnTotal` and that your fee calculation will always be 0 if you assume that you have _exactly_ the required amount of money to fullfil the contract. But of course a `Fee` of 0 is not correct, and I recommend not going into that rabbit hole. After reading [our documentation](https://cardanodocs.com/cardano/transaction-fees/) here is what I would do: I would reuse the formula in the docs:

```
a + b × size
```

Where `a` and `b` are constants, and `size` is the only unknown. Of course, we cannot determine `size` _apriori_ without having a valid Tx at hand, so I would suggest reusing the `txToLinearFee` function to do so:

``` haskell
txToLinearFee
    :: MonadError TxError m
    => TxSizeLinear -> TxAux -> m TxFee
txToLinearFee linearPolicy =
    fixedToFee .
    calculateTxSizeLinear linearPolicy .
    biSize @TxAux
```

Of course do don't have all the inputs required to generate a valid fee, but this shouldn't matter as this is a lower bound -- I would give a shot to `createFakeTxFromRawTx` to see if we can create a fake transaction with essentially all the inputs we own (and the outputs we need to pay for) -- granted, it won't be _valid_, but that's is the closest approximation in terms of bytes we have at the moment.

What do you think?

parsonsmatt: I don't think that approach buys us anything -- the `TxError` type that we get back is coming from one of these functions anyway, and if we get another "Need more money pls" error then it doesn't do much for us :)

I'll write some tests for the math and factor it out into it's own function.

wallet-new/test/API/TransactionsSpec.hs
```diff
@@ -0,0 +1,37 @@
...
+                `shouldBe`
+                    20
+
+        it "correctly calculates the amount #2" $ do
+            rawMkLowerBound
+                (TxnAmount 50)
+                (AvailableCoin 100)
+                (AdditionalCoin 10)
+                `shouldBe`
+                    60
```
parsonsmatt: @adinapoli-iohk I hope you forgive me for keeping the "estimate based on error message" implementation. I wrote some tests to make up for it :smile:

wallet-new/server/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
adinapoli-iohk: My point is that if we are not following a specification in order to compute even a lower bound, what is the value of reporting it 😀 ? Imho the right way to go would still to apply the formula, otherwise we are effectively using two different fees calculations in two different circumstances, no? 

To be sure to be on the same page, I'm not advocating to modify the internal or to go for any weird logic, but just to swap the code inside `mkLowerBound` to reuse the same formula we use for the accurate calculation, as that will likely give more realistic calculations 😉 

Happy to have a voice chat about that -- in the meantime you are asleep will ask also other folks who knows better than me when it comes to fee calculation, so see if we can find the sweet spot between the two approaches :)

nikolaglumac:
@parsonsmatt @adinapoli-iohk please see my comment on this Youtrack card: https://iohk.myjetbrains.com/youtrack/issue/CSL-2493#comment=93-20809
Thanks!

adinapoli-iohk:
Cheers @nikolaglumac !

@parsonsmatt I'm very sorry for having dragged you into the rabbit hole with this, and I should have taken more active interest in the planning of the feature (sadly for me, I was on annual leave when this was discussed and failed to catch up properly).

After a conversation with @dcoutts & the others, we all feel we took a wrong turn with this one. So, as reluctantly as I am, I need to abort this ticket and close it, sorry. 😢

parsonsmatt:
No worries! :) I'm glad we got it sorted out.

PR #2818, opened 2018-04-24
[CBR-190] Remove address balance from wallets' addresses

KtorZ:
## Description

Computing this balance is an expensive and innacurate operation that we most probably don't
want to support in the future. If needed, it will be 'easy' to re-introduce than to remove
in a non-backward compatible way. Let's therefore remove it before shipping API V1.

> **NOTE** It's normal if you feel like you have a déjà-vu (see #2817)

## Linked issue

[CBR-190](https://iohk.myjetbrains.com/youtrack/issue/CBR-190)

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/integration/Functions.hs
+++ b/wallet-new/integration/Functions.hs
@@ -415,10 +415,6 @@ runAction wc action = do
             log $ "Posting address: " <> ppShowT newAddress
             result  <-  respToRes $ postAddress wc newAddress
 
-            checkInvariant
-                (addrBalance result == minBound)
-                (AddressBalanceNotZero result)
-
             -- Modify wallet state accordingly.
             addresses  <>= [result]
             accounts . traverse . filtered (== account) %= \acct ->
@@ -450,7 +446,6 @@ runAction wc action = do
         -- Transactions
         PostTransaction -> do
             localAccounts  <- use accounts
-            localAddresses <- use addresses
 
             -- Some min amount of money so we can send a transaction?
             -- https://github.com/input-output-hk/cardano-sl/blob/develop/lib/configuration.yaml#L228
@@ -459,9 +454,9 @@ runAction wc action = do
 
             -- From which source to pay.
             accountSource <- pickRandomElement localAccsWithMoney
+            accountDestination <- pickRandomElement (localAccounts \\ [accountSource])
 
             let accountSourceMoney = accAmount accountSource
-                withoutSourceAddresses = filter (`notElem` accAddresses accountSource) localAddresses
                 reasonableFee = 100
 
             -- We should probably have a sensible minimum value.
@@ -478,7 +473,7 @@ runAction wc action = do
                         , psAccountIndex = accIndex    accountSource
                         }
 
-            addressDestination <- pickRandomElement withoutSourceAddresses
+            addressDestination <- pickRandomElement $ accAddresses accountDestination
 
             let paymentDestinations =
                     PaymentDistribution
@@ -535,7 +530,7 @@ runAction wc action = do
 
             let changeAddress = toList (txOutputs newTx) \\ toList paymentDestinations
                 -- NOTE: instead of this manual conversion we could filter WalletAddress from getAddressIndex
-                pdToChangeAddress PaymentDistribution{..} = WalletAddress pdAddress pdAmount True True
+                pdToChangeAddress PaymentDistribution{..} = WalletAddress pdAddress True True
                 realChangeAddressId = map addrId addressesAfterTransaction \\ map addrId addressesBeforeTransaction
                 changeWalletAddresses = filter ((`elem` realChangeAddressId) . addrId) addressesAfterTransaction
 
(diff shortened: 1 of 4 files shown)


wallet-new/integration/Functions.hs
```diff
@@ -460,8 +455,10 @@ runAction wc action = do
             -- From which source to pay.
             accountSource <- pickRandomElement localAccsWithMoney
 
+            let but xs x = filter (/= x) xs
+            accountDestination <- pickRandomElement (localAccounts `but` accountSource)
```
adinapoli-iohk: Isn't this virtually the same as `Data.List.(\\)`?

wallet-new/integration/Functions.hs
```diff
@@ -548,26 +545,26 @@ runAction wc action = do
...
-                            Nothing -> throwM $ CantFindAddress pdAddress
-
-            -- Check did addresses of all payment sources decrease by expected amount after transaction
-            checkWalletAddressAfter (txInputs newTx) $ flip unsafeSubCoin
-
-            -- Check did addresses of all payment destinations increase by expected amount after transaction
-            -- NOTE: we are using paymentDestinations instead of txOutputs to eliminate changeAddress
-            checkWalletAddressAfter paymentDestinations unsafeAddCoin
+
+            accountSourceAfter <-
```
adinapoli-iohk: I reckon this is just formatting, right?

wallet-new/integration/Functions.hs
KtorZ: Provided all accounts in the list are unique, yes (`\\` removes only the first occurrence). Will replace.

wallet-new/integration/Functions.hs
KtorZ: Nope, that's actually the interesting part. The invariant before was checking that each addresses balance either decreased or increased according to the payment distribution. I replaced that to have the check done on the accounts' balances instead.

KtorZ:
> Did we assess the integration tests still run OK?

I did make sure that the `PostTransaction` action runs correctly. Integration tests are still failing because of the estimation fees test but otherwise ran ok.

wallet-new/integration/Functions.hs
adinapoli-iohk: Ah, I have to admit I missed that subtle detail -- probably yours is safer, but in theory here it shouldn't matter as account (and their indexes) are definitely unique per-wallet 😉

EdwardPrentice:
Please forgive my lack of understanding - by removing this API how are we expecting clients/developers to get the balance of an arbitrary address?

> an expensive and inaccurate operation

Surely not both? How can this operation be inaccurate? Again I would be asking - what mechanism is there for a client/developer to accurately view the balance for a given address?

Thanks,

adinapoli-iohk:
Hey @EdwardPrentice , thanks for chiming in 😉 

> by removing this API how are we expecting clients/developers to get the balance of an arbitrary address?

At the present stage you cannot, as that's not a use case contemplated by our wallet and we prefer to ship an API which has the bare minimum set of operations we can support efficiently and add anything extra later, rather than shipping something we might need to deprecate very soon. It's a practicality, really 😉 

> Surely not both? How can this operation be inaccurate?

It's inaccurate because the balance associated to an address doesn't reflect any meaningful information about the state of the world. It merely says that on that particular address `A` there are some unspent ADA , but this doesn't help you answer any of the more practical questions (for example: "As a user, which is my balance?")

> Again I would be asking - what mechanism is there for a client/developer to accurately view the balance for a given address?

If I may ask, which is the practical use case for somebody willing to do that? Usually what a wallet's integrator/developer might want is to get the [balance of an entire wallet](https://cardanodocs.com/technical/wallet/api/v1/#tag/Wallets%2Fpaths%2F~1api~1v1~1wallets~1%7BwalletId%7D%2Fget) or the [balance associated to an account](https://cardanodocs.com/technical/wallet/api/v1/#tag/Accounts%2Fpaths%2F~1api~1v1~1wallets~1%7BwalletId%7D~1accounts~1%7BaccountId%7D%2Fget) of this wallet.

Why you would want to drill even deeper? Generally speaking, addresses are "ephemeral" resources -- you can spend ADA from them at any time, and the input selection process is not in control of the end user. Wallets and Accounts are instead much more "stable", and as such it makes sense to talk about balance for them.

I hope this helps clarifying! 😉

EdwardPrentice:
Hi @adinapoli-iohk thanks for your reply.

I'm learning fast. Would I be right in saying that the lowest level primitive Cardano Daedalus deals with is a Wallet? The fact that a wallet contains multiple addresses is fine and great but Daedalus and the backend APIs don't support addresses for, eg. querying an address's balance.

I think I understand that, and it seems like a somewhat reasonable decision. Are you concerned you are designing out certain ways of using Cardano by not treating addresses as first class concepts?

By not being able to generate/query/use a key pair ("address", "key") independently of a wallet, can something like [Brainblocks](https://brainblocks.io/) exist for Cardano?

Not being able to generate addresses without a wallet also means I can't generate one offline, I think. That to me feels like a problem as now I can't safely generate a paper wallet or paper address offline.

I was looking at building/experimenting with a view-only wallet, in which a user loaded their N addresses and could monitor their balance trivially, on a potentially-not-totally-secure device, like a phone - with a sync'd node remotely providing the backend APIs. On top of the same API one could also build IFTTT type integrations and plenty of other things. How do I learn of the balance of a bunch of addresses I don't own?

I'm learning still - so thanks for any thoughts you have. If there's somewhere else to discuss this just let me know. Thanks a lot, sorry it was a bit of a rambly mish mash of thoughts.

PR #2821, opened 2018-04-25
[CSL 2209] - Server side tls check

KtorZ:
## Description

Re-opening #2683 here. To summarize the discussions on this issues and the latest changes:

- This PR enables client certificate verification by the server. Which mean checking the followings:
    - A received client certificate is well-formed and not expired
    - The certificate has been signed by a known CA (note that the CA chain the server knows is provided to it upon starting via CLI)
    - The certificate is meant for client authentication (cf Key Usage Purpose)
    - +other default checks provided by the X509 library.

- @vincenthz & I discussed on slack and concluded that this should work but is perhaps overkill as an authentication scheme. We don't really leverage any of the identity-related and dynamic features that come along client-side certificates. 

- The documentation is now hosted by a _"different server"_ on a different port (8091). The check happens at the server level and there's no way to whitelist some endpoints to bypass the check (not mentioning the security risks). Since the documentation is meant to be accessed via a browser (where client certificate authentication becomes quite painful), I've simply decoupled the doc server from the api server and mount it elsewhere.

- I am considering adding a couple of integration tests solely focused on testing the client authentication in #2802. This PR already makes sure that the happy-flow works but it'd be nice to ensure that the certificate is necessary to contact the API and works as expected. @vincenthz any inputs on the best approach to thoroughly test that kind of things is highly appreciated.

### Related PRs

- #2822
- #2802

## Linked issue

[CSL-2209](https://iohk.myjetbrains.com/youtrack/issue/CSL-2209)

## Type of change

## Developer checklist

## Testing checklist


lib/src/Pos/Web/Server.hs
```diff
@@ -84,24 +91,87 @@ application = do
     server <- servantServer
     return $ serve nodeApi server
 
-serveImpl
-    :: (HasConfiguration, MonadIO m)
-    => m Application
+type TLSServer m =
+    m Application
```
parsonsmatt: I usually find these kinds of type synonyms to provide more indirection than readability -- I almost always need to jump-to-definition to find out *what* a `TLSServer` is. Could you write a comment explaining why you use this alias?

lib/src/Pos/Web/Server.hs
```diff
@@ -84,24 +91,87 @@ application = do
...
     -> String
     -> Word16
     -> Maybe TlsParams
     -> Maybe Settings
     -> m ()
-serveImpl app host port mWalletTLSParams mSettings =
+
+type GetTLSSettings =
+    String -> Word16 -> TlsParams -> TLSSettings
```
parsonsmatt: Kinda the same comment here -- Seeing `String -> Word16 -> TlsParams -> TLSSettings` in a signature immediately lets me know that it's a function expecting these things. The alias hides that.

lib/src/Pos/Web/Server.hs
KtorZ: Not sure where this comes from. Might have picked up somewhere or from before and it stayed as is. 
I do agree with you though. Will change :+1:

lib/src/Pos/Web/Server.hs
KtorZ: Yeah, this was _simply_ to avoid repeating this bit of signature twice in both functions. I tend to think on the moment that alias makes it more readable only to regret it later :/ ...

parsonsmatt: They do make it more readable, but less understandable. Unfortunately the two aren't directly correlated :(

lib/src/Pos/Web/Server.hs
```diff
@@ -84,24 +91,87 @@ application = do
...
+    tlsSettings =
+        tlsSettingsChain tpCertPath [tpCaPath] tpKeyPath
+
+    serviceID =
+        (host, BSC.pack (show port))
+
+    certificateUsageFromValidations =
+        maybe CertificateUsageAccept (CertificateUsageReject . CertificateRejectOther)
+
+    -- By default, X509.Validation validates the certificate names against the host
```
akegalj: :+1: for extra comments which clarifies things nicely.

PR #2822, opened 2018-04-25
[CSL-2406] - X509 Gen Utility

KtorZ:
## Description

(Re-Opening of #2742)

Haskell utility to generate X.509 certificates for TLS authentication.

This is a pure Haskell implementation (no openssl) to be used during installation to generate certificates now required for authenticating Daedalus with Cardano-SL (see CBR-82). A couple of things are configurable on certificates, such as the Subject Alternative Names for the server or the expiration date (default to 10 years for CA and 1 for other certs).

By default, the utility doesn't output the CA private key (which is therefore lost after having signed all certificates).

### Related PRs

- #2821 
- #2802

## Linked issue

[[CSL-2406]](https://iohk.myjetbrains.com/youtrack/issue/CSL-2406)

## Type of change

## Developer checklist

## Testing checklist


scripts/launch/connect-to-cluster/default.nix
```diff
@@ -73,18 +73,21 @@ in pkgs.writeScript "${executable}-connect-to-${environment}" ''
   export LC_ALL=en_GB.UTF-8
   export LANG=en_GB.UTF-8
   if [ ! -d ${stateDir}/tls ]; then
-    mkdir ${stateDir}/tls/
-    ${pkgs.openssl}/bin/openssl req -x509 -newkey rsa:2048 -keyout ${stateDir}/tls/server.key -out ${stateDir}/tls/server.cert -days 3650 -nodes -subj "/CN=localhost"
+    mkdir -p ${stateDir}/tls/server && mkdir -p ${stateDir}/tls/client
+    ${iohkPkgs.cardano-sl-tools.cardano-x509-certificates}       \
```
KtorZ: Not really sure about this line. @rvl Perhaps you can help on that one and how to wire that with nix.
What I am trying to do is to use the utility just created in `tools`.

scripts/launch/connect-to-cluster/default.nix
```diff
@@ -73,18 +73,21 @@ in pkgs.writeScript "${executable}-connect-to-${environment}" ''
...
     --configuration-key ${environments.${environment}.confKey}     \
-    ${ ifWallet "--tlscert ${stateDir}/tls/server.cert"}           \
-    ${ ifWallet "--tlskey ${stateDir}/tls/server.key"}             \
-    ${ ifWallet "--tlsca ${stateDir}/tls/server.cert"}             \
+    ${ ifWallet "--tlscert ${stateDir}/tls/server/server.cert"}    \
+    ${ ifWallet "--tlskey ${stateDir}/tls/server/server.key"}      \
+    ${ ifWallet "--tlsca ${stateDir}/tls/server/ca.cert"}          \
     --log-config ${configFiles}/log-config-connect-to-cluster.yaml \
     --topology "${configFiles}/topology.yaml"                      \
     --logs-prefix "${stateDir}/logs"                               \
```
KtorZ: Not sure about this whole file. How to actually test this :s ? 
Forgive my nix-noobness, I'll change I promise.

tools/src/gencerts/Configuration.hs
```diff
@@ -0,0 +1,245 @@
...
+--         ca: ...
+--         server: ...
+--         clients: ...
+--
+-- where the 'configuration-key' represents the target environment (dev, test,
+-- bench, etc.).
+decodeEitherConfigFile
+    :: ConfigurationKey -- ^ Target configuration Key
+    -> FilePath         -- ^ Target configuration file
+    -> ExceptT String IO TLSConfiguration
```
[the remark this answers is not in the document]
KtorZ: Yeah, the function isn't meant to be "safe" anyway, it used to return a proper error only for throwing it a bit later; it's not a library but just extracted to make the whole tool's sources more readable. 
In the end, a `MonadFail` would convey this a bit better :+1:

KtorZ:
@parsonsmatt 

- Ditch `ExceptT String IO` in favor of `MonadFail `
- Use `NonEmpty.List` for `serverAltNames`
- Define `newtype` for `ConfigurationKey`. Note that it's only a proxy to a string so far as we still want to allow any free string in there.
- Review some let / where declarations

scripts/launch/connect-to-cluster/default.nix
disassembler: This was mentioned in slack, but for anyone else wanting to test this:
`nix-build -A connectScripts.mainnetWallet -o "./launch_mainnet"`
`nix-build -A connectScripts.stagingWallet -o "./launch_staging"`

These wallets can be customized further by copying `sample-wallet-config.nix` to `custom-wallet-config.nix`, altering and rebuilding.

scripts/launch/connect-to-cluster/default.nix
```diff
@@ -73,18 +73,21 @@ in pkgs.writeScript "${executable}-connect-to-${environment}" ''
   export LC_ALL=en_GB.UTF-8
   export LANG=en_GB.UTF-8
   if [ ! -d ${stateDir}/tls ]; then
-    mkdir ${stateDir}/tls/
-    ${pkgs.openssl}/bin/openssl req -x509 -newkey rsa:2048 -keyout ${stateDir}/tls/server.key -out ${stateDir}/tls/server.cert -days 3650 -nodes -subj "/CN=localhost"
+    mkdir -p ${stateDir}/tls/server && mkdir -p ${stateDir}/tls/client
+    ${iohkPkgs.cardano-sl-tools}/bin/cardano-x509-certificates   \
+      --server-out-dir ${stateDir}/tls/server                    \
+      --client-out-dir ${stateDir}/tls/client                    \
```
disassembler: got an error, this should be `clients-out-dir`.

rvl: Change this to `--clients-out-dir` and add `-e` to line 61 (`#!${pkgs.stdenv.shell} -e`).

KtorZ: Yes, already fixed on my local branch indeed :+1:

KtorZ:
@disassembler 
> cardano-x509-certificates: user error (Error in $: Invalid TLS Configuration: property 'tls' not found in configuration file.)

There's now an extra configuration part related to the generation of these certificates (as we discussed initially). I've added some defaults in `lib/configuration.yaml` 
https://github.com/input-output-hk/cardano-sl/pull/2822/files#diff-61feea0f7a7bc722b74eedc3907be256R114

Not sure however where the different builds pull their configs from?


@rvl 
> The generator program seems ok but please remind me what was wrong with using openssl?
The main issues were about ease-of-use and cross-platform compatibility. `openssl` on windows works a bit differently than on mac & linux, which requires to maintain two different scripts to build a PKI and generate the associated keys. 

On the other hand, we wanted to move that logic from Daedalus (see https://github.com/input-output-hk/daedalus/pull/893) to cardano-sl and have it as a tool that would help us to setup a PKI for different scenarios (multiple clients, renewable certificates etc..). Hence the pure Haskell implementation here.

KtorZ:
Update:

I've slightly change the output to be in a `pem` encoding format rather than raw `der` which isn't very user friendly. Plus, `.pem` is the default encoding for most tools using TLS certificates. 
The freshly generated certificates are currently replacing those we have for development and have been successfully tested to encrypt their first requests :D

KtorZ:
Added extra bit of documentation and updated the existing ones :)

sample-wallet-config.nix
```diff
@@ -17,6 +17,11 @@
   ## double quotes.
   #stateDir = "./state-wallet-mainnet";
 
+  ## Preserve CA Key generated for wallet
+  ## Useful for exchanges that want to issue individual client certificates
+  ## instead of sharing one.
+  #keepCA = false;
```
KtorZ: :ok_hand:

scripts/tls-files/client.pem
```diff
@@ -0,0 +1,47 @@
+-----BEGIN RSA PRIVATE KEY-----
```
disassembler: Should we really have sample certs/keys in scripts/tls? Not sure why we wouldn't be generating these for tests.

tools/src/gencerts/configuration.yaml.example
```diff
@@ -0,0 +1,21 @@
+dev:
```
disassembler: Should we generate a custom config in connect-to-cluster instead of using `lib/configuration.yaml`? that can have some override parameters for altDNS, clients, expiration, etc?

scripts/tls-files/client.pem
KtorZ: I think we could. The only trouble I have at the moment is that the tool _can only_ be used as an executable. Unless we make it an independent module, we can't define a `library` for it in the cardano-tools `.cabal` file.

tools/src/gencerts/configuration.yaml.example
KtorZ: That would work, the tool doesn't make any assumption about the global shape of the config file so you could have a detached one with only the required fields for TLS.

KtorZ:
Note that I am clarifying some requirements with @dcoutts here before moving on.

parsonsmatt:
@rvl Do you feel satisfied with the justification of pure Haskell or do you have further objections?

rvl:
@parsonsmatt the review :x: was for the command line args which I see is fixed now.
Otherwise it looks good.

The justification for pure Haskell was not completely satisfactory because the real problem was horrible batch files/shell scripts not openssl -- but it doesn't really matter.

PR #2823, opened 2018-04-25
Split deterministic tests into separate modules

adinapoli-iohk:
This patch splits the deterministic tests into separate modules, so that
new tests can be added easily on them.

## Description

I have reached the conclusion that as much as we like having randomised tests, having a nice and ever-growing pool of deterministic tests is the only way to have a _witness_ of testability when it comes to QA testing during each release. Merely saying "for CSL-XXXX we are covering the test bit via fuzzy testing" won't cover it, imho. We should be in the position to say "for CSL-XXXX, we test it by running spec X".

This PR is the stepping stone into that direction.

## Linked issue

No linked issue, this spun via self-initiative (but can create one).

## Type of change

- [X] 🔨 New or improved tests for existing code

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -245,6 +245,10 @@ executable cardano-integration-test
                        Types
                        Functions
                        Error
+                       Util
+                       WalletSpecs
+                       AddressSpecs
+                       TransactionSpecs
 
   build-depends:       base
                      , aeson
--- a/wallet-new/integration/AddressSpecs.hs
+++ b/wallet-new/integration/AddressSpecs.hs
@@ -1,0 +1,48 @@
+{-# LANGUAGE RankNTypes    #-}
+{-# LANGUAGE TupleSections #-}
+
+module AddressSpecs (addressSpecs) where
+
+import           Universum
+
+import           Cardano.Wallet.Client.Http
+import           Control.Lens hiding ((^..), (^?))
+import           Test.Hspec
+
+import           Util
+
+
+addressSpecs :: WalletRef -> WalletClient IO -> Spec
+addressSpecs wRef wc = do
+    describe "Addresses" $ do
+        it "Creating an address makes it available" $ do
+            -- create a wallet
+            Wallet{..} <- sampleWallet wRef wc
+
+            -- create an account
+            accResp <- postAccount wc walId (NewAccount Nothing "hello")
+            acc@Account{..} <- wrData <$> accResp `shouldPrism` _Right
+
+            -- accounts should exist
+            accResp' <- getAccounts wc walId
+            accs <- wrData <$> accResp' `shouldPrism` _Right
+            accs `shouldContain` [acc]
+
+            -- create an address
+            addResp <- postAddress wc (NewAddress Nothing accIndex walId)
+            addr <- wrData <$> addResp `shouldPrism` _Right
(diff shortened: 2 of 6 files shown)


KtorZ:
Since we're at it, what about biting the bullet and also make a split of the `Function.hs` (moving the helpers into the newly create `Util.hs`) and having also separated files for each actions ?

wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -245,6 +245,10 @@ executable cardano-integration-test
                        Types
                        Functions
                        Error
+                       Util
+                       WalletSpecs
+                       AddressSpecs
+                       TransactionSpecs
```
parsonsmatt: Why split it into four modules? I feel like a single module would be fine.

adinapoli-iohk: I respectfully disagree 😉 Now it might look like it doesn't make sense to do so, but if in the future (and that's the goal I have in mind) we will have several dozens of tests, I feel that modifying just a monolith file is not the way to go, imho. 

Even just for the sake of contributing to the specs, different developers can contribute to different specs without merge conflicts every time they open a PR. The "glance" effect is rapidly lost once you start staking 20,30,50 specs in a file.

Said that, I would stick with democratic consensus, so I would pull @erikd as he has a great taste in these kind of things and he knows about testing more than I do anyway 😀

erikd: I don't have any strong opinions either way, but in general I don't like doing `X` now because we might do `Y` later. I usually prefer to do `X` when I *know* that the next thing I am going to do is `Y`.

adinapoli-iohk: > I usually prefer to do X when I know that the next thing I am going to do is Y.

That is what I was going to do next anyway (write more specs), but on the other hand I don't want to rub Matt the wrong way by trying to advocate my solution over his. Let's see what me and Matt can decide when he gets online and we'll take from there 😀 

Cheers Erik!

KtorZ: Side note but:

*.git/hooks/post-checkout*
```bash
#!/bin/sh
.git/hooks/ctags >/dev/null 2>&1 &
```

and 

*.git/hooks/ctags*
```bash
#!/bin/sh
set -e
hasktags --ctags --ignore-close-implementation .
```

to generate and refresh ctags does wonderful things for me with vim and everything going with the `:tags` feature. Not sure what editor you use it helps quite a lot to jump around files in the codebase.

parsonsmatt: I use and recommend `codex` so you get tags for dependencies, too :) but having a git hook for this is a great idea!

parsonsmatt:
I'm going to go ahead and merge this after resolving conflict as I want to finally fix the integration test for fees :)

wallet-new/cardano-sl-wallet-new.cabal
erikd: > We have a 1746 line monolithic API.V1.Types file 

Wow, when was the last time you people coded in C or C++?? 1746 lines is not that big :)

parsonsmatt: ... never? :sweat_smile: haskell 4 lyf

PR #2832, opened 2018-04-26
Fix integration test

parsonsmatt:
## Description

The integration test for fee estimation was broken. I have fixed it.

## Linked issue

CSL-2493, sort of, but we closed it out.

## Type of change

## Developer checklist

## Testing checklist

## QA Steps

Run the integration tests.

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
@@ -12,12 +12,16 @@ import           Data.Aeson
 import           Data.List.NonEmpty (NonEmpty ((:|)))
 import           Generics.SOP.TH (deriveGeneric)
 import qualified Network.HTTP.Types as HTTP
-import qualified Pos.Core as Core
 import           Servant
 import           Test.QuickCheck (Arbitrary (..), oneof)
 
+import qualified Pos.Client.Txp.Util as TxError
+import qualified Pos.Core as Core
+import qualified Pos.Crypto.Hashing as Crypto
+import qualified Pos.Data.Attributes as Core
+
 import           Cardano.Wallet.API.V1.Generic (gparseJsend, gtoJsend)
-import           Cardano.Wallet.API.V1.Types (SyncPercentage, SyncProgress (..),
+import           Cardano.Wallet.API.V1.Types (SyncPercentage, SyncProgress (..), V1 (..),
                                               mkEstimatedCompletionTime, mkSyncPercentage,
                                               mkSyncThroughput)
 
@@ -54,7 +58,7 @@ import           Cardano.Wallet.API.V1.Types (SyncPercentage, SyncProgress (..),
 -- TODO: change fields' types to actual Cardano core types, like `Coin` and `Address`
 data WalletError =
       NotEnoughMoney { weNeedMore :: !Int }
-    | OutputIsRedeem { weAddress :: !Text }
+    | OutputIsRedeem { weAddress :: !(V1 Core.Address) }
     | MigrationFailed { weDescription :: !Text }
     | JSONValidationFailed { weValidationError :: !Text }
     | UnknownError { weMsg :: !Text }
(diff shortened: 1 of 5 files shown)


wallet-new/src/Cardano/Wallet/Client/Http.hs
```diff
@@ -74,7 +75,14 @@ mkHttpClient baseUrl manager = WalletClient
...
     unNoContent = map void
     clientEnv = ClientEnv manager baseUrl
-    run       = fmap (over _Left ClientHttpError) . (`runClientM` clientEnv)
+    parseJsendError servantErr =
+        case servantErr of
+            FailureResponse resp ->
+                case decode (responseBody resp) of
+                    Just err -> ClientWalletError err
+                    Nothing  -> ClientHttpError servantErr
+            _ -> ClientHttpError servantErr
```
parsonsmatt: I wasn't converting the errors client-side! Oops!

parsonsmatt:
Feel free to merge if any of you approve it :smiley:

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -54,7 +58,7 @@ import           Cardano.Wallet.API.V1.Types (SyncPercentage, SyncProgress (..),
 -- TODO: change fields' types to actual Cardano core types, like `Coin` and `Address`
 data WalletError =
       NotEnoughMoney { weNeedMore :: !Int }
-    | OutputIsRedeem { weAddress :: !Text }
+    | OutputIsRedeem { weAddress :: !(V1 Core.Address) }
```
erikd: Sum types! Now with partial funtions!! :imp: 

Ok if I point Jordan at that data type?

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
```diff
@@ -73,6 +80,23 @@ data WalletError =
     -- node is still syncing with the blockchain.
     deriving (Show, Eq)
 
+convertTxError :: TxError.TxError -> WalletError
+convertTxError err = case err of
```
erikd: This error conversion function seems like a bit of a smell to me.

Why not have a constructor in `WalletError` that takes a value of type `TxError.TxError` to save converting?

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
parsonsmatt: yes please! do you mean to factor it out into individual exception types? that would be great

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
parsonsmatt: I have a feeling that a single flat type might be easier to deal with than nested types. 

Really want an open variant though :(

erikd: A single flat type makes sense when the errors are in a single package, but nesting errors like I suggest is necessary when dealing with errors outside the current package.

The problem with open variants is that there is no way for the compiler to check for pattern match completeness. The fact that sum types are closed is actually a nice property.

wallet-new/src/Cardano/Wallet/API/V1/Errors.hs
erikd: No the idea was just to remove all the field accessors (like `weAddress` in the constructor about) because all these field accessors are partial functions.

These are particularly odious in error type, because they are being used in the error handling path which is less likely to get exercised during regular execution or testing. What happens when you get an error that you'd like to log, but the error handling code gets a pattern match failure? You've just lost your logging.

adinapoli-iohk: @erikd It would be lovely if you guys could take a look at that type, but there are a bunch of pitfalls you folks should be aware of, as that type was created like that on purpose, as crazy as it might sound. Let me explain, hehe.

A `WalletError` is much more than a way to report an error in our backend, is also a particular structure used to expose nice error messages to API consumers. In particular, we are using the [Jsend](https://labs.omniti.com/labs/jsend) format to expose errors to the clients. The pattern we settled upon with the frontend folks was to expose a tag (essentially the type constructor converted to a `String`) in the `message` field, and _extra_ diagnostics in the `diagnostic` field. You can look at how [this renders](https://cardanodocs.com/technical/wallet/api/v1/#section/Errors) in the relevant section of our API documentation. A practical example is the following: you want to express the output is not a redeem address (like on like 61). What we do is to show the frontend the following:

``` json
{
   "status":"error",
   "diagnostic":{
      "address":"b10b24203f1f0cadffcfd16277125cf7f3ad598983bef9123be80d93"
   },
   "message":"OutputIsRedeem"
}
```

See what we did there? We converted (via `generic-sop`) our type so that the `weAddress` is spliced back into the `address` field of our final JSON, and that's why we alas require partial functions.

By all means I'm very open to other solutions, provided we can maintain the same error semantic and the same output structure. I hope it clarifies the rationale 😉

PR #2847, opened 2018-04-30
remove purscript from the default shell

cleverca22:


avieth:
Cherry-picked to the base branch. Thanks!

PR #2858, opened 2018-05-02
[CSL-2328] part 3: full diffusion layer without NetworkConfig/Topolog…

avieth:
## Description

Another step in the incremental merge of oustanding networking work from `networking/develop` into `develop`.

**NB: there are many whitespace-only changes. You can add "?w=1" to this URL to not show them, in which case we get +302 −217 changes**

- Interface for the *full* diffusion layer is changed so that it's more useful for other applications, such as testing, simulation, and benchmarking. For instance, you don't have to give a `NetworkConfig` anymore, which was an onerous requirement.
- There's a blockchain LCA suffix function which is used by the full diffusion layer. That's to say, LCA classification was formerly part of the diffusion layer, but is now part of the logic layer, which makes more sense.
- The type of `requestTip` is changed to be simpler: no callback, instead a map from peer identifiers to an action which gives their tip header.
- Many types simplified by removing the `kademlia` type variable.

## Type of change

## Developer checklist

## Testing checklist


avieth:
A wallet-new integration test is failing. Unfortunately it seems really hard to run it locally. It shows me the seed used, but this is not a typical HSpec test: it seems it requires running a separate server (how exactly? I don't know) and also does not allow the caller to specify the random seed.

wallet-new/integration/Error.hs
```diff
@@ -60,7 +62,8 @@ showConstr = \case
     InvalidTransactionState {} -> "InvalidTransactionState"
     InvalidTransactionFee {} -> "InvalidTransactionFee"
     UnexpectedChangeAddress {} -> "UnexpectedChangeAddress"
-    UnexpectedAccountBalance {} -> "UnexpectedAccountBalance"
+    UnexpectedAccountBalance explanation _ _ ->
+        T.unpack ("UnexpectedAccountBalance " <> explanation)
```
parsonsmatt: This function should not be showing the explanation -- if we need to render that info then it should be in a different function call imo

avieth: The issue here is that `UnexpectedAccountBalance` is used to describe 2 different unexpected cases, so that which one failed cannot be determined by looking at the test logs.

I'd be happy to remove this change, since it's not strictly related to the rest of the patch. I put it in so that I could see, in the failing test case that apparently only sometimes happens in wallet-new integration tests (seems the actual tests which are run are also chosen randomly), which unexpected balance happened: was the payer not deducted the transfer amount, or was the payee not granted the transfer amount?

parsonsmatt:
To run the integration tests, I run:

```
$ stack build --fast && ./scripts/test/wallet/integration.sh 20
```

There's a nix command to run them as well, though I don't recall it right now

parsonsmatt:
@avieth  Yeah, normalizing the integration test to use hspec is something I want to do. It's a little tricky right now because the suite is non-deterministic and we use `liftIO generate` a lot, which IIRC doesn't accept a seed :\

PR #2860, opened 2018-05-02
[CSL-2331] use more IO, particularly in diffusion

avieth:
## Description

Yet another piece of `networking/develop` brought into `develop`. Here, the networking stuff is rid of `Mockable`. It's all monomorphic in `IO` now.

But the logic layer stuff is still polymorphic, using mockable and other attempted generalizations like `MonadMask` and `MonadBaseControl`. The two styles are reconciled by hoisting. This isn't ideal, but it's also not a problem in practice:

```Haskell
-- It's always some reader with IO base.
lift :: forall t . IO t -> BigComplexRealMode t
lift = liftIO

-- The reader antecedent is always available when we create and run the diffusion layer.
lower :: forall t . BigComplexRealMode t -> IO t
```

## Type of change

## Developer checklist

## Testing checklist


avieth:
@erikd @CodiePP First step to removing mockable is in here.

erikd:
Wow, what a monster PR. Will been so go over this very carefully.

networking/src/Node/Internal.hs
```diff
@@ -64,22 +70,39 @@ import           Data.NonEmptySet (NonEmptySet)
...
+logInfo = Wlog.usingLoggerName "node" . Wlog.logInfo
+
+logDebug :: Text -> IO ()
+logDebug = Wlog.usingLoggerName "node" . Wlog.logDebug
+
+logWarning :: Text -> IO ()
+logWarning = Wlog.usingLoggerName "node" . Wlog.logWarning
+
+logError :: Text -> IO ()
+logError = Wlog.usingLoggerName "node" . Wlog.logError
```
avieth: Oops, this was done before I decided on using what's in `Pos.Util.Trace`. I can knock these out and use a `Trace` instead.

erikd: Yes, please do :)

infra/Pos/DHT/Real/Real.hs
```diff
@@ -8,62 +8,38 @@ module Pos.DHT.Real.Real
        , kademliaGetKnownPeers
        , startDHTInstance
        , stopDHTInstance
-         -- * Exported to avoid compiler warnings
-         -- TODO: If we really don't need these functions, remove
-       , foreverRejoinNetwork
-       , rejoinNetwork
-       , withKademliaLogger
```
erikd: :100:

infra/Pos/DHT/Real/Real.hs
```diff
@@ -72,77 +48,51 @@ stopDHTInstance KademliaDHTInstance {..} = liftIO $ K.close kdiHandle
...
     kdiKnownPeersCache <- atomically $ newTVar []
     pure $ KademliaDHTInstance {..}
   where
     catchErrorsHandler e = do
-        logError $ sformat ("Error launching kademlia with options: "%shown%": "%shown) kconf e
-        throwM e
-    catchErrors x = liftIO x `catchAny` catchErrorsHandler
+        traceWith logTrace (Error, sformat ("Error launching kademlia with options: "%shown%": "%shown) kconf e)
+        throwIO e
+    catchErrors x = x `catchAny` catchErrorsHandler
```
erikd: Is there a way to refactor that so that its more obvious that its catching `SomeException`, logging it and then re-throwing? Maybe by combining `catchErrors` and `catchErrorsHandler` into a single function?

Functions that catch `SomeException` always make me a bit nervous and I take a much closer look at the code.

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -3,19 +3,20 @@ module Pos.Diffusion.Subscription.Dns
     ( dnsSubscriptionWorker
     ) where
 
-import           Universum
+-- Why put MVars in a prelude? Weird.
```
erikd: Yes, there are a lot of "interesting" decisions made in `Universum`.

infra/Pos/Diffusion/Transport/TCP.hs
```diff
@@ -5,45 +5,42 @@ module Pos.Diffusion.Transport.TCP
...
-       , MonadIO n
-       , MonadThrow m
-       , MonadMask m
-       , WithLogger m
-       )
-    => Microsecond
+    :: Trace IO Text
+    -> Microsecond
     -> TCP.TCPAddr
-    -> (Transport n -> m a)
```
erikd: Wow, explicit parameter passing. Is this a new idea? /sarcasm :)

infra/Pos/DHT/Real/Real.hs
avieth: I'll just replace it with `onException`

avieth: Oh right, `onException` doesn't get the exception... I'll figure something out.

avieth:
I have pleased the gods. Thanks for the review @erikd

PR #2863, opened 2018-05-03
Rework storage layer to store blocks and undos in the same file.

nc6:
This is an attempt at "simplest possible change" which moves us in the
direction of using fewer files for the block storage. However, this also
tries to tidy up some of the way 'SerializedBlunds' were handled. It avoids
cases where we end up doing serialize/deserialize loops, and tries to remove
a bunch of unneeded functions.

I still don't like this way of handling things - in particular, the contract
on _how_ exactly a `Blund` is stored is completely implicit, yet not isolated
to this module.

## Type of change

## Developer checklist

## Testing checklist

## QA Steps
Run against an existing node DB and verify that existing blocks load correctly.

### DIFF
--- a/block/src/Pos/Block/Types.hs
+++ b/block/src/Pos/Block/Types.hs
@@ -4,7 +4,6 @@ module Pos.Block.Types
        ( SlogUndo (..)
        , Undo (..)
        , Blund
-       , SerializedBlund
 
        , LastKnownHeader
        , LastKnownHeaderTag
@@ -27,7 +26,6 @@ import           Pos.Communication.Protocol (NodeId)
 import           Pos.Core (HasConfiguration, HasDifficulty (..), HasHeaderHash (..))
 import           Pos.Core.Block (Block, BlockHeader)
 import           Pos.Core.Txp (TxpUndo)
-import           Pos.DB.Class (SerializedUndo)
 import           Pos.Delegation.Types (DlgUndo)
 import           Pos.Update.Poll.Types (USUndo)
 import           Pos.Util.Util (HasLens (..))
@@ -44,8 +42,6 @@ instance NFData Undo
 
 -- | Block and its Undo.
 type Blund = (Block, Undo)
-
-type SerializedBlund = (Block, SerializedUndo)
 
 instance HasConfiguration => Buildable Undo where
     build Undo{..} =
--- a/block/src/Pos/DB/Block.hs
+++ b/block/src/Pos/DB/Block.hs
@@ -39,23 +39,23 @@ import           Control.Lens (at)
 import qualified Data.ByteString as BS (hPut, readFile)
 import           Data.Default (Default (def))
 import           Formatting (formatToString)
-import           System.Directory (createDirectoryIfMissing, removeFile)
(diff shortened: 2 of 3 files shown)


erikd:
This looks good Nick, but I don't want to merge we get some pretty thorough testing on it. I've asked devops for some help on this.

I suspect that in it current state it, if the software is installed on a clean machine and syncs the whole blockchain, then everything will just work.

What happens on a machine that already has the separate block/undo files?

nc6:
Tested:

Create a new network with this version. Start daedalus. Create a new wallet, send it 500ADA from the bootstrap wallet. Remove the wallet DB, then restart the wallet and restore from backup. Confirm that wallet is successfully restored with 500ADA balance.

Create a new network on develop. Let a number of blocks get processed. Start daedalus. Create a new wallet, send it 500ADA from the bootstrap wallet. Stop this network, copy across the node DB to a clean checkout of this commit. Run the network without rebuilding the DB. Load daedalus and restore the wallet - confirm its balance.

parsonsmatt:
I would love to get all the testing infra for launching nodes etc into Haskell so we can more easily write *that exact integration test.* 

Can you paste  a list of commands you ran in order to do that, so I can reproduce it locally?

nc6:
@parsonsmatt I can give a rough idea, though a bunch of it was done through the actual Daedalus client. It should be possible by poking the API directly, though I haven't done this.

```
git worktree add ../cardano-sl.develop develop
cd ../cardano-sl.develop/
stack build
mkdir run && cp -r ./scripts/tls-files/ ./run/
./scripts/launch/demo-with-wallet-api.sh
```

At this point one can launch the 'develop' version. Import the genesis wallet from `daedalus/features/support/default-wallet.key`. Create a new wallet, transfer some ADA. Make sure it gets confirmed in some blocks.

```
./scripts/launch/kill-demo.sh
cp -R ./run ../cardano-sl/
cd ../cardano-sl
NO_REBUILD=1 ./scripts/launch/demo-with-wallet-api.sh
```

Launch `daedalus` again. It will take a while to connect since it's rebuilding the wallet DB. You should be able to restore the wallet from backup and confirm the balance.

erikd:
@cleverca22 asks (in Slack but copied here):

> 1: we need to somehow clean up all .undo files on startup, to regain the lost space

Yes, forgot about that. Is it worth, having a process that checks for separate block/undo files, compacting them into blund files and deleting the old ones?

> 2: do the old .block files still contain enough data to reconstruct those undos?

No. The solution chosen puts the block and undo in a single file, so we actually keep the undos for now.

> 3: what happens if an older cardano is ever ran on this state?

It would need to download all the blocks again. I can't see a way around this. We simply can't make old software compatible with the new code.

cleverca22:
one thought i had on how to migrate things

any time you read a `.block` and `.undo` pair, re-write them as a `.blund` and then delete the input pair

nc6:
> One option to address this issue would be to write a trivial tool that can be manually run and splits the blund files into separate block and undo files.

We could call it the "unblundler".

erikd:
Closing this as #2905 was built on top of this and has already been merged.

PR #2874, opened 2018-05-04
Remove field accessors from `KeygenCommand` data type

Jimbo4350:
## Description

Remove field accessors from `KeygenCommand` data type

## Linked issue

[https://iohk.myjetbrains.com/youtrack/issue/CDEC-284](https://iohk.myjetbrains.com/youtrack/issue/CDEC-284)

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/tools/src/keygen/KeygenOptions.hs
+++ b/tools/src/keygen/KeygenOptions.hs
@@ -36,8 +36,7 @@ data KeygenCommand
     | ReadKey FilePath
     | DumpAvvmSeeds DumpAvvmSeedsOptions
     | GenerateKeysBySpec GenKeysOptions
-    | DumpGenesisData { dgdPath      :: !FilePath
-                      , dgdCanonical :: !Bool }
+    | DumpGenesisData !FilePath !Bool
     deriving (Show)
 
 data DumpAvvmSeedsOptions = DumpAvvmSeedsOptions
@@ -100,7 +99,7 @@ keygenCommandParser =
         dgdCanonical <- switch $
             long "canonical" <>
             help "Whether genesis data should be in canonical json"
-        pure DumpGenesisData {..}
+        pure $ DumpGenesisData dgdPath dgdCanonical
 
 dumpAvvmSeedsParser :: Parser DumpAvvmSeedsOptions
 dumpAvvmSeedsParser = do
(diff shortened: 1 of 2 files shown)


tools/src/keygen/KeygenOptions.hs
```diff
@@ -29,15 +31,18 @@ data KeygenOptions = KeygenOptions
     , koConfigurationOptions :: ConfigurationOptions
     } deriving (Show)
 
+newtype DgdPath      = DgdPath FilePath deriving (Show)
+
+newtype DgdCanonical = DgdCanonical Bool deriving (Show)
```
parsonsmatt: For `newtype` over `Bool`, I typically like to just make a one-off type like

```
data DgdCanonical = DgdNonCanonical | DgdCanonical
```

Though, you can get similarly readable with `newtype DbgCanonical = DgdIsCanonical Bool`, as `DgcIsCanonical True` seems fairly clear to me

Jimbo4350: Thanks for pointing this out.

tools/src/keygen/KeygenOptions.hs
```diff
@@ -29,15 +31,18 @@ data KeygenOptions = KeygenOptions
...
 data KeygenCommand
     = RearrangeMask FilePath
     | GenerateKey FilePath
     | GenerateVss FilePath
     | ReadKey FilePath
     | DumpAvvmSeeds DumpAvvmSeedsOptions
     | GenerateKeysBySpec GenKeysOptions
-    | DumpGenesisData { dgdPath      :: !FilePath
-                      , dgdCanonical :: !Bool }
+    | DumpGenesisData DgdPath DgdCanonical
```
erikd: I see that you have `newtype`ed these. In this case I actually would not bother. Newtypes are useful when you have two constructor fields of the same type. like:
```
    | Command Bool Bool!
```
which I would usually `newtype` as:
```
newtype OpenDoors = OpenDoors Bool
newtype LaunchMissles = LauchMissiles Bool

data X
     = Command OpenDoors LauchMissiles
```
This means you reduces the chances of getting the two `Bool` fields around the wrong way.

So in the example above, I would be perfectly happy with:
```
    | DumpGenesisData !FilePath !Bool
```
because the compiler will not *let* anyone get it wrong.

parsonsmatt: Yes, that's true :) The primary advantage of a `newtype` here is that you can provide some additional meaning to the type, track provenance, not use it incorrectly, etc. For this, it is likely of small benefit.

PR #2886, opened 2018-05-08
[CDEC-286] Remove field accessors from 'FakeUtxoCoinDistribution' data type

intricate:
## Description

Remove field accessors from the `FakeUtxoCoinDistribution` data type.

## Linked issue

[https://iohk.myjetbrains.com/youtrack/issue/CDEC-286](https://iohk.myjetbrains.com/youtrack/issue/CDEC-286)
## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/tools/src/dbgen/Lib.hs
+++ b/tools/src/dbgen/Lib.hs
@@ -59,7 +59,7 @@ _exampleSpec = GenSpec
     { walletSpec = WalletSpec
         { accounts = 1
         , accountSpec = AccountSpec { addresses = 100 }
-        , fakeUtxoCoinDistr = RangeDistribution { amount = 1000, range = 100 }
+        , fakeUtxoCoinDistr = RangeDistribution 100 1000
         , fakeTxsHistory = SimpleTxsHistory { txsCount = 100, numOutgoingAddress = 3 }
         }
     , wallets = 1
@@ -97,6 +97,9 @@ data AccountSpec = AccountSpec
 instance FromJSON AccountSpec
 instance ToJSON AccountSpec
 
+type AddressRange = Integer
+type DistributionAmount = Integer
+
 -- TODO(ks): The question here is whether we need to support some other
 -- strategies for distributing money, like maybe using a fixed amount
 -- of `toAddress` to cap the distribution - we have examples where we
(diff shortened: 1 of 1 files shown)


tools/src/dbgen/Lib.hs
```diff
@@ -335,8 +337,8 @@ generateRealTxHistE outputAddresses = do
 
 
 generateFakeUtxo :: FakeUtxoCoinDistribution -> AccountId -> UberMonad ()
-generateFakeUtxo NoDistribution _          = pure ()
-generateFakeUtxo RangeDistribution{..} aId = do
```
erikd: :100:

tools/src/dbgen/Lib.hs
```diff
@@ -97,6 +97,9 @@ data AccountSpec = AccountSpec
 instance FromJSON AccountSpec
 instance ToJSON AccountSpec
 
+type AddressRange = Integer
+type DistributionAmount = Integer
```
parsonsmatt: I'd be happier to see a `newtype` here, as they're used in `RangeDistribution` side-by-side. It would be easy to forget which was which.

erikd: Good catch @parsonsmatt ! Missed that!

intricate: @parsonsmatt: I've made these changes. When you get a chance, let me know what you think.

tools/src/dbgen/Lib.hs
```diff
@@ -97,8 +99,19 @@ data AccountSpec = AccountSpec
...
+
+instance FromJSON AddressRange
+instance ToJSON AddressRange
+
+newtype DistributionAmount = DistributionAmount 
+    { unDistributionAmount :: Integer 
+    } deriving (Show, Eq, Generic)
+    
+instance FromJSON DistributionAmount
+instance ToJSON DistributionAmount
```
parsonsmatt: okay, so this change indicates that we are serializing/deserializing these fields! That's very good to know. These changes are going to break the serialization format.

We should probably go back and ensure that if we're serializing/desererializing using `Generic`s, the `ToJSON` and `FromJSON` instances produce the same output as before.

For this instance, we'll want to write a test that has the result of calling `Data.Aeson.encode` on an old value of these types (that has the field accessors). Then we must ensure that the *new* serialization matches that exactly for a similar value.

One way we might do that is replicating the data type in the test suite exactly as it used to be, and then creating a function `delabel :: FakeUtxoCoinDistribution -> Real.FakeUtxoCoinDistribution` that converts. Then we write a QuickCheck property that asserts that `forAll fakeUtxoCoinDistribution $ \x -> encode (delabel x) === encode x`.

erikd:
This is another one where this change may affect the serialisation format. This is in the tools package so the consequences of this breaking are possibly not too bad, but better to avoid merging this until we're sure.

Closing this until we have golden tests in place to make sure we're not breaking anything.

PR #2890, opened 2018-05-08
[PB-49] Develop new generator which can continuously create transactions between addresses.

365andreas:
## Description

This PR will create a new generator for auxx which will be able to continuously create transactions between addresses. Previously we could not exceed 50k transactions because that was the number
of addresses in genesis block.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/PB-49

## Type of change

## Developer checklist

## Testing checklist

## QA Steps

Run a cluster on staging and request more than 50k transactions.


scripts/common-functions.sh
```diff
@@ -281,7 +281,7 @@ function bench_cmd {
   echo -n " --peer 127.0.0.1:"`get_port $((i-1))`
   echo -n " $(logs "" node_auxx.log)"
   echo -n " --system-start $system_start"
-  echo -n " cmd --commands \"send-to-all-genesis $time $conc $delay $sendmode ./tps-sent.csv\""
+  echo -n " cmd --commands \"send-to-all-genesis $time $time $conc $delay $sendmode ./tps-sent.csv\""
```
CodiePP: is `$time` duplicated?

CodiePP: ah, I see: the first is "genesis" txs, the second the subsequent txs. 
Then, `$sendmode` is obsolete, I think. It's not part of the structure anymore.

auxx/src/Command/Tx.hs
```diff
@@ -92,28 +95,58 @@ sendToAllGenesis diffusion (SendToAllGenesisParams txsPerThread conc delay_ tpsS
...
+            -- construct transactions whose inputs does not belong in genesis
+            -- block. Send as many coins as were received in the previous round
+            -- back to yourself.
+            prepareTxs :: Int -> m ()
+            prepareTxs n
+                | n <= 0 = return ()
+                | otherwise = (atomically $ tryReadTQueue txPreparationQueue) >>= \case
+                    Just (tx, txOut1', senderKey) -> do
+                        let txInp = TxInUtxo (hash (taTx tx)) 0
+                            utxo' = M.fromList [(txInp, TxOutAux txOut1')]
```
kantp: So, what you're doing here is _expecting_ that the output from the previously sent transaction is included in the UTxO by the time this transaction will actually be sent, right?
Might be worth to add a comment explaining that.

client/src/Pos/Client/Txp/Network.hs
```diff
@@ -5,7 +5,6 @@
 
 module Pos.Client.Txp.Network
        ( TxMode
-       , submitTx
```
kantp: Was this just dead code?

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -557,7 +557,7 @@ createTx
     -> m (Either TxError TxWithSpendings)
 createTx pendingTx utxo ss outputs addrData =
     createGenericTxSingle pendingTx (\i o -> Right $ makePubKeyTx ss i o)
-    OptimizeForSecurity utxo outputs addrData
+    OptimizeForHighThroughput utxo outputs addrData
```
kantp: Is this used only in benchmarking, or also in the wallet?

lib/src/Pos/Launcher/Resource.hs
```diff
@@ -163,16 +170,22 @@ allocateNodeResources np@NodeParams {..} sscnp txpSettings initDB = do
...
-    closeNodeDBs nrDBs
-    releaseNodeContext nrContext
+releaseNodeResources NodeResources {..} =
+    case nrJsonLogConfig of
+        JsonLogDisabled -> return ()
+        JsonLogConfig mVarHandle _ -> do
+            h <- takeMVar mVarHandle
+            (liftIO . hClose) h
+            putMVar mVarHandle h
+            closeNodeDBs nrDBs
```
kantp: This should be outside of the `case`, right? Otherwise we will only close the DB and release the context if JSON logging is active.

networking/src/JsonLog/JsonLogT.hs
```diff
@@ -149,7 +149,7 @@ jsonLogDefault jlc x =
                 `catchAny` \e -> do
                     logWarning $ sformat ("error in deciding whether to json log: "%shown) e
                     return False
-            when b $ liftIO (withMVar v $ flip hPut $ encode event)
+            when b $ liftIO (withMVar v $ \h -> (hPut h (encode event) >> hFlush h))
```
kantp: Wouldn't it be better to set the buffering mode of the handle appropriately, instead of manually flushing after each log entry?

kantp: Speaking of which, it seems the buffering mode is set to `NoBuffering`, which seems a bit extreme (see lib/src/Pos/Launcher/Resource.hs)

auxx/src/Command/Proc.hs
```diff
@@ -217,8 +218,7 @@ createCommandProcs hasAuxxMode printAction mDiffusion = rights . fix $ \commands
         return ValueUnit
     , cpHelp = "create and send transactions from all genesis addresses \
                \ for <duration> seconds, <delay> in ms. <conc> is the \
-               \ number of threads that send transactions concurrently. \
-               \ <mode> is either 'neighbours', 'round-robin', or 'send-random'"
```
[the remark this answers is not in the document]
365andreas: With the new diffusion layer we cannot have this option.

client/src/Pos/Client/Txp/Network.hs
365andreas: Yes it is not used anywhere.

client/src/Pos/Client/Txp/Util.hs
365andreas: It is only used in `auxx`. So it is not used in the wallet.

lib/src/Pos/Launcher/Resource.hs
365andreas: Yes. Fixed.

networking/src/JsonLog/JsonLogT.hs
365andreas: I have seen that `NoBuffering` is set but I added `flush`, which is redundant in `NoBuffering` mode, in the case of somebody in the future changes it forgetting to add it.

lib/src/Pos/Launcher/Resource.hs
```diff
@@ -289,6 +302,14 @@ allocateNodeContext ancd txpSettings ekgStore = do
     logDebug "Created peersVar"
     mm <- initializeMisbehaviorMetrics ekgStore
 
+    logDebug ("Dequeue policy to core:  " <> (show ((ncDequeuePolicy networkConfig) NodeCore)))
+        `catch` \(e :: SomeException) -> logDebug $ show e
+    logDebug ("Dequeue policy to relay: " <> (show ((ncDequeuePolicy networkConfig) NodeRelay)))
+        `catch` \(e :: SomeException) -> logDebug $ show e
+    logDebug ("Dequeue policy to edge: " <> (show ((ncDequeuePolicy networkConfig) NodeEdge)))
+        `catch` \(e :: SomeException) -> logDebug $ show e
```
365andreas: Please review the 6 lines above also.

auxx/src/Command/Proc.hs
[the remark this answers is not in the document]
kantp: Ah, I see!

client/src/Pos/Client/Txp/Network.hs
kantp: Ok

client/src/Pos/Client/Txp/Util.hs
kantp: Are you sure? It is used in `submitTx` in client/src/Pos/Client/Txp/Network.hs, and that function is used quite a lot.

networking/src/JsonLog/JsonLogT.hs
kantp: But isn't `NoBuffering` too aggressive? I think it would be preferable to have `LineBuffering`, and use an `hPutLn`, which would also make the file much easier to load up in an editor and have a glance at it.

lib/src/Pos/Launcher/Resource.hs
kantp: Catching `SomeException` without immediately throwing it is generally not a good idea, since it can easily swallow arbitrary exceptions and mess up your error handling. You should either catch some specific exception (or class of exceptions) and handle them properly, or -- if, for example, you just want to log that there was some exception -- immediately rethrow the exception.

Also, I'm not sure why you're catching here in the first place. Are you expecting some exceptions from those loggings?

client/src/Pos/Client/Txp/Util.hs
365andreas: @kantp 
`submitTx` is removed by this PR because it was not used in the code.
If you see any `submitTx` in the `develop` branch it is because it is used also as a parameter of a function with the same name.

lib/src/Pos/Launcher/Resource.hs
365andreas: https://github.com/input-output-hk/cardano-sl/blob/5db86b391231f3eb655bb90fc879a3cfe7f53f2f/infra/Pos/Infra/Network/Policy.hs#L167-L173

Default dequeue policy of core throws an error for edge nodes.

client/src/Pos/Client/Txp/Util.hs
kantp: Ah, I see! Sorry about the confusion.

lib/src/Pos/Launcher/Resource.hs
kantp: Ok, that makes sense. And since there is nothing in the lines you're using `catch` on that could throw another exception, the usage is ok here (well, technically, `logDebug` itself could throw an exception, but you're also using that on in the code catching the exception.

parsonsmatt: Suggestion: Change the `error` call to a `throw` with an exception type (that isn't string). Then we can catch the specific error.

Alternatively, you can call:

```haskell
foo `catch` \(ErrorCall msg) -> logDebug (pack msg)
```

which will only catch the `error` call.

PR #2897, opened 2018-05-09
[CBR-255] TxMeta persistent storage

adinapoli-iohk:
## Description

This PR adds support for storing the transaction metadata (cfr. "Tracking Metadata" section of the Wallet Spec) on disk but outside acid-state, due to the fact such data is unbounded (in size) and as such it won't eventually all fit into main memory.

The concrete storage implementation is completely hidden via an opaque interface (i.e. via an `Handle` pattern) and the persistence of choice was Sqlite, accessed via the excellent `beam` library.

@edsko Feel free to review this already, as I expect lots of massaging to make it mergeable against latest `develop` and in order to please the CI (we are introducing a toolchain dependency on Sqlite, so I think this has all kinds of implications I will need to check with devops).

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-255

## Type of change

## Developer checklist

## Testing checklist

## Screenshots (if available)

screen shot 2018-05-09 at 11 50 59


wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,563 @@
...
+    | SortByAmount
+    -- ^ Sort the 'TxMeta' by the amount of money they hold.
+
+newtype OrdByCreationDate = OrdByCreationDate { _ordByCreationDate :: TxMeta } deriving (Show, Eq)
+
+instance Ord OrdByCreationDate where
+    compare a b = compare (_txMetaTableCreatedAt . _ordByCreationDate $ a)
+                          (_txMetaTableCreatedAt . _ordByCreationDate $ b)
+
+-- | TODO(adinapoli) Sorting.
```
edsko: Obsolete comment?

wallet-new/src/Cardano/Wallet/Kernel/DB/TxMeta/Types.hs
```diff
@@ -0,0 +1,173 @@
...
+      -- A transaction is outgoing when it decreases the wallet's balance.
+    , _txMetaIsOutgoing :: Bool
+    }
+
+makeLenses ''TxMeta
+
+-- | Deep equality for two 'TxMeta': two 'TxMeta' are equal if they have
+-- exactly the same data, and inputs & outputs needs to appear in exactly
+-- the same order.
+deepEq :: TxMeta -> TxMeta -> Bool
```
edsko: This naming is confusing. Usually deep versus shallow refers to how much of the substructures we compare. Here we are always comparing everything, we're just a bit more permissive. Maybe change to something like `exactlyEqualTo` and `isomorphicTo` or something like that?

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
...
+    let t1 = blueprint & over txMetaInputs  (const inputs)
+                       . over txMetaOutputs (const outputs)
+    let t2 = blueprint & over txMetaInputs  (const (NonEmpty.reverse inputs))
+                       . over txMetaOutputs (const (NonEmpty.reverse outputs))
+    return (t1, t2)
+
+
+-- | Handy wrapper to be able to compare things with the 'shallowEq'
+-- combinator, which ignores the different order of the inputs & outputs.
+data DeepEqual = DeepEqual TxMeta
```
edsko: If changing naming, need to change here too.

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
adinapoli-iohk: yup!

wallet-new/src/Cardano/Wallet/Kernel/DB/TxMeta/Types.hs
adinapoli-iohk: Yea, naming is hard, and I appreciate this nomenclature is a bit confusing -- your choice of names sounds good to me (bikeshed alert: I like `strictlyEqualTo` a bit better but I don't want the reader to get confused with lazy vs strict , so I won't pull the trigger on that :trollface: ).

Like `isomorphicTo` a lot -- the latin part of me approves 😀

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,562 @@
...
+    sqlValueSyntax txid = sqlValueSyntax (sformat hashHexF txid)
+
+instance HasSqlValueSyntax SqliteValueSyntax Core.Coin where
+    sqlValueSyntax = sqlValueSyntax . Core.unsafeGetCoin
+
+-- NOTE(adn) Terribly inefficient, but removes the pain of dealing with
+-- marshalling and unmarshalling of 'Integer'(s), as there is no 'HasSqlValueSyntax'
+-- defined for them.
+-- Using 'Scientific' would be a possibility, but it breaks down further as
+-- there is no 'FromField' instance for them.
```
parsonsmatt: We could treat the `Integer` as a `Word64`, which almost certainly has a relevant instance. A `Word64` can represent POSIX timestamps up to May 28th, 586524 :smile: 

```haskell
λ> import Data.Time
λ> import Data.Time.Clock.POSIX
λ> import Data.Word
λ> :set -XNumDecimals
λ> posixSecondsToUTCTime (fromIntegral ((maxBound :: Word64) `div` 1e6))
586524-01-19 08:01:49 UTC
```

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,562 @@
...
+-- | Generates a Beam's AST fragment for use within a SQL query, to order
+-- the results of a @SELECT@.
+toBeamSortDirection :: SortDirection
+                    -> SQL.QExpr (Sql92OrderingExpressionSyntax (Sql92SelectOrderingSyntax SqliteSelectSyntax)) s a
+                    -> SQL.QOrd (Sql92SelectOrderingSyntax SqliteSelectSyntax) s a
+toBeamSortDirection Ascending  = SQL.asc_
+toBeamSortDirection Descending = SQL.desc_
+
+-- | Generates a function suitable for 'sortBy' out of a 'Sorting'.
+toOrdering :: Sorting -> (Kernel.TxMeta -> Kernel.TxMeta -> Ordering)
```
parsonsmatt: why the parens?

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,562 @@
...
+
+import qualified Cardano.Wallet.Kernel.DB.TxMeta.Types as Kernel
+import qualified Pos.Core as Core
+import           Pos.Crypto.Hashing (decodeAbstractHash, hashHexF)
+
+
+-- | An opaque handle to the underlying 'Sqlite.Connection'.
+data MetaDBHandle = MetaDBHandle {
+    _mDbHandleConnection :: Sqlite.Connection
+}
```
parsonsmatt: in the interest of easier testing/mocking, I think I would rather see this as a "record of functions" that contain the API we're implementing :smile: Then we can swap SQLite out for a pure database if we ever need to for testing.

wallet-new/src/Cardano/Wallet/Kernel/DB/TxMeta/Types.hs
```diff
@@ -0,0 +1,173 @@
...
+                       <*> arbitrary
+                       <*> arbitrary
+                       <*> arbitrary
+
+-- | Generates 'NonEmpty' collections which do not contain duplicates.
+-- Limit the size to ~10 elements
+uniqueElements :: Gen (NonEmpty (Core.Address, Core.Coin))
+uniqueElements = do
+    (e :| es) <- arbitrary
+    return (e :| (List.take 1 $ List.filter (/= e) (List.nub es)))
```
parsonsmatt: Why not generate a `Set (Core.Address, Core.Coin)` of the desired size and turn it into a `NonEmpty`?

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
+{-# LANGUAGE RankNTypes #-}
+module TxMetaStorageSpecs (txMetaStorageSpecs) where
```
parsonsmatt: :heart_eyes_cat:

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
...
+    let t1 = blueprint & over txMetaInputs  (const inputs)
+                       . over txMetaOutputs (const outputs)
+    let t2 = blueprint & over txMetaInputs  (const (NonEmpty.reverse inputs))
+                       . over txMetaOutputs (const (NonEmpty.reverse outputs))
+    return (t1, t2)
+
+
+-- | Handy wrapper to be able to compare things with the 'isomorphicTo'
+-- combinator, which ignores the different order of the inputs & outputs.
+data DeepEqual = DeepEqual TxMeta
```
parsonsmatt: if `newtype`, then you can use `GeneralizedNewtypeDeriving` to get rid of some of the boilerplate here

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
...
+        sortFn (Isomorphic a) (Isomorphic b) =
+            (withDir direction compare) (a ^. txMetaCreationAt) (b ^. txMetaCreationAt)
+
+-- | Specs which tests the persistent storage and API provided by 'TxMeta'.
+txMetaStorageSpecs :: Spec
+txMetaStorageSpecs = do
+    describe "TxMeta equality" $ do
+        it "should be reflexive" $ do
+            (blueprint :: TxMeta) <- liftIO $ generate arbitrary
+            blueprint `exactlyEqualTo` blueprint `shouldBe` True
```
parsonsmatt: `Test.Hspec.QuickCheck` has a `prop` helper that lets you write:

```haskell
prop "should be reflexive" $ \blueprint ->
    bluePrint `exactlyEqualTo` (blueprint :: TxMeta)
```

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
...
+
+        it "should be strict when needed" $ do
+            (t1, t2) <- liftIO genSimilarTxMetas
+            t1 `exactlyEqualTo` t2 `shouldBe` False
```
parsonsmatt: For this, you would write 

```haskell
prop "it should be strict when needed" 
    $ forAll genSimilarTxMetas 
    $ \(t1, t2) ->
        not (t1 `exactlyEqualTo` t2)
```

(would require `genSimilarTxMetas` to be `Gen` not `IO`)

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
...
+
+        it "isomorphicTo is more lenient" $ do
+            (t1, t2) <- liftIO genSimilarTxMetas
+            t1 `isomorphicTo` t2 `shouldBe` True
```
parsonsmatt: likewise, 

```haskell
prop "isomoprhic is more leniant"
    $ forAll genSimilarTxMetas
    $ \(t1, t2) ->
        t1 `isomorphicTo` t2
```

The main reason we want this is because the *error* we get from failure will be much better. With your current code, we'll get something like:

```
isomorphic is more lenient FAILED
    False should be True
```

but using `prop` we'll get the actual values that failed the test.

parsonsmatt: Additionally, the `prop` things will generate 100 or so eaxmples. `liftIO (generate arbitrary)` only tests a single thing!

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
...
+
+    describe "TxMeta storage" $ do
+
+        it "can store a TxMeta and retrieve it back" $ do
+            withTemporaryDb $ \hdl -> do
+                testMeta <- liftIO $ generate arbitrary
+                void $ putTxMeta hdl testMeta
+                mbTx <- getTxMeta hdl (testMeta ^. txMetaId)
+                fmap DeepEqual mbTx `shouldBe` Just (DeepEqual testMeta)
```
parsonsmatt: We may want to use [`Test.QuickCheck.Monadic`](http://hackage.haskell.org/package/QuickCheck-2.11.3/docs/Test-QuickCheck-Monadic.html) here. I think this would be expressed as:

```haskell
it "can store a TxMeta and retrieve it back" $ monadicIO $ do
    testMeta <- pick
    run $ withTemporaryDb $ \hdl -> do
        void $ putTxMeta hdl testMeta
        mbTx <- getTxMeta hdl (testMeta ^. txMetaId)
        fmap DeepEqual mbTx `shouldBe` Just (DeepEqual testMeta)
```

That would generate 100 examples and try it on each of them, instead of just a single one.

wallet-new/test/unit/TxMetaStorageSpecs.hs
```diff
@@ -0,0 +1,167 @@
...
+import           System.Directory (getTemporaryDirectory, removeFile, withCurrentDirectory)
+import           System.IO.Error (isDoesNotExistError)
+import           Test.Hspec (shouldThrow)
+import           Test.QuickCheck (arbitrary, generate, resize, vectorOf)
+import           Util.Buildable.Hspec
+
+import qualified Cardano.Wallet.Kernel.DB.Sqlite as Storage
+
+-- | Handy combinator which yields a fresh database to work with on each spec.
+withTemporaryDb :: forall m a. (MonadIO m, MonadMask m) => (MetaDBHandle -> m a) -> m a
```
parsonsmatt: Why not use the in memory database? If we're just doing this for testing then touching the disk is going to be way slower than just keeping it in memory. Also simpler.

adinapoli-iohk:
Thanks for taking the time out of your day to review this PR, @parsonsmatt !

All in all they are all fair review points. I will need to do some massaging with CI and the windows-specific deps first, but then I will tackle your comments. Thanks!

wallet-new/test/unit/TxMetaStorageSpecs.hs
adinapoli-iohk: Not sure if I'm interpreting your comment correctly, but what I want to test here is specifically that my Sqlite/beam queries are sound, so disk I/IO is a price I'm willing to pay.

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
adinapoli-iohk: Just a visual cue that this really returns a function, but I don't have strong feelings about it.

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,562 @@
...
+    in fmap (mk txid) distribution
+  where
+      mk :: Core.TxId -> (Core.Address, Core.Coin) -> a
+      mk tid (addr, amount) = builder (TxCoinDistributionTable addr amount (TxIdPrimKey tid))
+
+-- | Convenient constructor of a list of 'TxInput' from a 'Kernel.TxMeta'.
+mkInputs :: Kernel.TxMeta -> NonEmpty TxInput
+mkInputs txMeta = mkCoinDistribution txMeta Kernel.txMetaInputs TxInput
+
+instance Beamable TxInputT
```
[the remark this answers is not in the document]
adinapoli-iohk: I guess it's a matter of style -- I prefer smart constructor first, then instances, but no strong objection in moving this closer to the type definition 😉

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
adinapoli-iohk: Probably by that date planet Earth won't exist anymore anyway, so I guess that's enough precision 😅

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,562 @@
...
+
+type TxCoinDistributionTable = TxCoinDistributionTableT Identity
+
+instance Prelude.Show TxCoinDistributionTable where
+    show txCoin =
+      let (TxIdPrimKey txid) = _txCoinDistributionTxId txCoin
+      in formatToString ("id = " % build % ", address = " % build % ", coin = " % build)
+                        txid
+                        (_txCoinDistributionTableAddress txCoin)
+                        (_txCoinDistributionTableCoin txCoin)
```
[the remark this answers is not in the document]
adinapoli-iohk: The reason why I went with a custom `Show` instance (which indeed I try to stir away from as well) was because deriving an instance of `TxPrimKey ...` was bringing me into a needless rabbit hole. Considering I don't think we are using this anyway (in fact, it is a relic of me trying to debug something else via `traceShow`) I'm inclined to simply remove it.

wallet-new/test/unit/TxMetaStorageSpecs.hs
adinapoli-iohk: Ah, sorry, after caffeine kicked in I see what you mean -- you probably refer to the fact that [sqlite can run in memory](https://www.sqlite.org/inmemorydb.html) and thus might be a nice win to avoid the disk I/O to have faster tests. Makes sense, will look into it 😉

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,562 @@
...
+newtype TxInputT f = TxInput  {
+  _getTxInput  :: (TxCoinDistributionTableT f)
+  } deriving Generic
+
+type TxInput = TxInputT Identity
+deriving instance Show TxInput
+
+-- | Generalisation of 'mkInputs' and 'mkOutputs'.
+mkCoinDistribution :: forall a. Kernel.TxMeta
+                   -> (Getter Kernel.TxMeta (NonEmpty (Core.Address, Core.Coin)))
```
k0001: I'm inclined to say one almost never wants a `Getter s a` as an argument, rather than an `s -> a`. One can use `to (_ :: s -> a) :: Getter s a` if necessary.

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
k0001: Be careful of not accidentally having only _seconds_ resolution. We surely want more than that, otherwise we may have to worry about rounding issues. Generally, a representation as a pair `(Int64, Word32)` is a good choice when serializing timestamps. `time` provides [`Data.Time.Clock.System.SystemTime`](https://hackage.haskell.org/package/time-1.9.1/docs/Data-Time-Clock-System.html#v:SystemTime) having that representation, which is exactly what's provided by system calls returning timestamps.

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -0,0 +1,562 @@
...
+                           ))
+
+--- | The full list of migrations available for this 'MetaDB'.
+-- For a more interesting migration, see: https://github.com/tathougies/beam/blob/d3baf0c77b76b008ad34901b47a818ea79439529/beam-postgres/examples/Pagila/Schema.hs#L17-L19
+migrateMetaDB :: MigrationSteps SqliteCommandSyntax () (CheckedDatabaseSettings Sqlite MetaDB)
+migrateMetaDB = migrationStep "Initial migration" initialMigration
+
+
+-- | Migrates the 'MetaDB', potentially mangling the input database.
+-- TODO(adinapoli): Make it safe.
```
k0001: Maybe don't worry about making it safe here. This SQL migration probably won't live here nor use the beam stuff anyway.

adinapoli-iohk:
> Be careful of not accidentally having only seconds resolution. We surely want more than that, otherwise we may have to worry about rounding issues.

@k0001 What you say is surely true and a great review point, but I _think_ (@edsko to confirm or disprove) that for metadata tracking (and for the API history tracking alike) we don't need anything more precise than seconds -- I cannot think of any substantial use case where we might need to distinguish between a Tx arriving into the system with milliseconds or microseconds precision.

Furthermore, there is no `FromField` instance for `(a,b)` (and quite rightly so), therefore a bit more massaging would be needed to store the time serialised that way, to which I might simply save myself the hassle and go with the less inefficient textual representation 😁 

In my local branch I have amended that instance to use `Word64` as per @parsonsmatt 's suggestion which I think strikes the sweet spot between efficiency, requirements and ease of implementation.

My 2 cents anyway :)

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
```diff
@@ -280,9 +260,9 @@ address :: DataType SqliteDataTypeSyntax Core.Address
 address = DataType (varCharType Nothing Nothing)
 
 -- | 'DataType' declaration to convince @Beam@ treating 'Core.Timestamp'(s) as
--- varchars of arbitrary length.
+-- a SQL @INTEGER@.
 timestamp :: DataType SqliteDataTypeSyntax Core.Timestamp
-timestamp = DataType (varCharType Nothing Nothing)
+timestamp = DataType intType
```
parsonsmatt: hmmm, so that may actually give us some truncation. we might want to use the [`UNSIGNED BIG INT`](https://www.sqlite.org/datatype3.html#affinity_name_examples) type instead, which is an 8 byte unsigned int (eg a `Word64)`. 

I mean, it's *fine*, if we really care about the difference between overflow in 52,000 CE and 26,000 CE, we should be using a string :laughing:

wallet-new/src/Cardano/Wallet/Kernel/DB/TxMeta/Types.hs
```diff
@@ -171,3 +188,36 @@ instance Buildable TxMeta where
...
+-- | An opaque handle to the underlying storage, which can be easily instantiated
+-- to a more concrete implementation like a Sqlite database, or even a pure
+-- K-V store.
+data MetaDBHandle = MetaDBHandle {
+      closeMetaDB   :: IO ()
+    , migrateMetaDB :: IO ()
+    , getTxMeta     :: Core.TxId -> IO (Maybe TxMeta)
+    , putTxMeta     :: TxMeta -> IO ()
+    , getTxMetas    :: Offset -> Limit -> Maybe Sorting -> IO [TxMeta]
+    }
```
parsonsmatt: so modular, wow,

the only other thing I'd say is "parameterize it over `m`" for ~*~pure tests~*~ but we can make that change later (eg via changing this to `data MetaDBHandleFor m = ...` and `type MetaDBHandle = MetaDBHandleFor IO`)

parsonsmatt: hey, reviews aren't what's blocking this PR now :trollface:

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
parsonsmatt: ah, but everything really returns a function :smile: :sheep: :dancer:

wallet-new/src/Cardano/Wallet/Kernel/DB/Sqlite.hs
parsonsmatt: The `Timestamp` is stored using `Microseconds` resolution, and that's what we're storing in the database. An integral number of microseconds since the epoch

adinapoli-iohk:
Ok, a bit of report from the trenches on why this PR is not building. In a nutshell, `beam-sqlite` depends on `Win32 >= 2.4`, but we are stuck in using `2.3.x` due to two major factors:

- It's what LTS 9.1 ships;
- Win32 has all kind of reverse deps, including core packages like `process`, `happy` (I believe) etc;

In a nutshell the only way out I see is to try manually backporting some `Win32 >= 2.4` features into a forked version of `beam-sqlite`, in order to remove or make this dependency more lenient.

On top of that my specs are failing on Hydra, but that is a separate problem. :trollface:

adinapoli-iohk:
@edsko @parsonsmatt @k0001 @cleverca22 Ok, I have tried to wrestle a little bit with this over the weekend by weeding out the deps and by changing the `cardano-integration-test` to a shorter name to circumvent the max_path limitation on Windows, but this is still borked. Not only on Appveyor but also on Darwin, despite the use of `enabledSharedExecutable = false` which should have fixed the infamous linker bug, but apparently it hasn't.

For now, I'm parking this PR and declaring defeat, as this turned out to be a huge time sink and I need to move to other tickets to get the velocity going.

Said that, it's quite worrisome to me that we are running so close to the edge of the tar pit: I can imagine that if not me, somebody else will soon hit this limit, as cardano seems to have incorporated quite a few dependencies in a [katamary damacy](https://en.wikipedia.org/wiki/Katamari_Damacy) fashion. 

I think the path forward is the following:

1. Park this PR here for a bit, as the part of the system it delivers is largely independent from the rest of the data layer, but will eventually block @k0001 in his migration task;

2. Add a ticket to CBR-277 to track the fact we need to merge this PR;

3. Have a valiant DevOps help us making this mergeable (this also reminds me our squad doesn't have a dedicated devops engineer assigned, but it would be nice to have, for situations like these).

adinapoli-iohk:
Outstanding problems (as a checklist) which prevents this PR to witness its full glory:

- [ ] AppVeyor fails due to `max_path` limitation (see [this YT ticket](https://iohk.myjetbrains.com/youtrack/issue/CSL-2344));

- [ ] Build on Darwin (via Hydra) fails due to linker space limit exceeded (likely due to too many deps);

- [ ] We are using a forked version of `Win32` to please AppVeyor. The proper fix is to switch to LTS-11.2, which is in the works and @avieth has an open patch for that.

adinapoli-iohk:
@cleverca22 @rvl Sigh, tried to rebase against `develop` but I think I have done more harm than good, and needless to say the linker bug in Darwin is still present.

Do you guys have any cunning plan on how we could escape this conundrum? Ideally we should [revive this script](https://github.com/input-output-hk/cardano-sl/pull/1248) and start thinking about another round of pruning, I suppose..

adinapoli-iohk:
@edsko , look, progress! Only AppVeyor to go 😉

adinapoli-iohk:
After speaking with @erikd , it seems in our brave new world we are temporarily overriding the Windows CI check as it's currently borked but we cannot afford to have development grind to an halt due to it.

@jcmincke , @edsko , little heads up that I'm going to merge this PR very soon (or any other I'm told) overriding CI, until we sort out the CI woes. If you want me to refrain, just shout 😀

adinapoli-iohk:
Drat, a conflict again, I should have merged yesterday!

PR #2903, opened 2018-05-09
[CDEC-238] Move tests for lrc package from lib to lrc

ruhatch:
## Description

The current tests for Lrc were in the `lib` package instead of the `lrc` package.

The code was tightly coupled and used a `configuration.yaml` file to set up configurations. This was only present in the `lib` package. To make things more composable, I create a `defaultCoreConfiguration` in the `core` package. This was the only part of the YAML file needed in the `lrc` tests.

- Move source for `core` and `lrc` to `src` folders in those packages
- Set up `cardano-sl-lrc-test` test-suite in `lrc` package
- Move single Lrc test from `lib` into the test suite
- Create a `defaultCoreConfiguration` in Haskell to avoid `configuration.yaml`

## Linked issue

[CDEC-238](https://iohk.myjetbrains.com/youtrack/issue/CDEC-238)

## Type of change

## Developer checklist

## Testing checklist


erikd:
Obviously a rebase is needed.

You have also managed to run into two separate problems with our CI infrastructure.

Firstly, the CI failure on Linux and Mac is because of the use of nix to in the CI builds for Linux and Mac (its not used on Windows). Some changes (particularly cabal file changes) require modification of the `pkgs/default.nix` file. This if is autogenerated, but needs to be stored in the repo so that CI builds can be pure. Personally, I think there must be a better way to do it, but don't have the time chase this. To get around this, under nix, run the command `pkgs/generate.sh` which will regenerate the `pkgs/default.nix` file and commit the changes and then push.

The second problem is due to you building locally with nix (most devs just use stack without nix) you ran into is is that the `stack2nix` program that is used for nix builds does not propagate the `-Werror` in the `stack.yaml` file down to the nix builds. This same nix build is used on the Linux and Mac CI builds, but the Windows just uses `stack` and hence promoted warnings to errors that were just warnings on your local build. I have a ticket open with devops to fix `stack2nix`.

lrc/cardano-sl-lrc.cabal
```diff
@@ -83,3 +84,26 @@ library
 
   build-tools: cpphs >= 1.19
   ghc-options: -pgmP cpphs -optP --cpp
+
+test-suite cardano-sl-lrc-test
```
erikd: Probably worth working towards some of the great conventions I've been talking about. In that light, that test executable should probably just be called "test".

lrc/cardano-sl-lrc.cabal
```diff
@@ -83,3 +84,26 @@ library
...
+  type:                exitcode-stdio-1.0
+  hs-source-dirs:      test
+  main-is:             Spec.hs
```
erikd: Under the conventions I'm trying to work towards, that file would normally be called "test.hs", but I'm not sure if that works with the `hspec-discover` magic that's being used.

I'm also not sure if this "Spec.hs" file is enough to actually make this a proper test executable. You may been to add a "test.hs" that has a `main` function. Something like `util/test/test.hs`.

You should also make sure that running something like `stack test cardano-sl-lrc` actually does run these tests that you have moved. No point moving them if they aren't going to be run :)

PR #2908, opened 2018-05-11
[DEVOPS-822] pkgs/generate.sh: Work with nix 2.0

rvl:
Fixes this error:

    error: Nix database directory ‘/nix/var/nix/db’ is not writable: Permission denied

Nix 2.0 no longer needs or sets the `NIX_REMOTE` variable, but when using nix-build (1.11.15) inside a shell it requires the variable.

So this provides a default for `NIX_REMOTE=daemon`.

The `NIX_REMOTE` variable setting can be removed completely when we update our nixpkgs pin to something that includes Nix 2.0.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-822

## Type of change

## QA Steps

    ./pkgs/generate.sh

### DIFF
--- a/pkgs/generate.sh
+++ b/pkgs/generate.sh
@@ -12,15 +12,20 @@ function runInShell {
 set -xe
 set -v
 
+# Ensure that nix 1.11 (used by cabal2nix) works in multi-user mode.
+if [ ! -w /nix/store ]; then
+    export NIX_REMOTE=${NIX_REMOTE:-daemon}
+fi
+
 # Get relative path to script directory
 scriptDir="$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")"
 
 pushd "${scriptDir}"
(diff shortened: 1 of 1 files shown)


rvl:
I have added on another one-liner so that hydra builds cardano-sl-explorer-frontend.

cleverca22:
i tried this already, if you force `NIX_REMOTE=daemon` in a single-user setup that lacks the daemon, then it fails
i think the only solution is to use nix2 in the nix-shell, or just not even include nix, the host already has nix

rvl:
Well something needs to be done - we can't just leave it broken. I like the idea of not including nix in the shell because it wouldn't have ever been necessary.

rvl:
It turns out that removing `nix` from the shell inputs doesn't help because cabal2nix uses nix 1.11.
I have updated so that it only applies the default in nix multi-user setups.

PR #2909, opened 2018-05-11
[CBR-265] Check for each file required for TLS instead of only the parent dir

KtorZ:
## Description

A bit ugly but it's not like the `cardano-launcher` gives us more flexibility. This will now at least check that files are there and regenerate all of them if one is missing. 

Note that I've removed the `tlsPath` option that I previously added to the launcher config to avoid duplicating information here. The downside is that it becomes complicated if we need to generate certificates in different directories (which we don't do). The configs in the `daedalus` 's PR will need to be adjusted as well.

## Linked issue

[[CBR-265]](https://iohk.myjetbrains.com/youtrack/issue/CBR-265)

## Type of change

## Developer checklist

## Testing checklist

## QA Steps

Start a cluster using the launcher (`./scripts/launcher/demo-with-wallet-api.sh`) in different scenario (with no `./run/tls-files` folder, with missing files in that folder, or with all files).

### DIFF
--- a/tools/src/launcher/Main.hs
+++ b/tools/src/launcher/Main.hs
@@ -38,8 +38,7 @@ import           Options.Applicative (Parser, ParserInfo, ParserResult (..), def
                                       header, help, helper, info, infoOption, long, metavar,
                                       progDesc, renderFailure, short, strOption)
 import           Serokell.Aeson.Options (defaultOptions)
-import           System.Directory (createDirectoryIfMissing, doesDirectoryExist, doesFileExist,
-                                   removeFile)
+import           System.Directory (createDirectoryIfMissing, doesFileExist, removeFile)
 import qualified System.Directory as Sys
 import           System.Environment (getExecutablePath, getProgName, setEnv)
 import           System.Exit (ExitCode (..))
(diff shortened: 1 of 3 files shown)


tools/src/launcher/Main.hs
```diff
@@ -365,9 +363,40 @@ main =
...
+data TlsPaths = TlsPaths
+    { tlsCaPath     :: FilePath
+    , tlsServerPath :: FilePath
+    , tlsKeyPath    :: FilePath
+    } deriving (Show)
+
+
+findTlsArgs :: [Text] -> M TlsPaths
+findTlsArgs args = do
+    tlsCaPath     <- findArg "--tlsca"   (pure . toString) args
```
akegalj: This will add one more runtime failure point  if someone changes how tls arguments are parsed from `NodeOptions`. Maybe exporting these constants "tlsca" "tlscert" and "tlskey" from `NodeOptions` module ?

KtorZ: Not sure if it worth the effort here :/ 

- The failure will be quite clear
- There's no reason to change option names and whoever would want to that will probably be very careful in doing so
- We are planning on ditching completely the launcher moving forward

PR #2914, opened 2018-05-11
[CDEC-161] 5th instalment of networking/develop merge

avieth:
## Description

5th installment of the incremental merge of `networking/develop` into `develop`.

Find here:

- Simplification of the full logic layer (one does not "run" it, one only creates it).
- `SendActions` are exposed in the full diffusion layer internals, useful for doing simulations / tests / benchmarks.
- The statsd reporting server is in `IO`, making it simpler to deal with.
- `runFullDiffusionLayer` uses proper concurrency, but is still buggy since we use an old async version (#2745 will fix this).
- Improvements to the monstrous dispatcher thread.
- Outbound queue shutdown fix.

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/networking/src/Node/Internal.hs
+++ b/networking/src/Node/Internal.hs
@@ -700,9 +700,9 @@ killNode node = do
         if _nodeStateClosed nodeState
         then throwIO $ userError "killNode : already killed"
         else pure (nodeState { _nodeStateClosed = True }, ())
-    nodeCloseEndPoint node
     uninterruptibleCancel (nodeDispatcherThread node)
     killRunningHandlers node
+    nodeCloseEndPoint node
 
 data ConnectionState peerData =
 
--- a/networking/src/Node/Internal.hs
+++ b/networking/src/Node/Internal.hs
@@ -38,19 +38,20 @@ module Node.Internal (
     ChannelOut(..),
     startNode,
     stopNode,
+    killNode,
     withInOutChannel,
     writeMany,
     Timeout(..)
   ) where
 
-import           Control.Concurrent (ThreadId, threadDelay)
+import           Control.Concurrent (threadDelay)
 import           Control.Concurrent.STM
 import           Control.Concurrent.Async
 import           Control.Concurrent.MVar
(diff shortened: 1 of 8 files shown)


avieth:
This is the penultimate instalment. All that remains is a change to the definition of subscriptions, to make it more modular: you don't have to use DNS, you can give a particular address, which is useful for testing.

networking/src/Node/Internal.hs
```diff
@@ -191,7 +201,8 @@ constantReceiveDelay = pure . Just
 -- | A 'Node' is a network-transport 'EndPoint' with bidirectional connection
 --   state and a thread to dispatch network-transport events.
 data Node packingType peerData = Node {
-       nodeEndPoint         :: NT.EndPoint
+       nodeTrace            :: Trace IO (Severity, Text)
```
coot: Good to see the `Trace`, so we can run  IO!

avieth:
@iohk-devops is appveyor stuck?

networking/src/Node/Internal.hs
```diff
@@ -665,21 +675,34 @@ startNode logTrace packing peerData mkNodeEndPoint mkReceiveDelay mkConnectDelay
...
+
+-- | Kill a 'Node', terminating its dispatcher thread, closing its endpoint,
+-- and killing all of its handlers.
+killNode :: Node packingType peerData -> IO ()
+killNode node = do
+    modifyMVar (nodeState node) $ \nodeState ->
+        if _nodeStateClosed nodeState
+        then throwIO $ userError "killNode : already killed"
+        else pure (nodeState { _nodeStateClosed = True }, ())
+    nodeCloseEndPoint node
```
coot: Maybe it's nicer to first kill the dispatcher thread and running handlers and close the endpoint afterwards?

avieth: It's not clear whether it makes any difference. In practice, with the TCP transport, closing the endpoint means closing the socket for each connected peer and then updating some state. This could take a bit of time and I suppose that we shouldn't have the dispatcher thread or any handlers running while it's done. I'll make the change, should be fine.

PR #2925, opened 2018-05-14
6th and final instalment of networking/develop merge

avieth:
## Description

In this final instalment find:

- The subscription mechanism was made more modular: you don't have to use DNS to resolve peers. This is useful for testing/simulation.
- The pure logic layer properly seeds the deterministic key generator.
- Full diffusion layer `OutSpecs` include the subscription stuff even if there's no subscription worker. This should be viewed as a workaround for the weird `OutSpecs` system: if we run a test/simulation setup in which we don't use a "subscription worker" thread (currently taken care of by the full diffusion layer) but instead wish to manually set up the subscription (`withNetworkSubscription`) it would fail because the `OutSpecs` wouldn't be correct.

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/infra/Pos/Diffusion/Subscription/Dns.hs
+++ b/infra/Pos/Diffusion/Subscription/Dns.hs
@@ -87,7 +87,7 @@ resolveOne logTrace defaultPort nodeAddrs = do
     let (errs, nids_) = partitionEithers mNodeIds
         nids = mconcat nids_
     when (null nids)       $ traceWith logTrace (Error, msgNoRelays)
-    when (not (null errs)) $ traceWith logTrace (Error, msgDnsFailure errs)
+    unless (null errs) $ traceWith logTrace (Error, msgDnsFailure errs)
     return nids
 
   where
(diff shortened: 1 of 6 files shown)


avieth:
@coot I juggled the subscription stuff. I hope I didn't mess anything up. One change I did intentionally make is to have the subscription duration _not_ shared amongst all concurrent subscriptions, which I think is better.

infra/Pos/Diffusion/Subscription/Dns.hs
```diff
@@ -1,170 +1,172 @@
...
+            oq
+            BucketBehindNatWorker
+            NodeRelay
+            peersVar
+            keepalive
+            subStatus
+            sendActions
+            nodeId
+        timeToWait <- retryInterval duration <$> slotDuration
+        threadDelay (fromIntegral timeToWait * 1000)
```
coot: Let me just note that there is a difference here; previously we computed the max of duration time for all alternative peers and we put `threadDelay` after we tried all the alternative peers; now we're using the previous duration time to wait until we subscribe to the next alternative peer.  This way we will punish more a good peer which is among bad ones (i.e. with short connection times): if there is one good peer and `n` bad ones, now we will wait `n` times for a long time, while previously we'd waited once for a short time and spin quickly through the list of bad peers.  There is a downside of this behaviour: lost bandwidth for trying bad peers too often.  But from the other side, I think we expect more often a single bad peer among good ones: and in this case this works more or less the same.

PR #2947, opened 2018-05-17
[CBR-245] Transaction submission according to the wallet spec

adinapoli-iohk:
## Description

This work-in-progress PR is a sneak-peek over a possible implementation of the transaction submission mechanism outlined in the wallet spec. It ships with a fully-functional scheduler and a bunch of stock resubmission functions, but this is still not complete because:

- [x] More tests could/should be added;
- [ ] The `WalletSubmission` layer should be wired to a green thread actually "ticking" it;
- [ ] `addPending` and `remPending` should be wired into `applyBlock` and `rollbackBlock`;
- [x] I need to think harder about the implications of receiving transactions which have a topological dependency on other, and how this retrofit to the scheduling. At the moment we try calling `topSortTxs` in a best-effort way when we compute the transactions due each slot.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-245

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
@@ -1,0 +1,106 @@
+module Cardano.Wallet.Kernel.Submission (
+    -- * Public API
+      newWalletSubmission
+    , addPending
+    , remPending
+    , tick
+
+    -- * Types and lenses
+    , WalletSubmission
+    , wsResubmissionFunction
+    ) where
+
+import           Universum
+
+import           Control.Lens.TH
+import           Data.HashPSQ (HashPSQ)
+import qualified Data.HashPSQ as PSQ
+import qualified Data.Map as M
+
+import           Cardano.Wallet.Kernel.DB.Spec (Pending (..), emptyPending)
+import qualified Pos.Core as Core
+
+import           Cardano.Wallet.Kernel.Diffusion (WalletDiffusion (..))
+
+-- | Wallet Submission Layer
+--
+-- This module implements section 9 of the Wallet spec,
+-- namely 'Transaction Submission'.
+--
+data WalletSubmission = WalletSubmission {
+      _wsResubmissionFunction :: ResubmissionFunction
+    , _wsState                :: WalletSubmissionState
+    }
+
+data WalletSubmissionState = WalletSubmissionState {
+      _wssPendingSet :: Pending
+    , _wssScheduler  :: SubmissionScheduler
+    , _wssDiffusion  :: WalletDiffusion
+    }
+
+data SubmissionScheduler = SubmissionScheduler {
+    _ssSubmissionQueue :: HashPSQ Core.TxId SubmissionPriority Core.TxAux
+    }
+
+-- | TODO(adn) Write better instances.
+newtype SubmissionPriority =
+  SubmissionPriority (Core.SlotId, SubmissionCount) deriving (Eq, Ord)
+
+type SubmissionCount = Int
+
+-- FIXME(adn) Just an ill-named placeholder.
+type ToResubmit = ()
+
+type ResubmissionFunction = (ToResubmit, SubmissionScheduler) -> Pending
+
+makeLenses ''WalletSubmission
+makeLensesFor [("_wssPendingSet", "wssPendingSet")] ''WalletSubmissionState
+
+--
+-- Public API, as written in the spec
+--
+
+newWalletSubmission :: ResubmissionFunction
+                    -> WalletDiffusion
+                    -> WalletSubmission
+newWalletSubmission resubmissionFunction diffusion = WalletSubmission {
+      _wsResubmissionFunction = resubmissionFunction
+    , _wsState = newEmptyState
+    }
+    where
+        newEmptyState :: WalletSubmissionState
+        newEmptyState = WalletSubmissionState {
+              _wssPendingSet = emptyPending
+            , _wssScheduler  = SubmissionScheduler PSQ.empty
+            , _wssDiffusion  = diffusion
+            }
+
+-- | Informs the 'WalletSubmission' layer of the new, updated 'PendingSet'.
+addPending :: Pending -> WalletSubmission -> WalletSubmission
+addPending updatedPending ws =
+    ws & over (wsState . wssPendingSet) (unionPending updatedPending)
+
+remPending :: Pending -> WalletSubmission -> WalletSubmission
+remPending updatedPending ws =
+    ws & over (wsState . wssPendingSet) (intersectPending updatedPending)
+
+-- | A \"tick\" of the scheduler.
+-- Returns the set transactions which needs to be droppped by the system as
+-- they likely exceeded the submission count and they have no chance to be
+-- adopted in a block.
+tick :: WalletSubmission -> (Pending, WalletSubmission)
+tick ws = (emptyPending, ws)
+
+--
+-- Internal API
+--
+
+-- TODO(adn): Better placed into module 'Cardano.Wallet.Kernel.DB.Spec'?
+unionPending :: Pending -> Pending -> Pending
+unionPending (Pending new) (Pending old) =
+    Pending (M.union <$> new <*> old)
+
+intersectPending :: Pending -> Pending -> Pending
+intersectPending (Pending new) (Pending old) =
+    Pending (M.intersection <$> new <*> old)
+
--- a/wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE BangPatterns #-}
 module Cardano.Wallet.Kernel.Submission (
     -- * Public API
       newWalletSubmission
@@ -10,10 +9,11 @@ module Cardano.Wallet.Kernel.Submission (
     -- * Types and lenses
     , Evicted
     , ResubmissionFunction
-    , SubmissionCount (..)
+    , Schedule
     , ScheduledTx (..)
-    , Schedule (..)
+    , SchedulingError (..)
     , Slot (..)
+    , SubmissionCount (..)
     , WalletSubmission
     , mapSlot
     , wsResubmissionFunction
@@ -40,12 +40,14 @@ import           Control.Lens.TH
 import           Data.IntMap.Strict (IntMap)
 import qualified Data.IntMap.Strict as IntMap
 import qualified Data.List as List
+import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map.Strict as M
 import           Data.Text.Buildable (build)
-import           Formatting (bprint, (%))
+import           Formatting (bprint, sformat, (%))
 import qualified Formatting as F
 import           Pos.Crypto.Hashing (WithHash (..))
 import           Pos.Txp.Topsort (topsortTxs)
+import qualified Prelude
 import           Serokell.Util.Text (listJsonIndent, mapBuilder, pairF)
 import           Test.QuickCheck
 
@@ -83,8 +85,18 @@ instance Buildable WalletSubmissionState where
 
 -- | A 'Schedule'. Despite modelled as in 'IntMap' it has to be intended
 -- as a mapping between 'Slot' and the list of transactions due that slot.
-newtype Schedule = Schedule {
-    _ssScheduled :: IntMap [ScheduledTx]
+data Schedule = Schedule {
+      _ssScheduled     :: IntMap [ScheduledTx]
+    , _ssUnsentNursery :: [ScheduledTx]
+    -- ^ A list of unsent transactions which couldn't be sent due to dependency
+    -- over transactions scheduled in some other slot. Practical example: Let
+    -- @A@ be a transaction scheduled for slot @N + 3@ and let @B@ be a transaction
+    -- @which depends on A@ scheduled for slot @N@. If we were to send @B@ we
+    -- would make a mistake, as it cannot be adopted before @A@ does. The solution
+    -- is to capture this event in 'dueThisSlot', and putting @B@ into the
+    -- nursery up until it can be sent.
+    -- NOTE(adn): Should we bump the 'SubmissionCount' of 'ScheduledTx' in
+    -- the nursery, to avoid them being stuck forever?
     }
 
 data ScheduledTx = ScheduledTx {
@@ -153,7 +165,7 @@ type ResubmissionFunction m =  Slot
                             -- ^ The transactions to remove together with
                             -- the new 'Schedule'.
 
-makeLenses ''Schedule
+makeLensesFor [("_ssScheduled", "ssScheduled")] ''Schedule
 makeLenses ''WalletSubmission
 makeLenses ''WalletSubmissionState
 
@@ -161,7 +173,6 @@ instance Buildable Schedule where
     build ss =
         let elems = ss ^. ssScheduled . to IntMap.toList
         in bprint (F.later mapBuilder) elems
-
 
 instance Arbitrary SubmissionCount where
     arbitrary = SubmissionCount <$> choose (0, 255)
@@ -174,9 +185,9 @@ instance Arbitrary SubmissionCount where
 genSchedule :: Pending -> Slot -> Gen Schedule
 genSchedule pending (Slot lowerBound) = do
     let pendingTxs  = pending ^. pendingTransactions . fromDb . to M.toList
-    slots <- vectorOf (length pendingTxs) (fmap Slot (choose (lowerBound, lowerBound + 10)))
-    retries    <- vectorOf (length pendingTxs) (choose (0, 255))
-    return $ Schedule $ List.foldl' updateFn mempty (zip3 slots pendingTxs retries)
+    slots    <- vectorOf (length pendingTxs) (fmap Slot (choose (lowerBound, lowerBound + 10)))
+    retries  <- vectorOf (length pendingTxs) (choose (0, 255))
+    return $ Schedule (List.foldl' updateFn mempty (zip3 slots pendingTxs retries)) mempty
     where
         updateFn acc (slot, (txId, txAux), retries) =
             let s = ScheduledTx txId txAux (SubmissionCount retries)
@@ -193,9 +204,8 @@ genWalletSubmission :: ResubmissionFunction m -> Gen (WalletSubmission m)
 genWalletSubmission rho =
     WalletSubmission <$> pure rho <*> arbitrary
 
-
 --
--- Public API, as written in the spec
+-- Public API, as written in the spec or mandated by real-world necessities.
 --
 
 newWalletSubmission :: ResubmissionFunction m -> WalletSubmission m
@@ -208,7 +218,7 @@ newWalletSubmission resubmissionFunction = WalletSubmission {
         newEmptyState = WalletSubmissionState {
               _wssPendingSet  = emptyPending
             , _wssCurrentSlot = Slot 0
-            , _wssSchedule   = Schedule IntMap.empty
+            , _wssSchedule   = Schedule IntMap.empty mempty
             }
 
 -- | Informs the 'WalletSubmission' layer about new 'Pending' transactions.
@@ -230,7 +240,7 @@ remPending updatedPending ws =
 -- local 'Pending' set. It will be caller's responsibility to call 'remPending'
 -- on the 'WalletSubmission'.
 --
-tick :: Monad m
+tick :: MonadThrow m
      => WalletSubmission m
      -- ^ The current 'WalletSubmission'.
      -> m (Evicted, WalletSubmission m)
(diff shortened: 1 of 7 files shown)


wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+
+-- | Removes the input 'Pending' from the local 'WalletSubmission' pending set.
+remPending :: WalletSubmission m -> Pending -> WalletSubmission m
+remPending ws updatedPending =
+    ws & over (wsState . wssPendingSet) (flip differencePending updatedPending)
+
+-- | A \"tick\" of the scheduler.
+-- Returns the set transactions which needs to be droppped by the system as
+-- they likely exceeded the submission count and they have no chance to be
+-- adopted in a block, or that have simply been successfully retrasmitted.
```
edsko: Typo. Also, I don't understand the last part "that have simply been successfully retransmitted". What do you mean?

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+     -> m (Evicted, WalletSubmission m)
+     -- ^ The set of transactions upper layers will need to drop, together
+     -- with the new 'WalletSubmission'.
+tick ws = do
+    let wss               = ws  ^. wsState
+        currentSlot       = wss ^. wssCurrentSlot
+        scheduler         = ws  ^. getScheduler
+        rho               = _wsResubmissionFunction ws
+        dueNow            = dueThisSlot currentSlot ws
+    (toPrune, scheduler') <- rho currentSlot dueNow scheduler
```
edsko: If we change the naming of `Scheduler`, we should remember to change it here also. (And I find it equally confusing here; `scheduler` isn't actually a scheduler, but rather a schedule).

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+    let newState = ws & over (wsState . wssScheduler) (const (purgeSlot currentSlot scheduler'))
+                      . over (wsState . wssCurrentSlot) succ
```
edsko: I'd suggest having `dueThisSlot` return a pair of the transactions that are due, as well as the remaining slots. Could even implement it as a `pattern` if you wanted to get fancy :) But that would be the idea: we're basically doing a pattern match `x : xs`; it's just that it happens to be a sparse implementation. But would still be nicer than later having to remove "the head"; and `purgeSlot` suggests that we're actually doing something with the stuff that we purged, but we're not really.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+        topSorted  = topsortTxs toTx slotted
+    in case topSorted of
+            Nothing     -> slotted
+            Just sorted -> sorted
+    where
+        toTx :: Scheduled -> WithHash Core.Tx
+        toTx (Scheduled (txId, txAux, _)) =  WithHash (Core.taTx txAux) txId
+
+-- | Overrides the 'Scheduler' with an input transaction and a
+-- custom 'Schedule'. Useful to force dispatching, especially in tests.
```
edsko: Would be good to document what's unsafe about this. Is this used only for the tests? If so, might want to document that.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+                toEntry (txId, txAux) = Scheduled (txId, txAux, Schedule 0)
+            in Scheduler (IntMap.insertWith mappend (succ currentSlot) txs s)
+
+-- | Purges all the scheduled transactions in the given 'Slot'. This is
+-- automatically called by 'tick', and it's responsibility of @rho@ to
+-- correctly reschedule all the transaction.
+purgeSlot :: Slot -> Scheduler -> Scheduler
+purgeSlot slot (Scheduler s) = Scheduler (IntMap.delete slot s)
+
+-- TODO(adn): Better placed into module 'Cardano.Wallet.Kernel.DB.Spec'?
```
edsko: Yeah, let's.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+
+
+-- | The 'Evicted' set represents the transactions which needs to be
+-- pruned from the local (and wallet) 'Pending' set.
+type Evicted = Pending
+
+-- | A 'ResubmissionFunction' (@rho@ in the spec), parametrised by an
+-- arbitrary @m@.
+type ResubmissionFunction m =  Slot
+                            -- ^ The current slot
```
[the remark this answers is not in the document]
edsko: Ah, I see (by reading the later code). Since this isn't a list anymore, this is the only way that we can say "insert 10 slots from now". Might be useful to mention this.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+addOneToPending txId aux (Pending p) = Pending (M.insert txId aux <$> p)
+
+-- Ready-to-use 'ResubmissionFunction's.
+
+-- | A 'RetryPolicy' is simply a function which instruct the 'Scheduler' when
+-- to attempt resubmitting the given 'Scheduled' item.
+type RetryPolicy = Slot -> Slot
+
+--
+-- Stock retry policies inspired by the 'retry' package.
```
edsko: Nice, I like this abstraction :)

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+
+constantRetry :: RetryPolicy
+constantRetry = succ
+
+exponentialBackoff :: RetryPolicy
+exponentialBackoff retries =
+    let normalised = (fromIntegral $ retries + 1) * (10.0 :: Double)
+    in retries + floor (logBase (2.0 :: Double) normalised)
+
+fibonacciBackoff :: RetryPolicy
```
edsko: Wait, what? Really? When is this useful? Fascinating idea :)

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+fibonacciBackoff base = fib (base + 1) (0, base)
+    where
+      fib 0 (a, _)    = a
+      fib !m (!a, !b) = fib (m-1) (b, a + b)
```
edsko: If we're actually serious about using this, did you realize that there is a closed form to compute fib? No need to recurse. (https://en.wikipedia.org/wiki/Fibonacci_number#Closed-form_expression). Of course that'll only be relevant for big values. Just thought I'd mention it.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
edsko: Do you think though that it might make sense to change it to `Slot -> Maybe Slot`, so that these retry policies can decide when to give up, rather than hardcoding the `255` limit in `defaultResubmissionFunction`?

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,393 @@
...
+            case retries >= maxRetries of
+                 True -> return (addOneToPending txId txAux evicted, acc)
+                 False -> do
+                     result <- send txAux
+                     case result of
+                         False -> return (evicted, Scheduler s')
+                         True  -> return (addOneToPending txId txAux evicted, acc)
+
+-- | A very lenient resubmitter which blindly schedules each pending transaction
+-- to be resubmitted the next slot.
```
edsko: Slightly worried that this will end up being used in the real code base. Maybe we should just parameterize this with the retry policy also, so that we force callers to make a decision, rather than just using the most convenient function.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+partitionScheduledTx pending xs = List.partition dependsOnPendingTx xs
+    where
+        -- | A 'ScheduledTx' is @not@ independent and should not be sent
+        -- over the wire if any of the inputs it consumes are mentioned in
+        -- the 'Pending' set.
+        dependsOnPendingTx :: ScheduledTx -> Bool
+        dependsOnPendingTx (ScheduledTx _ txAux _) =
+            let txs = pending ^. pendingTransactions . fromDb
+                inputs = List.foldl' updateFn mempty $ (Core.taTx txAux) ^. Core.txInputs . to NonEmpty.toList
+            in any (isJust . flip M.lookup txs) inputs
```
edsko: Is this quite right? Transactions that depend on other pending transactions but are scheduled for the same slot are okay, no?

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+
+        updateFn :: [Core.TxId] -> Core.TxIn -> [Core.TxId]
+        updateFn acc (Core.TxInUnknown _ _)   = acc
+        updateFn acc (Core.TxInUtxo txHash _) = txHash : acc
+
+
+-- | Overrides the 'Schedule' with an input transaction and a
+-- custom 'Schedule'. Useful to force dispatching, especially in tests.
```
edsko: I know I suggested this name but reading through this again, `override` is a bit of a misnomer actually no? It suggests that we replace something, but in fact we're appending. What do you think?

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+type Exponent   = Double
+type MaxRetries = Int
+
+--
+-- Stock retry policies inspired by the 'retry' package.
+--
+
+-- | Very simple policy which merely retries to resubmit the very next 'Slot',
+-- up until 'MaxRetries' attempts.
+constantRetry :: MaxRetries -> RetryPolicy
```
edsko: Do you think that it might be good to give this function an argument so that `constantRetry n` retires every `n` slot, forcing callers to think about how often they want to retry? Still a bit worried about this ending up in a code base.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+purgeSlot :: Slot -> Schedule -> Schedule
+purgeSlot slot (Schedule s n) = Schedule (IntMap.delete (castSlot slot) s) n
+
+-- Ready-to-use 'ResubmissionFunction's.
+
+-- | A 'RetryPolicy' is simply a function which instruct the 'Schedule' when
+-- to attempt resubmitting the given 'ScheduledTx' item.
+type RetryPolicy = SubmissionCount -> Slot -> Maybe Slot
+
+limited :: MaxRetries -> RetryPolicy -> RetryPolicy
```
edsko: Maybe add a comment stating that this is an internal function only, and that the policies we provide take `MaxRetries` as an argument to enforce that we _always_ have a maximum?

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+                reschedule targetSlot =
+                    IntMap.insertWith mappend
+                                      (castSlot targetSlot)
+                                      [incSubmissionCount entry]
+                                      s
+            case retryPolicy retries currentSlot of
+                 Nothing -> return (singletonPending txId txAux `unionPending` evicted, acc)
+                 Just newSlot -> do
+                     result <- send txAux
+                     case result of
```
edsko: Why this pattern match? If sending fails, do we really want to give up on it?

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+    where
+        updateFn :: (Evicted, Schedule) -> ScheduledTx -> m (Evicted, Schedule)
+        updateFn (evicted, acc@(Schedule s nursery)) entry = do
+            let (ScheduledTx txId txAux retries) = entry
+                reschedule targetSlot =
+                    IntMap.insertWith mappend
+                                      (castSlot targetSlot)
+                                      [incSubmissionCount entry]
+                                      s
+            case retryPolicy retries currentSlot of
```
edsko: The order here is correct (I _think_; this is the second time I'm writing this comment, I had the opposite conclusion first :) but it's confusing. We find a transaction that is due _now_, but instead of sending it, we _first_ check the retry policy to see if we should try again. That's.. odd. Not entirely sure what to suggest here, perhaps we should discuss it briefly. A comment or a different name or something may help. (Or maybe I'm just missing the point.)

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+    , _ssUnsentNursery :: [ScheduledTx]
+    -- ^ A list of unsent transactions which couldn't be sent due to dependency
+    -- over transactions scheduled in some other slot. Practical example: Let
+    -- @A@ be a transaction scheduled for slot @N + 3@ and let @B@ be a transaction
+    -- @which depends on A@ scheduled for slot @N@. If we were to send @B@ we
+    -- would make a mistake, as it cannot be adopted before @A@ does. The solution
+    -- is to capture this event in 'dueThisSlot', and putting @B@ into the
+    -- nursery up until it can be sent.
+    -- NOTE(adn): Should we bump the 'SubmissionCount' of 'ScheduledTx' in
+    -- the nursery, to avoid them being stuck forever?
```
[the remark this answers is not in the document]
adinapoli-iohk: > But actually I think that that gives us the clue out of this conumdrum. It should be the wallet's responsibility (not the submission layer's) to make sure that when it gives up ona transaction A, it also gives up on all transactions B that depend on A.

Ah, that's a good point -- will amend the documentation accordingly.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+
+        -- The current 'Schedule'.
+        schedule :: Schedule
+        schedule = ws ^. getSchedule
+
+        toTx :: ScheduledTx -> WithHash Core.Tx
+        toTx (ScheduledTx txId txAux _) =  WithHash (Core.taTx txAux) txId
+
+
+partitionScheduledTx :: Pending -> [ScheduledTx] -> ([ScheduledTx], [ScheduledTx])
```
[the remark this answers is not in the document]
adinapoli-iohk: Yea, I agree -- originally I was allowing it to take a function, a bit like the conventional `Data.List.partition` function, but later sacrificed it in name of brevity.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+    build wss = bprint ("{ pendingSet = " % F.build %
+                        ", scheduler  = " % F.build %
+                        ", slot       = " % F.build %
+                        " } "
+                       ) (_wssPendingSet wss) (_wssSchedule wss) (getSlot $ _wssCurrentSlot wss)
+
+-- | A 'Schedule'. Despite modelled as in 'IntMap' it has to be intended
+-- as a mapping between 'Slot' and the list of transactions due that slot.
+data Schedule = Schedule {
+      _ssScheduled     :: IntMap [ScheduledTx]
```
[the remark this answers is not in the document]
edsko: Err oops. Yes, let's just move that to where it belongs :)

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+dueThisSlot :: Slot
+            -- ^ The current 'Slot'.
+            -> WalletSubmission m
+            -- ^ The 'WalletSubmissionState'.
+            -> Either SchedulingError ([ScheduledTx], Schedule)
+            -- ^ An error if no schedule can be produced, or the scheduled
+            -- transactions together with the new, updated 'Schedule'.
+dueThisSlot currentSlot ws =
+    let localPending = ws ^. wsState . wssPendingSet
+        (dependent, independent) = partitionScheduledTx localPending candidates
```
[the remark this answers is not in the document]
adinapoli-iohk: Yes, fair point -- maybe `sendable` or some more specific name is more appropriate? We could surely add a disclaimer under the form of a comment if you think it's appropriate, in case we want to keep the `dependent vs independent` partition going.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
adinapoli-iohk: great catch -- this is quite (too) aggressive; we should really be lenient and check, if we find a match, if the input is in the same slot, and if yes, we allow it to be sent.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
```diff
@@ -0,0 +1,431 @@
...
+        topSorted  = topsortTxs toTx independent
+    in case topSorted of
+            Nothing     -> Left (LoopDetected localPending)
+            Just sorted ->
+                let newSchedule = schedule { _ssUnsentNursery = dependent }
+                in Right (sorted, purgeSlot currentSlot newSchedule)
```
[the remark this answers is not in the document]
adinapoli-iohk: I think I didn't appreciate the full power behind your idea -- I'm happy to discuss it further, I think there surely be something I'm missing here.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
adinapoli-iohk: I am a bit on the fence -- although I agree that this is just appending (so maybe `consSchedule` or `snocSchedule` might be a more suitable name) the interpretation that this is "overriding" a `Schedule` is also correct -- we are, in a sense, hijacking the Schedule (which handling would be normally internal or part of the resubmission function logic
).


No strong feelings, I'm open to both solutions.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
adinapoli-iohk: Fair point -- what if we move `constantRetry` to the test module and we leave only the exponential backoff available? I also agree `constantRetry` is a bit too powerful and I suppose having an `n` parameter won't stop the programmer to have it set to `1`, defeating the purpose. What do you think?

adinapoli-iohk: (Alternatively we could have something like a `bunnyHopRetry` policy which would retry every other slot, if that's marginally better, but I doubt 😉 )

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
adinapoli-iohk: yessir, will do!

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
adinapoli-iohk: > If sending fails, do we really want to give up on it?

We don't, I guess, and my intention with that pattern match was to capture that spirit:

```haskell
                     case result of
                         False -> return (evicted, Schedule (reschedule newSlot) nursery)
                         True  -> return (evicted, acc)

```

I thought it would have been a good choice to simply "drop" the transaction if submission succeeded or to re-add it to the scheduler (according to the retry policy) in case we fail sending. What did I miss?

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
edsko: We can't force callers to be sensible, but we can at least nudge them in the right direction :) I think with an argument they at least need to make a conscious decision to retry every single slot.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
edsko: Ah, then it's the _other_ case I should have asked about. We should never drop transactions here, should we? If they succeeded in transmission that doesn't mean that they'll make it into the blockchain. It's the responsibility of the wallet to inform us when they do.

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
adinapoli-iohk: Yea, I see what you mean, and I suspect we might need to do a bit of rubber duck debugging to see what's make most sense -- I originally would have though that checking if we were meant to resubmit immediately would have avoided work, but I'm not concerned that, in case of particular retry policy, we might even end up in a situation where transactions _never attempted_ would be erroneously dismissed by a premature check.. 🤔

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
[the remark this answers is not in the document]
adinapoli-iohk: A middle ground option might be naming them (canSend, cannotSend) ..

wallet-new/src/Cardano/Wallet/Kernel/Submission.hs
adinapoli-iohk: An alternative might be to parametrise our function with a `(Schedule -> Schedule)` override function.

PR #2956, opened 2018-05-18
[CBR-199][CBR-276] Sending raw data to network using diffusion layer 

coot:
## Description

Speed up passing blocks from db layer to the network.  Avoid deserialization followed by serialization.
@avieth I added `getSerializedBlock` in place of `getBlock` in the `Logic` layer.  It seems now we only need the former.

The PR includes two property tests:
* encoding of `MsgSerializedBlock` is the same as `MsgBlock`
* deserialization is (sort of) left inverse of serialization of `MsgSerializedBlock` 

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-199

## Type of change

## Developer checklist

## Testing checklist

- [] All new and existing tests passed.


lib/src/Pos/Binary/Communication.hs
```diff
@@ -66,6 +69,14 @@ instance Bi MsgBlock where
             1 -> MsgNoBlock <$> decode
             t -> cborError $ "MsgBlock wrong tag: " <> show t
 
+-- Serialize `MsgSerializedBlock` with the property
+-- ```
+-- serialize (MsgBlock b) = serializeMsgSerializedBlock (MsgSerializedBlock $ serialize b)
+-- ```
+serializeMsgSerializedBlock :: MsgSerializedBlock -> BS.ByteString
+serializeMsgSerializedBlock (MsgSerializedBlock b) = "\x82\x0" <> unSerialized b
+serializeMsgSerializedBlock (MsgNoSerializedBlock t) = serialize' (MsgNoBlock t)
```
avieth: Let's use a lazy bytestring. The `<>` in the first alternative won't cause a copy of the whole block. `serialize (MsgNoBlock t)` will make the second one well-typed.

lib/src/Pos/Diffusion/Full/Block.hs
```diff
@@ -454,8 +456,10 @@ handleGetBlocks logTrace logic recoveryHeadersMessage oq = listenerConv logTrace
                      " blocks to "%build%" one-by-one")
                     (length hashes) nodeId )
                 for_ hashes $ \hHash ->
-                    getBlock logic hHash >>= \case
-                        Just b -> send conv $ MsgBlock b
+                    getSerializedBlock logic hHash >>= \case
+                        -- TODO: we should get lazy bytestring from the db layer to
+                        -- stream the bytes to the network
```
avieth: Yes indeed we should. Would that be a huge change? Maybe we should make a ticket for this.

avieth:
https://iohk.myjetbrains.com/youtrack/issue/CBR-276

lib/src/Pos/Diffusion/Full/Block.hs
coot: I am not sure, maybe not, let me check.

avieth: On second thought, I think it's fine that rocksdb gives and takes strict bytestrings (it has to take them, since the write needs one pointer). Here we can just use lazy bytestrings to construct the `MsgSerializedBlock` encoding, and avoid one copy.

coot: For `RocksDB` it's true, but for `MonadDB` which writes to disc it may be fine to take lazy `ByteString`.  The only thing is that it may limit future options what to use for the block/undo storage layer.

avieth: Yeah I think as a rule it's better to take lazy bytestrings, since `fromStrict` is cheap but `toStrict` is not.

Don't worry about limiting future options, we can always change `MonadDB`. Nothing is sacred.

coot: Ok, so I'll update the `MonadDB`  to use lazy `ByteString`, anyway future option should be lazy ;)

coot:
Build is passing :), it's ready for review again after adding lazy bytestings to `MonadDB` part that handles block and undos.

block/src/Pos/Block/Network/Types.hs
```diff
@@ -71,3 +73,9 @@ data MsgBlock
     = MsgBlock Block
     | MsgNoBlock Text
     deriving (Eq, Show, Generic)
+
+-- | 'EncodedBlock' message
```
avieth: Should be `'SerializedBlock'`?

coot: Indeed it should be `'SerializedBlock'`.

coot:
Thanks @erikd!

PR #2982, opened 2018-05-23
[CBR-165] Bug Wallet Balance

KtorZ:
## Description

Some users reported weird money movement on their wallets. It appears that they were using a wallet with a same Id that can be created or restored by submitting an empty mnemonic backup phrase.

This PR prevent this and introduce a couple of tests that were written while trying to track down the issue. Note that I am still unsure about the complete impact on wallet that have already been created or restored using an empty backup phrase. They should continue to operate as normal but won't be recoverable anymore. We probably want to warn all those users about this and tell them to stop using the wallets anyway. 

## Linked issue

[[CBR-165]](https://iohk.myjetbrains.com/youtrack/issue/CO-256)

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet/src/Pos/Util/Mnemonics.hs
+++ b/wallet/src/Pos/Util/Mnemonics.hs
@@ -7,6 +7,7 @@ module Pos.Util.Mnemonics
          -- * Data types
          Entropy
        , Mnemonic
+       , defMnemonic
        , Seed
 
          -- * Entropy encoding and decoding
@@ -46,6 +47,12 @@ integerToBS i
     f 0 = Nothing
     f x = Just (fromInteger x :: Word8, x `shiftR` 8)
 
+-- | A default mnemonic that can be advertised as an example. Rejected by
+-- fromMnemonic.
+defMnemonic :: Mnemonic
+defMnemonic =
+    "squirrel material silly twice direct slush pistol razor become junk kingdom flee"
+
 -- | Provide intial entropy as a 'ByteString' of length multiple of 4 bytes.
 -- Output a mnemonic sentence.
 toMnemonic :: Entropy -> Either String Mnemonic
@@ -69,6 +76,8 @@ toMnemonic ent = do
 -- mnemonic.
 fromMnemonic :: Mnemonic -> Either String Entropy
 fromMnemonic ms = do
+    when (ms == defMnemonic) $
+        Left "fromMnemonic: forbidden mnemonic: an example mnemonic has been submitted. Please generate a fresh and private mnemonic from a trusted source."
     when (word_count <= 0) $
         Left "fromMnemonic: empty mnemonic is not allowed"
     when (isJust $ find (not . isAscii) ms) $
--- a/wallet/src/Pos/Util/Mnemonics.hs
+++ b/wallet/src/Pos/Util/Mnemonics.hs
@@ -50,6 +50,8 @@ integerToBS i
 -- Output a mnemonic sentence.
 toMnemonic :: Entropy -> Either String Mnemonic
 toMnemonic ent = do
+    when (length ent <= 0) $
+        Left "toMnemonic: entropy must be a non-zero sequence of bytes"
     when (remainder /= 0) $
         Left "toMnemonic: entropy must be a multiple of 4 bytes"
     when (cs_len > 16) $
@@ -67,6 +69,8 @@ toMnemonic ent = do
 -- mnemonic.
 fromMnemonic :: Mnemonic -> Either String Entropy
 fromMnemonic ms = do
+    when (word_count <= 0) $
+        Left "fromMnemonic: empty mnemonic is not allowed"
     when (isJust $ find (not . isAscii) ms) $
         Left "fromMnemonic: non-ASCII characters not supported"
     when (word_count > 48) $
--- a/wallet/test/Test/Pos/Util/MnemonicsSpec.hs
+++ b/wallet/test/Test/Pos/Util/MnemonicsSpec.hs
@@ -1,0 +1,25 @@
+module Test.Pos.Util.MnemonicsSpec (spec, Entropy(..)) where
+
+import           Universum
+
+import           Data.ByteString.Char8 (pack)
+import           Test.Hspec (Spec, describe, it)
+import           Test.Hspec.QuickCheck (modifyMaxSuccess)
+import           Test.QuickCheck (Arbitrary (..), property)
+import           Test.QuickCheck.Gen (oneof, vectorOf)
+
+import           Pos.Util.Mnemonics (fromMnemonic, toMnemonic)
+
+spec :: Spec
+spec = describe "Pos.Util.Mnemonics" $ modifyMaxSuccess (const 10000) $
+    it "toMnemonic >=> fromMnemonic = Right" $ property $
+        \(Entropy ent) -> (toMnemonic ent >>= fromMnemonic) == Right ent
+
+
+newtype Entropy = Entropy ByteString deriving (Eq, Show)
+
+-- | Initial seed has to be vector or length multiple of 4 bytes and shorter
+-- than 64 bytes.
+instance Arbitrary Entropy where
+    arbitrary =
+        Entropy . pack <$> oneof [ vectorOf (4 * n) arbitrary | n <- [1..16] ]
--- a/wallet/test/Test/Pos/Util/MnemonicsSpec.hs
+++ b/wallet/test/Test/Pos/Util/MnemonicsSpec.hs
@@ -4,11 +4,10 @@ import           Universum
 
 import           Data.ByteString.Char8 (pack)
 import           Data.Set (Set)
-import           Test.Hspec (Spec, describe, it, shouldSatisfy, xit)
-import           Test.Hspec.QuickCheck (modifyMaxSuccess)
-import           Test.QuickCheck (Arbitrary (..), property)
-import           Test.QuickCheck.Gen (Gen (..), oneof, vectorOf)
-import           Test.QuickCheck.Random (mkQCGen)
+import           Test.Hspec (Spec, it, shouldSatisfy, xit)
+import           Test.Hspec.QuickCheck (modifyMaxSuccess, prop)
+import           Test.QuickCheck (Arbitrary (..), forAll, property)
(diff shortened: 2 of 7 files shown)


wallet/src/Pos/Util/Mnemonics.hs
```diff
@@ -93,7 +97,7 @@ numCS :: Int -> Entropy -> Integer
 numCS len = shiftCS . bsToInteger
   where
     shiftCS = case 8 - len `mod` 8 of
-        0 -> identity
+        8 -> identity
```
KtorZ: cf https://github.com/haskoin/haskoin/commit/f00f09d392ca38bb4b68a65c8515ed70f8986841

wallet/src/Pos/Util/Mnemonics.hs
```diff
@@ -67,6 +69,8 @@ toMnemonic ent = do
 -- mnemonic.
 fromMnemonic :: Mnemonic -> Either String Entropy
 fromMnemonic ms = do
+    when (word_count <= 0) $
```
jmitchell: Would `word_count <= 12` be better? As far as I know, in every context we require at least 12.

KtorZ: We currently accept entropy between 4 and 64 bytes (i.e. 3 and 48 words). The BIP-39 official spec mentions the following:

> The mnemonic must encode entropy in a multiple of 32 bits. With more entropy security is improved but the sentence length increases. We refer to the initial entropy length as ENT. The allowed size of ENT is 128-256 bits. 

So, between 16 and 32 bytes (i.e. 12 and 24 words). I'd rather stick with BIP-39, however, I am afraid of the consequences of changing those boundaries at this stage :/

KtorZ:
I thought of something yesterday. Should we also _hard-code_ (:scream:) the list of mnemonics we advertise in the doc and reject them with a clear error message: "Please do not use example mnemonics; rely on trusted implementations to generate your own private words" ?

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,63 @@
...
+import           Pos.Util.Mnemonics (defMnemonic, fromMnemonic, toMnemonic)
+import           Pos.Wallet.Web.ClientTypes.Functions (encToCId)
+import           Pos.Wallet.Web.ClientTypes.Types (CId)
+
+import qualified Data.Set as Set
+
+
+spec :: Spec
+spec = do
+    describe "Pos.Util.Mnemonics" $ do
```
parsonsmatt: If this test is using the `hspec` autodiscovery, then this describe isn't necessary. `hspec`'s autodiscovery will take the `...Spec` module name, strip off `Spec`, and make a describe for that. So this module would get a line like:

```haskell
describe "Test.Pos.Util.Mnemonics" Test.Pos.Util.MnemonicsSpec.spec
```

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,63 @@
...
+        modifyMaxSuccess (const 10000) $ it "toMnemonic >=> fromMnemonic = Right" $ property $
```
parsonsmatt: `it msg (property some_property)` === `prop msg some_property` :)

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,63 @@
...
+
+                entropyToWalletId :: Entropy -> CId w
+                entropyToWalletId (Entropy ent) = cid
+                  where
+                    Right backupPhrase = (BackupPhrase . words) <$> toMnemonic ent
+                    Right cid          = (encToCId . fst) <$> safeKeysFromPhrase mempty backupPhrase
+
+                inputs = genList 1000000
+            in
+                length (inject entropyToWalletId inputs) == length inputs
```
parsonsmatt: Prefer `pendingWith` to `xit` -- same result, but you get to put a message *why* it is pending.

To keep indentation down, I usually also prefer to write:

```haskell
it "message" $ do
    let foo = bar
    foo `shouldBe` what
```

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,63 @@
...
+            (fromMnemonic "") `shouldSatisfy` isLeft
+
+        it "No empty entropy" $
+            (toMnemonic "") `shouldSatisfy` isLeft
+
+        xit "entropyToWalletId is injective " $
+            let
+                inject :: Ord b => (a -> b) -> [a] -> Set b
+                inject fn xs =
+                    foldl' (\acc x -> Set.insert (fn x) acc) Set.empty xs
```
parsonsmatt: Could be simpler: `Set.fromList . map f`

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,63 @@
...
+
+                genList :: Arbitrary a => Int -> [a]
+                genList n =
+                    unGen (vectorOf n arbitrary) (mkQCGen 14) 14
```
parsonsmatt: hmm why is this not a property? You could write:

```haskell
prop "entorpyToWalletId is injective" 
    $ forAll (vectorOf 10000 arbitrary) 
    $ \inputs ->
        length (inject entropyToWalletId inputs) === length inputs
```

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,63 @@
...
+
+                entropyToWalletId :: Entropy -> CId w
+                entropyToWalletId (Entropy ent) = cid
+                  where
+                    Right backupPhrase = (BackupPhrase . words) <$> toMnemonic ent
+                    Right cid          = (encToCId . fst) <$> safeKeysFromPhrase mempty backupPhrase
```
parsonsmatt: These partial pattern matches are safe due to the invariant on the arbitrary instance for `Entropy`? I think I'd rather see an `either (error ("Invariant failed on Entropy in toMenomic: " <> BS.unpack ent)) id` to get a clearer error message for diagnostics; or even `either (error . show) id`

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
KtorZ: Good one :+1:

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
KtorZ: :no_mouth: ... Indeed!

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
KtorZ: I was a property initially, but I switched back to a simple one-off check. Conceptually, it's a bit of the same; we want to test whether `entropyToWalletId` is injective, so we want to check that the function doesn't map two inputs to a same output. Testing 1000 times on lists of size n<sub>i</sub>, is the same as testing one time on a list of size Σn<sub>i</sub>. 

We aren't really interested in the various shapes of the inputs here, but just want to have a lot of them, ideally, all of them if it was possible.

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
KtorZ: They are safe indeed because the entropy we generate are valid (multiple of 4 bytes, given length, etc ...). Hence they _shouldn't fail_. You have a point however, if they ever do, message won't be much clearer.

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
KtorZ: Nevertheless... Having it as a quickcheck property gives us a progress feedback instead of just, hanging there for a minute.

KtorZ:
Addressed Matt's comment above :tada:

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,64 @@
...
+        fromMnemonic defMnemonic `shouldSatisfy` isLeft
+
+    it "No empty mnemonic" $
+        (fromMnemonic "") `shouldSatisfy` isLeft
+
+    it "No empty entropy" $
+        (toMnemonic "") `shouldSatisfy` isLeft
+
+    modifyMaxSuccess (const 10000) $ prop "toMnemonic >=> fromMnemonic = Right" $
+        \(Entropy ent) -> (toMnemonic ent >>= fromMnemonic) == Right ent
```
parsonsmatt: use `===` to get prettier output :)

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
```diff
@@ -0,0 +1,64 @@
...
+
+    -- Turn xit -> it to run, and go get a looooong coffee.
+    xit "entropyToWalletId is injective, (very long to run, used for investigation)"
```
parsonsmatt: `pendingWith` would allow this comment to be part of the pending test output

KtorZ: Except that we can't use `pendingWith` inside a QuickCheck property. It seems to be designed to go only on "standard" hspec tests ... Unless I am missing something :/

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
KtorZ: :o ! Sweet.

wallet/test/Test/Pos/Util/MnemonicsSpec.hs
parsonsmatt: oh, good point :(

hm, might be able to do

```haskell
it "foo" $ do
  pendingWith "asdf"
  property $ ...
```

but not a big deal really

PR #2983, opened 2018-05-23
[CBR-220] Fees included option (sketch)

parsonsmatt:
## Description

API clients would like to specify a payment amount that is inclusive of fees.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-220

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
@@ -181,12 +181,13 @@ distributeFeesInternal fee outputs
     | otherwise = Just result
   where
     result =
-        map (round . subtractFee) outputs
+        map (round . subtractFee . fromIntegral) outputs
+    fee' = fromIntegral fee
     total :: Rational
     total =
-        sum (map fromIntegral outputs)
+        foldl' (+) 0 (map fromIntegral outputs)
     subtractFee output =
-        fromIntegral output - (fromIntegral fee * (fromIntegral output / total))
+        output - (fee' * (output / total))
 
 -- | This subtraction implementation bottoms out at @0@ instead of underflowing.
 subBottomOut :: Word64 -> Word64 -> Word64
--- a/wallet-new/test/TransactionsSpec.hs
+++ b/wallet-new/test/TransactionsSpec.hs
@@ -1,6 +1,7 @@
 module TransactionsSpec where
 
 import Universum
+import qualified Prelude
 
 import Test.Hspec
 import Test.Hspec.QuickCheck
@@ -20,6 +21,7 @@ spec = do
                 -- nontermination. Or possibly some absolutely massive 'Integer'
                 -- values doing the same.
                 outputs = NE.fromList (NE.take 1000 outputs')
+                count = fromIntegral (length outputs) :: Double
                 distr = distributeFeesInternal fee outputs
              in
                 cover (isJust distr) 75 "Inputs did not satisfy invariants" $
@@ -33,15 +35,22 @@ spec = do
                                 sum (map toInteger distributed)
                             difference =
                                 abs (expected - actual)
+                            msg = Prelude.unlines
+                                [ "Difference was:   " <> show difference
+                                , "Count of outputs: " <> show count
+                                , "Fee / count:      " <>
+                                    show (fromIntegral fee / count)
+                                , "Fee:              " <> show fee
+                                ]
                          in
-                            counterexample ("Difference was: " <> show difference)
-                            $ difference <= 6
-                              .&&.
-                              all (0 <) distributed
-                              -- due to rounding error, we currently observe
-                              -- a maximum difference of 5 Lovelace between what
-                              -- we expect and the actual value. This amount
-                              -- would go to extra fees.
+                            counterexample msg $
+                                difference <= 4
+                                .&&.
+                                all (0 <) distributed
+                                -- due to rounding error, we currently observe
+                                -- a maximum difference of 5 Lovelace between what
+                                -- we expect and the actual value. This amount
+                                -- would go to extra fees.
 
         it "in case of fee that divides to X.5, goes to both" $ do
             let outputs = 21 :| [21]


client/src/Pos/Client/Txp/Util.hs
```diff
@@ -65,7 +65,7 @@ import           Pos.Core (Address, Coin, StakeholderId, TxFeePolicy (..), TxSiz
                            bvdTxFeePolicy, calculateTxSizeLinear, coinToInteger, integerToCoin,
                            isRedeemAddress, mkCoin, protocolMagic, sumCoins, txSizeLinearMinValue,
                            unsafeIntegerToCoin, unsafeSubCoin)
-import           Pos.Core.Configuration (HasConfiguration)
+import           Pos.Core.Configuration (HasConfiguration, HasProtocolMagic)
```
parsonsmatt: I cleaned up some of the excess constraints to make it easier to reuse these functions. Not necessary to roll in.

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -235,7 +235,7 @@ runTxCreator inputSelectionPolicy action = runExceptT $ do
 
 -- | Like 'makePubKeyTx', but allows usage of different signers
 makeMPubKeyTx
-    :: (HasConfiguration)
+    :: (HasProtocolMagic)
```
parsonsmatt: Simpler constraints = easier to reuse <3

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -304,55 +304,50 @@ type InputPickingWay = Utxo -> TxOutputs -> Coin -> Either TxError FlatUtxo
...
-    -- NOTE (adinapoli, kantp) There is another subtle corner case which involves such partitioning; it's now
-    -- in theory (by absurd reasoning) for the `confirmed` set to contain only dust, which would yes involve a
-    -- "high throughput" Tx but also a quite large one, bringing it closely to the "Toil too large" error
-    -- (The same malady the @OptimiseForSecurity@ policy was affected by).
+    -- NOTE (adinapoli, kantp) There is another subtle corner case which
+    -- involves such partitioning; it's now in theory (by absurd reasoning) for
+    -- the `confirmed` set to contain only dust, which would yes involve a "high
+    -- throughput" Tx but also a quite large one, bringing it closely to the
+    -- "Toil too large" error (The same malady the @OptimiseForSecurity@ policy
+    -- was affected by).
```
parsonsmatt: Comment is formatted to be under 80 columns.

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -304,55 +304,50 @@ type InputPickingWay = Utxo -> TxOutputs -> Coin -> Either TxError FlatUtxo
...
-
 plainInputPicker :: PendingAddresses -> InputPickingWay
 plainInputPicker (PendingAddresses pendingAddrs) utxo _outputs moneyToSpent =
-    evalStateT (pickInputs []) (InputPickerState moneyToSpent sortedUnspent)
+    case foldl' pick ([], moneyToSpent) sortedUnspent of
+        (inps, moneyLeft)
+            | moneyLeft == mkCoin 0 ->
+                pure inps
+            | otherwise ->
+                throwError (NotEnoughMoney moneyLeft)
```
parsonsmatt: I refactored this implementation to be simpler. No need for `StateT s (Except e)` when `foldl'` does what you need. I can revert this change if we want; it was mostly an exercise in understanding the code :)

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -448,10 +443,11 @@ prepareTxRawWithPicker inputPicker utxo outputs (TxFee fee) = do
...
             totalTxAmount <- sumTxOuts $ map snd inputsNE
-            let trInputs = map formTxInputs inputsNE
-                trRemainingMoney = totalTxAmount `unsafeSubCoin` moneyToSpent
-            let trOutputs = outputs
-            pure TxRaw {..}
+            pure TxRaw
+                { trInputs = map formTxInputs inputsNE
+                , trRemainingMoney = totalTxAmount `unsafeSubCoin` moneyToSpent
+                , trOutputs = outputs
+                }
```
parsonsmatt: banishing some record wildcards

client/src/Pos/Client/Txp/Util.hs
```diff
@@ -688,8 +684,10 @@ stabilizeTxFee pendingTx linearPolicy utxo outputs = do
     minFee <- fixedToFee (txSizeLinearMinValue linearPolicy)
     mtx <- stabilizeTxFeeDo (False, firstStageAttempts) minFee
     case mtx of
-        Nothing -> throwError FailedToStabilize
-        Just tx -> pure $ tx & \(S.Min (S.Arg _ txRaw)) -> txRaw
+        Nothing ->
+            throwError FailedToStabilize
+        Just (S.Min (S.Arg _ txRaw)) ->
+            pure txRaw
```
parsonsmatt: I found the `tx & \(S.Min (S.Arg _ txRaw))` thing confusing so I made it more ordinary

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -67,6 +67,25 @@ newTransaction submitTx Payment {..} = do
...
+                map subtractAmount (pmtDestinations pmt)
+            }
+        subtractAmount pmtDistribution = pmtDistribution
+            { pdAmount = V1 $
+                (unV1 (pdAmount pmtDistribution))
+                    `Core.unsafeSubCoin`
+                        Core.mkCoin coinPerOutput
+            }
+
+    newTransaction submitTx newPayment
```
parsonsmatt: So, this is a somewhat dumb first pass implementation. Basically, we calculate the estimated fee for the payment, and then subtract that amount from the `pmtDestinations` evenly (eg if there is one destination, then we subtract the entire fee from that destination, if there are two destinations, we subtract half of the fee from each destination, etc).

I'm not convinced it's a great strategy. But I think it *should* work. Some issues:

- A `PaymentDistribution` might be reduced to `0` coins. in this case, we probably want to throw an error, because that's very likely not what they want.
- The user might want to specify a single payment from which the fee comes out of. Embedding the `feesIncluded` into the `PaymentDistribution` object itself would allow that level of control, but would make the logic much trickier.
- We might want to specify different strategies for distributing the fee among the distributions. Right now it's 100% evenly split, though we might want to say "distribute the fee proportionally to each output depending on how large the output is." This isn't that much more difficult to implement logic-wise.

That we have multiple possible strategies suggests that `includeFees` query param should accept the strategy we want to use for distributing included fees. A few fixed strategies will be simpler and easier than embedding the `includedFees` into the payment type itself.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1066,15 +1065,15 @@ data PaymentDistribution = PaymentDistribution {
...
-  declareNamedSchema =
-    genericSchemaDroppingPrefix "pd" (\(--^) props -> props
-      & ("address" --^ "Address to map coins to.")
-      & ("amount"  --^ "Amount of coin to bind, in Lovelace.")
-    )
+    declareNamedSchema =
+        genericSchemaDroppingPrefix "pd" (\(--^) props -> props
+            & ("address" --^ "Address to map coins to.")
+            & ("amount"  --^ "Amount of coin to bind, in Lovelace.")
+        )
```
parsonsmatt: Indents

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
KtorZ: Although it seems more acceptable in the crypto world because of rather _unpredictable fees_, I am bit worried about divisions and rounding issues here.
I'd also find more logical to distribute fees according to the amounts sent to each destination as you suggest in your second point but I am pretty sure that we will find people on both sides. What about leaving this as a choice for the user? 

e.g.
```hs
data FeeHeuristic = Exclusive | InclusiveEvenly | InclusiveProportionally 
```

Then, we can build more heuristic over time if needed, that could also rely on fixed strategies ?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: My bad :no_mouth:
I hadn't yet setup the `.editorconfig` at that time ^^"

client/src/Pos/Client/Txp/Util.hs
KtorZ: I'd be great to have this comment in the commit message :p Would help to understand why this commit during a merge conflict or a refactoring!

KtorZ: My two cents on commits messages: A first line which is concise and give an overview of what the commit does and if needed, one blank line, and as many lines of comments as you want that describe things you can't solely express with code or in-code comment (intention, choices and reasoning). 
It's actually really helpful when digging into old code that is somewhat hard to get. Sometimes, a little note from the author help to understand a change.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: No worries! Everyone's doing it :sweat_smile:

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
parsonsmatt: mmm, division/rounding errors are a good point.

Another potential strategy is to entirely drain the "change output" before subtracting anything from the other outputs. If the change output is entirely drained (eg reduced to 0), then we can remove that output from the transaction and get a lower fee.

We'll also need to be careful of any non-change outputs that are reduced to 0 for this.

The more I'm thinking on this, the more I feel like this shouldn't *actually* submit the transaction, but should instead render a prepared transaction that can later be submitted. Eg: 

> Client: I give you this `Payment` and I want the fees to come out of the `PaymentDistributions` according to this `FeeStrategy`.
> Server: OK, here's the transaction that would result -- these outputs would be reduced by this much. If that is OK, then submit this `Payment` to the ordinary transaction endpoint and it'll do what you want.

KtorZ: I like the confirmation idea, though it's an extra step. This _simple_ tasks sounds more like a story to me and by that I mean, it should go over a proper design and discussion phase, testing, and implementation for there are many strategies we could take here. An informal specification on our approach that can be reviewed / documented would be nice. 

What do you think?

client/src/Pos/Client/Txp/Util.hs
adinapoli-iohk: Thank you! <3

client/src/Pos/Client/Txp/Util.hs
adinapoli-iohk: Yup, I also vote for the commit message scheme outlined by Matthias here.

Writing long and descriptive commit messages is certainly tedious, but it pays out immensely when trying to reason about code later on :)

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
parsonsmatt: I've decided that I want to only allow proportional distribution of fee among outputs; the other strategies that I can imagine are Bad (for reasons that will become evident when I push my latest commit with a ton of docs :sweat_smile: )

client/src/Pos/Client/Txp/Util.hs
parsonsmatt: I like this as well! I've been getting in the habit of "Squash and merge" with my PRs and I'll annotate the commit messages more thoroughly there. Unless you'd rather I not squash+merge :sweat_smile:

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -67,6 +67,134 @@ newTransaction submitTx Payment {..} = do
...
+    | 0 `elem` result = Nothing
+    | otherwise = Just result
+  where
+    result =
+        map (round . subtractFee) outputs
+    total :: Rational
+    total =
+        sum (map fromIntegral outputs)
+    subtractFee output =
+        fromIntegral output - (fromIntegral fee * (fromIntegral output / total))
```
parsonsmatt: This function distributes the fee evenly among the inputs, HOWEVER... with smaller numbers, we get rounding errors.

And, in fact, with any fee where proportion divides out to a multiple of 0.5, we'll overestimate the fee we actually do. The QuickCheck test demonstrates this.

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
```diff
@@ -67,6 +67,134 @@ newTransaction submitTx Payment {..} = do
...
+    | 0 `elem` outputs = Nothing
+    | fromIntegral fee >= total = Nothing
+    | 0 `elem` result = Nothing
+    | otherwise = Just result
+  where
+    result =
+        map (round . subtractFee) outputs
+    total :: Rational
+    total =
+        sum (map fromIntegral outputs)
```
parsonsmatt: there's a huge difference between `fromIntegral (sum outputs)` and `sum (map fromIntegral outputs)`, especially with types that overflow -_-

client/src/Pos/Client/Txp/Util.hs
adinapoli-iohk: > Unless you'd rather I not squash+merge

There are good reasons to do a normal merge vs a squash+merge. I tend to do those on a case-by-case analysis preferring the normal one (thus the need for a clean history) merely because it's what `git-flow` seems to dictate 😀

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Transactions.hs
adinapoli-iohk: @parsonsmatt @KtorZ Sorry, late to the party, had a looong hacking week 😉 

> We'll also need to be careful of any non-change outputs that are reduced to 0 for this.

I would exercise caution (I know this is a sketch, but still nice to know the fuller picture) to change anything which was originally scheduled in a transaction by the input selection mechanism, as this could have impacts on the anonymity of users. In particular, I would defer any change in the inner logic (or perhaps this ticket altogether) up until @edsko formalises the input selection policies as part of the new data layer (see: CBR-226, CBR-241, CBR-251). What I'm concerned about here is that we are stepping on each other toes -- looks like there are lots of valuable ideas from an API perspective in this PR and I think would be nice to see if they can retrofit into Edsko's policies, but at the same time we should make sure what we offer at the API level doesn't contradict directly the new data layer internal implementation. Am I making any sense? 😉

KtorZ: @adinapoli-iohk I got a similar feeling to be honest. This task seems a bit more complicated than it seems, with quite some important choices to make when it comes to selecting inputs. That's definitely not trivial. 
Should we put this task on hold for now and move it under CBR-227 ?

adinapoli-iohk: > This task seems a bit more complicated than it seems, with quite some important choices to make when it comes to selecting inputs. That's definitely not trivial.

Yea, unfortunately this _seems_ easy when it comes to the API side of it, but as it touches a very delicate part of the wallet, I have rapidly realised this was not going to be trivial. Matt will kill me as it's the second time to does work in this area and he has to pause his hacking 😬 

> Should we put this task on hold for now and move it under CBR-227 ?

I don't see nothing wrong with putting this in a user story (if you folks have one) related to "V2 API" or "API future improvements" and have it depends from CBR-227.

Different matter is if we think we cannot deliver this endpoint without having an extra requirement to be added in @edsko 's input selection policy mechanism, at which point it might make sense to consider the backend side of it as part of CBR-227, and the fronted side as per above.

KtorZ: > I don't see nothing wrong with putting this in a user story (if you folks have one) related to "V2 API" or "API future improvements" and have it depends from CBR-227.

Fun fact: this task is currently part of a story about "wrapping up remaining work for **V1**". I guess we were too optimistic about it. That's clearly not something that is part of the _minimal_ V1.

parsonsmatt: This logic is operating at a slightly higher level than input selection. Right now we're just taking a `Payment`, calculating the estimated fee, then reducing the amounts in the `Payment` record. It's possible that this new `Payment` gets a different fee, so I suppose we'd need to iterate that process until it stabilized.

We would not want to take anything from change outputs -- it turns out, that would defeat the point entirely, as the entire intent is to have the fee come from the amount we are *sending* and not the amount we *own*.

I don't feel like it's terribly complicated, just some fiddly bits to minimize the amount of wasted coin when dealing with the rounding. Moving it to a new user story would help get the V1 thing out the door though :)

edsko:
@parsonsmatt ,

> This logic is operating at a slightly higher level than input selection. Right now we're just taking a Payment, calculating the estimated fee, then reducing the amounts in the Payment record. It's possible that this new Payment gets a different fee, so I suppose we'd need to iterate that process until it stabilized.

> We would not want to take anything from change outputs -- it turns out, that would defeat the point entirely, as the entire intent is to have the fee come from the amount we are sending and not the amount we own.

> I don't feel like it's terribly complicated, just some fiddly bits to minimize the amount of wasted coin when dealing with the rounding. Moving it to a new user story would help get the V1 thing out the door though :)

So at the moment, the strategy in this PR as follows:

* Run coin selection (`estimateFee`), get a transaction out of it, compute the fee. 
* Remove that fee from the desired outputs (distributing it evenly).
* Then run coin selection all over again.

There are two issues that you already point out above: outputs can drop below zero, in which case this should fail (if it's recipient pays fees, but the recipient doesn't have enough funds in the transaction to pay those fees, the transaction cannot be generated), and secondly call to coin selection might generate a different kind of transaction since now the sum total is different, and hence this may need to be iterated. 

At the moment the second problem is _kind of_ benign, because the coin selection algorithm is deterministic and will therefore only generate a new transaction if the deduction in change outputs meant that we need fewer inputs; however, that's only because the current coin selection algorithm is bonkers. Once we move to other coin selection algorithms they will not be deterministic anymore and this strategy will not work.

However, the most important reason that this is awkward is that this should not be a post-processing step _after_ coin selection. Coin selection at the moment is selecting coins _on the premise_ that the sender must pay fees. This actually complicates coin selection: it may need to iterate to find a solution. Then _on top of that_ in this PR we now _again_ need to iterate to _change_ that payment again so that the recipient pays fees. This doesn't make too much sense.

If we adjust coin selection _itself_ to support recipient pays fees versus sender pays fees, _then_ you are absolutely right that in that case sender pays fees is a simple post-processing step, and we do almost exactly what you're doing here: we just compute the fee, and subtract it form the outputs. Note however that in this case there is no need to iterate _either_ in the coin selection process _nor_ in the post-processing step. So in this case recipient pays fees indeed is very simple (and easier than sender pays fees).

edsko:
Another problem here that occurred to me as I was trying to fall asleep last night (I know, I have no life :joy:) -- unless I am missing something, but doesn't the above approach mean fees get paid _twice_? You make a tranaction (which will have sender paid fees), then you compute the fees (which the sender already paid), then subtract those fees from the output, and construct another transaction -- in which the sender will make some fees again, and now the recipient also gets smaller outputs... No? If you wanted to do this as a post-processing step you'd need to _add_ the fees back to the change outputs.

All in all, easier to do this when the transaction is created :)

parsonsmatt:
I'm not sure that's how the logic flows. With the fee estimation endpoint, we don't get a *transaction* back. We just get a coin-amount of the estimated fee.

Likewise, we don't submit a transaction, we submit a `Payment` that is *converted* into a `Transaction` via the input selection policy, etc.

edsko:
> I'm not sure that's how the logic flows. With the fee estimation endpoint, we don't get a transaction back. We just get a coin-amount of the estimated fee.

> Likewise, we don't submit a transaction, we submit a Payment that is converted into a Transaction via the input selection policy, etc.

Right, and that's precisely the problem. The logic above reads something like

```
let newPayment = ... twiddle with outputs, reducing the m...
newTransaction submitTx newPayment
```

Unless I'm missing the point (entirely possible) -- but `newTransaction` will not have any information that these outputs are supposed to have been adjusted for fees. As far as it is concerned, this is just another payment request, albeit for slightly smaller outputs. So it will now select inputs from the UTxO to cover that payment request, and it will construct a transaction in which the sender pays the transaction fee. So now the sender is paying fees _and_ the recipient is receiving less.

PR #2984, opened 2018-05-23
Add core support for protocol level MultiSig

vincenthz:
## Description

* Extend AddressSpendingData to support having multiple parties computing
  an address using the hash of the participants' public key.
* Extend TxWitness to take a list of MultisigWitness which are either the public key
  hash if not participating or as a signatory the public key and it's associated signature.

The multisig signature is valid if the number of signatures is equal or above
to the expected threshold. if any present signature is also invalid, the
overall multisig signature is considered invalid too even if the expected
number of valid signatures is met.

While the order of participants is not enforced, it is necessary to specify the
same order between the witnesses and how the address has been generated, since
the order is required to compute the expected address.

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/core/src/Pos/Arbitrary/Core.hs
+++ b/core/src/Pos/Arbitrary/Core.hs
@@ -195,8 +195,14 @@ instance Arbitrary Types.AddrType where
             [ pure Types.ATPubKey
             , pure Types.ATScript
             , pure Types.ATRedeem
-            , Types.ATUnknown <$> choose (3, maxBound)
+            , pure Types.ATMultisig
+            , Types.ATUnknown <$> choose (4, maxBound)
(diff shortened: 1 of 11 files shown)


erikd:
The code here looks fine, but shouldn't there be tests as well?

vincenthz:
@erikd yes it should ideally, but I couldn't find any existing tests for AddressSpendingData/Witnessing

erikd:
None of this code has enough tests, and I'd really like to avoid adding still more code without tests. As it is, my team is having to write tests in order to some of the refactorings this code so badly needs.

jcmincke:
Does not seem relevant at the moment

PR #2999, opened 2018-05-27
Hard coded fee estimates

matt-noonan:
## Description

This replaces the fee estimate in the wallet unit tests with a formula derived from the actual serialized size of a `Tx`.

There are a number of supporting changes made to the tests, stemming from two things:
1. The new estimator needs to know the `Coin` amounts in each tx output, not just the total number of tx outputs.
2. The linear size -> fee function has coefficients that are derived from the current configuration, leading to a `HasGenesisBlockVersionData` constraint in a few places.

## Linked issue

CBR-262

## Type of change

- [X] 🔨 New or improved tests for existing code

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/test/unit/Test/Infrastructure/Generator.hs
+++ b/wallet-new/test/unit/Test/Infrastructure/Generator.hs
@@ -131,7 +131,24 @@ cardanoModel boot = GeneratorModel {
     , gmMaxNumOurs    = 5
     }
 
--- | Estimate the size of a transaction, in bytes.
+{-| Estimate the size of a transaction, in bytes.
+
+     The magic numbers appearing in the formula have the following origins:
+
+       5 = 1 + 2 + 2, where 1 = tag for Tx type, and 2 each to delimit the
+           TxIn and TxOut lists.
+
+      42 = 2 + 1 + 34 + 5, where 2 = tag for TxIn ctor, 1 = tag for pair,
+           34 = size of encoded Blake2b_256 Tx hash, 5 = max size of encoded
+           CRC32 (range is 1..5 bytes, average size is just under 5 bytes).
+
+      11 = 2 + 2 + 2 + 5, where the 2s are: tag for TxOut ctor, tag for Address
+           ctor, and delimiters for encoded address. 5 = max size of CRC32.
+
+      32 = 1 + 30 + 1, where the first 1 is a tag for a tuple length, the
+           second 1 is the encoded address type. 30 = size of Blake2b_224
+           hash of Address'.
+-}
 estimateSize :: Int      -- ^ Average size of @Attributes AddrAttributes@.
              -> Int      -- ^ Size of transaction's @Attributes ()@.
              -> Int      -- ^ Number of inputs to the transaction.
--- a/wallet-new/test/unit/Test/Infrastructure/Generator.hs
+++ b/wallet-new/test/unit/Test/Infrastructure/Generator.hs
@@ -141,23 +141,18 @@ estimateSize saa sta ins outs
     = fromBytes . fromIntegral $
       5
     + 42 * ins
-    + (11 + size (32 + (fromIntegral saa))) * length outs
-    + sum (map coinSize outs)
+    + (11 + listSize (32 + (fromIntegral saa))) * length outs
+    + sum (map intSize outs)
     + fromIntegral sta
   where
-    coinSize s =
+    intSize s =
         if | s <= 0x17       -> 1
            | s <= 0xff       -> 2
            | s <= 0xffff     -> 3
            | s <= 0xffffffff -> 5
-           | otherwise      -> 9
+           | otherwise       -> 9
 
-    size s =
-        if | s <= 0x17       -> 1 + s
-           | s <= 0xff       -> 2 + s
-           | s <= 0xffff     -> 3 + s
-           | s <= 0xffffffff -> 5 + s
-           | otherwise      -> 9 + s
+    listSize s = s + intSize s
 
 -- | Estimate the fee for a transaction that has @ins@ inputs
 --   and @length outs@ outputs. The @outs@ lists holds the coin value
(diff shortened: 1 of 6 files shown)


wallet-new/test/unit/Test/Infrastructure/Genesis.hs
```diff
@@ -30,10 +33,13 @@ data GenesisValues h = GenesisValues {
 
       -- | Hash of the bootstrap transaction
     , hashBoot :: h (Transaction h Addr)
+
+      -- | Fee policy
+    , txFee :: Int -> [Value] -> Value
```
matt-noonan: Adding the fee policy here helps avoid many other `HasGenesisBlockVersionData` constraints.

wallet-new/test/unit/Test/Spec/Translation.hs
```diff
@@ -147,21 +144,29 @@ example1 GenesisValues{..} = OldestFirst [OldestFirst [t3, t4]]
...
                trFresh = 0
              , trFee   = fee4
              , trHash  = 4
              , trIns   = Set.fromList [ Input (hash t3) 1 ]
-             , trOuts  = [ Output r2 (initR0 - 1000 - fee3 - fee4) ]
+             , trOuts  = [ Output r2 (initR0 - 1000 - trFee t3 - fee4) ]
              , trExtra = ["t4"]
              }
 
+-- | Use this to construct a transaction where parts of the transaction depend on
```
matt-noonan: Several of the tests had transactions where the txos depended somehow on that transaction's fee. But the fee estimator needs to know the value of each txo, because the coin values have a variable-length encoding. Circular!

This function takes a "transaction with a hole for the fee", and turns it into a normal transaction. Will it converge? Who knows! (I suspect it can be made to cycle, but I haven't proven it). I'm pretty sure it can't grow without bound, so maybe it is enough to say "iterate until a fixpoint or the next fee estimate is smaller than the current one".

Anyway, I haven't actually witnessed this ever failing to converge.

wallet-new/test/unit/Test/Infrastructure/Generator.hs
```diff
@@ -124,8 +131,43 @@ cardanoModel boot = GeneratorModel {
...
+estimateSize :: Int      -- ^ Average size of @Attributes AddrAttributes@.
+             -> Int      -- ^ Size of transaction's @Attributes ()@.
+             -> Int      -- ^ Number of inputs to the transaction.
+             -> [Value]  -- ^ Coin value of each output to the transaction.
+             -> Byte     -- ^ Estimated size of the resulting transaction.
+estimateSize saa sta ins outs
+    = fromBytes . fromIntegral $
+      5
+    + 42 * ins
+    + (11 + listSize (32 + (fromIntegral saa))) * length outs
```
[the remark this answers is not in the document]
matt-noonan: I could do this, though I should warn that e.g. the 5 is actually a 2 + 2 + 1 with each term coming from a different part of the transaction. So it might end up a little messier.

The only values that are averages are the saa and sta parameters, plus 5 bytes per address for a CRC32 (it's actually encoded as variable length, but in practice it will almost always be 5 bytes)

wallet-new/test/unit/Test/Spec/Translation.hs
matt-noonan: That seems like a good idea, updated.

edsko: > Several of the tests had transactions where the txos depended somehow on that transaction's fee. But the fee estimator needs to know the value of each txo, because the coin values have a variable-length encoding. Circular!

Yeah, this is what the `PreChain` was trying to convey. I thought that with the fee estimation function this problem is circumvented -- but I guess the reason that it's not is that the fee estimation now depends on the _size_ of the outputs, rather than merely their number. I'd still prefer to keep out this fixpoint though. It seems to me that it would be possible to define an upper bound fee calculation using `maxCoin` for each output -- would that work? Then we don't need the fixpoint anymore and we can keep the examples simple.

wallet-new/test/unit/Test/Spec/Translation.hs
```diff
@@ -147,21 +144,31 @@ example1 GenesisValues{..} = OldestFirst [OldestFirst [t3, t4]]
...
+    computedFee tx = getFee (length $ trIns tx) (map outVal $ trOuts tx)
+    step (tx, idx) = (make $ computedFee tx, idx + 1 :: Int)
+
 {-------------------------------------------------------------------------------
   Verify chain
 -------------------------------------------------------------------------------}
 
-intAndVerifyPure :: (GenesisValues GivenHash -> Chain GivenHash Addr)
+intAndVerifyPure :: HasGenesisBlockVersionData
+                 => (GenesisValues GivenHash -> Chain GivenHash Addr)
```
edsko: Given that you added the fee policy to `GenesisValues`, is this `HasGenesisBlockVersionData` will required? If so, why?

wallet-new/test/unit/UTxO/Generator.hs
```diff
@@ -259,16 +259,17 @@ genTransaction GenTrParams{..} removeUsedInputs makeOutputsAvailable notThese =
         nextHash   <- gtsNextHash <<+= 1
         numOutputs <- lift $ choose (1, gtpMaxNumOutputs)
 
-        let fee     = gtpEstimateFee numInputs numOutputs
-            inValue = sum (map (outVal . snd) inputs)
+        let inValue = sum (map (outVal . snd) inputs)
+            gos = initOutState inValue
 
-        if inValue <= fee
+        (outputs, gos') <- lift $ (`runStateT` gos) $
```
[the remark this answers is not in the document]
edsko: I don't mind necessarily, but if it would make the code significantly simpler, can use the same upper bound suggestion as above? But if there was no fixpoint calculation necessary here, it doesn't really matter so much.

wallet-new/test/unit/Test/Spec/Translation.hs
matt-noonan: It's due to the use of `genesisValues`, which constructs the `GenesisValues` value. It does so by setting the `txFee` field to `estimateCardanoFee`, and `estimateCardanoFee` uses the `GenesisBlockVersionData` part of the configuration to get the correct parameters for the linear fee computation.

PR #3011, opened 2018-05-29
Merge release/1.2.0 into master

jcmincke:
Merge release/1.2.0 into master


jcmincke:
Please, have a look at pkgs/default.nix and check it

jmitchell:
DevOps' records indicate rev 55dce87 was used in the recent Daedalus release, and it looks like all subsequent commits since are for the changelog. LGTM.

adinapoli-iohk:
This is probably a highly-unscientific method, but considering the size of the PR I have resorted to a shortcut, i.e. comparing `release/1.2.0` with `jcm/merge-in-master/release/1.2.0` by visiting `https://github.com/input-output-hk/cardano-sl/compare/release/1.2.0...jcm/merge-in-master/release/1.2.0` and it seems to me we should be in business:

screen shot 2018-05-29 at 17 05 48

I can see how there are 4 commits missing, but I suspect those are just operational commits from devops and don't add any extra features. The only one I wasn't sure is the top one (the one which was compressing the logs etc) but I think this commit was superseded by a more appropriate fix -- @avieth and @jmitchell should be in the position to confirm, I believe.

PR #3031, opened 2018-06-01
Fix Latex compilation (on some systems)

georgeee:
Fixed Latex compilation on my system (minor fixes).

@erikd 
@aspiwack

### DIFF
--- a/docs/rules/rules.tex
+++ b/docs/rules/rules.tex
@@ -71,7 +71,7 @@
 
 \newcommand{\kronecker}[2]{[#1 = #2]}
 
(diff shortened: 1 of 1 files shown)


georgeee:
I used `make` on my Arch installation and it didn't work well unless I put `{}` in all places which use `_` and many symbols following.

aspiwack:
It's a bit hard to test on my side. I can pop a bunch of dockers until I find one where I can reproduce.

But I can't accept as-is since some of the fixes are wrong. I'll see what I can do.

jcmincke:
Closing: no clear conclusion

PR #3043, opened 2018-06-04
[CBR-289] Review mnemonic implementation

KtorZ:
## Description

This is actually an implementation of BIP-39. The current implementation is inspired from Haskoin and is rather hard to grasp. This refactor merges together `Pos.Util.BackupPhrase` and `Pos.Util.Mnemonic`. The entire implementation has been reviewed to rely on smart constructors and maintain invariant this way. 

Quite some tests have been added, and I am open to any idea of things that could also be tested and that aren't currently included. 

> Note
> At the moment, the PR still contains the old code, moved in a test folder and used for comparison 
> with Quickcheck. This is to make sure that the new code behaves exactly as the old one
> since it touches to a quite sensitive part of the code. I am not sure if this has to stay or can be removed if we all get convinced that the behavior has been preserved (keeping only the relevant tests).

## Linked issue

[[CBR-289]](https://iohk.myjetbrains.com/youtrack/issue/CBR-289)

## Type of change

## Developer checklist

## Testing checklist

## QA Steps

- We need to make sure that once this get merged, users with an existing backup phrase can still restore their wallet (and it has to be the same wallet). Said differently, changing the implementation shouldn't have changed the behavior.

- It shouldn't be possible to use the example mnemonic advertised in the API doc anymore

- Paper wallet already emitted should also not be affected and we must verify they can still be redeemed as expected.


wallet/src/Pos/Wallet/Web/Account.hs
```diff
@@ -95,16 +96,14 @@ getSKByAddressPure secrets scp passphrase addrMeta = do
...
+    -> Mnemonic
     -> m EncryptedSecretKey
-genSaveRootKey passphrase ph = do
-    sk <- either keyFromPhraseFailed (pure . fst) $
-        safeKeysFromPhrase passphrase ph
-    addSecretKey sk
-    return sk
-  where
-    keyFromPhraseFailed msg =
-        throwM . RequestError $ "Key creation from phrase failed: " <> msg
```
akegalj: This will potentially break Deadelus if it depends on parsing this error to handle UI.

First iteration of Daedalus parsed string error msgs to figure out what went wrong, I am not sure was it changed recently. 

EDIT: quick search in deadalus repo didn't reveal anything so I guess this error is handled internally by our backend (or msg is changed - or its not handled). Anyway, looks ok to me

wallet/src/Pos/Wallet/Web/Methods/Redeem.hs
```diff
@@ -56,19 +55,17 @@ redeemAdaPaperVend
...
-        $ deriveAesKeyBS <$> toSeed pvBackupPhrase
+    let aesKey = AesKey (mnemonicToAESKey pvBackupPhrase)
     seedDecBs <- either decryptionFailed pure
         $ aesDecrypt seedEncBs aesKey
     redeemAdaInternal submitTx passphrase pvWalletId seedDecBs
   where
     invalidBase58 =
         throwM . RequestError $ "Seed is invalid base58 string: " <> pvSeed
-    invalidMnemonic e =
-        throwM . RequestError $ "Invalid mnemonic: " <> toText e
```
akegalj: hm, this is potentially a regression.

Previous implementation would rise an error if client submitted backup phrase with illegal number of mnemonics. This rised an error and UI reacted on that error. I am not sure this is how it is done now though.

the old implementation rised an error if `fromMnemonic` returned left and your new implementation will create aesKey and then fail on seed decryption with a different error.  Not sure will this affect UI (or is it some sort of regression)

Running acceptance tests from Daedalus might be useful here.

wallet/src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -857,7 +857,8 @@ deriveSafeCopySimple 0 'base ''WebTypes.CHash
 deriveSafeCopySimple 0 'base ''WebTypes.CId
 deriveSafeCopySimple 0 'base ''WebTypes.Wal
 deriveSafeCopySimple 0 'base ''WebTypes.Addr
-deriveSafeCopySimple 0 'base ''BackupPhrase
```
akegalj: hm, why do we store backup phrase at all :/ I doesn't make sense to keep this peace of info on the disk. (will check later why is this so)

EDIT: this was a mistake and is not used in acid (and it doesn't make sense to be used)

wallet/test/Test/Pos/Util/MnemonicOld.hs
```diff
@@ -106,7 +106,7 @@ numCS :: Int -> Entropy -> Integer
 numCS len = shiftCS . bsToInteger
   where
     shiftCS = case 8 - len `mod` 8 of
-        8 -> identity
+        0 -> identity
```
akegalj: hm, I see this change here as well https://github.com/haskoin/haskoin/blob/690c7ae8619f8f478e26cff08ada7bc24f758e8d/haskoin-core/src/Network/Haskoin/Crypto/Mnemonic.hs#L94 . Not sure how this affects things exactly

EDIT: I see https://github.com/haskoin/haskoin/blob/690c7ae8619f8f478e26cff08ada7bc24f758e8d/haskoin-core/src/Network/Haskoin/Crypto/Mnemonic.hs#L94 . All good

wallet/src/Pos/Wallet/Web/State/Storage.hs
```diff
@@ -857,7 +857,8 @@ deriveSafeCopySimple 0 'base ''WebTypes.CHash
...
+deriveSafeCopySimple 0 'base ''Entropy
```
akegalj: this can/should be removed (both Entropy and Mnemonic). See comment on the left

wallet/src/Pos/Wallet/Aeson/ClientTypes.hs
```diff
@@ -38,7 +37,6 @@ deriveJSON defaultOptions ''CWalletInit
 deriveJSON defaultOptions ''CPaperVendWalletRedeem
 deriveJSON defaultOptions ''CTxMeta
 deriveJSON defaultOptions ''CProfile
-deriveJSON defaultOptions ''BackupPhrase
```
akegalj: I believe this would be a breaking change for V0 as json serilization will change to your manually written instances of ToJson/FromJson .

wallet/src/Pos/Wallet/Web/ClientTypes/Types.hs
```diff
@@ -269,7 +269,7 @@ instance Default CAccountMeta where
 -- | Query data for wallet creation
 data CWalletInit = CWalletInit
     { cwInitMeta     :: !CWalletMeta
-    , cwBackupPhrase :: !BackupPhrase
+    , cwBackupPhrase :: !Mnemonic
```
akegalj: same as in L299 - I believe this is a breaking change for V0 users as json serilization format changed.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,664 @@
...
+import           Pos.Util.LogSafe (SecureLog)
+
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Char8 as B8
+import qualified Data.Text as Text
+
+
+-- | A backup-phrase in the form of a non-empty of Mnemonic words
+-- Constructor isn't exposed.
+data Mnemonic = Mnemonic
```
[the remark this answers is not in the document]
akegalj: Note that this might be useful when we add redeem functions to V1 - to distinguish between mnemonics of different length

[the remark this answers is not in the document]
KtorZ: I thought of this. And ideally, I'd even like to have different types for them, and define the conversion functions as type-class. Otherwise, there's nothing really preventing us to use a `Mnemonic12` in a context where we want a `Mnemonic15`. Having the number of words at the type level with a `Nat` could do the trick. I'll spend some time today to see if I can get something to work and if the complexity it brings isn't too prevalent.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,664 @@
...
+instance ToSchema Mnemonic where
+    declareNamedSchema _ = do
+        NamedSchema _ schema <- declareNamedSchema (Proxy @[Text])
+        return $ NamedSchema (Just "Mnemonic") schema
+
+
+-- | Smart-constructor for the Entropy
+mkEntropy :: ByteString -> Either Text Entropy
+mkEntropy bytes = do
+    when (BS.length bytes <= 0) $
```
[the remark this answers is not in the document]
KtorZ: I totally agree with you and... yes.. we diverge from the official spec quite a lot (especially when it comes to generating the seed). 
For now, I kept the behavior the same to avoid any issue and not introduce regression although I believe the range could be reduced indeed.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,664 @@
...
+        Left "Forbidden Mnemonic: an example Mnemonic has been submitted. Please generate a fresh and private Mnemonic from a trusted source"
+    indices <- findIndices enWordList wordsm
+    let bits = concatMap word11ToBits indices
+    let (entropyBits, checksum) = splitAt ((length bits `quot` 33) * 32) bits
+    entropy <- mkEntropy (bitsToBytes entropyBits)
+    when (checksum /= calcChecksum entropy) $
+        Left "mnemonic checksum failed: cannot convert mnemonic to entropy"
+    pure $ Mnemonic
+        { getEntropy = entropy
+        , getWords   = zip indices wordsm
```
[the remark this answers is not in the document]
KtorZ: Those are hidden from outside of the module, I put them in the Mnemonic object directly to avoid having to re-compute them when doing conversions. Since Mnemonic can only be created using the public functions of the module, it's safe (no one has access to `getWords` or `indices` in the end).

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,664 @@
...
+entropyToByteString :: Entropy -> ByteString
+entropyToByteString (Entropy e) =
+    e
+
+
+-- | Convert a mnemonic to a seed that can be used to initiate a HD wallet.
+-- Note that our current implementation deviates from BIP-39 as:
+--
+--    - We do not use the password to produce the seed
+--    - We rely on a fast blake2b hashing function rather than a slow PKBDF2
```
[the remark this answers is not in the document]
KtorZ: From my discussion with @vincenthz on the matter:

> The bip39 uses a password to actually generate its root key, something we don't do on our side that's the reason to use pbkdf2, since the password is not enforced to be particularly anything good more generally every time you accept a user supplied password to do something cryptographic, to prevent having terrible entropy and really fast brute forcing, you go through a layer of slowhashing (pbkdf2, scrypt, bcrypt, argon2).

So, because we don't actually use the user passphrase to compute the seed, there's _no need_ for slow hashing. The entropy from the mnemonic is already big enough to cope with brute-force.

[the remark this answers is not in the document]
KtorZ: Beside, I don't know why it was done this way.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,664 @@
...
+-- the Pos.Binary module. And, when creating an AESKey seed in which case we
+-- simply pass the `identity` function.
+mnemonicToSeed :: Mnemonic -> ByteString
+mnemonicToSeed =
+    serialize' . blake2b . serialize' . entropyToByteString . mnemonicToEntropy
+
+
+-- | Convert a mnemonic to a seed AesKey. Almost identical to @MnemonictoSeed@
+-- minus the extra serialization.
+mnemonicToAESKey :: Mnemonic -> ByteString
```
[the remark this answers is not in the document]
KtorZ: Well... my two cents is that they _should_ actually be the same function, used with the same purpose :| ... Though, we ended up computing the seed in two different ways in two different part of the code and, dis-tangling the thing yielded these two functions that are nearly identical (the `serialize'` calls away). 

Though, it might be valuable to directly return an `AESKey` here. I'll see whether we can do that without tightening modules too much (Maybe a type-class).

wallet/src/Pos/Wallet/Web/Methods/Redeem.hs
KtorZ: Note that the new implementation still does mnemonic validation. Actually, it's not possible to create a mnemonic that is invalid because the validation is done via the smart constructor. Therefore, if you hold a `Mnemonic` type _you know_ it's valid. That's also why the conversion functions are now pure and can't fail. 

You are raising a very valid point however. I didn't thought of Daedalus "pattern matching" on the error messages. That's probably something we want QA to be involved with. I'll ask.

wallet/test/Test/Pos/Util/MnemonicOld.hs
KtorZ: See https://github.com/input-output-hk/cardano-sl/pull/3013

wallet/src/Pos/Wallet/Web/State/Storage.hs
KtorZ: Will do! Thanks for investigating this.

wallet/src/Pos/Wallet/Aeson/ClientTypes.hs
KtorZ: It _shouldn't_ since the manually written one preserve the same form. Only difference is that is will now fail to parse invalid mnemonic, getting an early error rather than holding a potentially wrong list of words that only explodes when we start using it. 
For any valid inputs however, the structure of the `ToJSON` and `FromJSON` instances have been preserved (a list of strings).

wallet/src/Pos/Wallet/Web/ClientTypes/Types.hs
KtorZ: See above comment, this shouldn't matter in the end.

wallet/src/Pos/Wallet/Web/Methods/Redeem.hs
akegalj: right, I have missed validation at json deserilization. My bad (althought we should still check for error msg format)

vantuz-subhuman:
Hi! Can someone, please, clear one thing for me - do I get it right that changes in mnemonics are **completely** backward-compatible? So old mnemonics can be restored in new version, and new mnemonics can be restored in an old Daedalus version and will give **the same** wallet in the result?

Would really appreciate this clarification, thank you :)

KtorZ:
@vantuz-subhuman 
Yes, the whole point of this change is to review the implementation to get clearer code without changing anything to the current behavior. That's why we currently compare side-by-side the old and new implementations, making sure we generate the exact same wallet from a given mnemonic.

vantuz-subhuman:
> That's why we currently compare side-by-side the old and new implementations, making sure we generate the exact same wallet from a given mnemonic.

@KtorZ, this is awesome! Nice clean changes. So the result will be the exact following of the BIP-39, do I get this right?

KtorZ:
@vantuz-subhuman 
Unfortunately not. There are some discrepancy between the current implementation and BIP-39. As much as I would like this to be fully compliant with BIP-39, it conflicts with the "backward-compatible" approach. Applying changes to be fully BIP-39 compliant would invalidate every existing wallets :) .. which is well, not good.

KtorZ:
Addressed most of the comments above:

- Created a new module `Pos.Util.Bits` which contains all the binary conversions, using `newtype` instead of aliases for more type-safety and tested with QuickCheck.

- Differentiated `mnemonicToSeed` and `mnemonicToAesKey` by using an explicit type for `AesKey` (instead of raw bytestring)

- Introduce number of words at the type-level (e.g. `Mnemonic 12` and `Entropy 12`). This makes the code slightly more complex but also removes a whole category of possible error. The main downside (which is incidentally the main benefit of that change) is that we can't interchange mnemonic phrases with different number of words, they are now different types. If really needed at some point, we may simply define a dedicated `ADT` with the supported types. Like:

```hs
data MultiMnemonic 
  = Multi12 (Mnemonic 12)
  | Multi15 (Mnemonic 15)
```

Note also that this last change is to-be-discussed and can easily be removed by discarding the underlying commit.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+    SupportedWords 24 = ()
+    SupportedWords 27 = ()
+    SupportedWords 30 = ()
+    SupportedWords 33 = ()
+    SupportedWords 36 = ()
+    SupportedWords 39 = ()
+    SupportedWords 42 = ()
+    SupportedWords 45 = ()
+    SupportedWords 48 = ()
+    SupportedWords _  = TypeError ('Text "Unsupported number of Mnemonic words")
```
KtorZ: I feel bad about this. _In theory_, to maintain 100% compatibility with the current implementation, we have to support this whole list. However, I don't think we used any mnemonics but with 9 and 12 words. 
@akegalj May you confirm / disconfirm this?

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
+-- | Module providing restoring from backup phrase functionality
+{-# LANGUAGE DataKinds      #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE TypeFamilies   #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
KtorZ: Some of the type-family `SupportedWords` are judged _redundant_ while they're there to enforce we use a supported `Nat`

wallet/src/Pos/Util/Mnemonic.hs
akegalj: right. To my knowledge we only use 9 and 12 word mnemonics. I am not sure about recently developed paper certificates. Maybe @darko-mijic can fill in the gaps.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+    { getEntropy :: Entropy n
+    , getWords   :: [(Word11, Text)]
+    } deriving (Eq, Show, Generic)
+
+-- | Entropy as a non-empty sequence of bytes, multiple of 4 bytes
+newtype Entropy (n :: Nat) = Entropy ByteString deriving (Eq, Show, Generic)
+
+type family SupportedWords (n :: Nat) :: Constraint where
+    SupportedWords 3  = ()
+    SupportedWords 6  = ()
```
akegalj: just to eliminate possible errors, I would rather remove 3 and 6 word mnemonics - as they don't provide enough entropy (to my knowledge) and even bip39 defines it should be 12 or more words (we used 9 for a specific corner use case)

wallet/src/Pos/Util/Bits.hs
```diff
@@ -0,0 +1,117 @@
...
+groupOf :: Int -> [a] -> [[a]]
+groupOf =
+    iter []
+  where
+    iter acc _ [] = acc
+    iter acc size xs =
+        let
+            (h, q) = splitAt size xs
+        in
+            iter (acc ++ [h]) size q
```
parsonsmatt: http://hackage.haskell.org/package/split-0.2.3.3/docs/Data-List-Split.html#v:chunksOf

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+-- Constructor isn't exposed.
+data Mnemonic (n :: Nat) = Mnemonic
+    { getEntropy :: Entropy n
+    , getWords   :: [(Word11, Text)]
+    } deriving (Eq, Show, Generic)
+
+-- | Entropy as a non-empty sequence of bytes, multiple of 4 bytes
+newtype Entropy (n :: Nat) = Entropy ByteString deriving (Eq, Show, Generic)
+
+type family SupportedWords (n :: Nat) :: Constraint where
```
parsonsmatt: Our kind is too big!

```haskell
data WordCount = Nine | Twelve

type family ToNat w where
  ToNat 'Nine = 9
  ToNat 'Twelve = 12

newtype Entropy (n :: WordCount) = ...
```

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+        return $ NamedSchema (Just "Mnemonic") schema
+            & minItems ?~ n
+            & maxItems ?~ n
+
+
+-- | Smart-constructor for the Entropy
+mkEntropy
+    :: forall n. (SupportedWords n, KnownNat n)
+    => ByteString
+    -> Either Text (Entropy n)
```
parsonsmatt: Instead of `Text`, can we make a datatype that represents the failure? Seems like:

```haskell
data EntropyError = EntropyError
  { requiredBytes :: Int
  , providedBytes :: Int
  }
```

oculd cover it

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+-- | Generate Entropy of a given size
+genEntropy
+    :: forall n. (SupportedWords n, KnownNat n)
+    => IO (Entropy n)
+genEntropy =
+    let
+        size =
+            fromIntegral (entropyLength (Proxy :: Proxy (Entropy n))) `quot` 8
+        invariant =
+            either (error . ("Failed to create Entropy:" <>) . show) identity
```
parsonsmatt: We should prefer `throwIO` or `throwM` over `error` in `IO` code. Otherwise the exception can be confusing.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+    in
+        invariant . mkEntropy <$> Crypto.getEntropy size
+
+
+-- | Smart-constructor for the Mnemonic
+mkMnemonic
+    :: forall n. (SupportedWords n, KnownNat n)
+    => [Text]
+    -> Either Text (Mnemonic n)
```
parsonsmatt: Instead of `Text`, let's make an ADT that represents the various failure modes.

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+-- Output a mnemonic sentence.
+entropyToMnemonic
+    :: forall n. (SupportedWords n, KnownNat n)
+    => Entropy n
+    -> Mnemonic n
+entropyToMnemonic entropy@(Entropy bytes) =
+    let
+        checksum = calcChecksum entropy
+        bits     = toBits (BS.unpack bytes) ++ checksum
+        -- Smart-constructors maintain an invariant, fromBits _can't fail_.
```
parsonsmatt: I was going to ask "JUSTIFY THOSE INVARIANTS" but, uh, you did :joy:

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,683 @@
...
+    , "what", "wheat", "wheel", "when", "where", "whip"
+    , "whisper", "wide", "width", "wife", "wild", "will"
+    , "win", "window", "wine", "wing", "wink", "winner"
+    , "winter", "wire", "wisdom", "wise", "wish", "witness"
+    , "wolf", "woman", "wonder", "wood", "wool", "word"
+    , "work", "world", "worry", "worth", "wrap", "wreck"
+    , "wrestle", "wrist", "write", "wrong", "yard", "year"
+    , "yellow", "you", "young", "youth", "zebra", "zero"
+    , "zone", "zoo"
+    ]
```
parsonsmatt: Since we're indexing into this, we might want to make it a `Vector` or `IntMap` instead of a list.

wallet/src/Pos/Wallet/Aeson/ClientTypes.hs
parsonsmatt: To provide evidence of safety, would you be willing to write a few golden tests?

wallet/src/Pos/Wallet/Web/Account.hs
parsonsmatt: We should not expect that users are handling stringly-typed errors.

wallet/test/Test/Pos/Util/BitsSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+        \(ws :: [Word8]) -> (length . toBits) ws == ((8 *) . length) ws
+
+    prop "(Word11) length . toBits == (11 *) . length" $
+        \(ws :: [Word11]) -> (length . toBits) ws == ((11 *) . length) ws
+
+    it "toBits (42 :: Word8) = [0,0,1,0,1,0,1,0]" $
+        toBits ([42 :: Word8]) `shouldBe` [zero, zero, one, zero, one, zero, one, zero]
+
+    it "toBits (14 :: Word11) = [0,0,0,0,0,0,0,1,1,1,0]" $
+        toBits ([14 :: Word11]) `shouldBe` [zero, zero, zero, zero, zero, zero, zero, one, one, one, zero]
```
parsonsmatt: Nice

wallet/test/Test/Pos/Util/BitsSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+instance Arbitrary Word11 where
+    arbitrary =
+        fromIntegral <$> suchThat (arbitrary @Int) predicate
+      where
+        predicate x =
+            let
+                l = fromIntegral (minBound @Word11)
+                h = fromIntegral (maxBound @Word11)
+            in
+                x >= l && x <= h
```
parsonsmatt: I think this could be `arbitrary = fromIntegral <$> choose (toInteger (minBound @Word11), toInteger (maxBound @Word11))`. Right now it's going to have to pick random ints (eg a 32 bit range) and discard them until it gets one within the range, when you know the range it should be picking anyway.

wallet/src/Pos/Util/Bits.hs
akegalj: ah, I am always surprised needing to discover this function over and over again.

wallet/src/Pos/Util/Mnemonic.hs
KtorZ: Good idea :)

wallet/src/Pos/Util/Mnemonic.hs
KtorZ: Hmmm... Interesting indeed. I kept this one from the previous implementation but there's room for improvement as well.

vincenthz:
ahem ? https://github.com/input-output-hk/cardano-crypto/blob/master/src/Crypto/Encoding/BIP39.hs

KtorZ:
@vincenthz Yes.. I actually found that one during the weekend and I was about to ask you. I believe we can avoid the duplication here ^.^

KtorZ:
@parsonsmatt 
Sooooooo. I've made use of `cardano-crypto` implementation to avoid duplicating code here, simply building a thin wrapper around it with functions that are ready-to-use in cardano-sl. For those, I've defined a `MnemonicErr` error type. We lost some ... granularity along the way though. `cardano-crypto` actually returns `Maybe *` for various functions, which we turn in a generic `FailedToCreate` error. 
Let me know what you think of that.

@vincenthz 
Your review of the new Pos.Util.Mnemonic is highly welcome :)

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,357 @@
...
+                (mnemonicPhraseToMnemonicSentence Dictionary.english)
+                (mnemonicPhrase @12 (toUtf8String <$> wordsm))
+
+            entropy = fromMaybe
+                (error $ show $ UnexpectedMnemonicErr MnemonicErrFailedToCreate)
+                (wordsToEntropy @(EntropySize 12) sentence)
+        in Mnemonic
+            { mnemonicToSentence = sentence
+            , mnemonicToEntropy  = entropy
+            }
```
KtorZ: Note, we cannot use `mkMnemonic` here because it uses this particular `Default` instance. This would create a mutual dependency in a way that won't converge.

wallet/src/Pos/Wallet/Web/Methods/Redeem.hs
KtorZ: After checking with @Nikola on the matter, Daedalus doesn't currently rely on any of the error messages that were triggered by the Mnemonic or BackupPhrase modules. 
I have opened a ticket [DDW-318](https://iohk.myjetbrains.com/youtrack/issue/DDW-318) to actually leverage the messages from the new implementation :+1:

wallet/src/Pos/Util/Mnemonic.hs
```diff
@@ -0,0 +1,357 @@
...
+        (Left MnemonicErrFailedToCreate)
+        Right
+        (wordsToEntropy sentence :: Maybe (Entropy n))
+
+    -- We only advertise a 12-word backup-phrase, and that's the one we want
+    -- to block.
+    let exampleMnemonic =
+            mnemonicSentenceToString Dictionary.english $ mnemonicToSentence (def @(Mnemonic 12))
+
+    when (mnemonicSentenceToString Dictionary.english sentence == exampleMnemonic) $
```
vincenthz: this is a textbook side channel timing measurement. not a very straightforward/useful one, but this could be used to determine some information about the mnemonics.

KtorZ: Good catch. Will use a time-constant comparison here :+1:

KtorZ:
Thanks @vincenthz, I will borrow a couple of test vectors from `cardano-crypto` :)

PR #3072, opened 2018-06-07
[CDEC-328] Implement `deriveIndexedBi`

mhuesch:
## Description

Currently Pos.Binary.Class defines `deriveSimpleBi` that uses TH to generate binary serialization/deserialization functions.

However it only works on data types with accessors which is fine for structs, but really bad for sum types. For sum types, each field accessor function is partial.

This PR implements a new Bi-instance derivation function which uses integer indices instead of field names. This approach doesn't require Sum types to have partial-function field accessors.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CDEC-328

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/binary/test/Test/Pos/Binary/BiSerialize.hs
+++ b/binary/test/Test/Pos/Binary/BiSerialize.hs
@@ -1,0 +1,152 @@
+{-# LANGUAGE DeriveAnyClass  #-}
+{-# LANGUAGE TemplateHaskell #-}
+module Test.Pos.Binary.BiSerialize
+    ( tests
+    ) where
+
+import           Universum
+
+import           Hedgehog (Gen, Property, discover, (===))
+import qualified Hedgehog as H
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+import           Pos.Binary.Class (Cons (..), Field (..), cborError, deriveIndexedBi,
+                                   deriveSimpleBi, serialize')
+
+import qualified Serokell.Util.Base16 as B16
+
+
+--------------------------------------------------------------------------------
+
+data TestSimple
+    = TsInt
+        { unTsInt :: Int }
+    | TsIntList
+        { unTsIntList :: [Int] }
+    | TsChar2
+        { unTsChar2L :: Char
+        , unTsChar2R :: Char }
+    | TsInteger
+        { unTsInteger :: Integer }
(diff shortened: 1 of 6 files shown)


binary/test/test.hs
```diff
@@ -1 +1,20 @@
...
+import qualified Test.Pos.Binary.BiSerialize
+
+main :: IO ()
+main = do
+    hspec spec
+    runTests
+        [ Test.Pos.Binary.BiSerialize.tests
+        ]
+
+runTests :: [IO Bool] -> IO ()
```
mhuesch: This is an obvious duplication of https://github.com/input-output-hk/cardano-sl/blob/3feb70bfb399dfb8d1b2ba2730c044544e91738a/crypto/test/Test/Pos/Crypto/TempHelpers.hs#L60
Since that function is in `crypto`'s helpers, though, I wasn't sure if we should put it into a more general test utility package. Thoughts?

erikd: Actually, the whole of that `TempHelpers.hs` module should be moved to the `binary` package (new name of `Helpers.hs` maybe) and then exposed via the `cardano-sl-binary-test` package where other package can import it.

Up to you whether you cant to make that part of this PR or submit another PR that then replaces `TempHelpers.hs` in `crypto` and other packages. In fact, I'm a little surprised you didn't need other functionality from the `TempHelpers` module.

binary/test/Test/Pos/Binary/BiSerialize.hs
```diff
@@ -0,0 +1,152 @@
...
+data TestIndexed
+    = TiInt Int
+    | TiIntList [Int]
+    | TiChar2 Char Char
+    | TiInteger Integer
+    | TiMaybeInt (Maybe Int)
+    | TiChar2Permuted Char Char
+    | TiPair TestIndexed TestIndexed
+    deriving (Eq, Show, Typeable)
+
```
erikd: Bingo! This is exactly the sort of thing I was looking for!

binary/test/test.hs
mhuesch: 👍 I'll make the changes in a separate PR.

binary/test/Test/Pos/Binary/BiSerialize.hs
mhuesch: 👍👍

PR #3078, opened 2018-06-08
Use newline-style for imports in stylish haskell

parsonsmatt:
This PR uses a different stylish-haskell import formatting setting which should make the import section less sensitive to whitespace diffs.

Contract the PR to #3079 which runs the current settings on the repo.


.stylish-haskell.yaml
```diff
@@ -198,7 +198,7 @@ steps:
 # Note: we tend to write code which fits into 80 characters, in some cases
 # 100 is acceptable. For imports we always permit 100 characters because it
 # decreases verbosity of diffs and makes merging easier.
-columns: 100
+columns: 80
```
jmitchell: The comment should be revised if we go with 80.

erikd: I *really* wonder about 80. I can see not wanting to go above about 120 (github side-by-side diffs obscure the far right of a line), but 80 just doesn't seem to make sense, especially with 4 space indent. I could possibly live with 80 columns if the indent was 2 spaces.

The current code does roughly adhere to 100 columns. It does not adhere to 80.

auxx/Main.hs
```diff
@@ -4,38 +4,59 @@ module Main
...
+import           Formatting
+    (sformat, shown, (%))
+import           Mockable
+    (Production (..), runProduction)
+import qualified Network.Transport.TCP as TCP
+    (TCPAddr (..))
 import qualified System.IO.Temp as Temp
-import           System.Wlog (LoggerName, logInfo)
+import           System.Wlog
+    (LoggerName, logInfo)
```
erikd: That whole line would have fit in 80 columns! Why is it being wrapped?

auxx/Main.hs
```diff
@@ -4,38 +4,59 @@ module Main
...
+import           Control.Exception.Safe
+    (handle)
+import           Data.Maybe
+    (fromMaybe)
+import           Formatting
+    (sformat, shown, (%))
+import           Mockable
+    (Production (..), runProduction)
+import qualified Network.Transport.TCP as TCP
+    (TCPAddr (..))
```
erikd: Quite honestly, I *really* dislike that. It will take me a *long* time to come to terms with it.

And as I stated above, what does qualified and explicit imports buy us here? if that was just a qualified import:
```
import qualified Network.Transport.TCP as TCP
```
uses of the `TCPAddr` constructors still need to be explicit, eg `TCP.Unaddressable`.

auxx/Main.hs
```diff
@@ -4,38 +4,59 @@ module Main
...
+import           Pos.Util
+    (logException)
+import           Pos.Util.CompileInfo
+    (HasCompileInfo, retrieveCompileTimeInfo, withCompileInfo)
+import           Pos.Util.Config
+    (ConfigurationException (..))
+import           Pos.Util.UserSecret
+    (usVss)
+import           Pos.WorkMode
+    (EmptyMempoolExt, RealMode)
```
erikd: I value vertical space higher than horizontal space. This doesn't do it for me.

auxx/src/Command/Tx.hs
```diff
@@ -12,41 +12,66 @@ module Command.Tx
...
+import           Data.Aeson
+    (eitherDecodeStrict)
 import qualified Data.ByteString as BS
-import           Data.Default (def)
+import           Data.Default
+    (def)
 import qualified Data.HashMap.Strict as HM
-import           Data.List ((!!))
+import           Data.List
+    ((!!))
```
erikd: That is a tool applying overly simple and in my opinion poorly thought out rules.

erikd:
I really cannot get behind this. In my opinon this PR makes a minor problem (some minor issues with the way imports are currently dones and code that doesn't quite conform to standards) and makes it far worse.

Like I said above, when editing code I value vertical space higher than horizontal so:
```
import           Data.List
     ((!!))
```
is just crazy. I also think it actively *hurts* readability.

If it were my code base, I would have rules along the lines:

* If the explicit imports can fit within a single like they should (so the `Data.List` above would not be wrapper).
* If wrapping is needed, it should not wrap as it currently does but to four spaces futher than the module name. eg:
```
import qualified Data.List as List
import              Pos.Wibble.Foo.Baz.Bar (Apple, Cherry, Lemon, Lime, banana,
                        grapefruit, grapes, peach)
import              Test.Pos.Wibble (testWibble)
```

If stylish haskell cannot currently do this, then its probably worth investing the small amount of resources required to fix it.

.stylish-haskell.yaml
adinapoli-iohk: Yeah, I second Erik here. For me 80 columns is a bit too short in terms of screen-estate, especially since we are using 4 spaces indentation 😁

auxx/Main.hs
adinapoli-iohk: I have to say, I agree with Erik again here. I know that's just a visual thing, but it's different from 99% of project I have worked on (or 99% of the Haskell projects out there), and I think being consistent is a good thing, in general 😢

.stylish-haskell.yaml
```diff
@@ -63,7 +63,7 @@ steps:
       #   >     (concat, foldl, foldr, head, init, last, length)
       #
       # Default: after_alias
-      list_align: after_alias
+      list_align: new_line
```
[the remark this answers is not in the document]
parsonsmatt: I'm not sure what you mean -- I'm pretty sure that this just puts the import list on an indented line unconditionally.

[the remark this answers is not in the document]
parsonsmatt: If there *are* any qualified/explicit imports, I'd be in favor of removing them, I don't see why you would go through that trouble :thinking:

auxx/Main.hs
parsonsmatt: :man_shrugging:

auxx/Main.hs
parsonsmatt: I 100% agree that having a qualified import with explicit imports is silly.

auxx/Main.hs
parsonsmatt: That's a reasonable trade off :) I tend to value horizontal more than vertical as I tend to have many vertical splits at a time (2, sometimes 3)

auxx/src/Command/Tx.hs
parsonsmatt: Yeah, I'd agree that the exact logic is ... bad here, though I like the "concept" of it.

parsonsmatt:
I like your proposal @erikd more than this one :smile:

.stylish-haskell.yaml
parsonsmatt: you shouldn't be nesting indents so much :wink: 

The current code should be adhering to 80 columns -- that's what's in the .editorconfig and the style guide. If we want to change that then we should change the relevant docs/editorconfig stuff.

PR #3079, opened 2018-06-08
run stylish-haskell script

parsonsmatt:
This PR runs the `scripts/haskell/stylish.sh` script on the repository. It has become a bit stale.

This PR is unnecessary if we do #3078 , but we should merge one of them (along with a CI step to check stylish-haskell output).


jmitchell:
The [other PR](https://github.com/input-output-hk/cardano-sl/pull/3078/files#diff-b792fcf4ba1eee910f3150e4a9f7a67aR201) sets `columns: 80`. Can we either make that change here too or remove it there?

boothead:
One point on the newline style: If you're using your editor to search for the occurrence of an function in a file and you're looking for the import line, the search result might not now show where the function was imported from... Fairly minor point, but might make life a little more difficult for some.

vincenthz:
I'm really dubious that it's worth the trouble of touching 314 files (or the 916 files of the other PR) and causing ongoing merge/rebase misery for all in-fly works.

parsonsmatt:
I'm not 100% attached to this. An additional style check in CI would be useful, and we'd want to merge this along with it.

It'd be great if we could only run stylish-haskell on files that changed, to make this a more gradual transition.

erikd:
@parsonsmatt says:
> It'd be great if we could only run stylish-haskell on files that changed, to make this a more gradual transition

That can be done by making it part of a git pre-commit hook that only runs on files that are about to be commited.

erikd:
I raised what I consdier major concerns about this in PR #3078.

parsonsmatt:
@erikd Your concerns seem relevant to #3078 but I don't see how they apply here. Do you mean the "only indent four spaces past the module name" comment? That would likely require a patch to stylish-haskell to implement. I'm happy putting this on hold until that's done if that's the consensus on import-style.

erikd:
@parsonsmatt Closing this because I'm pretty sure its been superseded.

PR #3116, opened 2018-06-22
[CBR-304] Track block metadata during ApplyBlock

uroboros:
## Description

Implement BlockMeta as per the wallet spec

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-304

## Type of change

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/Kernel.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel.hs
@@ -53,7 +53,6 @@ import           Cardano.Wallet.Kernel.Types (WalletId (..))
 import           Cardano.Wallet.Kernel.DB.AcidState (ApplyBlock (..),
                      CancelPending (..), CreateHdWallet (..), DB,
                      NewPending (..), NewPendingError, Snapshot (..), defDB)
-import           Cardano.Wallet.Kernel.DB.BlockMeta (BlockMeta (..))
 import           Cardano.Wallet.Kernel.DB.HdWallet
 import qualified Cardano.Wallet.Kernel.DB.HdWallet as HD
 import qualified Cardano.Wallet.Kernel.DB.HdWallet.Create as HD
--- a/wallet-new/src/Cardano/Wallet/Kernel.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel.hs
@@ -210,11 +210,8 @@ applyBlock :: PassiveWallet
 applyBlock pw@PassiveWallet{..} b
     = do
         blocksByAccount <- prefilterBlock' pw b
-        -- TODO(@uroboros/ryan) do proper metadata initialisation (as part of CBR-239: Support history tracking and queries)
-        let blockMeta = BlockMeta . InDb $ Map.empty
-
         -- apply block to all Accounts in all Wallets
-        void $ update' _wallets $ ApplyBlock (blocksByAccount, blockMeta)
+        void $ update' _wallets $ ApplyBlock blocksByAccount
 
 -- | Apply multiple blocks, one at a time, to all wallets in the PassiveWallet
 --
--- a/wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
@@ -151,12 +151,12 @@ applyBlock blocksByAccount = runUpdateNoErrors $ zoom dbHdWallets $
                            modify $ Spec.applyBlock prefBlock)
         blocksByAccount
   where
-    -- Accounts are discovered during wallet creation (if the account was given
-    -- a balance in the genesis block) or otherwise during ApplyBlock. For
-    -- accounts discovered during ApplyBlock, we can assume that there was no
-    -- genesis utxo for this Account (since if there was, the account would have
-    -- been created with the wallet). Hence we use empty initial utxo for accounts
-    -- discovered during applyBlock.
+    -- NOTE: When we create the new wallet, we look at the genesis UTxO and create
+    -- an initial balance for all accounts that we recognize as ours. This means that
+    -- when we later discover a new account that is also ours, it cannot appear
+    -- in the genesis UTxO, because if it did, we would already have seen it
+    -- (the genesis UTxO is static, after all). Hence we use empty initial utxo
+    -- for accounts discovered during applyBlock.
     -- The Addrs need to be created during account initialisation and so we pass them here.
     initUtxoAndAddrs :: PrefilteredBlock -> (Utxo,[AddrWithId])
     initUtxoAndAddrs pb = (Map.empty, pfbAddrs pb)
--- a/wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
@@ -152,10 +152,12 @@ applyBlock blocksByAccount = runUpdateNoErrors $ zoom dbHdWallets $
         blocksByAccount
   where
     -- Accounts are discovered during wallet creation (if the account was given
-    -- a balance in the genesis block) or otherwise, during ApplyBlock. For
+    -- a balance in the genesis block) or otherwise during ApplyBlock. For
     -- accounts discovered during ApplyBlock, we can assume that there was no
-    -- genesis utxo, hence we use empty initial utxo for such new accounts.
-    -- The Addrs will need to created during account initialisation and so we pass them here.
+    -- genesis utxo for this Account (since if there was, the account would have
+    -- been created with the wallet). Hence we use empty initial utxo for accounts
+    -- discovered during applyBlock.
+    -- The Addrs need to be created during account initialisation and so we pass them here.
     initUtxoAndAddrs :: PrefilteredBlock -> (Utxo,[AddrWithId])
     initUtxoAndAddrs pb = (Map.empty, pfbAddrs pb)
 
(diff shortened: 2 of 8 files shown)


wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -217,8 +222,7 @@ createPrefiltered initUtxoAndAddrs applyP accs = do
                 , _checkpointUtxoBalance = InDb $ Spec.balance utxo'
                 , _checkpointExpected    = InDb Map.empty
                 , _checkpointPending     = Pending . InDb $ Map.empty
-                -- TODO(@uroboros/ryan) proper BlockMeta initialisation (as part of CBR-239: Support history tracking and queries)
-                , _checkpointBlockMeta   = BlockMeta . InDb $ Map.empty
+                , _checkpointBlockMeta   = mempty::BlockMeta
```
adinapoli-iohk: Weird, why did you need the explicit type annotation here? Wasn't `mempty` enough?

wallet-new/src/Cardano/Wallet/Kernel/DB/InDb.hs
```diff
@@ -58,6 +58,10 @@ instance SafeCopy (InDb Core.Coin) where
   getCopy = error "TODO: getCopy for (InDb Core.Coin)"
   putCopy = error "TODO: putCopy for (InDb Core.Coin)"
 
+instance SafeCopy (InDb Core.SlotId) where
+  getCopy = error "TODO: getCopy for (InDb Core.SlotId)"
+  putCopy = error "TODO: putCopy for (InDb Core.SlotId)"
```
adinapoli-iohk: @edsko told me but I have forgot: which is the name of the precise ticket where we are going to "fill" all this stubbed implementations?

wallet-new/src/Cardano/Wallet/Kernel/DB/Resolved.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE DeriveGeneric, DeriveDataTypeable             #-}
```
adinapoli-iohk: Why do we need this one? This PR doesn't add any extra generic derivation, I'm confused 😀

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
uroboros: `mempty` was enough, yes, I was just being explicit (I'm removing it though)

wallet-new/src/Cardano/Wallet/Kernel/DB/Resolved.hs
uroboros: that was shrapnel - removed

wallet-new/src/Cardano/Wallet/Kernel/DB/InDb.hs
uroboros: we could do it as part of this  "[CBR-300] Address all the TODOs left dangling after CBR-238"

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
adinapoli-iohk: That's a laudable goal -- I usually resort to an explicit (yet banal) "smart constructor" like `newBlockMeta`, but here I guess the fact `mempty` is contextualised by the `_checkpointBlockMeta` field makes things explicit enough 😉 

I reckon it's just a matter of preferences, though!

wallet-new/src/Cardano/Wallet/Kernel/DB/InDb.hs
uroboros: agreed, it should happen asap, I'll get onto it before rollback for that matter https://iohk.myjetbrains.com/youtrack/issue/CBR-305

wallet-new/src/Cardano/Wallet/Kernel.hs
```diff
@@ -203,11 +202,8 @@ applyBlock :: PassiveWallet
 applyBlock pw@PassiveWallet{..} b
     = do
         blocksByAccount <- prefilterBlock' pw b
-        -- TODO(@uroboros/ryan) do proper metadata initialisation (as part of CBR-239: Support history tracking and queries)
-        let blockMeta = BlockMeta . InDb $ Map.empty
-
         -- apply block to all Accounts in all Wallets
-        void $ update' _wallets $ ApplyBlock (blocksByAccount, blockMeta)
+        void $ update' _wallets $ ApplyBlock blocksByAccount
```
edsko: Yup, like this structure better (having the block metadata part of the prefiltered block). The spec stipulates that, too :)

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -122,21 +120,20 @@ newPending accountId tx = runUpdate' . zoom dbHdWallets $
...
         (\prefBlock -> zoom hdAccountCheckpoints $
-                           modify $ Spec.applyBlock (prefBlock,meta))
+                           modify $ Spec.applyBlock prefBlock)
         blocksByAccount
   where
     -- Accounts are discovered during wallet creation (if the account was given
     -- a balance in the genesis block) or otherwise, during ApplyBlock. For
     -- accounts discovered during ApplyBlock, we can assume that there was no
     -- genesis utxo, hence we use empty initial utxo for such new accounts.
-    initUtxoAndAddrs :: PrefilteredBlock -> (Utxo, [AddrWithId])
```
edsko: This is unrelated to this PR per se, but it's something that I've been pondering for a while. Why is this assumption safe? I don't think it is, I think we need to prefilter the genesis UTxO here. After all, there is no guarantee that the account ID we discover doesn't have an initial balance in the genesis UTxO, is there?

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -160,15 +157,21 @@ switchToFork n blocks = runUpdateNoErrors $
 --
 --  Given prefiltered utxo's, by account, create an HdAccount for each account,
 --  along with HdAddresses for all utxo outputs.
+--
+-- NOTE: we don't have the slotId of transactions occuring in the Utxo,
+--       which prevents us from recording the BlockMeta for the initial utxo
 createHdWallet :: HdRoot
-               -> Map HdAccountId PrefilteredUtxo
+               -> Map HdAccountId (Utxo,[AddrWithId],[Core.TxId])
```
edsko: I don't understand this. You added an additional `[Core.TxId]` argument here, but you also add a comment saying we can't use them, and indeed, in the body of this function we don't seem to use it at all, do we?

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
```diff
@@ -217,8 +221,7 @@ createPrefiltered initUtxoAndAddrs applyP accs = do
                 , _checkpointUtxoBalance = InDb $ Spec.balance utxo'
                 , _checkpointExpected    = InDb Map.empty
                 , _checkpointPending     = Pending . InDb $ Map.empty
-                -- TODO(@uroboros/ryan) proper BlockMeta initialisation (as part of CBR-239: Support history tracking and queries)
-                , _checkpointBlockMeta   = BlockMeta . InDb $ Map.empty
+                , _checkpointBlockMeta   = mempty
```
edsko: This could do with a comment, saying something like "Since this is the first checkpoint _before_ we have applied any blocks, the block metadata is empty."

wallet-new/src/Cardano/Wallet/Kernel/DB/InDb.hs
edsko: As per the meeting yesterday, let's give rollback a higher priority.

wallet-new/src/Cardano/Wallet/Kernel/DB/Resolved.hs
```diff
@@ -56,6 +58,9 @@ data ResolvedTx = ResolvedTx {
 data ResolvedBlock = ResolvedBlock {
       -- | Transactions in the block
       _rbTxs  :: [ResolvedTx]
+
+      -- | Source MainBlock SlotId
```
edsko: Don't really understand this comment. How about something like "The slot number of the slot this block appeared in"?

wallet-new/src/Cardano/Wallet/Kernel/DB/Spec/Update.hs
```diff
@@ -98,14 +98,17 @@ applyBlock (prefBlock, _bMeta) checkpoints
...
+        pending''           = updatePending   prefBlock (checkpoints ^. currentPendingTxs)
+        blockMeta''         = updateBlockMeta prefBlock (checkpoints ^. currentBlockMeta)
+        -- TODO(@uroboros/ryan) applyBlock.updateExpected
         -- (as part of CBR-150 Extend pure data layer to support rollback)
         expected''          = checkpoints ^. currentExpected
-        blockMeta''         = checkpoints ^. currentBlockMeta
+
+updateBlockMeta :: PrefilteredBlock -> BlockMeta -> BlockMeta
+updateBlockMeta PrefilteredBlock{..} meta
+    = pfbMeta `mappend` meta
```
edsko: It probably doesn't really matter, but the order of the operands to `mappend` here makes me a little uneasy.

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
uroboros: I think the new comment explains the empty initial Utxo better...

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
uroboros: Fixed: The TxIds are useful in other settings, and createHdWallet was carrying that burden here...

wallet-new/src/Cardano/Wallet/Kernel/DB/AcidState.hs
edsko: Hmm, I think it can still be made clearer. Something like this:

NOTE: When we create the new wallet, we look at the genesis UTxO and create an initial balance for _all_ accounts that we recognize as ours. This means that when we later discover a new account that is also ours, it _cannot_ appear in the genesis UTxO, because if it did, we would already have seen it (the genesis UTxO is static, after all).

Something along those lines.

uroboros: that's nice, I'll use your version...

erikd:
> But this lacks any and all tests. 

I thought the policy was that all new code should be accompanied by tests :)

edsko:
> I thought the policy was that all new code should be accompanied by tests :)

I also thought you were arguing that we shouldn't leave branches open for too long. This doesn't affect any existing code, and there is a new ticket for the tests.

erikd:
My problem is less with long running branches and more with the difficulty of reviewing large PRs containing large chunks of work. I am also a fan of code and tests for that code being developed at the same time.

PR #3159, opened 2018-06-28
[CBR-312] Refined exceptions

parsonsmatt:
## Description

This PR refines the type of exceptions that the new wallet API code throws.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-312

## Type of change

- [] 🐞 Bug fix (non-breaking change which fixes an issue)
- [] 🛠 New feature (non-breaking change which adds functionality)
- [] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
- [] ⛑ git-flow chore (backport, hotfix, etc)

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/V1/Types.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Types.hs
@@ -1888,6 +1888,16 @@ instance BuildableSafeGen NodeInfo where
 -- If constructor does not have additional data (like in case of `WalletNotFound` error),
 -- then "diagnostic" field will be empty object.
 --
+-- Note that this exception type should only be used for errors that we throw to
+-- short-circuit an API request, and this type is primarily used for formatting
+-- to the JSend format generically. Because the type has so many constructors,
+-- it should not be used in the 'Left' case of an 'Either' result -- you lose
+-- type-safety in which constructors you should be throwing and handling.
+-- Instead, use a smaller type that doens't have an 'Exception' instance, and do
+-- your best to handle it using 'Either', 'ExceptT', etc. without it bubbling
+-- out to the client. Add the constructor case here and convert to the
+-- 'WalletError' only if you need to tell the end user of it.
+--
 -- TODO: change fields' types to actual Cardano core types, like `Coin` and `Address`
 data WalletError =
       NotEnoughMoney { weNeedMore :: !Int }


wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -1888,6 +1888,16 @@ instance BuildableSafeGen NodeInfo where
...
 --
+-- Note that this exception type should only be used for errors that we throw to
+-- short-circuit an API request, and this type is primarily used for formatting
+-- to the JSend format generically. Because the type has so many constructors,
+-- it should not be used in the 'Left' case of an 'Either' result -- you lose
+-- type-safety in which constructors you should be throwing and handling.
+-- Instead, use a smaller type that doens't have an 'Exception' instance, and do
+-- your best to handle it using 'Either', 'ExceptT', etc. without it bubbling
+-- out to the client. Add the constructor case here and convert to the
+-- 'WalletError' only if you need to tell the end user of it.
```
parsonsmatt: My initial plan was to attack the `WalletError` type, but after reviewing how it's actually used, it makes more sense to comment how it's intended to be used -- not as something for the API backend logic to use to signal error conditions, but as something for the API backend to use to render API errors.

I don't think this ticket is profitable to pursue further at this point :\

KtorZ: I've always found this nomenclature quite wrong for API. I mean, we make sort of a clear distinction between errors and responses whereas _"errors"_ are just non-positive responses. I mean, as long as we've been able to handle the request correctly, might it be by saying no, nothing went "wrong" right? 
Though we kinda _need_ this exception mechanism to short-circuit stuff indeed; yet I wonder what it would look like with continuations instead, treating user _"errors"_ simply as a type of possible response. 

Anyway, thanks for digging into this, let's simply extend this `WalletError` with what @akegalj finds and hunt down those remaining raw strings being thrown.

KtorZ:
Ignoring `appveyor` failure here, we aren't merging to develop and it's seriously 10 lines of comments in a file -___- ...

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: It's a gray area. Is a segfault really an "error" if the computer noticed it and responded appropriately (by killing the offending program)? Well, it's an error in the *program*, but it's correct behavior for the computer. Likewise, these errors mean that, at some level, a developer or client did something wrong, and now we have to report that higher up in the system to handle appropriately.

PR #3170, opened 2018-07-02
[CBR-315] Have only pure getters for the Kernel's DB

adinapoli-iohk:
This PR exposes only a single getter for the Kernel, which is the
'getWalletSnapshot' API, explaining library authors and users what they
are expected to do and how to interact with the wallet.

## Description

We don't want to have developers pick up bad old habits when it comes to the new data layer. We want to enforce a way to access the data layer internal state in a consistent manner, and in line with what we have done in the past.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-315

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/Kernel.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel.hs
@@ -9,9 +9,8 @@
 module Cardano.Wallet.Kernel (
     -- * Passive wallet
     PassiveWallet -- opaque
+  , DB -- opaque
   , WalletId
-  , accountUtxo
-  , accountTotalBalance
   , applyBlock
   , applyBlocks
   , bracketPassiveWallet
@@ -20,6 +19,10 @@ module Cardano.Wallet.Kernel (
   , walletLogMessage
   , walletPassive
   , wallets
+    -- * The only effectful getter you will ever need
+  , getWalletSnapshot
+    -- * Pure getters acting on a DB snapshot
+  , module Getters
     -- * Active wallet
   , ActiveWallet -- opaque
   , bracketActiveWallet
@@ -34,8 +37,6 @@ import           Control.Lens.TH
 import qualified Data.Map.Strict as Map
 import           Data.Time.Clock.POSIX (getPOSIXTime)
 
-import           Formatting (build, sformat)
-
 import           System.Wlog (Severity (..))
 
 import           Data.Acid (AcidState)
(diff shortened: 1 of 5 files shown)


wallet-new/test/unit/Wallet/Inductive/Cardano.hs
```diff
@@ -223,12 +223,12 @@ equivalentT activeWallet (pk,esk) = \mkWallet w ->
...
             => Text
             -> (Wallet h Addr -> a)
-            -> (Kernel.PassiveWallet -> IO (Interpreted a))
+            -> (Kernel.DB -> Interpreted a)
             -> TranslateT (EquivalenceViolation h) m ()
         cmp fld f g = do
           let dsl = f inductiveCtxtWallet
           translated <- toCardano ctxt fld dsl
-          kernel     <- liftIO $ g passiveWallet
+          kernel     <- g <$> liftIO (Kernel.getWalletSnapshot passiveWallet)
```
uroboros: @adinapoli-iohk `checkWalletState` calls `cmp` with multiple queries, so you can move the `getWalletSnapshot` there instead and pass the snapshot into `cmp`. That will also demonstrate the intent of this PR nicely.

wallet-new/src/Cardano/Wallet/Kernel.hs
```diff
@@ -300,24 +304,44 @@ cancelPending passiveWallet cancelled =
...
+  The @only@ effectful function we expose is 'getWalletSnapshot', which reads
+  the full DB 'Snapshot' and returns it.
+  All the other getters are completely pure and take the 'Snapshot' as input,
+  so that users of the wallet are forced to re-use the same 'Snapshot' in case
+  they want to read the state of the wallet multiple times within the same
+  code block / handler.
+
 -------------------------------------------------------------------------------}
 
+-- | A type synonym to stress the fact that what we are carrying around is not
```
uroboros: Which type synonym does this refer to?

adinapoli-iohk: Good catch! I've previously introduced a type synonym `type DbSnapshot = DB` but decided later on it was not worth the cognitive overload. Will remove the stale comment!

wallet-new/test/unit/Wallet/Inductive/Cardano.hs
adinapoli-iohk: I see! The reason I didn't do it right off the bat is that I wasn't sure we needed access to the full `PassiveWallet`. Will change!

wallet-new/src/Cardano/Wallet/Kernel.hs
```diff
@@ -20,6 +19,11 @@ module Cardano.Wallet.Kernel (
   , walletLogMessage
   , walletPassive
   , wallets
+    -- * The only effectful getter you will ever need
+  , getWalletSnapshot
+    -- * Read-only, pure getters
+  , accountUtxo
```
edsko: What about not exporting these at all, and having a separate module with the read-only functions on the snapshot? (I think we might already have such a module, but it's possible that we don't have a "top-level" one.) That way when people add _new_ read only functions they're not tempted to make them effectful here, because they'd be adding them to the module with all the rest of them and there's not a single effect in sight.

adinapoli-iohk:
> I wouldn't personally re-export the whole module but I know I'm in the minority on that one :)

I don't have a strong opinion on this either, but I thought I would have been handy for developers to use 😉

PR #3172, opened 2018-07-02
Revert "[CDEC-299] Remove field accessors for `GenesisConfiguration`"

disassembler:
This reverts commit b258f926580387f113d746f7f897444c0dcfc570 from PR #3140.

The record fields are used for parsing of `lib/configuration.yaml`.


mhuesch:
@rvl @disassembler Yikes, sorry we broke that! Good call to revert it. If we re-remove the partial function field accessors we'll rewrite the JSON instances to make sure they are backwards compatible.

Are there tests that we should've run to catch the regression?

cc/ @Jimbo4350

Jimbo4350:
Apologies guys!

PR #3174, opened 2018-07-02
[DEVOPS-810] Fix nix-shell env

rvl:
## Description

The change to `justStaticExecutablesGitRev` broke the `.env` attribute of haskell derivations.

Also there is a new nixpkgs lib function for filtering sources.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-810
https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-937

## Type of change
- [x] Build scripts/dev environment bug fix
- [x] Build scripts refactor

## QA Steps

    nix-shell default.nix -A cardano-sl-wallet-new.env
    nix-build -A all-cardano-sl

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -113,29 +113,7 @@ let
         # Provide a dummy installPhase for benchmark packages.
         installPhase = "mkdir -p $out";
       }) // optionalAttrs (args ? src) {
-        src = let
-           cleanSourceFilter = with pkgs.stdenv;
-             name: type: let baseName = baseNameOf (toString name); in ! (
-               # Filter out .git repo
-               (type == "directory" && baseName == ".git") ||
-               # Filter out editor backup / swap files.
-               lib.hasSuffix "~" baseName ||
-               builtins.match "^\\.sw[a-z]$" baseName != null ||
-               builtins.match "^\\..*\\.sw[a-z]$" baseName != null ||
-
-               # Filter out locally generated/downloaded things.
-               baseName == "dist" ||
-
-               # Filter out the files which I'm editing often.
-               lib.hasSuffix ".nix" baseName ||
-               # Filter out nix-build result symlinks
-               (type == "symlink" && lib.hasPrefix "result" baseName)
-             );
-
-          in
-            if (builtins.typeOf args.src) == "path"
-              then builtins.filterSource cleanSourceFilter args.src
-              else args.src or null;
+        src = localLib.cleanSourceTree args.src;
       } // optionalAttrs enableDebugging {
         # TODO: DEVOPS-355
         dontStrip = true;
--- a/default.nix
+++ b/default.nix
@@ -132,10 +132,12 @@ let
     walletIntegrationTests = pkgs.callPackage ./scripts/test/wallet/integration { inherit gitrev; };
     validateJson = pkgs.callPackage ./tools/src/validate-json {};
     demoCluster = pkgs.callPackage ./scripts/launch/demo-cluster { inherit gitrev; };
-    tests = {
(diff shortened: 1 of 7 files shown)


lib.nix
```diff
@@ -12,10 +12,23 @@ let
        then result
        else default;
 
+  filterSource = src:
```
jmitchell: There's a `builtin.` by the same name. Using a different name, like `filterSrc`, may reduce confusion/mistakes.

default.nix
```diff
@@ -113,29 +113,7 @@ let
...
-               lib.hasSuffix ".nix" baseName ||
-               # Filter out nix-build result symlinks
-               (type == "symlink" && lib.hasPrefix "result" baseName)
-             );
-
-          in
-            if (builtins.typeOf args.src) == "path"
-              then builtins.filterSource cleanSourceFilter args.src
-              else args.src or null;
+        src = localLib.filterSource args.src;
```
jmitchell: As `filterSource` is implemented I think some filters are getting dropped, like:

* `# Filter out .git repo`
* `# Filter out editor backup / swap files.`
* `# Filter out nix-build result symlinks`

Is that what we want?

rvl: The good thing about `pkgs.lib.cleanSourceWith` (as opposed to `builtins.filterSource`) is that it's chainable. In `localLib.filterSource` it calls the "default" `pkgs.lib.cleanSource` function which provides the above filters. I verify this by running nix-build and checking that a cached build is used.

lib.nix
rvl: OK, will rename to `localLib.cleanHaskellSource`.

lib.nix
```diff
@@ -12,10 +12,23 @@ let
...
 
+  cleanHaskellSource = src:
+    if (builtins.typeOf src) == "path"
+      then lib.cleanSourceWith {
+        filter = with pkgs.stdenv;
+          name: type: let baseName = baseNameOf (toString name); in ! (
+            # Filter out cabal build products
+            baseName == "dist" ||
+            # Filter out files which don't affect cabal build
+            lib.hasSuffix ".nix" baseName
```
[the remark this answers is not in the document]
rvl: Thanks for the review. The goal of these source filters is to avoid nix-build rebuilds of cabal packages when only "irrelevant" files have been changed. For example, the common case of when I edit a `.nix` file within the package's source directory. Try this:

    nix-build -A cardano-sl-wallet-new  # should download from hydra cache
    touch wallet-new/hello
    nix-build -A cardano-sl-wallet-new  # now it wants to rebuild everything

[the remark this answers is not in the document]
k0001: Yeah, the purpose is clear. It's just that one wouldn't normally expect a function named “clean haskell source” to pay attention to `.nix` files.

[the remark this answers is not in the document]
rvl: OK, now it's called `localLib.cleanSourceTree`.

rvl:
I have added a few more useful filters to `localLib.cleanSourceTree`, and changed other instances of `builtins.filterSource` in the source code tests to use `lib.cleanSourceWith`.

lib.nix
```diff
@@ -12,10 +12,32 @@ let
        then result
        else default;
 
+  # Removes files within a Haskell source tree which won't change the
+  # result of building the package.
+  # This is so that cached build products can be used whenever possible.
+  # It also applies the lib.cleanSource filter from nixpkgs which
+  # removes VCS directories, emacs backup files, etc.
+  cleanSourceTree = src:
```
disassembler: Can we add doc files to this function as well so changing md files doesn't trigger new builds?

rvl: Probably not because there is usually `extra-source-files: README.md` in the cabal file.

PR #3175, opened 2018-07-02
[CDEC-405]

coot:
## Description

This solves the recent problem with `lib/configuration.yaml` not being parsable anymore.  This happend because core team rightly removed some partial accessor functions.   I just updated the yaml file and I alsoI added a test.  This is a short term solution, long term I think we should write `FromJSON` instances for the types we are reading, because just removing fields from config file will make it less and less readable.  @erikd it's up to you to decide if this maintaining `FromJSON` instances is worth the effort vs readability of the config file.  I guess this really depends how many more partial functions we will remove.

## Linked issue
https://iohk.myjetbrains.com/youtrack/issue/CDEC-405

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/lib/cardano-sl.cabal
+++ b/lib/cardano-sl.cabal
@@ -277,6 +277,7 @@ test-suite cardano-test
                        Test.Pos.ConstantsSpec
                        Test.Pos.Diffusion.BlockSpec
                        Test.Pos.Genesis.CanonicalSpec
+                       Test.Pos.Launcher.ConfigurationSpec
                        Test.Pos.MerkleSpec
                        Test.Pos.Infra.Slotting.TypesSpec
                        Test.Pos.Types.Identity.SafeCopySpec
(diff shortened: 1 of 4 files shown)


lib/configuration.yaml
```diff
@@ -14773,8 +14773,8 @@ mainnet_full: &mainnet_full
     <<: *mainnet_base_core
     genesis:
       src:
-        file: mainnet-genesis.json
-        hash: 5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb
+        - mainnet-genesis.json
+        - 5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb
```
erikd: I suspect that is wrong, or rather that there should not be any change to that file. Commit b258f926580387f1 changed the way the config was parsed and has since been reverted.

You can probably drop these changes to this file and still have the test pass (as long as you have rebased).

coot:
@erikd, @parsonsmatt, @avieth any chance for taking a look at this tiny PR.

lib/configuration.yaml
coot: Great, let me try!

PR #3183, opened 2018-07-03
[CO-307] Stub API endpoints

parsonsmatt:
## Description

Initial stub of the endpoints we'll be using for CO-307. These endpoints are subject to change.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CO-307

## Type of change

- [] 🐞 Bug fix (non-breaking change which fixes an issue)
- [] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
- [] 🏭 Refactoring that does not change existing functionality but does improve things like code readability, structure etc
- [] 🔨 New or improved tests for existing code
- [] ⛑ git-flow chore (backport, hotfix, etc)

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/V1/Daedalus.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Daedalus.hs
@@ -8,12 +8,24 @@ module Cardano.Wallet.API.V1.Daedalus where
 type API =
     "daedalus"
         :> ( "update"
-            :> ( "apply" :> Get '[ValidJSON] (WalletResponse ())
-            :<|> "postpone" :> Get '[ValidJSON] (WalletResponse ())
+            :> ( "apply"
+                :> Post '[ValidJSON] NoContent
+            :<|> "postpone"
+                :> Post '[ValidJSON] NoContent
             )
-        :<|> "papervend" :>  "redemptions" :> "ada" :> Get '[ValidJSON] (WalletResponse ())
-        :<|> "redemptions" :> "ada" :> Get '[ValidJSON] (WalletResponse ())
+        :<|> "redemptions"
+            :> "ada"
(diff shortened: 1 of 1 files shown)


wallet-new/src/Cardano/Wallet/API/V1/Daedalus.hs
```diff
@@ -0,0 +1,74 @@
+-- | This module contains the top level API definition for frontend-related
+-- tasks.  The API endpoints presented here are intended for use with the
+-- Daedalus client, and aren't useful for wallets, exchanges, and other users.
+module Cardano.Wallet.API.V1.Daedalus where
+
+-- migrate everything except for import/export
+
+type API =
+    "daedalus"
```
adinapoli-iohk: @parsonsmatt Ah, interesting change in perspective -- essentially we are moving away from a "kitchen-sink" style of API where everything is collated together, but we are instead keeping "namespace-free" only the "core APIs" which are specific to the wallet state manipulation, and instead cleanly separating deadalus-specific operations into the `daedalus` namespace.

I think it surely make sense. 

Another advantage I can see with this approach is that it will allow us to separate the "let's update daedalus" vs "let's vote or apply an update at the blockchain level", which are two aspects at the moment confused but that in principle we might to keep disjointed. Nice!

My RESTful OCD would make me incline to say that, if possible, we shouldn't use verbs for endpoints -- if these endpoints are going to stay, can we think about anything different?

parsonsmatt: Yeah, that's a good point. We can change it to `v1/internal/update?action=[apply|postponse]` to clean it up

parsonsmatt:
This PR was made under the other name; the CO-334 one covers this task.

PR #3184, opened 2018-07-03
Update text rendering of dependency graph

mhuesch:
This is the current dependency graph. It is an ASCII rendering of
the PNG output from `stack dot | ... | dot`. `-test` and `-bench`
packages are omitted.

This graph was run through `tred` which eliminates redundant deps
(those which are implied by transitivity).

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -6,11 +6,46 @@ flags:
 
 extra-package-dbs: []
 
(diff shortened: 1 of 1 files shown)


mhuesch:
That sounds reasonable. It would be a pain to keep that thing up-to-date... And since everyone has `stack dot` accessible, it's just a matter of having `graphviz`, which is no biggie.

PR #3187, opened 2018-07-04
[CBR-314] Kernel Keystore

adinapoli-iohk:
## Description

This PR introduces an opaque `Keystore` abstraction and a set of APIs meant to be able to store and retrieve `EncryptedSecretKey` from disk "out-of-acid-state", in the sense that those delicate sensitive information shouldn't be stored inside `acid-state`, to not incur in the risk of accidentally leaking them in the transaction logs.

The presented API models the one for `Data.Map` and should be quite general thanks to the use of the `Kernel.WalletId` type which incapsulate the notion of different ids, being them coming from HD wallets or elsewhere.

The concrete implementation is quite inefficient and relies on the already-present `UserSecret`, as reinventing the wheel here would be detrimental as it makes sense to piggyback on what we have already instead of rewriting it from scratch.

This `Keystore` is now hooked into the `PassiveWallet` and this means we can also get rid of the artificial creation of a `HdWallet` inside the worker thread, but rather start using this abstraction (cc @matt-noonan).

Tests are provided.

Let me know what you think!

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-314

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/client/src/Pos/Client/KeyStorage.hs
+++ b/client/src/Pos/Client/KeyStorage.hs
@@ -19,7 +19,6 @@ module Pos.Client.KeyStorage
        , KeyData
        , KeyError (..)
        , AllUserSecrets (..)
-       , keyDataFromFile
        ) where
 
 import           Universum
@@ -27,12 +26,11 @@ import           Universum
 import qualified Control.Concurrent.STM as STM
 import           Control.Lens ((<%=), (<>~))
 import           Serokell.Util (modifyTVarS)
-import           System.Wlog (WithLogger)
 
 import           Pos.Crypto (EncryptedSecretKey, PassPhrase, SecretKey, hash,
                      runSecureRandom, safeKeyGen)
-import           Pos.Util.UserSecret (HasUserSecret (..), UserSecret,
-                     peekUserSecret, usKeys, usPrimKey, writeUserSecret)
+import           Pos.Util.UserSecret (HasUserSecret (..), UserSecret, usKeys,
+                     usPrimKey, writeUserSecret)
 
 type KeyData = TVar UserSecret
 
@@ -108,9 +106,6 @@ newSecretKey pp = do
 
 containsKey :: [EncryptedSecretKey] -> EncryptedSecretKey -> Bool
 containsKey ls k = hash k `elem` map hash ls
-
-keyDataFromFile :: (MonadIO m, WithLogger m) => FilePath -> m KeyData
-keyDataFromFile fp = peekUserSecret fp >>= liftIO . STM.newTVarIO
 
 data KeyError =
     PrimaryKey !Text -- ^ Failed attempt to delete primary key
--- a/wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
@@ -125,8 +125,18 @@ deriveSafeCopy 1 'base ''HasSpendingPassword
   HD wallets
 -------------------------------------------------------------------------------}
 
--- | HD wallet root ID
-data HdRootId = HdRootId (InDb (Core.AddressHash Core.PublicKey))
+-- | HD wallet root ID. Conceptually, this is just an 'Address' in the form
+-- of 'Ae2tdPwUPEZ18ZjTLnLVr9CEvUEUX4eW1LBHbxxxJgxdAYHrDeSCSbCxrvx', but is,
+-- in a sense, a special breed as its derived from the 'PublicKey' (derived
+-- from some BIP-39 mnemonics, typically) and which does not depend from any
+-- delegation scheme, as you cannot really pay into this 'Address'. This
+-- ensures that, given an 'EncryptedSecretKey' we can derive its 'PublicKey'
+-- and from that the 'Core.Address'.
+-- On the \"other side\", given a RESTful 'WalletId' (which is ultimately
+-- just a Text) it's possible to call 'decodeTextAddress' to grab a valid
+-- 'Core.Address', and then transform this into a 'Kernel.WalletId' type
+-- easily.
+data HdRootId = HdRootId (InDb Core.Address)
   deriving (Eq, Ord)
 
 -- | HD wallet account ID
@@ -297,7 +307,9 @@ instance HasPrimKey HdAddress where
     type PrimKey HdAddress = HdAddressId
     primKey = _hdAddressId
 
-type HdRootIxs    = '[]
+-- | An 'HdRoot' can be accessed via its 'Core.Address', which is the
+-- Cardano 'Address' derived by the PublicKey associated with it.
+type HdRootIxs    = '[Core.Address]
 type HdAccountIxs = '[HdRootId]
 type HdAddressIxs = '[HdRootId, HdAccountId, Core.Address]
 
@@ -308,6 +320,8 @@ type instance IndicesOf HdAddress = HdAddressIxs
 instance IxSet.Indexable (HdRootId ': HdRootIxs)
                          (OrdByPrimKey HdRoot) where
     indices = ixList
+                (ixFun (\hdRoot -> case view hdRootId hdRoot of
+                                        HdRootId (InDb addr) -> [addr]))
 
 instance IxSet.Indexable (HdAccountId ': HdAccountIxs)
                          (OrdByPrimKey HdAccount) where
@@ -439,7 +453,7 @@ assumeHdAccountExists _id = return ()
 
 -- | Computes the 'HdRootId' from the given 'EncryptedSecretKey'.
 eskToHdRootId :: Core.EncryptedSecretKey -> HdRootId
-eskToHdRootId = HdRootId . InDb . Core.addressHash . Core.encToPublic
+eskToHdRootId = HdRootId . InDb . Core.makePubKeyAddressBoot . Core.encToPublic
 
 {-------------------------------------------------------------------------------
   Pretty printing
--- a/wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
@@ -58,6 +58,8 @@ module Cardano.Wallet.Kernel.DB.HdWallet (
   , zoomOrCreateHdAddress
   , assumeHdRootExists
   , assumeHdAccountExists
+    -- * General-utility functions
+  , eskToHdRootId
   ) where
 
 import           Universum
@@ -430,6 +432,14 @@ assumeHdRootExists _id = return ()
 -- Helper function which can be used as an argument to 'zoomOrCreateHdAddress'
 assumeHdAccountExists :: HdAccountId -> Update' HdWallets e ()
 assumeHdAccountExists _id = return ()
+
+{-------------------------------------------------------------------------------
+  General-utility functions
+-------------------------------------------------------------------------------}
+
+-- | Computes the 'HdRootId' from the given 'EncryptedSecretKey'.
+eskToHdRootId :: Core.EncryptedSecretKey -> HdRootId
+eskToHdRootId = HdRootId . InDb . Core.addressHash . Core.encToPublic
 
 {-------------------------------------------------------------------------------
   Pretty printing
--- a/wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
@@ -127,7 +127,7 @@ deriveSafeCopy 1 'base ''HasSpendingPassword
 
 -- | HD wallet root ID. Conceptually, this is just an 'Address' in the form
 -- of 'Ae2tdPwUPEZ18ZjTLnLVr9CEvUEUX4eW1LBHbxxxJgxdAYHrDeSCSbCxrvx', but is,
--- in a sense, a special breed as its derived from the 'PublicKey' (derived
+-- in a sense, a special breed as it's derived from the 'PublicKey' (derived
 -- from some BIP-39 mnemonics, typically) and which does not depend from any
 -- delegation scheme, as you cannot really pay into this 'Address'. This
 -- ensures that, given an 'EncryptedSecretKey' we can derive its 'PublicKey'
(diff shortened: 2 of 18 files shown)


client/src/Pos/Client/KeyStorage.hs
```diff
@@ -19,20 +19,18 @@ module Pos.Client.KeyStorage
        , KeyData
        , KeyError (..)
        , AllUserSecrets (..)
-       , keyDataFromFile
```
adinapoli-iohk: It turns out this was never used.

wallet-new/src/Cardano/Wallet/Kernel/Types.hs
```diff
@@ -59,11 +61,9 @@ data WalletId =
 
     deriving (Eq, Ord)
 
--- | Map of Wallet Master keys indexed by WalletId
---
--- TODO: We may need to rethink having this in-memory
--- ESK should _not_ end up in the wallet's acid-state log
-type WalletESKs = Map WalletId EncryptedSecretKey
```
adinapoli-iohk: This type was effectively superseded by the `Keystore`.

KtorZ:
Interesting @adinapoli-iohk  ... I am facing quite a lot of issues recently with integration tests; let me have a look.

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,194 @@
...
+     Keystore <$> newMVar (InternalStorage us StorageInitialised)
+
+-- | Creates a 'Keystore' out of a randomly generated temporary file (i.e.
+-- inside your $TMPDIR of choice). Suitable for testing.
+newInMemoryTransientKeystore :: MonadIO m => m Keystore
+newInMemoryTransientKeystore = liftIO $ runIdentityT $ fromKeystore $ do
+    tempDir         <- liftIO getTemporaryDirectory
+    (tempFile, hdl) <- liftIO $ openTempFile tempDir "keystore.key"
+    liftIO $ hClose hdl
+    us <- takeUserSecret tempFile
```
edsko: This needs a comment about why this works with an empty (unparseable) file.

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,194 @@
...
+-- probably don't want, as it will effectively leave you with an empty
+-- 'MVar' and your next call to an operation on the keystore would block
+-- indefinitely.
+-- This has still its use in some teardowns or tests.
+-- Note that this operation will always succeed. Use with care.
+destroyKeystore :: MonadIO m => Keystore -> m ()
+destroyKeystore (Keystore ks) = liftIO $ do
+    (InternalStorage us _st) <- takeMVar ks
+    writeUserSecretRelease us
+    removeFile (getUSPath us)
```
edsko: Leaving this in an empty `MVar` state is not a good idea, if this ever happens the code will break horribly with an opaque "blocked indefinitely" exception. Moreover, there is code duplication here. Let's call `releaseKeystore` first, avoiding code duplication, and then remove the file.

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,194 @@
...
+insert _walletId esk (Keystore ks) =
+    liftIO $ modifyMVar_ ks $ \(InternalStorage us st) -> do
+        return $ if view usKeys us `contains` esk
+                     then InternalStorage us st
+                     else InternalStorage (us & over usKeys (esk :)) st
+    where
+      -- Comparator taken from the old code which needs to hash
+      -- all the 'EncryptedSecretKey' in order to compare them.
+      contains :: [EncryptedSecretKey] -> EncryptedSecretKey -> Bool
+      contains ls k = hash k `elem` map hash ls
```
edsko: Not your fault, but let's add a TODO (or open a ticket): keeping all keys in memory until the node gets shut down is madness. If you lose power or something, you lose all data. This needs to be fixed.

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,194 @@
...
+lookup :: MonadIO m
+       => WalletId
+       -> Keystore
+       -> m (Maybe EncryptedSecretKey)
+lookup (WalletIdHdRnd walletId) (Keystore ks) =
+    liftIO $ withMVar ks $ \(InternalStorage us st) ->
+        return $ case st of
+            StorageInitialised ->
+                Data.List.find (\k -> eskToHdRootId k == walletId) (us ^. usKeys)
+            StorageReleased      -> Nothing
```
edsko: We should distinguish between "key not found" versus "you already shut down the keystore you idiot".

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,194 @@
...
+{-------------------------------------------------------------------------------
+  Deleting things from the keystore
+-------------------------------------------------------------------------------}
+delete :: MonadIO m
+       => WalletId
+       -> Keystore
+       -> m ()
+delete walletId (Keystore ks) = do
+    mbEsk <- lookup walletId (Keystore ks)
+    liftIO $ modifyMVar_ ks $ \(InternalStorage us st) ->
```
edsko: Let's avoid the race condition here. Split `lookup` into a top-level function and an internal pure one, then here use the internal one with a single `modifyMVar`.

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,194 @@
...
+                let erase = Data.List.deleteBy (\a b -> hash a == hash b)
+                let us' = maybe us (\esk -> us & over usKeys (erase esk)) mbEsk
+                return (InternalStorage us' st)
+            StorageReleased -> return (InternalStorage us st)
+
+-- | Returns all the 'EncryptedSecretKey' known to this 'Keystore'.
+toList :: MonadIO m => Keystore -> m [(WalletId, EncryptedSecretKey)]
+toList (Keystore ks) =
+    liftIO $ withMVar ks $ \(InternalStorage us _st) ->
+        return (over mapped (\k -> (WalletIdHdRnd (eskToHdRootId k), k)) (us ^. usKeys))
```
edsko: Why `over mapped` rather than `map`?

wallet-new/test/unit/Test/Spec/Keystore.hs
```diff
@@ -0,0 +1,111 @@
...
+import           Util.Buildable (ShowThroughBuild (..))
+
+{-# ANN module ("HLint: ignore Reduce duplication" :: Text) #-}
+
+-- | Creates, operate on a keystore and finally destroys it.
+withKeystore :: (MonadIO m, MonadMask m) => (Keystore -> m a) -> m a
+withKeystore action =
+    bracket (Keystore.newInMemoryTransientKeystore)
+            (\ks -> Keystore.releaseKeystore ks >>
+                    Keystore.destroyKeystore ks
```
edsko: Destroy also releases, so can just call `destroyKeystore`. Might make sense to call it `releaseAndDestroy`?

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,194 @@
...
+
+-- | Release the resources associated with this 'Keystore'.
+-- This function is idempotent and can be called multiple times.
+releaseKeystore :: MonadIO m => Keystore -> m ()
+releaseKeystore (Keystore ks) = liftIO $ modifyMVar ks $ \(InternalStorage us st) ->
+    case st of
+         StorageReleased      -> return ((InternalStorage us StorageReleased), ())
+         StorageInitialised   -> do
+             writeUserSecretRelease us
+             return ((InternalStorage us StorageReleased), ())
```
edsko: If would be nicer to export a `bracketKeystore` instead.

edsko: Indeed, if it is true that _when_ the wallet is initialized using the legacy keystore, our release process here should be different, then it's even more important to only export `bracket` style functions, to guarantee that we do the right thing.

wallet-new/test/unit/Test/Spec/Keystore.hs
```diff
@@ -0,0 +1,111 @@
...
+                    Keystore.delete wid ks
+                    mbKey <- Keystore.lookup wid ks
+                    (fmap hash mbKey) `shouldBe` Nothing
+
+        prop "Deletion of keys is idempotent" $ monadicIO $ do
+            forAllM genKeypair $ \(STB wid, STB esk) -> run $ do
+                withKeystore $ \ks -> do
+                    Keystore.insert wid esk ks
+                    Keystore.delete wid ks
+                    Keystore.delete wid ks
```
edsko: I'm not convinced this is the property we want. I feel that the second call should throw an exception.

wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -295,7 +307,9 @@ instance HasPrimKey HdAddress where
     type PrimKey HdAddress = HdAddressId
     primKey = _hdAddressId
 
-type HdRootIxs    = '[]
+-- | An 'HdRoot' can be accessed via its 'Core.Address', which is the
+-- Cardano 'Address' derived by the PublicKey associated with it.
+type HdRootIxs    = '[Core.Address]
```
edsko: This index is superfluous, it adds nothing on top of the primary key index. Perhaps the confusion stems from the name. How about we rename this to `HdRootAdditionalIxs` (and similarly for the other two)?

wallet-new/src/Cardano/Wallet/Kernel/DB/HdWallet.hs
```diff
@@ -431,6 +447,14 @@ assumeHdRootExists _id = return ()
 assumeHdAccountExists :: HdAccountId -> Update' HdWallets e ()
 assumeHdAccountExists _id = return ()
 
+{-------------------------------------------------------------------------------
+  General-utility functions
+-------------------------------------------------------------------------------}
+
+-- | Computes the 'HdRootId' from the given 'EncryptedSecretKey'.
```
edsko: We need a comment here (or a reference to a comment) why `Core.makePubKeyAddressBoot` is okay here ("held together by ducktape").

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,239 @@
...
+
+module Cardano.Wallet.Kernel.Keystore (
+      Keystore -- opaque
+      -- * Constructing a keystore
+    , bracketKeystore
+    , bracketLegacyKeystore
+    -- * Destroying a keystore (something you rarely should do)
+    , releaseAndDestroyKeystore
+    -- * Releasing a keystore and its associated resources
+    , releaseKeystore
```
edsko: We should not have `releaseKeystore` exported anymore (it breaks the invariants established by the brackets).

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,239 @@
...
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
+
+module Cardano.Wallet.Kernel.Keystore (
+      Keystore -- opaque
+      -- * Constructing a keystore
+    , bracketKeystore
+    , bracketLegacyKeystore
+    -- * Destroying a keystore (something you rarely should do)
+    , releaseAndDestroyKeystore
```
edsko: I think this function doesn't work so well anymore, it doesn't make sense to have destruction from _within_ the bracket, and from _outside_ the bracket we have no keystore to restore.

edsko: I think the correct API here is to have the bracket functions themselves take an argument telling it whether or not the file should be deleted on release.

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,239 @@
...
+bracketLegacyKeystore us withKeystore =
+    bracket (newLegacyKeystore us)
+            (\_ -> return ()) -- Leave teardown to the legacy wallet
+            withKeystore
+
+-- | Creates a 'Keystore' out of a randomly generated temporary file (i.e.
+-- inside your $TMPDIR of choice). Suitable for testing.
+-- We don't offer a 'bracket' style here as the teardown is irrelevant, as
+-- the file is disposed automatically from being created into the
+-- OS' temporary directory.
```
edsko: Let's rename this to `newTestKeystore`, and bang in a comment that this function has some theoretical race conditions with respect to filenames and should only be used in tests.

wallet-new/src/Cardano/Wallet/Kernel/Keystore.hs
```diff
@@ -0,0 +1,239 @@
...
+import           System.Wlog (CanLog (..), HasLoggerName (..), LoggerName (..),
+                     logMessage)
+
+import           Cardano.Wallet.Kernel.DB.HdWallet (eskToHdRootId)
+import           Cardano.Wallet.Kernel.Types (WalletId (..))
+
+-- Internal storage necessary to smooth out the legacy 'UserSecret' API.
+data InternalStorage =
+      StorageInitialised !UserSecret
+    | StorageReleased
```
edsko: There is no need for this anymore, since there is no explicit release anymore.

adinapoli-iohk:
@rvl @jmitchell This PR have been waiting for Buildkite to spawn a worker for the past ~3 hours at this stage. As this seems to be a common problem and I am very keen on merging this PR, _and I have built the PR myself on a Darwin machine, I hereby grant the right to you Devops to bring shame to me and my dynasty if merging this PR would result in a broken `develop` branch 😁

PR #3197, opened 2018-07-05
[CO-334] Implement remaining V0 endpoints to V1

parsonsmatt:
## Description

This topic branch tracks the implementation of the last few V0 endpoints in V1, under a special `daedalus` namespace.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CO-334

## Type of change

## Developer checklist

## Testing checklist


wallet-new/src/Cardano/Wallet/API/V1/Internal.hs
```diff
@@ -0,0 +1,85 @@
...
+-- data CPaperVendWalletRedeem = CPaperVendWalletRedeem
+--     { pvWalletId     :: !CAccountId
+--     , pvSeed         :: !Text -- TODO: newtype!
+--     , pvBackupPhrase :: !(CBackupPhrase 9)
+--     } deriving (Show, Generic)
+
+-- The redemption datatypes differ only by the backup phrase. Otherwise the
+-- wallet ID and seed are both present. Potentially we can unify these into the
+-- same endpoint. Seems like a query parameter like `paper_vend_passphrase=...`
+-- would do nicely.
```
KtorZ: I agree. I see no point for two endpoints here. I'd rather unify them and have a switch kind of parameter to target the right behavior. Perhaps it's not even needed and we can, at parsing-time, identify which structure is being passed and pick the right behavior after that.

wallet-new/src/Cardano/Wallet/API/V1/Internal.hs
```diff
@@ -0,0 +1,85 @@
...
+-- redeemAdaPaperVend
+--     :: MonadWalletTxFull ctx m
+--     => ProtocolMagic
+--     -> (TxAux -> m Bool)
+--     -> PassPhrase
+--     -> CPaperVendWalletRedeem
+--     -> m CTx
+-- data CPaperVendWalletRedeem = CPaperVendWalletRedeem
+--     { pvWalletId     :: !CAccountId
+--     , pvSeed         :: !Text -- TODO: newtype!
```
KtorZ: Yes please!

wallet-new/src/Cardano/Wallet/API/V1/Internal.hs
KtorZ: Note that, I am not sure this `redeem` should be in the _internal_ part of the API. We probably want to have them in the public-facing part?

KtorZ:
Note (wasn't maybe really clear from the requirements). What we use to call `development API` and this new `internal API` are the same thing. 
 "Development" just conveys a wrong idea of what it is. Therefore, we have to unify this with:

https://github.com/input-output-hk/cardano-sl/blob/develop/wallet-new/src/Cardano/Wallet/API/Development.hs

adinapoli-iohk:
> What we use to call development API and this new internal API are the same thing.

I agree in principle. The way I see it, there are two subtle faces of the same coin, as the original spirit of that `Development` endpoints was to capture that there were endpoints what frontend folks were using only for the sake of testing or troubleshooting the wallet, the endpoint to dump the full `WalletStorage` being the perfect example. 

I think that in principle though unifying everything into only one category make sense, although I'm not sure how this would work in practice: for example (if I recall correctly) we do not even expose certain dev endpoints if the `--wallet-debug` option is not set.

KtorZ:
@adinapoli-iohk Yes, and this has still the same meaning / behavior although we will also expose the documentation for these endpoints somewhere. These `update apply` and `update postpone` things aren't especially used in testing but they fall in the category of frontend-specific endpoints that are irrelevant for exchanges.

parsonsmatt:
Nice job with the commit history @KtorZ :) I'll get this merged once CI passes.

KtorZ:
CI was green, but something was merged to develop conflicting in `generate-swagger-file/Main.hs` ... Rebased, fixed, aaaaaaaaand going for another 2-hour trip :)

parsonsmatt:
Everything is green, no conflicts, GOTTA GO FAST

PR #3210, opened 2018-07-09
[CO-324] Accounts per-field endpoints

KtorZ:
## Description

There are needs for exchanges to retrieve only the balance of a given account. A call to `/api/v1/wallets/{walletId}/accounts/{accountId} already contains this field but is unpractical for exchanges since they usually have accounts with MANY addresses, resulting in a huge response from the API.

## Linked issue

[CO-324](https://iohk.myjetbrains.com/youtrack/issue/CO-324)

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/integration/AccountSpecs.hs
+++ b/wallet-new/integration/AccountSpecs.hs
@@ -1,0 +1,57 @@
+{-# LANGUAGE RankNTypes    #-}
+{-# LANGUAGE TupleSections #-}
+
+module AccountSpecs (accountSpecs) where
+
+import           Universum
+
+import           Cardano.Wallet.API.Indices (accessIx)
+import           Cardano.Wallet.Client.Http
+import           Control.Lens
+import           Pos.Core.Common (mkCoin)
+import           Test.Hspec
+import           Util
+
+import qualified Pos.Core as Core
+import qualified Prelude
+
+
+accountSpecs :: WalletRef -> WalletClient IO -> Spec
+accountSpecs _ wc =
+    describe "Accounts" $ do
+        it "can retrieve only an accounts balance" $ do
+            let zero = V1 (mkCoin 0)
+            (Wallet{..}, Account{..}) <- randomAccount wc
+            eresp <- getAccountBalance wc walId accIndex
+
+            partialAccount <- wrData <$> eresp `shouldPrism` _Right
+            partialAccount `shouldBe` AccountBalance zero
+
+        it "can retrieve only an account's addresses" $ do
+            pair@(Wallet{..}, Account{..}) <- randomAccount wc
+            addresses <- createAddresses wc 10 pair
+            let addr = Prelude.head addresses
+            let tests =
+                    [ PaginationTest (Just 1) (Just 5) NoFilters NoSorts
+                        (expectNAddresses 5)
+                    , PaginationTest (Just 1) (Just 5) (filterByAddress addr) NoSorts
+                        (expectExactlyAddresses [addr])
+                    , PaginationTest (Just 2) (Just 5) (filterByAddress addr) NoSorts
+                        (expectExactlyAddresses [])
+                    ]
+
+            forM_ tests $ \PaginationTest{..} -> do
+                eresp <- getAccountAddresses wc walId accIndex page perPage filters sorts
+                expectations . acaAddresses . wrData =<< eresp `shouldPrism` _Right
+  where
+    filterByAddress :: WalletAddress -> FilterOperations WalletAddress
+    filterByAddress addr =
+        FilterOp (FilterByIndex $ accessIx @_ @(V1 Core.Address) addr) NoFilters
+
+    expectNAddresses :: Int -> [WalletAddress] -> IO ()
+    expectNAddresses n addrs =
+        length addrs `shouldBe` n
+
+    expectExactlyAddresses :: [WalletAddress] -> [WalletAddress] -> IO ()
+    expectExactlyAddresses as bs =
+        sort as `shouldBe` sort bs
--- a/wallet-new/src/Cardano/Wallet/API/Indices.hs
+++ b/wallet-new/src/Cardano/Wallet/API/Indices.hs
@@ -51,6 +51,10 @@ instance ToIndex Transaction (V1 Core.Timestamp) where
     toIndex _ = fmap V1 . Core.parseTimestamp
     accessIx Transaction{..} = txCreationTime
 
+instance ToIndex WalletAddress (V1 Core.Address) where
+    toIndex _ = fmap V1 . either (const Nothing) Just . Core.decodeTextAddress
+    accessIx WalletAddress{..} = addrId
+
 -- | A type family mapping a resource 'a' to all its indices.
 type family IndicesOf a :: [*] where
     IndicesOf Wallet        = WalletIxs
@@ -117,7 +121,7 @@ type family IndexToQueryParam resource ix where
     IndexToQueryParam Wallet  WalletId                = "id"
     IndexToQueryParam Wallet  (V1 Core.Timestamp)     = "created_at"
 
-    IndexToQueryParam WalletAddress (V1 Core.Address) = "id"
+    IndexToQueryParam WalletAddress (V1 Core.Address) = "address"
 
     IndexToQueryParam Transaction (V1 Core.TxId)      = "id"
     IndexToQueryParam Transaction (V1 Core.Timestamp) = "created_at"
--- a/wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
@@ -2,10 +2,13 @@ module Cardano.Wallet.API.V1.Accounts where
 
 import           Servant
 
+import           Cardano.Wallet.API.Request
 import           Cardano.Wallet.API.Response
 import           Cardano.Wallet.API.Types
 import           Cardano.Wallet.API.V1.Parameters
 import           Cardano.Wallet.API.V1.Types
+
+import qualified Pos.Core as Core
 
 
 type API
@@ -34,6 +37,9 @@ type API
     :<|> "wallets" :> CaptureWalletId :> "accounts"
           :> CaptureAccountId :> "addresses"
           :> Summary "Retrieve only account's addressees."
+          :> WalletRequestParams
+          :> FilterBy '[V1 Core.Address] WalletAddress
+          :> SortBy '[V1 Core.Address] WalletAddress
           :> Get '[ValidJSON] (WalletResponse AccountAddresses)
     :<|> "wallets" :> CaptureWalletId :> "accounts"
           :> CaptureAccountId :> "balance"
--- a/wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
@@ -42,7 +42,7 @@ type API
           :> SortBy '[V1 Core.Address] WalletAddress
           :> Get '[ValidJSON] (WalletResponse AccountAddresses)
     :<|> "wallets" :> CaptureWalletId :> "accounts"
-          :> CaptureAccountId :> "balance"
+          :> CaptureAccountId :> "amount"
           :> Summary "Retrieve only account's balance."
           :> Get '[ValidJSON] (WalletResponse AccountBalance)
 
--- a/wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
@@ -31,4 +31,13 @@ type API
           :> Summary "Update an Account for the given Wallet."
           :> ReqBody '[ValidJSON] (Update Account)
           :> Put '[ValidJSON] (WalletResponse Account)
+    :<|> "wallets" :> CaptureWalletId :> "accounts"
+          :> CaptureAccountId :> "addresses"
+          :> Summary "Retrieve only account's addressees."
+          :> Get '[ValidJSON] (WalletResponse AccountAddresses)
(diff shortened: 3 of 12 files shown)


wallet-new/src/Cardano/Wallet/API/Indices.hs
```diff
@@ -117,7 +121,7 @@ type family IndexToQueryParam resource ix where
     IndexToQueryParam Wallet  WalletId                = "id"
     IndexToQueryParam Wallet  (V1 Core.Timestamp)     = "created_at"
 
-    IndexToQueryParam WalletAddress (V1 Core.Address) = "id"
+    IndexToQueryParam WalletAddress (V1 Core.Address) = "address"
```
KtorZ: NOTE: Wasn't used before, so there's no breaking change here. I believe `address` makes more sense even though it filters on the `id` field of an address. As a reminder, we represent addresses as follows:

image

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -868,13 +870,25 @@ data Account = Account
...
+-- | Datatype wrapping addresses for per-field endpoint
+newtype AccountAddresses = AccountAddresses
+    { acaAddresses :: [WalletAddress]
+    } deriving (Show, Ord, Eq, Generic)
+
+-- | Datatype wrapping balance for per-field endpoint
+newtype AccountBalance = AccountBalance
+    { acbAmount    :: V1 Core.Coin
+    } deriving (Show, Ord, Eq, Generic)
+
```
KtorZ: This is really open to debate. This will preserve a similar structure than the one used for account. For instance, retrieving the account balance will return a JSON with:

```json
{
    "status": "success",
    "data": {
        "amount": 14
    },
    "metadata": ...
}
```

Another option would be to remove one level of nesting and put the value directly under `data`:

```json
{
    "status": "success",
    "data": 14,
    "metadata": ...
}
```

I am really open to discuss this.

Anviking: What is the argument for the flattened one? I have no idea how this would affect usage in various client libraries, maybe there is some specific argument for either one. I wouldn't instinctively argue against the nested approach though.

Anviking: Could one enforce the "keeping things roughly the same" part? When first seeing this PR I came to think about lenses and GraphQL…

adinapoli-iohk: I always thought about `data` as a container of, well, data, so it would feel unnatural to me to go from `data :: Object` to `data :: Value`, if you see what you mean. I suspect I lean towards symmetry and consistency, and as such having `data` always index a JSON object seems like the best approach, to me, but of course, that's just my opinion 😉

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
```diff
@@ -31,4 +34,16 @@ type API
           :> Summary "Update an Account for the given Wallet."
           :> ReqBody '[ValidJSON] (Update Account)
           :> Put '[ValidJSON] (WalletResponse Account)
+    :<|> "wallets" :> CaptureWalletId :> "accounts"
+          :> CaptureAccountId :> "addresses"
+          :> Summary "Retrieve only account's addressees."
+          :> WalletRequestParams
+          :> FilterBy '[V1 Core.Address] WalletAddress
+          :> SortBy '[V1 Core.Address] WalletAddress
```
adinapoli-iohk: Does sorting on address really make sense? They are just a bunch of base58-encoded strings after all, so I'm wondering whether or not we should allow sorting, as it buys virtually nothing 😉

wallet-new/src/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -654,6 +654,27 @@ curl -X GET \
 $readAccounts
 ```
 
+Partial Representations
```
adinapoli-iohk: Good lad for adding documentation! 😻

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -868,13 +870,25 @@ data Account = Account
...
     } deriving (Show, Ord, Eq, Generic)
 
+-- | Datatype wrapping addresses for per-field endpoint
+newtype AccountAddresses = AccountAddresses
+    { acaAddresses :: [WalletAddress]
+    } deriving (Show, Ord, Eq, Generic)
+
+-- | Datatype wrapping balance for per-field endpoint
+newtype AccountBalance = AccountBalance
+    { acbAmount    :: V1 Core.Coin
```
adinapoli-iohk: Bikeshedding -- `acbAmount` or `acbBalance`? 🚴 

I suspect the latter reads better in the output JSON -- what do you think?

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: I agree that consistency and convention are important (especially without types), and right now, as far as JS is concerned, we've got a type like `{ data :: Object }`. Deviating from that would a potential source of runtime errors.

wallet-new/integration/AccountSpecs.hs
```diff
@@ -0,0 +1,57 @@
...
+            addresses <- createAddresses wc 10 pair
+            let addr = Prelude.head addresses
+            let tests =
+                    [ PaginationTest (Just 1) (Just 5) NoFilters NoSorts
+                        (expectNAddresses 5)
+                    , PaginationTest (Just 1) (Just 5) (filterByAddress addr) NoSorts
+                        (expectExactlyAddresses [addr])
+                    , PaginationTest (Just 2) (Just 5) (filterByAddress addr) NoSorts
+                        (expectExactlyAddresses [])
+                    ]
```
parsonsmatt: love it

wallet-new/src/Cardano/Wallet/API/Indices.hs
parsonsmatt: man, that does feel really weird. i agree that this is an improvement, but this is confusing.

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
parsonsmatt: It allows pagination based on an indexed part of the datatype, which makes for super efficient querying.

wallet-new/src/Cardano/Wallet/API/V1/LegacyHandlers/Accounts.hs
```diff
@@ -68,3 +70,25 @@ updateAccount wId accIdx accUpdate = do
...
+    :: (V0.MonadWalletLogic ctx m)
+    => WalletId
+    -> AccountIndex
+    -> RequestParams
+    -> FilterOperations WalletAddress
+    -> SortOperations WalletAddress
+    -> m (WalletResponse AccountAddresses)
+getAccountAddresses wId accIdx pagination filters sorts = do
+    resp <- respondWith pagination filters sorts (getAddresses <$> getAccount wId accIdx)
+    return resp { wrData = AccountAddresses . wrData $ resp }
```
parsonsmatt: is `WalletResponse` a functor/foldable? it totally should be!

wallet-new/src/Cardano/Wallet/API/V1/Swagger.hs
```diff
@@ -771,14 +792,16 @@ Make sure to carefully read the section about [Pagination](#section/Pagination)
...
+    createAccount        = decodeUtf8 $ encodePretty $ genExample @(WalletResponse Account)
+    createAddress        = decodeUtf8 $ encodePretty $ genExample @(WalletResponse WalletAddress)
+    createWallet         = decodeUtf8 $ encodePretty $ genExample @(WalletResponse Wallet)
+    readAccounts         = decodeUtf8 $ encodePretty $ genExample @(WalletResponse [Account])
+    readAccountBalance   = decodeUtf8 $ encodePretty $ genExample @(WalletResponse AccountBalance)
+    readAccountAddresses = decodeUtf8 $ encodePretty $ genExample @(WalletResponse AccountAddresses)
+    readAddresses        = decodeUtf8 $ encodePretty $ genExample @(WalletResponse [Address])
+    readFees             = decodeUtf8 $ encodePretty $ genExample @(WalletResponse EstimatedFees)
+    readNodeInfo         = decodeUtf8 $ encodePretty $ genExample @(WalletResponse NodeInfo)
+    readTransactions     = decodeUtf8 $ encodePretty $ genExample @(WalletResponse [Transaction])
```
parsonsmatt: rather than large whitespace changes like this i would prefer an indentation to make the start point equal for each line. minor quibble :)

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
parsonsmatt: Weak vote for `acbBalance` but I think both are fine

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -886,13 +900,33 @@ instance ToSchema Account where
...
 instance Arbitrary Account where
     arbitrary = Account <$> arbitrary
                         <*> arbitrary
                         <*> arbitrary
                         <*> pure "My account"
                         <*> arbitrary
 
+instance Arbitrary AccountAddresses where
+    arbitrary =
+        AccountAddresses <$> arbitrary
```
parsonsmatt: with deriving strategies, this is `deriving newtype (Arbitrary)` :smile:

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: :upside_down_face: ... I went for `acbBalance` at first, because it reads better. Then I notice the fireld was actually named `amount` in the `Account` representation. Therefore, I changed for `acbAmount` as well for consistency.

wallet-new/src/Cardano/Wallet/API/V1/Swagger.hs
KtorZ: And I do agree with you!

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
KtorZ: I am always (without any good reasons) afraid of deriving newtypes automagically :p ...

wallet-new/src/Cardano/Wallet/API/V1/Accounts.hs
adinapoli-iohk: > It allows pagination based on an indexed part of the datatype, which makes for super efficient querying.

@parsonsmatt @KtorZ  Isn't what `FilterBy` buys you already? Unless I am mistaken, note that the type-level sorting & filtering has no effect on the (already-defined) IxSet indices 😉 Once those are defined (as in `instance Indexable ...`) which one you use for filter & sort (or the lack thereof) really doesn't undermine querying efficiency 😉 

Imho keeping `SortBy` here is a mistake, as it would encourage exchanges to sort by addresses, which is an operation terribly wasteful, as there is no meaningful ordering for addresses, as the lexicographical order doesn't really help here. My 2 cents anyway, very happy to be proved wrong! 😉

KtorZ: In a way, pagination without ordering doesn't make much sense to me, does it? Do we even have guarantees with the acid-state db that we will yield the same elements from two successive get queries? Or is it completely non-deterministic on non-indexed items?

Though, I'd rather follow your intuition on that @adinapoli-iohk. Note that the PR has been merged in a feature-branch and not yet to develop, so there's still time to do some minor adjusments!

edsko: Haven't followed the details here but I agree that for pagination it is important to have an indexed attribute available to sort on. Not sure how much control we need to give the user over this, since as @adinapoli-iohk points out, sorting on base58 strings isn't particularly meaningful. Moreover, in the new wallet we clearly distinguish between the primary key and the rest of the data structure and don't require Ord instances for the whole type; in this case, the natural way to sort by "address" would be to sort on its key (which is basically just the pair of integers telling it how that address is derived from the root). So I think it's important that "sort by address" doesn't necessarily mean "sort lexicographically", but rather "give me some kind of pagination I can depend on" -- so that we are free to switch over the sorting to the primary key once we move to the new wallet, rather than being forced to still support this lexicographical sort.

adinapoli-iohk: > In a way, pagination without ordering doesn't make much sense to me, does it?

Indeed, but _I think_ we should be fine in this case, as in the current implementation we simply call `IxSet.toList` which, AFAIK, uses descending ordering, which is a reasonable default, I believe.

My main gripe was that, if we leave that `SortBy` as-it-is, we are really exposing `sort_by=address[DES|ASC]` to the API, which means exchanges could enforce an explicit sorting on the addresses, which, considering the amount of addresses an exchanges has, I would prefer to not pay the explicit price, if possible 😀 

Am I being mental? 😛

KtorZ: I usually say: If users can do something, eventually they will. I wonder if we shouldn't indeed remove that sorting for now, until we have something more meaningful we can actually sort on.

adinapoli-iohk: > If users can do something, eventually they will.

Aye, that is usually my experience -- that's why I have flagged this up 😉

KtorZ: NOTE: Sorting was removed on the main story branch

PR #3215, opened 2018-07-09
[CO-327] Rely on Example type-class for Swagger API schema

Anviking:
## Description

Use `Example` typeclass instead of `Arbitrary` when creating Swagger schema.

1. Import and use `Example` typeclass in `../V1/Types.hs`
instead of `Arbitrary`
2. Resolve circular dependencies
`Errors.hs` -> `Types.hs` -> `Example.hs` -> `Response.hs` -> `Errors.hs`
by moving `Example` instances from `Example.hs` to where the datatypes
are defined, so that `Example.hs` does not need to import `Response.hs`
or `Types.hs`.

Regarding name collisions between `Data.Swagger` and `Cardano.Wallet.API.V1.Swagger.Example`, I suspect `hiding` is the wrong way to go?

## Linked issue

[CO-327](https://iohk.myjetbrains.com/youtrack/issue/CO-327)

## Type of change

## Developer checklist

- [~] I have updated any documentation accordingly, if needed. Documentation changes can be reflected in opening a PR on [cardanodocs.com](https://github.com/input-output-hk/cardanodocs.com), amending the inline [Haddock](https://www.haskell.org/haddock/) comments, any relevant README file or one of the document listed in the [docs](https://github.com/input-output-hk/cardano-sl/tree/develop/docs) directory.

## Testing checklist

- [x] All new and existing tests passed. **Having problems running tests. Want to see CI.**

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/V1/Swagger/Example.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Swagger/Example.hs
@@ -2,22 +2,14 @@ module Cardano.Wallet.API.V1.Swagger.Example where
 
 import           Universum
 
-import           Test.QuickCheck (Arbitrary (..), Gen, listOf1, oneof)
+import           Test.QuickCheck (Arbitrary (..), Gen, listOf1)
 
-import           Cardano.Wallet.API.Response
-import           Cardano.Wallet.API.V1.Types
 import           Cardano.Wallet.Orphans.Arbitrary ()
-import           Data.Default (Default (def))
-import           Node (NodeId (..))
 import           Pos.Arbitrary.Wallet.Web.ClientTypes ()
-import           Pos.Client.Txp.Util (InputSelectionPolicy (..))
-import           Pos.Util.Mnemonic (Mnemonic)
 import           Pos.Wallet.Web.ClientTypes (CUpdateInfo)
 import           Pos.Wallet.Web.Methods.Misc (WalletStateSnapshot (..))
 
 import qualified Data.Map.Strict as Map
-import qualified Pos.Core.Common as Core
-import qualified Pos.Crypto.Signing as Core
 
 
 class Arbitrary a => Example a where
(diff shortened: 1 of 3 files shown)


Anviking:
Hm, CI is failing, but it seems to only be warnings I didn't see in ghci. I will look into this.

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
```diff
@@ -83,12 +83,14 @@ import           Universum
...
 
+import           Cardano.Wallet.API.V1.Swagger.Example (Example, example)
 import           Control.Lens (At, Index, IxValue, at, ix, makePrisms, to, (?~))
 import           Data.Aeson
 import           Data.Aeson.TH as A
 import           Data.Aeson.Types (toJSONKeyText, typeMismatch)
 import qualified Data.Char as C
-import           Data.Swagger as S
+import           Data.Swagger as S hiding (Example, example)
+import qualified Data.Swagger as SE (example)
```
KtorZ: We can keep `S` instead of `SE` here. They refer to the same module anyway :)

wallet-new/src/Cardano/Wallet/API/V1/Swagger/Example.hs
```diff
@@ -39,107 +37,15 @@ instance Example a => Example (Maybe a) where
...
-instance Example Payment where
-    example = Payment <$> example
-                      <*> example
-                      <*> example -- TODO: will produce `Just groupingPolicy`
-                      <*> example
 
 instance Example WalletStateSnapshot
 
 
-
```
KtorZ: What about removing this whole file, moving the remaining instances in places where they belong :thinking:  ?

Anviking: @KtorZ The remaining ones are 
- `WalletSnapshot` from *../wallet/src/Pos/Wallet/Web/Methods/Misc.hs*
- `CUpdateInfo` from *../wallet/src/Pos/Wallet/Web/ClientTypes/Types.hs*
- `Mnemonic` from *../wallet/src/Pos/Util/Mnemonic*

Can we really depend on *wallet-new* from the old *wallet*?

Then, also, the class itself, and instances for `()`, `NonEmpty`, `[]`, `Maybe` and `Map` need to be somewhere.

So it would seem difficult…?

KtorZ: Hmmmm... Fair point. Not sure why we have this `Example` in the `wallet-new` in a first place since it's quite an isolated module. But it doesn't make sense to import anything from the wallet in the old one,

wallet-new/src/Cardano/Wallet/API/V1/Types.hs
Anviking: Actually, there was a reason. This way I could simply copy-paste the `Example`-instances and have `example` refer to `Swagger.Example.example`, and accessing `Data.Swagger.example` as `SE.example`.

So we *can*, but have to make 50+ references to `Example` and `example` qualified. Feels a little bit wrong, but I guess explicitly is good?

KtorZ: Sorry for the delay, but what I meant was simply to have:

```hs
import           Data.Swagger hiding (Example, example)
import qualified Data.Swagger as S
```

Even though, I'd rather list everything we use from `Data.Swagger` if it's realistic.

Anviking: Seems borderline unrealistic (2k LoC and lots of things from `Data.Swagger`). Should I spend time on it? Could perhaps get [importify](https://github.com/serokell/importify) working.

Anviking: ("unrealistic"  as in I got to like 10-20 symbols and then gave up, and not sure if there are 25 in total or 100)

KtorZ:
Weird failure on appveyor, merging into the feature branch anyway. This doesn't seem related to the PR itself :+1:

PR #3217, opened 2018-07-09
[CDEC-427] Move Arbitrary instances from ssc to ssc-test

intricate:
## Description

As it stands, the Arbitrary instances for data types in `ssc` all exist in `ssc` when they should actually be exposed from a separate package, `ssc-test`.

In addition to this, `Pos.Arbitrary.Ssc` was actually used within `Pos.Ssc.Worker`. It was used to [generate a random Opening](https://github.com/input-output-hk/cardano-sl/blob/develop/ssc/src/Pos/Ssc/Worker.hs#L235) in the event that we call [`onNewSlotOpening`](https://github.com/input-output-hk/cardano-sl/blob/develop/ssc/src/Pos/Ssc/Worker.hs#L206) while passing an instance of [`SscOpeningParams`](https://github.com/input-output-hk/cardano-sl/blob/develop/ssc/src/Pos/Ssc/Behavior.hs#L36) constructed with [`SscOpeningWrong`](https://github.com/input-output-hk/cardano-sl/blob/develop/ssc/src/Pos/Ssc/Behavior.hs#L39). This was changed not only because the `Arbitrary` instance deterministically generated an `Opening` (does anyone know whether there was any kind of security risk in this?), but also because `ssc` can't depend on `ssc-test` (circular dependency).

## Linked issue

[https://iohk.myjetbrains.com/youtrack/issue/CDEC-427](https://iohk.myjetbrains.com/youtrack/issue/CDEC-427)
## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/block/cardano-sl-block.cabal
+++ b/block/cardano-sl-block.cabal
@@ -154,6 +154,7 @@ test-suite test
                      , cardano-sl-crypto-test
                      , cardano-sl-delegation-test
                      , cardano-sl-ssc
+                     , cardano-sl-ssc-test
                      , cardano-sl-txp-test
                      , cardano-sl-update
                      , cardano-sl-update-test
--- a/block/test/Test/Pos/Block/Arbitrary.hs
+++ b/block/test/Test/Pos/Block/Arbitrary.hs
@@ -29,8 +29,6 @@ import           Test.QuickCheck (Arbitrary (..), Gen, choose, suchThat,
 import           Test.QuickCheck.Arbitrary.Generic (genericArbitrary,
                      genericShrink)
 
-import           Pos.Arbitrary.Ssc (SscPayloadDependsOnSlot (..), genSscPayload,
-                     genSscPayloadForSlot)
 import           Pos.Binary.Class (biSize)
 import qualified Pos.Block.Logic.Integrity as T
 import           Pos.Block.Slog (SlogUndo)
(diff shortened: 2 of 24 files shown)


ssc/src/Pos/Ssc/Worker.hs
```diff
@@ -202,6 +201,16 @@ onNewSlotCommitment pm slotId@SlotId {..} sendCommitment
         sscProcessOurMessage (sscProcessCommitment pm comm)
         sendOurData sendCommitment CommitmentMsg comm siEpoch 0
 
+-- | Generate a random Opening.
+randomOpening :: IO Opening
+randomOpening = snd <$> secureRandCommitmentAndOpening
+  where
+    secureRandCommitmentAndOpening = runSecureRandom $ do
+        t       <- randomNumberInRange 3 10
+        n       <- randomNumberInRange (t*2-1) (t*2)
```
erikd: That range does not look right :). It looks to me like the range only has two valid values.

erikd: I suspect that its actually supposed to be
```
       n       <- randomNumberInRange (2 ^ (t -1)) (2 ^ t)
```
but then the value of `t` should be restricted more than it is.

ssc/src/Pos/Ssc/Worker.hs
```diff
@@ -232,7 +241,7 @@ onNewSlotOpening pm params SlotId {..} sendOpening
         mbOpen' <- case params of
             SscOpeningNone   -> pure Nothing
             SscOpeningNormal -> pure (Just open)
-            SscOpeningWrong  -> Just <$> liftIO (QC.generate QC.arbitrary)
+            SscOpeningWrong  -> Just <$> liftIO randomOpening
```
parsonsmatt: This is a lot better, certainly. But.. *why* are we generating a random opening here? This seems deeply weird to me.

intricate: I'm going to be digging into this today to really understand.

PR #3232, opened 2018-07-12
Matt noonan/size estimates

matt-noonan:
## Description

This PR implements the "static size estimate" work from CBR-318. The idea is to provide a way to compute he upper- and lower-bounds on the encoded size for a given type.

To get this capability, I've added a `encodedSizeExpr` method to the `Bi` typeclass. At some type `a`, this method returns a symbolic expression describing the range of sizes that an encoded `a` could have. `encodedSizeExpr` has a default implementation that just returns a symbolic representation of the type: something like "the size of `a` is "the size of `a`"". I have only provided more specific implementations for basic types and whatever was needed to get estimates for `TxAux`.

The expressions are represented by the `Size` type. A variety of utility functions let you carry out basic operations like simplification, evaluation, forcing "thunks", and substituting values in the expressions.

The tests work by randomly generating values `x :: Bi a => a`, then comparing the actual encoded size of `x` to `encodedSizeExpr @a`. This can involve some annoying fiddling around, e.g. to set the number of `TxIn`s to the same amount as are in the generated `x`.

## Linked issue

CBR-318

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/binary/cardano-sl-binary.cabal
+++ b/binary/cardano-sl-binary.cabal
@@ -35,6 +35,8 @@ library
                      , formatting
                      , hashable
                      , lens
+                     , micro-recursion-schemes
+                     , mtl
                      , safecopy
                      , safe-exceptions
                      , serokell-util
--- a/binary/cardano-sl-binary.cabal
+++ b/binary/cardano-sl-binary.cabal
@@ -89,6 +89,7 @@ test-suite test
   other-modules:
                        Spec
                        Test.Pos.Binary.BiSerialize
+                       Test.Pos.Binary.BiSizeBounds
                        Test.Pos.Binary.Cbor.CborSpec
                        Test.Pos.Binary.Helpers
                        Test.Pos.Binary.Helpers.GoldenRoundTrip
--- a/binary/cardano-sl-binary.cabal
+++ b/binary/cardano-sl-binary.cabal
@@ -41,6 +41,7 @@ library
                      , tagged
                      , template-haskell
                      , text
+                     , text-format
                      , th-utilities
                      , time-units
                      , universum
--- a/binary/cardano-sl-binary.cabal
+++ b/binary/cardano-sl-binary.cabal
@@ -124,6 +124,7 @@ test-suite test
                      , quickcheck-instances
                      , safecopy
                      , serokell-util >= 0.1.3.4
+                     , tagged
                      , template-haskell
                      , text
                      , formatting
--- a/binary/src/Pos/Binary/Class/Core.hs
+++ b/binary/src/Pos/Binary/Class/Core.hs
@@ -5,8 +5,11 @@
 {-# LANGUAGE GADTs            #-}
 {-# LANGUAGE LambdaCase       #-}
 {-# LANGUAGE DeriveFunctor    #-}
+{-# LANGUAGE CPP              #-}
 
 -- | Bi typeclass and most basic functions.
+
+#include "MachDeps.h"
 
 module Pos.Binary.Class.Core
     ( Bi(..)
@@ -30,13 +33,12 @@ module Pos.Binary.Class.Core
     -- * Utils
     , toCborError
     , cborError
-
+    , withWordSize
+    
     , Range(..)
     , szEval
     , Size
     , Length(..)
-    , punch
-    , plug
     , isTodo
     , szCases
     , szLazy
@@ -58,6 +60,7 @@ import qualified Codec.CBOR.Write as CBOR.Write
 import qualified Data.Binary as Binary
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as BS.Lazy
+import qualified Data.Char as Char
 import           Data.Coerce
 import           Data.Fixed (Fixed (..), Nano)
 import qualified Data.HashMap.Strict as HM
@@ -71,6 +74,7 @@ import           Data.Time.Units (Microsecond, Millisecond)
 import           Data.Typeable (typeRep, TypeRep)
 import qualified Data.Vector as Vector
 import qualified Data.Vector.Generic as Vector.Generic
+import           Foreign.Storable (sizeOf)
 import qualified GHC.Generics as G
 import           Serokell.Data.Memory.Units (Byte, fromBytes, toBytes)
 
@@ -104,6 +108,26 @@ matchSize :: Int -> Text -> Int -> D.Decoder s ()
 matchSize requestedSize lbl actualSize =
   when (actualSize /= requestedSize) $
     cborError (lbl <> " failed the size check. Expected " <> show requestedSize <> ", found " <> show actualSize)
+
+withSize :: Integral s => s -> a -> a -> a -> a -> a -> a
+withSize s a1 a2 a3 a4 a5 =
+    if | s <= 0x17 -> a1
+       | s <= 0xff -> a2
+       | s <= 0xffff -> a3
+#if WORD_SIZE_IN_BITS == 64
+       | s <= 0xffffffff -> a4
+       | otherwise -> a5
+#elif WORD_SIZE_IN_BITS == 32
+       | otherwise -> a4
+#define ARCH_32bit
+#else
+#error expected WORD_SIZE_IN_BITS to be 32 or 64
+#endif
+
+-- | Compute size of a @'Word'@, this is used for computation of @'Word8'@, ..,
+-- @'Word32'@ encoded size.
+withWordSize :: (Integral s, Integral a) => s -> a
+withWordSize s = withSize s 1 2 3 5 9
 
 ----------------------------------------
 
@@ -173,7 +197,7 @@ instance Bi Char where
     decodeList    = do txt <- D.decodeStringCanonical
                        return (toString txt) -- unpack lazily
 
-    encodedSizeExpr _ pxy = encodedSizeRange pxy
+    encodedSizeExpr _ pxy = encodedSizeRange (Char.ord <$> pxy)
     encodedListSizeExpr size _ =
         let bsLength = size (Proxy @(Length [Char])) * size (Proxy @Char)
         in  bsLength + apMono "(withSize _ 1 2 3 4 5)" (\x -> withSize x 1 2 3 4 5) bsLength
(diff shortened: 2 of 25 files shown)


adinapoli-iohk:
@matt-noonan Regardless from the PR content (which I have yet to review) would you please prefix all your commits with the `[CBR-318]` tag, as per the new practices enforced by our Release Manager? Thanks a lot 😉

binary/src/Pos/Binary/Class/Core.hs
```diff
@@ -82,6 +112,26 @@ matchSize requestedSize lbl actualSize =
   when (actualSize /= requestedSize) $
     cborError (lbl <> " failed the size check. Expected " <> show requestedSize <> ", found " <> show actualSize)
 
+withSize :: Integral s => s -> a -> a -> a -> a -> a -> a
```
adinapoli-iohk: Out loud thought -- as this seems to be used only in two places, would be possible/better to unroll the actual calls into a sort of static map, maybe without relying on CPP at all? I suspect the answer is no, but feel free to me as a rubber duck here 😉

matt-noonan: This function started life in @coot 's CBR-96 branch, where it gets considerably more use. So I think it should stay, or else whatever I do here will eventually become redundant if/when CBR-96 is merged.

binary/cardano-sl-binary.cabal
```diff
@@ -37,13 +37,15 @@ library
                      , half
                      , hashable
                      , lens
+                     , micro-recursion-schemes
```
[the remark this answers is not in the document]
matt-noonan: I think so. `micro-recursion-schemes` doesn't have the huge Kmettian dependencies of `recursion-schemes` and is maintained by vmchale (at IOHK), so it isn't a very risky dependency. And without it, `SizeF` will just end up re-implementing a half-baked version of `Fix`, `cata`, etc.

binary/src/Pos/Binary/Class/Core.hs
matt-noonan: I found a bug in the second place where this was used, so I ended up making this static map anyway :)

binary/src/Pos/Binary/Class/Core.hs
```diff
@@ -26,8 +30,27 @@ module Pos.Binary.Class.Core
...
+    , szLazy
+    , szGreedy
+    , szForce
+    , szWithCtx
+    , szSimplify
+    , apMono
+    , szBounds
     ) where
 
+import           Formatting (bprint, build, shown, string, (%))
```
parsonsmatt: style nit: this isn't the right place for this import, it should be below in the non-Prelude and non-local dependencies

matt-noonan:
WIP no longer, this now builds and all tests pass (locally, still waiting for CI to wrap up)

core/src/Pos/Core/Common/AddrStakeDistribution.hs
```diff
@@ -35,9 +35,9 @@ data AddrStakeDistribution
     -- has a value, portion of this value is stake). The constructor
     -- is unsafe because there are some predicates which must hold:
     --
-    -- • the sum of portions must be @maxBound@ (basically 1);
-    -- • all portions must be positive;
-    -- • there must be at least 2 items, because if there is only one item,
+    --   the sum of portions must be @maxBound@ (basically 1);
+    --   all portions must be positive;
+    --   there must be at least 2 items, because if there is only one item,
```
parsonsmatt: is your editor automatically deleting all non-ASCII characters?

core/src/Pos/Core/Common/AddrSpendingData.hs
```diff
@@ -128,6 +138,6 @@ addrSpendingDataToType =
 
 -- Define these at the end of the file to avoid TH staging issues.
 deriveSafeCopySimple 0 'base ''AddrSpendingData
-deriveSafeCopySimple 0 'base ''AddrType -- ☃
+deriveSafeCopySimple 0 'base ''AddrType --
```
parsonsmatt: goodbye ☃

core/src/Pos/Core/Common/AddrStakeDistribution.hs
matt-noonan: Looks like my emacs session was holding on to some non-IOHK version of `stylish-haskell`. Nothing re-starting emacs can't fix!

core/src/Pos/Core/Common/AddrSpendingData.hs
adinapoli-iohk: Goodbye ⛄️

binary/src/Pos/Binary/Class/Core.hs
```diff
@@ -82,6 +109,15 @@ matchSize requestedSize lbl actualSize =
   when (actualSize /= requestedSize) $
     cborError (lbl <> " failed the size check. Expected " <> show requestedSize <> ", found " <> show actualSize)
 
+-- | Compute encoded size of an integer.
```
edsko: I think this code is only valid for unsigned values, but you're using it also for signed values below.

binary/src/Pos/Binary/Class/Core.hs
```diff
@@ -659,3 +760,252 @@ instance (i ~ G.C, GSerialiseProd f) => GSerialiseSum (G.M1 i c f) where
...
+
+instance Buildable (Fix SizeF) where
+    build x = bprint build (unfix x)
+
+-- | Create a case expression from individual cases.
+szCases :: [Case Size] -> Size
+szCases = Fix . CasesF
+
+-- | An individual labeled case.
+data Case t = Case String t deriving (Typeable, Functor)
```
edsko: Would it make sense to have an ADT instead of a string here, to give consumers of these sizes more of a hint what these cases represent? For instance, I think you use the strings `minBound` and `maxBound` in various places.. it  might be useful for consumers to _know_ that, so that they can derive best and worst cases (just as one example).

binary/src/Pos/Binary/Class/Core.hs
```diff
@@ -659,3 +760,252 @@ instance (i ~ G.C, GSerialiseProd f) => GSerialiseSum (G.M1 i c f) where
...
+caseValue :: Case t -> t
+caseValue (Case _ x) = x
+
+instance Buildable t => Buildable (Case t) where
+  build (Case n x) = bprint (string % "=" % build) n x
+
+-- | A range of values. Should satisfy the invariant @forall x. lo x <= hi x@.
+data Range b = Range { lo :: b, hi :: b }
+
+instance Num b => Num (Range b) where
```
edsko: Does this instance really make sense? I'm not sure what it means to add, multiply and subtract ranges pointwise. Is that actually used?

binary/src/Pos/Binary/Class/Core.hs
```diff
@@ -659,3 +760,252 @@ instance (i ~ G.C, GSerialiseProd f) => GSerialiseSum (G.M1 i c f) where
...
+        AddF x y  -> binOp (+) x y
+        MulF x y  -> binOp (*) x y
+        SubF x y  -> binOp (-) x y
+        NegF x    -> unOp negate x
+        AbsF x    -> unOp abs x
+        SgnF x    -> unOp signum x
+        ApF _ f (Right x) -> Right (Range { lo = f (lo x), hi = f (hi x) })
+        ApF n f (Left x)  -> Left  (apMono n f x)
+
+    normalize = id -- TODO
```
edsko: Stray TODO or not needed?

binary/src/Pos/Binary/Class/Core.hs
```diff
@@ -659,3 +760,252 @@ instance (i ~ G.C, GSerialiseProd f) => GSerialiseSum (G.M1 i c f) where
...
+    | MulF t t -- ^ Product of two sizes.
+    | SubF t t -- ^ Difference of two sizes.
+    | AbsF t   -- ^ Absolute value of a size.
+    | NegF t   -- ^ Negation of a size.
+    | SgnF t   -- ^ Signum of a size.
+    | CasesF [Case t] -- ^ Case-selection for sizes. Used for sum types.
+    | ValueF Byte     -- ^ A constant value.
+    | ApF String (Byte -> Byte) t
+      -- ^ Application of a monotonic function to a size.
+    | forall a. Bi a => TodoF (forall x. Bi x => Proxy x -> Size) (Proxy a)
```
edsko: I'm not sure I understand this `TodoF` constructor --- that is, I don't quite understand why it's needed and how it's used. I think this needs a bit more of an explanation.

binary/src/Pos/Binary/Class/Core.hs
matt-noonan: This would be great, but slightly messy. The issue is that some cases are just named by purpose ("minBound" etc), while others are the actual constructor names used in a sum type (for example, "PubKeyASD", "ScriptASD", "RedeemASD" for `Bi AddrSpendingData`).

I could cover all of this with some ADT along the lines of

```haskell
data CaseLabel
    = MinBound
    | MaxBound
    | Constructor Con -- from TH?
```

but it isn't clear to me if the benefits outweigh the complexity. It still wouldn't let you do something like say "I'm only interested in the lower bound".

binary/src/Pos/Binary/Class/Core.hs
matt-noonan: Thanks, it looks like a bug was lingering here. Fixed.

binary/src/Pos/Binary/Class/Core.hs
matt-noonan: Yes, it is used in `szSimplify` to fold a `Size` up into a `Range Byte`. The `AddF` constructor for `Size` turns into `(+)` for `Range Byte`, and so on.

The various operations aren't defined pointwise, but as set operations. For example, `{ x | lo <= x <= hi } + `{ x' | lo' <= x' <= hi' }` is defined as `{ x + x' | lo <= x <= hi, lo' <= x' <= hi' }`. But it turns out that you can reproduce all of these calculations as calculations on the endpoints.

I think I should add an implementation for `abs` for completeness. As usual, I'm not sure if I can write a lawful `signum`.

matt-noonan: Initially this instance was written under the assumption that the range only contained non-negative values, but that really doesn't make sense given the implementation of `negate`. So I went ahead and implemented general interval arithmetic here.

edsko: Hmmm I think that needs a comment clarifying that. For me if you say "as set operations", then I'd expect, say, "+" to work as set union (minimum on `lo`, maximum on `hi`). Not saying it doesn't make sense -- can pick any semantics we need of course. Just saying it needs some clarification in the comments :)

edsko:
Would appreciate some minor changes as listed above but I think it can be merged as is also.

PR #3234, opened 2018-07-12
Add an example get pre-commit hook

erikd:
Runs stylish-haskell and hlint on the haskell files about to be
committed. Documentation in the script itself.

## Description

N/A


mhuesch:
Merging this so I can use it 🙂

PR #3238, opened 2018-07-12
[CDEC-432] Move `Mockable` and `JsonLog` to `core`

mhuesch:
## Description

Move `Mockable` and `JsonLog` into core, so that networking can be freed up. The goal is to let `networking` float upwards in the dependency graph.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CDEC-432

## Type of change

## Developer checklist

## Testing checklist
^ this PR moves code around & doesn't change functionality. Existing tests should be good enough.


core/src/Pos/Core/JsonLog/JsonLogT.hs
```diff
@@ -0,0 +1,181 @@
...
+type instance Promise (JsonLogT m) = Promise m
+type instance SharedAtomicT (JsonLogT m) = SharedAtomicT m
+type instance SharedExclusiveT (JsonLogT m) = SharedExclusiveT m
+type instance ChannelT (JsonLogT m) = ChannelT m
+
+
+
+
+
+
```
erikd: Wow, that is a lot of empty lines. Is that just the way it was in the original file?

erikd: It seems that was indeed the case. Weird!

erikd:
As you can see from the [dependency graph](https://github.com/input-output-hk/core-team-notes/blob/master/images/cardano-sl-pkg-deps.png), the `networking` package is in this werid place at the bottom of the graph, with nearly everything else depending on it.

This is why there is also networking code in `infra` and `lib`. By moving `Mockable` and the `JsonLog` stuff to `core`, `networking` can be moved up in the graph. After this commit is merged we would then work up the dependency graph replacing the `networking` versions of `Mockable` and the `JsonLog` with the `core` version until nothing more depends on the `networking` versions and we can then drop the re-exports.

We do it like this so the transition happens in a number of smaller PRs that are easier to review (this one is by far the worst).

It should also be noted that the dependency graph above is about a week out-of-date and that `infra` has already been moved up near `cardano-sl`. When `networking` is also moved up the graph we would like to move all the networking code to a single package.

core/src/Pos/Core/JsonLog/JsonLogT.hs
mhuesch: Yep 🤷‍♂️ In the spirit of minimal changes I left it. But several of these modules definitely could use tidying up.

PR #3247, opened 2018-07-13
Increase `genCoeff` range & document why

mhuesch:
Bump range from 10^12 to 10^15. 10^12 is not enough to cover the
range used in Cardano, but 10^15 is.

This is a follow up to PR #3236, which [bumped the range of `Coeff`](https://github.com/input-output-hk/cardano-sl/pull/3236/files#diff-6ec2f26e2869d3c19cb621038f455659R433), but not by enough.

### DIFF
--- a/core/test/Test/Pos/Core/Gen.hs
+++ b/core/test/Test/Pos/Core/Gen.hs
@@ -430,7 +430,11 @@ genChainDifficulty = ChainDifficulty <$> genBlockCount
 
 genCoeff :: Gen Coeff
 genCoeff = do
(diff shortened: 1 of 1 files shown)


mhuesch:
More context: the values in [1] and [2] show that our gen range wasn't sufficient (too low by 10^3), so I increased the upper bound by 10^3.

[1]: https://github.com/input-output-hk/cardano-sl/blob/develop/core/src/Pos/Core/Configuration/Core.hs#L85
[2]: https://github.com/input-output-hk/cardano-sl/blob/develop/lib/configuration.yaml#L247

PR #3255, opened 2018-07-16
[CBR-317] wallet-new: Add withMonadDBRead

k0001:
## Description

Allowing the use of the of old code depending on `MonadDBRead` in the `wallet-new` code. 

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-317

## Type of change

## Developer checklist

## Testing checklist


adinapoli-iohk:
Thank you for this PR, @k0001 !

Before I delve into reviewing it, let's turn CI green and let me write down the general recipe for situations like these:

- Every time you add a new dependency (like you did in this case for `resourcet`), the `stack2nix` job is going to fail. The best way to fix this is to hop over the [build log](https://buildkite.com/input-output-hk/cardano-sl/builds/7246#4b0b351f-cce5-4ca3-8831-9a2f0056f011) , find the 'Artifacts' section and download the ready-to-be-applied `stack2nix.patch`:

screen shot 2018-07-19 at 08 45 04

- Once you have the patch on your desktop, simply apply it with `git apply stack2nix.patch`.

- We use `stylish-haskell` and formatting is checked as part of CI, therefore we need to ensure that your patch abides to the coding style. Exactly like `stack2nix`, we can ask (Hydra in this case) for an [patch ready to apply](https://hydra.iohk.io/build/174394):

screen shot 2018-07-19 at 08 47 48

The other failures might be red herrings, as lately the wallet integration tests had some hiccups and the AppVeyor builds currently times out due to the 90 minutes build time limit.

wallet-new/cardano-sl-wallet-new.cabal
```diff
@@ -125,7 +126,7 @@ library
                        Cardano.Wallet.WalletLayer.Legacy
                        Cardano.Wallet.WalletLayer.Error
                        Paths_cardano_sl_wallet_new
-  ghc-options:         -Wall
+  ghc-options:         -Wall -Werror -Wcompat -fwarn-redundant-constraints
```
edsko: Why did you make this change?

wallet-new/default.nix
```diff
@@ -3,7 +3,8 @@
 #
 # Running `nix-build` builds the `wallet-new` code.
 let
-  drv = (import ../. {}).cardano-sl-wallet-new;
+  top = import ../. { forceDontCheck = true; enableBenchmarks = false; };
+  drv = top.cardano-sl-wallet-new;
```
edsko: How is this related to this PR? Should this be changed for everyone? If so, maybe open a separate PR with motivation? It seems unrelated to this problem.

wallet-new/src/Cardano/Wallet/Kernel/Compat.hs
```diff
@@ -0,0 +1,103 @@
...
+  -> Maybe Integer
+  -- ^ Optional seed which overrides one from testnet initializer if
+  -- provided.
+  -> IO ( ProtocolMagic
+        , Maybe GeneratedSecrets
+        , GenesisData
+        , GenesisHash
+        , BlockVersionData
+        , ProtocolConstants )
+getCoreConfigurations cc fp yts yseed =
```
edsko: I don't think this is quite right. We don't want to read the `configuration.yaml`. Instead, we want to query the node context/state.

wallet-new/src/Cardano/Wallet/Kernel/Compat.hs
```diff
@@ -0,0 +1,103 @@
...
+-- | This monad transformer exists solely to provide a 'MonadRealDB' instance,
+-- as required by upstream libraries.
+newtype DBReadT m a = DBReadT { unDBReadT :: ReaderT NodeDBs m a }
+  deriving (Functor, Applicative, Monad, MonadThrow, MonadTrans, MonadIO)
+
+instance MonadUnliftIO m => MonadUnliftIO (DBReadT m) where
+  askUnliftIO =
+    DBReadT (withUnliftIO (\u -> pure (UnliftIO (unliftIO u . unDBReadT))))
+
+instance (HasConfiguration, MonadThrow (DBReadT m), MonadRealDB NodeDBs (ReaderT NodeDBs m))
```
edsko: This seems strange to me. You are requiring instances for the monad you are defining here...? Why not give a `MonadThrow` instance for `DBReadT`? (That would then probably reduce this constraint to `MonadThrow m`, which is more reasonable.)

edsko: I'm very confused by the `MonadRealDB NodeDBs (ReaderT NodeDBs m))` constraint. You are requiring something for the _underlying_ (unwrapped) monad of `DBReadT`? I don't quite get it. Also, doesn't this mean that in order to use the `DBReadT` thing, I then am required, as a user, to somehow satisfy this `MonadRealDB` requirement -- for a monad I shouldn't even see, because it's an implementation detail of `DBReadT`?

wallet-new/default.nix
k0001: It's not directly related. I don't think anybody other than me is using this. I can take it out if necessary and just keep the changes locally.

wallet-new/src/Cardano/Wallet/Kernel/Compat.hs
k0001: There is a `MonadThrow` instance already. This was just a mistake, I replaced this with `MonadThrow m`.

k0001: Yes, this is horrible. However, `DBReadT` is supposed to be a type internal to this module. I just removed it from the export list.

wallet-new/src/Cardano/Wallet/Kernel/Compat.hs
k0001: Hmm... but how do these values reach the “node context/state” in the first place? They need to be read from somewhere, don't they?

wallet-new/cardano-sl-wallet-new.cabal
k0001: Because these are the settings that CI seem to be using when building. It makes sense to try and solve anything that will trigger a CI error before submitting the changes. If building with `stack`, then these GHC flags are already being used.

k0001:
> I strongly suggest that you actually use your adapter

This is my intention. I am in the process of doing that in CBR-243.

> ... but in order to plug in your DBReadT ...

`DBReadT` is supposed to be internal. I just removed it from the export list.

> it would be great if you could squash and clean up the commit history before marking PRs ready for review.

I would squash the commits before merging, of course, but doing that while a PR is open seems rather counterproductive for everyone.

wallet-new/src/Cardano/Wallet/Kernel/Compat.hs
adinapoli-iohk: @k0001 I _think_ you can use something like `NodeContext` or `NodeResources`. Going by my memory here, you might need to do a bit of grepping inside the codebase, heh.

wallet-new/default.nix
edsko: Fair enough. I'd prefer we keep the PR on topic. Especially since I know nothing about `nix` and don't feel very comfortable approving changes to the nix infrastructure :flushed: :grin:

edsko:
> DBReadT is supposed to be internal. I just removed it from the export list.

Oh! I see. Then I misunderstood how this meant to be used. Fair enough. That constraint on the `ReaderT` thing still wrecks my head though, can't quite figure out what's going on there :) Can we really not simplify that constraint? Weere is it coming from? And where is it satisfied? 

> I would squash the commits before merging, of course, but doing that while a PR is open seems rather counterproductive for everyone.

I'd prefer to review a PR that's actually reading for merging, rather than a PR that will be "cleaned up once reviewed". Maybe that's just me. Also, in general (not so much for this particular PR), it's often easier to review PRs commit by commit, provided that they are cleaned up and are self-consistent. (Bit of a pipe-dream, I realize, but something to aim for :) .

wallet-new/src/Cardano/Wallet/Kernel/Compat.hs
k0001: @adinapoli-iohk only `NodeDBs` seems to be there (which this `getCoreConfiurations` function doesn't return, so it will likelly need to be obtained from there).

edsko:
> I am doing that in CBR-243.

But that means that other people who need access to the rocks DB cannot use this adapter until you finish work on wallet restoration, which is a bit of a shame.

k0001:
> Oh! I see. Then I misunderstood how this meant to be used. Fair enough. That constraint on the ReaderT thing still wrecks my head though, can't quite figure out what's going on there :) Can we really not simplify that constraint? Weere is it coming from? And where is it satisfied?

I suppose you are talking about the `MonadRealDB NodeDBs (ReaderT NodeDBs m)` constraint on the `MonadDBRead (DBReadT m)` instance. This `MonadDBRead` instance defers most of it's work  (e.g., the `dbGetDefault` function relies on `MonadRealDB`) to the `MonadRealDB NodeDBs (ReaderT NodeDBs m)` instance so as to not do any of the terrible choices necessary in order satisfy `MonadRealDB NodeDBs (DBReadT m)` itself.

Notice that `MonadRealDB` is this horrible monster:

```
type MonadRealDB ctx m =                                                               
   ( MonadReader ctx m
   , HasLens NodeDBs ctx NodeDBs
   , MonadIO m
   , MonadCatch m
   )
```

Meaning that if we wanted `DBReadT NodeDBs m` itself to satisfy this, then we would need to at least make the controversial choice of giving a `MonadReader NodeDBs (DBReadT m)` instance with the surprising behavior of not behaving as pass-though to an underlying `ReaderT NodeDBs m`. Presumably, none of this matters that much, because all is happening internally in this module and we are not exporting things. But by this same reasoning, we might as well just defer most of the work to the `MonadRealDB NodeDBs (ReaderT NodeDBs m)` instance, seeing how `DBReadT m` is essentially a dummy newtype around `ReaderT NodeDBs m`. 

An alternative would be to just give a `MonadDBRead (ReaderT NodeDBs m)` instance. However, I think that's as terrible an idea as the already existing `MonadRealDB NodeDBs (ReaderT NodeDBs m)` is: Making `ReaderT` part of a public API is almost always a bad choice. Moreover, this awkward instance will be always exported from this module. 


> I'd prefer to review a PR that's actually reading for merging, rather than a PR that will be "cleaned up once reviewed". Maybe that's just me. Also, in general (not so much for this particular PR), it's often easier to review PRs commit by commit, provided that they are cleaned up and are self-consistent. (Bit of a pipe-dream, I realize, but something to aim for :) .

That's fair, and I agree with that. However, keep in mind that this PR has been open for a while already, and it continued to add commits even after it was submitted for review the first time. Perhaps the issue here is that the PR was submitted for review prematurely.

k0001:
>>  I am doing that in CBR-243.
>
> But that means that other people who need access to the rocks DB cannot use this adapter until you finish work on wallet restoration, which is a bit of a shame.

On the contrary, the whole reason for having CBR-317 be a piece of work separate from CBR-243 was so that folks can use this without waiting for CBR-243 to be completed.

In CBR-243 there are working use cases of this `withMonadDBRead` function (e.g., calling `(withMonadDBRead .... (foo :: forall n. MonadDBRead n => n ()) :: forall m. (MonadCatch m, MonadUnlift m) => m ()` compiles just fine), but that's besides the point of this function existing alongside `getCoreConfigurations` as a means to obtaining many of the inputs required by this function.

PR #3256, opened 2018-07-16
[CBR-313] Implement "newTransaction" and "estimateFees" endpoints (new data layer)

adinapoli-iohk:
## Description

This PR delivers two new Servant handlers implemented using the new wallet data layer, specifically `newTransaction` and `estimateFees`. Do do so, I've also added the relevant `WalletLayer` and `Kernel` functions, as well as tests.

@kderme You will notice I have reshuffled a bit the `getTransactionsHistory` stub you got there from the top-level `Handlers` file into a specific `API.Handlers.Transactions` module. I hope it doesn't create too much mayhem 😉 

@edsko A note on review: this PR still lacks two aspects:

- [ ] Extra fallback strategy in case we exceed the number of inputs;
- [ ] When crafting a new `V1.Transaction`, there are two fields which needs to be populated using the TxMeta storage. I have opened CBR-324 for that.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-313

## Type of change

## Developer checklist

- [~] I have read the [style guide](https://github.com/input-output-hk/cardano-sl/blob/develop/docs/style-guide.md) document, and my code follows the code style of this project.
- [~] I have updated any documentation accordingly, if needed. Documentation changes can be reflected in opening a PR on [cardanodocs.com](https://github.com/input-output-hk/cardanodocs.com), amending the inline [Haddock](https://www.haskell.org/haddock/) comments, any relevant README file or one of the document listed in the [docs](https://github.com/input-output-hk/cardano-sl/tree/develop/docs) directory.

## Testing checklist


wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,279 @@
...
+               -> CoinSelectionOptions
+               -- ^ The options describing how to tune the coin selection.
+               -> HdAccountId
+               -- ^ The source @(root, account) from where the payment was
+               -- originated
+               -> NonEmpty (Address, Coin)
+               -- ^ The payees
+               -> IO (Either NewTransactionError TxAux)
+newTransaction ActiveWallet{..} genChangeAddr signAddress options accountId payees = do
+    snapshot <- getWalletSnapshot walletPassive
```
edsko: Something we have to be wary of here -- the address generation function lives in `IO`, not just because it generates random addresses but presumably also because it registers those addresses with the DB. This means that we're not _actually_ just using a single snapshot, we're in fact modiying the DB as we go. 

In fact, it wouldn't really be necessary to do this. We could change this so it _doesn't_ actually modify the DB, creating new addresses but not registering them. In that way we are _actually_ using just a single snapshot (could modify the type to be `Snapshot -> IO Address` or whatever) and then register them when it has succeeded. That would address your cocern below and also make the flow of data a bit clearer.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,279 @@
...
+-- from a fixed seed obtained from hashing the payees. This guarantees that
+-- when we estimate the fees and later create a transaction, the coin selection
+-- will always yield the same value, making the process externally-predicatable.
+
+newtype PayMonad a = PayMonad {
+      buildPayment :: ReaderT (NonEmpty (Address, Coin)) IO a
+    } deriving (Functor, Applicative, Monad, MonadIO, MonadReader (NonEmpty (Address, Coin)))
+
+-- | \"Invalid\" 'MonadRandom' instance for 'PayMonad' which generates
+-- randomness using the hash of 'NonEmpty (Address, Coin)' as fixed seed.
```
edsko: Er, not sure I understand this. Is this ti make coin selection deterministic? Are we running coin selection in this monad? What do we use this for?

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,279 @@
...
+estimateFees activeWallet@ActiveWallet{..} genChangeAddr signAddress options accountId payees = do
+    snapshot         <- getWalletSnapshot walletPassive
+    let originalUtxo = accountUtxo snapshot accountId
+    res <- newTransaction activeWallet genChangeAddr signAddress options accountId payees
+    case res of
+         Left e  -> return . Left . EstFeesTxCreationFailed $ e
+         Right tx -> -- calculate the fee as the difference between inputs and outputs.
+             -- NOTE(adn) Apparently we shouldn't worry about the 'ExpenseRegulation'
+             -- affecting the way we sum, as no matter who pays for the fee,
+             -- the final difference won't be affected.
```
edsko: Hmmm I don't think that's _entirely_ true. For sender pays fees we may need to add additional inputs (where recipient pays fees would just fail). But it's close enough I suppose.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,279 @@
...
+             return $ Right
+                    $ sumOfInputs tx originalUtxo `unsafeSubCoin` sumOfOutputs tx
+  where
+      sumOfInputs :: TxAux -> Utxo -> Coin
```
edsko: Is there really no function for this in core? :-o

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,270 @@
...
+
+
+-- NOTE(adn) At the moment we are passing
+-- the full set of Hd wallets as input, which means our lookup function
+-- for an 'HdAddress' can span across the whole wallets and can be very
+-- costly. However, in the way the 'DB' is modelled at the moment, we
+-- store the addresses in a \"flat\" representation (i.e. in an 'IxSet'),
+-- so indexing directly by 'Core.Address' might not be any less efficient
+-- than performing two lookups, one on 'HdAccountId' followed by one on
+-- 'Core.Address'.
```
edsko: I wouldn't worry about efficiency here. However, passing in a substructure of the DB does eliminate potential bugs as  it makes it clearer what we should do.

wallet-new/src/Cardano/Wallet/Kernel/Addresses.hs
```diff
@@ -65,6 +67,11 @@ instance Buildable CreateAddressError where
     build (CreateAddressHdRndAddressSpaceSaturated hdAcc) =
         bprint ("CreateAddressHdRndAddressSpaceSaturated " % F.build) hdAcc
 
+instance Show CreateAddressError where
+    show = formatToString build
+
+instance Exception CreateAddressError
```
adinapoli-iohk: Side note: we should open a ticket to be sure we track down all the possible exceptions that the `Kernel` (and/or the `WalletLayer`) can throw, so that we can catch this into the `wai` exception handler at the top level and @KtorZ 's squad will eventually be able to curate them into proper V1 error types.

wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -342,3 +349,59 @@ largestFirst opts changeAddr mkTx maxInps =
...
+           34 = size of encoded Blake2b_256 Tx hash, 5 = max size of encoded
+           CRC32 (range is 1..5 bytes, average size is just under 5 bytes).
+
+      11 = 2 + 2 + 2 + 5, where the 2s are: tag for TxOut ctor, tag for Address
+           ctor, and delimiters for encoded address. 5 = max size of CRC32.
+
+      32 = 1 + 30 + 1, where the first 1 is a tag for a tuple length, the
+           second 1 is the encoded address type. 30 = size of Blake2b_224
+           hash of Address'.
+-}
```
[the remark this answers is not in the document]
adinapoli-iohk: Indeed we do, I can bang a comment tracking that PR and add another cleanup ticket for that.

wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -342,3 +349,59 @@ largestFirst opts changeAddr mkTx maxInps =
...
+             -> Byte     -- ^ Estimated size of the resulting transaction.
+estimateSize saa sta ins outs
+    = fromBytes . fromIntegral $
+      5
+    + 42 * ins
+    + (11 + listSize (32 + (fromIntegral saa))) * length outs
+    + sum (map intSize outs)
+    + fromIntegral sta
+  where
+    intSize s =
```
edsko: See coment https://github.com/input-output-hk/cardano-sl/pull/3232#discussion_r202620067.

wallet-new/src/Cardano/Wallet/Kernel/DB/Util/IxSet.hs
```diff
@@ -146,6 +148,15 @@ member pk = isJust . view (Lens.at pk)
 size :: IxSet a -> Int
 size = IxSet.size . unwrapIxSet
 
+-- | Safely returns the 'head' of this 'IxSet', but only if it is a singleton
+-- one, i.e. only if it has @exactly@ one element in it. Usually this is
+-- used in tandem with 'getEQ' to witness the existence of exactly one element
+-- in the set indexed by a particular index.
+onlyOne :: IxSet a -> Maybe a
+onlyOne ixset = case IxSet.toList . unwrapIxSet $ ixset of
```
adinapoli-iohk: We can actually rename this `getOne` as the twin function on the underlying `IxSet`, for consistencies' sake.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,279 @@
...
+    -> NonEmpty (Address, Coin)
+    -- ^ The payees
+    -> IO (Either PaymentError Tx)
+pay activeWallet genChangeAddr signAddress opts accountId payees = do
+        retrying retryPolicy shouldRetry $ \rs -> do
+            tx <- newTransaction activeWallet genChangeAddr signAddress opts accountId payees
+            case tx of
+                 Left e      -> return (Left $ PaymentNewTransactionError e)
+                 Right txAux -> do
+                     succeeded <- newPending activeWallet accountId txAux
```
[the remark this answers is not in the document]
adinapoli-iohk: Perhaps we should do this as part of CBR-324, or CBR-239 , perhaps whichever @kderme prefers here.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,270 @@
...
+                     case succeeded of
+                          Left e   -> do
+                              print (sformat build e)
```
adinapoli-iohk: We shouldn't print the error here, that's a relic of debugging -- will fix!

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,270 @@
...
+                              retriesLeft <- applyPolicy retryPolicy rs
+                              return . Left $ case retriesLeft of
+                                   Nothing ->
+                                       PaymentSubmissionMaxAttemptsReached
+                                   Just _  ->
+                                       PaymentNewPendingError e
+                          Right () -> return . Right . taTx $ txAux
+    where
+        retryPolicy :: RetryPolicyM IO
+        retryPolicy = constantDelay 5000000 <> limitRetries 6
```
adinapoli-iohk: Fair point from @edsko : we might want to randomise the constantDelay to avoid threads simultaneously competing on the shared resources here -- perhaps we can leverage something like `fullJitterBackoff` from the `retry` package.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
adinapoli-iohk: Me and @edsko had a chat on the above and we concluded that a better design is to factor out the concern of generating an `Address` from the concern of actually running the coin selection algorithm.

I will refactor the `MkTx` function to actually deal with address creation and the `runCoinSelT` will be completely agnostic to the entire change-address generation business.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -0,0 +1,270 @@
...
+    } deriving (Functor, Applicative, Monad, MonadIO, MonadReader (NonEmpty (Address, Coin)))
+
+-- | \"Invalid\" 'MonadRandom' instance for 'PayMonad' which generates
+-- randomness using the hash of 'NonEmpty (Address, Coin)' as fixed seed.
+instance MonadRandom PayMonad where
+    getRandomBytes _ =
+        ask >>= return . ByteArray.convert
+                       . encodeUtf8 @Text @ByteString
+                       . sformat build
+                       . hash
```
adinapoli-iohk: This is indeed bollocks, as the initial seed is correctly computed, but successive calls to `getRandomBytes` yields exactly the same result, so that's no randomness at all here. We should probably use the hash plus something like an incremental counter tucked into an `IORef` to ensure we get deterministic yet random results.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
adinapoli-iohk: Agreed, this needs a better comment to explain things properly.

wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -222,22 +228,14 @@ runCoinSelT opts genChangeAddr pickUtxo mkTx policy request utxo = do
...
                         i:is -> i :| is
-            allOuts = case outs ++ changeOuts of
-                        []   -> error "runCoinSelT: empty list of outputs"
-                        o:os -> o :| os
+            originalOuts = case outs of
+                               []   -> error "runCoinSelT: empty list of outputs"
+                               o:os -> o :| os
         -- TODO: We should shuffle allOuts
-        mkTx allInps allOuts
+        mkTx allInps originalOuts (concatMap coinSelChange css)
```
edsko: I like it :)

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
edsko: But why does it still take a `genChangeAddr` (and a signer function fort hat matter?) Shouldn't that just be internal responsibility of the wallet?

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -163,17 +176,28 @@ newTransaction ActiveWallet{..} genChangeAddr signAddress options accountId paye
...
+-- This ensures that the coin selection algorithm runs in a random environment
+-- which is yet deterministically reproduceable by feeding the same set of
+-- payees.
 instance MonadRandom PayMonad where
-    getRandomBytes _ =
-        ask >>= return . ByteArray.convert
-                       . encodeUtf8 @Text @ByteString
-                       . sformat build
-                       . hash
+    getRandomBytes _ = do
```
edsko: Although it is better, I still feel somewhat uncomfortable with this code. We have IO available. Couldn't we call into a proper random number generator, but using your `PayMonad` state to give it a seed? That way we'd have predictability but we wouldn't be trying to roll our own PRNG.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -208,9 +232,10 @@ estimateFees activeWallet@ActiveWallet{..} genChangeAddr signAddress options acc
     case res of
          Left e  -> return . Left . EstFeesTxCreationFailed $ e
          Right tx -> -- calculate the fee as the difference between inputs and outputs.
-             -- NOTE(adn) Apparently we shouldn't worry about the 'ExpenseRegulation'
-             -- affecting the way we sum, as no matter who pays for the fee,
-             -- the final difference won't be affected.
+             -- NOTE(adn) In case of 'SenderPaysFee' is practice there might be a slightly
+             -- increase of the projected fee in the case we are forced to pick "yet another input"
+             -- to be able to pay the fee, which would, in turn, also increase the fee due to
+             -- the extra input being picked.
```
edsko: It is also important to note that _if_ this happens, it may in fact be the case that we give you an estimated fee where in reality we would _fail_ to produce a transaction at all.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -213,22 +253,22 @@ instance Arbitrary EstimateFeesError where
...
              -> HdAccountId
-             -- ^ The source @(root, account) from where the payment was
-             -- originated
+             -- ^ The source HD Account from where the payment should originate
              -> NonEmpty (Address, Coin)
              -- ^ The payees
              -> IO (Either EstimateFeesError Coin)
-estimateFees activeWallet@ActiveWallet{..} genChangeAddr signAddress options accountId payees = do
+estimateFees activeWallet@ActiveWallet{..} spendingPassword options accountId payees = do
     snapshot         <- getWalletSnapshot walletPassive
```
adinapoli-iohk: TODO: have `newTransaction` returns the utxo it originally used so that the calculation of the fee is more precise.

PR #3258, opened 2018-07-16
Set date of publishing new version

edsko:
Literally just changing the date.


adinapoli-iohk:
Merging overriding CI as this is touching only the spec.

PR #3264, opened 2018-07-17
[CDEC-439] Cleaning ntp client

coot:
## Description

I cleaned the ntp client a bit. 

The ntp client resolves addresses and takes first ipv4 and ipv6 addresses. It now will also try the other address in case of failure (it may happen that ipv4/6 is not supported by the os).

And I added tests:
* serialization / deserialization of ntp packets
* computation of clock offset
*  the coercions between ntp time and unix time

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CDEC-439

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/networking/src/Ntp/Client.hs
+++ b/networking/src/Ntp/Client.hs
@@ -1,9 +1,12 @@
 {-# LANGUAGE ConstraintKinds     #-}
 {-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE GADTs               #-}
+{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE NoImplicitPrelude   #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
 
 -- | This module implements functionality of NTP client.
 
@@ -11,35 +14,35 @@ module Ntp.Client
     ( NtpClientSettings (..)
     , NtpStatus (..)
     , withNtpClient
-    , withoutNtpClient
-    , ntpSingleShot
     ) where
 
-import           Universum
+import           Universum hiding (Last)
 
 import           Control.Concurrent (threadDelay)
-import           Control.Concurrent.Async (async, concurrently, forConcurrently,
-                     race, withAsync)
-import           Control.Concurrent.STM (TVar, check, modifyTVar')
+import           Control.Concurrent.Async (async, cancel, concurrently_, race,
+                     withAsync)
+import           Control.Concurrent.STM (TVar, modifyTVar', retry)
 import           Control.Exception.Safe (Exception, catchAny, handleAny)
 import           Control.Monad (forever)
-import           Data.Binary (decodeOrFail, encode)
+import           Data.Binary (decodeOrFail)
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.List.NonEmpty as NE
-import           Data.Maybe (catMaybes, isNothing)
+import           Data.Semigroup (Last (..))
 import           Data.Time.Units (Microsecond, TimeUnit, toMicroseconds)
 import           Data.Typeable (Typeable)
 import           Formatting (sformat, shown, (%))
-import           Network.Socket (AddrInfo, SockAddr (..), Socket, addrAddress,
-                     addrFamily, close)
-import           Network.Socket.ByteString (recvFrom, sendTo)
-import           System.Wlog (LoggerNameBox)
+import qualified Network.Socket as Socket
+import           Network.Socket.ByteString (recvFrom)
 import qualified System.Wlog as Wlog
 
-import           Ntp.Packet (NtpOffset, NtpPacket (..), evalClockOffset,
-                     mkCliNtpPacket, ntpPacketSize)
-import           Ntp.Util (createAndBindSock, resolveNtpHost, selectIPv4,
-                     selectIPv6, udpLocalAddresses, withSocketsDoLifted)
+import           Ntp.Packet (NtpOffset, NtpPacket (..), clockOffset,
+                     mkNtpPacket, ntpPacketSize)
+import           Ntp.Util (AddrFamily (..), Addresses, EitherOrBoth (..),
+                     Sockets, WithAddrFamily (..), createAndBindSock,
+                     foldEitherOrBoth, logDebug, logInfo, logWarning, ntpTrace,
+                     resolveNtpHost, runWithAddrFamily, sendPacket,
+                     udpLocalAddresses)
+import           Pos.Util.Trace (traceWith)
 
 data NtpStatus =
       -- | The difference between ntp time and local system time
@@ -66,7 +69,7 @@ data NtpClientSettings = NtpClientSettings
 data NtpClient = NtpClient
     { ncSockets  :: TVar Sockets
       -- ^ Ntp client sockets: ipv4 / ipv6 / both.
-    , ncState    :: TVar (Maybe [NtpOffset])
+    , ncState    :: TVar [NtpOffset]
       -- ^ List of ntp offsets and origin times (i.e. time when a request was
       -- send) received from ntp servers since last polling interval.
     , ncStatus   :: TVar NtpStatus
@@ -77,10 +80,10 @@ data NtpClient = NtpClient
       -- ^ Ntp client configuration.
     }
 
-mkNtpClient :: MonadIO m => NtpClientSettings -> TVar NtpStatus -> Sockets -> m NtpClient
+mkNtpClient :: NtpClientSettings -> TVar NtpStatus -> Sockets -> IO NtpClient
 mkNtpClient ncSettings ncStatus sock = liftIO $ do
     ncSockets <- newTVarIO sock
-    ncState  <- newTVarIO Nothing
+    ncState   <- newTVarIO []
     return NtpClient{..}
 
 data NoHostResolved = NoHostResolved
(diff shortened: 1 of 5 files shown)


networking/src/Ntp/Client.hs
```diff
@@ -77,280 +80,209 @@ data NtpClient = NtpClient
...
--- passes.  If at least one server responded `handleCollectedResponses` will
--- update `ncStatus` in `NtpClient`.
-startSend :: NtpClient -> [SockAddr] -> IO ()
-startSend cli addrs = do
     let respTimeout = ntpResponseTimeout (ncSettings cli)
-    let poll = ntpPollDelay (ncSettings cli)
-
+    let poll        = ntpPollDelay (ncSettings cli)
+
+    _ <- withAsync
```
erikd: That looks like you are discarding the result of a computation. Since the last statement in the computation is `cancel a` whose result type is `IO ()`, the above could be written as:
```
    () <- withAsync
```
which is more obviously correct and hence easier for the reviewer.

networking/src/Ntp/Client.hs
```diff
@@ -77,280 +80,209 @@ data NtpClient = NtpClient
...
-
-socketsToList :: Sockets -> [Socket]
-socketsToList (BothSock s1 s2) = [s1, s2]
-socketsToList (IPv4Sock s1)    = [s1]
-socketsToList (IPv6Sock s1)    = [s1]
+    liftIO $ logInfo "withNtpClient"
+    ncStatus <- newTVarIO NtpSyncPending
+    -- using async so the ntp thread will be left running even if the parent
+    -- thread finished.
+    _ <- liftIO $ async (spawnNtpClient ntpSettings ncStatus)
```
erikd: How many times is this function likely to be called? Does the `spawnNtpClient` ever terminate?

If it never terminates,  how do we make sure we don't get hundreds, thousands or even more?

If if just runs, does it job and then terminates, it would be nice to have a comment to that effect.

networking/src/Ntp/Client.hs
```diff
@@ -77,280 +80,209 @@ data NtpClient = NtpClient
...
+            mkSockets settings
+  where
+    doMkSockets :: IO (Option Sockets)
+    doMkSockets =
+        do
+            addrs <- udpLocalAddresses
+            (<>) <$> (Option <$> createAndBindSock IPv4 addrs)
+                 <*> (Option <$> createAndBindSock IPv6 addrs)
+
+    handlerE :: SomeException -> IO (Option Sockets)
```
erikd: We're using `Control.Exception.Safe` in this module so `SomeException` does not include async exxecptions like stack and heap overflow, right?

networking/src/Ntp/Client.hs
coot: That's true.  PureScript has a class for types which can be ignored without a warning (with an instance for `()`), it would be nice to have it in Haskell.

networking/src/Ntp/Client.hs
coot: It just runs once, a comment should be in place indeed.

networking/src/Ntp/Client.hs
coot: Yes, it's a bit inconsistent though: `Ntp.Client` is using safe-exceptions while `Ntp.Util` is using `Contorl.Exception` (but it's never catching async exceptions, only `IOException` and one custom exception).  I think we should just catch `IOExceptions` in `Ntp.Client` as well rather than `SomeException`.

coot:
@erikd I updated with your comments with two additions:

* now whole `Ntp` is using `Control.Exception` catching `IOException`, instead of mixing `safe-exceptions` with `Control.Exception`.
* I moved `NtpConfiguration` from `infra` to `networking`

networking/src/Ntp/Util.hs
```diff
@@ -1,97 +1,319 @@
...
+getAddrFamily (WithIPv6 _) = IPv6
+getAddrFamily (WithIPv4 _) = IPv4
+
+-- |
+-- Keep either of the two types or both.
+data EitherOrBoth a b
+    = EBFirst  !a
+    | EBSecond !b
+    | EBBoth   !a !b
+    deriving (Show, Eq, Ord)
```
parsonsmatt: aka the [`these`](http://hackage.haskell.org/package/these-0.7.4/docs/Data-These.html) datatype, though this is strict

networking/test/Test/NtpSpec.hs
```diff
@@ -0,0 +1,101 @@
+{-# LANGUAGE TypeApplications #-}
+
+module Test.NtpSpec
+    ( spec
```
parsonsmatt: :raised_hands:

networking/test/Test/NtpSpec.hs
```diff
@@ -0,0 +1,101 @@
...
+        return $ NtpMicrosecond $ fromMicroseconds $ x
+
+spec :: Spec
+spec = describe "NtpClient" $ do
+    describe "clockOffsetPure" $ do
+        it "should return clock offset" $ property $ \(NtpPacketWithOffset {..}) ->
+            let offset = clockOffsetPure npoNtpPacket npoDestinationTime
+            in npoOffset === offset
+    describe "realMcsToNtp" $ do
+        it "should be right inverse of ntpToRealMcs" $ property $ \(NtpMicrosecond x) ->
```
parsonsmatt: `hspec` has a shorthand for this; `it msg (property f)` can be `prop msg f`

networking/test/Test/NtpSpec.hs
```diff
@@ -0,0 +1,101 @@
...
+            x === uncurry ntpToRealMcs (realMcsToNtp x)
+        it "should be left inverse of ntpToRealMcs" $ property $ \(NtpTime x@(sec, frac)) ->
+            let (sec', frac') = realMcsToNtp (uncurry ntpToRealMcs x)
+            in sec === sec' .&&. frac `div` 4310 === frac' `div` 4310
```
parsonsmatt: 4310 is a bit of a magic number here. What's the significance? Could it be named?

networking/src/Ntp/Util.hs
coot: Indeed, the strictness doesn't matter that much, since we are just holding max three values, so if you prefer we can pull `these` as a dependency. `these` hasn't been yet used anywhere in `cardano-sl` but it's a small dependency so I don't mind.

networking/test/Test/NtpSpec.hs
coot: It is `1000000 `div` 232`, in the ntp spec time is represented in seconds + a fraction of a second which resolved to `232` picoseconds, so there are exactly `4310` of them in a milisecond.

networking/src/Ntp/Packet.hs
```diff
@@ -4,92 +4,142 @@
...
+    , ntpReceivedTime :: Microsecond  -- ^ when server got request
+    , ntpTransmitTime :: Microsecond  -- ^ when client sent request
+    } deriving (Show, Eq)
 
 instance Binary NtpPacket where
     put NtpPacket{..} = do
         putWord8 ntpParams
+        putWord8 0
+        putInt8 ntpPoll
+        putWord8 0
```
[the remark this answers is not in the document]
coot: Not really, I added bits just so we have a test that `put` and `get` are inverse of each other.

networking/src/Ntp/Client.hs
```diff
@@ -77,280 +86,239 @@ data NtpClient = NtpClient
...
+    -- Restart the @loop@ in case of errors; wait 5s before recreating the
+    -- socket.
+    handleIOException
+        :: AddrFamily
+        -> IOException
+        -> IO ()
+    handleIOException addressFamily e = do
+        logDebug $ sformat ("startReceive failed with reason: "%shown) e
+        threadDelay 5000000
+        udpLocalAddresses >>= createAndBindSock addressFamily >>= \case
```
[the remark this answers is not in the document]
coot: Yeah, good idea.

networking/src/Ntp/Client.hs
```diff
@@ -66,7 +75,7 @@ data NtpClientSettings = NtpClientSettings
 data NtpClient = NtpClient
     { ncSockets  :: TVar Sockets
       -- ^ Ntp client sockets: ipv4 / ipv6 / both.
-    , ncState    :: TVar (Maybe [NtpOffset])
+    , ncState    :: TVar [NtpOffset]
```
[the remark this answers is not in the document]
coot: There was, but it wasn't necessary.  It was used to not accept request which came late.  Every packet comes with information when it was sent so we can discard it based on that information.  This is now done in [clockOffset function](https://github.com/input-output-hk/cardano-sl/blob/fac69409a55fdbb6576acb17a5ccd2a4a8c1dacc/networking/src/Ntp/Packet.hs#L132)

networking/src/Ntp/Client.hs
```diff
@@ -77,280 +86,239 @@ data NtpClient = NtpClient
...
-        Nothing        -> False
-        Just responses -> length responses >= length servers
+    let respTimeout = ntpResponseTimeout (ncSettings cli)
+    let poll        = ntpPollDelay (ncSettings cli)
+
+    () <- withAsync
+        (do
+            -- wait for responses and update status
+            _ <- timeout respTimeout waitForResponses
+            updateStatus cli
```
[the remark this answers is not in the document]
coot: Good point, I was actually thinking about doing this too.  I haven't done it because we need `ncState :: NtpClient -> TVar [NtpOffset]` variable to communicate between the listener thread and sending thread.  But we can change how the client is run: if we start the listener thread in this place we will not need this tvar.  The communication will be done just using `ncStatus`.

networking/src/Ntp/Client.hs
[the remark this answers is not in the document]
coot: Giving this a second thought: in this place we are just recreating the socket that failed, while `mkSockets` creates both of them, i.e. `IPv4` and `IPv6`; so I don't think we should reuse it here.

networking/src/Ntp/Client.hs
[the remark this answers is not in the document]
coot: We need to listen on two sockets, so I think there's no way around impure `ncState`. Maybe you have some other idea?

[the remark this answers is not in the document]
avieth: I wonder if having a receiving thread (`startReceive`) is the wrong way around. Do we ever expect to get a message without first sending one? That's to say, are we acting as an ntp *server* as well as a client? If not, maybe we should `recvFrom` the socket after sending on it. That would give a simpler implementation, with less concurrency and fewer mutable cells.

networking/src/Ntp/Packet.hs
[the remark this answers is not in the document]
avieth: They weren't inverse prior to this change?
So I'm guessing the `replicateM_ 4 $ putWord8 0` was there beecause `ntpPoll` was assumed to be `0`. Here there's a `putInt8` rather than `putWord8`. Same encoding?
Then there's the `replicateM_ 9 $ PutWord32be 0`, which is replaced by `replicateM_ 5 $ putWord32be 0` but where are the remaining 4? `ntpOriginTime ^.. each` and `ntpReceivedTime ^.. each` give 2 `Word32`s each? Lens combinators are annoying hard to search for.

networking/src/Ntp/Packet.hs
```diff
@@ -4,92 +4,142 @@
...
+-- |
+-- For pairs @(x, y) :: (Word32, Word32)@ with @y \`mod\` 4294 == 0@ it is
+-- be right inverse of @'realMsgToNtp'@.   In general it is not injective (for that
+-- we'd need to use @'Picosecond'@ instead of @'Microsecond'@).
+ntpToRealMcs :: Word32 -> Word32 -> Microsecond
+ntpToRealMcs sec frac =
+    let -- microseconds
+        secMicro :: Integer
+        secMicro = (fromIntegral sec - ntpTimestampDelta) * 1000000
+        -- Each fraction resolves to 232 pico seconds (`232 ≈ 10000000/4294`)
```
avieth: There's an extra `0` in the numerator.

The explanation could be better. We divide `1` second into `2 ^ 32` parts, giving `2.3283064365386963e-10` as the quantum. A picosecond is `10e-12` of a second, so this is `232` picoseconds.

networking/src/Ntp/Client.hs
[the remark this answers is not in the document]
coot: It is, but I haven't changed that.  I just didn't want to rewrite everything. I will add a todo, and let's move on.

networking/src/Ntp/Packet.hs
coot: Sorry, I put this comment under wrong thread.

PR #3274, opened 2018-07-19
[CBR-340] 'createWallet' implemented with the new data layer

adinapoli-iohk:
## Description

This PR allows the creation of a new wallet via the V1 API. It does so by exposing the relevant functions both in the WalletLayer and in the Kernel.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CBR-340

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/scripts/launch/demo-with-new-wallet-api.sh
+++ b/scripts/launch/demo-with-new-wallet-api.sh
@@ -2,4 +2,4 @@
 
 base=$(dirname "$0")
 
-WALLET_EXTRA_ARGS="--new-wallet" WALLET_EXE_NAME='cardano-node' WALLET_TEST=1 "$base"/demo.sh "$@"
+WALLET_EXE_NAME='cardano-node' WALLET_TEST=1 "$base"/demo.sh "$@"
--- a/wallet-new/src/Cardano/Wallet/API/V1/Handlers/Addresses.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Handlers/Addresses.hs
@@ -7,19 +7,19 @@ import           Universum
 
 import           Servant
 
-import           Cardano.Wallet.WalletLayer.Types (ActiveWalletLayer,
-                     PassiveWalletLayer (..), walletPassiveLayer)
+import           Cardano.Wallet.WalletLayer.Types (PassiveWalletLayer (..))
 
 import           Cardano.Wallet.API.Request
 import           Cardano.Wallet.API.Response
 import qualified Cardano.Wallet.API.V1.Addresses as Addresses
 import           Cardano.Wallet.API.V1.Types
 
+
 import           Pos.Core.Mockable.Production (Production, runProduction)
 
-handlers :: ActiveWalletLayer Production -> ServerT Addresses.API Handler
+handlers :: PassiveWalletLayer Production -> ServerT Addresses.API Handler
 handlers w =  listAddresses
-         :<|> newAddress (walletPassiveLayer w)
+         :<|> newAddress w
          :<|> getAddress
 
 listAddresses :: RequestParams -> Handler (WalletResponse [WalletAddress])
--- a/wallet-new/src/Cardano/Wallet/Kernel/BIP39.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/BIP39.hs
@@ -1,0 +1,12 @@
+{-- | This module trivially re-exports everything from 'Pos.Util.Mnemonic',
+      in preparation for the final deprecation of 'wallet' code once the new
+      data layer will be delivered.
+--}
+
+module Cardano.Wallet.Kernel.BIP39 (
+    module LegacyCompat
+    ) where
+
+import           Pos.Util.Mnemonic as LegacyCompat
+
+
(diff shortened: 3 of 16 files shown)


wallet-new/src/Cardano/Wallet/Kernel/BIP39.hs
```diff
@@ -0,0 +1,12 @@
+{-- | This module trivially re-exports everything from 'Pos.Util.Mnemonic',
+      in preparation for the final deprecation of 'wallet' code once the new
+      data layer will be delivered.
+--}
+
+module Cardano.Wallet.Kernel.BIP39 (
+    module LegacyCompat
+    ) where
+
+import           Pos.Util.Mnemonic as LegacyCompat
```
adinapoli-iohk: @KtorZ A potential strategy we might adopt regarding our previous conversation on how to identify modules which needs migration.

Here I have "wrapped" the legacy stuff in a presentable module name which is re-export everything. If we follow the same pattern, it should be easy to search for "LegacyCompat" and gradually replace things up until we can decommission the `wallet` project 😀

wallet-new/test/unit/Wallet/Inductive/Cardano.hs
```diff
@@ -159,18 +165,29 @@ equivalentT activeWallet esk = \mkWallet w ->
...
+                                                 False
+                                                 walletName
+                                                 assuranceLevel
+                                                 esk
+                                                 utxo
+        case res of
+             Left e -> createWalletErr (STB e)
+             Right hdRoot -> do
+                 let keystore = passiveWallet ^. Internal.walletKeystore
+                 liftIO $ Keystore.insert (WalletIdHdRnd $ hdRoot ^. HD.hdRootId) esk keystore
```
adinapoli-iohk: Fixing this test highlighted something orthogonal but equally important: if, when applying a block, we discover that the wallet is in an inconsistent state (like in this case where we had a wallet in the acid-state but no corresponding key in the keystore) we should fail hard and not silently ignore the fact.

I have opened https://iohk.myjetbrains.com/youtrack/issue/CBR-341 to track this.

wallet-new/src/Cardano/Wallet/API/V1/Handlers/Addresses.hs
```diff
@@ -7,19 +7,19 @@ import           Universum
...
 import qualified Cardano.Wallet.API.V1.Addresses as Addresses
 import           Cardano.Wallet.API.V1.Types
 
+
 import           Pos.Core.Mockable.Production (Production, runProduction)
 
-handlers :: ActiveWalletLayer Production -> ServerT Addresses.API Handler
+handlers :: PassiveWalletLayer Production -> ServerT Addresses.API Handler
 handlers w =  listAddresses
-         :<|> newAddress (walletPassiveLayer w)
```
adinapoli-iohk: I have changed the type signature of these handlers to take only the `PassiveWalletLayer`, as they don't need anything else.

scripts/launch/demo-with-new-wallet-api.sh
```diff
@@ -2,4 +2,4 @@
 
 base=$(dirname "$0")
 
-WALLET_EXTRA_ARGS="--new-wallet" WALLET_EXE_NAME='cardano-node' WALLET_TEST=1 "$base"/demo.sh "$@"
+WALLET_EXE_NAME='cardano-node' WALLET_TEST=1 "$base"/demo.sh "$@"
```
adinapoli-iohk: For some reason having the `WALLET_EXTRA_ARGS` was causing the `--new-wallet` option to be added twice in the final launch script, causing the demo cluster not to run when invoked via `./scripts/launch/demo-with-new-wallet-api.sh`.

wallet-new/src/Cardano/Wallet/Kernel/Wallets.hs
```diff
@@ -0,0 +1,138 @@
...
+    let (_, esk) = safeDeterministicKeyGen (BIP39.mnemonicToSeed mnemonic) spendingPassword
+    -- STEP 2: Atomically generate the wallet and the initial internal structure in
+    -- an acid-state transaction.
+    let newRootId = eskToHdRootId esk
+    res <- createWalletHdRnd pw
+                             (spendingPassword /= emptyPassphrase)
+                             walletName
+                             assuranceLevel
+                             esk
+                             mempty
```
adinapoli-iohk: Add a comment here and in the type signature specifying the invariants of this function, i.e. that the `Mnemonic` shouldn't have been used anywhere else before, otherwise this won't work as it passes in the empty Utxo and thus `prefilterUtxo` won't work correctly.

wallet-new/test/unit/Test/Spec/CreateWallet.hs
```diff
@@ -0,0 +1,133 @@
...
+                                 return ()
+                             Left unexpectedErr ->
+                                 fail $ "expecting different failure than " <> show unexpectedErr
+                             Right _ -> fail "expecting wallet not to be created, but it was."
+
+        prop "supports Unicode characters" $ withMaxSuccess 1 $ do
+            monadicIO $ do
+                request <- genNewWalletRq
+                withLayer $ \layer _ -> do
+                    let w' = request { V1.newwalName = "İıÀļƒȑĕďŏŨƞįťŢęșťıİ" }
```
adinapoli-iohk: Let's also add some kanjis.

PR #3283, opened 2018-07-20
[CDEC-456] Remove usages of -fno-warn-orphans in lib

intricate:
## Description

This PR reunites orphan instances (where possible) in `cardano-sl`.

## Linked issue

[https://iohk.myjetbrains.com/youtrack/issue/CDEC-456](https://iohk.myjetbrains.com/youtrack/issue/CDEC-456)
## Type of change

## Developer checklist

## Testing checklist


lib/src/Pos/Communication/Message.hs
```diff
@@ -1,312 +0,0 @@
-{-# OPTIONS_GHC -fno-warn-orphans #-}
```
intricate: :fire:

lib/src/Pos/Recovery/Instance.hs
```diff
@@ -1,57 +0,0 @@
-{-# OPTIONS_GHC -fno-warn-orphans #-}
```
intricate: :wave:

networking/src/Node/Message/Class.hs
```diff
@@ -45,6 +50,22 @@ class Message m where
     default formatMessage :: F.Buildable m => m -> T.Text
     formatMessage = F.sformat F.build
 
+instance Message Void where
+    messageCode _ = 0
+    formatMessage _ = "Void"
+
+instance Message UpdateVote where
```
intricate: I was a little concerned about placing these particular `Message` instances in the module along with their typeclass definitions. I ended up having to do this, as opposed to how I typically couple the instances with their type definitions, because these types are from `core` and `core` cannot depend on `networking` since `networking` already depends on `core`.

If anyone has a better idea for how I should do this, please let me know!

ssc/cardano-sl-ssc.cabal
```diff
@@ -65,6 +65,7 @@ library
                      , cardano-sl-crypto
                      , cardano-sl-db
                      , cardano-sl-lrc
+                     , cardano-sl-networking
```
intricate: @erikd Is this a problematic dependency?

networking/src/Node/Message/Class.hs
intricate: I was a little concerned about placing these particular Message instances in the module along with the typeclass definition. I ended up having to do this, as opposed to how I typically couple the instances with their type definitions, because these types are from core and core cannot depend on networking since networking already depends on core.

If anyone has a better idea for how I should do this, please let me know!

ssc/cardano-sl-ssc.cabal
intricate: @erikd Is this a problematic dependency?

lib/src/Pos/Communication/Message.hs
intricate: :fire:

infra/src/Pos/Infra/Recovery/Info.hs
```diff
@@ -76,16 +85,61 @@ class Monad m => MonadRecoveryInfo m where
     -- are enumerated in 'SyncStatus'.
     getSyncStatus :: SlotCount -> m SyncStatus
 
+type MonadRecoveryInfoConstraints ctx m =
```
intricate: @erikd @parsonsmatt @mhuesch Thoughts on this? The need for this came about as the result of reuniting the `MonadRecoveryInfo` instance below (line 97). These constraints still need to be satisfied for functions which previously had a constraint of `MonadRecoveryInfo` (from when it was an orphan). I wasn't too sure how else I should handle this.

ssc/cardano-sl-ssc.cabal
mhuesch: @intricate Yeah, I think that'll be a problem. We want `networking` up by `infra` & `lib` so we can combine/reshuffle the "network" functionality among them.

infra/src/Pos/Infra/Recovery/Info.hs
erikd: Wow, this is kind of interesting. `MonadRecoveryInfo` is a typeclass that has a single method and a single implementation. In functional programming, we call this a *function!*

I think the correct solution here is to:
* Remove `MonadRecoveryInfoConstraints` (its un-needed)
* Change `MonadRecoveryInfo` into a set of type constraints.
* Turn `getSyncStatus` into a function.

The second part might be something like:
```
type MonadRecoveryInfo m =
         ( Monad m
         , MonadIO m
         , MonadDBRead m
         , MonadSlots ctx m
         , MonadReader ctx m
         , HasLens RecoveryHeaderTag ctx RecoveryHeader
         , HasProtocolConstants
         )
```
Does that make sense?

erikd: Also, everywhere that currently has `MonadRecoveryInfo m` in  a constraint now needs `MonadRecoveryInfo ctx m`.

erikd: Also, somewhere I think you added a comment about `Mockable` which @mhuesch removed in the last couple of days :)

mhuesch:
LGTM, besides the merge conflicts. I'll let @erikd have the final say since he requested the changes.

infra/src/Pos/Infra/Recovery/Types.hs
```diff
@@ -0,0 +1,13 @@
...
+       ) where
+
+import           Control.Concurrent.STM (TMVar)
+
+import           Pos.Core.Block (BlockHeader)
+import           Pos.Infra.Communication.Types.Protocol (NodeId)
+
+data RecoveryHeaderTag
+
+type RecoveryHeader = TMVar (NodeId, BlockHeader)
```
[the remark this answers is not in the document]
erikd: Ah now I see. All good :+1: !

PR #3284, opened 2018-07-21
Change the hard-coded size estimates to use Bi.

matt-noonan:
## Description

This PR takes out the hard-coded transaction size estimates in the new wallet , replacing them with uses of the new size-expression code in `Bi`. 

## Linked issue

CBR-291, perhaps?

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
+++ b/wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
@@ -22,6 +22,7 @@ module Cardano.Wallet.Kernel.CoinSelection.FromGeneric (
   , dummyTxAttrSize
     -- * Estimating transaction limits
   , estimateMaxTxInputs
+  , estimateHardMaxTxInputs
     -- * Testing & internal use only
   , estimateSize
   ) where
@@ -31,18 +32,19 @@ import           Universum hiding (Sum (..))
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
-import           Data.Typeable (typeRep)
+import           Data.Typeable (TypeRep, typeRep)
 
 import           Pos.Binary.Class (LengthOf, Range (..), SizeOverride (..),
                      encode, szSimplify, szWithCtx, toLazyByteString)
 import qualified Pos.Chain.Txp as Core
 import qualified Pos.Client.Txp.Util as Core
-import           Pos.Core (AddrAttributes, Coin (..), TxAux, TxIn, TxInWitness,
-                     TxOut, TxSigData, TxSizeLinear, calculateTxSizeLinear)
+import           Pos.Core (AddrAttributes, Coin (..), TxSizeLinear,
+                     calculateTxSizeLinear)
 import qualified Pos.Core as Core
+import           Pos.Core.Attributes (Attributes)
+import           Pos.Core.Txp (TxAux, TxIn, TxInWitness, TxOut, TxSigData)
 import           Pos.Crypto (Signature)
 import qualified Pos.Crypto as Core
-import           Pos.Data.Attributes (Attributes)
 import qualified Pos.Txp as Core
 import           Serokell.Data.Memory.Units (Byte, toBytes)
 
@@ -371,7 +373,7 @@ estimateSize :: Byte     -- ^ Average size of @Attributes AddrAttributes@.
              -> Byte     -- ^ Size of transaction's @Attributes ()@.
              -> Int      -- ^ Number of inputs to the transaction.
              -> [Word64] -- ^ Coin value of each output to the transaction.
-             -> Byte     -- ^ Estimated size of the resulting transaction.
+             -> Range Byte -- ^ Estimated size bounds of the resulting transaction.
 estimateSize saa sta ins outs =
     -- This error case should not be possible unless the structure of `TxAux` changes
     -- in such a way that it depends on a new type with no override for `Bi`'s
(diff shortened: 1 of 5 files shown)


wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
```diff
@@ -145,9 +146,10 @@ newTransaction :: ActiveWallet
                -> IO (Either NewTransactionError TxAux, Utxo)
 newTransaction ActiveWallet{..} spendingPassword options accountId payees = do
 
-    -- | NOTE(adn) This number was computed out of the work Matt Noonan did
-    -- on the size estimation. See [CBR-318].
-    let maxInputs = 350
+    -- | NOTE(mn) 65536 is the current maximum transaction size, but this can change
+    --   over time. @newTransaction@ should be parameterized over this value, perhaps
+    --   adding it to @CoinSelectionOptions@.
+    let maxInputs = estimateMaxTxInputs dummyAddrAttrSize dummyTxAttrSize 65536
```
adinapoli-iohk: In theory, we should be able to read this value directly from the database -- have a look at `verifyGState` and follow the rabbit hole, i think they fetches the current "block meta params" directly from the rocksDB.

It should be relatively straightforward to hook in, and we can be sure we won't need to amend this code further -- leaving the magic constant here is tempting but dangerous, as during the next soft fork (if any) to bump the max Tx size, I'm pretty sure we would forget tweaking this number accordingly 😉

wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -408,4 +444,53 @@ estimateSize saa sta ins outs
...
+--
+--   We use a conservative over-estimate for the encoded transaction sizes, so the
+--   number of transaction inputs computed here is a lower bound on the true
+--   maximum.
+estimateMaxTxInputs
+  :: Byte -- ^ Size of @Attributes AddrAttributes@
+  -> Byte -- ^ Size of @Attributes ()@
+  -> Byte -- ^ Maximum size of a transaction
+  -> Word64
+estimateMaxTxInputs addrAttrSize txAttrSize maxSize = fromIntegral $
```
adinapoli-iohk: Do you think it's possible to write trivial unit tests for this function? It's used in a crucial part of the new wallet and is not a one-liner, which might suggest testing it out is a good thing to do 😉 

Also, minor bikeshed -- I'm wondering whether we should rehome these calculations into a more specific module, leaving `FromGeneric` free from this "fee internals", and re-export in this module only the top-level functions.

I would also curious to hear what @edsko thinks 😉

wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -408,4 +444,53 @@ estimateSize saa sta ins outs
...
+    txins0 = fromIntegral $ maxSize `div` 200
+    estSize txins = estimateSize addrAttrSize txAttrSize txins [Core.maxCoinVal]
+    step0 = 1 + (txins0 `div` 10)
+
+    bsearchUp f x y step =
+      if | f x < y -> bsearchUp f (x + step) y step
+         | f x == y -> x
+         | otherwise -> if step > 1
+                       then let step' = step `div` 2 in bsearchDown f (x - step') y step'
+                       else x - 1
```
parsonsmatt: Rather than `MultiWayIf`, why not write as a regular ol guard?

```
bsearchUp f x y step
  | f x < y = bsearchUp f (x + step) y step
  | f x == y = x
  | otherwise = ...
```

wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
```diff
@@ -377,27 +388,52 @@ largestFirst opts maxInps =
...
-estimateSize saa sta ins outs
-    = fromBytes . fromIntegral $
-      5
-    + 42 * ins
-    + (11 + listSize (32 + (fromIntegral saa))) * length outs
-    + sum (map intSize outs)
-    + fromIntegral sta
+estimateSize saa sta ins outs =
+    case szSimplify (szWithCtx (Map.fromList ctx) (Proxy @TxAux)) of
+        Left  sz    -> error ("Size estimate failed to simplify: " <> pretty sz)
```
parsonsmatt: Since this can give an error, I'd *really* prefer to see this whole function return `Either` as well. An `unsafeEstimateSize` that called `error` can be used for backwards compatibility.

matt-noonan: This is actually a little weird, and I'd love to figure out the best approach. It's true that this function could return `Left`, but only if the structure of a `TxAux` changes in such a way that the symbolic formula can't simplify to a value (say, because it suddenly starts depending on the encoded size of some type `T` that we haven't handled yet).

And there is a test in `core` that ensures the expression for `TxAux` really does reduce to a range value, and the range encloses the true encoded size: https://github.com/input-output-hk/cardano-sl/blob/develop/core/test/Test/Pos/Core/Bi.hs#L1809

So in one sense it doesn't make sense to bubble up an error, because the linked test is supposed to catch any change to `TxAux` that could cause a `Left` here. But on the other hand, this connection *does* feel slightly more tenuous than I'd have liked...

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
matt-noonan: That seems reasonable, I'll go down this rabbit hole when I get back on Wednesday morning.

wallet-new/src/Cardano/Wallet/Kernel/CoinSelection/FromGeneric.hs
parsonsmatt: Would you be willing to put this justification in the file as a comment? I'm satisfied that it's probably safe now.

wallet-new/src/Cardano/Wallet/Kernel/Transactions.hs
matt-noonan: As discussed on slack, this will involve some refactoring of `ActiveWallet` to remove the max tx size constant. I'll do this work on a separate PR to keep it logically separated.

edsko: Don't worry about this for now, this is actually part of a larger problem (CBR-250) which I will take a look at.

PR #3290, opened 2018-07-23
[CO-335] Implement legacy functionality

parsonsmatt:
## Description

This PR exposes the internal update API using the legacy handlers.

## Linked issue

https://iohk.myjetbrains.com/youtrack/issue/CO-335

## Type of change

## Developer checklist

## Testing checklist

### DIFF
--- a/wallet-new/src/Cardano/Wallet/API/V1/Internal/Update.hs
+++ b/wallet-new/src/Cardano/Wallet/API/V1/Internal/Update.hs
@@ -1,0 +1,13 @@
+module Cardano.Wallet.API.V1.Internal.Update where
+
+import           Servant
+
+import           Cardano.Wallet.API.Response (ValidJSON)
+
+type API =
+    "update"
+    :> ( "apply"
+        :> Post '[ValidJSON] NoContent
+    :<|> "postpone"
+        :> Post '[ValidJSON] NoContent
+    )
--- a/wallet-new/cardano-sl-wallet-new.cabal
+++ b/wallet-new/cardano-sl-wallet-new.cabal
@@ -46,10 +46,12 @@ library
                        Cardano.Wallet.API.V1.Handlers.Transactions
                        Cardano.Wallet.API.V1.Info
                        Cardano.Wallet.API.V1.Internal
+                       Cardano.Wallet.API.V1.Internal.Update
                        Cardano.Wallet.API.V1.LegacyHandlers
                        Cardano.Wallet.API.V1.LegacyHandlers.Accounts
                        Cardano.Wallet.API.V1.LegacyHandlers.Addresses
                        Cardano.Wallet.API.V1.LegacyHandlers.Info
+                       Cardano.Wallet.API.V1.LegacyHandlers.Internal.Update
                        Cardano.Wallet.API.V1.LegacyHandlers.Settings
                        Cardano.Wallet.API.V1.LegacyHandlers.Transactions
                        Cardano.Wallet.API.V1.LegacyHandlers.Wallets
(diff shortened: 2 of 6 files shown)


wallet-new/src/Cardano/Wallet/API/V1/Internal/Update.hs
```diff
@@ -0,0 +1,13 @@
...
+
+import           Cardano.Wallet.API.Response (ValidJSON)
+
+type API =
+    "update"
+    :> ( "apply"
+        :> Post '[ValidJSON] NoContent
+    :<|> "postpone"
+        :> Post '[ValidJSON] NoContent
+    )
```
KtorZ: Very small remark but, I am rather _annoyed_ by these endpoints trying to play it RESTful while they are completely not. I'd rather have plain:

- `api/internal/apply-update`
- `api/internal/postpone-update`

rather than the old nesting `/update/apply`. What is the `update` resource exactly :p ? It's very minor, yet I believe we could be very honest about what those endpoint are, and they aren't RESTful.
You can probably leave it as it is now, I'll make a final "cosmetic" PR with a few stuff like that we can discuss once done.

adinapoli-iohk:
> Though, I recall @adinapoli-iohk mentioning it wouldn't be "that" straightforward to migrate these endpoints. Do you remember what you had in mind exactly?

Yes and no, @KtorZ , hehe. Although what @parsonsmatt got there I bet is correct (haven't reviewed the PR) it will work up until we start tackling the Shelley-specific parts, namely the delegation centre (where each update can be voted etc).

I have written in the past a few proposals on how to overhaul the entire update mechanism, but I suspect that for now we should go with the path of least resistance and have the V1 endpoints implement the old, reliable behaviour so that Daedalus can rely on it, and plan for V2 the proper overhaul, which will need to be done with the Shelly requirements in mind 😉 

Thoughts?

wallet-new/src/Cardano/Wallet/API/V1/Internal/Update.hs
adinapoli-iohk: It's a bit of a bittersweet feeling indeed -- in my [update overhaul](https://github.com/input-output-hk/cardano-sl/pull/1884) proposal I have tried to make these all RESTful but unfortunately in the current state of things these are not. We are paying the "original sin" of the old API, in a sense. These are "imperative" endpoints to begin with, so trying to mask 'em as RESTful is probably going to be possible for clunky.

Perhaps you are right, we should be explicit here (my only consolation is that this in an internal API, hehe) and do things properly for the upcoming decentralisation.

parsonsmatt: @adinapoli-iohk Do you think it's worthwhile to just continue using V0 update API until the new stuff is ready to go?

parsonsmatt:
In light of the proposal to overhaul the updates, and since this is literally just calling the old code in a new API, I think it might make sense to table this and just have V0 API called for updates for now. Thoughts?

adinapoli-iohk:
@parsonsmatt I _think_ we do want to migrate Daedalus to use nothing but the V1 API, so I think what you did here is good -- porting the old behaviour using the new API. If anything, we gain the fact we can seriously start thinking about deprecating V0 as soon as our major user (i.e. Daedalus) switches to V1.

I would say 👍 to merge this PR albeit we know it's just an interim step in the right direction. What does @KtorZ think?

KtorZ:
@adinapoli-iohk For me, it seems that almost the whole V1 as we know it is an interim step waiting for the new data-layer. Most handlers rely on legacy code and are screaming for a good redesign / refactoring. 
So, these update endpoints are just adding a small rock to the pile we already have ^^" ... Then one by one, we are going to give them some fresh air.

PR #3293, opened 2018-07-24
start wiring cross-compile into hydra

cleverca22:


scripts/launch/connect-to-cluster/default.nix
```diff
@@ -18,6 +18,7 @@
 , disableClientAuth ? false
 , extraParams ? ""
 , iohkPkgs
+, gitrev # workarround for a minor bug
```
angerman: Could you elaborate on that "minor bug"? I'd just like to understand what I missed.

release.nix
```diff
@@ -1,7 +1,7 @@
 let
   fixedNixpkgs = (import ./lib.nix).fetchNixPkgs;
 in
-  { supportedSystems ? [ "x86_64-linux" "x86_64-darwin" ]
+  { supportedSystems ? [ "x86_64-linux" "x86_64-darwin" "win64" ]
```
angerman: Maybe we want to be more explicit here in the future? To specify that "win64" is only a `supportedTarget`?
Lumping it together with the other `supportedSystems`, might suggest (and confuse) that it's built on "win64"?

lib.nix
```diff
@@ -1,24 +1,32 @@
 let
+  todo = builtins.fromJSON (builtins.readFile ./nixpkgs-src.json);
+  fixme = { account, project, rev, sha256, sha256unpacked }:
+    pkgs.fetchFromGitHub {
+      owner = account;
+      repo = project;
```
angerman: I guess, we could just call them `owner`, and `repo` in the json file, and be done with `inherit` them, or even just calling the function with the map.

scripts/launch/connect-to-cluster/default.nix
cleverca22: my new `mapOnTest'` function passes gitrev to everything, including the new connect scripts

